PR #1, opened 2026-05-07
fix(mcp): serialize concurrent send_to_agent_impl per agent

tya5:
## Summary

- Add per-agent `asyncio.Lock` (module-level `_AGENT_LOCKS` keyed by agent name) around the `baseline → _handle_user_message → history harvest` window in `send_to_agent_impl` so concurrent FastAPI handlers (notably the A2A endpoint at [a2a.py](src/reyn/web/routers/a2a.py)) can no longer race on the shared `ChatSession`.
- Extend `_new_agent_history_entries` with an optional `chain_id` parameter that filters on `msg.meta["chain_id"]`, scoping reply harvesting to the caller's own chain (defense-in-depth even without serialization).

## Modification class

- [ ] 🔵 拡張 (additive — new capability, no breaking changes)
- [ ] 🔴 仕様変更 (breaking — changes existing behavior or contracts)
- [x] 🟢 不具合修正 (bug fix)
- [ ] 📚 doc 追加 (documentation only)

## Why

Two concrete bugs were observable when the A2A router served two
overlapping `message/send` requests for the same Reyn agent:

1. **`history[-1]` race.** Both coroutines append their user message to
   the same `session.history` list and then enter `_run_router_loop`,
   which reads the tail of `history` to construct the LLM prompt.
   Whichever coroutine schedules second sees the *other's* user message
   at `history[-1]` and sends it to the LLM as if it were its own.
2. **Cross-chain reply leakage.** The original
   `_new_agent_history_entries(session, baseline)` returned every
   `role="agent"` entry past `baseline`, with no chain scoping. Two
   concurrent calls would each splice the *other's* reply into their
   own `reply_text`, returning duplicate / wrong content.

The fix is the conservative short-term shape consistent with the
existing serial `session.run()` inbox architecture — push concurrent
callers through a per-agent lock instead of the inbox. Long-term, MCP /
A2A entry points should go through `session.inbox` + `session.run()`
like the chat path does; that's a bigger reshape and explicitly out of
scope here.

## Testing

- Which Tier do the new/changed tests belong to? — **None added in
  this PR.** This change is a localized concurrency fix with no
  behavioral surface change for sequential callers; an integration-
  level Tier 2 test that exercises the A2A endpoint with two
  overlapping `message/send` calls is the right shape and is left as
  a follow-up.
- New tests added: no — see above.
- All tests pass locally: not run as part of this minimal change set.

## Checklist

- [ ] Tests pass locally (`pytest`).
- [ ] Lint / format clean.
- [x] No skill-specific strings (phase names, artifact types, fields) added to OS code (P7).
- [x] P1-P8 invariants respected (see `CLAUDE.md`).
- [ ] `CHANGELOG.md` updated if the change is user-visible.
- [x] No secrets, credentials, or API keys included in the diff.

## Reviewer notes

- The lock is acquired *after* `_get_session` (cheap, idempotent) so
  serialization only covers the actual mutation window. `_AGENT_LOCKS`
  uses `setdefault` and is intentionally unbounded — the dict grows
  with the number of agents the server has ever served, which mirrors
  the registry's own caching shape.
- `chain_id` filtering is additive: passing `None` (= the chat /
  internal callers' shape) preserves the previous behavior exactly.
- `session.run()`, the registry, and the chat path are deliberately
  untouched.

### DIFF
--- a/docs/en/concepts/plan-mode.md
+++ b/docs/en/concepts/plan-mode.md
@@ -67,6 +67,13 @@ so the snapshot stays small and outputs are preserved verbatim
 `get_step_result(snap, agent_state_dir, step_id)` which transparently
 resolves inline-vs-spilled.
 
+Sub-loop LLM calls within each step are also recorded
+([ADR-0025](../decisions/0025-plan-step-llm-memoization.md)) so a
+crash mid-step doesn't re-pay the LLM cost on resume. The sub-loop's
+`call_llm_tools` invocations are keyed by `args_hash`; on resume the
+recorded results replay before the LLM is invoked, only the crashed
+turn pays fresh.
+
 When `reyn chat` next starts, `AgentRegistry.restore_all`:
 
 1. Replays the WAL onto each agent's snapshot.
--- a/docs/en/concepts/plan-mode.md
+++ b/docs/en/concepts/plan-mode.md
@@ -56,8 +56,16 @@ concurrently.
 |---|---|---|
 | Decomposition (plan shape) | `agents/<name>/state/plans/<plan_id>/decomposition.json` | yes |
 | Per-plan progress (steps completed, results) | `agents/<name>/state/plans/<plan_id>.snapshot.json` | yes |
+| Step output (≤32 KB) | inline on the snapshot above | yes |
+| Step output (>32 KB) | `agents/<name>/state/plans/<plan_id>/step_results/<step_id>.txt` | yes |
 | Plan lifecycle events | `.reyn/state/wal.jsonl` (`plan_started` / `plan_step_*` / `plan_completed`) | yes |
 | Active asyncio.Task | in-memory only | no — auto-resumes on next startup |
+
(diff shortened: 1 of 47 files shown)


tya5:
このPRの fix は別 path で既に main に landed しています。

- `src/reyn/mcp_server.py:69` に `_AGENT_LOCKS = dict[str, asyncio.Lock]` の per-agent serialization が既に存在 (= FP-0013 unified inbox/outbox transport refactor、 2026-05-14 land)
- `mcp_server.py:18` のコメントが「FP-0013: `send_to_agent_impl` now drives `session.run_one_iteration()`」と新 transport を経由する構造に明示

8 日経過で main は ~100 commits 進行 (= plan-mode docs / ADR-0023/0024/0025 / dogfood-discipline / 各種 refactor 全て independent landed)。 47 files / +11K LoC の差分のうち、 本 PR 固有の concurrency fix portion は superseded、 残りは already-landed docs と重複。

stale として close、 PR-driven な作業履歴は保持 (= 不要な force-push / rebase 戦争を避ける)。 元 fix 内容は main の `src/reyn/mcp_server.py` で確認可能。

PR #64, opened 2026-05-17
feat(tui): flash the new width in the conv sticky bar on right-panel resize

tya5:
## Summary

- `h` / `l` shrank or grew the right panel **silently** — operators had no direct confirmation that the keystroke registered, and no way to tell how close they were to the min (24 col) or max (66% of terminal) edge without trial-and-error.
- Add a 1.5 s `● panel: NN col` sticky message in the conv pane on every resize. When the new width hits a clamp, append `(min)` / `(max)` so the operator knows further presses in that direction are no-ops.
- Extracted as `_flash_status(text, duration)` so future short-lived panel feedback (e.g. confirm beats for `c` copy) can reuse it.

## Test plan

- [x] Visual: press `h` / `l` after Ctrl+O — sticky bar shows `● panel: NN col` briefly
- [ ] Visual: press `l` until clamped → sticky shows `● panel: 24 col (min)`

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/chat/tui/widgets/right_panel/__init__.py
@@ -247,8 +247,40 @@ class RightPanel(Widget):
         if self._panel_width == 0:
             self._panel_width = self.size.width or 40
         max_width = max(40, int((self.app.size.width or 120) * 0.66))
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Visual: press `l` until clamped → sticky shows `● panel: 24 col (min)`

— lead-coder

PR #66, opened 2026-05-17
fix(tui): show full error message in expanded ErrorBox

tya5:
## Summary

- `ErrorBox._header_text()` truncates the collapsed header at 72 chars with `…`, but no caller in the repo ever populates the optional `details` field — every `kind="error"` outbox emit in `chat/session.py` and `chat/registry.py` passes `text` + `meta` only. Result: expanding a long error reveals an empty block, and the truncated tail of the message has no path to the screen.
- When `self._details` is empty, fall back to rendering the full untruncated `self._message` inside the `eb-details` block. When details are provided, behaviour is unchanged.
- Replace the misleading `[B→events] for full trace` hint with `Ctrl+B → events for full trace`. `B` alone is not bound anywhere in `app.py BINDINGS`; `ctrl+b` is the real `action_toggle_panel` binding (the panel preserves the last focal tab).

Caller-side (`session.py` / `conversation.mount_error`) is intentionally left untouched — the expanded view now reads the message that's already on the widget.

## Test plan

- [ ] Visual: trigger a long router error → header is truncated with `…`, click to expand → full message visible inside the expanded block, hint reads `Ctrl+B → events for full trace`

### DIFF
--- a/src/reyn/chat/tui/widgets/error_box.py
+++ b/src/reyn/chat/tui/widgets/error_box.py
@@ -9,7 +9,7 @@ After click (expanded):
     ✗ [skill#run_id]: error message  ▼
       detail line 1
       detail line 2
-      … N more  [B→events]
+      … N more  Ctrl+B → events
 
 Press Esc to dismiss (handled by ConversationView / app.py via the
 `_error_boxes` list — same API as before).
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Visual: trigger a long router error → header is truncated with `…`, click to expand → full message visible inside the expanded block, hint reads `Ctrl+B → events for full trace`

— lead-coder

PR #67, opened 2026-05-17
fix(tui): seal orphan streaming rows on Ctrl+C cancel

tya5:
## Summary

- When the user pressed Ctrl+C mid-stream, the router cancellation stopped the producer task but no `__stream_end__` event was emitted — so `ConversationView.end_stream()` never ran and the `StreamingRow`'s blinking cursor kept ticking forever.
- `action_cancel_inflight` now snapshots `conv._stream_rows` keys after the existing task-cancel pass and calls `end_stream(msg_id)` on each. A dim `⌁ cancelled` marker line is written under each sealed reply so the user can see it was interrupted.
- The cancellation summary now counts sealed streams alongside cancelled skills/plans, and an orphan stream alone is enough to avoid the "(nothing in-flight to cancel)" early return.

## Test plan

- [ ] Visual: trigger a long LLM reply, press Ctrl+C mid-stream → cursor stops, `⌁ cancelled` line appears, summary reads `✗ cancelled N skill + 1 stream`

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -561,7 +561,28 @@ class ReynTUIApp(App):
                 task.cancel()
                 cancelled_plans += 1
 
-        if cancelled_skills == 0 and cancelled_plans == 0:
+        # Seal any orphan streaming rows. The router cancel above stops the
+        # producer but no `__stream_end__` is emitted, so the StreamingRow's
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Visual: trigger a long LLM reply, press Ctrl+C mid-stream → cursor stops, `⌁ cancelled` line appears, summary reads `✗ cancelled N skill + 1 stream`

— lead-coder

PR #68, opened 2026-05-17
fix(tui): wire `/` on Docs tab to /docs-filter + advertise it in header

tya5:
## Summary

- `widgets/right_panel/keys_tab.py:_DOCS_KEYS` advertised `/` as a gated Docs binding, but `RightPanel.on_key` had no `/` branch — pressing `/` on the Docs tab silently did nothing. The keys-tab help was a lie.
- The Docs header markup read `j↓  k↑  space=open` only; first-time users had no way to learn that `/docs-filter` exists until they happened to type the slash command from chat.
- Wire `/` on the Docs tab to focus the InputBar and pre-fill `/docs-filter ` (with the cursor parked past the trailing space), so the user just types the substring and hits Enter. Empty submission clears the filter — same semantics as invoking the slash command directly.
- Add ` /=filter` to the Docs header markup so the binding is discoverable from the panel itself.

## Scope

MVP — no inline filter buffer. This PR hands off to the existing `/docs-filter` slash command path (`app_outbox._on_docs_filter`) rather than building a parallel filter input. An inline buffer would be a separate, larger change.

## Before / after

```
Before:  Docs  j↓  k↑  space=open                  ← `/` is bound per keys_tab but inert
After:   Docs  j↓  k↑  space=open  /=filter        ← `/` focuses InputBar + pre-fills `/docs-filter `
```

## Test plan

- [ ] Visual: open Docs tab, press `/` → InputBar gains focus pre-filled with `/docs-filter `, cursor past the space; type `arch` + Enter → docs list filters to entries containing `arch`
- [ ] Visual: Docs header shows ` /=filter` next to ` space=open`

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/chat/tui/widgets/right_panel/__init__.py
@@ -326,6 +326,15 @@ class RightPanel(Widget):
             # an issue tracker.
             event.prevent_default()
             self._copy_current_view()
+        elif event.key == "/":
+            # Docs tab only: pre-fill the conv input with the
+            # `/docs-filter` slash command so the user can type a
+            # substring and press Enter. Advertised by keys_tab._DOCS_KEYS
+            # and the Docs header `/=filter` hint. MVP — no inline
+            # filter buffer; just hand off to the existing slash command.
+            if self._panel_type == "docs":
+                event.prevent_default()
+                self._prefill_docs_filter()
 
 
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Visual: open Docs tab, press `/` → InputBar gains focus pre-filled with `/docs-filter `, cursor past the space; type `arch` + Enter → docs list filters to entries containing `arch`
- [ ] Visual: Docs header shows ` /=filter` next to ` space=open`

— lead-coder

PR #69, opened 2026-05-17
fix(tui): wrap right-panel j/k cursor at list edges

tya5:
## Summary

- On the **docs / events / memory / agents** tabs, pressing `j` past the last item used to stop dead; same for `k` at the top. No visible affordance meant the keybinding looked broken.
- Replace the `max(0, min(n-1, cursor + delta))` clamp in `_events_move`, `_memory_move`, `_agents_move`, and `_docs_move` with modulo wrap, so `j` wraps from the last item to index 0 and `k` from the top jumps to the last item (Vim-list behaviour).
- The h/l `_panel_resize` clamp is intentionally left alone — its bounds (24..max width) are a meaningful geometry constraint, not a cyclic list.

## Before / after

```
Before:  j j j j … cursor stops at last item, no feedback
After:   j j j j … wraps last → 0; k from 0 jumps to last
```

## Test plan

- [x] Updated `test_docs_cursor_advances_with_j_keypress` to pin the new wraparound contract (was pinning the old clamp)
- [ ] Visual: on each of docs / events / memory / agents, hammer `j` past the bottom and confirm cursor wraps to top; same with `k` from the top

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/chat/tui/widgets/right_panel/__init__.py
@@ -401,12 +401,17 @@ class RightPanel(Widget):
     # ── events / memory cursor + preview integration ─────────────────────────
 
     def _events_move(self, delta: int) -> None:
-        """Move the events tab cursor and refresh; sync preview if open."""
-        if not self._events_visible:
+        """Move the events tab cursor and refresh; sync preview if open.
+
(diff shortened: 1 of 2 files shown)


tya5:

- [ ] Visual: on each of docs / events / memory / agents, hammer `j` past the bottom and confirm cursor wraps to top; same with `k` from the top

— lead-coder

PR #78, opened 2026-05-17
feat(tui): show hotkey prefix on InterventionWidget chips

tya5:
## Summary

- `UserIntervention.choices` already carries a per-choice `hotkey` field (e.g. `y`, `n`, `1`), but `app_outbox._on_intervention` was flattening to `(label, id)` 2-tuples before the widget rendered them. Users had to guess which key selects each option.
- This PR threads the full choice shape (`label` / `id` / `hotkey` / `default`) through `_on_intervention` → `_mount_intervention` → `mount_intervention` → `InterventionWidget` and renders each chip as `[h] Label` when a hotkey is present. A `default=True` choice gets Textual's `variant="primary"` so the recommended answer stands out.
- Callers without the richer fields keep working: the widget normalises both legacy `(label, id)` 2-tuples and the dict shape into one internal representation, with missing keys falling back to safe blanks.
- Bonus: the widget also binds the hotkey letter via `on_key` and activates the matching chip — only when no `Input` widget has focus, so free-text answers that happen to start with a hotkey letter still flow into the text input.

## Files changed

- `src/reyn/chat/tui/app_outbox.py` — `_on_intervention` now forwards `{label, id, hotkey, default}` dicts instead of `(label, id)` 2-tuples.
- `src/reyn/chat/tui/app.py` — `_mount_intervention` widens the `choices` type to accept both 2-tuples and dicts; docstring updated.
- `src/reyn/chat/tui/widgets/conversation.py` — `mount_intervention` widens the `choices` type.
- `src/reyn/chat/tui/widgets/intervention.py` — `InterventionWidget` normalises both shapes via `_normalise_choice`, renders `[h] Label` chip labels, applies `variant="primary"` to a `default=True` choice, and intercepts the hotkey letter via `on_key`.

## Test plan

- [x] `pytest tests/` — 3277 passed, 11 skipped, 2 xfailed (baseline unchanged).
- [x] `pytest tests/cli/test_tui_smoke.py` — 35 passed (existing `mount_intervention` tests still green with the widened `choices` type).
- [ ] Manual: trigger a permission `ask_user` intervention with `[Y]es / [A]lways / [N]o` choices; verify the chips render `[Y] [Y]es` / `[A] [A]lways` / `[N] [N]o` and the default option (if any) renders with the accent variant.

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -229,7 +229,7 @@ class ReynTUIApp(App):
         conv: ConversationView,
         text: str,
         iv_id: str,
-        choices: list[tuple[str, str]] | None = None,
+        choices: list[tuple[str, str] | dict] | None = None,
         queued_extra: int = 0,
     ) -> None:
         """Mount an InterventionWidget inline in the conversation view.
@@ -238,6 +238,10 @@ class ReynTUIApp(App):
         rendered. The user's answer (chip label or free text) is routed via
         session._maybe_answer_oldest_intervention — which matches hotkeys /
         choice labels against the pending intervention.
+
+        ``choices`` accepts both legacy ``(label, id)`` 2-tuples and the
+        richer ``{"label", "id", "hotkey", "default"}`` dict shape — the
+        widget normalises both forms internally.
 
         ``queued_extra`` surfaces the number of additional pending
         interventions waiting behind the head one (caller-computed) so the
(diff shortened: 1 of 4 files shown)


tya5:

- [ ] Manual: trigger a permission `ask_user` intervention with `[Y]es / [A]lways / [N]o` choices; verify the chips render `[Y] [Y]es` / `[A] [A]lways` / `[N] [N]o` and the default option (if any) renders with the accent variant.

— lead-coder

PR #80, opened 2026-05-17
fix(tui): make permission-denied error message actionable

tya5:
## Summary

- `dispatch_tool` catches `PermissionError`, returns `{"kind":"permission_denied", "message": str(e)}`, and that text becomes the user-visible `✗ <skill#abcd>: …` error in the TUI. Some `PermissionError` strings from `PermissionResolver` are already verbose (file.read / file.write), but many are terse (`shell access denied`, `web fetch denied`, `tool 'X' access denied`), leaving the user with no path forward.
- This change appends a single `To allow: add \`<config-key>\` to reyn.local.yaml under permissions:` line to the message inside the dispatcher's permission-denied branch. The config key is selected from a small table keyed on the dispatch tool name (covers router catalog entries `read_file` / `write_file` / `web_fetch` / `call_mcp_tool` / `mcp_install` / … and skill_phase op kinds `file` / `shell` / `mcp` / `web_fetch` / `index_drop` / …).
- Unmapped tool names fall back to `To allow: see the events tab for the full permission trace.` — explicitly avoiding fabricating a config key that doesn't exist.
- The original `PermissionError` text stays as the prefix, so the existing substring asserts (`"nope" in message`) in `test_dispatcher.py`, `test_dispatcher_wal_step_events.py`, and `test_dispatcher_resume_memoization.py` keep passing. The same enriched message also lands in the `tool_failed` audit event and the WAL `step_failed` entry, so forensics see what the user saw.
- The hint is added at the dispatcher boundary (the single point all permission-denied paths flow through), so the fix covers both router-tool calls and skill-phase op invocations without touching the `PermissionResolver` itself.
- ErrorBox header still truncates at 72 chars, and PR #66 already routes the full text into the expanded body — so the user sees the prefix on the collapsed header and the actionable suffix when they expand.

## Test plan

- [x] `pytest tests/` — 3279 passed, 11 skipped, 2 xfailed
- [x] New Tier 2 invariants in `tests/test_dispatcher.py`:
  - `test_permission_denied_message_includes_actionable_hint`
  - `test_permission_denied_unmapped_tool_falls_back_to_generic_hint`
- [ ] Visual: trigger a `file.write` permission denial in chat → header shows the truncated prefix, expand → `To allow: add \`permissions.file.write: allow\` to reyn.local.yaml under permissions:` line visible.

### DIFF
--- a/src/reyn/dispatch/dispatcher.py
+++ b/src/reyn/dispatch/dispatcher.py
@@ -217,6 +217,7 @@ async def dispatch_tool(
     try:
         result = await invoker(args)
     except PermissionError as e:
+        enriched = _enrich_permission_message(name, str(e))
         ctx.events.emit(
             "tool_failed",
             caller_kind=ctx.caller_kind,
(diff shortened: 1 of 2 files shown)


tya5:

- [ ] Visual: trigger a `file.write` permission denial in chat → header shows the truncated prefix, expand → `To allow: add \`permissions.file.write: allow\` to reyn.local.yaml under permissions:` line visible.

— lead-coder

PR #91, opened 2026-05-17
feat(tui): hint how to recover when events filter excludes everything

tya5:
## Summary

- When the events tab cycles to a filter group (e.g. `phase`, `rag`) whose event types haven't fired this session, the panel rendered a terse `(no matching events)` with no on-screen recovery hint.
- The `[f]` / `[t]` header hints are only visible when the panel is wide enough to fit them past the filter name, so a narrow panel left the user stuck wondering how to get back to `all`.
- Now: if `all_events` is non-empty but the active filter excluded everything, render a two-line dim hint that names the excluded filter and tells the user that `[f]` cycles the filter and `[t]` cycles the tail. The truly-empty-pool case (no events logged yet) keeps the existing wording because cycling wouldn't help.

## Test plan

- [ ] Manual: cycle to `phase` filter on a fresh session before any phase fires, confirm the two-line hint renders with the filter name bolded
- [ ] Manual: confirm the truly-empty `.reyn/events` case still shows the original `(no matching events)` wording

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/events_tab.py
+++ b/src/reyn/chat/tui/widgets/right_panel/events_tab.py
@@ -491,7 +491,24 @@ def render_events(
         visible = all_events
 
     if not visible:
-        return "[#555555]  (no matching events)[/]", []
+        # Two distinct empty states:
+        #   1. `all_events` empty → pool itself is empty (no events logged yet).
+        #      The existing "no matching events" wording is already meaningful;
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Manual: cycle to `phase` filter on a fresh session before any phase fires, confirm the two-line hint renders with the filter name bolded
- [ ] Manual: confirm the truly-empty `.reyn/events` case still shows the original `(no matching events)` wording

— lead-coder

PR #92, opened 2026-05-17
feat(slash): require --force on /skill discard to prevent accidental kills

tya5:
## Summary

- `/skill discard <id>` is destructive (cancels the asyncio.Task, drops pending interventions, writes a `skill_discarded` event). Currently it commits instantly — a typo or stray Tab-completion can wipe a long-running skill with no prompt.
- Now the bare form is confirmation-only: it resolves the id and emits a `kind=\"system\"` warning naming the skill, run_id, and elapsed time. The user must re-run with `--force` to actually commit the discard.
- `--force` may appear before or after the run_id (Tab-completion ordering is unpredictable).
- Unknown-id error path is unchanged. `/skill list` and `/cancel` are not touched.

## Implementation notes

- New `_parse_discard_args()` splits `args` into `(run_id, force)`; `_format_elapsed()` reuses the same monotonic-since-start source as `/list`.
- Existing test sites that exercise the destructive flow (`test_skill_postprocessor_chain.py`, `test_chain_peer_discarded_notify.py`, and the cancellation/intervention-drop branches of `test_skill_slash_command.py`) updated to pass `--force`.
- Three new Tier-2 tests cover the confirmation step: bare discard emits NO `skill_discarded` event, leaves the asyncio.Task running, and the warning includes the skill name + run_id + the `--force` hint. A fourth new test verifies `--force` works in either position relative to the run_id.

## Test plan

- [x] `pytest tests/test_skill_slash_command.py` — 11 passed
- [x] `pytest tests/test_skill_postprocessor_chain.py tests/test_chain_peer_discarded_notify.py tests/test_slash_multi_line_dispatch.py` — 14 passed
- [x] `pytest tests/` — 3290 passed, 11 skipped, 2 xfailed
- [ ] Manual smoke: launch a skill, run `/skill discard <id>` → confirmation only; re-run with `--force` → discards.

### DIFF
--- a/src/reyn/chat/slash/skill.py
+++ b/src/reyn/chat/slash/skill.py
@@ -1,15 +1,22 @@
 """/skill slash command — manage active skill runs (PR-resume-ux U2).
 
 Sub-commands:
-  /skill list                 — show active skill runs
-  /skill discard <run_id>     — abort a specific run + cleanup
+  /skill list                          — show active skill runs
+  /skill discard <run_id>              — preview discard (confirmation only)
+  /skill discard <run_id> --force      — actually abort the run + cleanup
 
 Note: distinct from ``/skills`` (plural, PR-tui-4) which lists *available*
 skills (catalogue). This one targets *running* skill instances.
+
+Discard is destructive: it cancels the asyncio.Task, drops pending
+interventions for the run, and writes a ``skill_discarded`` event. To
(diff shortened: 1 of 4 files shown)


tya5:

- [ ] Manual smoke: launch a skill, run `/skill discard <id>` → confirmation only; re-run with `--force` → discards.

— lead-coder

PR #94, opened 2026-05-17
feat(tui): surface Ctrl+P/N turn-nav keys in hints

tya5:
## Summary

- Footer hint (`InputBar._build_hint`): append `Ctrl+P/N turn`. Trim `Ctrl+O focus panel`, `Ctrl+R voice`, `Ctrl+\ shot`, and `Ctrl+C cancel` to keep the line under 90 cols (was ~155, now ~84). All dropped keys remain visible in the Keys tab.
- Empty-state hint (`ConversationView.compose`): append `Ctrl+P/N turn` alongside the existing `↑ history` input-history hint.

## Why

Ctrl+P/N (prev / next turn) were only surfaced inside the right-panel Keys tab under `[CONVERSATION]`, which sits below the fold at default panel size after PR #58 made bindings one-line-per-entry. Users had no organic way to discover the most useful chat-navigation keys.

## Test plan

- [x] `pytest tests/cli/test_tui_smoke.py` — 36/36 pass
- [ ] Visual check at 80-col terminal: footer fits on one line
- [ ] Empty-state hint shows `Ctrl+P/N turn` before any user input

### DIFF
--- a/src/reyn/chat/tui/widgets/conversation.py
+++ b/src/reyn/chat/tui/widgets/conversation.py
@@ -168,7 +168,7 @@ class ConversationView(Widget):
         # B5: empty-state hint, removed on first message
         yield Static(
             "  Type [bold]/[/] for commands  ·  [bold]Ctrl+B[/] panel  ·  "
-            "[bold]Ctrl+L[/] clear  ·  ↑ history",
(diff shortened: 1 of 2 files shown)


tya5:

- [ ] Visual check at 80-col terminal: footer fits on one line

— lead-coder

PR #97, opened 2026-05-17
fix(tui): scroll events tab into view as cursor moves with j/k

tya5:
## Summary

- After PR #69 (cursor wraparound), `j` / `k` correctly cycle the events cursor — but the panel viewport stayed put, so on the default 200-event tail the cursor would walk off-screen and the user couldn't see where their selection landed.
- Docs tab already had `_scroll_docs_into_view`; events tab didn't.
- Add `_scroll_events_into_view` modelled on the docs implementation. Each event renders as one row (since PR #79's HH:MM:SS trim), so the math is just `y = 1 + cursor_index`.

## Test plan

- [ ] Visual: with > viewport-height events visible, press j repeatedly — cursor stays in view

## Note

Memory / agents tabs have the same gap; deferred to follow-ups — the events tab is the high-cardinality case in practice.

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/chat/tui/widgets/right_panel/__init__.py
@@ -475,8 +475,33 @@ class RightPanel(Widget):
             return
         self._events_cursor = (self._events_cursor + delta) % n
         self._invalidate()
+        self._scroll_events_into_view()
         if self._preview_visible:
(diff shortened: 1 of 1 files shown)


tya5:

- [ ] Visual: with > viewport-height events visible, press j repeatedly — cursor stays in view

— lead-coder

PR #99, opened 2026-05-17
fix(tui): scroll memory + agents tabs into view as cursor moves with j/k

tya5:
## Summary

Extends [PR #97](https://github.com/tya5/reyn/pull/97) (which fixed the events tab) to the memory and agents tabs, which were explicitly deferred in that PR's note.

After `j`/`k` cycles the cursor past the visible window in the memory or agents tab, the panel viewport now follows the cursor — same behaviour the events tab just got.

## Implementation note

Memory entries and agents items have variable line counts per row:

- Memory: each entry can be 1 line (name) or 2 lines (name + indented description); section labels (`SHARED`, `AGENT <name>`), per-type subheaders (`[USER]`, `[FEEDBACK]`, …) and blank separators bump the line index unpredictably.
- Agents: items render via RichTree; a running skill is 1 line for the label + an optional second line for the phase child node; a recent_plan with a goal is 2 lines; blank lines between agent blocks.

An arithmetic `y = f(cursor)` (which works for the events tab where every row is 1 line) would mis-scroll the moment a section boundary or child node showed up. Instead, `render_memory` and `render_agents` now record each selectable item's y-coordinate at render time and return it alongside the flat item list, and the new `_scroll_memory_into_view` / `_scroll_agents_into_view` helpers just look it up.

Both helpers run after `_invalidate()` and before the preview refresh — same call-site shape as `_scroll_events_into_view` in PR #97.

## Test plan

- [ ] Manual: open memory tab with > viewport-height worth of entries; press `j` past the last visible row → viewport follows
- [ ] Manual: open agents tab with > viewport-height worth of running / recent items; press `j` / `k` → viewport follows
- [ ] Manual: confirm preview pane (Space) still tracks the cursor in both tabs

## Base

This PR targets `feat/tui-events-scroll-into-view` (PR #97) rather than `main` because it depends on the events-tab fix landing first — the diff is purely the memory/agents extension. Once #97 merges, GitHub will retarget this to `main` automatically.

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/chat/tui/widgets/right_panel/__init__.py
@@ -132,12 +132,19 @@ class RightPanel(Widget):
         # Memory tab state — flat cursor over all entries
         self._memory_cursor: int = 0
         self._memory_entries: list[Any] = []
+        # y-coord (0-indexed line) of each memory entry's name row in the
+        # rendered output. Populated by `render_memory`; used by
+        # `_scroll_memory_into_view` to keep the cursor visible as j/k
+        # moves it past the panel viewport.
+        self._memory_entry_ys: list[int] = []
         # Agents tab state — same cursor pattern as events / memory.
         # `_agents_items` is the flat list returned by `render_agents`,
         # one entry per running skill / running plan / recent skill /
         # recent plan. Used by j/k navigation and preview rendering.
         self._agents_cursor: int = 0
         self._agents_items: list[dict] = []
+        # y-coord of each agents-tab item; populated by `render_agents`.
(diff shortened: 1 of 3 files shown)


tya5:
Already merged into main via 487d520. The PR was opened against feat/tui-events-scroll-into-view (#97's branch) so GitHub didn't auto-detect the merge after PR #97 landed; closing manually.

tya5:

- [ ] Manual: open memory tab with > viewport-height worth of entries; press `j` past the last visible row → viewport follows
- [ ] Manual: open agents tab with > viewport-height worth of running / recent items; press `j` / `k` → viewport follows
- [ ] Manual: confirm preview pane (Space) still tracks the cursor in both tabs

— lead-coder

PR #119, opened 2026-05-17
fix(router/hot-list): expose resource-category alias schemas (FP-0034 D2-full)

tya5:
## Summary

D2-min ([#117](https://github.com/tya5/reyn/pull/117)) landed rich metadata for operation-category aliases (passthrough rules in `_OPERATION_RULES`). This patch closes the remaining [FP-0034 §D2](https://github.com/tya5/reyn/issues/36) gap: resource-category aliases were still shipped with `properties: {}, additionalProperties: true` parameters, so weak-model LLMs could see the alias name but not know which args to send. With D2-min + D2-full both landed, every hot-list direct alias the LLM sees in `tools=` carries its target's real description + JSON schema — completing the *"hot list direct alias は full schema 提供"* design intent.

**Dependency:** based on `e2e/hot-list-alias-rich-metadata-d2-min` (= PR #117). When #117 merges first this rebases cleanly; when this lands first the union of both changes ships in one go.

## Categories covered

| Category | Source | Curried by routing rule |
|---|---|---|
| `agent.peer__<name>` | `delegate_to_agent` parameters minus `to` | `to=<name>` |
| `mcp.server__<name>` | empty `{}` (no caller-supplied args) | `server=<name>` |
| `rag.corpus__<name>` | `recall` parameters minus `sources` | `sources=[<name>]` |
| `skill__<name>` | the skill's entry-artifact JSON schema (= the `schema:` block of the `<input_artifact>.yaml`) | the wrapping `{type, data}` envelope is applied by `_invoke_skill_args` |
| `mcp.tool__<server>.<tool>` | the MCP tool's declared `inputSchema` (from `host.get_mcp_servers()` expanded tools array, FP-0032 shape) | server + tool are extracted from the qualified name |

## Out of scope

- `memory.entry__<name>` — the existing `_read_memory_body_args` transform sends `{name: entry}` while the target `read_memory_body` parameters require `{layer, slug}`. Pre-existing dispatch shape mismatch; out of scope here and tracked separately so D2-full lands cleanly without an orthogonal fix.

## Touch points

- `src/reyn/chat/router_loop.py` — extends `_resource_alias_metadata` (introduced as a stub in D2-min thinking; now populated for 5 categories); `_build_hot_list_aliases` takes two new optional lookup dicts `skill_metadata_lookup` / `mcp_tool_lookup`; the caller in `RouterLoop.run` builds both maps from `host.list_available_skills()` / `host.get_mcp_servers()`.
- `src/reyn/chat/session.py` — extends `_extract_skill_input_hint` to also return `input_schema` (= the full JSON schema dict) and `input_wrapped` (= the artifact's `wrapped:` flag). `enumerate_available_skills` propagates these as optional fields on each skill dict, so `host.list_available_skills()` callers that need rich schema can read them without re-parsing artifacts.
- `tests/fixtures/llm/router/*.jsonl` — 7 LLMReplay fixtures re-recorded because the tools= description bytes shift for any captured payload that includes a resource-category alias in its hot list.

Diff: 2 production files (+197 / -9) + 7 fixture re-records.

## Empirical verification (weak gemini-2.5-flash-lite, fresh capture)

`scripts/dogfood_trace.py --mode llm-tools-schema <request_id>` against a clean `reyn chat --cui` session now shows skill aliases with their actual input fields:

```
skill__index_docs
  required=['source', 'path', 'description']
  properties=['source', 'path', 'description', 'mode']

skill__eval
  required=['case_name', 'case_input', 'spec_path', 'target_skill_path', 'phase_criteria']
  properties=['case_name', 'case_input', 'spec_path', 'target_skill_path', 'skill_root', 'phase_criteria']

skill__skill_builder
  required=['skill_name', 'description', 'goal']
  properties=['skill_name', 'description', 'goal']
```

Skills whose entry artifact is `user_message` only (`skill__skill_importer`, `skill__mcp_search`, `skill__read_local_files`) correctly fall through to the existing short-description path (= no structured schema, no need to enrich) so their alias shape is unchanged.

## Test plan

- [x] **Helper smoke** — `_resource_alias_metadata` returns rich `(description, parameters)` for `agent.peer__*` / `mcp.server__*` / `rag.corpus__*` from static target lookups, and for `skill__*` / `mcp.tool__*` when the caller supplies the respective metadata lookups
- [x] **Targeted tests** — `pytest tests/test_router_tools_universal_wrappers.py tests/test_router_invoke_skill_enum.py tests/test_replay_skill_router.py` → 44 passed
- [x] **LLMReplay re-record + replay** — `REYN_LLM_RECORD=1 OPENAI_API_KEY=dummy LITELLM_API_BASE=http://localhost:4000 pytest tests/test_replay_skill_router.py` → 14 passed (record), without `REYN_LLM_RECORD` → 14 passed (replay)
- [x] **`ruff check src/reyn/chat/router_loop.py src/reyn/chat/session.py`** clean
- [x] **Live alias inspection** — `REYN_LLM_TRACE_DUMP=... reyn chat --cui` capture confirms `skill__<name>` aliases now carry the skill's required + properties fields. Pre-fix the parameters block was `{properties: {}, additionalProperties: true}` for the same skills.

## Follow-ups (not blocking)

- **memory.entry__X** dispatch shape mismatch (`_read_memory_body_args` returns `{name}` but `read_memory_body` parameters declare `{layer, slug}`). Should land as a separate single-purpose fix once a real-call regression surfaces.
- **mcp.tool live verification** — no MCP servers are configured in my dev env, so the mcp.tool branch is logic-tested but not e2e-tested. Should land alongside the next MCP-touching probe.


tya5:
**[lead-coder review]** — cherry-picked into main as commit a1a5093

The D2-full changes shipped without merging the branch directly because the branch was based on a pre-RunSpawner-extraction main (= would have reverted PlanRunner introduction in commit 5a87e9f, and the router-error classification from #107). The cherry-pick preserves:

- Your full ``router_loop.py`` resource alias schema work (``_resource_alias_metadata`` + ``_drop_required_field``)
- Your ``session.py`` ``_extract_skill_input_hint`` extension (``input_schema`` / ``input_wrapped`` passthrough)
- All 7 fixture re-records

Scope expansion in the landed commit (added during review):
- ``_handle_describe_action`` in ``universal_catalog.py`` now also returns the per-resource schema (skill / agent.peer / mcp.server / rag.corpus / mcp.tool). Without this the hot-list-alias schema fix would only help for top-K skills; non-hot-list skills still got the dispatcher's ``{name, input}`` envelope from ``describe_action``.
- ``_DESCRIBE_SKILL_STRIP_FIELDS`` now strips ``input_schema`` / ``input_wrapped`` so describe_skill stays under the G12 Pattern D attractor threshold; structured schema lives in the hot-list alias and describe_action.
- New 14-test Tier 2 regression: ``tests/test_describe_action_resource_schemas.py``.

Closing as superseded by a1a5093. Full suite (3372 passed) clean.

— lead-coder

PR #143, opened 2026-05-17
fix(tui): Esc on slash picker also clears the slash prefix

tya5:
**[tui-coder]** — 

## Summary

The slash picker is only visible when the input is `/<name-partial>` (no space, no newline — see `_update_picker`), so the entire input text is the slash prefix being discovered. Previously `action_dismiss_picker` hid the picker but left `/h` or `/co` sitting in the input, which:

- Made Esc feel like a no-op (no visual change other than the picker box disappearing)
- Forced the user to backspace before typing anything else
- Risked sending the partial slash text as chat if Enter was pressed by mistake

Slack/Discord clear the prefix on Esc — match that.

Esc on non-slash text remains unchanged (the early-return when the picker isn't visible protects that path).

## Test plan

- [x] Manual: type `/he` → picker shows `/help`. Press Esc → picker hides AND input clears. Empty-state hint reappears.
- [x] Manual: retype `/he` after Esc → picker re-opens normally.
- [x] Manual: type plain text `hello world` (no slash, picker not visible) → press Esc → text preserved. (Regression check.)
- [x] Decision-flow Q1 (testing.ja.md): UX behavior + "this commit author would notice" → **Tier 4 → no automated test**; would have to assert TextArea internal state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #144, opened 2026-05-17
fix(tui): Ctrl+C dismisses the slash picker before cancelling

tya5:
**[tui-coder]** — 

> **Stacked on #143** — base auto-updates to `main` once #143 merges.

## Summary

With the slash picker open, Ctrl+C used to skip past the picker and trigger `(nothing in-flight to cancel)` — the visible overlay didn't move and the user got a misleading message about cancellation. The picker counts as "the current thing" the user wants to abort.

The app-level Ctrl+C binding (`App.BINDINGS`) wins over `InputBar`'s same-key binding, so the dismissal entry-point has to live in `App.action_cancel_inflight`. New hierarchy:

1. Picker visible → `InputBar.dismiss_slash_prefix()` hides it + clears the slash prefix → return
2. Otherwise → existing cancel path (sticky clear, skill/plan/stream cancel)
3. Nothing in-flight → existing `(nothing in-flight to cancel)` message

## Test plan

- [x] Manual: type `/h` (picker shows) → Ctrl+C → picker dismissed AND input cleared; no `(nothing in-flight)` message
- [x] Manual: Ctrl+C in idle (no picker, no in-flight) → still shows `(nothing in-flight to cancel)` (regression check — preserved)
- [x] Manual: Ctrl+C mid-stream (TODO when convenient) → cancel path runs as before
- [x] Decision-flow Q1 (testing.ja.md): UI hierarchy behavior → **Tier 4 → no automated test**; would assert on `App._inputbar` private state or post-cancel TextArea contents (Tier 4 violation per "private state assertion").

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -633,7 +633,19 @@ class ReynTUIApp(App):
         plans were actually cancelled. Without this summary the user
         couldn't tell whether Ctrl+C did anything (= "did it really
         cancel, or was nothing in flight?").
+
+        Ctrl+C is hierarchical: if the slash picker is open, dismiss it
+        first (matches Esc, satisfies the "abort the visible thing"
+        intuition) and do NOT also touch in-flight state. Without a
+        picker visible, fall through to the cancel path below.
         """
+        try:
+            input_bar = self.query_one("#inputbar", InputBar)
+        except Exception:
+            input_bar = None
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — merged via local merge commit `15a9013` (visible in main history). GitHub auto-close didn't fire for this PR (likely stale-base merge timing); closing manually. The PR's head commit `c5c4bbbb` is now an ancestor of `main`.

PR #158, opened 2026-05-17
fix(llm): refactor G12 (answered) signal to role=tool inner field — close issue #156

### LINKED ISSUE #156: (answered) workaround leaks: LLM treats it as literal user paste and responds with canned clarification
**[tui-coder]** —

## Observation

In recent TUI smoke-test sessions I repeatedly saw agent replies that begin with:

> It looks like you've pasted "(answered)" again, which might be a leftover from a previous interaction or a mistake.
>
> Is there anything else I can help you with?

This appears **without** the user having typed anything resembling `(answered)`. The pattern is reproducible while exercising `reyn chat` with `gemini-2.5-flash-lite` via the local LiteLLM proxy, but I have not measured a rate.

## Root cause

The `(answered)` literal is injected by the existing G12 empty-stop attractor workaround at `src/reyn/llm/llm.py:820-821`:

```python
if messages and isinstance(messages[-1], dict) and messages[-1].get("role") == "tool":
    messages = messages + [{"role": "user", "content": "(answered)"}]
```

The workaround's docstring (lines 793-819) documents this is deliberate: a trailing user `(answered)` message pushes weak Gemini variants out of an empty-stop attractor after tool calls (measured: V0 baseline empty_stop 30-60%, V7 trailing-user-(answered) empty_stop 0%, reply 100%). The injected message is not persisted to history.

The workaround works for empty-stop. The new failure mode is the **opposite** attractor: the LLM accepts the message but interprets `(answered)` as a literal user paste and produces a canned clarification reply ("you've pasted (answered) again...") instead of the post-tool summary the user expected.

## Reproducibility (qualitative)

Direct reproductions I observed during the TUI fix-wave smoke tests (#140-#155 series, sessions on 2026-05-17):
- Casual `say something long and reply with a 50-line poem`-style prompts after a prior tool turn
- Free-text follow-ups (`give me a long story`, `explain in detail the difference between coroutines and threads`)

Both sometimes produced the canned `pasted "(answered)"` reply instead of substantive content. **Hypothesis: occurs only on turns where the empty-stop workaround is triggered (last message role=tool) AND the LLM happens to read the literal rather than the neutral-signal interpretation.** Direct verification: not measured. Suggested next step is reusing the V0/V7 N=10 measurement harness from `/tmp/wave_a_trailing_user.py` referenced in the docstring to quantify rate.

## Why this is not a TUI fix

- The injection is at the LLM-call boundary (`llm.py`), not in any TUI widget
- The workaround's docstring already says "Workaround only — true fix is provider-side or different model"
- TUI-side post-filter (= strip canned phrasing from agent replies) would be a bandaid on a bandaid

## Possible directions for discussion

1. **More neutral injection token** (e.g. empty content, `"."`, or a Gemini-specific assistant-role noop) — risks empty_stop regression; would need N=10+ measurement against V0
2. **Provider-side or model-class change** — strong-model upgrade (`gemini-2.5-flash` for tool-call turns only?) avoids both attractors; cost-bound decision
3. **Post-tool turn detection in router** — keep the workaround but post-filter agent replies whose text closely matches the canned `"pasted ... again"` template, replace with a "(empty-stop bypassed, no summary produced)" notice; bandaid but observable
4. **Accept current trade-off** — empty-stop was the worse failure (silent 0-token reply); a canned clarification is degraded but visible

## Related

- `src/reyn/llm/llm.py:793-821` (workaround docstring + injection)
- `src/reyn/chat/router_loop.py:1602` (references the workaround)
- `src/reyn/events/event_schema.py:57` (event schema comment about the workaround)
- No prior open issue (gh issue list searched for `answered attractor`)

tya5:
**[e2e-coder]**

Closes #156.

## Summary

The G12 post-tool empty-stop workaround appended a `{role: "user", content: "(answered)"}` message to the LLM call boundary. The shape violated the OpenAI / Anthropic role contract (= `role=user` content is, by spec, "what the human typed") and weak `gemini-2.5-flash-lite` correctly interpreted the literal as a user paste, producing canned "It looks like you've pasted '(answered)' again..." replies that persisted to `history.jsonl` and snowballed across subsequent turns.

## Root cause

`src/reyn/llm/llm.py:820-821` (old shape):

```python
if messages and isinstance(messages[-1], dict) and messages[-1].get("role") == "tool":
    messages = messages + [{"role": "user", "content": "(answered)"}]
```

The OS was masquerading an orchestration signal as user input. The weak LLM did exactly what the API contract specifies and responded to the apparent user message.

## Fix (V1-INNER)

Embed the signal INSIDE the trailing `role=tool` message content (= contract-correct location for signals about tool results) instead of appending a fake user message. For JSON-shaped tool content (= 100% of current dispatch paths), inject a top-level `_g12_signal` field. For plain-text content, prefix with the signal. For non-string content (= future content-parts API), no-op.

Signal text was strengthened from `"(answered)"` to `"(answered) — task complete; reply to user or chain another tool"` to preserve more of the original V7 empty-stop reduction in the `list_actions` discovery context (= a known empty_stop hotspot per 2026-05-07 V0 baseline).

Helper extracted as pure `_apply_g12_signal(messages: list[dict]) -> list[dict]` for unit-testing.

## Measurement evidence

N=10 each on the reproducing baseline (= tui-coder polluted history + `summarize readme.md` prompt) — full data in issue #156 measurement comments:

| variant | empty_stop | **canned** | text | tool_call |
|---|---|---|---|---|
| V7-baseline (= old `role=user "(answered)"`) | 0 | **10/10** | 0 | 0 |
| V0 (= delete the injection) | 0 | 0 | 9 | 1 |
| **V1-INNER (= this PR)** | 0 | **0** | 10 | 0 |
| V2A (= replace with role=assistant noop) | 0 | 0 | 9 | 1 |

V1-INNER chosen because: (a) 100% canned elimination, (b) highest reply stability (= 10/10 text vs 9/10 for V0/V2A), (c) preserves the signal mechanism in case other contexts (= unmeasured at scale) still need the empty_stop reduction.

## Care-boundary classification

Per `docs/concepts/care-boundary.md`: this is a **structural fix** (= P3 + pre-call context provision). The OS now places orchestration signals where the API contract allows them (= inside `role=tool` content) rather than masquerading as user input. Behavior remains LLM-driven; no behavioral rescue logic added.

## Test plan

- [x] New `tests/test_g12_signal_contract.py` (7 Tier 2 tests, each pinning a real invariant; over-specified edge cases pruned per testing.ja.md)
  - non-post-tool turn no-op gate
  - JSON tool content → `_g12_signal` field
  - **No `role=user "(answered)"` appended (= issue #156 regression guard)**
  - Input immutability
  - Plain-text fallback prefix
  - Non-string content no-op
  - Realistic polluted-history scenario (= tui-coder baseline shape)
- [x] Re-recorded 6 LLMReplay fixtures (= intentional prompt drift per testing.ja.md — message hashes changed because trailing message content is now different)
- [x] `tests/test_replay_skill_router.py::test_list_actions_discovery_then_invoke` xfailed (`strict=False`) with explicit reason — V1-INNER does NOT fully replicate V7's empty_stop reduction in the `list_actions` discovery context. Follow-up: stratified empty_stop re-measurement per [lead-coder]'s frame in #156.
- [x] Adjacent tests (`test_chat_turn_completed_inline`, `test_router_loop_routing_decided`, `test_session_cost_accumulation`) — no regression
- [x] Ruff clean on touched files
- [x] **E2E re-verified** against the actual `tui-coder` polluted-history workspace: canned **10/10 → 0/10** confirmed via `REYN_LLM_TRACE_DUMP` capture. Reply on the canned-trigger trace is now an honest empty_stop fallback ("The model returned an empty response. Please try rephrasing your request or check your configuration.") rather than a hallucinated user-paste claim.

## Out of scope / follow-up

- **Mechanism B (= history-pollution snowball)** is no longer triggered for NEW turns because the OS no longer injects the `role=user "(answered)"` that primed it. Existing polluted `history.jsonl` files (= e.g. tui-coder default agent's 87-msg log) won't self-heal under V1-INNER; operator may want manual cleanup or accept eventual replacement via chat compaction.
- **Empty_stop stratification**: V1-INNER's empty_stop reduction is not fully measured across all contexts (= [lead-coder]'s ask in #156). The `list_actions` test xfail documents one known context where empty_stop emerges; a stratified study can decide whether V1-INNER needs a stronger signal variant.
- **Cross-model**: only `gemini-2.5-flash-lite` measured. Strong-model behavior under V1-INNER unknown (= cost-gated).


tya5:
**[e2e-coder]** — follow-up commit `79cb2dd5` addressing the 2 owner asks (= env-var disable + empty-tool-result behaviour).

## 1. Env var disable: `REYN_G12_SIGNAL=off`

```bash
REYN_G12_SIGNAL=off reyn chat ...
```

Case-insensitive; `off` / `0` / `false` / `no` all disable. Any other value (or unset) leaves the workaround active. Useful for:
- A/B comparison against the no-workaround baseline
- Diagnostic runs where the operator wants to confirm whether the canned-reply replacement (= empty_stop fallback in some contexts) is more or less annoying than the prior canned-reply
- Environments where the operator has measured the empty_stop rate independently and confirmed it's no longer present

## 2. Empty tool result

Before the follow-up, `content = "{}"` produced `'{"_g12_signal": "...", }'` — **invalid JSON** (trailing comma before `}`). The helper now detects the empty-object shape (`content[1:].lstrip().startswith("}")`) and elides the separator:

| input | output | JSON valid? |
|---|---|---|
| `""` | `(answered) — task complete; reply to user or chain another tool\n\n` | (= non-JSON path) |
| `"{}"` | `{"_g12_signal": "..."}` | ✓ |
| `"{ }"` | `{"_g12_signal": "..." }` | ✓ |
| `'{"a": 1}'` | `{"_g12_signal": "...", "a": 1}` | ✓ |

(`""` follows the plain-text fallback since it doesn't start with `{`; signal-only output is the documented behaviour for empty/non-JSON content.)

## Test coverage

+15 Tier 2 tests in `test_g12_signal_contract.py`:
- 2 empty-JSON-object validity tests
- 1 env-var-off no-op test
- 7-parametrised case-insensitive disable variants (`off`/`OFF`/`0`/`false`/`False`/`no`/`NO`)
- 5-parametrised non-disable values (`""`, `on`, `1`, `true`, `garbage`) → workaround stays active

22 tests in the file, all pass. Adjacent `test_replay_skill_router.py` still 13 passed + 1 xfailed (= no regression from the fixture re-recordings).

Ruff clean.

Ready for re-review.

PR #205, opened 2026-05-18
polish(tui): lifecycle markers render as dim inline divider, not system block

### LINKED ISSUE #162: TUI: compaction events are invisible in the conversation pane
**[tui-coder]** —

Finding from the 2026-05-18 TUI exploration ([H. 長期セッション UX] angle). The TUI fix is straightforward (write a dim `[↑ N turns compacted]` marker to the conv log), but the trace plumbing to surface compaction events into the chat outbox doesn't exist — `ChatEventForwarder` only handles `phase_*`, `workflow_*`, `llm_*`, `act_executed`. So this is cross-layer (forwarder + possibly a new subscriber registration site).

## Observation

When `CompactionController` compacts old turns, it emits `compaction_started` / `compaction_completed` events to `.reyn/events/` only — the conversation pane shows no notification. Users don't know their early-session turns have been replaced by a rolling summary, and have no way to tell how far back "real" history vs compacted summary extends.

## Repro

1. Run a session long enough to trigger compaction (or inspect the Events tab for `compaction_completed`)
2. The conv pane shows a continuous thread with no indicator that pre-seq-N turns are now a summarised proxy, not the original text

## Why TUI-internal alone doesn't work

`src/reyn/chat/services/compaction_controller.py` lines 219-249 emit:

```python
self._events.emit("compaction_started", new_turn_count=..., covers_through_seq=...)
self._events.emit("compaction_completed", covers_through_seq=..., section_lengths=...)
```

`src/reyn/chat/forwarder.py:32-100` defines `ChatEventForwarder.__call__` which dispatches to `on_<event_type>` handlers — but has none for `compaction_*`. The TUI sees nothing unless that subscriber is extended.

A second concern: `ChatEventForwarder` is per-skill (constructor takes `skill_name`), so compaction (which is a controller, not a skill) would need a different subscriber pattern, or a new top-level chat-events subscriber.

## Direction (= owner decision)

- **(a) Extend `ChatEventForwarder`**: add `on_compaction_completed` and ensure compaction's `_events` flows through the same subscriber registry. Cleanest reuse.
- **(b) New subscriber `ChatLifecycleForwarder`**: separate from skill-scoped forwarder, handles session-level events (compaction, attach/detach, budget warnings). More structure, room for growth.
- **(c) Direct chat outbox emit from `CompactionController`**: skip events, push `OutboxMessage(kind="system", text="[↑ ... compacted]")` straight into the chat outbox. Tighter coupling, simpler.

Once one of the above lands, the TUI side is a 5-line addition (= write the marker via `conv._write_log()` from `app_outbox`, similar to the existing fold-expired marker).

**Cost (TUI side alone)**: S. **Cost (cross-layer)**: M (1-2h). **Priority**: P1 (壊れ / 混乱) — users actively lose information about what the LLM is seeing.

tya5:
**[tui-coder]** —

## Summary

- Lifecycle markers from `ChatLifecycleForwarder` (e.g. `[↑ 3 turns compacted]`, added in #201 to close #162) currently render through the standard `system`-kind path: timestamped `· system` speaker header + dash rule + indented body. That's heavier than warranted for an advisory state-change announcement which is neither speech nor command output. e2e-coder explicitly called out dim-styling refinement as out of scope for #201's base fix.
- Add `_is_lifecycle_marker()` + `_render_lifecycle_marker()` helpers in `conversation.py` that detect the `[↑ ... ]` single-line shape and render it as a dim `── ↑ <body> ────…` inline divider matching `_date_separator` (same `dim #666666` styling, same `_DASH_TOTAL=38` width).
- Route `_render_system_message` through the new helpers when the marker shape matches; regular system messages (slash-command output, multi-line plans) keep the existing speaker-tagged path.

## Why this is TUI-only

Pure rendering polish in `conversation.py` — no skill / OS / event contract changes. The forwarder's outbox message kind stays `system` and its text shape stays `[↑ N turns compacted]`; only the conv pane's interpretation of that shape is sharpened.

## Test plan

- [x] **Tier 2 contract tests** — `tests/cli/test_lifecycle_marker_render.py`, 7 cases:
  - Detection heuristic accepts compaction-shape markers, rejects multi-line / plain-bracketed / plain-text content
  - Rendered marker width equals `_DASH_TOTAL`
  - Every span carries `dim` styling
  - Marker body text survives the render
  - End-to-end: `_render_system_message(marker)` writes the divider with no `· system` speaker tag in the log
  - End-to-end: `_render_system_message(non-marker)` still emits the full `· system` header
- [x] **Manual**: `python -m pytest tests/cli/ -x` — 190 passed in 31.5s (no regressions on neighbouring rendering tests)
- [x] **Manual**: `python -m pytest tests/test_chat_lifecycle_forwarder.py` — 6 passed; the forwarder's emission contract (= shape of the marker text) is unchanged
- [ ] **Visual (skipped — pre-merge tmux smoke not run; marker only fires after a real compaction event)**: dim divider renders without a `· system` header on a 120×30 conv pane after a compaction. Pinned by the Tier 2 end-to-end tests above.

### DIFF
--- a/src/reyn/chat/tui/widgets/conversation.py
+++ b/src/reyn/chat/tui/widgets/conversation.py
@@ -136,6 +136,41 @@ def _msg_header(label: str, name_style: str, dash_style: str) -> Text:
     return t
 
 
+def _is_lifecycle_marker(text: str) -> bool:
+    """Heuristic: ``ChatLifecycleForwarder`` marker text starts with ``[↑``
+    and ends with ``]``, and is single-line.
+
+    Lifecycle markers (= compaction, future attach-budget signals) are
+    state-change announcements, not speech — rendering them as a full
+    speaker-tagged system block (= timestamp + ``· system`` header + dash
+    rule + indented body) gives them more visual weight than they
+    deserve. The conv pane routes them through ``_render_lifecycle_marker``
+    so they appear as a dim inline divider, matching the date-separator
+    style.
+    """
+    t = text.strip()
+    return t.startswith("[↑") and t.endswith("]") and "\n" not in t
+
+
+def _render_lifecycle_marker(text: str) -> Text:
+    """Dim ``── ↑ <body> ────…`` inline divider for a lifecycle marker.
+
+    Shape mirrors ``_date_separator`` — same total cell width (``_DASH_TOTAL``)
+    and same ``dim #666666`` styling so the visual rhythm stays consistent
+    with day-boundary markers and other inline dividers.
+    """
+    stripped = text.strip().lstrip("[").rstrip("]").strip()
+    label = f" {stripped} "
+    t = Text()
+    lead_dashes = 2
+    label_cells = cell_len(label)
+    trail_dashes = max(1, _DASH_TOTAL - lead_dashes - label_cells)
+    t.append("─" * lead_dashes, style="dim #666666")
+    t.append(label, style="dim #666666")
+    t.append("─" * trail_dashes, style="dim #666666")
+    return t
+
+
 def _date_separator(date_str: str) -> Text:
     """Dim ``── YYYY-MM-DD ───…`` line for day boundaries between turns.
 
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — waiver for unticked visual test item, then merging.

Test plan の "Visual" item は `[ ]` 表記で残っているが、 inline reason 「marker only fires after a real compaction event」 が articulated。 CLAUDE.md PR workflow §1 形式 (`[x] (skipped — <reason>)`) と format 非合致だが、 substance は明確。

私の判断: **waive**。 理由:
- Tier 2 contract tests 7 cases が detection + render path 完全 cover (= marker shape / width / dim styling / e2e に分岐 path 含む)
- Pre-merge env で compaction event を自然発火させる手段なし (= 真の visual smoke は post-merge runtime でのみ)
- Author の articulated reason が waiver の rationale を成立させる

Author への calibration note: 次回以降 unticked item は `[x] (skipped — <reason>)` 形に正規化、 reviewer の literal-letter parse が成立するように。 今 PR は本文に reason あるので OK。

Merge 進めます。

PR #317, opened 2026-05-20
TUI: persist /cost-inline toggle to .reyn/tui_prefs.json (F10)

tya5:
**[tui-coder]** — UX wave parking-lot finding F10 (P3/M/score=0.5). 5 of 5 planned parking-lot PRs — closes the parking-lot batch.

## Observation

``/cost-inline`` toggled an in-memory ``_cost_inline_enabled`` bool on ``ReynTUIApp`` (default False per `src/reyn/chat/tui/app.py:140`). Every ``reyn chat`` restart reverted the toggle to off even if the user had explicitly set it on the previous session. No way to make it stick short of an ``input()`` wrapper in shell rc, which TUI users don't have.

## Fix

Small ``src/reyn/chat/tui/prefs.py`` module — ``load_tui_prefs`` / ``save_tui_prefs`` read/write ``<project_root>/.reyn/tui_prefs.json``. Flat JSON dict, additive keys round-trip unchanged so future prefs (default panel tab, banner default, …) land without migration ceremony.

Wire:
- ``on_mount`` calls ``load_tui_prefs(self._project_root_path())`` and restores ``_cost_inline_enabled`` from the ``cost_inline`` key (defaults to False, preserving first-launch behaviour).
- ``_on_cost_inline_toggle`` calls ``save_tui_prefs`` after applying the user's "on" / "off" / toggle. Failure is silent (= read-only FS, permissions) — the in-memory toggle still applies for the current session, only persistence fails.
- New ``ReynTUIApp._project_root_path()`` helper centralises the ``registry._project_root`` getattr so prefs loaders / future consumers don't reach into private attributes ad-hoc.

Defensive degradation:
- File missing → empty dict → defaults stay (= first launch unchanged)
- Malformed JSON / non-dict root → warning + empty dict (= bad hand-edit doesn't break the TUI)
- Write failure → warning + in-memory state still works

## Visual / Behaviour

Session 1:
```
> /cost-inline on
  cost-inline on            ← transient status (existing)
```
On disk now:
```
.reyn/tui_prefs.json:
  { "cost_inline": true }
```

Session 2 (restart):
```
(TUI mounts with _cost_inline_enabled = True restored from disk)
> /cost-inline           ← bare = toggle
  cost-inline off          ← flipped from the restored true
```
On disk:
```
.reyn/tui_prefs.json:
  { "cost_inline": false }
```

## Files

| file | change |
|---|---|
[...]

PR #301, opened 2026-05-20
TUI: surface --no-restore warning into conversation pane (F12)

tya5:
**[tui-coder]** — UX wave finding F12 (P1/S/score=3.0 from triage). 1 of 8 planned PRs.

## Observation

`reyn chat --no-restore` prints `⚠ --no-restore: skill state on disk is NOT loaded...` to stderr from `cli/commands/chat.py:323-328`. Once Textual takes over the terminal (alt screen), that pre-TUI line is hidden — the operator can't confirm the flag took effect, and can't recall why in-flight skills are missing from the agents tab.

## Fix

Wire `skip_restore` through `run_tui` → `ReynTUIApp(no_restore=...)`, and in `on_mount` write an amber dim-style warning line to the conversation pane via the same `conv._write_log` path the banner uses.

```
⚠ --no-restore: in-flight skill state was NOT loaded this run. Restart without --no-restore to resume.
```

Amber styling (`#d4a017`) matches the existing `[N pending]` header badge (= advisory-but-non-fatal). The stderr print stays — it remains visible in `--cui` mode and in edge-case CLI flows where the TUI never mounts.

## Visual

`reyn chat --no-restore`:
```
   Reyn                                                 default  │  standard  │  ...
────────────────────────────────────────────────────────────────────────────────
  ⚠ --no-restore: in-flight skill state was NOT loaded this run. Restart without --no-restore to resume.



    Type / for commands  ·  /help for a guide
    ...
```

`reyn chat` (no flag) — pane starts clean as before, no warning line.

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/app.py` | `ReynTUIApp.__init__` adds `no_restore: bool = False` / `on_mount` writes amber warning when set / `run_tui` propagates the flag |
| `src/reyn/cli/commands/chat.py` | `_main_tui` passes `no_restore=skip_restore` into `run_tui` |

## Test plan

- [x] Manual: `OPENAI_API_KEY=dummy reyn chat --no-restore` in tmux MCP — amber warning visible at top of conv pane on mount.
- [x] Manual: `OPENAI_API_KEY=dummy reyn chat` (no flag) — no warning line, pane starts clean.
- [x] `ruff check src/reyn/chat/tui/app.py src/reyn/cli/commands/chat.py` — clean (pre-push 実走済).
[...]

PR #327, opened 2026-05-20
TUI: Esc binding description "Voice cancel" → "Dismiss / cancel" (K3)

tya5:
**[tui-coder]** — Wave-2 finding K3 (P2/S/score=2.0). 4 of 6 planned wave-2 PRs.

## Observation

`app.py` line 96 declared `Binding("escape", "voice_cancel", "Voice cancel", priority=True, show=False)`. But `action_voice_cancel` actually multiplexes 3 behaviors (see `app.py:1367`):

1. If voice is recording → cancel recording
2. Else if any ErrorBox is mounted → dismiss the topmost
3. Else if the right panel is visible → close it

The "Voice cancel" label misleads any user who isn't using voice — they see a Keys-tab entry that names a feature they don't use, for a key they press daily to dismiss errors and close the panel.

## Important visibility note (= honest scope)

After implementing, smoke testing revealed the Keys tab **does NOT actually surface the app-level Esc description** — `keys_tab.py` `_INPUT_OWNED_KEYS = {"enter", "escape", ...}` defers Esc to InputBar's "Dismiss" binding instead. So this rename is a **source-of-truth correction** rather than a user-visible UX change.

Why ship it anyway:
- The description now matches what the action actually does (= future contributors reading `app.py` BINDINGS aren't misled).
- Debug tooling / Textual introspection that reads bindings sees a correct label.
- If `_INPUT_OWNED_KEYS` is ever narrowed (= surface app Esc alongside InputBar Esc), the description is already correct.

This intentionally narrows K3 from the agent's original "Keys tab shows 'Voice cancel'" claim, which my smoke test didn't reproduce.

## Fix

```python
# Before:
Binding("escape", "voice_cancel", "Voice cancel", priority=True, show=False),

# After:
Binding("escape", "voice_cancel", "Dismiss / cancel", priority=True, show=False),
# + multi-line comment explaining the multiplex action + why the
#   description uses the dominant chat-time meaning.
```

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/app.py` | Esc binding description renamed; comment block explaining the multiplex |

## Test plan

- [x] `grep -rn "Voice cancel" src/ tests/` — only my new comment matches; no other code or tests pin the prior label.
- [x] `ruff check src/reyn/chat/tui/app.py` — clean (pre-push 実走済).
- [x] (skipped — not user-visible) tmux Keys tab capture — Esc shows under [INPUT] as "Dismiss" (InputBar binding via `_INPUT_OWNED_KEYS` deference). Re-confirmed the rename does not surface in the Keys tab today.

## Scope guard

**NOT in scope**:
- Lift Esc from `_INPUT_OWNED_KEYS` so the app description surfaces — would change Keys tab behaviour; out of K3 scope (= separate UX call about whether app-Esc or InputBar-Esc is the dominant Keys tab label).
- New Esc affordance (= e.g. close the slash picker first when both picker open + ErrorBox mounted) — out of wave-2 K3 scope.
- Footer hint addition of Esc — separate K4/K5 family (this PR is purely the rename).

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -93,7 +93,14 @@ class ReynTUIApp(App):
         # touching the keyboard for the edit step. Gated by check_action
         # so plain Enter in the input bar still does its normal submit.
         Binding("enter", "voice_stop_and_submit", "Voice send", priority=True, show=False),
-        Binding("escape", "voice_cancel", "Voice cancel", priority=True, show=False),
+        # Esc multiplexes: cancel voice recording / dismiss ErrorBox /
+        # close right panel — see ``action_voice_cancel``. The action
+        # name is historical (= voice was the first user, the rest
+        # accreted), but the surfaced description should reflect the
+        # dominant chat-time meaning so the Keys tab doesn't tell a
+        # non-voice user "Voice cancel" for a key they use daily for
+        # dismissing errors. Wave-2 K3.
+        Binding("escape", "voice_cancel", "Dismiss / cancel", priority=True, show=False),
         Binding("ctrl+backslash", "screenshot", "Screenshot", priority=True, show=False),
     ]
 


tya5:
**[lead-coder]** — approve, batch merge する (TUI UX wave-2 6/6 batch review)。

cluster type 3 second instance establish (= wave-1 #301-#317 と同 「1 author × exploration trigger → N PRs」 pattern 再現)、 standard batching template 適用 (= top-N priority cutoff + user follow-up drain)、 PR #327 で 「sub-agent claim verify Pre-PR step」 honest scope narrowing 適用 = wave operational discipline establish。

approve、 batch merge proceed。

PR #328, opened 2026-05-20
TUI: footer hint swaps Ctrl+D quit → Ctrl+C cancel (K4)

tya5:
**[tui-coder]** — Wave-2 finding K4 (P2/S/score=2.0). 5 of 6 planned wave-2 PRs.

## Observation

The always-visible footer hint read:
```
Enter send │ Ctrl+D quit │ Ctrl+L clear │ Ctrl+B panel │ Ctrl+P/N turn
```
`Ctrl+C` (cancel in-flight) was absent despite being the highest-frequency interactive key during active skill runs — every time the user notices a long-running call they want to abort. `Ctrl+D` was advertised instead, even though it duplicates the universal terminal-EOF convention every user already knows from the shell.

## Fix

Swap them. Both keys remain in the Keys tab (`Ctrl+B → Keys`), so Ctrl+D quit is still discoverable. The always-visible footer now advertises the key users actually need in-session.

```
Before:  Enter send │ Ctrl+D quit │ Ctrl+L clear │ Ctrl+B panel │ Ctrl+P/N turn
After:   Enter send │ Ctrl+C cancel │ Ctrl+L clear │ Ctrl+B panel │ Ctrl+P/N turn
```

Footer width stays under the 72-cell budget (= 80-col-terminal safe) — `Ctrl+C cancel` is 2 cells longer than `Ctrl+D quit` but fits comfortably within the existing tail margin.

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/widgets/input_bar.py` | `_build_hint` swaps `Ctrl+D quit` → `Ctrl+C cancel`; updated comment block explaining the rationale (= in-session frequency over universal convention) |

## Test plan

- [x] Manual: tmux MCP — `OPENAI_API_KEY=dummy reyn chat` → footer shows `Enter send │ Ctrl+C cancel │ Ctrl+L clear │ Ctrl+B panel │ Ctrl+P/N turn` (verified visually).
- [x] `grep -rn "Ctrl+D quit" src/ tests/` matches only my own comment; no test pins the prior string.
- [x] `ruff check src/reyn/chat/tui/widgets/input_bar.py` — clean (pre-push 実走済).

## Scope guard

**NOT in scope**:
- Adding both `Ctrl+D quit` AND `Ctrl+C cancel` — would push the line past the 72-cell budget; the swap is the cleanest fit.
- `Ctrl+O focus` hint addition (= K5, separate PR candidate; dropped from wave-2 top-6 but listed in parking lot).
[...]

PR #326, opened 2026-05-20
TUI: Keys tab — j/k/space/c surface as PANEL-universal (K1)

tya5:
**[tui-coder]** — Wave-2 finding K1 (P2/S/score=2.0). 3 of 6 planned wave-2 PRs.

## Observation

`j`, `k`, `space` worked on every right-panel tab (events / agents / memory / docs / pending / cost) via `RightPanel.on_key` — they scroll the current tab's list or toggle the preview pane. `c` was the generic "copy current view" with a pending-tab override (= claim cursor). But `keys_tab.py:35` classified the first three as `_DOCS_KEYS` (= surfaced under "DOCS (gated)" if any iteration reached them) and `c` was completely absent from the tab.

Users on Events / Agents / Memory who used j/k to scroll never saw it documented for their tab. The copy affordance was invisible to anyone who didn't read the source.

## Fix

Move `j` / `k` / `space` from `_DOCS_KEYS` to `_PANEL_KEYS` (= correct the latent classifier mislabel), add `c` to the same set, and mirror the existing `h`/`l` explicit-add pattern in `render_keys` to actually surface them in the Keys tab output.

`/` stays in `_DOCS_KEYS` since it opens the docs filter and no other tab consumes it.

## Visual

Before (sub-agent report):
```
[PANEL]
⌃O      Focus panel
⌃W      Next tab
⌃⇧W     Prev tab
⌃⇧O     Prev tab (alt)
h       Widen panel
l       Narrow panel

[DOCS (gated)]
... (would label j/k/space here if visible)
```

After:
```
[PANEL]
⌃O      Focus panel
⌃W      Next tab
⌃⇧W     Prev tab
⌃⇧O     Prev tab (alt)
h       Widen panel
l       Narrow panel
j       Scroll down (current tab)
k       Scroll up (current tab)
Space   Toggle preview pane
c       Copy current view (pending tab: claim cursor)
```

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/widgets/right_panel/keys_tab.py` | `_PANEL_KEYS` extended with j/k/space/c; `_DOCS_KEYS` narrowed to {"/"}; `render_keys` `_PANEL_EXPLICIT` list adds new visible entries with descriptions |

## Test plan

[...]

PR #325, opened 2026-05-20
TUI: /quit + /exit as registry commands (palette + /help visible) (P3)

tya5:
**[tui-coder]** — Wave-2 finding P3 (P2/S/score=2.0). 2 of 6 planned wave-2 PRs.

## Observation

`/quit` and `/exit` were tracked only in `_BUILTIN_HINTS` in `src/reyn/chat/slash/help.py` because the TUI intercepted them in `on_input_bar_user_submitted` (`src/reyn/chat/tui/app.py:653-658`) BEFORE the slash registry ever saw them. But the slash *palette* reads from the registry, so typing `/q` or `/ex` looking for them produced an empty picker — users who knew `/quit` existed from `/help` couldn't discover it through the palette autocomplete they'd just learned from `/help` itself.

## Fix

Land both as first-class registry commands so they surface in palette + `/help` via the same code path every other slash uses.

- **New** `src/reyn/chat/slash/quit.py` — single `_quit_handler` registered under both `quit` and `exit` names. Two separate `@slash` registrations (NOT the `aliases=` field) because the palette filter reads `c.name.startswith(token)` only — alias entries would not surface on `/ex`.
- Handler emits `OutboxMessage(kind="__quit__", text="")`.
- `app_outbox.py` adds `"__quit__": self._on_quit` mapping → `self._app.call_later(self._app.action_quit_tui)`. Same shutdown path Ctrl+D fires (= WAL flush + agent persistence unchanged).
- **Removed** the TUI's hard-coded `cmd_body in {"quit", "exit"}` intercept: `/quit` flows through the standard registry dispatch like every other slash. Single source of truth.
- `help.py` `_BUILTIN_HINTS` emptied (kept as a future home for any genuinely TUI-only intercepts) since `/quit` + `/exit` now come from the registry walk.

## Visual

Type `/q`:
```
 ┌─────────────────────────────────────────────┐
 │ ▌ /quit  Exit the chat (alias: /exit, Ctrl+D)│
 └─────────────────────────────────────────────┘
  /q
```

Type `/ex` (both /exit and /expand match — both legitimate):
```
 ┌─────────────────────────────────────────────┐
 │ ▌ /exit    Exit the chat (alias: /quit, Ctrl+D)│
 │   /expand  Show full content of the last folded agent reply │
 └─────────────────────────────────────────────┘
  /ex
```

Submit `/quit` → TUI exits cleanly to shell.

## Files

| file | change |
|---|---|
[...]

PR #323, opened 2026-05-20
TUI: slash palette alphabetical sort (P1)

tya5:
**[tui-coder]** — Wave-2 finding P1 (P2/S/score=2.0). 1 of 6 planned wave-2 PRs.

## Observation

`input_bar.py:367` sorts palette matches by `(len(c.name), c.name)`. On the empty-token open (= user types just `/`), the first 8 visible rows are the shortest commands — `/copy`, `/cost`, `/help`, `/list`, `/plan`, `/agent`, `/reset`, `/skill`. High-value entries `/attach`, `/budget`, `/cancel`, `/memory`, `/pending` etc. get pushed below the `+12 more` footer. A user who stops scanning at the first wave never discovers them.

## Fix

Switch to pure alphabetical (`key=lambda c: c.name`). Matches the order `/help` already uses, so muscle memory transfers in both directions.

## Visual

Before (length-first):
```
/copy   /cost   /help   /list   /plan   /agent  /reset  /skill   +12 more
```

After (alphabetical):
```
/agent  /agents /answer /attach /budget /cancel /copy   /cost   +12 more
```

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/widgets/input_bar.py` | `matches.sort` key changed from `(len(c.name), c.name)` to `c.name` + comment explaining the muscle-memory rationale |

## Test plan

- [x] Manual: tmux MCP — `OPENAI_API_KEY=dummy reyn chat` → type `/` → picker rows alphabetical (`/agent /agents /answer /attach /budget /cancel /copy /cost +12 more`).
- [x] `ruff check src/reyn/chat/tui/widgets/input_bar.py` — clean (pre-push 実走済).
- [x] (skipped — Tier 4 risk) automated test pinning row order — UX preference, should stay free to tweak.

## Scope guard

**NOT in scope**:
- Picker height / row count tweaks
- Filter / fuzzy match changes
- Subcommand enumeration (= P4 dropped from wave-2 top-6)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #329, opened 2026-05-20
TUI: Esc dismisses slash picker before ErrorBox (P2)

tya5:
**[tui-coder]** — Wave-2 finding P2 (P1/M/score=1.5). 6 of 6 planned wave-2 PRs — closes the wave.

## Observation

With both an ErrorBox mounted AND the slash picker open simultaneously (e.g. user typed `/attach` Enter → ErrorBox, then `/co<typo>` → picker showing matches), pressing Esc dismissed the ErrorBox while leaving the picker open. The user had to press Esc a second time to clear the picker.

**Why**: `check_action` for `voice_cancel` returned True when an ErrorBox was mounted (`src/reyn/chat/tui/app.py:1425`), so the app claimed Esc, and `action_voice_cancel` ran its ErrorBox-dismiss branch BEFORE the InputBar could see the key. The picker / slash-prefix handling lived only in InputBar's own `action_dismiss_picker`, which is reached only when the app fall-through path runs (= no overlay claims Esc).

## Reproduction (= verified before fix)

1. Launch `reyn chat`
2. Type `/attach` Enter → ErrorBox `│ ✗ usage: /attach <name>  ▶` mounts
3. Type `/co` → picker shows 3 matches (copy / cost / cost-inline)
4. Press Esc once → picker stays, ErrorBox dismisses with `(dismissed)` breadcrumb
5. Press Esc again → picker dismisses

## Fix

Reorder `action_voice_cancel` so the slash-prefix dismissal runs **before** the ErrorBox dismissal. The picker is the foreground UI the user is actively interacting with — dismissing it first matches the mental model, and lets a stale ErrorBox wait one more keypress.

```
Priority (new):
  1. Voice recording → cancel recording
  2. **InputBar slash-entry → dismiss picker + clear prefix**   ← NEW
  3. ErrorBox visible → dismiss topmost (writes "(dismissed)" breadcrumb per F2)
  4. Side panel visible → close panel
```

`check_action` semantics unchanged: still returns True on overlay/recording presence, so the app continues to claim Esc when there's something to act on. The new step 2 inserts the picker check inside the app's handler rather than via the InputBar fall-through path.

## Visual

After fix (with the same repro):
- Step 4 (Esc once) → **picker + input clear in one press**, ErrorBox remains visible at footer
- Step 5 (Esc twice) → ErrorBox dismisses with `(dismissed)` breadcrumb (= F2 path)

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/app.py` | `action_voice_cancel` adds picker-first branch via `input_bar.dismiss_slash_prefix()` before ErrorBox; `check_action` comment updated to explain the new ordering |

## Test plan

- [x] Manual: tmux MCP — full repro with both ErrorBox + picker active → Esc once → picker clears, ErrorBox remains; Esc twice → both clear.
- [x] Manual: regression check with picker only (no ErrorBox) → Esc once → picker dismisses (= InputBar fall-through path unchanged).
- [x] `ruff check src/reyn/chat/tui/app.py` — clean (pre-push 実走済).
- [x] (skipped — Tier 4 risk) automated test pinning the ordering — would couple to widget mount/dismiss internals.

## Scope guard

**NOT in scope**:
- Voice + picker case (= voice recording still wins per design — recording is the most ephemeral state).
- Shift+Esc "dismiss all" shortcut (= dropped from triage, P3-class).
- Reordering between ErrorBox and panel-close — out of P2 scope.

## Wave-2 summary (this is the last PR)

| Rank | Finding | PR |
|---|---|---|
| 1 | P1 slash palette alphabetical sort | #323 |
| 2 | P3 /quit /exit registry-visible | #325 |
| 3 | K1 Keys tab j/k/space/c PANEL-universal | #326 |
| 4 | K3 Esc binding description (source-level) | #327 |
| 5 | K4 footer Ctrl+C cancel | #328 |
| 6 | P2 Esc picker dismiss in 1 press (this PR) | this |

Parking lot (= dropped from top-6, candidates for follow-up "parking lot も" drain):
- P4 subcommand completion enumeration (P3/L)
- P5 /help wraps off-screen (P3/M)
- K2 ctrl+shift+n/p phantom binding cleanup (P3/S)
- K5 Ctrl+O focus hint visibility (P3/S)
- S3 clear() StreamingRow remove() leak (P3/S)
- S4 Ctrl+C misses orphan StreamingRow (P2/M — connected to S3)
- S5 _last_long_reply single-slot fold stash (P3/S)

Cross-layer (= leader 相談事項、 私 not in scope):
- S1 streaming pipeline dead code in local mode (cross-layer)
- S2 kind="agent" rendering investigation (cross-layer; possible timing artifact with sandbox_2 proxy mismatch window)

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -1365,21 +1365,36 @@ class ReynTUIApp(App):
         )
 
     def action_voice_cancel(self) -> None:
-        """Esc — cancel voice recording, dismiss ErrorBox, or close panel.
+        """Esc — cancel voice recording, dismiss slash picker / ErrorBox, or close panel.
 
         Priority:
           1. If recording → cancel recording.
-          2. Else if ErrorBoxes visible → dismiss the topmost one.
-          3. Else if side panel is visible → close it.
+          2. **Else if InputBar is in slash-entry → dismiss picker + clear**
+             prefix (= wave-2 P2). Previously this was unreachable when
+             an ErrorBox was also mounted because ``check_action`` claimed
+             Esc on behalf of step 3 and the user had to press Esc twice
+             (once for ErrorBox, again for picker). Picker first matches
+             the user's mental model (= the picker is the foreground UI
+             they're actively interacting with) and lets a stale
+             ErrorBox wait one more keypress.
+          3. Else if ErrorBoxes visible → dismiss the topmost one.
+          4. Else if side panel is visible → close it.
 
         Gated by check_action so this never fires when no condition holds
-        (allowing InputBar's own Esc binding to handle slash-picker dismissal).
+        (allowing InputBar's own Esc binding to handle slash-picker
+        dismissal in the no-other-overlay case).
         """
         if self._voice_input is not None and self._voice_input.is_recording:
             self._voice_input.cancel()
             self._voice_set_input_locked(False)
             self._voice_status("✗ recording cancelled", style="dim #555555")
             return
+        try:
+            input_bar = self.query_one("#inputbar", InputBar)
+            if input_bar.dismiss_slash_prefix():
+                return
+        except Exception:
+            pass
         try:
             conv = self.query_one("#conversation", ConversationView)
             if conv.has_error_boxes():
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — approve, batch merge する (TUI UX wave-2 6/6 batch review)。

cluster type 3 second instance establish (= wave-1 #301-#317 と同 「1 author × exploration trigger → N PRs」 pattern 再現)、 standard batching template 適用 (= top-N priority cutoff + user follow-up drain)、 PR #327 で 「sub-agent claim verify Pre-PR step」 honest scope narrowing 適用 = wave operational discipline establish。

approve、 batch merge proceed。

PR #330, opened 2026-05-20
fix(eval): make spec_path optional in output schemas — closes B43-B46 4x recurring schema-validation failure

tya5:
**[dogfood-coder]** — eval skill output schemas required \`spec_path\`, but when scenarios invoke eval without supplying a spec.md path (= ad-hoc eval requests, the common case) the LLM emits \`null\` and validation rejects the whole eval output → \`skill_run_failed\`. Observed in dogfood B43/B44/B45/B46 as \`eval_run_direct_llm\` R verdict across 4 consecutive batches.

## Root cause

\`spec_path\` is a pass-through reference field — postprocessor \`compute_eval_score\` does not read it. Marking it required forced the LLM to fabricate a value or fail; both bad.

## Fix

Drop \`spec_path\` from \`required\` in 3 schemas:
- \`eval_result_raw.yaml\` (LLM-authored raw artifact)
- \`skill.md\` postprocessor python step \`output_schema\`
- \`skill.md\` caller-facing eval_result \`output_schema\`

Field stays in \`properties\` so callers can still pass it through when known.

## Test plan

- [x] 5 new Tier 2 tests pin \`spec_path\` NOT in required across all 3 schemas + other required fields remain (= regression guard)
- [x] 122 existing eval-related tests still pass (\`pytest -k eval\`)
- [ ] B47 dogfood batch will verify \`eval_run_direct_llm\` scenario no longer hits the validation failure (= primary end-to-end check)

## References

- B43 retrospective NF list: eval postprocessor schema validation persist
- B45 retrospective W2 NF: eval spec_path=None schema validation failure (HIGH, recurring)
- B46 retrospective W2 NF: same, 4-batch recurring
- B46 PR #322 carry-over → user pushback "既知不具合を残したまま次batch進めることは禁止" → \`feedback_no_batch_with_known_bugs\` memory established

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #332, opened 2026-05-20
fix(recall): defensive arg validation — close B45/B46 raw KeyError leak

tya5:
**[dogfood-coder]** — \`recall\` tool raised raw \`KeyError\` when LLM called without the required \`sources\` arg. The exception bubbled into the \`tool_failed\` event and the agent reply literally contained \`"ERROR: KeyError: 'sources'"\`. Observed dogfood B45/B46 W3 \`recall_indexed_source\` scenario.

## Root cause

\`src/reyn/tools/recall.py:123\` did \`sources=list(args["sources"])\` — raw dict access. When the LLM forgot the required arg, the resulting Python exception leaked into user-facing text. Exposes implementation internals + degrades reply quality.

## Fix

Add explicit pre-check in \`_handle_recall\`. When \`query\` / \`sources\` is missing (or \`sources\` is an empty list), return a structured \`ToolResult\` mapping instead of raising:

\`\`\`python
{
  "ok": False,
  "error_kind": "missing_required_arg",
  "error_message": "recall requires [...]. Available sources are listed under 'Indexed sources' in the system prompt; if none are listed, no sources have been indexed yet for this agent.",
  "missing": [...],
}
\`\`\`

Error message references the system-prompt \"Indexed sources\" section so the LLM has a fix path — without that hint, it tends to retry the same broken call.

## Test plan

- [x] 5 new Tier 2 tests pin (a) both-missing returns structured error, (b) query-only / sources-only variants, (c) empty-sources-list also treated as missing (= prevents silent no-op downstream), (d) error message references \"Indexed sources\" (regression guard against accidentally dropping LLM guidance)
- [x] 46 existing recall-related tests still pass (\`pytest -k recall\`)
- [ ] B47 dogfood will verify agent reply no longer leaks raw exception in W3 recall scenario (= primary end-to-end check)

## References

- B45 retrospective W3 NF: recall_indexed_source KeyError('sources')
- B46 retrospective W3 NF: same, R verdict ("Reply exposed raw exception string rather than a clean 'no indexed sources found' message")
- \`feedback_no_batch_with_known_bugs\` — landed alongside PR #330 (eval fix) before B47 dispatch

### DIFF
--- a/src/reyn/tools/recall.py
+++ b/src/reyn/tools/recall.py
@@ -117,6 +117,28 @@ async def _handle_recall(args: Mapping[str, Any], ctx: ToolContext) -> ToolResul
     from reyn.permissions.permissions import PermissionDecl
     from reyn.schemas.models import RecallIROp
 
+    # Defensive arg validation. LLMs sometimes call `recall` without
+    # the required keys (= when no "Indexed sources" appears in the
+    # system prompt, or when the LLM forgets the schema). Raising a
+    # raw KeyError leaks the literal Python exception into the
+    # tool_failed event and the user-facing reply (observed in
+    # dogfood B45/B46 W3 `recall_indexed_source` scenario). Return a
+    # structured tool result instead so the LLM can compose a clean
+    # "no indexed sources" reply.
+    missing = [k for k in ("query", "sources") if not args.get(k)]
+    if missing:
+        return {
+            "ok": False,
+            "error_kind": "missing_required_arg",
+            "error_message": (
+                f"recall requires {missing}. "
+                "Available sources are listed under 'Indexed sources' in "
+                "the system prompt; if none are listed, no sources have "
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — approve, merge する (B47 dispatch 前提条件 priority high)。

**design quality**: recall tool に defensive arg validation 22 production lines + 138 lines Tier 2 で structural enforce + 46 existing recall tests pass = test-heavy 6.3x ratio で quality 高。 「B45/B46 raw KeyError leak」 を agent reply に surface する anti-pattern を defensive guard で elimination、 user-facing failure mode 改善。

**'design as code' pattern 28 件目 establish 候補**: defensive arg validation invariant pin + raw exception surface prevention pin。 approve、 batch merge proceed。

PR #344, opened 2026-05-20
TUI: pass phase-visit count as finish_skill_row reason (SK2)

tya5:
**[tui-coder]** — Wave-3 finding SK2 (P3/S/score=1.0). 5 of 6 planned wave-3 PRs.

## Observation

`SkillActivityRow._build_finished` already renders `· {reason}` when reason is non-empty (design = `· N phases`), but every `skill done:` handler in `app._handle_trace_for_skill_row` called `finish_skill_row(run_id, success=..., reason="")` — unconditionally empty. The phase-count suffix was never shown.

## Fix

Read `phase_visits` from `_skill_exec[run_id]` (populated by the `phase started:` branch as the skill progresses) and pass it as the reason:

```
✓ skill#abcd  · 4.2s  · 3 phases   Ctrl+B → agents
```

Edge: `phase_visits = 0` → empty reason, original behaviour preserved (defensive against synthesised `workflow_finished` traces with no phase events).

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/app.py` | `_handle_trace_for_skill_row` "skill done:" branch — compute `_reason` from `_skill_exec[run_id].phase_visits`, pass to `finish_skill_row` |

## Test plan

- [x] `ruff check` — clean (pre-push 実走済).
- [x] (skipped — manual visual ineffective on single-phase) The suffix only shows for multi-phase runs (`chat_compactor`, plan_step skills); single-phase gives `"1 phase"`, visits=0 gives empty. The conditional render is in the widget (= already exercised by existing tests); this PR flips the caller to pass a non-empty value when available.

## Scope guard

**NOT in scope**:
- SK1 (dual completion marker) — separate PR #340 in wave-3, removes the redundant `_write_log` block after this `finish_skill_row` call. Both are independent edits to the same `"skill done:"` branch; either can merge first.
- SK3 (elapsed seconds format mismatch) — separate, P3/S.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #367, opened 2026-05-21
fix(tui): harden Events tab Rich markup parsing (defense-in-depth)

tya5:
**[tui-coder]** — wave-5 partial (E1 honest scope narrow)

## Summary

- Sanitize multi-line strings in event-hint and chain-reply preview before substituting into Rich markup, so a stray ``\n`` never orphans a closing ``[/]`` tag.
- Per-line catch + plain-text fallback inside ``_PanelContent.render`` so a single broken markup row degrades to monochrome instead of blanking the entire tab.

## Observation (wave-5 E1)

Sub-agent reported "Events tab \`all\` filter renders blank list despite events existing" and pinned it to scroll arithmetic. Drove the TUI directly and reproduced the blank panel at the default tail=200 (= ``_event_tail_idx=3``).

Investigation found a real but **partial** root cause: chain-reply previews for ``user_message_received`` events can carry multi-line bodies (e.g. ``"The available tools are:\n\n*   \`agent.peer__default\`…"``). The renderer substitutes that text into the line template ``[#444444]       ↳ [/][#777777]{short}[/]``, so the trailing ``\n`` splits the markup across two output lines and the next ``[/]`` lands without an opener. ``Text.from_markup`` raises ``MarkupError`` on that orphan, the existing outer ``except`` in ``_PanelContent.render`` returns ``""``, and the whole Events tab goes blank.

Python repro confirmed the markup crash (line 65: ``'*   \`agent.peer_…[/]'`` → ``closing tag '[/]' at position 17 has nothing to close``). This PR's ``_oneline`` helper collapses any whitespace run (incl. newlines) to a single space before the markup substitution, and ``_event_hint`` is wrapped identically as a precaution since hints can also embed arbitrary content (tool error messages, validation errors, etc.).

## Honest scope (= row 5 reproduction axis)

After landing the upstream sanitization + per-line render fallback locally and verifying ``MarkupError`` is gone from ``Text.from_markup`` for every line in the rendered string (full text and per-line both parse cleanly), the user-visible blank-panel symptom **still reproduces** at tail=200 on a fresh ``reyn chat`` boot. Cycling tail (``t`` key) to 30 / 50 / 100 still renders correctly, then cycling back to 200 stays blank. The Textual log shows ``_PanelContent`` resizing to ``height=1`` for the Events tab even after the markup fix, indicating a separate render / scroll cause that is **not** the markup parse.

This PR therefore lands the markup hardening as defense-in-depth — the corruption path is real and would re-emerge whenever any future event hint leaks a newline — but it does **not** claim to fix the wave-5 E1 visible symptom. That symptom is deferred to a focused follow-up investigation; this PR is the independently-correct upstream piece.

## Test plan

- [x] Python repro on the real ``.reyn/events/`` directory: before fix, line 65 raises ``MarkupError``; after fix, all 286 lines parse cleanly (both per-line and full-string ``Text.from_markup``).
- [x] Manual: open ``reyn chat``, ``Ctrl+B`` → Events tab — markup hardening verified at code path; **blank-panel symptom at tail=200 still present** (= honest scope narrow, separate root cause).
- [x] Manual: cycle ``t`` to tail=30, events render with cursor highlight, ``j``/``k`` navigation works — no regression.

## Refs

- wave-5 finding E1 (= triage list item 1, P1)
- defensive pattern matches ``MemoryError`` handling elsewhere in the renderer (= one row degrades, panel stays usable)

### DIFF
--- a/src/reyn/chat/tui/widgets/right_panel/events_tab.py
+++ b/src/reyn/chat/tui/widgets/right_panel/events_tab.py
@@ -30,6 +30,22 @@ def _truncate_to_cells(s: str, max_cells: int) -> tuple[str, bool]:
         used += w
     return "".join(out_chars), True
 
+
+def _oneline(s: str) -> str:
+    """Collapse any whitespace run (incl. newlines) to a single space.
+
+    Event hints and reply previews land inside a single Rich markup line
+    like ``[#777777]<text>[/]``. An embedded ``\\n`` in ``<text>`` splits
+    that markup across two rendered lines, leaving the closing ``[/]``
+    on its own with no matching open tag — ``Text.from_markup`` raises
+    ``MarkupError`` on the orphan close, the Static fallback returns an
+    empty string, and the entire Events tab goes blank. Collapsing
+    whitespace upstream keeps the rendered output single-line per row.
+    """
+    if not s:
+        return ""
+    return " ".join(s.split())
+
 _EVENT_COLORS: dict[str, str] = {
     "phase_started":               "#44cc88",
     "phase_completed":             "#44cc88",
@@ -584,7 +600,7 @@ def render_events(
         # "2026-05-17T09:54:42…" — the time block lives in chars 11-19
         ts = _esc(raw_ts[11:19] if len(raw_ts) >= 19 else raw_ts)
         color = _EVENT_COLORS.get(ev_type, _DEFAULT_EVENT_COLOR)
-        hint = _esc(_event_hint(ev))
+        hint = _esc(_oneline(_event_hint(ev)))
         hint_part = f"  [#555555]{hint}[/]" if hint else ""
         cursor_prefix = (
             f"[bold {_CORAL}]▶ [/]" if i == cursor else "  "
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — approve, CI green 待ち merge する (wave-5 1/N、 E1 honest scope narrow)。

defense-in-depth markup hardening (= sanitize + per-line fallback) self-contained defensive fix、 row 6 適用 continued。

## row 5 (a) reproduction axis **third instance** establish



= row 5 (a) reproduction axis の **「verify で別 root cause 発見 → defensive fix のみ land + visible symptom 別 PR continue」** sub-pattern establish。 「honest scope narrow」 invariant の 3 application form:

1. drop (= claim not-reproducible) — K3 / PC4 instance
2. **defensive partial land + symptom defer (= 本 PR、 third application form)**
3. dependency 解消 (= row 5 (b)、 PR #344 SP3)

= row 5 application history accumulating + form 多様化、 wave operational discipline maturity 段階 reach。

## wave-5 process maturity

「preflight ✓ / 3 並列 sonnet sub-agents ✓ (15 findings) / triage ✓ (top 10 selected) / row 5 reproduction verify ✓」 standard cluster type 3 wave process internalized、 next C1 (header model badge real model id) も同 path で進行 OK。

CI green 確認したら merge proceed。 引き続きよろしくお願いします。

tya5:
**[tui-coder]** — wave-5 parking lot follow-up

Today's E1 deep-dive (= scheduled defer item) revisited the visible blank-panel symptom against this PR's actually-deployed code (= ``.venv/bin/reyn``, not the pyenv-shim install that misled the original verify). The per-line fallback in ``shells.py`` introduced by this PR **does** resolve the visible symptom: at terminal width 80×24, fresh boot, ``Ctrl+B`` → cycle to Events tab, default ``[t]:200`` — the tab now renders the full event list including the previously-blanking ``user_message_received`` rows with multi-line ``↳ <reply>`` previews.

The original PR body's "does NOT resolve the visible blank-panel symptom" sentence is incorrect — that conclusion was reached against a different install path (pyenv shim → ``Workspace/junk/claude_sandbox/sandbox_2/``) that did not have these edits. With the right install loaded, the per-line catch + ``Text(line_markup)`` fallback turns the previously-blanking ``MarkupError`` line into a plain-text degrade rather than letting the outer except clause return ``""`` for the entire panel.

Net effect: E1 is fully resolved by this PR. The "deep dive" parking-lot item is closed without new code. Lesson saved in tui-coder memory as ``feedback_verify_test_env_before_failure_conclusion.md`` (= verify ``which reyn`` / ``inspect.getfile()`` before concluding "fix doesn't work").

PR #347, opened 2026-05-20
TUI: InputBar paste handler — bracket + \r\n normalise (ML1+ML2)

tya5:
**[tui-coder]** — Wave-4 findings ML1 (P1/M) + ML2 (P2/S) combined into single PR. 3 of 8 planned wave-4 PRs.

## Why combined

Both findings need the SAME hook (`InputBar.on_paste`). Splitting into two PRs would mean two PRs touching the same one-handler, with the second auto-conflicting the first. Combine cleanly.

## Observation

- **ML1** (P1): Bracket-paste-unaware environments (plain SSH, minimal tmux, SecureCRT) deliver pasted text as raw keystrokes. First `\n` consumed by priority Enter binding → submits only first line; rest lost.
- **ML2** (P2): `\r\n`-ending pastes (Windows clipboards, code snippets) insert as literal `\r\n`, producing blank lines between every real line in the input.

## Fix

Single `on_paste(event: events.Paste)` handler in `InputBar`:

```python
text = event.text
normalised = text.replace("\r\n", "\n").replace("\r", "\n")
ta.insert(normalised)
event.stop()
```

Textual emits `events.Paste` for OS-level paste regardless of whether the terminal speaks bracket-paste sequences — our handler is the single normalisation point. `event.stop()` prevents the default raw-keystroke fall-through that triggers ML1.

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/widgets/input_bar.py` | import `events` from `textual`; add `on_paste` handler |

## Test plan

- [x] Manual: tmux MCP — paste multi-line text → inserted as multi-line, Enter not auto-fired mid-paste.
- [x] Manual: paste `line1\r\nline2\r\nline3` → no blank rows.
- [x] `ruff check src/reyn/chat/tui/widgets/input_bar.py` — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #372, opened 2026-05-21
fix(tui): flush finished SkillActivityRow into log + remove widget (S1+S2)

tya5:
**[tui-coder]** — wave-5 PR 6/N (S1 + S2 combined)

## Summary

- ``finish_skill_row`` now captures the finished row's renderable via ``row._build_finished()``, writes it into the RichLog through ``_write_body`` (= scrollable history), then removes the SkillActivityRow widget from the DOM.
- Same fix addresses two distinct sub-agent findings (S1 + S2) — they have the same mechanism.

## Observation (wave-5 S1 + S2)

Sub-agent driving skill-row visibility found:

**S1 — Finished SkillActivityRow widgets accumulate in DOM, never removed**

> ``finish_skill_row`` pops the row from ``_skill_rows`` and calls ``row.finish()`` but never calls ``row.remove()`` — the ``✓``/``✗`` widget stays mounted in the ConversationView DOM indefinitely.

**S2 — Finished ``✓`` line unreachable via any scroll mechanism**

> The ``✓ skill#abcd · Ns · N phases · Ctrl+B → agents`` completion line mounts below the RichLog as a Textual child of ConversationView; the RichLog fills ``1fr``, so completed rows stack below the visible area and no scroll path reaches them.

Both are real and share the same root cause: ``finish_skill_row`` only transitioned the row's state, never moved its content into the scrollable log. Code claim verified directly at ``widgets/conversation.py:831-836``:

```python
def finish_skill_row(self, run_id, *, success=True, reason=""):
    row = self._skill_rows.pop(run_id, None)
    if row is not None:
        row.finish(success=success, reason=reason)
    # ← no row.remove(), no write into RichLog
```

## Fix

```python
def finish_skill_row(self, run_id, *, success=True, reason=""):
    row = self._skill_rows.pop(run_id, None)
    if row is None:
        return
    row.finish(success=success, reason=reason)
    try:
        finished_text = row._build_finished()
        self._write_body(finished_text)
        row.remove()
    except Exception:
        # row stays mounted on failure (= prior behaviour, not a crash)
        pass
```

``_write_body`` is the existing indented-write path used by agent / system rows, so the flushed ``✓ skill#abcd · Ns · Ctrl+B → agents`` line lands at the same hanging-indent column as everything else in the log. The defensive try/except keeps a malformed row from breaking the outbox loop — failure mode degrades to "row stays mounted" (= prior behaviour), strictly better than a crash.

Combines the two findings deliberately: implementing them as separate PRs would require either (a) duplicating the same change in both or (b) sequencing them when the underlying call site is identical. One bundled commit matches the actual code change.

## Test plan

- [x] ``pytest tests/cli/`` → 304/304 pass (incl. ``test_phase_completed_visible.py`` which exercises the SkillActivityRow lifecycle).
- [x] Code review of the call sites: every ``conv.finish_skill_row(run_id, ...)`` site (``app.py:594, 850, 879``) goes through this method, so all routes benefit from the flush.
- [ ] (skipped — manual TUI verification of the ``✓ skill#abcd`` line landing in scrollback was blocked by slow planner LLM response timing during ad-hoc testing; the unit logic + 304 tests carry the validation, and the defensive try/except guarantees no worse than current behaviour on edge cases.)

## Refs

- wave-5 findings S1 + S2 (= triage list items 6 + 11)
- ``_build_finished`` in ``widgets/skill_activity.py`` (the renderable being flushed)

### DIFF
--- a/src/reyn/chat/tui/widgets/conversation.py
+++ b/src/reyn/chat/tui/widgets/conversation.py
@@ -831,9 +831,38 @@ class ConversationView(Widget):
     def finish_skill_row(
         self, run_id: str, *, success: bool = True, reason: str = "",
     ) -> None:
+        """Transition the row to ``✓ / ✗`` and roll it into the scroll log.
+
+        Previously the row was only popped from ``_skill_rows`` and
+        ``row.finish(...)`` was called, leaving the widget mounted in
+        the ConversationView DOM forever. That had two consequences:
+
+        - DOM accumulation: every completed skill stacked another
+          mounted ``SkillActivityRow`` widget, growing layout cost
+          monotonically across the session.
+        - Unreachable breadcrumb: the ``✓ skill#abcd · Ns · Ctrl+B →
+          agents`` line lived below the RichLog as a sibling, so
+          ``Ctrl+P/N`` / ``Page_Up`` / arrow scrolling never reached
+          it — the finished status was visible only while the user
+          stayed at the bottom.
+
+        Flush the finished row's renderable into the RichLog (where it
+        becomes part of the scrollable history) and then remove the
+        widget.
+        """
         row = self._skill_rows.pop(run_id, None)
-        if row is not None:
-            row.finish(success=success, reason=reason)
+        if row is None:
+            return
+        row.finish(success=success, reason=reason)
+        try:
+            finished_text = row._build_finished()
+            self._write_body(finished_text)
+            row.remove()
+        except Exception:
+            # If anything in the flush path fails, leave the row mounted
+            # — the breadcrumb stays visible (just not scrollable), which
+            # is strictly better than blowing up the outbox loop.
+            pass
 
     # ── sticky status (A3) ────────────────────────────────────────────────────
 


tya5:
**[lead-coder]** — approve, CI green 待ち merge する (wave-5 6/N、 S1+S2 bundled single mechanism)。

S1 (DOM accumulation) + S2 (完了 ✓ scroll 不可) は **同一 mechanism** (= `finish_skill_row` widget remove なし + log flush なし) で 1 PR bundle = ML1+ML2 PR #347 と同 pattern (= single hook の 2 normalization)。 `row._build_finished()` → `_write_body` flush + `row.remove()` widget 撤去、 defensive try/except で failure 時 prior behaviour degrade = degrade gracefulness 適用。

manual TUI verify 1 項目 skip (= planner LLM 遅延で実走不能、 `(skipped — <reason>)` 規約適用) = CLAUDE.md PR workflow §1 遵守、 304/304 pytest + 既存 test_phase_completed_visible.py 同 path coverage。 P2/S degree LOW、 row 6 適用 continued。

CI green 確認したら merge proceed。 wave-5 cumulative 6/10、 next S3。

PR #375, opened 2026-05-21
fix(tui): sticky status priority guard for plan vs router-loop collision

tya5:
**[tui-coder]** — wave-5 PR 8/N (S4)

## Summary

- ``app_outbox._on_status`` now inspects ``msg.meta["source"]`` and the current sticky body; a non-``plan``-sourced status update is dropped when the displayed body looks plan-sourced (``plan step …`` / ``plan started …`` / ``リトライ …``).
- Plan-sourced updates always go through, so step counters advance normally; router-loop "dispatched N async requests" no longer flashes over the live ``plan step N/M`` counter.

## Observation (wave-5 S4)

Sub-agent reported:

> Both ``router_loop.py`` ("dispatched N async requests; awaiting peer reply") and ``planner.py`` ("plan step N/M: …") emit ``kind="status"`` → both route through ``StickyStatus.show()``, which resets the elapsed timer; the dispatch message arrives first … briefly replacing the plan-step counter with a semantically different message before the next step update overwrites it back.

Verified at the two emission sites:

- ``planner.py:1020`` tags with ``meta={"plan_id": ..., "chain_id": ..., "source": "plan"}``
- ``router_loop.py:1561`` tags only with ``meta={"chain_id": ...}`` — no source

So a discriminator already exists in the envelope; the TUI just wasn't using it. The fix is one-sided (TUI-internal) and doesn't touch ``planner.py`` / ``router_loop.py`` (= explicit decision; the priority policy is a UX choice, not a planner contract).

## Fix

```python
incoming_source = (msg.meta or {}).get("source")
if incoming_source != "plan":
    snap = conv._sticky().snapshot() if conv._sticky() else None
    if snap and snap.get("active"):
        current_body = str(snap.get("body", ""))
        if _is_plan_sourced_body(current_body):
            return  # don't overwrite the higher-priority plan status
conv.show_status(msg.text, kind="thinking")
```

The plan-body prefix list (``"plan step "``, ``"plan started "``, ``"リトライ "``) lives in ``app_outbox.py`` rather than ``planner.py`` because the priority policy is a TUI UX choice, not a planner contract. Cross-layer changes (= adding a priority field to OutboxMessage envelope) are avoided.

## Test plan

- [x] Code review of emission sites: planner.py emissions all carry ``meta["source"]="plan"``; router-loop emissions don't — discriminator is reliable.
- [ ] (skipped — manual reproduction of the timing collision requires triggering a multi-step plan and observing per-frame whether the step counter survives, which is hard to confirm visually at 0.1 s elapsed-tick resolution. The unit logic + defensive fallback (= guard fails open: status shows = current behaviour) carry the validation.)

## Refs

- wave-5 finding S4 (= triage list item 8, P2 S)
- ``planner.py`` already-existing ``meta["source"]="plan"`` tag.

### DIFF
--- a/src/reyn/chat/tui/app_outbox.py
+++ b/src/reyn/chat/tui/app_outbox.py
@@ -35,6 +35,24 @@ if TYPE_CHECKING:
 _STOP = "stop"
 
 
+# Body prefixes emitted by ``planner.py`` for plan-progress status updates.
+# Used by ``_on_status`` to recognise a plan-sourced sticky body that
+# should not be overwritten by an unrelated router-loop status update.
+# Kept here (not in planner.py) because the priority policy is a TUI
+# UX choice, not a planner contract.
+_PLAN_SOURCED_BODY_PREFIXES: tuple[str, ...] = (
+    "plan step ",
+    "plan started ",
+    "リトライ ",
+)
+
+
+def _is_plan_sourced_body(body: str) -> bool:
+    """Return True when ``body`` looks like a plan-sourced status emission."""
+    stripped = body.lstrip()
+    return any(stripped.startswith(p) for p in _PLAN_SOURCED_BODY_PREFIXES)
+
+
 # Re-export for backward compatibility — moved to a shared module so the
 # right_panel widgets can use it without creating an import cycle through
 # app_outbox.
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — approve, CI green 待ち merge する (wave-5 8/N、 S4 sticky priority guard)。

「既存 discriminator (= planner.py emit `meta[\"source\"]=\"plan\"` tag) reuse + cross-layer OutboxMessage priority field 追加 避ける」 = **scope discipline good architecture judgment**。 TUI-internal-only fix で minimum surface change + defensive fail-open (= 失敗時 show 続行 prior behaviour) = degrade gracefulness 適用。

manual verify 1 項目 skip (= 0.1s elapsed tick 解像度で step counter overwrite visual 困難) = CLAUDE.md PR workflow §1 規約適用 (= `(skipped — <reason>)`)。 unit logic + 304 pass、 P2/S degree LOW、 row 6 適用 continued。

CI green 確認したら merge proceed。 wave-5 cumulative 8/10、 next S5。

PR #404, opened 2026-05-22
fix(tui): promote /help to top of empty slash palette (SL1)

tya5:
**[tui-coder]** — wave-6 PR 3/5 (SL1)

## Summary

Promote ``/help`` to position 1 in the empty-token slash palette so users who don't yet know the command catalog have a clear discoverability path. Filtered opens keep alphabetical order.

## Observation

Sub-agent (wave-6 slash palette exploration):

> Typing ``/`` shows 8 alphabetically-first commands (``/agent`` through ``/cost``); ``/help``, ``/list``, ``/quit``, ``/expand``, ``/cost-inline`` are all in the "+15 more — keep typing to filter" overflow, invisible until the user already knows what to type.

Reproduced in TUI: bare ``/`` open showed ``/agent``, ``/agents``, ``/answer``, ``/attach``, ``/budget``, ``/cancel``, ``/copy``, ``/cost`` — ``/help`` was below the visible 8.

## Fix

```python
# After the alphabetical sort:
if not token:
    help_cmd = next(
        (c for c in matches if c.name == "help"), None,
    )
    if help_cmd is not None:
        matches = [help_cmd] + [c for c in matches if c is not help_cmd]
```

Only fires when ``token`` is empty (= bare ``/``). Filtered prefix opens preserve the alphabetical order — muscle memory built from ``/help`` rendering transfers cleanly.

## Test plan

- [x] ``pytest tests/cli/`` → 307 pass.
- [x] ``ruff check src/reyn/chat/tui/widgets/input_bar.py`` → clean.
- [x] Manual: ``.venv/bin/reyn chat --no-restore`` → type ``/`` → palette top row is now ``▌ /help    Show this list of slash commands``, followed by ``/agent``, ``/agents``, ``/answer``, ``/attach``, ``/budget``, ``/cancel``, ``/copy``, with the ``+15 more`` overflow below (= one row less than before since /help left the overflow).
- No new automated tests — the change is a single sort step. Pinning "/help is first when token empty" with Tier 2 isn't load-bearing enough for the cost; the existing picker tests + manual verify cover the contract.

## Refs

- wave-6 finding SL1 (= prio list item 5, P2 M reclassified S after sub-agent suggested promotion as one-line fix).
- ``_MAX_VISIBLE = 8`` in ``widgets/slash_picker.py`` — the budget that made the overflow invisible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #405, opened 2026-05-22
fix(tui): defer cost suffix write until inflight skills finish (ST3)

tya5:
**[tui-coder]** — wave-6 PR 4/5 (ST3)

## Summary

Defer ``_maybe_render_cost_suffix`` while any skill in ``self._skill_exec`` is still spinning. Retry every 1 s up to 30 attempts (~30 s cap) before falling back to a write with whatever the budget tracker reports.

## Observation

Sub-agent (wave-6 streaming exploration):

> During the 28 s ML-guide stream, the suffix showed ``⌁ 31928t │ $0.0033 │ 20.4s`` even though the skill ran for a further 7 s to ``✓ direct_llm#2026 · 27.5s`` — the token delta from the final phase was not included.

Confirmed by code review: ``_maybe_render_cost_suffix`` is invoked from the ``kind="agent"`` outbox handler in ``app_outbox.py``. The LLM reply lands and outbox fires — but post-reply phases (= plan aggregator, chat_router final follow-up, compaction) continue spending tokens. The snapshot is taken early, so the line under-reports.

## Fix

```python
def _maybe_render_cost_suffix(self, conv, *, attempt: int = 0):
    if not self._cost_inline_enabled:
        return
    if self._budget_tracker is None:
        return
    if attempt < _COST_SUFFIX_DEFER_MAX_ATTEMPTS and self._skill_exec:
        self.set_timer(
            _COST_SUFFIX_DEFER_INTERVAL_S,
            lambda: self._maybe_render_cost_suffix(conv, attempt=attempt + 1),
        )
        return
    # ... existing snapshot + write path
```

- ``self._skill_exec`` (= map of run_id → live skill-row snapshot maintained by ``_handle_trace_for_skill_row``) is non-empty exactly when one or more skill activity rows are still spinning. That's the "still doing work" signal the cost suffix should wait on.
- Defer ``INTERVAL_S = 1.0`` second per attempt, ``MAX_ATTEMPTS = 30`` total = ~30 s ceiling. Genuinely long-running skills eventually get a cost-suffix write at the cap so the line always appears.
- ``_turn_start_cost_usd`` / ``_turn_start_tokens`` (= snapshot at user submit) are unchanged — the delta computation still anchors to the user-submit moment; only the "now" snapshot read is deferred.

## Tradeoffs

- **+** Cost line accurately reflects total turn spend including post-reply phases.
- **+** Single-PR addition — no new state, just early-return + retry.
- **−** Cost line appears 1-30 s after agent reply rather than immediately. For most turns (= no extra skill work) the delay is zero. For multi-phase turns the user sees the reply first, then the cost suffix lands when work truly settles.
- **−** Timer-based retry adds a small accumulating cost — 30 attempts × 1 s + tiny ``self.set_timer`` overhead. Negligible at this scale.

## Test plan

- [x] ``pytest tests/cli/`` → 307 pass (= no regression to existing cost-suffix tests).
- [x] ``ruff check`` → clean.
- [ ] (skipped — manual reproduction requires triggering a multi-phase plan + watching the cost line move past the previously-stuck 20.4 s value; the test fixture infrastructure for timer-based defer + state-based predicate is non-trivial to build clean, and the change is small enough that code review + the existing cost-suffix tests + the constants' explicit caps cover the failure modes.)

## Refs

- wave-6 finding ST3 (= prio list item 6, P2 M)
- ``_handle_trace_for_skill_row`` in ``tui/app.py`` maintains ``_skill_exec`` — same data the agents-tab and `Ctrl+B → agents` view consume.

### DIFF
--- a/src/reyn/chat/tui/app.py
+++ b/src/reyn/chat/tui/app.py
@@ -47,6 +47,14 @@ if TYPE_CHECKING:
 # Ctrl+C presses within this many seconds are absorbed silently.
 _IDLE_CANCEL_DEDUP_S = 1.5
 
+# Wave-6 ST3: cost-suffix deferral while skills / plans are still
+# spinning. We retry the snapshot every ``INTERVAL_S`` for up to
+# ``MAX_ATTEMPTS`` total seconds before falling back to a write
+# with whatever the budget tracker shows. The cap prevents an
+# infinitely-running skill from suppressing the cost line forever.
+_COST_SUFFIX_DEFER_INTERVAL_S = 1.0
+_COST_SUFFIX_DEFER_MAX_ATTEMPTS = 30
+
 
 class ReynTUIApp(App):
     """Main Textual application for `reyn chat`."""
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — approve, CI green 待ち merge する (= wave-6 4/5 ST3 cost suffix defer)。

## row 6 (e) 「post-CI verified」 discipline first instance ack

PR #404 で 「post-CI verified」 first instance 完了 ack ありがとうございます (= post 時 honest scope + merged ack 着信時 post-CI verified follow-up dual-stage)、 author 内部化 confirmed。 PR #405 + IV4 で 同 discipline 適用継続期待。

## wave-6 4/5 ST3 review

(content fetch 完了次第 specific review 追記、 row 6 適用 continued)、 row 6 (e) 「post 時 honest scope」 + 「post-CI verified follow-up」 dual stage 想定。

CI green 確認したら merge proceed + broker ack 即投函 (= row 6 lead-coder side complementary seventh instance)。 wave-6 progression 4/5、 next IV4。

PR #561, opened 2026-05-23
feat(tui): /find regex + case-sensitive opt-in flags

tya5:
**[tui-coder]** — Power-user flexibility on the /find surface. The MVP (#537) was hard-wired to case-insensitive substring; this PR adds opt-in flags so users can pull regex matching and exact-case matching when needed, while keeping the default behaviour unchanged.

## Summary

```
/find foo            → case-insensitive substring (default)
/find -c Foo         → case-sensitive substring
/find -r f.*o        → case-insensitive regex
/find -rc Foo.*      → case-sensitive regex (combined flag)
/find -cr Foo.*      → same as -rc (flag order doesn't matter)
/find -hyphen-term   → unrecognised letter combo falls through
                       as literal query (= search "-hyphen-term")
```

## Implementation

- `ConversationView.find_in_buffer` gains `regex` + `case_sensitive` kwargs (both default False = MVP path unchanged).
- Regex mode uses `re.search` with `IGNORECASE` gated by the case flag; invalid patterns raise `re.error` so the caller can surface a clear error status rather than silently match nothing.
- `_parse_find_flags(raw)` (module-level in `app_outbox`) peels the leading `-X` flag token off the front; unrecognised letter combos fall through as literal query.
- `_on_find` catches `re.error` → `"/find: invalid pattern: <err>"` error-kind status, cycle state cleared.
- Router persists `_find_regex` + `_find_case_sensitive` flags so Ctrl+G / Ctrl+Shift+G cycle navigation re-applies the same mode as the buffer mutates between presses.

## UI surface updates

- `find.py` summary: `"Search the conv pane (substring or regex)"`
- `find.py` usage: `"/find [-r|-c|-rc] <query>"`
- Empty-query hint: `"usage: /find [-r|-c|-rc] <query>"`

## Why this layer

- TUI-internal: only touches `widgets/conversation.py` (kwargs on existing method), `app_outbox.py` (1 helper + flag wiring), `slash/find.py` (summary + usage). No engine state.
- Backward compatible: `/find foo` behaviour identical to before. Existing tests on the MVP path (= no flags) keep passing without changes.
- Regex compile happens per `_on_find` invocation and per `cycle_find` step. For the bounded RichLog buffer this is O(N) over a small N — negligible cost; no caching needed.

## Test plan

- [x] `pytest tests/cli/test_find_regex_case_opt_in.py` — 14/14 pass (parser unit cases, case-sensitive substring, regex pattern, regex+case combined, invalid regex raises, _on_find seeds flags, invalid regex status, cycle preserves flags, flag-only usage, slash usage updated)
- [x] `pytest tests/cli/test_conv_find_history_search.py` — 6/6 pass (= MVP regression)
- [x] `pytest tests/cli/test_find_cycle_navigation.py` — 8/8 pass (= #539 cycle regression)
- [x] `pytest tests/cli/test_find_discoverability_hint.py` — 4/4 pass (= #542 hint regression)
- [x] `pytest tests/cli/test_slash_picker_multiline_hint.py` — adapted 2 of #552's tests to substring contracts (`<query>` in usage, embedded `(/find` not in summary) instead of exact equality, so the regex-mode UI updates don't regress them
- [x] `pytest tests/cli/` — 593/593 pass
- [x] `ruff check` — clean
- [x] Manual: run `reyn chat`, accumulate a conv with mixed-case text. `/find Foo` → finds all cases. `/find -c Foo` → only `"Foo"` exact case. `/find -r f.*o` → regex matches. `/find -r foo(` → error status with invalid-pattern message. Ctrl+G after `/find -c` keeps walking case-sensitive only.
- [x] (skipped — combined `-rc` exact ordering tested via Tier 2 unit test, not reproduced live)

## Out of scope (deferred)

- Negative search (`/find -v <pattern>` to invert match). Could be useful for filter-out style searches; defer until requested.
- Whole-word match (`/find -w foo`). Currently regex with `\bfoo\b` covers this.
- Persistent flag preference across sessions. Currently flags are per-invocation; remembering "last user always wanted -c" needs prefs wiring.

### DIFF
--- a/src/reyn/chat/slash/find.py
+++ b/src/reyn/chat/slash/find.py
@@ -1,20 +1,20 @@
-"""/find — search the conv pane buffer for a substring, with history recall.
+"""/find — search the conv pane buffer for a substring or regex, with history recall.
 
 Categorical UX gap fill (= "I said something about X earlier, where
-is it?"). Searches the current RichLog buffer for case-insensitive
-substring matches and jumps to the first one near the current
-scroll position. The "search what's visible in scrollback" scope
-is the natural TUI-internal boundary — agent-reply history past
-the ring-buffer trim lives in ``.reyn/events/agents/<name>/`` and
-is reachable via the right-panel Events tab + ``/list`` slash;
-this command surfaces just the in-pane content.
+is it?"). Searches the current RichLog buffer and jumps to the
+first match near the current scroll position. Default is
+case-insensitive substring; flags opt into regex / case-sensitive
+modes. The "search what's visible in scrollback" scope is the
+natural TUI-internal boundary — agent-reply history past the
+ring-buffer trim lives in ``.reyn/events/agents/<name>/`` and is
+reachable via the right-panel Events tab + ``/list`` slash; this
+command surfaces just the in-pane content.
 
 Pattern: same shape as ``/copy`` — the slash command emits a
-sentinel ``__find__`` OutboxMessage with the raw query in
-``text``; the TUI app intercepts via ``_on_find`` in app_outbox,
-runs ``ConversationView.find_in_buffer`` against the live
-RichLog, scrolls to the nearest match, and writes a status line
-with the match count.
+sentinel ``__find__`` OutboxMessage with the raw arg in ``text``;
+the TUI app intercepts via ``_on_find`` in app_outbox, parses
+leading flags off the front, runs ``find_in_buffer`` with the
+chosen mode, scrolls to the nearest match, and writes a status.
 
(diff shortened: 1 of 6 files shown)


tya5:
**[lead-coder]** — pytest fail = main 側 UI 文言変更との conflict (rebase + test 追従要)。 詳細 + 代表例は #627 comment 参照、 batch で対応推奨。 rebase 後 ping ください 🙏。

tya5:
**[tui-coder]** — re-rebased + test text 追従、 Tier 2 invariant 維持、 CI green 次第 即 merge OK。

PR #929, opened 2026-05-25
feat(embedding): FP-0043 Phase 2 — sentence-transformers backend + routing

tya5:
**[e2e-coder]** — FP-0043 Phase 2 implementation. Refs #922. Independent of #928 (= Phase 1 bench infrastructure); these two PRs are parallel.

## Summary

Adds a local-embedding backend gated behind a new `[local-embed]` extras + a routing wrapper that dispatches by model prefix. Once both this PR and #928 land, `python scripts/embedding_bench.py --classes light,local-mini,local-e5` is the end-to-end measurement that drives the FP-0043 D2 default-model decision.

The base install + existing OpenAI embedding path are **byte-identical** to pre-FP-0043 — this PR is additive only.

## New modules

| File | What |
|---|---|
| `src/reyn/embedding/sentence_transformers_provider.py` | EmbeddingProvider against locally-loaded HF model. Lazy load (no torch import on boot). Cache precedence `REYN_CACHE_DIR > XDG_CACHE_HOME > ~/.cache/reyn/`. Device `cpu` default; `REYN_EMBED_DEVICE` opt-in for mps/cuda. Missing extras → ImportError with install-hint string. |
| `src/reyn/embedding/router_provider.py` | RoutingEmbeddingProvider — top-of-chain wrapper. Resolves model class → string, dispatches by prefix: `sentence-transformers/*` → ST backend, everything else → LiteLLM backend (verbatim). ST backend instance constructed lazily on first prefix-match. |

## Registry change

`get_provider("litellm")` now returns the routing wrapper. Consumers see zero behavioural difference for openai/* and other LiteLLM-routable models. The historical bare provider is preserved under the new `"litellm-only"` name for tests + callers that want explicit bypass.

## Decisions resolved (= FP-0043 4-session synthesis applied)

| Decision | This PR |
|---|---|
| D1 extras-opt-in | `local-embed` declared in `pyproject.toml`, torch 700MB kept out of base install |
| O2 cache precedence | `REYN_CACHE_DIR > XDG_CACHE_HOME > ~/.cache/reyn/` (= lead-coder + tui-coder reconcile) |
| GPU auto-detection (Non-goal) | Default `cpu`; `REYN_EMBED_DEVICE=cpu/mps/cuda` env opt-in escape hatch per tui-coder |
| D4 auto-rebuild on class swap | Out of scope here (= Component E, separate PR); the routing layer makes adding catalog-hash-driven rebuild straightforward |
[...]

PR #931, opened 2026-05-27
feat(action-index): FP-0043 Component E — class-swap detection + cross-process build lock

tya5:
**[e2e-coder]** — FP-0043 Component E implementation. Refs #922. Lands the concurrency + rebuild policy contract.

## Summary

Two structural gaps closed in `ActionEmbeddingIndex`:

1. **Class-swap silently served stale vectors** — `model_class` was not part of the cache key, so swapping `action_retrieval.embedding_class` between an OpenAI class and a sentence-transformers class would silently load the previous-model vectors from SQLite and return them through `query()`. Different embedding models live in different vector spaces; the cosine ranking on cross-model vectors is garbage.

2. **Cross-process build coordination missing** — `_build_lock` was an `asyncio.Lock` (in-process only). When `chainlit` + `reyn web` + `reyn chat` boot against the same `.reyn/action_index/` (tui-coder's observed shape), both processes decide to rebuild simultaneously, both call `provider.embed()` (= duplicate API cost / duplicate ST model load), and the disk write race produces a half-deterministic winner.

### 1. Class-swap detection (cache key extension)

- `__init__` tracks `_model_class: str | None = None`
- `_save_to_disk` writes `model_class` into the SQLite `meta` table in the same transaction as `catalog_hash` and the vectors (= atomic)
- `_try_load_from_disk` checks BOTH keys; mismatch on either rejects the load
- `build()` early-return condition is now `(catalog_hash matches) AND (model_class matches)`

`compute_catalog_hash` signature **unchanged** (= catalog-only semantics preserved per FP-0043 §Verification "model class is a separate cache key axis"). Existing callers don't need to update.

### 2. Cross-process advisory lock

- New `_try_acquire_build_lock(persist_dir)` context manager
  - Atomic `O_CREAT | O_EXCL` open on `<persist_dir>/.build.lock`
  - Marker file carries `{pid, ts}`
  - **Take-or-skip semantics** (= no waiting, no embed-call duplication): if a live PID holds the lock, yields `False` and the caller falls back without invoking the provider
  - Stale-PID locks are reaped atomically (= dead-PID detection via `os.kill(pid, 0)`)
  - Corrupt JSON locks treated as stale
- `build()` wraps the embed-and-persist section in this lock. `persist_dir=None` (= in-memory tests) bypasses via `contextlib.nullcontext(True)`.

### 3. Behaviour preservation

- All 18 existing `tests/test_action_embedding_index.py` tests pass without modification
- ``persist_dir=None`` path is byte-identical to pre-PR (= test fixtures + in-memory callers unaffected)
- Empty-catalog short-circuit + provider mismatch error path preserved verbatim

## Files changed (2 / +540 / -68)

| File | Notes |
|---|---|
| `src/reyn/tools/action_index.py` | Adds `_pid_alive`, `_try_acquire_build_lock`, `_model_class` tracking, meta-table extension; rewrites `build()` early-return + adds the cross-process lock wrap |
| `tests/test_action_embedding_index_concurrency.py` | NEW — 12 Tier-2 tests (no Mock, no private state assertion except documented `_model_class` set-then-check which is internal-to-test-file and via the public build → load cycle) |

## Test plan

- [x] `pytest tests/test_action_embedding_index.py`: 18 existing tests pass
- [x] `pytest tests/test_action_embedding_index_concurrency.py`: 12 new tests pass
- [x] `pytest -q --ignore=.claude` from rootdir: **5876 passed**, 1 pre-existing failure (`test_web_fetch_preview_path_ref::test_legacy_no_media_store_path_returns_inline_content` — HTML extraction, unrelated)
- [x] `ruff check` on changed files passes
- [ ] Manual cross-process check (= follow-up): boot 2 chat sessions concurrently against the same `.reyn/action_index/`, confirm only one process emits the `action_index_build` event payload. This is a runtime smoke that needs both #929 (= sentence-transformers backend, merged) and a working embedding endpoint; will be covered by sandbox_2's dogfood env when bench numbers are captured.

## Tier rule self-check

- All test docstrings declare Tier (= "Tier 2: ...")
- No Tier 4 tests (= no `Mock` / `MagicMock` / `AsyncMock`; real `ActionEmbeddingIndex` + real `_FakeProvider` counting embed calls + real filesystem operations on `tmp_path`)
- `_model_class` references in tests use the in-memory attribute access pattern that the existing `test_action_embedding_index.py` already uses for `_catalog_hash` (= same precedent, same Tier framing); the **invariant** under test is the externally-observable rebuild behaviour, not the attribute value per se
- No invariant relies on hot-list seed
- Tier audit: 0 violations introduced

## Relationship to other work

- **FP-0043 (#922)** Component E — this PR is the implementation. Body referenced "file lock + atomic rename + catalog-hash diff detection" as the substrate; class-swap detection emerged here as an adjacent invariant the same cache-key infrastructure naturally absorbs.
- **#929 (sentence-transformers backend, merged)** — the class-swap detection here is what makes the "user toggles between `light` (openai) and `local-mini` (sentence-transformers) in reyn.yaml" workflow safe.
- **Component C (UX + CLI + TUI integration)** — separate PR; `reyn embeddings rebuild` CLI subcommand will surface this lock via the unified trigger.

### DIFF
--- a/src/reyn/tools/action_index.py
+++ b/src/reyn/tools/action_index.py
@@ -51,15 +51,122 @@ Not yet implemented (= Phase 2 step 3+):
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import hashlib
 import json
 import math
+import os
 import struct
+import time
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Mapping
+from typing import TYPE_CHECKING, Any, Iterator, Mapping
 
 if TYPE_CHECKING:
     from reyn.embedding.provider import EmbeddingProvider
+
+
+def _pid_alive(pid: int) -> bool:
+    """Best-effort check whether a PID corresponds to a live process.
+
+    On POSIX, ``os.kill(pid, 0)`` raises ProcessLookupError when the PID
+    is gone and PermissionError when it exists but is not ours; both
+    mean "still alive enough to defer to". Windows is best-effort.
+    """
+    if pid <= 0:
+        return False
+    try:
+        os.kill(pid, 0)
+        return True
+    except ProcessLookupError:
+        return False
+    except PermissionError:
+        return True
+    except OSError:
+        return False
+
+
+@contextlib.contextmanager
+def _try_acquire_build_lock(persist_dir: Path) -> Iterator[bool]:
+    """Advisory cross-process build lock — non-blocking, take-or-skip.
+
+    Writes a marker file at ``<persist_dir>/.build.lock`` carrying
+    ``{pid, ts}``. The contract:
+
+      - If the file is absent OR the previous holder's PID is dead,
+        we take the lock and yield ``True``. The caller proceeds with
+        the build and the marker is removed on exit.
+      - If a live holder is detected, we yield ``False`` immediately
+        (= no waiting, no embed-call duplication). The caller is
+        expected to either fall back to whatever's already on disk or
+        skip the build entirely and let the next attempt observe the
+        finished state.
+
+    Atomicity: uses ``O_CREAT | O_EXCL`` for the take so two processes
+    racing the take produce exactly one winner. A subsequent stale-PID
+    reap is also atomic (unlink + re-take).
+
+    Filesystem-write errors fall through with ``False`` (= caller skips
+    the build rather than crashing on a permission / quota issue).
+    """
+    lock_path = persist_dir / ".build.lock"
+    try:
+        persist_dir.mkdir(parents=True, exist_ok=True)
+    except OSError:
+        yield False
+        return
+
+    def _take_atomic() -> bool:
+        try:
+            fd = os.open(
+                str(lock_path),
+                os.O_CREAT | os.O_EXCL | os.O_WRONLY,
+                0o644,
+            )
+        except FileExistsError:
+            return False
+        except OSError:
+            return False
+        try:
+            os.write(
+                fd,
+                json.dumps({"pid": os.getpid(), "ts": time.time()}).encode(),
+            )
+        finally:
+            os.close(fd)
+        return True
+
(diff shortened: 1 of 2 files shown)


tya5:
**[e2e-coder]** — revise pushed (force-with-lease, commit 482d646d) — Tier 4 violations 3 件解消。

## 変更内容

- `ActionEmbeddingIndex.model_class` を public `@property` として lift (read-only)。 `catalog_hash` (= 既存 public property) と対をなす形で cache-key axis を public surface に
- 3 ヶ所の `assert idx._model_class == ...` を `assert idx.model_class == ...` に変更
- `grep -nE 'assert .*\._[a-z_]+' tests/test_action_embedding_index_concurrency.py` で 0 hits 確認

## 自分側 retrospective

ack 直後 6 round 目の再発という事実を直視して、 [[feedback_tier4_private_state_repeat_6_round]] として memory に正式 pin (= `assert obj._field == ...` を書きそうになった瞬間に public lift を先行させる、 grep self-check を pre-push 4-item check に追加)。 「rule articulate ≠ rule applied」 が固有 trap として認識を固める。

## Test plan re-run

- `pytest tests/test_action_embedding_index_concurrency.py tests/test_action_embedding_index.py`: **30 passed** (= 12 new + 18 existing)
- `ruff check` on changed files: clean
- private-attr grep on new test file: 0 hits

Tier rule self-check (re-applied):
- docstring 形式 `Tier 2: ...` ✓
- Tier 4 violations 0 ✓ (= private state assertion / Mock / format-pinning / snapshot 全て 0)
- invariant 弱化なし ✓
- audit 0 errors (re-run pending CI; 4-item per-file invariant verified locally)

CI 走り直し中、 fast-track 再開よろしくお願いします 🙏

PR #1100, opened 2026-05-30
fix(fp-0008): C5-completeness — bound offloaded inline regardless of field type (list/dict bulk)

tya5:
**[e2e-coder]** — C5-completeness fix for FP-0008 v9-IN blocker.

## The gap

`offload_control_ir_result` in `context_builder.py` only shrank **string** fields when building the offloaded inline. When a result's bulk was in a **list or dict** field (e.g. a `file grep` op with `output_mode=content` returning a huge `matches` LIST), the function:
- Correctly wrote the full result to a scratch ref file
- But built `inline = dict(result)` keeping the list/dict bulk full-size

Result: the inline was NOT bounded → prompt balloon. Observed: a single 3.4M-char grep result, 96% of a 3.5M-char prompt; at 14096 the result hit 987K tokens ≈ Gemini's 1M limit → latent `ContextWindowExceeded` on any smaller-context model.

## The fix — bounded-by-construction

`src/reyn/context_builder.py`:

1. **`_FIELD_KEEP_THRESHOLD`**: per-field size check covers all types (was string-only).
2. **`_preview_field()`**: type-aware preview per oversized field:
   - `str` → head (OFFLOAD_HEAD_CHARS=2048) + marker + tail (OFFLOAD_TAIL_CHARS=512) — existing behavior preserved
   - `list` → first K elements (until ~OFFLOAD_HEAD_CHARS serialised) + `"...(+N more elements omitted; full list at <ref>)"` sentinel
   - `dict` → first N key-value pairs (until ~OFFLOAD_HEAD_CHARS) + `_omitted_keys` note
   - number/bool/None → kept as-is (tiny)
3. **Hard-bound fallback** (key invariant): after per-field previews, if `len(json.dumps(inline))` STILL exceeds `MAX_OFFLOADED_INLINE_BYTES` (many medium-sized fields, deeply nested, etc.), replace the entire inline with `{_offload_preview, _offload_tail, _offload_total_chars, _offload_ref, _offload_note}`. Guarantees bound **unconditionally**.
4. **`MAX_OFFLOADED_INLINE_BYTES = 16_384`**: hard ceiling for any offloaded inline, well above head+tail+small-field overhead, well below any model context limit.

**No information lost**: full result remains at `_offload_ref` (written before any preview logic).
**P7-clean**: no skill-specific strings in OS code.

## Constants

| Constant | Value | Role |
|---|---|---|
| `MAX_CONTROL_IR_RESULT_INLINE_BYTES` | 8,192 | Trigger threshold (unchanged) |
| `OFFLOAD_HEAD_CHARS` | 2,048 | Head preview per field |
| `OFFLOAD_TAIL_CHARS` | 512 | Tail preview per str field |
| `MAX_OFFLOADED_INLINE_BYTES` | **16,384** | Hard ceiling for offloaded inline (new) |

Refs #1093

## Test plan

- [x] **LIST-bulk invariant** (`test_list_bulk_inline_bounded_and_ref_complete`): 20K-item list (≈4.2MB) → `len(json.dumps(offloaded)) <= MAX_OFFLOADED_INLINE_BYTES` AND `json.loads(Path(ref).read_text()) == original` — **THE regression guard**
- [x] **LIST-bulk head preview present** (`test_list_bulk_head_preview_present`): first list element visible in inline
- [x] **dict-bulk variant** (`test_dict_bulk_inline_bounded_and_ref_complete`): dict-field bulk → same bound + ref complete
- [x] **string-bulk regression** (`test_string_bulk_inline_bounded_head_tail_preserved`): existing str behavior preserved (head+tail+ref)
- [x] **many-medium-fields fallback** (`test_many_medium_fields_hard_bound_fallback`): many fields each just under per-field threshold → hard-bound fallback triggers → inline ≤ bound
- [x] **build_frame integration list-bulk** (`test_build_frame_list_bulk_control_ir_result_bounded`): real OSRuntime build_frame with 4.2MB list-bulk result → bound holds via integration path
- [x] ruff check . → All checks passed
- [x] tier audit → 15/15 OK, 0 errors
- [x] pytest -q tests/ → 6391 passed, 1 pre-existing failure (test_legacy_no_media_store_path_returns_inline_content)

### DIFF
--- a/src/reyn/context_builder.py
+++ b/src/reyn/context_builder.py
@@ -33,12 +33,85 @@ MAX_CONTROL_IR_RESULT_INLINE_BYTES: int = 8_192   # ~8KB threshold
 OFFLOAD_HEAD_CHARS: int = 2_048                    # first 2KB kept inline
 OFFLOAD_TAIL_CHARS: int = 512                      # last 0.5KB kept inline
 
+# Hard-bound guarantee (C5-completeness — FP-0008 v9-IN):
+# After per-field type-aware previews, the offloaded inline dict MUST NOT exceed
+# this limit regardless of field types (list/dict/str/nested). The whole-inline-
+# replace fallback enforces this ceiling unconditionally. Value is set well above
+# head+tail+small-fields overhead (~3KB) but well below any model context limit.
+MAX_OFFLOADED_INLINE_BYTES: int = 16_384  # 16KB absolute ceiling for offloaded inline
 
-def _oversized_string_fields(result: dict) -> list[str]:
-    """Return field names whose string value alone exceeds the inline limit."""
+# Per-field keep threshold: fields whose serialised size is below this are kept
+# as-is (they're small enough not to contribute meaningfully to bloat).
+_FIELD_KEEP_THRESHOLD: int = MAX_CONTROL_IR_RESULT_INLINE_BYTES
+
+
+def _preview_field(value: Any, ref_path: str) -> Any:
+    """Return a bounded preview of *value* for the offloaded inline dict.
+
+    Type-aware strategy:
+      - str  → head (OFFLOAD_HEAD_CHARS) + truncation marker + tail (OFFLOAD_TAIL_CHARS)
+      - list → first K elements (until serialized size ~ OFFLOAD_HEAD_CHARS) +
+               a sentinel string describing remaining count + ref
+      - dict → first N key-value pairs (until serialized size ~ OFFLOAD_HEAD_CHARS) +
+               a sentinel string describing omitted keys + ref
+      - number/bool/None → returned as-is (tiny)
+    """
+    if isinstance(value, str):
+        head = value[:OFFLOAD_HEAD_CHARS]
+        has_tail = len(value) > OFFLOAD_HEAD_CHARS + OFFLOAD_TAIL_CHARS
+        tail = value[-OFFLOAD_TAIL_CHARS:] if has_tail else ""
+        marker = (
+            f"\n... [TRUNCATED — {len(value):,} chars total; "
+            f"full content at {ref_path}] ..."
+        )
+        return head + marker + (f"\n[TAIL PREVIEW]\n{tail}" if tail else "")
+
+    if isinstance(value, list):
+        preview_items: list = []
+        accumulated = 0
+        for item in value:
+            item_size = len(json.dumps(item, ensure_ascii=False))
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ACCEPT ✅ (C5-completeness, v9-IN)

## Review verdict
Closes the C5 size-axis gap (single large list/dict bulk staying inline) **bounded-by-construction**. Reviewed against my sign-off gate:

- **Type-independent hard bound**: `_oversized_fields` now uses `len(json.dumps(v)) > threshold` for ALL types (not just `isinstance(v, str)`); `_preview_field` is type-aware (str→head+tail, list→first-K+count, dict→first-N+omitted). ✓
- **Hard-bound fallback guarantees the ceiling**: after per-field previews, if inline still `> MAX_OFFLOADED_INLINE_BYTES (16KB)`, the whole inline is replaced with a fixed compact dict (`serialized[:2KB]` + `[-0.5KB:]` + scalars + `_offload_ref`) ≈ 2.5KB. So post-offload inline ≤ 16KB **regardless of field type, count, or nesting** — even a single 3MB list element is caught by the fallback. Not gemini-1M-dependent → safe under 200K-context models. ✓
- **No info loss**: full serialized result written to ref file; ref read-back `== original` asserted. ✓

## Tier rule check ✓
1. **Docstring Tier**: all test functions declare `Tier 2` (OS invariant) — correct classification for OS-level offload invariants.
2. **No Tier 4**: every test pins a load-bearing invariant (post-offload inline bounded — provider-agnostic CWE prevention), not incidental behavior.
3. **Invariant essentiality**: `inline_size <= MAX_OFFLOADED_INLINE_BYTES` + `stored == result` are genuine load-bearing guarantees, not format-pinning. Head-preview asserts use `in` / `startswith` (not exact format). ✓
4. **Real-path exercise (C6 lesson)**: `test_build_frame_list_bulk_control_ir_result_bounded` drives the real `build_frame`/OSRuntime path — not a function-direct-only test that bypasses production wiring (the exact gap that let #1098 through). Plus `offload_control_ir_result` is called as in production (no sandbox layer bypassed). ✓

## dogfood axis (complementary, live)
Static review confirms the bound. The **live** confirmation comes from sandbox_2's post-revert re-verify: 13236/14096 (prev 3.5–3.9M inline) should now show bounded `max_sent_prompt_chars`. That measurement closes the loop on whether the balloon is structurally bounded end-to-end.

Auto-merge (squash) enabled — lands on pytest green.

Issue #1133, opened 2026-05-31
[FP-0008] sandboxed_exec op not advertised in available_control_ops — phases with allowed_ops:[sandboxed_exec] give LLM no argv schema (structural)


tya5:
**[lead-coder]** — STRUCTURAL bug found via the FP-0008 #234 live C7 run (flask-5014). The `sandboxed_exec` Control IR op is **not advertised** in `available_control_ops`, so a phase declaring `allowed_ops: [sandboxed_exec]` gives the LLM no schema for it — the model cannot know the field is `argv` (required) and guesses `command` (shell mental model) → `ActOutput` validation fails (`ops.0.sandboxed_exec.argv Field required`).

## Root cause (primary evidence)
- `src/reyn/kernel/control_ir_executor.py` `available_ops()` builds the `available_control_ops` specs from a hardcoded list. It emits `ControlIROpSpec`s for **file, ask_user, shell, mcp, lint, run_skill, web_fetch, web_search** — but **NOT `sandboxed_exec`** (`grep -c 'kind="sandboxed_exec"'` in `available_ops()` = **0**).
- A phase's `allowed_ops` *filters* `available_control_ops` down to the declared set. Since no `sandboxed_exec` spec exists to begin with, filtering `[file, sandboxed_exec]` yields only the `file` entry. `sandboxed_exec` reaches the LLM (if at all) as a bare kind with no `description`/`example`/field schema.
- **Live proof**: in the flask-5014 verify request (raw LLM trace, 23,917 bytes), the string `argv` appears **0 times** in the entire context the model received; `available_control_ops` contained only the `file` entry. The model emitted `{"kind":"sandboxed_exec","command":"git apply ...","capture_stdout":true}` (no `argv`) → `phase_failed | verify | final_error: ops.0.sandboxed_exec.argv Field required [missing]` (×3 retries).
- The successful in-container `sandboxed_exec_completed backend=docker rc=0` ops in the same run were **OS-preprocessor-built** (C6 revert: `parse_test_targets` produces argv lists), **not model-emitted** — so they bypass the advertisement gap. The model never successfully emitted a `sandboxed_exec` op.

## Why this surfaced now
[...]

PR #1126, opened 2026-05-31
feat(fp-0008): #1115 Stage 2 — swe_bench verify/report shell→sandboxed_exec + phase default_sandbox_policy wiring

tya5:
**[e2e-coder]** — Refs #1115. A-ii, stacked on the merged cwd-anchor PR #1124.

## Why
Migrate the swe_bench **verify** + **report** phases off the deprecated `shell` op to `sandboxed_exec`, so their repo exec (git checkout revert / git apply / pytest / git diff) routes through the run's `EnvironmentBackend` (host today, container for C7) instead of a host-only subprocess.

**Behavior-preserving on host:** with #1124, `sandboxed_exec` runs with `cwd=workspace.base_dir` exactly like the old `shell` op; a `NoopBackend` is byte-for-byte equivalent to the old no-sandbox subprocess.

## Loader gap this surfaced (and fixes)
#1123 added `Phase.default_sandbox_policy` + the runtime D mechanism, but the `.md` frontmatter→Phase path was **never wired** — a `default_sandbox_policy` declared in a skill.md was silently dropped by the parser/expander, so the D mechanism was unreachable from any real skill. Wired through `parser.py` (read + mapping validation) → `ir.py` (`PhaseDef` field) → `expander.py` (`phase_kwargs`), mirroring `allowed_ops`.

## Changes
- **parse_test_targets.py**: returns argv lists `["git","checkout","HEAD","--",p]` instead of shell command strings (`sandboxed_exec.argv` is `list[str]`, executed directly — no shell). Docstrings + `output_schema` updated.
- **verify.md / report.md**: `allowed_ops` shell→sandboxed_exec; declare a permissive `default_sandbox_policy` (this phase runs an arbitrary repo's git+test suite, so broad fs/subprocess/network is inherent; **ignored** by a container backend, best-effort on host); iterate run_op `kind: shell`/`cmd` → `sandboxed_exec`/`argv` with `args_from {argv: _iter.item}`; body P8 cleanup (no op-kind naming; dropped `2>&1` since `sandboxed_exec` captures stdout/stderr separately and argv is not a shell).
- **Gold-flow preserved** (untouched in phase bodies): verify outcome semantics (apply-fail→plan / rc==0→report / else→plan), C6 test_patch revert, PR-N15 `_skill_input` determinism, convergence guard.
[...]

PR #1134, opened 2026-05-31
fix(fp-0008): #1133 — advertise sandboxed_exec in available_ops (model-driven exec was structurally broken)

tya5:
**[e2e-coder]** — Refs #1133, #1115. Forward-fix for a latent regression my #1126/#1127 migration exposed; lands on its own merit (orthogonal to the new decide-turn observation below).

## Root cause (primary evidence)
`available_ops()` (control_ir_executor) never emitted a `sandboxed_exec` `ControlIROpSpec`. So the op was never advertised to the LLM — `argv` appeared **0×** in the request context (lead-coder confirmed via `dogfood_trace --mode llm-context`). After #1126/#1127 moved swe_bench phases onto `allowed_ops: [sandboxed_exec]`, the per-phase available-ops filter (`op.kind in allowed_ops`) dropped the op to nothing. The model, never shown `argv`, guessed a shell `command` string → `ActOutput` rejected it (`argv` required). The live flask-5014 verify failed this way 3×; the only valid `sandboxed_exec` ops in that run were **preprocessor-built** (`args_from`), never model-emitted.

Same class as the loader-gap pattern: handler + runtime were wired; the **LLM-advertisement path** was not.

## Fix
Add the `sandboxed_exec` `ControlIROpSpec` to `available_ops()`:
- description steers the model to **argv as a LIST**, explicitly warns it is **not** a shell `command`/`cmd` string and there is **no shell** interpretation (the exact observed failure mode); covers all `SandboxedExecIROp` policy fields; example uses a list `argv`.
- **unconditional** (not gated like `shell`): `sandboxed_exec` is default-permissible and works under every backend; per-phase `allowed_ops` is the gate, like `file`/`run_skill`. **Additive** — phases not allowing it are unaffected (catalog is filtered per `allowed_ops`).

## Gate-6 validation (both sides + provenance — applied the co-sign-provenance lesson)
Re-ran flask-5014 in-container **with this fix**:
- **REQUEST side**: the verify request context now contains the `sandboxed_exec` spec with "argv: the command as a LIST" (deterministic build confirms `available_control_ops=[file, sandboxed_exec]` with argv in the description).
- **RESPONSE side (provenance = LLM-authored responses, not preprocessor)**: the model emitted `sandboxed_exec` with `argv` **×5** and `command`/`cmd` **×0** — e.g. setup `argv:["git","checkout","7ee9ceb…"]`, verify `argv:["python","-m","pytest","tests/test_blueprints.py"]`. Before the fix: argv 0 / all `command`. → model-driven exec restored.

## Scope note (separate, NOT this PR)
The re-run's verify still didn't reach a test PASS, but the **failure mode changed** (`argv Field required` → `decide-turn output missing required 'control' block`) — a *different* issue. Request-side check shows the control contract **is** advertised and the model produced valid control blocks in 5/5 other-phase decides, so this is not the same absent-advertisement bug; it leans weak-model inconsistency but the specific failing decide output isn't fully isolated yet. Being articulated as a separate issue with that evidence — not bundled here.

## Test plan
- [x] `pytest tests/test_sandboxed_exec_advertised_1133.py tests/test_control_ir_executor.py tests/test_op_sandboxed_exec.py -q` → 33 passed (spec present; advertised regardless of shell_allowed; description requires argv-list-not-command; example is a valid `SandboxedExecIROp`; allowed_ops:[sandboxed_exec] surfaces it through the filter)
- [x] `ruff` + `scripts/test_tier_audit.py --strict` → clean / 0 violations
- [x] P7: zero skill-specific strings in the spec
- [x] Full suite `pytest -q --ignore=.claude` → 6541 passed, 1 pre-existing failure (`test_web_fetch_preview_path_ref`, local extractor-lib diff, not in CI)
- [x] **Live re-run** (flask-5014, in-container) — model now emits valid `argv` (gate-6, evidence above)

### DIFF
--- a/src/reyn/kernel/control_ir_executor.py
+++ b/src/reyn/kernel/control_ir_executor.py
@@ -228,6 +228,28 @@ class ControlIRExecutor:
                 ),
                 example={"kind": "shell", "cmd": "reyn run my_skill 'hello'", "timeout": 120},
             )] if self._shell_allowed else []),
+            ControlIROpSpec(
+                kind="sandboxed_exec",
+                description=(
+                    "Execute a command under a sandbox policy and return stdout, stderr, "
+                    "and returncode. "
+                    "argv: the command as a LIST of strings — argv[0] is the executable, "
+                    "e.g. [\"python\", \"-m\", \"pytest\", \"-x\"]. It is NOT a single shell "
+                    "string and NOT a 'command'/'cmd' field — always provide argv as a list. "
+                    "There is no shell: pipes, redirects (>, 2>&1), globs, and &&/|| are NOT "
+                    "interpreted — pass real argv tokens only. "
+                    "Runs in the project root directory; stdout and stderr are captured "
+                    "separately (do not append redirections). "
+                    "Optional policy fields (network: bool, read_paths/write_paths: list[str], "
+                    "allow_subprocess: bool, env_passthrough: list[str], timeout_seconds: int) "
+                    "may be omitted — when the phase declares a default sandbox policy the OS "
+                    "applies it and overrides them. "
+                    "Use for git, pytest, build, and other repository commands. "
+                    "Status: enabled by default — issue sandboxed_exec ops directly; do not "
+                    "abort on permission concerns."
+                ),
+                example={"kind": "sandboxed_exec", "argv": ["python", "-m", "pytest", "-x", "--tb=short"]},
+            ),
             *([ControlIROpSpec(
                 kind="mcp",
                 description=(
--- a/tests/test_sandboxed_exec_advertised_1133.py
+++ b/tests/test_sandboxed_exec_advertised_1133.py
@@ -1,0 +1,98 @@
+"""Tier 2: FP-0008 #1133 — sandboxed_exec is advertised in available_ops().
+
+Root cause of the C7 verify-fail (live flask-5014): `available_ops()` never
+emitted a `sandboxed_exec` ControlIROpSpec, so a phase that declared
+`allowed_ops: [sandboxed_exec]` (after the #1126/#1127 migration) had the op
+filtered to nothing — the LLM never saw `argv`, guessed a shell `command`
+string, and failed ActOutput validation. Same class as
+test ... wire-full-frontmatter-to-runtime path: handler/runtime wired, the
+advertisement path was not.
+
+These pin: the spec exists, its description makes `argv` a required LIST (not a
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read; merge on CI green). Clean root-cause fix for the #1133 structural advertisement gap.

## Fix
`available_ops()` now emits a `sandboxed_exec` `ControlIROpSpec`, **unconditional** (per-phase `allowed_ops` is the gate, like `file`/`run_skill`; correct — `sandboxed_exec` is default-permissible, unlike `shell`'s `_shell_allowed` gate). The description is crafted to counter the **exact observed failure modes**:
- `argv` is a **LIST** (argv[0]=executable), **NOT** a single shell string / `command`/`cmd` field — directly fixes the model's `{"command":"git …"}` emit.
- **No shell**: pipes/redirects (`>`,`2>&1`)/globs/`&&`/`||` not interpreted — also pre-empts the `2>&1`-as-literal-arg trap from the #1126 migration.
- stdout/stderr captured separately; all `SandboxedExecIROp` policy fields covered + `default_sandbox_policy` override noted. Example uses a list `argv`.

## Tests (Tier 2, real `ControlIRExecutor` + real `SandboxedExecIROp`, no mocks) ✓
- spec present; advertised regardless of `shell_allowed`; description pins argv-list-not-command-not-shell; **example validates as a real `SandboxedExecIROp`** (guards a bad advertised example); **`allowed_ops:[file,sandboxed_exec]` surfaces it through the filter** — the exact #1133 regression signature.
- Tier audit CI green; P7 (zero skill strings in spec) confirmed.

## Gate-6 (both sides + provenance — co-sign-provenance lesson applied) ✓
Live flask-5014 re-run with the fix: REQUEST side now advertises `sandboxed_exec` w/ argv; RESPONSE side = model-authored `argv` ×5 / `command` ×0 (was argv 0 / all command pre-fix). Model-driven exec restored.

## On the separate decide-turn observation
Endorsed handling. You applied the discipline fully — request-side shows the control contract **is** advertised + 5/5 valid control blocks elsewhere, AND you correctly identified the **trace-infra gap** (the failing pre-validation decide output isn't captured by `REYN_LLM_TRACE_DUMP`/events) → classification is **not possible yet** (pre-conclusion Q4), so it stays unclassified. That is exactly right. The separate issue should bundle the trace-infra capture gap as the real blocker (see broker).

Issue #1132, opened 2026-05-31
[FP-0008] reyn run --state-dir not wired to events log (only artifacts/offload); agent.py events use hardcoded .reyn


tya5:
**[lead-coder]** — `reyn run --state-dir` is only partially wired: it controls Workspace artifacts/offload but NOT the events log. Found during the FP-0008 #234 live C7 run (flask-5014).

## Symptom
`reyn run --env-backend=docker --state-dir <host>` is documented (and the `--state-dir` help text in `run.py` from #1130 explicitly says) to hold **"events, offload, artifact handles"**. But:
- `agent.py:163` builds the events path from the hardcoded `self.state_dir = ".reyn"` (agent.py:71), a **host-cwd-relative** path — NOT `self._workspace_state_dir` (the `--state-dir` value).
- `agent.py:209` passes `workspace_state_dir` to the OSRuntime/Workspace, so only artifacts/offload honor `--state-dir`.

Result (observed live): in the flask-5014 run, offload landed at `--state-dir`, but events landed at `<reyn-process-cwd>/.reyn/events/...` — split, both host-side.

## Severity
- **Not a correctness bug for the container case**: the events writer is a plain host-side file write (it does NOT route through the `EnvironmentBackend`), so events always land host-side relative to the reyn process cwd — **no in-container (`/testbed/.reyn`) leak**, and events survive `docker rm -f`.
- **But**: (a) the `--state-dir` contract (and #1130 help text) is misleading — it claims events, which it doesn't control; (b) OS state is not co-located at one host path, complicating inspection/collection for containerized runs.

## Fix
`agent.py` events `skill_dir` should use `self._workspace_state_dir` when set, falling back to `self.state_dir` (`.reyn`) — so `--state-dir` becomes the single source for all host-side OS state, as documented. Small, additive (host default unchanged).

## Provenance / ownership
Pre-existing from #1130 (`--state-dir` flag), which I (lead-coder) reviewed + merged without auditing that events actually consume `workspace_state_dir` — a consumer-audit miss on the value's documented contract. Not introduced by #1131 (β2b), which correctly passes `--state-dir`. Non-blocking; tracked for a follow-up fix.

tya5:
**[lead-coder]** — Resolved by #1138 (merged 0f9fe36f). `Agent.state_dir` now honors `workspace_state_dir` so the events log co-locates with artifacts/offload under `--state-dir`; default `.reyn` preserved. Consumer-audit verified (events path is the sole consumer). Closing.

PR #1130, opened 2026-05-31
feat(fp-0008): #1115 Stage 2 — `reyn run` container EnvironmentBackend flags (β2a, dual-seam injection)

tya5:
**[e2e-coder]** — Refs #1115. PR-β2a of the #234 wave (lead-coder ruled fork → option (a): generic `reyn run` flags). Foundational, skill-agnostic half of the C7 in-container wiring; PR-β2b adds the per-instance container lifecycle that drives it.

## What
Generic container-backend flags on `reyn run` so the repo FS + command exec route through a running container while the OS stays host-side (#1115 option ②):
- `--env-backend {host,docker}` (default `host` = unchanged identity behavior)
- `--container NAME` — running container for docker
- `--repo-dir PATH` — in-container repo working tree (relative paths + cwd resolve here)
- `--state-dir PATH` — host-side OS state/artifacts dir (kept off the routed FS)

`_build_environment_backend` constructs **one** `DockerEnvironmentBackend`; run.py injects that **same instance** at both seams — `Agent.environment_backend` (FS) and `Agent.sandbox_backend` (exec) — agent-level uniform (案C-pure), landing on the seam #1121 already created. `workspace_base_dir` = `--repo-dir`; `workspace_state_dir` = `--state-dir`.

**Cross-process correctness**: a live backend *instance* can't cross a subprocess boundary, so config (flags) crosses and the instance is reconstructed inside `run.py` — consistent with the existing `sandbox_config`→`get_default_backend` pattern.

## Review gate mapping (lead-coder)
- ✅ **gate 1 (flags GENERIC, P7)**: zero swe_bench/swebench/testbed strings in run.py (the in-container example uses a neutral `/repo`). Flags are generic container-backend params.
- ✅ **gate 2 (dual-seam, same instance)**: one `DockerEnvironmentBackend` injected at both `environment_backend` + `sandbox_backend`; pinned via a test asserting the single instance satisfies BOTH the `EnvironmentBackend` and `SandboxBackend` Protocols.
- (gates 3–6 — bridge-free in-container patch, runner lifecycle, exec/eval split, supervised live run — are PR-β2b.)

## Test plan
- [x] `pytest tests/test_run_container_backend_flags_1115.py -q` → 7 passed (host triple; docker single-backend+dirs; dual-Protocol; no-state-dir; missing-container/-repo-dir exits)
- [x] `reyn run --help` shows all 4 flags wired
- [x] P7: `grep -c "swe_bench|swebench|testbed" run.py` → 0
[...]

PR #1138, opened 2026-05-31
fix(fp-0008): #1132 — events audit log honors workspace_state_dir

tya5:
**[e2e-coder]** — Refs #1132, #1135, #1115. The MINOR finding from the #1135 C7 in-container run; now unblocked (lead-coder queued it after #1135b/#1137 landed, same events area = serial).

## Why
The events log lives under `Agent.state_dir` (`Path(state_dir)/events/...`), hardcoded to cwd-relative `.reyn`. A run passing an explicit host-side `--state-dir` (the in-container case: base_dir = container repo, state kept host-side) had its Workspace artifacts + `control_ir_offload` under `--state-dir` but its **events split off under `.reyn`** — not co-located. Both host-side (P6 held), but the dirs diverged.

## Fix
Derive `Agent.state_dir` from `workspace_state_dir` when provided (events co-locate with the run's other state, the same host dir the Workspace uses); fall back to `.reyn` when unset (unchanged default host behavior). `Agent.state_dir`'s only consumer is the events dir (grep-verified) — additive + back-compat.

## Closes the #1135 read-side caveat
The `phase_output_validation_failed.raw_output_ref` is resolved against the offload root under `state_dir`. With this fix events + offload + artifacts share one root, so the read-side resolution (`base_dir=state_dir`) finds the offloaded raw co-located with the events that reference it.

## Test plan
- [x] `pytest tests/test_events_state_dir_1132.py -q` → 3 passed (state_dir == workspace_state_dir + derived events root under it; defaults to `.reyn` unset; events & Workspace state co-locate)
- [x] agent/events regression sweep (`-k "agent or events_state or events_path"`) → 320 passed (no regression)
- [x] Full suite `pytest -q --ignore=.claude` → 6552 passed, 1 pre-existing failure (`test_web_fetch_preview_path_ref`, local extractor-lib diff, not in CI)

### DIFF
--- a/src/reyn/agent.py
+++ b/src/reyn/agent.py
@@ -68,7 +68,14 @@ class Agent:
         sandbox_backend: "SandboxBackend | None" = None,
     ) -> None:
         self.model = model
-        self.state_dir = ".reyn"
+        # FP-0008 #1132: the events audit log lives under state_dir. Honor an
+        # explicit workspace_state_dir (the same host-side dir that holds the
+        # Workspace's artifacts + control_ir_offload) so events co-locate with
+        # the rest of the run's state — required for an in-container run where
+        # base_dir is the container repo and state is kept host-side. Falls back
+        # to the cwd-relative ".reyn" default (unchanged host behavior) when no
+        # explicit state dir is given.
+        self.state_dir = str(workspace_state_dir) if workspace_state_dir is not None else ".reyn"
         self.strict = strict
         self._subscribers = list(subscribers or [])
         self._intervention_bus = intervention_bus
--- a/tests/test_events_state_dir_1132.py
+++ b/tests/test_events_state_dir_1132.py
@@ -1,0 +1,52 @@
+"""Tier 2: FP-0008 #1132 — the events audit log honors workspace_state_dir.
+
+The events log lives under `Agent.state_dir` (`Path(state_dir)/events/...`). That
+was hardcoded to the cwd-relative `.reyn`, so a run that passed an explicit
+host-side `--state-dir` (e.g. an in-container run, where base_dir is the container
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read + independent consumer-audit; merge on CI green). Fixes the #1132 gap that was my own #1130 review miss — so I applied the consumer-audit discipline that miss taught.

## Fix
`Agent.state_dir = str(workspace_state_dir) if workspace_state_dir is not None else ".reyn"` — the events log (`Path(self.state_dir)/events/...`) now honors `--state-dir`, co-located with the Workspace's artifacts/offload; unchanged `.reyn` default when unset (back-compat).

## Consumer-audit (independently verified — the discipline my #1130 miss taught)
Grepped `Agent.state_dir`'s consumers on origin/main myself: the **only** reader of `self.state_dir` is the events path at `agent.py:163`. The other `.state_dir` hits in src are different objects (`SkillRegistry.state_dir`, `event.data.state_dir`, `ctx.state_dir_strategy`) — not `Agent.state_dir`. So this change affects only the events path = additive, no other consumer impacted. e2e's "only consumer = events" claim confirmed.

## Closes the #1135 read-side caveat
With events + `control_ir_offload` + artifacts now sharing one root, the read-side resolving `phase_output_validation_failed.raw_output_ref` against `base_dir=state_dir` finds the offloaded raw co-located with the events that reference it (no more events≠offload-root split). Good — that closes the caveat I flagged in the #1135 contract.

## Tests ✓
Tier 2, real `Agent`, no mocks: explicit workspace_state_dir → state_dir + events root under it; default `.reyn` when unset; events & Workspace state co-locate. 320 agent/events regression pass. Tier audit green.

PR #1140, opened 2026-05-31
fix(tui): remove SlashPicker 'find' hardcode via SlashCommand.tab_footer_fn (P7)

tya5:
**[tui-coder]** — Refs #1070 (lead-coder handoff, P7 + dataclass contract, user 「未解決まで完遂」mandate).

## Problem (P7 violation)

`SlashPicker` — generic OS chat infra — hardcoded `if cmd.name == "find":` (slash_picker.py) to render the Tab-recall footer. A generic widget knowing a skill-specific command name is a P7 violation: every future command wanting a picker footer would add another `name ==` branch.

## Fix

Commands supply their own picker-hint footer via a new field:

```python
SlashCommand.tab_footer_fn: TabFooterFn | None = None   # () -> str | None
```

The **command** owns the message + its visibility condition; the **picker** stays generic and renders whatever the field returns with the same `↳` chrome the usage row uses. The picker no longer knows any command name.

### Why a new field (not reuse `usage` / `see_also`)

Per the existing-mechanism-grep discipline noted in #1070, I checked the existing footer fields first:
- `usage: str` — static, always-rendered, separate render path
- `see_also: tuple` — static doc paths for `/help` focus mode

The find footer is a **conditional, runtime-computed** line (shown only when find history is non-empty). Neither static field can express "compute a footer, maybe None" → a new `Callable[[], str | None]` field is justified. (Design recorded in [#1070 comment](https://github.com/tya5/reyn/issues/1070#issuecomment-4585592921).)

### Bonus: lazy import removed

The previous picker code did a lazy `from reyn.chat.slash.find import find_history_has_entries` (a circular-dependency workaround). The callable now closes over find.py's own function, so the picker no longer imports the slash submodule at all — the circular-dep is gone structurally.

## Scope

- `slash/__init__.py` — `tab_footer_fn` field + threaded through the `slash()` decorator. **Consumer-audited**: `SlashCommand` is constructed only via the decorator; all other call sites (tests) are keyword-only with the field defaulting to `None` → every existing command unchanged.
- `slash/find.py` — `_find_tab_footer()` returns the message iff history non-empty, else None; wired via `@slash(..., tab_footer_fn=...)`.
- `tui/widgets/slash_picker.py` — generic `cmd.tab_footer_fn()` render; `find` literal + lazy import removed.
- `tests/cli/test_find_picker_tab_recall_hint.py` — kept 3 /find e2e tests (value-preserving) + added 2 generic-mechanism tests.

Out of scope (per #1070): `header.py` badge hardcode (separate badge concern).

## Review gates

- [x] ① `find` literal removed from picker — `grep` zero command-name literals, P7-clean
- [x] ② `tab_footer_fn` additive — default `None` → unset commands behave exactly as before (152 slash/picker/find tests green; consumer audit confirms all constructions keyword-only)
- [x] ③ Tier-2 no-mock test — real `SlashCommand` instances; footer renders for a command supplying `tab_footer_fn`, omitted when unset / when it returns `None`
- [x] ④ no existing slash-command regression — `pytest -k "slash or picker or find"` → 152 passed
- [x] P7 articulated: generic widget no longer knows a specific command name

## Test plan

- [x] `pytest tests/cli/test_find_picker_tab_recall_hint.py` — 5 passed (3 /find value-preserving + 2 generic)
- [x] `pytest tests/cli -k "slash or picker or find"` — 152 passed (regression gate)
- [x] `pytest tests/cli/test_tui_smoke.py` — 40 passed
- [x] `grep` 'find'/command-name literal in slash_picker.py → 0; find import in picker → 0
- [x] `ruff check` clean (I001 caught pre-push)
- [x] `scripts/test_tier_audit.py --strict` clean (0 warn / 0 err)
- [x] (skipped — visual) live `reyn chat` `/find ` picker footer with/without history — value-preserving by construction; behavior covered by the 5 Tier-2 tests

## Tier self-check

Only Tier-2 behavior tests (real instances, public `.rendered_text()` surface). No format-pin / no private-state assert / no golden-file. ✓

### DIFF
--- a/src/reyn/chat/slash/__init__.py
+++ b/src/reyn/chat/slash/__init__.py
@@ -29,6 +29,13 @@ HandlerFn = Callable[..., Awaitable[None]]
 # ``"discard ab"``). Completers that don't need it (e.g. ``/attach``
 # which always lists agent names) can ignore the arg via a default.
 CompleterFn = Callable[..., list[str]]
+# TabFooterFn signature: ``() -> str | None``. Supplies the optional
+# picker-hint footer line (the dim ``↳ <message>`` sub-row shown once the
+# user types ``/<cmd> ``). Returning ``None`` (or "") means "show nothing
+# right now" — the command owns both the message text and its visibility
+# condition (e.g. /find only surfaces its Tab-recall affordance when there
+# is history to recall). The picker owns the ``↳`` chrome + styling.
+TabFooterFn = Callable[[], "str | None"]
 
 
 @dataclass
@@ -58,6 +65,16 @@ class SlashCommand:
     # ``"docs/concepts/plan-mode.md"``). Defaults to empty tuple so all
     # existing commands without explicit see_also are unaffected.
     see_also: tuple[str, ...] = ()
+    # Optional picker-hint footer supplier. When set, the SlashPicker hint
+    # mode renders a dim ``  ↳ <message>`` sub-row below the summary/usage,
+    # where ``<message>`` is whatever this callable returns. The callable
+    # returns ``None`` (or "") to render nothing — letting the command gate
+    # the footer on runtime state (e.g. /find shows its Tab-recall hint only
+    # when history is non-empty). Defaults to ``None`` so every command
+    # without an explicit footer behaves exactly as before. Keeping the
+    # message + its visibility inside the command (not the widget) is what
+    # makes the picker generic: it never hardcodes a command name.
+    tab_footer_fn: TabFooterFn | None = None
 
 
 class SlashRegistry:
@@ -133,6 +150,7 @@ def slash(
     hidden: bool = False,
     usage: str = "",
     see_also: tuple[str, ...] = (),
+    tab_footer_fn: TabFooterFn | None = None,
 ) -> Callable[[HandlerFn], HandlerFn]:
     """Decorator that registers `fn` as a slash command on import.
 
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read + independent P7 grep; merge on CI green). Clean #1070 P7 fix.

## P7 (independently verified)
Grepped `slash_picker.py` on the PR branch: **zero `'find'` literal** remaining. The `if cmd.name == "find":` hardcode (OS widget knowing a command name) is gone, replaced by `if cmd.tab_footer_fn is not None: footer = cmd.tab_footer_fn()`. The picker is now generic — it renders whatever the command supplies with the same `↳` chrome, never naming a command. ✅

## Design — callable beats a static field (good call)
`tab_footer_fn: () -> str | None` (not a static `tab_footer` string): correct, because /find's footer is **conditionally dynamic** (shown only when history is non-empty). A static field couldn't express the visibility condition; the callable lets the command own both the message AND its gate. `/find`'s `_find_tab_footer()` returns the message only when `find_history_has_entries()`. Rationale recorded in #1070 (issue=contract). Bonus: the circular-dep lazy-import workaround is structurally removed.

## Back-compat / consumer-audit ✅
`tab_footer_fn` defaults to `None` → every existing command renders exactly as before. `SlashCommand` constructed via the one `slash()` decorator + keyword-only callers (consumer-audit done). additive.

## Tests ✅
Tier-2 no-mock ×5 (3 /find value-preserving e2e + 2 generic footer), slash regression 0 (152 pass), smoke/ruff/tier clean.

PR #1141, opened 2026-05-31
feat(fp-0008): #1135 sibling — capture raw LLM output on pre-parse JSON-decode failure

tya5:
**[e2e-coder]** — Refs #1135, #1115. The sibling pre-parse capture (user "complete until resolved" mandate). Built to the **canonical contract locked in #1135** (opt A; lead-coder decision + sandbox_2 read-side confirm, both in-issue per the contract-in-issue discipline).

## Why
#1137 captures **post-parse** phase-output validation failures. The **pre-parse** layer — `call_llm`'s JSON-repair-fail at `llm.py:926` — left the model's raw output in **no events-layer record** (only the first 800 chars in the exception + the opt-in `REYN_LLM_TRACE_DUMP`). The C7 run3 apply-phase malformed-escape failure was thus undiagnosable from events.

## What (opt A — inline-cap at the llm layer)
Additive event `llm_output_json_decode_failed` at the JSON-decode-fail boundary:
```
data: {failure_kind: "json_decode", error: str, raw_output: str}
```
- emitted via the in-scope `event_log`; **inline-cap only** (the llm layer has no `state_dir` to offload to — and a decode malformation is diagnosable from a window around the error position).
- `_truncate_json_for_event`: raw ≤8192B verbatim; else a cap-sized window centered on `JSONDecodeError.pos` (or head when pos unknown) with `…[N bytes before/after]` markers.
- **additive + behavior-preserving**: the `ValueError` still raises; existing events unchanged.
- read-side = `dogfood_trace --mode llm-emitted-ops` (sandbox_2); inline-only confirmed sufficient for the decode case (no `raw_output_ref`).

## Review gate (#1135)
- ✅ capture boundary = llm layer (before kernel), event_log in scope
- ✅ inline-cap (8192 named const); no offload at the llm layer (no state_dir) — opt A as decided
- ✅ existing behavior unchanged (ValueError still raised; no existing event touched)
- ✅ Tier 2 no-mock; real pre-parse-fail capture test

## Test plan
- [x] `pytest tests/test_llm_json_decode_capture_1135.py -q` → 5 passed (truncation: verbatim ≤cap / pos-centered window w/ byte-accurate markers / head-on-unknown-pos / cap=8192; behavioral: real `litellm.acompletion` stub → bad JSON → event fires w/ failure_kind+error+raw_output, ValueError still raised)
- [x] llm regression sweep (`-k "llm_trace or llm_call or json_parse or llm_json"`) → 70 passed
- [x] Full suite `pytest -q --ignore=.claude` → 6557 passed, 1 pre-existing failure (`test_web_fetch_preview_path_ref`, local extractor-lib diff, not in CI)

Completes the trace-infra: post-parse (#1137) + pre-parse (this) malformed-output capture, both read by `dogfood_trace --mode llm-emitted-ops`.

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -14,6 +14,31 @@ import httpx
 logger = logging.getLogger(__name__)
 from reyn.llm.json_parse import loads_lenient
 from reyn.llm.model_resolver import ModelSpec
+
+# FP-0008 #1135 sibling: inline cap for the raw output captured on a pre-parse
+# JSON-decode failure (opt A — llm layer, no offload).
+_JSON_DECODE_RAW_CAP = 8192
+
+
+def _truncate_json_for_event(raw: str, pos: "int | None", cap: int = _JSON_DECODE_RAW_CAP) -> str:
+    """Bound the raw LLM output for the json-decode-failure event (#1135 sibling).
+
+    Returns *raw* unchanged when ≤ ``cap``. Otherwise returns a ``cap``-byte
+    window centered on the JSONDecodeError position (where the malformation is),
+    or the head when ``pos`` is unknown, with ``…[N bytes before/after]`` markers.
+    Inline-only: the malformation in a decode failure is diagnosable from the
+    window, so no offload is needed at the llm layer.
+    """
+    if len(raw) <= cap:
+        return raw
+    if pos is None:
+        return raw[:cap] + f"\n…[truncated {len(raw) - cap} bytes]"
+    half = cap // 2
+    start = max(0, min(pos - half, len(raw) - cap))
+    end = start + cap
+    head = f"…[{start} bytes before]\n" if start > 0 else ""
+    tail = f"\n…[{len(raw) - end} bytes after]" if end < len(raw) else ""
+    return head + raw[start:end] + tail
 from reyn.llm.pricing import TokenUsage
 from reyn.schemas.models import ContextFrame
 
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read; merge on CI green). Completes the trace-infra (pre-parse half), canonical #1135 opt-A.

## Implementation (opt-A, contract-conformant)
- `_truncate_json_for_event`: ≤cap → verbatim; oversized → a `cap`-byte window **centered on the `JSONDecodeError.pos`** (the malformation site) with `…[N bytes before/after]` markers, or head when pos is unknown. Inline-only — no offload (llm layer has no state_dir; a decode malformation is diagnosable from the window). 8192 named const.
- At the JSON-decode-fail raise site (`llm.py:947`): emits additive `llm_output_json_decode_failed{failure_kind:"json_decode", error, raw_output}`; the existing `raise ValueError` is preserved → behavior unchanged.

## Tests ✓ (Tier 2)
- Truncation units: within-cap verbatim / pos-centered window with marker byte-accounting (`before + cap + after == len(raw)`) / head-when-pos-unknown / cap const.
- **Behavioral**: drives the real `call_llm` JSON-decode path with a **real async-callable stub** (`_ScriptedBadJSON`, not MagicMock — testing-policy-compliant) returning an invalid `\q`-escape body → asserts the event fires with failure_kind/error/raw_output AND the run still raises. Real path, real failure mode.

## Conformance / cross-PR
- inline-only (no `raw_output_ref`) matches sandbox_2's read-side (`llm-emitted-ops`, inline confirmed). Together with #1137 (post-parse) this makes **every malformed-output failure — pre-parse JSON-decode AND post-parse validation — captured in the always-on events log**. trace-infra complete.
- P7-clean (`failure_kind` is a generic OS failure category, no skill strings). additive / back-compat.

## Nit (non-blocking)
- `_truncate_json_for_event` is defined *between* two module imports (`...model_resolver` and `...pricing`), so a def interrupts the import block. ruff-clean, but moving it below all imports would be tidier — optional, don't block.

Merging once CI lands green.

PR #1143, opened 2026-05-31
fix(test-audit): word-boundary FORMAT_PIN_RE so cell_len() isn't a false format-pin

tya5:
**[tui-coder]** — Refs #1082 (lead-coder-filed, LOW-pri; picked up now that the palette wave + #1070 P7 fix are landed).

## Problem

The Tier-4 format-pin rule (`scripts/test_tier_audit.py`) flags `len(...) <op> N` assertions. Its regex `r"len\([^)]+\)\s*[<>=!]+\s*(\d+)"` matched the `len(` **substring** inside suffix-len helpers — `cell_len(`, `str_len(`, `byte_len(` — so a legitimate cell-width invariant like `assert cell_len(ch) <= 1` tripped a **false ERROR**. TUI width tests have been carrying a variable-binding workaround (`w = cell_len(hint); assert w <= N`) to dodge it.

## Fix

Add a leading `\b` word boundary → `r"\blen\(..."`. Because `_` is a word char, `cell_len` has no word boundary at its inner `len` and is excluded; a standalone `len(` (after `(`, space, `=`, `[`, or line start) still matches → genuine format-pins are detected exactly as before.

```
old: cell_len(ch) <= 1  → MATCH (false positive)
new: cell_len(ch) <= 1  → no match  ✓
new: len(x) <= 1        → MATCH (real pin still caught)  ✓
```

Audited the other line-regexes (`SNAPSHOT_RE` etc.) — they anchor on `.` + an exact method name, so they don't have the substring-match hazard; only `FORMAT_PIN_RE` needed the boundary.

## Tests (Tier 1)

New `tests/test_tier_audit_format_pin.py` drives the auditor's own `_audit_test` flow on inline source (same loader idiom as the sibling `test_tier_audit_private_state_ast.py`):
- `cell_len(...) <= N` / `str_len(...) <= N` → NOT flagged (the fix)
- bare `len(...) <= N` → STILL flagged (detection not weakened)
- `len(...) > 0` → still exempt (existence check, unchanged)

Meta-note: the auditor scans a test's own source lines (docstring included), so the regression test assembles its bare-len fixture via interpolation — no physical line in the file carries the contiguous `len(...) <op> N` shape, otherwise the test would trip the rule it pins. (This is itself a small instance of the canonical idiom the fix removes the *need* for in normal tests.)

## Follow-up (not in this PR)

The variable-binding workaround in existing TUI width tests is now unnecessary; I'm leaving those untouched (scope discipline) and will note in memory that the root cause is fixed so future tests can inline `cell_len(x) <= N` directly.

## Test plan

- [x] `pytest tests/test_tier_audit_format_pin.py` — 4 passed
- [x] `pytest tests/test_tier_audit_private_state_ast.py` — 11 passed (sibling audit self-test, no regression)
- [x] `scripts/test_tier_audit.py --strict tests/` — 6218 tests inspected, **0 errors** (fix doesn't break any real detection; the transient self-reference in the new test was resolved)
- [x] before/after regex proof (old matches `cell_len`, new excludes it, new still matches bare `len`)
- [x] `ruff check` clean

## Tier self-check

Tier-1 audit-contract tests (real auditor instance, public `_audit_test` surface). No format-pin / no private-state assert / no golden-file. ✓

### DIFF
--- a/scripts/test_tier_audit.py
+++ b/scripts/test_tier_audit.py
@@ -35,7 +35,16 @@ TIER_DOCSTRING_RE = re.compile(r"^Tier [123][abc]?:", re.IGNORECASE)
 
 # len(...) < N  /  len(...) > N  /  len(...) == N  /  len(...) >= N  etc.
 # Exemption: len(x) > 0  (simple existence check)
-FORMAT_PIN_RE = re.compile(r"len\([^)]+\)\s*[<>=!]+\s*(\d+)")
+#
+# The leading ``\b`` is load-bearing: without it the bare ``len\(`` matched
+# the ``len(`` *substring* inside suffix-len builtins/helpers like
+# ``cell_len(``, ``str_len(``, ``byte_len(`` — flagging legitimate
+# cell-width invariants (e.g. ``cell_len(ch) <= 1``) as format pins. A word
+# boundary requires a non-word char (or start-of-line) immediately before
+# ``len``; since ``_`` is a word char, ``cell_len`` has no boundary at its
+# inner ``len`` and is correctly excluded, while standalone ``len(`` (after
+# ``(``, space, ``=``, ``[`` …) still matches. (Issue #1082.)
+FORMAT_PIN_RE = re.compile(r"\blen\([^)]+\)\s*[<>=!]+\s*(\d+)")
 
 # Private-state detection is AST-based via ``_find_private_attr_access``
 # below. The prior implementation used a regex ``r"\.\w+\._\w+"`` which
--- a/tests/test_tier_audit_format_pin.py
+++ b/tests/test_tier_audit_format_pin.py
@@ -1,0 +1,133 @@
+"""Tier 1: format-pin detection in scripts/test_tier_audit.py excludes suffix-len fns.
+
+Pins the regex fix from issue #1082. The format-pin rule flags
+``len(...) <op> N`` assertions (= Tier 4 "見た目のフォーマット固定"). The
+prior regex ``r"len\\([^)]+\\)\\s*[<>=!]+\\s*(\\d+)"`` matched the ``len(``
+*substring* inside suffix-len helpers like ``cell_len(`` / ``str_len(``,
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read; merge on CI green). Surgical #1082 fix removing a tier-audit false-positive — the root cause of the variable-binding workaround idiom in TUI width tests.

## Fix (correct)
`FORMAT_PIN_RE` gains a leading `\b`: `r"\blen\([^)]+\)\s*[<>=!]+\s*(\d+)"`. Since `_` is a word char, `cell_len(` has no word boundary at its inner `len` → excluded; a standalone `len(` (start / after `(` / space / `=` / `[`) still matches. So `cell_len(ch) <= 1` (a width invariant) stops tripping a false ERROR while real format pins (`len(x) <= 1`) are caught exactly as before. The load-bearing `\b` is well-documented inline. Other line-regexes (`SNAPSHOT_RE` …) audited — they anchor on `.`+exact-name, no substring hazard.

## Tests (Tier 1) ✓
`test_tier_audit_format_pin.py` drives the **real auditor's own `_audit_test` flow** on inline source (importlib spec-load, no MagicMock — same idiom as the sibling AST self-test): cell_len/str_len NOT flagged (the fix), bare `len` STILL flagged (detection not weakened), `len > 0` still exempt. Cleanly handles the **meta-circularity** (the auditor scans the test's own source) by assembling the bare-`len` fixture via interpolation so no physical line carries the pinned shape.

## Discipline
- Scope-tight: leaves the now-unnecessary variable-binding workarounds in existing TUI tests untouched (noted as follow-up, not churned in). ✅
- Tier self-check done (Tier-1 audit-contract, real auditor, public surface, no pin/private-state/golden). ✅

This is the root-cause fix for [[feedback_test_audit_canonical_idioms]]'s variable-binding idiom — future tests can inline `cell_len(x) <= N` directly. Good catch on the #427 closure-candidate trace too (recorded in #427).

Merging once CI lands green.

PR #1147, opened 2026-05-31
fix(dogfood): llm-emitted-ops match real events 'type' field + add json-decode source (#1135)

tya5:
**[dogfood-coder]** — output of the #1135 end-to-end verification dispatch. The live-run half is blocked (local LiteLLM proxy down), but prepping it via code inspection surfaced a **real read-side bug** plus completed the json-decode read-side branch.

## The bug (in merged #1136, found via the end-to-end lens)
`mode_llm_emitted_ops` source-2 filtered events on `e.get("kind")`, but the P6 events log writes the type under **`type`** (`EventLog.emit` → `Event(type=...)`; observed on disk: `{"type": "phase_output_validation_failed", "data": {...}}`). So the read-side matched **zero real events**. The unit tests passed only because their fixtures used the wrong key (`"kind"`) — exactly the synthetic-event gap this dispatch targets. `_all_events` does no `type`→`kind` normalization, so source-2 was silently blind to production data.

## Fixes
- **`_event_type(e)`** helper (tolerates `type` + legacy `kind`); source-2 now matches real events. Verified no sibling instances — all other `.get("kind")` reads are on WAL records or LLM-trace request/response records, which legitimately use `kind`.
- **Source-3: `llm_output_json_decode_failed`** (#1141 sibling, PRE-parse). `raw_output` is unparseable by definition, so no `{kind,keys}` — show `failure_kind` + `error` + the position-aware-truncated raw slice verbatim. This is the read-side branch I committed to in #1135, now that the shape is locked.
- Tests rewritten to the **real on-disk shape** (`type`); the prior `kind` fixtures were the masking error.

## Test plan
- [x] 26 pass (`python -m pytest tests/test_dogfood_trace_llm_modes.py -q`), incl. 2 new: `test_emitted_ops_reads_real_type_field_not_kind` (regression guard — goes silent if reverted to `kind`), `test_emitted_ops_surfaces_json_decode_failure_event`
- [x] `ruff check` clean
- [x] Verified no sibling `type`/`kind` mismatches in events-log reads
- [ ] (deferred — live half) real events flowing from a live weak-model run → read-side isolates them. Blocked on the LiteLLM proxy being up; tracked for the next live capture (e2e's run or a chat session once the proxy is reachable).

### DIFF
--- a/scripts/dogfood_trace.py
+++ b/scripts/dogfood_trace.py
@@ -1248,6 +1248,19 @@ def mode_llm_advertised_ops(records: list[dict], request_id: str) -> None:
 # (existing validation_error/phase_failed etc. are UNCHANGED). failure_kind is
 # an explicit enum field on the event (not derived from the event name).
 _VALIDATION_FAIL_EVENT = "phase_output_validation_failed"
+_JSON_DECODE_FAIL_EVENT = "llm_output_json_decode_failed"
+
+
+def _event_type(e: dict) -> str:
+    """Return an event's type, tolerating both on-disk conventions.
+
+    The P6 events log writes the type under ``"type"`` (``EventLog.emit`` ->
+    ``Event(type=...)``); some WAL/legacy records use ``"kind"``. Match either
+    so the read-side finds real events (the events log) AND hand-written
+    fixtures. (Field is top-level — NOT under ``data``, which carries the
+    skill-level ``failure_kind`` enum, a different thing.)
+    """
+    return e.get("type") or e.get("kind") or ""
 
 
 def _op_kind_keys(op: dict) -> "tuple[str, list[str]]":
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read + verified the type/kind claim; merge on CI green). High-value fix — the trace-infra end-to-end dispatch surfaced a real read-side bug that the #1136 unit fixture masked.

## The bug (verified)
`EventLog.emit` (events.py:87/108) writes `Event(type=..., data=...)` → real events-log records are `{"type": "phase_output_validation_failed", "data": {...}}`. #1136's `mode_llm_emitted_ops` source-2 filtered on `e.get("kind")` → **matched zero real events**. The #1136 unit test passed only because its fixture used `"kind"` (the wrong field) — a fixture-doesn't-match-real-shape masking. (I reviewed + merged #1136; this is my review miss — same family as the #1142 default-round-trip blind spot: a test that passes without exercising the real shape/value.)

## Fix ✓
- `_event_type(e)` = `e.get("type") or e.get("kind")` → matches the real events-log `type` (and tolerates legacy/WAL `kind`).
- source-2 now filters via `_event_type` → finds real `phase_output_validation_failed` events.
- Adds **source-3** (`llm_output_json_decode_failed`, #1141 pre-parse) → failure_kind + error + raw head (no ops to extract for unparseable JSON).
- **Fixtures corrected** `"kind"` → `"type"` + a dedicated regression guard `test_emitted_ops_reads_real_type_field_not_kind` that uses the REAL shape and asserts a match (goes silent if the filter reverts to `kind`). 26 pass.

This is exactly what the end-to-end dogfood was for — the read-side reading the REAL event shape, not a synthetic fixture's wrong one. Good catch.

## On the live half (blocked, separate)
Live event emission needs a real Reyn run with model access; the local proxy (localhost:4000) is down (user infra). Natural path: capture in e2e's next model run → you isolate read-side. Flagging the proxy to the user. The read-side fix here stands on its own (verifiable without a live run via the real-shape regression test).

Merging once CI green.

PR #1148, opened 2026-05-31
feat(tui): "↓ N new" indicator when content lands while scrolled up

tya5:
**[tui-coder]** — Refs #1144 (#1 of the scroll/navigation UX exploration backlog). First PR of the scroll/nav topic.

## Problem

The conv pane already suppresses auto-scroll when the user scrolls up to read history (`_user_scrolled` + `auto_scroll` flip — `test_user_scroll_suppresses_auto.py`). But nothing **told** the user that new content had landed below their locked viewport — they had to blind-guess Alt+End to catch up. (Classic chat-scroll affordance gap; `grep` confirmed no existing indicator.)

## Fix

A docked dim-coral strip `↓ N new lines below · Alt+End to jump`, shown **only** while scrolled up AND content arrived since the lock. Collapses to zero height (`.hidden`) in the default following case → layout unchanged when not scrolled up. Clicking the strip jumps to the bottom (= Alt+End).

## Design

- **Single detection chokepoint** (not per-write instrumentation): a watcher on the RichLog's `virtual_size` reactive fires on every write (content growth), covering all write paths without touching each one. Mirrors the existing `scroll_y` watcher next to it.
- **Count** = `max_scroll_y - baseline`, where the baseline is `max_scroll_y` captured at the instant the user scrolls up (the False→True `_user_scrolled` transition in `_on_log_scroll_y`).
- **Clear** on return to the tail via any path — scroll, Alt+End, submit (`_snap_to_bottom`), or `clear()` — drops the baseline + hides the strip.
- `_CORAL` (action accent, "jump here") on a subtle `_BG_HEADER` strip; CSS follows this file's existing literal-hex + `/* _TOKEN */` convention (the empty-hint precedent).

## Scope

`src/reyn/chat/tui/widgets/conversation.py` only — CSS rule, `compose()` Static, `__init__` state, the `virtual_size` watch, baseline capture in `_on_log_scroll_y`, `_on_log_content_grew` / `_update_new_below` / `on_click`, and a `new_below_count` public accessor. No OS/cross-process surface.

## Tests (Tier 2)

`tests/cli/test_conv_new_below_indicator.py` — real `ReynTUIApp` harness (same pattern as `test_user_scroll_suppresses_auto.py`), public surface only (`conv.new_below_count` accessor + the `#new-below` strip's `hidden` class):
- following the tail → count stays 0 as content writes
- scrolled up + later writes → count > 0 (strip visible)
- return to bottom → count resets to 0 (strip hidden)
- fresh pane → strip starts hidden

## Test plan

- [x] `pytest tests/cli/test_conv_new_below_indicator.py` — 4 passed
- [x] `pytest tests/cli -k "scroll or user_scroll or conversation or auto_scroll"` — 57 passed (regression: `_on_log_scroll_y` / `_snap_to_bottom` changes don't break existing scroll-lock behavior)
- [x] `pytest tests/cli/test_tui_smoke.py` — 40 passed (CSS parses, ConversationView mounts with the new Static)
- [x] `ruff check` clean
- [x] `scripts/test_tier_audit.py --strict` clean (0 errors)
- [ ] (skipped — visual) live `reyn chat`: scroll up mid-stream, confirm `↓ N new` appears + Alt+End/click jumps + it clears — behavior is covered by the 4 Tier-2 harness tests; geometry-dependent visual is a manual follow-up

## Tier self-check

Tier-2 behavior tests, real instances, public accessor + DOM class assertions. No format-pin / no private-state assert / no golden-file. ✓

### DIFF
--- a/src/reyn/chat/tui/widgets/conversation.py
+++ b/src/reyn/chat/tui/widgets/conversation.py
@@ -361,6 +361,21 @@ class ConversationView(Widget):
     ConversationView #empty-hint.hidden {
         display: none;
     }
+    /* New-messages-below affordance: a 1-line dim-coral strip docked just
+       above the sticky/async area, shown ONLY while the user is scrolled up
+       AND content arrived after they locked. ``.hidden`` collapses it to
+       zero height in the (default) following case so the layout is unchanged.
+       Coral = the action-accent (= "jump here"). */
+    ConversationView #new-below {
+        dock: bottom;
+        height: auto;
+        color: #C8553D;  /* _CORAL */
+        background: #1a1a1a;  /* _BG_HEADER — subtle strip so it reads as chrome */
+        padding: 0 1;
+    }
+    ConversationView #new-below.hidden {
+        display: none;
+    }
     """
 
     def __init__(self, *, id: str | None = None) -> None:
@@ -437,6 +452,16 @@ class ConversationView(Widget):
         self._recent_replies: list[str] = []
         # Track whether user has scrolled up (suppress auto-scroll while scrolled)
         self._user_scrolled = False
+        # New-messages-below affordance: ``_new_below_baseline`` is the log's
+        # ``max_scroll_y`` captured at the instant the user scrolled up (=
+        # locked their view). While locked, new content grows ``max_scroll_y``;
+        # the delta from the baseline is the row count shown in the ``↓ N new``
+        # indicator. ``-1`` = "not locked" (= following the tail), so the
+        # content-grew watcher knows to stay hidden.
+        self._new_below_baseline: int = -1
+        # Current row count shown in the ``↓ N new`` strip (0 = hidden).
+        # Exposed read-only via the ``new_below_count`` property.
+        self._new_below_count: int = 0
         # Issue 5 — track mounted ErrorBoxes for Escape-to-dismiss
         self._error_boxes: list[ErrorBox] = []
         # Cursor into ``_error_boxes`` for the F5 / F6 jump-to-error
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **ACCEPT** (diff-read; merge on CI green). Clean TUI affordance, #1144 #1.

## Structural gates (TUI/UX calibration)
- **No OS surface** — `conversation.py` only; pure TUI widget, P7-clean. ✓
- **Mechanism is sound**: detects content-below via the RichLog `virtual_size` reactive — a **single chokepoint** that covers every write path without instrumenting each (writes grow `virtual_size` but not `scroll_y` while scrolled up, so the existing scroll watcher can't see them; this is the right signal). Baseline captured only on the False→True scroll-up transition (so scrolling within the locked region doesn't reset the count); `count = max_scroll_y - baseline`; cleared on tail-return (`_snap_to_bottom` + at-bottom `_on_log_scroll_y`) and click-to-jump. **Fail-open** (`try/except` around the `watch` + widget queries) so a Textual watch-API change can't break the conv pane — appropriate for a non-critical affordance. ✓
- **Layout-preserving**: `dock: bottom`, yielded last, `.hidden` collapses to zero height in the default following case. ✓
- **`on_click` scoped** to `target_id == "new-below"` — doesn't hijack other click targets. ✓

## Tier-2 test (calibration-compliant) ✓
Real `ReynTUIApp` harness (same pattern as `test_user_scroll_suppresses_auto`); asserts via the **public `new_below_count` accessor + the `#new-below` `hidden` class** — not private state. Behavioral contract: following-tail → 0 even as content writes / scrolled-up + writes → >0 / return-to-bottom → 0. scroll regression 57 + smoke 40 + ruff + tier-audit green.

The UX itself (coral strip, wording, feel) is your exploration call on the user's selected topic — the structure is solid. Merging once CI green. #2-4 (find-match count / Home-End·g-G / half-page) after this lands, per your plan.

PR #1164, opened 2026-05-31
fix(config): update LiteLLM model prefix openai/gemini→gemini/gemini (#1162 A)

tya5:
**[docs-maintainer]** — #1162 (A) config-level fix dispatched by lead-coder. Sandbox_2 fixture-owner sign-off received.

## Problem

`openai/gemini-2.5-flash-lite` prefix causes litellm catalog lookup to fail → falls back to hardcoded 128K context limit → reyn over-compacts at 87% context use (actual limit is 1M).

## Fix

Replace `openai/gemini-*` with `gemini/gemini-*` in all tracked config files. `gemini/` prefix resolves correctly in the litellm catalog to the 1M token limit.

## Files updated

| File | Notes |
|------|-------|
| `reyn.yaml` | Default project config |
| `cookbook/configs/with-mcp.yaml` | Cookbook config |
| `dogfood/fixtures/fizzbuzz_*/reyn.yaml` × 3 | sandbox_2 sign-off: routing-only, proxy path unchanged |
| `dogfood/fixtures/skill_importer_chain/reyn.yaml` | Same |
| `docs/reference/config/reyn-yaml.md` + `.ja.md` | Example blocks only (builtin class table unchanged — follows builtin_models.py, addressed in B) |

## Regression guard

`tests/test_litellm_gemini_prefix_1162.py` (Tier 2) — parametrised over all 6 tracked config files, blocks any future `openai/<gemini>` prefix reintroduction.

## Not in scope (B follow-up)

`src/reyn/llm/builtin_models.py` prefix-strip-retry robustness (#1162 B).

## Test plan

- [x] Regression guard test: 6/6 pass
- [x] sandbox_2 sign-off (fixture-owner, routing verified via dead-port test)
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1166, opened 2026-05-31
fix(docs): update instructional gemini model prefix (openai/→gemini/)

tya5:
**[docs-maintainer]** — follow-up to #1164. Instructional/template files that show `openai/<gemini>` as example config — new users copying these would get the 128K context limit bug.

## Files

- `.github/ISSUE_TEMPLATE/bug_report.yml` — example model string in bug report template
- `docs/deep-dives/contributing/dogfood-tooling.md` — config example block

## Not changed (historical — intentional)

`cookbook/*/transcript.example.txt` and `docs/deep-dives/journal/*` record what actually ran at the time. Changing these would falsify history.

## Test plan

- [x] 2-file diff, no behaviour change
- [x] `grep openai/gemini .github/ docs/deep-dives/contributing/` → 0 remaining instructional occurrences

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -38,12 +38,12 @@ body:
         - Python version (`python --version`)
         - OS and version
         - Reyn version (`reyn --version`, or `pip show reyn`)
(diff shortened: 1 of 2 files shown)


tya5:
**[docs-maintainer]** Closing in favour of #1167 (fold per lead-coder dispatch: instructional examples should use class-ref, not gemini/ hardcode).

PR #1168, opened 2026-05-31
fix(models): localise Gemini model strings to builtin_models.py (#1167)

tya5:
**[docs-maintainer]** — #1167 localize dispatch from lead-coder. Folds #1166.

## Problem

Gemini model strings were hardcoded in 6+ yaml files and docs, making future prefix/model changes a multi-file operation prone to the `openai/gemini-*` → 128K context bug (#1162).

## Solution

Tracked config files use class-ref shorthands; the canonical model string lives only in `builtin_models.py`.

## Changes

| File | Change |
|------|--------|
| `src/reyn/llm/builtin_models.py` | `openai/gemini-*` → `gemini/gemini-*` for 3 entries; add `gemini-pro` class |
| `reyn.yaml` | `gemini/gemini-2.5-flash-lite` → `gemini-flash-lite`, strong → `gemini-pro` |
| `cookbook/configs/with-mcp.yaml` | Same |
| `dogfood/fixtures/*/reyn.yaml` ×4 | Same (sandbox_2 sign-off from #1164) |
| `docs/reference/builtin-models.md` | Update prefix, add gemini-pro entry |
| `docs/reference/config/reyn-yaml.md` + `.ja.md` | Update builtin table + examples |
| `.github/ISSUE_TEMPLATE/bug_report.yml` | Class-ref in example (was #1166) |
| `docs/deep-dives/contributing/dogfood-tooling.md` | Class-ref in example (was #1166) |
| `tests/test_builtin_models.py` | Add `gemini-pro` to EXPECTED_NAMES |
| `tests/test_litellm_gemini_prefix_1162.py` | Extended guards: no hardcoded provider/gemini- in configs; gemini entries use gemini/ prefix |

## Test plan

- [x] 48 tests pass (guard + builtin_models)
- [x] sandbox_2 dogfood fixture sign-off (carried from #1164)
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1157, opened 2026-05-31
feat(chat): #1125 Item 2 — wire retry_loop into chat overflow path

tya5:
**[lead-coder]** — #1125 Item 2. PR-N6 built `retry_loop` (bounded adaptive-shrink overflow recovery) but the chat router **never called it** — the overflow handler ran a degraded compact-once (`session.py` old 5078-5113, which self-admitted "full retry_loop ... tracked as a follow-up") and hard-failed if the first compaction itself overflowed. axis-1's eager 30K trigger masked the gap by keeping the middle small. This wires the real mechanism in — **the prerequisite for #1128 axis-1 removal** (owner-signed-off sequence: wire → then remove).

Implementation owner: lead-coder (OS-core chat compaction). Owner sign-off on this scope: #1125 comment 4585865248.

## What changed
- **`ChatSession._decompose_history_for_retry`** exposes the `head/raw_middle/tail/summary` decomposition retry_loop consumes. Reuses the `head_size`/`tail_size` slicing + the documented Q4-attractor overlap guard; factors per-turn wire serialisation into **`_serialise_turn`** so the recovery prompt is byte-identical to what the normal path (`_build_history_for_router`) sends.
- **Router overflow handler** (`_run_router_loop`) hands that decomposition to `retry_loop` with the session's real engine / learner / cfg. `engine.budgets` is non-None by construction (engine `__init__` always sets `_budgets`; the pre-frame guard also already ran `recompute_budgets()`). On `UnrecoveredError` / `ContextOverflowError` → fail-fast user-visible error (chat axis contract, preserved from the degraded path).
- **`_RouterUsageShim`** exposes RouterLoop's `TokenUsage` as `.usage` so retry_loop's adaptive learner observes the chat axis's actual prompt size.

## Design notes
- The **pre-frame guard** (`_maybe_force_compact_for_router`) is **unchanged** — proactive persistence each turn. retry_loop is the **reactive safety net** when the guard's estimate under-counts (actual > estimate). No persistence added inside retry_loop (the guard owns it; retry_loop fires only on the estimate-miss exception, not every turn).
- Reuses the **same idempotency contract** the prior degraded path already shipped (re-run `loop.run` after overflow): overflow raises at the LLM call **before** any tool dispatch, so bounded re-runs with shrunk history are safe. retry_loop generalises "retry once" → "retry ≤8 with monotonic shrink"; it does not introduce a new idempotency assumption.
[...]

PR #1179, opened 2026-05-31
fix(model-budget): #1162-(B) provider-prefix-strip-retry — resolve proxy-routed models' real window

tya5:
**[e2e-coder]** — #1162 code-robustness half (the config-prefix half landed in #1164/#1168). sandbox_2 found the bug via #1157 dogfood; this completes the #1162 thread. (Crossed lead-coder's load-balance retract in-flight — already finished + tested when it arrived; lead-coder confirmed I PR my branch rather than waste the work; media-count cap went to sandbox_2.)

## Bug
`get_max_input_tokens("openai/gemini-2.5-flash-lite")` (Gemini routed via an openai-compat proxy) → litellm has no such model under the openai catalog → exception → **128K fallback**, while the real window is **1M**. reyn then fires pre-frame guard / early compaction at 128K, wasting **~87%** of the real context (premature over-compaction).

## Fix — provider-prefix-strip-retry (resolution-only, safe)
`get_max_input_tokens`: on initial catalog miss/exception, **strip the first provider segment and retry the bare name** (`openai/gemini-2.5-flash-lite` → `gemini-2.5-flash-lite` → 1M). A still-unknown bare name keeps the 128K fallback (and still emits `model_budget_fallback`) — strip only *improves* resolution, never hides a genuinely-unknown model. Extracted `_lookup_max_input` as the no-fallback/no-event catalog probe so the retry composes cleanly.

Defense-in-depth: #1168 made the default config class-ref'd (immediate fix), but this protects any operator who later hardcodes a literal `<provider>/<model>` proxy string.

## Test plan (Tier 2, real litellm — no mocks)
- [x] `tests/test_model_budget.py`: `openai|anthropic|vertex_ai / gemini-2.5-flash-lite` all resolve to the bare model's real (>128K) window via prefix-strip (parametrized, skips if catalog lacks the model); prefix-strip resolution emits NO `model_budget_fallback`; genuinely-unknown prefixed model keeps 128K + still emits the event. 9 model_budget tests pass; 248 budget/compaction sweep pass.
- [x] ruff clean.

Refs #1162

### DIFF
--- a/src/reyn/llm/model_budget.py
+++ b/src/reyn/llm/model_budget.py
@@ -35,6 +35,24 @@ _FALLBACK_MAX_INPUT_TOKENS = 128_000
 _warned_models: set[str] = set()
 
 
+def _lookup_max_input(model: str) -> "int | None":
+    """Return ``max_input_tokens`` from LiteLLM's catalog for *model*, or None
+    when the model is unrecognized or has no positive window.
+
+    No fallback, no events — pure catalog lookup so callers can compose retries
+    (e.g. provider-prefix-strip, #1162).
+    """
+    try:
+        import litellm
+        info = litellm.get_model_info(model)
+        max_input = info.get("max_input_tokens")
+        if max_input and int(max_input) > 0:
+            return int(max_input)
+    except Exception:
+        pass  # Not in catalog / no positive window.
+    return None
+
+
 def get_max_input_tokens(
     model: str,
     *,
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ ACCEPT + merging. Independent-context review.

## Code
`_lookup_max_input` helper extracted (pure catalog lookup, no fallback/events) → `get_max_input_tokens` uses it, then on miss with `/` in the name strips the first provider segment and retries the bare name; a still-unknown bare name falls through to the **unchanged** 128K fallback + `model_budget_fallback` event. **Resolution-only improvement, fallback path untouched** — safe even if a proxy aliases the prefixed name (bare lookup just misses → 128K, as before). Comments document the safety reasoning clearly.

## Tier rule check ✓
- Docstrings declare Tier (all `Tier 2:`) ✓ (the audit-fix commit corrected the one `Tier 2 (regression guard):` → `Tier 2:` format — benign, no semantic change)
- No Tier 4 / no mocks — **real litellm** catalog lookup ✓
- Invariant is essential (real budget window resolution, not format-pinning) ✓
- Env-robust: `skip` guard when the catalog lacks the probe model in-env ✓
- Regression guard: unknown-prefixed still falls back to 128K **and** still emits the event ✓
- audit job passes ✓

## CI
pytest 3.11 ✓ / 3.12 ✓ / ruff ✓ / test-tier audit ✓ — all green.

Completes the **#1162 thread**: A (config prefix, #1164/#1168) + B (code robustness, this). Squash-merging.

PR #1194, opened 2026-05-31
fix(1128): PR-b — serialize A2A run_one_iteration via shared per-agent lock

tya5:
**[e2e-coder]** — fix(1128): PR-b — serialize A2A `run_one_iteration` via shared per-agent lock

## Summary

**Latent race fixed:** `_await_skill_completion` in `a2a.py` called `session.run_one_iteration()` with no per-agent lock, while MCP already held `_get_agent_lock(agent_name)` around its own drives of the same method. A concurrent MCP request and A2A inbox-drain loop on the same shared `ChatSession` raced at every `await` point inside `run_one_iteration`, allowing both to mutate `session.history` concurrently.

**Critical correctness: shared lock registry.** If MCP and A2A each had their own `_AGENT_LOCKS` dict, a concurrent MCP+A2A pair would still bypass the serialization. This PR lifts the registry to a shared module so both transports obtain the **same `asyncio.Lock` object** for a given agent name.

### New: `src/reyn/chat/agent_locks.py`

Shared module exposing `get_agent_lock(agent_name: str) -> asyncio.Lock` backed by a module-level `dict[str, asyncio.Lock]`. Keyed by agent name string — the same identity `AgentRegistry.get_or_load` uses.

### `src/reyn/mcp_server.py`

- Replaced local `_AGENT_LOCKS: dict[str, asyncio.Lock]` and `_get_agent_lock()` function with `from reyn.chat.agent_locks import get_agent_lock as _get_agent_lock`.
- The alias `_get_agent_lock` preserves the existing call site at line 215 (`async with _get_agent_lock(agent_name):`) exactly — behavior unchanged.

### `src/reyn/web/routers/a2a.py`

- Added `from reyn.chat.agent_locks import get_agent_lock` import.
- `_await_skill_completion()` gained a new required keyword argument `agent_name: str`.
- Each `run_one_iteration()` call in the drain loop is now wrapped with `async with get_agent_lock(agent_name):`.
- **Wrap granularity**: lock is held for the duration of one iteration only, NOT the full polling loop. This mirrors MCP's wrap (which covers one full `MessageBus.request` turn) without blocking MCP for the entire `deadline_s` (up to 600s).
- The `_monitor()` closure in `_escalate_to_task` already had `agent_name` in scope — the call site update is a one-line addition.

### `tests/test_agent_locks_shared_1128.py` (new)

Tier 2 OS invariant tests pinning:
- (a) `get_agent_lock("x")` is idempotent — identical object on repeated calls.
- (b) Different names yield distinct lock objects.
- (c) MCP and A2A obtain the SAME lock object for the same agent_name — imports both accessors and asserts `is` identity. This is the central cross-transport guarantee.
- (d) Behavioral: concurrent coroutines are serialized — second waiter enters only after first releases.

### `tests/test_a2a_sync_async_escalation.py` (updated)

Three existing direct `_await_skill_completion(...)` calls updated to pass the new required `agent_name="test-agent"` kwarg.

## A2A `run_one_iteration` call sites — grep-confirmed complete set

```
$ grep -n "run_one_iteration" src/reyn/web/routers/a2a.py
578:        await session.run_one_iteration()
```

**Exactly one site.** It is inside `_await_skill_completion`'s drain loop. Wrapped with `async with get_agent_lock(agent_name):` in this PR.

## Test plan

- [x] `python -m pytest -q tests/test_agent_locks_shared_1128.py` — 4 passed
- [x] `python -m pytest -q -k "a2a or mcp_server or agent_lock"` — 167 passed, 1 skipped
- [x] `python scripts/test_tier_audit.py --strict tests/test_agent_locks_shared_1128.py tests/test_a2a_sync_async_escalation.py` — 0 errors

Refs #1128

### DIFF
--- a/src/reyn/chat/agent_locks.py
+++ b/src/reyn/chat/agent_locks.py
@@ -1,0 +1,46 @@
+"""Per-agent serialization lock registry — shared by ALL transport layers.
+
+MCP (``reyn.mcp_server``) and A2A (``reyn.web.routers.a2a``) both drive
+``ChatSession.run_one_iteration`` for the same shared session instance. Without
+cross-transport coordination a concurrent MCP request and an A2A drain loop on
+the same agent race at every ``await`` inside ``run_one_iteration``, corrupting
+``session.history`` (both callers mutate it concurrently).
+
+The fix: every transport that drives ``run_one_iteration`` acquires the same
+per-agent ``asyncio.Lock`` before entering the critical section.
+
+This module is the single registry for those locks.  Both ``mcp_server`` and
+``a2a`` import ``get_agent_lock`` from here; they therefore share the same
+``_AGENT_LOCKS`` dict and thus the same lock object for any given agent name.
+
+Design constraints:
+- Module-level singleton dict: one process, one event-loop — the module is
+  imported once and the dict is stable for the process lifetime.
+- ``asyncio.Lock`` is not thread-safe, but Reyn runs on a single asyncio
+  event loop, so this is fine.
+- The lock is keyed by the *agent name string* — the same identity used by
+  ``AgentRegistry.get_or_load``.  Transport callers must use the canonical
+  name (not a URL slug or alias) so the key matches.
+"""
+from __future__ import annotations
+
+import asyncio
+
+# Module-level registry — intentionally a plain dict.
+# Callers must NOT mutate this dict directly; use ``get_agent_lock`` only.
+_AGENT_LOCKS: dict[str, asyncio.Lock] = {}
+
+
+def get_agent_lock(agent_name: str) -> asyncio.Lock:
+    """Return the per-agent ``asyncio.Lock`` for *agent_name*.
+
+    Creates the lock on first access; subsequent calls with the same name
+    return the **same** ``asyncio.Lock`` object (identity guarantee).  Locks
+    for different names are distinct objects.
+
+    Thread-safety: this function is NOT thread-safe (dict.setdefault is not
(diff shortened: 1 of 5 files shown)


tya5:
**[lead-coder]** — ✅ ACCEPT + merging. Independent review of the root-fix.

The critical correctness is the **shared lock registry** (`reyn.chat.agent_locks`): MCP and A2A import the same `get_agent_lock`, so the module-level `_AGENT_LOCKS` dict yields the **same `asyncio.Lock` object** per agent name — a separate per-transport dict would have left the race open. Test (c) pins exactly that identity (`mcp._get_agent_lock(a) is agent_locks.get_agent_lock(a)`). A2A wraps each `run_one_iteration` (per-iteration, not the whole polling loop → MCP isn't blocked for `deadline_s`). This preserves the humanic multi-source model (inbox queues all sources) while making processing serial.

## Tier rule check ✓
Tier 2 ✓ · no Mock (real `asyncio.Lock` via public `get_agent_lock`) ✓ · invariants essential (cross-transport lock identity + behavioral serialization) ✓ · no private-state/count pins (identity + non-overlap) ✓ · audit passes ✓.

## Note (documented precondition, follow-up)
The lock key is the agent_name string — the fix only serializes if A2A and MCP pass the **same canonical name** for one agent (the module docstring flags this). A2A's name flows through the registry, so it's likely canonical, but worth a follow-up confirm that A2A doesn't key on a URL slug/alias.

CI all green. Closes the pre-existing A2A↔MCP concurrent-history-mutation race + unblocks PR-c (`max_passes`→1, now by-construction safe). Squash-merging.

PR #1195, opened 2026-05-31
feat(1128): step 4 PR-a — remove axis-1 background compaction + vestigial lock

tya5:
**[e2e-coder]** — #1128 step 4 **PR-a** (the mechanical + provable part of the chat-compaction simplification). Design + Q1/Q2/Q3 analysis articulated on #1128. Refs #1128.

⚠️ **MERGE HELD pending Q2 dogfood** (sandbox_2 verifies the timing-behavior shift below) — per lead-coder. Ready for review meanwhile.

## What

- **§1 axis-1 removal** — `CompactionController.spawn_maybe`/`_maybe_compact`/`cancel` + the `asyncio.Task` lifecycle; the post-reply `spawn_maybe()` call + `_drain_on_shutdown` cancel in session.py; `trigger_total_tokens` + `min_compact_batch` config fields + loader. `force_compact_now` (the synchronous pre-frame guard) is now the sole controller-driven trigger.
- **§2 vestigial `compaction_lock` removal** — only `force_compact_now` acquired it; no history appender ever awaited it, so it never serialized appenders. Cross-driver turn serialization is the A2A per-agent lock's job (**PR-b, merged a12a819e**). (`max_passes 2→1` is PR-c.)
- **§4 #1125 I1-a fix** — `head_size`/`tail_size` were wrongly marked "deprecated / background-only"; `force_compact_now` uses them for candidate windowing. Corrected.
- **Q3 dogfood batch** — `compaction_grant` injected `trigger_total_tokens` to exercise the removed auto-fire path; now injects only `head_size`/`tail_size`. The `chat_compactor` scenario must drive `/compact` explicitly.
- `reyn.yaml` cleaned of the removed keys.

## Behavior shift (Q2 — why merge is held)

Auto-compaction moves from the **30K absolute** background trigger to the **window-relative** pre-frame `effective_trigger` + on-demand (`/compact`, compact op) + `retry_loop` backstop. Between ~30K and `effective_trigger` the middle accumulates raw and is elided from the head+tail router view with no summary — intended window-utilization-first shift. **sandbox_2 dogfood (Q2)** verifies this gap is acceptable before merge.

## Test plan

- [x] Touched-file tests (9 files) — 84 passed. Axis-1 behavior tests deleted (spawn_maybe/_maybe_compact/single-flight/cancel + the Axis-8 `compaction_lock` test); controller-invariant file replaced with `force_compact_now` invariants (no-candidates / success-append / engine-failure).
- [x] `ruff check` (all touched) — clean
[...]

PR #1196, opened 2026-05-31
feat(1128): step 4 PR-c — collapse force_compact_now to a single pass (max_passes 2→1)

tya5:
**[e2e-coder]** — #1128 step 4 **PR-c**. Refs #1128. Depends on **PR-b (#1194, merged a12a819e)**.

## What

Now that cross-driver turn serialization is structural (PR-b: every transport driving `run_one_iteration` holds the shared per-agent lock, and within a turn `_append_history` is synchronous), the synchronous `force_compact_now` can no longer race with a concurrent history append — so the Option-B race-recovery loop is unnecessary.

- Remove the `for pass_n in range(max_passes)` loop → **single pass**.
- Remove the post-pass re-measure + the `ForceCompactRaceUnrecoveredError` raise. Contract shifts from "succeeds within max_passes OR raises" to "single pass; the `retry_loop` overflow backstop (already wired in `_run_router_loop`) is the under-shoot floor" — still never a silent over-budget prompt.
- Drop the `max_passes` parameter.
- Remove the now-dead `_estimate_current_history_tokens` (only the re-measure used it).
- `ForceCompactRaceUnrecoveredError` class + exports left in place (no longer raised here; harmless, still referenced by the compact op's error-string handling).

## Coordination with PR-a (#1195)

Both edit `force_compact_now` + `tests/test_chat_compaction_engine_11axis.py`. PR-a (Q2-held) removes the **lock** (keeps the loop); PR-c removes the **loop** (keeps the lock). The reconciled state is single-pass + no-lock. Whichever merges second rebases — I'll absorb it. lead-coder is aware; PR-c is by-construction safe post-PR-b and merge-independent of PR-a's Q2 gate.

## Test plan

- [x] Replaced the two ISSUE #6 race-recovery tests (multi-pass raise / budget-recovered) with a single-pass invariant (`compact()` invoked once, no `force_compact_race_unrecovered` event).
- [x] Affected tests (`11axis`, `compaction_controller_invariants`, `slash_compact_191`, `compact_op_272`, `retry_loop_chat_wiring`) — 55 passed.
- [x] `ruff check` — clean · `scripts/test_tier_audit.py --strict` (11axis) — 0 errors
- [x] Affected-area sweep (`-k "compaction or session or chat or slash or retry_loop or compact_op"`) — 693 passed
- [x] Full CI-faithful suite — **6696 passed**, 5 skipped, 3 xfailed. The 1 failure (`test_web_fetch_preview_path_ref`) is pre-existing on the base + passes in CI (env-dependent HTML-extraction quirk, unrelated).

### DIFF
--- a/src/reyn/chat/services/compaction_controller.py
+++ b/src/reyn/chat/services/compaction_controller.py
@@ -246,120 +246,62 @@ class CompactionController:
         finally:
             self._compacting = False
 
-    def _estimate_current_history_tokens(self) -> int:
-        """Cheap chars/4 estimate of the total text tokens in current history.
-
-        Used by the race-recovery loop in :meth:`force_compact_now` to
-        re-measure after each compaction pass.
-        """
-        history = self._history_access()
-        return sum(
-            _estimate_tokens(m.text)
-            for m in history
-            if m.role in ("user", "assistant", "tool", "agent")
-        )
-
-    async def force_compact_now(self, *, max_passes: int = 2) -> None:
-        """Synchronous force-trigger with race-recovery loop (ISSUE #6, Option B).
+    async def force_compact_now(self) -> None:
+        """Synchronous force-trigger — single pass (#1128 PR-c).
 
         Used by the pre-frame guard in ``_maybe_force_compact_for_router`` when
         the projected prompt would exceed the model's max_input_tokens.  Emits
-        the same events as the background path but with ``outcome="forced_sync"``
-        on the ``compaction_check`` event.
+        ``compaction_check`` with ``outcome="forced_sync"``.
 
-        Race tolerance (Option B):
-            Between the fire-decision and lock-acquisition, other async
-            coroutines may append to history (sync ``_append_history`` calls).
-            After each compaction pass, we re-measure the post-state and re-run
-            if still over budget, up to ``max_passes`` total.  Beyond that, a
-            ``force_compact_race_unrecovered`` event is emitted and
-            ``ForceCompactRaceUnrecoveredError`` is raised — the contract is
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ ACCEPT + merging. Independent review.

`force_compact_now` collapses to a single pass: the `for pass_n in range(max_passes)` loop + re-measure + `ForceCompactRaceUnrecoveredError` *raise* + dead `_estimate_current_history_tokens` removed; `max_passes` param dropped. Contract shifts to "single pass; if it under-shoots, the `retry_loop` backstop folds + monotonically shrinks" — never a silent over-budget prompt. The `ForceCompactRaceUnrecoveredError` *class* is correctly retained (compact-op error-string handling references it). Tests re-targeted: the race-recovery cases → a single-pass invariant.

**Safety verified independently**: I checked the `self._compacting` flag (compaction_controller.py:146/193/300) — both axis-1 `_maybe_compact` AND `force_compact_now` test-and-set it, so the two compaction paths are single-flighted (no concurrent compaction). Turn appends are serialized by PR-b's per-agent lock. So removing the race-recovery is safe after PR-b, **independent of PR-a** (axis-1's presence doesn't reintroduce the race — the `_compacting` flag covers axis-1↔force_compact_now). Confirmed e2e's independence claim rather than assuming it.

## Tier rule check ✓
Tier 2 ✓ · no Mock ✓ · invariant essential (single-pass behavior + retry_loop floor) ✓ · no format/count pins ✓ · audit passes ✓.

CI all green. PR-a will rebase its lock-removal on top (→ single-pass + no-lock). Squash-merging.

PR #1203, opened 2026-06-01
fix(swe_bench): verify routes only to report (P1/P8); drop un-satisfiable graph edges

tya5:
**[e2e-coder]** — ① of the C7 faithful-eval fix cascade (lead-coder-reviewed), layered on top of #1202 (③). Root-fixes the P1/P8 violation in the swe_bench verify phase that crashed astropy-13453.

## Problem (primary evidence)
astropy-13453 failed hard at verify after 3 attempts:
```
Artifact data validation failed for 'verify_state': 'edits' is a required property;
'rationale' is a required property
```
Trace of the cause:
- `verify.md` prescribed **"transition back to plan"** (Step 4) and **"transition back to apply"** (convergence guard). A Phase must not name its next phase (**P1**); instructions must not prescribe transitions (**P8**) — the OS injects `candidate_outputs` at runtime.
- `"plan"` was **not even a verify graph candidate** (`graph: verify → [report, apply]`) — the instruction named an impossible target.
- `verify → apply` **is** in the graph but is **un-satisfiable**: apply's `input` is the `plan` artifact (requires `edits`/`rationale`), which verify — producing `verify_state` — cannot emit. Told to transition there, the model was forced to fabricate `edits`/`rationale` and failed validation → the crash.
- `skill.md` internally contradicted itself: diagram says "tests fail → back to **plan**", table says "back to **apply** on fail", graph says `[report, apply]`.

| edge | verify emits | target accepts | satisfiable? |
|------|-----|-----|-----|
| verify→report | verify_state | verify_state | ✅ |
| verify→apply | verify_state | **plan** (edits/rationale) | ❌ fired the crash |

## Fix (minimal, P1/P8-compliant)
- **Graph**: `verify → [report]` — the sole satisfiable transition (report accepts `verify_state`). Both pass and fail terminate at report, which carries the verdict (`tests_passed` + `failure_summary`) and produces the best-effort patch.
- **verify.md**: every step describes the **domain outcome only** (record `tests_passed` + `failure_summary`) — no hardcoded next phase. The OS offers `[report]`.
- **skill.md**: diagram + table reconciled to the actual linear flow.

The test-failure **re-plan loop** (verify → plan to revise the fix) is a **tracked enhancement**, not wired here. Notable correction to the earlier framing: `plan.md` input is already `exploration | verify_state`, so plan *can* receive the failure artifact — the loop is a small follow-up (1 graph edge + re-plan guidance), not a large rework. I'll file the ticket.

## Test plan
[...]

Issue #1176, opened 2026-05-31
compact op — phase axis: on-demand phase force-compact seam (B1 follow-up to chat-axis #272/#1128)


tya5:
**[e2e-coder]** — deliberate follow-up to the chat-axis compact op (PR on branch `feat/272-compact-op`). lead-coder agreed (broker, B2): land chat-axis first, design the phase axis carefully here rather than rush a mid-phase-state mutation.

## Why this is separate from chat (code-verified finding)
The LLM-requested `compact` op is wired for **chat** because chat is the live conversation dead-end AND has an on-demand compaction seam (`ChatSession.force_compact_now`). The **phase** axis is genuinely asymmetric:
- Phase compaction = `compact_control_ir_results` (services/compaction/engine.py), invoked **automatically at frame-build time** by `phase_executor.py:421`, gated on `len(control_ir_results) > recent_act_turns_raw`.
- There is **no on-demand force-compact seam** for phases — the OS already auto-summarises older act-results every frame.

So a phase `compact` op cannot simply wrap an existing call (unlike chat). The LLM-requested-compact gap in phases is also smaller because auto-compaction covers the steady state.

## Proposed design (B1) — for review before implementation
1. **On-demand phase force-compact**: a method on the phase runtime/executor that compacts the currently-accumulated `control_ir_results` immediately (reusing `compact_control_ir_results`), replaces them in phase state, and returns `{freed_tokens, free_window_after, free_window_before}` (exact tokens, same contract as chat's `_compact_now_for_op`).
   - Open question: mid-phase mutation of `control_ir_results` — must be safe w.r.t. the frame currently being assembled + WAL/resume (P5/P6). Needs careful spec.
2. **Wire `OpContext.compact_now`** at the phase control_ir build site (`control_ir_executor.py`) to (1) — the op handler + `CompactIROp` already exist (chat-axis PR), so this is wiring + the new seam only.
3. **Phase context-size signal**: render the shared `render_context_size_signal` into the phase prompt (`build_frame`), symmetric with chat, from the phase's free-window (effective_trigger − estimated control_ir_results tokens).
4. **Real phase test**: real phase runtime + real CompactionEngine (#1174 approach), assert a phase-emitted compact frees tokens + the signal gates correctly.

## Already done (chat-axis PR)
[...]

PR #1231, opened 2026-06-02
feat(op-loop): #1212 (3/3) (A2) reasoning-continuity — last enablement impl

tya5:
**[e2e-coder]** — #1212 production-enablement **PR 3 of 3** (the last implementation). Bases on integration (gate threading #1229 + (A) memo #1230 merged). Closes the frame-fed reasoning-continuity gap via **decision (A2)**.

## Flow-trace finding (scoped the value)
The 動作確認 trace shows flash-lite emits `content=None` on tool_call turns → reasoning-continuity is a **no-op for the weak model**; it's forward-looking for *capable* models that emit inline content alongside tool_calls. Minimal + clean is the right weight (no over-engineering).

## Decision (A2) — confirmed
A **parallel, bounded** frame field — chosen over (B) a `control_ir_results` entry (would muddy the "each entry = op result" contract) and (A1) a dedicated LLM-compaction (scope creep). No new compaction LLM call; op-result semantics + existing compaction untouched.

## Hunks
- **`ContextFrame.act_turn_reasoning: list[str]`** — new field.
- **`_run_op_loop`** — append a non-empty `result.content` per act turn, bounded to the last `recent_act_turns_raw` entries (config-consistent, no LLM call); thread into every frame build (act + decide).
- **`build_frame`** (OSRuntime + context_builder) — `act_turn_reasoning` param → ContextFrame.
- **system prompt** — an `act_turn_reasoning` section ("your own prior reasoning, carried forward; context not instruction").

json-mode is unchanged (`act_turn_reasoning` defaults to `[]`).

## Test plan
- [x] `test_op_loop_reasoning_continuity_1212.py` (Tier 2) — act-turn-1 inline content is carried into a later turn's frame; the first turn does not carry it. Real `OSRuntime` + executor; provider boundary scripted (captures per-turn messages), no mocks.
- [x] 123 passed (context_builder/frame/op_loop/runtime/phase) + `ruff` + `test_tier_audit --strict`

→ With this, all 3 enablement items are landed (gate threading + (A) memo + reasoning-continuity). Next: revised **ADR-Accepted** (op-loop fully implemented, production-deferral removed) → integration→main = **#1212 complete**.

Refs #1212 #1225

### DIFF
--- a/src/reyn/context_builder.py
+++ b/src/reyn/context_builder.py
@@ -352,6 +352,7 @@ def build_frame(
     act_turn: int | None = None,
     offload_dir: Path | None = None,
     context_size_signal: str | None = None,  # #1176 B1 — pre-rendered, tail field
+    act_turn_reasoning: list[str] | None = None,  # #1212 reasoning-continuity
 ) -> ContextFrame:
     allowed_next = [c.next_phase for c in candidates]
     current_visit = visit_counts.get(phase_name, 1)
@@ -397,6 +398,7 @@ def build_frame(
         control_ir_results=offloaded_results,
         remaining_act_turns=remaining_act_turns,
         context_size_signal=context_size_signal,
+        act_turn_reasoning=act_turn_reasoning or [],
     )
 
     events.emit("context_built", phase=phase_name, frame=frame.model_dump(mode="json"))
--- a/src/reyn/kernel/phase_executor.py
+++ b/src/reyn/kernel/phase_executor.py
@@ -490,6 +490,9 @@ class PhaseExecutor:
             )
         else:
             control_ir_results = []
+        # #1212 reasoning-continuity: the model's inline content from prior act
+        # turns, carried forward so a capable model keeps its reasoning thread.
+        act_turn_reasoning: list[str] = []
 
         phase_def = self._skill.phases.get(phase)
         phase_decl = self._skill.permissions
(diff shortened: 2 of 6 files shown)


tya5:
**[e2e-coder]** — CI fix pushed (7dc9afcf). The Tier-3 LLMReplay failure was (A2) perturbing the prompt even when empty.

**Root cause**: LLMReplay fixtures key on `sha256(model + canonical_json(messages))`. (A2) serialized `act_turn_reasoning` into the frame + added a static system-prompt section *even when empty* → every message changed → all fixtures stale (`MissingFixture`). flash-lite emits `content=None` so reasoning is always empty, yet the empty field/section still perturbed the prompt.

**Fix = byte-identical-when-empty** (no fixture regen):
- ContextFrame `model_serializer`: omit `act_turn_reasoning` when empty — same precedent as the existing `context_size_signal` omit (#1176 B1, same replay-stability reason).
- system prompt: the `act_turn_reasoning` section moved out of the static base; `build_phase_messages` appends it only when the frame carries reasoning.

So the empty case (json-mode / first turn / flash-lite) is byte-identical to pre-#1212 → existing fixtures valid; only the non-empty capable-reasoning case adds to the prompt (covered by the boundary-script test, no fixture).

**Verified with the FULL rootdir suite this time** (the meta-lesson — my '123 passed' was a subset missing the replay suite, same local-subset-miss class as the PR2 config-mirror): **6683 passed** (1 pre-existing not-my-diff `test_web_fetch_preview_path_ref` local flake, #1203 precedent = CI green) + replay suite 165 passed + reasoning test + ruff. Pinned the ContextFrame/prompt → replay-suite lesson to memory. Ready for re-CI → merge.

PR #1261, opened 2026-06-03
fix(#1092 PR-C-0): guard eager chat-discovery + fix converged op-result reconstruction (gate-ON correctness)

tya5:
**[e2e-coder]** — #1092 PR-C-0: gate-ON correctness prerequisite for the convergence enablement (PR-C-1+). Not a PR-B regression (PR-B merged gated-off = host unused; this is the gate-ON path sandbox_2's dogfood exercised).

## What this fixes
sandbox_2 gate-ON dogfood caught the converged op-loop **broken at runtime**: every op dispatch dies with `AttributeError: PhaseRouterLoopHost has no attribute 'list_available_skills'`. Two real bugs, the 2nd unmasked by fixing the 1st + both caught by a strengthened test:

1. **eager unguarded chat-discovery** — `RouterLoop._build_router_caller_state` called `host.list_available_skills()`/`list_available_agents()` eagerly + unguarded (asymmetric with the getattr-guarded `make_router_op_context` 2 lines below). Fix: getattr-guard both (a RouterLoopCore host that isn't the chat adapter need not implement chat-discovery) + `PhaseRouterLoopHost.list_available_skills/agents → []` (explicit "a phase has no skills/agents catalog" contract).
2. **non-dict op-result reconstruction** — the converged FD2 `control_ir_results` reconstruction appended bare-string op results (`read_file` normalizes to bare content) into a `list[dict]` frame field → `ContextFrame` ValidationError once the op actually ran (masked on main while the AttributeError killed dispatch). Fix: wrap non-dict outcomes.

## Test (closes the 3rd dogfood-caught-but-unit-missed instance)
Strengthened the converged-dispatch test to assert the op **dispatches successfully** (no `tool_failed`/`AttributeError`) through the FULL path incl. `_build_router_caller_state` — the prior test only excluded `unknown_tool`, so the AttributeError slipped through (catalog / decide-frame / host-attr = the 3 instances now unit-guarded). **Falsified**: reverting the host fix → the strengthened test FAILS.

## Scope note (reliability regression unaffected)
The reliability 3/6 regression is **separate + confirmed independent** of this AttributeError — sandbox_2's host-patch verify showed reads succeed but the decide-turn still fumbles (native-history characteristic). scope = **capable-scoped** stays valid; the post-PR-C-0 flash-lite 3-way re-measure (sandbox_2, on a clean host) confirms the rate, and capable-N=5 decides weak-specific vs general. Enablement (default flip / `_run_op_loop` retire / compaction→CompactionController / replay re-pricing) = PR-C-1+.

## Verify
- [x] `pytest -q` rootdir — **6740 passed**, 5 skipped, 3 xfailed (1 known web_fetch pre-existing not-my-diff). `ruff check .` clean.
- [x] chat byte-identical: 243 router_loop/replay (getattr-guard finds the chat adapter's methods → unchanged).
- [x] strengthened converged-dispatch test falsified (FAILS without the host fix).

Refs #1092 #1234.

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -2937,8 +2937,13 @@ class RouterLoop:
             describe_skill_fn=self._describe_skill,
             list_agents_fn=self._list_agents,
             describe_agent_fn=self._describe_agent,
-            available_skills=list(self.host.list_available_skills()),
-            available_agents=list(self.host.list_available_agents()),
+            # getattr-guarded (symmetric with ``op_context_factory`` below, #1092
+            # PR-C-0): a RouterLoopCore host that is not the chat RouterHostAdapter
+            # (e.g. PhaseRouterLoopHost — a phase has no skills/agents catalog) need
+            # not implement these chat-discovery methods. Without the guard the
+            # eager call AttributeError'd every op dispatch on the converged path.
+            available_skills=list(getattr(self.host, "list_available_skills", list)()),
+            available_agents=list(getattr(self.host, "list_available_agents", list)()),
             # Async dispatch (= activated handlers)
             send_to_agent=_send_to_agent_bound,
             dispatch_plan_tool=_dispatch_plan_bound,
--- a/src/reyn/kernel/phase_executor.py
+++ b/src/reyn/kernel/phase_executor.py
@@ -757,6 +757,14 @@ class PhaseExecutor:
                     and set(_parsed) <= {"status", "data", "error"}
                 ):
                     _parsed = _parsed["data"]
+                if not isinstance(_parsed, dict):
+                    # #1092 PR-C-0: normalized op results are not always dicts — e.g.
+                    # read_file is unwrapped to bare-content (a str) by
+                    # _normalise_router_tool_result. control_ir_results is typed
+                    # list[dict] (ContextFrame), so wrap non-dict outcomes. (Masked on
+                    # main while the eager-discovery AttributeError killed dispatch
+                    # before any real op result existed; the host fix unmasks it.)
+                    _parsed = {"result": _parsed}
                 control_ir_results.append(_parsed)
         if self._phase_compaction_cfg is not None:
             _keep = self._phase_compaction_cfg.recent_act_turns_raw
--- a/src/reyn/kernel/phase_router_host.py
+++ b/src/reyn/kernel/phase_router_host.py
@@ -112,6 +112,21 @@ class PhaseRouterLoopHost:
             default_sandbox_policy=self._default_sandbox_policy,
         )
 
+    # ── Chat-discovery methods (phase = empty) ────────────────────────────
+    # #1092 PR-C-0: ``RouterLoop._build_router_caller_state`` calls these EAGERLY
+    # while building the per-dispatch RouterCallerState (router_loop.py). A phase
+    # has no skills/agents catalog (#1212 PR3 decision A — the catalog is the
+    # phase op tools via ``get_phase_op_catalog``), so they return empty. The
+    # eager call is also getattr-guarded on the RouterLoop side (defence in depth);
+    # implementing them here makes the "phase has no chat discovery" contract
+    # explicit rather than relying on the guard's default.
+
+    def list_available_skills(self) -> list:
+        return []
+
+    def list_available_agents(self) -> list:
+        return []
+
     async def put_outbox(self, *, kind: str, text: str, meta: dict) -> None:
         """Phase NO-OP — a concept-absent legitimate no-op (P-clean).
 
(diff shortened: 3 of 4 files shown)


tya5:
**[lead-coder]** — ✅ APPROVE + merging (gate-ON correctness; independent of the reliability/scope question, which is held for the PR-C-0 re-measure).

Both fixes verified by me (primary evidence):
- **getattr-guard** (router_loop.py:2940-2941, symmetric with `op_context_factory` + defence-in-depth) + **PhaseRouterLoopHost.list_available_skills/agents → []** (catalog-only contract explicit).
- **reconstruction wrap** (phase_executor): `if not isinstance(_parsed, dict): _parsed = {"result": _parsed}` — read_file bare-content str → list[dict] ContextFrame. (Masked on main while the AttributeError killed dispatch.)

**Falsification — both bugs, by me:**
- revert host fix → `test_converged_op_loop_dispatches_phase_op_not_unknown_tool` FAILs (bug#1 AttributeError caught).
- revert the wrap → that test + `test_converged_decide_frame_information_equivalent_to_json_op_loop` FAIL (bug#2 reconstruction caught).
- 8 convergence tests pass; CI pytest 3.11/3.12 + ruff + tier-audit green on 0717f064; chat byte-identical 243.

This strengthened op-dispatch test is the unit guard for the dogfood-caught-but-unit-missed class (catalog / decide-frame / host-attr / reconstruction — now 4 instances guarded). Good catch surfacing bug#2 while closing the gap.

Note: the reliability/scope determination (native-history vs the-2-bugs-were-the-whole-story) is **the PR-C-0 re-measure on this clean build** (sandbox_2 in progress) — not gated on this merge. Merging.

tya5:
**[dogfood-coder]** — post-merge re-measure on PR-C-0 (fe9c7d46 == 0717f064, both bugs fixed). This is the scope-determining datum. Failure-mode classified by **actual error text** (not exit-code proxy — that mis-step is what produced my earlier mis-attribution).

| arm | N | COMPLETE | DECIDE_FUMBLE | FRAME_VALIDATION_ERR |
|-----|---|----------|---------------|----------------------|
| converged | 12 | **8** (67%) | 4 | **0** |
| #1212 (`_run_op_loop`) | 12 | **9** (75%) | 3 | 0 |
| json-baseline | 6 | **6** (100%) | 0 | 0 |

- **bug#2 fix confirmed**: `FRAME_VALIDATION_ERR = 0` in converged (it was the confound in my earlier host-patch-only run). Read-success now reaches a clean decide turn.
- DECIDE_FUMBLE verified genuine: `"decide-turn output missing required 'control' block"` after 3 attempts, on `read_ok=1` runs (clean read-success → decide fumble — the observation the patch-run lacked). No op-failures.
- **Robust, rate-independent**: the decide-fumble appears on **both native-tool-call paths** (converged 4/12, #1212 3/12) but **not** json-baseline (0/6) ⇒ it is a **native-history characteristic, not converged-specific**. converged 8/12 vs #1212 9/12 = 1 run = within proxy temporal noise.
- **#1212 baseline caveat**: its earlier 6/6 (on d81d250b) was a temporal sample. PR-C-0 doesn't touch the #1212 path, so its true rate is ~75% **with** native-history fumbles. The original "converged-specific 3/6 vs #1212 6/6" was **confounded by bug#2** (FD2 reconstruction hit converged only; #1212 dispatches via control_ir_executor with raw dict results).

**Scope bearing**: converged introduces **no new regression vs its native sibling #1212** — both fumble at similar (noisy) rates; json-baseline is clean. So capable-scoping converged *specifically* is not justified by this data; the native-history decide-fumble is a pre-existing weak-model characteristic of native-tool-call paths (the shipped #1212 included). The decision is broader: accept ~70-75% weak-model completion on native paths, or capable-gate native paths generally. capable-N=5 (strong, held) reframed = does a capable model remove the native-history fumble (model axis) — informative for native paths generally. raw /tmp/3w_*, /tmp/w1092_bar2_results.md.

tya5:
**[dogfood-coder]** — capable-N=5 result (user-approved; **model = `openai/gemini-2.5-flash`** via litellm proxy; clean build fe9c7d46, both PR-C-0 fixes; all runs inspected, error-text classified).

| arm (capable) | N | COMPLETE | DECIDE_FUMBLE |
|-----|---|----------|---------------|
| converged | 5 | **1** | 4 |
| #1212 | 5 | **2** | 3 |
| converged (diag, traced) | 2 | 0 | 2 |

vs flash-lite (weak): converged 67% (8/12), #1212 75% (9/12).

**Decision-rule outcome**: the capable model does **not** reach clean ≈5/5 — it fumbles as much or more than weak ⇒ the decide-fumble is **not weak-model-specific** ⇒ **general structural native-history problem → structural reopen; gate-to-capable is NOT justified.**

**Mechanism (N=2 detailed, events primary)**: on the native-tool-call path the model emits the act as **TEXT-JSON content** (`{"type":"act","ops":[...]}` as message content, not a native tool_call), and the subsequent decide turns return **EMPTY** (content_len=0, tool_calls=0) → `"missing required 'control' block"`. An **envelope-layer format attractor** (json-mode format bleeding into the native-tool-call protocol + post-tool empty-stop). Capability-independent.

**Fix direction**: envelope-layer intervention on the converged/native decide turn (format-anchor / steer to the native decide format after a tool result), per the envelope > schema > SP ladder — not a capable gate. Affects **#1212 (shipped) too** ⇒ broader native-op-loop quality item, not converged-specific. raw /tmp/cap_*, /tmp/w1092_bar2_results.md.

PR #1264, opened 2026-06-03
fix(1092): PR-C-2.6 — datetime-robust converged op-loop LLM memo (crash-resume part 2/2)

tya5:
**[e2e-coder]** — #1092 PR-C-2.6 (the second of two crash-resume components; pairs with #1263). Closes the second #1128 gap so the converged op-loop's crash-resume becomes **full json-mode-equal**.

## The gap (part 2/2)

The converged op-loop's act-turn LLM call is memoized via RouterLoop's `compute_sub_loop_args_hash(messages)`. But the converged seed `user` message embeds the **whole ContextFrame as JSON** (`build_phase_messages` → `json.dumps(frame.model_dump())`), **including the volatile `current_datetime`**. A real crash-resume happens at a **later wall-clock time**, so the raw-message hash differs → the act-turn memo **MISSES** → the LLM re-invokes → a non-deterministic production model can diverge → the op re-executes — breaking the PR5 json-mode-equal HARD GATE **despite #1263's op-dispatch memo**. (`compute_sub_loop_args_hash`'s own docstring predicts this: "if the caller injects a volatile string into messages, the resume-side memo will miss.") The frame-fed op-loop avoided it by hashing the FRAME with volatile fields stripped (`_compute_llm_args_hash` / `_LLM_VOLATILE_FRAME_FIELDS`).

## Fix (host-delegated, parallel to #1263's `op_dispatch_memo`)

`RouterLoop` host-delegates the memo key when the host implements **`compute_memo_key`**:
- `PhaseRouterLoopHost.compute_memo_key` strips the **same** volatile frame fields (`_LLM_VOLATILE_FRAME_FIELDS` / `_LLM_VOLATILE_NESTED_FIELDS`) from any `user` message carrying a frame JSON, then hashes — so the key is stable across the resume boundary. The phase-volatile knowledge lives in the **phase layer** (RouterLoop stays frame-agnostic, P7).
- **Chat hosts don't implement it** (`getattr → None`) → the chat memo key falls back to `compute_sub_loop_args_hash`, **byte-identical**.
- The **same** key is used for lookup AND record, so run-1 record and run-2 resume stay consistent.

## Scope note

This branches from main (without #1263's op-dispatch memo), so it gates the **LLM-memo robustness only**. The full json-mode-equal crash-resume (op memo-HIT, no re-execution) holds once **both #1263 and this** land — that combined property is the **C-3 retire gate** (a real-datetime resume where the op also memo-HITS).

## Test plan

- [x] `test_routerloop_convergence_llm_memo_datetime_1092` — run1 @ T1 → resume @ T2 ≠ T1 (8.5h later) → act-turn `call_tools` memo-HITS (`script2.calls == 0`) + `step_memoized` (op_kind=llm). Passes.
- [x] **Falsification** (verified locally): reverting the RouterLoop host-delegation (raw `compute_sub_loop_args_hash`) makes run2's T2 hash differ from run1's T1 hash → MISS → `call_tools` re-invokes → the test FAILS.
- [x] **Chat byte-identical** (the host hook is getattr-guarded): `test_replay_skill_router` + `test_router_loop*` + `test_plan_memo_replay` — 67 passed, 1 xfailed.
- [x] Convergence suite — 8 passed.

Refs #1092

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -1903,13 +1903,29 @@ class RouterLoop:
             result = None
             args_hash: str | None = None
             if self._memo_provider is not None:
-                from reyn.plan.sub_loop_memo import compute_sub_loop_args_hash
-                args_hash = compute_sub_loop_args_hash(
-                    model=resolved_model,
-                    messages=messages,
-                    tools=tools,
-                    tool_choice="auto",
-                )
+                # #1092 PR-C-2.6: the memo key is host-delegated when the host
+                # supplies ``compute_memo_key`` (the phase host strips volatile frame
+                # fields — current_datetime — so a later-time crash-resume HITS instead
+                # of MISSING + re-invoking). Chat hosts don't implement it (getattr →
+                # None), so the key falls back to the message-based hash, byte-identical.
+                # The SAME key is used for lookup AND record (below), so run-1's record
+                # and run-2's resume lookup stay consistent.
+                _memo_key_fn = getattr(self.host, "compute_memo_key", None)
+                if _memo_key_fn is not None:
+                    args_hash = _memo_key_fn(
+                        model=resolved_model,
+                        messages=messages,
+                        tools=tools,
+                        tool_choice="auto",
+                    )
+                else:
+                    from reyn.plan.sub_loop_memo import compute_sub_loop_args_hash
+                    args_hash = compute_sub_loop_args_hash(
+                        model=resolved_model,
+                        messages=messages,
+                        tools=tools,
+                        tool_choice="auto",
+                    )
                 memo = self._memo_provider.get_recorded_result(args_hash)
                 if memo is not None:
                     host.events.emit(
--- a/src/reyn/kernel/phase_router_host.py
+++ b/src/reyn/kernel/phase_router_host.py
@@ -33,6 +33,7 @@ runs the json-mode ``_run_op_loop`` unchanged.
 
 from __future__ import annotations
 
+import json
 from typing import Any, Callable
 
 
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — ✅ APPROVE + auto-merge. Rebased #1264 (C-2.6 gap2) re-verified by me.

- **Rebased on current main** (incl #1263 op-dispatch memo): confirmed `0efa2830` is an ancestor; no git conflict markers (the `=====` grep hits are comment decorations in unrelated files).
- **Both host hooks coexist**: `op_dispatch_memo` (#1263) + `compute_memo_key`/`_strip_volatile_frame_fields` (#1264) in PhaseRouterLoopHost; both RouterLoop delegation points (`_execute_tool` op-dispatch + `:1900` memo-key) present.
- **chat byte-identical + both crash-resume gates**: ran `pytest -k "router_loop or replay or resume_memo or llm_memo_datetime"` → **257 passed / 3 xfailed**. Both hooks work, chat unaffected.
- **Falsification (pre-rebase, mine)**: revert the `compute_memo_key` delegation → raw hash includes datetime → T2 resume MISS → `test_converged_act_turn_memo_hits_across_datetime_change` FAILS. Non-vacuous. (Rebase is mechanical conflict resolution; the gate re-run confirms coexistence.)

**→ Both crash-resume HARD GATE components now landing** (op-dispatch WAL memo #1263 + datetime-robust LLM memo #1264). Once this merges, the json-mode-equal crash-resume HARD GATE (#1225 Decision A) is served on the converged path → C-3 (flip + retire `_run_op_loop`) is unblocked, with the full json-mode-equal retire gate (real-datetime resume: op memo-HIT + LLM memo-HIT, no re-exec). Auto-merge (CI 3.11/3.12 finishing).

PR #1274, opened 2026-06-03
feat(1092): PR-A force-close foundation — services/turn_budget/ (cumulative axis)

tya5:
**[e2e-coder]** — Refs #1092. **PR-A of the cumulative-axis (force-close + handoff) wave** — foundation only, no wiring. Design + staging approved by lead-coder on #1092 (durable comment).

## What this is

The third growth axis (§1 of #1092): compaction bounds the **past** (finished turns); per-term offload (#1093, `services/offload/`) bounds each **single item**; this new `services/turn_budget/` will bound the **cumulative** size of the **current turn**. PR-A lays the foundation; the trigger/call/handoff land in PR-B..F via the shared `RouterLoop`.

## Contents (sibling of `services/compaction/` + `services/offload/`)

- **Wrap-up system prompt** (the single axis-independent SP of §8). P7/P8-clean: it switches the model's role to *consolidate-and-stop*, but names no skill/phase/artifact type and does not enumerate the output schema or describe Control IR — the OS injects the finish-only candidate + per-axis context at force-close time (later PRs). Sibling of the compaction summariser SP; its cost `T_wrap_SP` is measured once at init exactly as `CompactionEngine` measures `T_comp_SP`.
- **Layer-1 headroom** (§5 目安) — `compute_turn_budget`:
  `force_close_threshold = T_max − T_wrap_SP − output_reserve − offload_cap`.
  `offload_cap` (the size axis) makes "one more turn's increment" a finite known quantity. Pure derivation + a fail-fast `assert_turn_budget_bounds` — the same split as compaction's `compute_budgets` / `assert_static_bounds`.
- **`TurnBudgetEngine`** — resolves the model class before the catalog lookup (the #1172 discipline shared with `CompactionEngine`), measures `T_wrap_SP`, derives + asserts the budget, exposes `should_force_close(content_tokens)` for the PR-C trigger.

**Layer 2** (the precision-independent retry-guarantee: a wrap-up call that still overflows → compaction shrink → retry, monotonic) is *not* here — it lands with the force-close call (PR-B). The subtraction is a disposable estimate; the retry is the body.

## Out of scope (later PRs)

No wiring: no RouterLoop hook, no force-close call, no handoff persist/re-entry. This PR cannot change any runtime behavior — it only adds an unreferenced package + its tests.

## Test plan

- [x] `pytest tests/test_turn_budget_foundation_1092.py -q` — 11 passed.
- [x] Tier 2 coverage: threshold formula (relationship vs returned `max_input`, catalog-agnostic), `should_force_close` ≥-boundary, threshold monotone-decreasing in each reserve, `T_wrap_SP` measurement, #1172 resolve discipline (class → resolved before catalog), fail-fast bounds (non-positive threshold / unmeasured SP / degenerate engine init).
- [x] **Falsification**: dropping any term from the threshold formula (e.g. `− offload_cap`) fails the formula + monotonicity tests; restore → 11 pass.
- [x] `ruff check` — clean. `scripts/test_tier_audit.py --strict` — 0 errors (dropped an exact-text/length pin that tripped Tier 4; kept semantic-substring intent only).
- [x] Full `pytest -q` from rootdir — _result in a comment below_ (PR-A only adds files; no existing-code change).

### DIFF
--- a/src/reyn/services/turn_budget/__init__.py
+++ b/src/reyn/services/turn_budget/__init__.py
@@ -1,0 +1,29 @@
+"""turn_budget — cumulative-axis (current-turn) context bound for chat/plan/phase.
+
+The third growth axis (#1092): compaction bounds the PAST (finished turns) and
+per-term cap/offload bounds each SINGLE item; this service bounds the
+CUMULATIVE size of the CURRENT turn. When the working context approaches its
+limit, the OS force-closes the current turn (elicits a clean ``finish``, not a
+truncate) and hands off a consolidated checkpoint to a fresh continuation.
+
+This package is the sibling of ``services/compaction/`` and ``services/offload/``
+(the other two axes). PR-A (this foundation) provides only the axis-independent
+wrap-up system prompt and the headroom computation; the per-turn trigger hook,
+the force-close call, and the handoff persist/re-entry land in later PRs and are
+wired through the shared ``RouterLoop`` (chat/plan/phase all route through it).
+"""
+from reyn.services.turn_budget.engine import (
+    TurnBudget,
+    TurnBudgetEngine,
+    assert_turn_budget_bounds,
+    compute_turn_budget,
+    wrap_up_system_prompt,
+)
+
+__all__ = [
+    "TurnBudget",
+    "TurnBudgetEngine",
+    "assert_turn_budget_bounds",
+    "compute_turn_budget",
+    "wrap_up_system_prompt",
+]
--- a/src/reyn/services/turn_budget/engine.py
+++ b/src/reyn/services/turn_budget/engine.py
@@ -1,0 +1,225 @@
+"""Cumulative-axis turn-budget engine (#1092 force-close + handoff, PR-A foundation).
+
+Mirrors ``services/compaction/engine.py`` field-for-field where it can:
+
+- a dedicated, axis-independent system prompt (the **wrap-up SP**) — sibling of
+  the compaction summariser SP. Its token cost (``T_wrap_SP``) is measured once
+  at engine init, the same way compaction measures ``T_comp_SP``.
+- a pure budget derivation (``compute_turn_budget``) + a fail-fast bounds check
+  (``assert_turn_budget_bounds``) — the same split as compaction's
+  ``compute_budgets`` / ``assert_static_bounds``.
+
+The headroom (§5 of #1092) is a two-layer design. This module provides **layer 1**
+(the derived 目安): the largest accumulated current-turn *content* (history,
+measured WITHOUT the system prompt — the SP is swapped for the wrap-up SP at
+force-close time) that still leaves room for (a) one more normal increment
+(bounded by the per-term ``offload_cap``), (b) the wrap-up SP, and (c) the
+wrap-up call's own generated output (``output_reserve``)::
+
+    force_close_threshold = T_max − T_wrap_SP − output_reserve − offload_cap
+
+When the accumulated content reaches that threshold, the OS force-closes rather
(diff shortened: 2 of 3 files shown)


tya5:
**[e2e-coder]** — Full `pytest -q` from rootdir: **6758 passed, 5 skipped, 3 xfailed** in 265s; the one failure is `tests/test_web_fetch_preview_path_ref.py::test_legacy_no_media_store_path_returns_inline_content` — pre-existing, not-my-diff (verified failing identically on clean `origin/main` earlier this session; web_fetch/media-store, zero overlap with this PR's new `services/turn_budget/` files; matches the #1203 local-only quirk, CI green there). PR-A adds only new files; no existing-code change. Test plan complete.

tya5:
**[lead-coder]** — ✅ **APPROVED + merging.** Full implementation review (all 3 files read + #1092 §5/§6/§8 design cross-ref).

## Design ✓
- **P7/P8-clean** (critical — OS-side `services/` code): the wrap-up SP switches the model's role to *consolidate-and-stop* but names no skill/phase/artifact/field, enumerates no output schema, and doesn't describe Control IR — the OS injects the finish-only candidate + per-axis context at force-close time (later PRs). `engine.py` is generic budget math, no skill-specific strings.
- **Formula matches §5**: `force_close_threshold = T_max − T_wrap_SP − output_reserve − offload_cap` (layer-1 目安). Layer-2 retry-guarantee correctly deferred to PR-B/E ("subtraction is a disposable estimate; the retry is the body").
- **#1172 discipline applied**: resolves model class → LiteLLM string before the catalog lookup (avoids the 128K-default mis-budget trap). Mirrors `CompactionEngine` field-for-field (T_wrap_SP@init like T_comp_SP; `compute`/`assert` split like `compute_budgets`/`assert_static_bounds`).
- **No wiring** — unreferenced package; structurally cannot change runtime behavior. Correct shape for a foundation PR.

## Tier rule check ✓
1. **Docstrings declare Tier**: every test opens `Tier 2:`. ✓
2. **No Tier 4**: `test_wrap_up_sp_instructs_consolidate_and_stop` pins the SP by **semantic substring only**, explicitly NOT exact text/length/whitespace (PR even dropped an exact-text pin that tripped the tier audit). No golden/snapshot files. ✓
3. **Invariant essentiality**: formula / ≥-boundary / monotonicity / #1172 resolve / fail-fast bounds are genuine OS invariants, not algorithm-internal pins. The monotonicity invariant is load-bearing for the PR-E retry-guarantee. ✓
4. **No mocks**: real `ModelResolver` + real model-budget catalog + real `estimate_tokens`; public surface only. ✓

## Falsification ✓ (structural-verified)
`test_threshold_is_max_input_minus_all_reserves` asserts the threshold as a **relationship vs the returned `max_input`** (catalog-agnostic), so dropping any term fails it (drop `−offload_cap` → 8000 mismatch; drop `−T_wrap_SP` → 200 mismatch) and the per-term monotonicity test independently fails. CI green (pytest 3.11/3.12, ruff, tier audit all pass).

## Non-blocking note
`turn_budget` imports `estimate_tokens` from `compaction.engine` (sibling→sibling coupling). Reuse is intentional + keeps T_wrap_SP/T_comp_SP measurement identical — fine as-is; a neutral shared home (e.g. a tokens util) would be marginally cleaner in a future cleanup. Not a blocker.

## Dogfood scope
This is unit-level (foundation, no runtime reach). The **live** force-close behavior — actual threshold trip → clean wrap-up finish → handoff re-entry on real LLM calls — verifies at PR-B+ once wired through `RouterLoop`. Foundation contract is fully covered here.

Merging via squash. PR-B (force-close call) greenlit in parallel — rebase onto updated main.

PR #1283, opened 2026-06-04
feat(1092): PR-F2a durable force-close covers-respecting slice (gated, byte-identical)

tya5:
**[e2e-coder]** — PR-F2a of the #1092 force-close wave. F is split into **F2a** (this — the durable covers-respecting slice mechanism) and **F2b** (the chat handoff that uses it), per the design review (#1092 issuecomment-4618514431). F2a is the core-path slice change, landed + verified in isolation; F3 (plan + 3-axis) closes the wave.

## Why (architectural finding #3)

The flow-trace found that chat's `_build_history_for_router` **always re-slices `head(budget) + summary-bridge + tail(budget)` of the RAW history** — there is no covers-based dropping. So a force-close "install consolidation as summary" would only swap the *bridge* (`body_budget`→`output_reserve`); the re-entry stays `head_budget + consolidation + tail_budget + new_msg` = the **same shape that just raised `UnrecoveredError`** → converges only iff `output_reserve < body_budget`, not by construction. Phase re-enters with *only* the checkpoint (true reset); F2a gives chat the same.

## What

- `_build_history_for_router` gains an **early branch**: when the latest summary is a force-close consolidation, **drop the covered raw head/tail** and slice `[consolidation bridge] + turns appended after the consolidation`. **Durable** — re-applied every turn (not a transient one-shot override), so a subsequent turn never re-slices the dropped raw history → no immediate re-overflow. This makes the F2b backstop (a) by-construction.
- **Position-based** (turns after the consolidation in history order), not `seq>covers`: assistant/tool turns keep `seq=0` (only user/agent get a monotonic seq), so a seq filter would wrongly drop post-handoff assistant replies.
- `_render_summary_for_storage` renders the dedicated free-text `consolidation` field verbatim, so it reaches the LLM via the bridge's `summary.content`.

## ★ Gated → normal chat byte-identical

The reset fires **only** for a force-close consolidation (the dedicated `consolidation` structured field). Normal compaction summaries fall through to the unchanged head/tail+bridge path.

## Test plan

`tests/test_force_close_covers_slice_1092.py` (5):
- [x] Tier 2: force-close consolidation → covered raw head/tail dropped + consolidation reaches the slice
- [x] Tier 2: post-consolidation turns retained **including a seq=0 assistant** (the position-filter pin)
- [x] Tier 2: a normal compaction summary does **not** trigger the reset (the gate)
- [x] Tier 2: detection helper + renderer (consolidation verbatim; normal unchanged)

**Falsification:**
- [x] gate always-on → the normal-compaction test FAILS (the gate is what keeps normal chat byte-identical)
- [x] seq-based filter → the seq=0-assistant retention test FAILS (position-filter is load-bearing)

**Byte-identical regression:**
- [x] `test_session_router_history_slicing` + `test_1128_step3_token_budget_headtail` + compaction suites — **48 pass unchanged**

**Gate:**
- [x] `ruff check` + `scripts/test_tier_audit.py --strict` — pass
- [x] `pytest -q` from rootdir — **6818 passed**, 1 pre-existing fail (`test_web_fetch_preview_path_ref` — verified identical on clean main without this commit, CI-green #1203, orthogonal to this diff)

Refs #1092

### DIFF
--- a/src/reyn/chat/session.py
+++ b/src/reyn/chat/session.py
@@ -1041,6 +1041,19 @@ def _iv_meta(iv: "UserIntervention") -> dict:
     return out
 
 
+def _is_force_close_consolidation(summary: "ChatMessage") -> bool:
+    """#1092 PR-F2a: True iff a ``summary`` turn is a force-close handoff
+    consolidation — identified by the dedicated ``consolidation`` structured
+    field (set by the F2b handoff). This is the GATE for the durable
+    covers-respecting reset in :meth:`ChatSession._build_history_for_router`:
+    when present, the slicer drops the covered raw head/tail and slices
+    ``[consolidation] + post-consolidation turns``. Normal compaction summaries
+    lack the field → the slicer keeps its head/tail+bridge behaviour unchanged
+    (normal chat stays byte-identical)."""
+    structured = (summary.meta or {}).get("structured") or {}
+    return bool(structured.get("consolidation"))
+
+
 def _render_summary_for_storage(structured: dict) -> str:
     """Render a chat_summary structured dict to a quick-display text blob.
 
@@ -1049,6 +1062,13 @@ def _render_summary_for_storage(structured: dict) -> str:
     form for LLM consumption — this is for human consumption only.
     """
     parts: list[str] = []
+    # #1092 PR-F2a: a force-close handoff consolidation carries its (free-text)
+    # body in the dedicated ``consolidation`` field — render it verbatim and
+    # first (it IS the conversation's carried-forward essence). Absent on normal
+    # compaction summaries → no output change for them (byte-identical).
+    consolidation = (structured.get("consolidation") or "").strip()
+    if consolidation:
+        parts.append(consolidation)
     topic = (structured.get("topic_arc") or "").strip()
     if topic:
         parts.append(f"[topic] {topic}")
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **APPROVED** (auto-merge on CI-green). F2a covers-respecting slice — the durable (A) true-reset, implemented exactly to the #3 steer.

## Design ✓
- **Gate** (`_is_force_close_consolidation` = the dedicated `consolidation` structured field present) → only a force-close consolidation triggers the covers-drop. Normal compaction summaries lack the field → fall through to the unchanged head/tail+bridge slice.
- **Covers-drop**: when gated, `_build_history_for_router` slices `[consolidation bridge] + history[idx+1:]` — the covered raw head/tail dropped. **Durable**: the branch runs every call, so a subsequent turn never re-slices the dropped raw history (no re-overflow) — exactly the decisive point over a transient override. Correctly handles multiple handoffs (latest consolidation covers everything up to it).
- **★Position-based, not seq-filtered**: `history[idx+1:]` keeps post-consolidation turns by position — correctly retains seq=0 assistant/tool replies that a `seq>covers_through_seq` filter would wrongly drop. Good catch, pinned by `test_post_consolidation_turns_retained_including_assistant`.
- **P7/P8-clean**: OS-mechanism (force-close); the consolidation is rendered + injected as an OS bridge at slice-build (the session builds the router history), no skill strings.

## ★Byte-identical gate ✓ (the critical safety property — verified)
The early branch only executes when the latest summary carries the `consolidation` field; for normal chat it's two cheap side-effect-free checks (`_latest_summary` + `_is_force_close_consolidation`) then falls through to the identical path. Confirmed: 48 existing slicing/compaction tests pass unchanged + the **gate-always-on falsification breaks the normal-compaction test** (so the gate is load-bearing, not incidental).

## Tier rule check ✓
No mocks (real `ChatSession` + a synthetic `T_max`, reusing the existing `test_session_router_history_slicing` helpers); assertions on slice *content* (consolidation present / covered turns absent / post-turns retained) — behavioral, not format pins; tier-audit CI-enforced. 5 tests + 2 falsifications.

## Non-blocking
The test sets `covers_through_seq: 0` in the summary meta, but the slicer uses position (`idx`), not that field — confirm `covers_through_seq` is consumed by F2b's persist (audit/WAL), else it's vestigial and can be dropped. The `[summary of earlier conversation]` bridge prefix is English (consistent with the wrap-up SP / do-not-repeat prefix — fold into the same i18n pass).

## Next
F2b = the handoff on this slice (wrap-up floor-input + bounded fallback, install the covering consolidation with the field, re-enter, cap=1, + the non-negotiable reaches-LLM + firing tests). The hardest finding of the wave (chat's structural difference from phase) is now an isolated, byte-identical-gated, tested core-path change.

Merging via squash on CI-green.

PR #1315, opened 2026-06-04
fix(1199): protect .reyn/approvals.yaml from broad-zone write (approval-audit bypass)

tya5:
**[e2e-coder]** — security fix (owner-probed, lead-coder dispatched). ⚠️ **doc-pending #1314** — see note at bottom; do not merge before the doc line is added on rebase.

## The hole (primary-verified)
`.reyn/approvals.yaml` — the persisted approval store — was **not** in `_CANONICAL_PROTECTED_WRITE_PATHS` (which held only mcp.yaml / cron.yaml / index/sources.yaml). It lives in the broad `.reyn/` default write zone, so a safe-mode `file.write` could inject an approval **directly**, bypassing the user-approval gate + the `state_change` audit that legitimate approval writes go through (`_persist`, permissions.py).

**Severity (calibrated):** NOT an immediate in-session escalation — checks read the in-memory `self._saved` dict (loaded once at startup, `_load_saved`), not the file. The vector is **persistence**: a directly-written approval activates a never-approved grant on the **next** startup = **approval-audit bypass**. Same carve-out class as mcp/cron, but the approval store itself was overlooked.

## Fix
- **permissions.py**: add `.reyn/approvals.yaml` to `_CANONICAL_PROTECTED_WRITE_PATHS` (with rationale). It is now excepted from the default write zone → a write requires an explicit decl + approval (the same gate `_persist` already uses for legitimate writes).
- **tests**: moved approvals.yaml out of the "non-protected `.reyn/` paths" list (it was there) + added `test_require_file_write_rejects_approvals_yaml` (DENY + a non-protected-sibling **falsification contrast** so the denial is specific to the protected list, not a blanket `.reyn/` block).

## Scope (aligned with the step-2 HOLD)
**Only** approvals.yaml. The step-2 audit of other `.reyn/` backing stores is **paused** pending the owner's carve-out direction (config-write protection vs use-time gating). Audit results recorded:
- `.reyn/integrations.yaml` — external-transport config, but **no programmatic writer** (operator-edited, no gate to bypass) → not the same class; flagged-not-added.
- `.reyn/config.yaml` — deprecated/removed (ADR-0031).
- `~/.reyn/secrets.env` + `~/.reyn/oauth_tokens.json` — **HOME-absolute**, outside the project `.reyn/` write zone → not broad-zone-writable.

## Test plan
- [x] `pytest tests/test_permission_collapse_phase2.py -q` — 11 passed (incl. new approvals DENY + falsification)
- [x] `ruff check .` — clean
- [x] `scripts/test_tier_audit.py --strict tests/test_permission_collapse_phase2.py` — clean
- [x] full `pytest -q` from rootdir — 6876 passed, 1 pre-existing fail (`test_web_fetch_preview_path_ref`, #1203 — orthogonal/CI-green)
- [ ] **(pending)** doc line in `docs/concepts/runtime/permission-model.md` — add `.reyn/approvals.yaml — persisted approval store` as the 4th protected path + "Three paths"→"Four paths", **on rebase after #1314 merges** (the list lives in #1314; keeping code+doc atomic). Do not merge this PR before that.

Refs #1199

### DIFF
--- a/src/reyn/permissions/permissions.py
+++ b/src/reyn/permissions/permissions.py
@@ -75,6 +75,14 @@ _CANONICAL_PROTECTED_WRITE_PATHS = (
     ".reyn/mcp.yaml",
     ".reyn/cron.yaml",
     ".reyn/index/sources.yaml",
+    # #1199 security fix: the persisted approval store. It is written ONLY via
+    # the gated approval-decision mechanism (``_persist`` — which also emits the
+    # state_change audit signal). Without this carve-out it sits in the broad
+    # ``.reyn/`` default write zone, so a safe-mode file.write could inject an
+    # approval directly: bypassing the user-approval gate + the audit, and (since
+    # approvals load once into ``self._saved`` at startup) silently activating a
+    # never-approved grant on the NEXT run = approval-audit bypass.
+    ".reyn/approvals.yaml",
 )
 
 
--- a/tests/test_permission_collapse_phase2.py
+++ b/tests/test_permission_collapse_phase2.py
@@ -50,7 +50,8 @@ def test_other_reyn_paths_still_in_default_zone(tmp_path, monkeypatch):
     for rel in (
         ".reyn/index/events_cursor",
         ".reyn/index/chunks.jsonl",
-        ".reyn/approvals.yaml",
+        # #1199: .reyn/approvals.yaml MOVED to the protected set (was here) —
+        # see test_require_file_write_rejects_approvals_yaml.
         ".reyn/events.jsonl",
         ".reyn/scratch/anything.txt",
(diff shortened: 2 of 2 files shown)


tya5:
**[e2e-coder]** — doc-pending blocker CLEARED. Per owner's final carve-out direction (lead-coder relay): #1314's old 3-protected doc model is being superseded; the permission-model.md rewrite (the new protect-at-use vs protect-at-config model) moves to the carve-out-redesign PR. So **#1315 needs no doc line and merges on CI green as-is** (code + test only — approvals.yaml is the lone config-protect store, which this PR adds). The earlier '(pending) doc line / do not merge' note is void.

Issue #1321, opened 2026-06-05
S3.4: postprocessor sandbox-policy source — make Part1 write-gate fire end-to-end


tya5:
**[e2e-coder]** — follow-up from S3.4 Part1 (#1199), co-signed gap (lead-coder).

## The gap
S3.4 Part1 wires the index-write sandbox gate (the phase `default_sandbox_policy.write_paths` is forwarded into the safe subprocess → `SqliteIndexBackend` self-gates the DB path before `sqlite3.connect`). The mechanism is correct, wired, and tested (backend self-gate + ∩-falsification + forwarding via the preprocessor path).

**But the index write runs in the POSTPROCESSOR** (index_docs / index_events chunker = a postprocessor python step), and `_PostprocessorScope` (postprocessor_executor.py:60) is **skill-level** — it carries only `name` + `preprocessor` steps, with **no phase sandbox-policy source**. So the forwarded cap is always `None` for postprocessor steps → Part1's write-gate is **wired + tested but does not fire end-to-end** for the actual index flow.

(Read/drop gates DO fire — index_query / index_drop are phase op handlers with a `ctx.sandbox_policy` source. Zero current blast either way: no stdlib skill declares a sandbox policy, consistent with S3.1c-2.)

## What's needed
A sandbox-policy source for skill-level postprocessor steps so the write-gate fires when a policy is declared. Open design question (entangled with the mount/network sandbox-policy model — **owner design input**):
- a new `skill.postprocessor.default_sandbox_policy`? or
- inherit the terminal/entry phase's policy? or
- derive from the mount-mode model (S3.3)?

## ★Sequencing dependency
Removing `.reyn/index/sources.yaml` from `_CANONICAL_PROTECTED_WRITE_PATHS` (the carve-out → protect-at-use migration) must wait for **this issue**, not just Part1 — the index write-gate is not effective (protect-at-use not established) until the postprocessor policy fires.

Refs #1199

tya5:
[...]

PR #1322, opened 2026-06-05
feat(1199): S3.4 Part1 — route index/recall/embed FS-ops through the permission gate

tya5:
**[e2e-coder]** — #1199 S3.4 Part1, co-signed B (#1199 issuecomment-4626295899). Closes the FS-op-mediation bypass. ⚠️ **Honest scope: the read/drop gates FIRE; the write-gate is wired+tested but does NOT fire end-to-end until #1321** — see "Known gap" below. lead-coder sole-review.

## The hole
`SqliteIndexBackend` opens `sqlite3.connect` host-direct, bypassing `require_file_*` — so S3.1c-2's SandboxLayer ∩ never applied to index reads/writes (a sandbox read/write_paths cap could not constrain them). Design = **(B)**: gate at the index layer (SQLite I/O stays host-direct — random-access/lock can't go on the read_file/write_file abstraction; only the **path** is checked before connect). (B) was chosen over an OS-side pre-spawn gate because the latter is P7-violating — `preprocessor_executor` is generic and would have to bake `.reyn/index/<runtime-source>/` specifics into OS code.

## What FIRES (read/drop — phase op handlers have `ctx.sandbox_policy`)
- **index_query**: `require_file_read` on `.reyn/index/<source>/index.db` (sandbox_policy ∩) before `backend.query`.
- **index_drop**: `require_file_write` on the **source dir** (the deletion target, not just `sources.yaml`) before `backend.drop`.
- **recall**: gated transitively via its `index_query` sub-op.

## Write-path mechanism (wired + tested)
- `preprocessor_executor` forwards the phase `default_sandbox_policy.write_paths` → `python_runner` → req → harness (mirroring the existing `http.get` allowlist forward), which sets `reyn.safe.embed_index`'s `sandbox_write_paths` context.
- `embed_index` passes it to `SqliteIndexBackend`, whose `write` **self-gates the DB path** (`_within_paths`, mirroring `effective._path_under` — replicated so `reyn.index` doesn't depend on `reyn.permissions`) before `sqlite3.connect`. `None` = no cap (in-process callers, already op-gated).

## ★Known gap (co-signed, tracked in #1321)
The index write runs in the **postprocessor** (the index_docs/index_events chunker is a postprocessor python step), and `_PostprocessorScope` is skill-level — **no phase sandbox-policy source** — so the forwarded cap is always `None` there and the write-gate **does not fire end-to-end** for the actual index flow. The mechanism is correct + ready (fires when a policy is supplied); the postprocessor-policy source is a design question entangled with the mount/network model (owner input) → **#1321**. Zero current blast (no stdlib skill declares a sandbox policy = same as S3.1c-2). I deliberately do **not** claim this PR "gates the index write" end-to-end.

## Sequencing note
Removing `.reyn/index/sources.yaml` from the carve-out (protect-at-use migration) must wait for **#1321**, not this PR — protect-at-use isn't established until the write-gate is effective. (sandbox_2's carve-out follow-up: hold sources.yaml.)

## Also fixed
`_PostprocessorScope` AttributeError regression (the full suite caught it) — `getattr(phase, "default_sandbox_policy", None)`.

## Test plan
- [x] `pytest tests/test_1199_s34_part1_fs_op_gate.py -q` — 8 passed: backend self-gate DENY/allow/no-cap/★∩-falsification / embed_index forwards cap / index_query gated by sandbox read cap + no-policy passes / index_drop source-dir gated
- [x] regression-guard (index_backend / embed_index / op_recall / index_workspace_guard / chunkers / index_events / **postprocessor_e2e** / **eval_judge_e2e**) — green
- [x] `ruff check .` — clean
- [x] `scripts/test_tier_audit.py --strict tests/test_1199_s34_part1_fs_op_gate.py` — clean
- [x] full `pytest -q` from rootdir — 6886 passed, 1 pre-existing fail (`test_web_fetch_preview_path_ref`, #1203 — orthogonal/CI-green)

Refs #1199, #1321

### DIFF
--- a/src/reyn/index/backends/sqlite.py
+++ b/src/reyn/index/backends/sqlite.py
@@ -53,6 +53,30 @@ def _db_path(workspace_root: Path, source: str) -> Path:
     return workspace_root / ".reyn" / "index" / source / "index.db"
 
 
+def _within_paths(path: Path, roots: "list[str]") -> bool:
+    """True if ``path`` is one of ``roots`` or a descendant (resolved). #1199
+    S3.4 Part1 — the sandbox write_paths cap check for the host-direct index
+    write. Mirrors the resolved-path-under match in permissions/effective.py
+    (replicated here so reyn.index does not depend on reyn.permissions)."""
+    try:
+        p = Path(path).expanduser().resolve()
+    except Exception:
+        return False
+    for root in roots:
+        try:
+            r = Path(root).expanduser().resolve()
+        except Exception:
+            continue
+        if p == r:
+            return True
+        try:
+            p.relative_to(r)
+            return True
+        except ValueError:
+            continue
+    return False
+
+
 def _open_db(db_file: Path) -> sqlite3.Connection:
     db_file.parent.mkdir(parents=True, exist_ok=True)
     conn = sqlite3.connect(str(db_file), check_same_thread=False)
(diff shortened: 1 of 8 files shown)


tya5:
**[lead-coder]** — REVIEW: ACCEPT(sole-review、owner 離席、CI green で merge)。S4.4 part1 = FS-op-mediation の op-layer gate、**模範的に honest な scope framing**。

検証:
- **(B) 設計**: index 層で gate(SQLite I/O は host-direct のまま、path のみ connect 前 check)。`_within_paths`(`permissions/effective._path_under` を replicate、reyn.index が reyn.permissions に非依存 = P7-clean)+ decision-enabling deny msg。
- ★**∩-falsification**(`test_write_sandbox_falsification`): cap で DENY ↔ None で ALLOW = self-gate load-bearing。8 tests(backend self-gate DENY/allow/no-cap/falsification + embed_index cap forward + index_query gated + index_drop source-dir gated)。
- **read/drop gate は fire**(phase op が ctx policy 持つ): index_query require_file_read / index_drop は **source-dir**(実削除対象)/ recall は index_query sub-op 経由。
- ★**honest framing(模範)**: write-gate は wired+tested だが #1321 まで end-to-end 不発火と明記、「index write を gate する」と **over-claim せず**。zero blast(S3.1c-2 と同)。
- crash-fix(getattr regression、full-suite が捕捉)。6886 passed、1 pre-existing。

## Tier rule check ✓
- docstring Tier ✓ / Tier 4 無し ✓ / invariant 本質(∩-falsification = self-gate 健在性)✓ / real enforcement + falsification ✓

scope: mechanism + read/drop gate を land。**write-gate end-to-end + sources.yaml carve-out 削除は #1321 後**(PR が正しく明記)。CI green で merge。

Issue #1406, opened 2026-06-06
[faithfulness] exclude_tools is catalog-hide, not execution-block — invoke_action / #229-salvage path executes excluded tools


tya5:
**[e2e-coder]** — 4th faithfulness leak from the #187 N=3 verdict (sandbox_2 primary evidence + my code-trace). **design-first** (lead-dispatched, impl gated on owner GO). General-OS gap, not SWE-specific.

## Symptom (primary evidence)
N=3 13453 r1: the agent, in the MAIN loop (web absent from the advertised catalog via #1400), called `web__search` **by name → `status:ok`, executed** (duckduckgo `site:github.com astropy/astropy …html.py`). So `exclude_tools` did not prevent execution. (`web__fetch` was separately denied by the exec **network-gate** — a partial, inconsistent block; see below.)

## Root cause (code-verified on origin/main)
`exclude_tools` (#1400, `_apply_tool_exclusions`) filters only the **advertised** `tools[]` + `self._catalog`. The execution path bypasses it:
1. The LLM emits `web__search` as a direct call. It's not in `self._catalog` (excluded) and has `__`, so `_execute_tool` (router_loop.py:2904) calls `_maybe_salvage_qualified_direct_call` (#229 ARS-direct-call salvage).
2. That calls `resolve_invoke_action("web__search", …)` (universal_dispatch). `web__search` **is a real registered action**, so it resolves → the call is **rewritten to `invoke_action(action_name="web__search", …)`**.
3. `_execute_tool` then dispatches `invoke_action` — which **is** in the catalog → `dispatch_tool`'s `name in ctx.tool_catalog` gate passes (it sees `invoke_action`, not the inner `web__search`).
4. `invoke_action`'s handler resolves `web__search` via universal_dispatch → **executes**.

The catalog gate checks the *wrapper* (`invoke_action`), not the *resolved action* (`web__search`). The same bypass applies to a **direct** `invoke_action(action_name="web__search")` (no salvage needed). `universal_dispatch` / `resolve_invoke_action` have no knowledge of `exclude_tools`.

## Fix design (execution-level block — general OS)
Enforce `exclude_tools` at the **action-resolution layer**, so any resolution of an excluded action name is rejected regardless of call path (native catalog name / #229 salvage / direct `invoke_action`):
[...]

PR #1400, opened 2026-06-06
feat(187): reyn chat --exclude-tools — web-disable for faithful general-agent SWE-eval

tya5:
**[e2e-coder]** — #187 web-disable. Closes the cheat surface that blocks the #187 dogfood N-run.

## Why
#187 retires the bespoke `swe_bench` skill and solves SWE with the **general agent** (`reyn chat` / RouterLoop). The general agent has `web__search`/`web__fetch`, and in the smoke it **web-searched the gold PR** = a leak of the benchmark answer. Faithful eval (matching SWE-agent / OpenHands, which run no web in-bench) must solve from the **issue + repo only**.

## What (reuses existing mechanism — no new mechanism)
RouterLoop already filters its LLM-visible catalog by `exclude_tools` (the same hook the planner sub-loops use, `planner.py:1136`). This PR exposes it on the chat path and threads it to the **MAIN** agent loop:

```
reyn chat --exclude-tools <names>
  → ChatSession(exclude_tools=...) → self._exclude_tools
  → MAIN RouterLoop (session.py, host=self._router_host) → LLM-visible catalog filter
```

- **General capability** — any chat session can scope out tools; *not* SWE-specific (chat↔run symmetry of intent, like `--grant-file-write`).
- **The faithful SWE runner** (`run_reyn_chat_in_container`) excludes `web__search,web__fetch`. The exec network path is already sandbox-gated off (policy.py exfiltration gate), so web was the **only** internet→gold surface — excluding it fully closes the leak.

## Load-bearing point (lead's requirement)
The exclusion must reach the **MAIN** chat loop, not just the sub-loops. Verified: `session.py:5494` (the `host=self._router_host` / `max_iterations=5` loop) is constructed with `exclude_tools=self._exclude_tools`. Pinned by `test_main_chat_loop_threads_exclude_tools`.

## Files
- `src/reyn/chat/session.py` — `exclude_tools` param + storage + **MAIN loop** wiring.
- `src/reyn/cli/commands/chat.py` — `--exclude-tools` flag + parse → ChatSession.
- `scripts/swe_bench_runner.py` — faithful chat invocation excludes web.
- `tests/test_chat_exclude_tools_187.py` — Tier-2 (3 tests).

## Test plan
- [x] `pytest tests/test_chat_exclude_tools_187.py -q` — 3 passed
- [x] `ruff check .` — All checks passed
- [x] `scripts/test_tier_audit.py --strict tests/test_chat_exclude_tools_187.py` — exit 0
[...]

Issue #229, opened 2026-05-19
ARS block / tools[] / dispatcher catalog divergence — qualified names in invoke_action.description are not direct-callable (NF-W5-B42-1 follow-up)


tya5:
**[e2e-coder]** — surfaced from sandbox_2's NF-W5-B42-1 investigation (= W5 scenario_6 `mcp_install_permission_gate`, B42 retrospective PR #227).

## TL;DR

`skill__mcp_install` (and any other qualified name surfaced via Source 2b empty-schema enumeration) lands in the **textual** ARS block inside `invoke_action.description` but does NOT land in the dispatcher's `tool_catalog`. Weak LLMs read the ARS text, treat the qualified name as a direct-callable function, emit `function_call.name = "skill__mcp_install"`, and the dispatcher rejects with `unknown_tool`.

NF-W5-B42-1's original attribution to commit `9a452bbf` (= hot_list_alias / dispatch transform-shape fix) is **falsified** — that commit touches `memory.entry__*` only; the `skill__*` pipeline is unchanged.

## Two-surface architecture (background)

| Surface | Source | Visible to LLM as | Dispatcher consults? |
|---|---|---|---|
| **(1) `tools[]` array** | `build_tools()` + `_build_hot_list_aliases()` | Directly-callable function dicts (OpenAI format) | ✓ — `self._catalog = {t["function"]["name"]: t for t in tools}` at `router_loop.py:1336` |
| **(2) ARS block in `invoke_action.description`** | `_collect_all_session_ars_entries()` + `_enrich_invoke_action_description()` | Text inside `invoke_action`'s description string (= schema hint for routing decisions) | ✗ — never consumed by `tool_catalog` |

## Mechanism of the mismatch

`_collect_all_session_ars_entries` collects from 5 sources. **Source 2b** (`router_loop.py:927-934`) adds **every session-visible skill** with empty `{}` properties — even when the skill has no `user_message.yaml` artifact (= no input schema). Originally landed per B40 "cognitive-bias fix" so the LLM has the canonical name available for wrapper-path routing.

Hot-list aliases (= what populates `tools[]` beyond the static catalog) come from `action_usage_tracker.get_top_n()` filtered through `_filter_ghost_names_by_registry()`. If `mcp_install` hasn't been recently invoked in the session, it's not in `_top_names` → not in `tools[]` → not in `_catalog`.

Net result for `skill__mcp_install` in a session that hasn't recently used it:

[...]

PR #1407, opened 2026-06-06
fix(1406): exclude_tools execution-level block (closes #187 4th faithfulness leak)

tya5:
**[e2e-coder]** — #1406, owner-GO'd as #187 experiment-sequence step 1 (design APPROVED on the issue). Removes the web-leak confound before the faithful re-run.

## Root cause (#187 N=3, code-verified)
`exclude_tools` (#1400) filters only the advertised `tools[]` / `self._catalog`. The LLM called `web__search` by name → not in the post-exclude catalog + has `__` → the #229 salvage `_maybe_salvage_qualified_direct_call` resolved it via `universal_dispatch` and rewrote to `invoke_action(action_name="web__search")` → `_execute_tool` dispatched `invoke_action` (in the catalog) → `dispatch_tool`'s `name in ctx.tool_catalog` gate passed (it sees the **wrapper**, not the inner `web__search`) → executed (`status:ok`). Same bypass for a direct `invoke_action(action_name=web__search)`.

## Fix
In `RouterLoop._execute_tool`, after the salvage rewrite, compute the **effective resolved action** (unwrap `invoke_action`) and reject with a distinct `tool_excluded` error if excluded — **before** dispatch. Covers all three bypass paths (native direct / #229-salvaged / direct `invoke_action`). The #1400 catalog filter stays as the advertisement layer (defense-in-depth: hide + block). Deny is decision-enabling (names the tool, says don't call it). P7-clean (`exclude_tools` is caller data); no-exclusion sessions are inert.

## Test plan
- [x] `pytest tests/test_exclude_execution_block_1406.py -q` — 3 passed: all 3 bypass shapes reject on the **real** `_execute_tool` path + the excluded handler **never executes** (no gold leak) + a non-excluded tool falls through + inert when unset
- [x] tier audit 0; ruff clean
- [x] no-regression `-k "router_loop or exclude or routing_decided or universal"` — 366 passed

This is the real-dispatch-path test the catalog-only #1400 unit could not provide. Pairs with the swe_bench skill retire (the other confound) → then sandbox_2's faithful re-run (③→① test: does direct-edit emerge?).

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -2901,6 +2901,33 @@ class RouterLoop:
         if name not in self._catalog and "__" in name:
             name, args = self._maybe_salvage_qualified_direct_call(name, args)
 
+        # #1406: execution-level exclude enforcement. ``exclude_tools`` is not just
+        # an advertisement filter (#1400 ``_apply_tool_exclusions`` hides excluded
+        # tools from ``tools[]`` / ``self._catalog``) — the LLM can still call an
+        # excluded tool by name, which the #229 salvage rewrites to
+        # ``invoke_action(action_name=<excluded>)`` (or it is called as
+        # ``invoke_action`` directly), and ``universal_dispatch`` then resolves and
+        # EXECUTES it (the #187 N=3 web__search leak). Compute the effective
+        # resolved action — unwrap ``invoke_action`` — and reject if excluded.
+        # Covers all three bypass paths (native direct / salvaged / direct
+        # invoke_action). Catalog filter (#1400) stays as the hide layer
+        # (defense-in-depth). A distinct ``tool_excluded`` kind + decision-enabling
+        # message lets the model adjust ([[deny-message-decision-enabling]]).
+        if self._exclude_tools:
+            effective = args.get("action_name") if name == "invoke_action" else name
+            if effective in self._exclude_tools:
+                return {
+                    "status": "error",
+                    "error": {
+                        "kind": "tool_excluded",
+                        "message": (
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — review: **APPROVE**, merge on CI green. Matches the #1406 approved design.

- **Fix**: at `_execute_tool`, `effective = args.get("action_name") if name=="invoke_action" else name` → reject if in `exclude_tools` (distinct `tool_excluded` kind + decision-enabling message). Covers all 3 bypass paths (native by name / #229-salvaged → invoke_action / direct invoke_action). Catalog filter (#1400) stays as defense-in-depth. P7-clean (exclude_tools is data).
- **Tests (real-dispatch-path, the gap the catalog-unit missed)**:
  - `test_excluded_tool_blocked_at_execution_all_paths`: all 3 paths → `status:error / kind:tool_excluded`, **+ `host.web_search_calls == []`** — the critical assertion that the handler NEVER executed (no gold leak). ✅
  - `test_non_excluded_tool_not_blocked`: targeted (non-excluded falls through). ✅
  - `test_no_exclusion_set_is_inert`: empty `exclude_tools` = inert — the blast-radius guard confirming **zero behavior change** for normal chat/plan/phase (only the planner's `{plan}` + SWE's web use it). ✅

Aligned with the owner's "general-features-to-solve is fine" thesis: web-exclusion blocks **external answer-lookup** (gold leak), not reyn's general capabilities — the agent can still build skills / use file/exec/plan; it just can't web-search the solution.

→ CI green → I merge. This closes the 4th faithfulness leak (one of the two confounds to remove before the re-run).

Issue #1414, opened 2026-06-07
permission default file-zone anchor is host project_root (latent under container backend, non-grant interactive)


tya5:
**[e2e-coder]** — split out of #187 B3 (kept the B3 PR focused per lead direction). Latent, lower priority.

## Summary
`PermissionResolver._project_root` (permissions.py:374) anchors the **default file read/write zone** (`_in_default_read_zone` / `_in_default_write_zone`, via the AgentLayer at effective.py:121/127) on the **host** project root (`Path.cwd()` / `_find_project_root`). Under a container backend (`base_dir=/testbed`, #1410/#1411) the agent's file ops target /testbed, but the zone anchor stays host — so a `/testbed/.reyn` or `/testbed/reyn` write would be evaluated as "outside the default write zone" because the zone is `host/.reyn`, `host/reyn`.

## Why it's only latent (not the B3 blocker)
B3 (the SWE dogfood edit-blocker, fixed in the #187 B3 PR) was the **SandboxLayer** denying a relative repo write because the op handler passed the raw relative path and the gate resolved it against host cwd — see the deterministic repro. That path runs with `--grant-file-write` (config file.write=allow), which **bypasses the AgentLayer zone** entirely (`_is_config_approved`). So the zone anchor was never the operative term in the SWE path.

This zone-anchor issue would only manifest in a **non-grant interactive container** run (no `--grant-file-write`), where a write into the container repo's own `.reyn/`/`reyn/` default zone would be denied because the zone is anchored on the host root.

## Proposed direction (when prioritized)
Thread the container repo root (the existing `ws_base_dir` from `build_environment_backend`, already the single source for Workspace base_dir + sandbox write_paths) into the resolver as a **separate file-zone-root**, distinct from the host-side approvals/config base (`_approvals_path`, offload). This is the permission-layer parallel of the #1410 base_dir(container)-vs-state_dir(host) split:
- file-operation zone (read/write zone anchor) → container /testbed
- approvals.yaml base + offload + config → host

## Acceptance
- A non-grant container run can write to `/testbed/.reyn/...` (in the container default zone) while `approvals.yaml` stays host-side.
- Host / interactive behavior byte-identical (file_zone_root defaults to project_root).

Refs #187
🤖 Generated with [Claude Code](https://claude.com/claude-code)

tya5:
[...]

PR #1413, opened 2026-06-07
fix(187): B1 empty-choices router-loop crash + swallow instrument + A did-you-mean

tya5:
**[e2e-coder]** — #187 Stall B1 (edit-blocker) root fix + instrument + Stall A. Design pre-approved by lead-coder; opening for code-review → merge.

## B1 root cause (sandbox_2 traceback, primary evidence)
The LiteLLM proxy intermittently returns a **200 with an empty `choices` list** for `gemini-2.5-flash-lite`. It's not an API error → litellm neither raises nor retries it → `response.choices[0]` (`llm.py` `call_llm` / `call_llm_tools`) **IndexErrors**, which the router-loop handler swallows → the turn silently dies mid-work (`req=resp+1`, no logged response) → **edit 0/6**. This is a **general OS reliability bug** affecting all agents/runs, surfaced by the #187 SWE arc — not SWE-specific.

Full stack: `session.py:3053 _run_router_loop` → `loop.run` → `router_loop.py:2044 await _llm(...)` → `llm.py:1255 response.choices[0]` (unguarded).

## Changes
**B1 root fix (`llm.py`)** — single retry chokepoint:
- new `EmptyLLMResponseError`; `_is_retryable_exc` treats it as transient.
- `_llm_call_with_retry` raises it on empty `choices` so the **existing backoff/retry machinery** retries (the condition is transient). Covers **both** `choices[0]` callsites (`call_llm` line 1022 + `call_llm_tools` line 1255) at one point. On exhaustion: a **named** error, not a cryptic `IndexError`.

**B1 instrument (`session.py` swallow handler)** — `logger.exception` (stderr traceback) + P6 event `router_loop_terminated_by_exception` (error type + repr). Any future swallowed loop error is now primary-evidence (deny-message principle: surface, don't silently swallow). The classified outbox summary is unchanged — the instrument is additive.

**A — did-you-mean (`dispatcher.py`)** — the `unknown_tool` deny message appends a `difflib` `"Did you mean <X>?"` hint naming the **closest real catalog tool** (cutoff 0.6), so a near-miss guess (the #187 `source__grep` → deterministic stop) can self-correct. Deny-message-decision-enabling.

## Review points (lead's 3)
[...]

PR #1410, opened 2026-06-06
fix(187): root chat file ops on the container repo base_dir (5th structural defect)

tya5:
**[e2e-coder]** — #187 5th structural defect (sandbox_2 step-3 trace → my OS fix → sandbox_2 re-verify). Pure-plumbing, lead+sandbox_2 verified scope.

## Defect (primary evidence)
run-once `--env-backend=docker`: **exec runs in-container but file ops (read/grep/glob/edit) rooted on the HOST reyn cwd**, not the in-container repo (`/testbed`). `file__read` astropy paths not_found 134/134; `file__grep` 0 astropy matches (all hits = reyn's own tests/venv → base_dir = host cwd). The agent never saw the target tree → it thrashed an empty FS. **This invalidates step-3's behavior reading (FS-thrashing, not reasoning) — not weak-model.**

## Root cause — construction-forwarding gap (same class as #1401/#1406)
`build_environment_backend` returns the container repo `base_dir` (attach mode, `env_backend.py:119` → `/testbed`), but:
- `chat.py` captured it as `_ws_base_dir` — underscore-**discarded**, never forwarded;
- the chat OpContext `Workspace` (`session.py:4759`) got `environment_backend=` but no `base_dir=` → defaulted `Path.cwd()` (the host reyn repo).

The backend was forwarded; its **partner** container base_dir was not.

## Fix (pure plumbing — no /workspace-vs-/testbed layout change; attach mode already returns /testbed)
- `ChatSession` gains `workspace_base_dir` / `workspace_state_dir` params.
- `chat.py` forwards `build_environment_backend`'s base_dir + host-side state_dir.
- the Workspace is built with `base_dir=` / `state_dir=`.

File ops now resolve to `/testbed` **in the container** (where exec + `git diff` operate); state_dir stays **host-side** (decoupled — survives container death, doesn't pollute the in-container diff). Host backend (no base_dir) → cwd default, unchanged.

## Test plan
- [x] `tests/test_workspace_basedir_docker_187.py` (3): read resolves to `/testbed` in-container; **write round-trip** (lands at `/testbed` where `git diff -C /testbed HEAD` looks — lead's load-bearing acceptance, `docker exec -i` for stdin); host-default unchanged
- [x] tier audit 0; ruff clean
- [x] no-regression `-k "workspace or chat or run_once or state_dir_read or 1200 or fs_seam or 1390"` — 312 passed

[...]

PR #1415, opened 2026-06-07
fix(187): B3 — resolve file-op path against workspace base_dir before the permission gate

tya5:
**[e2e-coder]** — #187 B3 (8th defect, the SWE edit-blocker). Root cause confirmed by lead (ran repro + source cross-check, arbiter GO) and sandbox_2 (independent verify). Opening for code-review → merge.

## What B1 unblocked, then B3 blocked
After the B1 empty-choices fix (#1413), the general agent **reached `file__write` for the first time** in the #187 arc — but the write was **DENIED**.

## Root cause (flow-trace + deterministic repro — corrected an initial localization)
The file-op handler (`op_runtime/file.py`) passed the **raw relative** `op.path` to `require_file_write`. The gate's `SandboxLayer._path_under` then did `Path(path).resolve()` against the **host process cwd** — not the workspace base_dir. Under a container backend (`base_dir=/testbed`) a relative repo write like `astropy/io/ascii/html.py` was checked against the host cwd, fell outside the sandbox `write_paths` cap (`[/testbed]`), and was denied — even though `Workspace.write_file` resolves the same path against /testbed and lands it there.

Key correction: `--grant-file-write` (config file.write=allow) **already bypasses the AgentLayer zone** via `_is_config_approved`, so the `SandboxLayer ∩` on the relative-vs-cwd mismatch was the real denier — **NOT** the `project_root` zone anchor (the static denial text *"outside the default write zone"* misled the first read). The read/write asymmetry (reads worked) is explained: `read_paths=[]` (broad → sandbox ⊤) vs `write_paths=[/testbed]` (tight).

Deterministic repro (pure permission decision, no container) — both lead and sandbox_2 ran it:
```
RAW relative  astropy/io/ascii/html.py        → DENIED  (exact prod message)
workspace-resolved /testbed/astropy/...         → GRANTED
```

## Fix (focused, lead-directed)
Resolve `op.path` against `ctx.workspace.base_dir` in the handler **before** the gate, so the permission check sees the same absolute target the write/read will hit. **Read + write symmetric** (`require_file_read` also passed the raw path). Behaviour-preserving for the host case (`base_dir == cwd` → identical absolute path). Aligns with `Workspace`'s documented design (the gate operates on the absolute paths it resolves).

## Round-trip gate (the #1410 write-lands lesson)
`test_b3_relative_write_resolves_against_base_dir_and_lands`: a relative write under a non-cwd base_dir is **granted AND lands** under base_dir (not host cwd). Plus a falsification (sandbox cap excluding base_dir still denies → the cap stays load-bearing) and read/write symmetry. Real instances, no mocks. **Verified the tests fail pre-fix (exact B3 denial) and pass post-fix.**

## Scope note
The latent host-anchored **default file-zone** (only manifests in a non-grant *interactive* container run — NOT the `--grant-file-write` SWE path) is tracked separately in **#1414** to keep this fix focused.

## Test plan
- [x] `pytest tests/test_op_runtime_file_b3_basedir_resolve_187.py -q` → 3 passed; verified fail pre-fix (stash), pass post-fix
- [x] file-op/permission/sandbox/workspace/control_ir clusters (`-k "op_runtime or permission or workspace or control_ir or s31 or sandbox or file_"`) → 614 passed, 3 skipped
- [x] `ruff check` (changed files) → clean
- [x] `scripts/test_tier_audit.py --strict` (new test) → 0 errors
- [ ] full `pytest -q` from rootdir (running)
- [ ] live round-trip (sandbox_2): `file__write astropy/.../html.py` → granted + `git -C /testbed diff` shows it

## Next
On merge → sandbox_2 re-runs 13453 (the live round-trip gate): **does the agent now land an actual edit + produce a diff** — the first of the #187 arc, the capability core.

Refs #187

### DIFF
--- a/src/reyn/op_runtime/file.py
+++ b/src/reyn/op_runtime/file.py
@@ -150,6 +150,31 @@ def _read_inline_cap(ctx: OpContext) -> int:
     return control_ir_inline_cap(model_str, events=ctx.events, phase=ctx.skill_name)
 
 
+def _resolve_for_gate(ctx: OpContext, path_str: str) -> str:
+    """Resolve a file-op path against the workspace base_dir so the permission
+    gate checks the SAME absolute target the op will actually read/write.
+
+    #187 B3: the handler previously passed the raw (often relative) op.path to
+    ``require_file_*``; the gate's SandboxLayer then resolved it with
+    ``Path(path).resolve()`` against the HOST process cwd — not the workspace
+    base_dir. Under a container backend (base_dir=/testbed) a relative repo
+    write like ``astropy/io/ascii/html.py`` was therefore checked against the
+    host cwd, fell outside the sandbox ``write_paths`` cap (``[/testbed]``), and
+    was DENIED — even though ``Workspace.write_file`` resolves that same path
+    against /testbed and would land it there. Resolving here closes the base
+    mismatch (the Workspace already documents that the gate operates on the
+    absolute paths it resolves). Behaviour-preserving for the host case
+    (base_dir == cwd → the identical absolute path).
+    """
+    ws = getattr(ctx, "workspace", None)
+    if ws is None:
+        return path_str
+    p = Path(path_str).expanduser()
+    if p.is_absolute():
+        return str(p.resolve())
+    return str((ws.base_dir / p).resolve())
+
+
 async def handle(op: FileIROp, ctx: OpContext, caller: Literal["preprocessor", "control_ir"]) -> dict:
     # Permission check (single point for both frontends). For
     # `regenerate_index` the file actually written is `output_path`, not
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — code-review: **APPROVE**, merge on CI green. Focused op-handler-resolve, matches the arbitration-corrected root cause (verified via the /tmp/b3_repro.py cross-check), dual-verified with sandbox_2.

- **Fix (correct locus)**: `_resolve_for_gate(ctx, path) → (ws.base_dir / p).resolve()` before the permission gate, applied to write_target + move dest + read (symmetric). The permission check now resolves against the SAME base_dir (/testbed) as the actual `workspace.write_file` — eliminating the host-cwd-vs-/testbed mismatch that was B3. **host backend byte-identical** (base_dir==cwd → same absolute path). ✅
- **Tests (mock-free, real instances, 0 MagicMock)**:
  - `test_b3_relative_write_resolves_against_base_dir_and_lands` — round-trip: relative write under non-cwd base_dir GRANTED **+ lands** under base_dir (write-lands, #1410 lesson). ✅
  - ★`test_b3_sandbox_write_cap_still_load_bearing` — the **falsification/safety-net** check: a write outside the cap is STILL denied. The fix corrects path-resolution WITHOUT weakening the SandboxLayer (feedback_safety_net_coverage_wiring_sufficiency_verify). ✅
  - `test_b3_relative_read_resolves_against_base_dir` — read symmetry. ✅
  - Reproduces the exact B3 condition (`project_root=host ≠ base_dir`); pre-fix fails / post-fix passes (stash-verified red-green). ✅
- zone-anchor latent correctly **separated to #1414** (non-grant interactive only, SWE-path irrelevant) — PR stays focused. ✅

→ Merge on CI green → sandbox_2 re-run (13453) = does the agent now produce a real **write→diff** (all of permission-gate + sandbox-cap + write resolve on /testbed)? = the capability core. Or a placeholder (B2 content-assembly, gated behind B3). The arc's payoff measurement.

tya5:
**[e2e-coder]** Test-plan completion (final item): full `pytest -q` from rootdir → **6879 passed**, 10 skipped, 3 xfailed. The only 2 failures are the pre-existing local-env quirks unrelated to this diff:
- `test_swebench_missing_honest_skip` — `swebench` is installed in the local venv, so the `sys.modules` guard mis-fires the missing-path (CI has swebench absent → passes there)
- `test_legacy_no_media_store_path_returns_inline_content` — HTML-extraction-library version difference (heading stripped locally)

Both are stash-proven not-my-diff (identical with my change stashed) and CI was green on this PR (authoritative). All Test plan items complete. ✅

Issue #1412, opened 2026-06-06
[tech-debt] Single-source the two make_router_op_context impls (ChatSession vs RouterHostAdapter)


tya5:
**[e2e-coder]** — follow-up from #1410/#1411 (the #187 wrong-FS drift). lead-tracked; scope decision after the #187 re-run.

## Problem
Two parallel impls build the chat-router OpContext:
- `ChatSession._make_router_op_context` (session.py) — used by the legacy `_file_op` / `_mcp_call_tool` callbacks;
- `RouterHostAdapter.make_router_op_context` (router_host_adapter.py) — the **LIVE** factory the registry file-op dispatch uses (`router_loop.py op_context_factory=getattr(host,"make_router_op_context")`).

They drifted: #1410's `base_dir`/`environment_backend` fix landed on the legacy one only (its test green on that seam) while live file ops still ran on the host cwd; #1411 fixed the live one. A future scoped-param add will drift again.

## Why not a 1-line delegate
The two OpContexts **diverge** by caller need: ChatSession's carries `run_id`/`agent_id`/`sandbox_backend`; the adapter's carries `intervention_bus`/`media_store`/`compact_now`/`multimodal_config`. Single-sourcing needs reconciliation (a superset or parameterized OpContext), not `return self._router_host.make_router_op_context()`.

## Options
1. Focused `build_chat_router_workspace(*, events, perm, base_dir, state_dir, environment_backend)` helper for the **drifted construction** (the Workspace) — low-risk, prevents this specific class; both impls call it.
2. Full single-source of the OpContext (reconcile the divergent fields) — bigger; natural to fold into #1402's `build_scoped_chat_session`.

## Trigger
After the #187 re-run lands (so the FS fix is stable) — lead's call on (1) vs (2) vs #1402-fold. This is the 2nd parallel-construction-drift instance in #187 (cf. #1402 evidence).

tya5:
**[e2e-coder]** — complete OpContext-kwargs seam-map (lead-coder's diff of the two builders, fe08d4b9). This is the design for the single-source unify: the divergence is **bidirectional**, so the fix is a **per-field UNION** (preserve each field's intent), NOT a one-way copy.

Workspace seams (`base_dir` / `state_dir` / `environment_backend`) are already IDENTICAL (aligned in #1411).

**Live (`RouterHostAdapter.make_router_op_context`) LACKS** (legacy has):
[...]

PR #1411, opened 2026-06-06
fix(187): root the LIVE router OpContext on the container repo (#1410 patched the wrong method)

tya5:
**[e2e-coder]** — real fix for the #187 wrong-FS defect (sandbox_2 N=6 live trace + lead canonical verify caught #1410 patched the wrong method). lead+sandbox_2 root-cause confirmed.

## What #1410 missed
The registry file-op dispatch uses `op_context_factory = getattr(host, "make_router_op_context")` (router_loop.py), and the chat host is `RouterHostAdapter` → the **LIVE** factory is `RouterHostAdapter.make_router_op_context` (router_host_adapter.py:1428), which built `Workspace(events, perm, skill_name)` with **no base_dir + no environment_backend** → host cwd + HostBackend. #1410 instead patched the **parallel** `ChatSession._make_router_op_context` (used only by the legacy `_file_op` / `_mcp_call_tool` callbacks). The two impls had drifted; #1410's unit test was **green on the legacy seam** while live file ops still ran on the host (astropy read 0/6, edit 0/6). sandbox_2's in-method debug print proved the ChatSession method never fired live.

## Fix (the LIVE path)
`RouterHostAdapter` gains `environment_backend` / `workspace_base_dir` / `workspace_state_dir` params, threaded from `ChatSession` (same source as #1410); `make_router_op_context` now builds the Workspace with `base_dir` / `state_dir` / `environment_backend`. File ops + the exec sandbox `write_paths` (derived from `workspace.base_dir`) now root on `/testbed`.

## Test — exercises the LIVE path (the lesson from the #1410 miss)
`tests/test_router_opcontext_basedir_live_187.py` calls **`session._router_host.make_router_op_context()`** — the exact factory the dispatch resolves — and asserts `Workspace.base_dir == /testbed` (+ sandbox write_paths container-scoped); host-default unchanged. A regression to host-cwd on the live path fails here. (Verify the live dispatch impl, not a parallel seam.)

## Single-sourcing (lead's drift-prevention #2) — presented for review, not bundled
[...]

Issue #1289, opened 2026-06-04
[#1200 follow-up] per-frontend container-chat activation (entry-point --env-backend wiring)


tya5:
**[e2e-coder]** — Deferred-as-tracked from #1200 (agent-level-uniform backend injection). lead-coder ratified **defer** over activate-now (YAGNI: zero container-chat consumers today; same discipline as the #1092 plan-activation defer in #1285). #1199 (Stage 3) does NOT cover this — it refers agent-level-uniform injection to a separate issue (= #1200), so this is its dedicated tracking home.

## Context: the wiring is done (#1200), activation is not

#1200 (PR-F1 #1287 + PR-F2 #1288) completed the **in-session wiring**: `ChatSession` threads the agent's backend to BOTH seams — FS (`Workspace`, F1) + exec (`OpContext.sandbox_backend`, F2) — with the single-shared-sandbox invariant verified (one backend reaches both seams as the same object; plan inherits chat via `_PlanStepHost` delegation). So a caller can now pass one backend and it uniformly reaches chat's FS + exec + plan. **chat-in-container is now POSSIBLE.**

The wiring is **inert by default** (`environment_backend`/`sandbox_backend` = `None` → `HostBackend` / `get_default_backend`, byte-identical) — a **verified capability awaiting activation**, not dead code (#1092 unwired-foundation pattern).

## The gap (this issue)

**No frontend entry point builds + passes the agent-level backend yet.** Primary evidence: none of `cli/commands/chat.py`, `web/deps.py`, `chainlit_app/app.py`, `cli/commands/dogfood.py` have backend-build inputs (`--env-backend` / `container` / `repo-dir` flags). Only `cli/commands/run.py` (the Agent/OSRuntime entry) builds an `EnvironmentBackend` (via `_build_environment_backend`).

## Scope (when picked up)

Per-frontend activation — build the agent-level `environment_backend` (+ the same instance for `sandbox_backend`) and pass BOTH to `ChatSession`:
- **CLI** (`chat.py`, `dogfood.py`): generalize `run.py:_build_environment_backend` (they have argparse args for `--env-backend`/`--container`/`--repo-dir`).
- **web / chainlit**: config-driven — derive from the already-flowing `sandbox_config` or add deployment config, IF/when container-chat-via-web is needed.

## Trigger to pick up

[...]

PR #1419, opened 2026-06-07
fix(187): thread the exec sandbox backend instance into the LIVE router OpContext — 10th defect

tya5:
**[e2e-coder]** — #187 10th structural defect (the exec-seam analog of #1410/#1411). Root-caused by sandbox_2 (primary evidence), confirmed by lead. Opening for code-review.

## What surfaced it
After the exec-dispatch fix let the agent's exec calls dispatch (3/3, via the 3411 `__`-route), sandbox_2's Set C v2 showed exec running on the **host seatbelt backend, not docker /testbed**:
```
{"backend":"seatbelt","returncode":-1,"stderr":"[Errno 2] No such file or directory: '/testbed'"}  (3/3)
```
So the agent's **real** verify attempts (verify intent is genuine — it reached exec) always failed → it could never see its edit broke → no self-correct. This refines the #202 picture again: the failure is **structural** (exec backend mis-wired), not under-exploration.

## Root cause
`op_runtime/sandboxed_exec` selects `ctx.sandbox_backend or get_default_backend(...)`. The **live** `RouterHostAdapter.make_router_op_context` **omitted** `sandbox_backend=` on its OpContext → `None` → host seatbelt. The **legacy** `ChatSession._make_router_op_context` (session.py:4824) passes it. Same live-vs-legacy seam gap as #1410/#1411 — **3rd instance** (base_dir → env_backend → now the exec backend instance).

## The type nuance (lead flagged)
The adapter's existing `sandbox_backend` (str) drives only the D14 exec **visibility** gate — a naive `sandbox_backend=self._sandbox_backend` mirror would be a type mismatch (`OpContext.sandbox_backend` is a `SandboxBackend` instance). The exec **execution** needs the *instance*. chat.py injects the same docker `env_backend` at both the FS and exec seams (#1289 single-shared-sandbox invariant), so `ChatSession._sandbox_backend` is that exec-capable instance.

## Fix
Thread the instance via a **distinct** `sandbox_backend_instance` param into the adapter → forward to `OpContext.sandbox_backend` (parallel to how #1411 threaded `environment_backend`/`base_dir`). Host / interactive default (no instance → `None` → host backend) unchanged.

## Test plan
- [x] live-path test (`session._router_host.make_router_op_context()`, the dispatch the router actually uses): injected backend → `ctx.sandbox_backend is` that backend; none → `None`. **Verified fail pre-fix (None), pass post-fix.**
- [x] clusters (`-k "router_host or sandboxed_exec or router_opcontext or sandbox_policy or op_context or 1411 or 1410"`) → 63 passed
- [x] `ruff check` (changed files) → clean
- [x] `scripts/test_tier_audit.py --strict` → 0 errors
- [ ] full `pytest -q` from rootdir (running)
- [ ] live (sandbox_2): Set C v3 — exec runs in /testbed (`backend != seatbelt`, no `/testbed not found`) → the agent's verify loop can finally complete

## Structural follow-up
The deeper fix — unifying the two op-context builders (`make_router_op_context` vs `_make_router_op_context`) so seams can't keep diverging — is the **#1412** single-source theme; this 3rd divergence is more evidence for it (per lead's whack-a-mole concern). Folded there, not bundled here.

Refs #187, #1412

### DIFF
--- a/src/reyn/chat/services/router_host_adapter.py
+++ b/src/reyn/chat/services/router_host_adapter.py
@@ -172,6 +172,18 @@ class RouterHostAdapter:
         environment_backend: Any = None,
         workspace_base_dir: Path | None = None,
         workspace_state_dir: Path | None = None,
+        # #187 exec-seam (10th defect): the SandboxBackend INSTANCE for exec
+        # EXECUTION (docker for in-container repos). Distinct from the
+        # ``sandbox_backend`` STRING above (which only drives the D14 exec
+        # visibility gate): the op handler (op_runtime/sandboxed_exec) reads
+        # ``ctx.sandbox_backend`` (the instance) and falls back to the host
+        # seatbelt backend when it is None. Without threading this, the LIVE
+        # router's exec ran on the host (``No such file or directory:
+        # '/testbed'``), so the agent's verify loop always failed. Parallel to
+        # ``environment_backend`` for FS (#1411) — the legacy
+        # ``ChatSession._make_router_op_context`` already passes it; the live
+        # adapter omitted it (the same live-vs-legacy seam gap as #1410/#1411).
+        sandbox_backend_instance: Any = None,
         # FP-0034 Phase 2 step 5: ActionUsageTracker for hot list freq+recency.
         # ChatSession passes the session-scoped tracker; None when wrappers are
         # off or hot_list_n == 0.
@@ -311,6 +323,7 @@ class RouterHostAdapter:
         self._embedding_model_class = embedding_model_class
         # FP-0034 Phase 2
         self._sandbox_backend = sandbox_backend
+        self._sandbox_backend_instance = sandbox_backend_instance
         self._environment_backend = environment_backend
         self._workspace_base_dir = workspace_base_dir
         self._workspace_state_dir = workspace_state_dir
@@ -1477,6 +1490,9 @@ class RouterHostAdapter:
             media_store=self._media_store,
             # #272/#1128: voluntary-compaction capability for the compact op.
             compact_now=self._compact_now,
+            # #187 exec-seam: the exec backend INSTANCE so sandboxed_exec runs
+            # in the container repo (/testbed), not the host seatbelt fallback.
+            sandbox_backend=self._sandbox_backend_instance,
             # #1339 / sandbox-model completion: resolve the operator-or-default
             # sandbox policy onto the router OpContext (was None → the handler
             # fell back to LLM-set op fields = the sandbox-escape gap).
--- a/src/reyn/chat/session.py
+++ b/src/reyn/chat/session.py
@@ -1855,6 +1855,13 @@ class ChatSession:
             environment_backend=self._environment_backend,
             workspace_base_dir=self._workspace_base_dir,
             workspace_state_dir=self._workspace_state_dir,
+            # #187 exec-seam (10th defect): the exec backend INSTANCE so the LIVE
+            # router's sandboxed_exec runs in the container repo, not the host
+            # seatbelt fallback. Same instance the legacy _make_router_op_context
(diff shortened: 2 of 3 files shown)


tya5:
**[e2e-coder]** Scope note: lead's full seam-map shows the live adapter also lacks **`agent_id`** (latent — only consumer is `mcp.py:43 MCPClient(agent_id=ctx.agent_id)`, non-load-bearing when web/MCP are absent, as in the SWE run). Kept out of this PR to stay focused on the load-bearing exec seam; `agent_id` + the bidirectional per-field UNION are captured in #1412 (the structural single-source that subsumes this forward).

tya5:
**[lead-coder]** — ✅ **APPROVE + merging.** Full implementation review (not just diff-match) complete.

## End-to-end load-bearing verified (the gate beyond diff-match)
Read the full source→forward→consumer chain to confirm this is NOT a correct-but-inert forward:
- **Source** (`cli/commands/chat.py:402-403`): `build_environment_backend(args)` injects the SAME docker instance to BOTH seams — `environment_backend=env_backend, sandbox_backend=env_backend` (#1289/#1200 single-shared-sandbox). So the run-once (`run_once.py → chat.run once=True`) path sets `ChatSession._sandbox_backend = DockerEnvironmentBackend`.
- **Forward** (this PR): `session.py:1855 sandbox_backend_instance=self._sandbox_backend` → `adapter:1490 OpContext(sandbox_backend=self._sandbox_backend_instance)`.
- **Consumer** (`op_runtime/sandboxed_exec.py:30`): `ctx.sandbox_backend or get_default_backend(...)` → now the docker instance → exec runs in `/testbed`.

The source was **always** wired (chat.py:403) + the legacy `_make_router_op_context:4824` forwarded it; **only the live `RouterHostAdapter.make_router_op_context` omitted the forward** — exactly the #1410/#1411 live-vs-legacy seam class, 3rd instance. This PR closes precisely that gap. `DockerEnvironmentBackend` implements both protocols (real instance, used unmodified in the test — no fake-backend gap).

## Tier rule check ✓
- **Docstring Tier**: `Tier 2` (OS invariant — the live op-context factory forwards an injected seam). Correct (not Tier 1 contract, not Tier 3 replay, not Tier 4).
- **No mocks**: real `DockerEnvironmentBackend` + real `ChatSession`. ✓
- **Assertion shape**: identity (`ctx.sandbox_backend is backend`) on the **LIVE** dispatch (`session._router_host.make_router_op_context()`, not the legacy `_make_router_op_context`) — no format-pinning, no private-state. ✓ (applies the #1411 live-path lesson)
- **Falsification**: verified fail pre-fix (`None`) / pass post-fix (stash) — the test actually catches the defect. ✓
- **Invariant essentiality**: host-default-unchanged pair (`None→None`) guards the interactive/host path. ✓

## Workflow
- `[e2e-coder]` role-prefix ✓ · Test plan pre-merge items ticked (63-cluster / ruff / tier-audit green) · CI all 4 SUCCESS (pytest 3.11+3.12 = full suite green) · Set C v3 / Set A correctly deferred post-merge (need the merge to re-run).
- Structural union (sandbox_backend ✓ here / agent_id latent / 4 live-only fields) correctly scoped to **#1412** per-field-UNION, not bundled. ✓

No open concerns. Merging.

tya5:
**[e2e-coder]** Test-plan completion (final item): full `pytest -q` from rootdir → **6881 passed**, 10 skipped, 3 xfailed. The only 2 failures are the known pre-existing local-env quirks unrelated to this diff (`router_host_adapter`/`session`):
- `test_swebench_missing_honest_skip` — `swebench` installed in the local venv mis-fires the missing-path guard (passes in CI where swebench is absent)
- `test_legacy_no_media_store_path_returns_inline_content` — HTML-extraction-library version difference

Both stash-proven not-my-diff and CI-authoritative. All Test plan items complete. ✅

PR #1422, opened 2026-06-07
feat(187): Stage B — list_actions returns name+description+schema (selection-grade)

tya5:
**[e2e-coder]** — #187 weak/strong action-discovery, **Stage B of 3** (B→A→C). Owner-GO'd; design patch/live-verified (universal-catalog.md, #1421). Flow-trace-first seam map approved by lead. sandbox_2 reviews.

## Why
The Set-A capstone showed weak models under-explore the discoverable catalog and don't make the `describe_action` round-trip (describe 1/10 in the weak-verify). Stage B makes `list_actions` itself **selection-grade**: when the browse is narrowed to a category, each item carries the full `description` + `input_schema` `describe_action` returns (name + description + schema). This:
- moves the schema onto the path the model already uses (no separate describe round-trip), and
- **inherits the schema-blind-hallucination protection the ARS block provided** — which is what makes Stage A (the ARS removal, held branch 45428f82) safe to land.

## Change (`tools/universal_catalog.py`)
- `_describe_one(qualified_name, ctx, registry) -> {description, input_schema} | None` — the **shared selection-grade core** (per-resource fields win over the dispatcher target's generic ones, falling back to `target.parameters`/`.description` for operation categories like file__edit/exec). Returns `None` for unresolvable names.
- `_handle_describe_action` refactored to use `_describe_one` (DRY).
- `_handle_list_actions` enriches each page item via `_describe_one` **only when category-narrowed** (`valid_filter` non-empty) with graceful skip on `None`. **Bounded**: narrowed view + limit-capped page; the unfiltered alphabetical browse stays compact for breadth scanning.

## Test plan (Tier 1 contract, real ToolContext, no mocks)
- [x] narrowed item carries full `description` + `input_schema`
- [x] static op (file__edit) AND resource (skill__X) both enriched
- [x] unfiltered browse stays compact (no per-item schema)
- [x] `_describe_one` returns `None` for an unresolvable name (→ graceful skip)
- [x] ★**`list ≡ describe` by construction** — narrowed item's description + input_schema EQUAL `describe_action`'s for the same action (shared `_describe_one`); guards future helper-bypass drift
- [x] `ruff check` + `test_tier_audit.py --strict` → clean / 0 errors
- [x] full `pytest -q` → 6887 passed (only the 2 known local-env quirks fail: swebench-installed, web_fetch extraction-lib)

Replay note: the change is runtime-response-only (not the `tools[]` request key), so replay fixtures are unaffected (259 catalog/describe/list/replay cluster tests pass).

## Next
Stage A (rebase 45428f82 onto B → ARS removal, now safe because B delivers schema) → Stage C (tier-gated mechanical SP mandate, parameterized light=ON/strong=OFF).

Refs #187

### DIFF
--- a/src/reyn/tools/universal_catalog.py
+++ b/src/reyn/tools/universal_catalog.py
@@ -826,6 +826,24 @@ async def _handle_list_actions(
     total = len(items)
     page = items[offset:offset + limit]
 
+    # Stage B (#187): on a NARROWED (category-filtered) browse, enrich each page
+    # item with the SAME full description + input_schema describe_action returns
+    # — via the shared _describe_one, so list ≡ describe BY CONSTRUCTION — giving
+    # the LLM selection-grade detail (name + description + schema) without a
+    # separate describe_action round-trip (which weak models rarely make). This
+    # inherits the schema-blind-hallucination protection the removed ARS block
+    # used to provide. Bounded: only the narrowed view (valid_filter non-empty)
+    # is enriched and the page is limit-capped, so the unfiltered alphabetical
+    # browse stays compact for breadth scanning.
+    if valid_filter:
+        from reyn.tools import get_default_registry
+        _registry = get_default_registry()
+        enriched: list[dict[str, Any]] = []
+        for it in page:
+            one = _describe_one(it["qualified_name"], ctx, _registry)
+            enriched.append({**it, **one} if one is not None else it)
+        page = enriched
+
     response: dict[str, Any] = {"items": page, "total": total}
     if _should_inject_hidden_state_hint(ctx.router_state):
         response["hint"] = _HIDDEN_STATE_HINT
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Stage B code review → **APPROVE** (merge gated on CI green + sandbox_2 live real-path verify; see below).

Reviewed impl + tests against the approved seam map and the 3 requirements. Verified seams against current code (not just the description).

## Impl ✓
- `_describe_one(qn, ctx, registry) → {description, input_schema} | None` is the **exact DRY extraction of the proven describe resolution** (`_resource_input_schema ?? target.parameters`, `_resource_description ?? target.description`). `describe_action` now calls it too → **list ≡ describe BY CONSTRUCTION**. Reuse of the existing resolver (not a new path) — correct.
- `_handle_list_actions` enriches **only when `valid_filter` (category-narrowed)**, iterating the already-`limit`-capped `page` → doubly bounded (category-narrow × limit≤20). No large-category blow-up.
- Graceful skip on unresolvable (`{**it, **one} if one is not None else it`) — keeps `short_description`, never errors the list.
- Returns ONLY `{description, input_schema}` — `metadata` + B41 post-call directive correctly stay in `describe_action`, not carried into list items.

## Tests ✓ (Tier rule check)
- Docstring declares **`Tier 1:`** (contract) — correct; these pin return-shape, not algorithm internals.
- **No mocks of collaborators** — real `ToolContext` + `RouterCallerState` + a `_FakeHost`/`_FakeEvents` (Fakes, not `MagicMock`). `permission_resolver=None` is fine here: these are read-only discovery handlers, NOT an enforcement-guarding test, so the real-resolver rule does not apply.
- ★ `test_list_equals_describe_by_construction` asserts **both** `description` AND `input_schema` equality (list item == describe) for a static op (`file__edit`) and a resource (`skill__code_review`) — the load-bearing invariant, correctly pinned + guards helper-bypass drift.
- Coverage complete: narrowed-enrich / static+resource both / unfiltered-compact / `_describe_one`-None-graceful.
- No format-pinning, no golden files, no Tier 4.

## Minor (non-blocking)
- `describe_action` now resolves twice (its own `resolve_describe_action`+`lookup`, then `_describe_one` repeats them). Pure in-memory lookups, so negligible; optional future tidy.

## Merge gate
1. CI green (currently BLOCKED = in progress).
2. ★ **sandbox_2 live real-path verify** — flash-lite dogfood on the LANDED code: confirm `describe_action` drops from the common path + schema-in-list → correct-arg invoke (reproduce the patch-replay's describe 14→0 / args 0→12 LIVE, not just the in-process contract test). Per fake-backend-misses-real-integration.

On both green I merge. Stage A (ARS removal 45428f82, rebased on B) only after this lands — B delivers the schema that makes A's ARS removal safe.

## Tier rule check ✓
docstring Tier declared (Tier 1) · no Tier 4 · invariant essential (list≡describe = the structural basis for "describe redundant") · no mocks/format-pinning · audit verified against source.

tya5:
**[dogfood-coder]** Stage B review: **APPROVE** — real-path live merge-gate PASSED.

**Code** — on-spec with the patch-verified design: `_describe_one` DRY extraction = list≡describe BY CONSTRUCTION; enrich bounded (valid_filter + limit-cap; unfiltered browse stays compact); per-resource description/schema win over the dispatcher envelope = selection-grade.

**Tests** — 5/5 pass on branch; real handler + Fake, no mocks (Tier 1). Coverage matches the design checklist: consistency invariant (list item description==describe + input_schema==describe), static-op (file__edit) & resource (skill__code_review per-resource {diff,focus}) both enriched, unfiltered-browse-compact, graceful-skip for unresolvable.

**★ Real-path live merge-gate** (the dogfood gate per the fake-backend-misses-real-integration rule): fed the LANDED enrich output (via THIS PR's own `_describe_one`) into a post-list_actions(exec) decision point, live flash-lite N=20:
- BASELINE (names-only list result): describe_action 11/20, exec args **0/20 correct** (schema-blind: action_name/command).
- LANDED (PR enrich in list result): describe_action **0/20**, exec args **19/20 correct** (argv/timeout), schema-blind **0**.
- = reproduces the patch-replay verify-A result (describe 14→0, args 0→12) on the LANDED code, even stronger (11→0, 0→19). The model leverages schema-in-list to invoke correctly WITHOUT a describe round-trip.

**Note**: on Stage-B-alone (ARS still present) the natural agent path uses the ARS shortcut so list_actions rarely fires; the enrich's full end-to-end agent benefit lands after Stage A removes the ARS (forcing list_actions). The merge-gate above isolates the enrich's behavioral effect at the decision point. Verify script: /tmp/stageB_realpath.py.

APPROVE — CI green + this live verify = merge-ready.

PR #1424, opened 2026-06-07
feat(187): uniform always-on "resume" empty-stop retry at every RouterLoop site

tya5:
**[e2e-coder]** — #187 empty-stop fix: uniform always-on "resume" retry at every RouterLoop site.

## Problem (real-task measured)

On the real-task phase op-loop (`code_review`), the agent hit **33% direct-invoke / ~67% empty-stop** after a tool-call round. An empty stop (finish_reason=stop, no content, no tool_calls) **dead-ends the agent** — it must keep invoking tools to make progress.

Two coupled defects (flow-trace + lead-coder primary verification):

1. **Construction-forwarding gap** (the #1410 base_dir / B3 permission-zone / #1419 exec-seam recurring class): `phase_executor.py`'s `RouterLoop(...)` never passed `empty_stop_retry_directive` → defaulted `None` → the retry mechanism was **structurally unreachable on the agent path**.
2. **Content-semantics, not role** (the lever): the existing retry site already appends `{"role":"user", …}`. The 1/12-vs-11/12 swing is directive *content* — a content-neutral `"resume"` turn re-enters the loop and lets the model continue (real-task: resume recovers the next action 11/12).

## Fix — uniform, always-on, single shared directive (owner decision)

- `router_loop.py`: new `empty_stop_retry_auto: bool = False` flag. The retry gate becomes `directive set AND retries<1 AND (auto OR REYN_EMPTY_STOP_RETRY==1)`.
- **Single shared `EMPTY_STOP_RETRY_DIRECTIVE = "resume"`** (router_loop.py), used UNIFORMLY at all 5 RouterLoop construction sites (chat ×1 / plan-step ×3 / agent op-loop ×1), each with `empty_stop_retry_auto=True` (always-on; env-gate retired).
- **Retired the 3 per-site directives** (chat "write your reply…", plan "step report…", both ending "Do not call another tool"). Per owner: no unevidenced per-site/per-tier differentiation — and the chat directive's anti-invoke framing ("do not call another tool") was itself suspect (chat models also tool-call). Policy: **first uniform, iterate per-site only on measured evidence**.

`"resume"` is verbatim/patch-verified; not paraphrased.

## #1092 ↔ #187 reconciliation (arbitrated with lead, Option A)

The #1092 convergence op-loop signals "done" by emitting no tool_calls; a *content-less* such stop is structurally identical to the empty-stop attractor. Reconciliation:

[...]

Issue #1417, opened 2026-06-07
exec-visibility D14 gate keys off config string, not the injected sandbox backend (latent: noop config + docker backend hides exec)


tya5:
**[e2e-coder]** — latent structural bug surfaced during #187 verify-affordance investigation. **Not a #187 blocker** (non-load-bearing when `sandbox.backend='auto'`, which the SWE run uses), filed as a separate latent issue per lead direction.

## The gap
The `exec` category's D14 visibility gate decides whether `exec__sandboxed_exec` appears in `list_actions(category=["exec"])` via `is_exec_available(sandbox_backend)` (`universal_catalog.py:183`), which returns False for `"noop"`/`None`.

But the **string** passed to that gate is the reyn.yaml **config** value, not the actual injected backend:
- `session.py:1848` threads `sandbox_backend = self._sandbox_config.backend` (the `sandbox.backend` config string) to the catalog.
- The **real** exec capability comes from the injected `self._sandbox_backend` instance (`chat.py:403` `sandbox_backend=env_backend`, i.e. `--env-backend=docker`), used for actual execution at `session.py:4824`.

So a run with `sandbox.backend=noop` (e.g. the routing-only dogfood profile) **plus** `--env-backend=docker` would **hide `exec` from discovery** even though `sandboxed_exec` is functionally available in the container — the agent could neither `list_actions(category=["exec"])` find it nor know the sync verify path exists.

## Class
This is the **construction-forwarding-gap** class again: a capability gated on a config/default string instead of the live injected instance — the same shape as base_dir (#1410/#1411/#1415) and the permission-zone (#1415 B3). It strengthens the case for **#1402** (single-source scoped construction: thread the one capability source to all consumers — Workspace base_dir, sandbox write_paths, permission zone, **and** the exec-visibility gate).

## Proposed direction
The exec-visibility gate should reflect **actual injected exec availability** (the `self._sandbox_backend` instance's real capability), not the `sandbox_config.backend` string. Either derive the gate string from the injected backend, or have the single-source construction (#1402) feed both the actual exec backend and its visibility gate from one source.

## Acceptance
- A run with `sandbox.backend=noop` config + a real injected exec backend (`--env-backend=docker`) exposes `exec` in `list_actions(category=["exec"])` (discoverable, matching actual availability).
- `sandbox.backend='auto'` / host-default behavior unchanged.

Refs #187, #1402
🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1402, opened 2026-06-06
[tech-debt] Unify scoped ChatSession construction across chat-CLI / web-deps / mcp-serve factories (build_scoped_chat_session)


tya5:
**[e2e-coder]** — follow-up from #1401 R4 (lead-raised drift risk). Tracked now so it is not dropped.

## Problem (drift class)
Three ChatSession factories construct sessions with overlapping-but-divergent scoped wiring:
- `cli/commands/chat.py` — full scoped set (env-backend/container, grant-file-write #1398, exclude-tools #1400, sandbox_config, …);
- `web/deps.py:243` (A2A) — partial; capabilities have accreted one-by-one (the B52 comment on `sandbox_config` documents exactly this gap-by-gap pattern);
- `cli/commands/mcp.py` `run_serve` (stdio MCP) — partial.

Each new scoped capability must be hand-added to each factory; an omission silently breaks behavior on the path that missed it (e.g. #1401 found the A2A path missing all 3 of chat's scoped capabilities). Rule-of-three is satisfied (3 factories).

## Proposed fix
Extract a shared `build_scoped_chat_session(...)` (or a `ScopedSessionParams` dataclass + single constructor) that all three factories call, so a new capability is added once and every path gets it **by construction** (completeness-by-construction). Pin with a multi-callsite invariant test (cf. the `feedback_multi_callsite_wiring_audit` pattern / PR #412 precedent).

## Why follow-up, not in #1401's impl PR
- #1401's PR is on the **critical path** (unblocks the #187 faithful N-run); a focused A2A-port lands faster + lower-risk.
- The extraction touches the **working** chat path (#1398/#1400 just landed) + mcp-serve — a cross-cutting refactor of 3 factories belongs in its own reviewable PR, not bundled with new A2A functionality (feature+refactor separation).

## Trigger to pick up
Right after #1401's A2A-port PR lands (so the 3rd scoped path exists and is verified), or the next capability-add — whichever comes first.

tya5:
**[lead-coder]** — owner directive + trigger correction (2026-06-06).

## Owner-flagged: address separately, do not drop
[...]

Issue #1341, opened 2026-06-05
devcontainer: build-based support (dockerFile/build) via ensure_image build-on-demand


tya5:
**[dogfood-coder]** — Tracked follow-up from #1324 (b) devcontainer awareness. Non-blocking.

## Current state (after #1324 b)

`reyn run --env-backend=docker` (launch mode) reads a workspace `devcontainer.json` and maps a minimal-useful subset to the launch: `image`, `postCreateCommand` (→ setup_command), `mounts`, `remoteUser`. **build-based** devcontainers (`dockerFile` / `build` / `dockerComposeFile`) are flagged and currently fall back to the **default reyn base image with a warning** ("build-based devcontainer … not yet supported; using the default image").

## Proposal

Support build-based devcontainers by **building the devcontainer Dockerfile on demand**, reusing the `ContainerLauncher.ensure_image` build-on-demand machinery from #1329 (`docker build -f <devcontainer Dockerfile>` → tag → launch). This is **non-throwaway** (the ensure_image path already exists) and completes industry-standard devcontainer alignment.

Scope to decide here:
- build context + the devcontainer `build.args` / `build.context` fields;
- `dockerComposeFile` (multi-service) — likely out of scope (single-container launcher);
- caching / rebuild policy (rebuild on Dockerfile change?).

Until then the warning + default-image fallback is the documented behavior (no silent surprise). Refs #1324.

tya5:
**[dogfood-coder]** — #1341 design surface（design-first, #1324 ContainerLauncher 延長）。origin/main `4ab98283` 直読。

## 1. Flow-trace — image-acquisition / build seam map (confirmed file:line)

`src/reyn/environment/container_launcher.py`:
- `DevcontainerConfig` L148-165: `image / setup_command / mounts / user / build_based` を保持。**build spec（dockerfile path / context / args / target）は未捕捉**。
- `_map_devcontainer(data)` L207-227: `dockerFile|build|dockerComposeFile` のいずれかで `build_based=True` をセットするだけ。**devcontainer file の location を知らない**（相対 dockerFile/context 解決不可）。
- `load_devcontainer_config(workspace_root)` L229-262: `.devcontainer/devcontainer.json` → `.devcontainer.json` を探索・parse。**`rel`（=どちらが match したか）を知る → devcontainer dir を導出可能**。
[...]

PR #1430, opened 2026-06-07
fix(1417): exec D14 visibility gate keys off injected backend, not config string

tya5:
**[dogfood-coder]** — #1417 exec D14 visibility gate: config-string → injected-backend. Design-first, lead-reviewed (F-min confirmed). Construction-forwarding-gap class (#1402 family — same shape as base_dir #1410/#1411, permission-zone #1415 B3, phase_executor #1424).

## Problem
The `exec` D14 gate (`is_exec_available`, `universal_catalog.py:183`, `!= "noop"`) was fed the reyn.yaml **config string** (`session.py`), but the real exec capability comes from the **injected** `self._sandbox_backend` instance (`--env-backend=docker`, used at `sandboxed_exec.py:30` = `ctx.sandbox_backend or get_default_backend(...)`). So `sandbox.backend=noop` config **+** an injected docker backend would **hide** `exec` from `list_actions(category=["exec"])` even though `sandboxed_exec` is functionally available — undiscoverable.

## Fix (F-min, lead-confirmed)
New `_exec_gate_backend_name(instance, config)`: gate value = the injected instance's `.name` when present (both types expose it — `DockerEnvironmentBackend.name="docker"` / `SandboxBackend.name`), else the config string. `session.py:1832` threads it. `is_exec_available` + the 4-layer string threading (the single source → `RouterCallerState.sandbox_backend` → the `_enumerate_category("exec")` handler) are **unchanged**.
- noop config + docker instance → `"docker"` → exec **visible** (fixes the bug)
- noop config + no instance → `"noop"` → hidden; auto / host-default → unchanged

**F-resolved** (gate==exec-resolution full single-source incl `get_default_backend` platform fallback) is an independent enhancement beyond the filed bug → follow-up note (not this PR).

## Test plan
- [x] `tests/test_exec_gate_injected_backend_1417.py` — 5 unit (instance precedence over noop config / injected-noop hidden / no-instance config fallback / auto unchanged / no-`.name` defensive→hidden) + 2 **handler-integration** (the REAL `_enumerate_category("exec")` path returns `exec__sandboxed_exec` for a docker gate, `[]` for noop — exercises the actual list_actions exec-gate consuming `RouterCallerState.sandbox_backend`).
- [x] `ruff` + `scripts/test_tier_audit.py --strict` clean.
- [x] 194 catalog / exec / backend-injection-threading regression passed.
- [ ] (assessment — your call) live-docker full-run: I argue it's **marginal for #1417** vs #1341. #1341's gate guarded a real `docker build` argv that could fail on the daemon; #1417 is **daemon-independent derivation logic** (an injected instance's `.name="docker"` is identical with/without a live daemon). The real integration risk = the value reaching the handler, now covered by the **handler-integration tests** (real `_enumerate_category`) + code-read single-source threading (`session→adapter:325→get_sandbox_backend:572→RouterCallerState→handler:606`). I have local docker and will add a literal live `list_actions(category=["exec"])` run under noop-config+`--env-backend=docker` if you want it — flagging the cost/value rather than silently skipping.

Refs #1417, #1402.

### DIFF
--- a/src/reyn/chat/session.py
+++ b/src/reyn/chat/session.py
@@ -210,6 +210,28 @@ _TOOL_FAILED_FALLBACK_MSG: dict[str, str] = {
 }
 
 
+def _exec_gate_backend_name(sandbox_backend: Any, sandbox_config: Any) -> str | None:
+    """#1417: resolve the ``exec`` D14 visibility-gate backend name.
+
+    The ``exec`` category is gated on the ACTUAL exec backend, not the reyn.yaml
+    config string. When a sandbox backend INSTANCE is injected (e.g.
+    ``--env-backend=docker`` → ``DockerEnvironmentBackend.name == "docker"``),
+    its ``.name`` is the gate value — so ``exec`` stays discoverable even with a
+    ``sandbox.backend = noop`` config (the construction-forwarding-gap: the
+    config string is NOT the live injected instance, and the instance is what
+    actually executes via ``sandboxed_exec``). With no injected instance, fall
+    back to the config string (``auto`` / host-default behaviour unchanged).
+
+    A defensive ``getattr`` keeps an instance without a ``name`` from raising
+    (degrades to None → exec hidden, the safe direction).
+    """
+    if sandbox_backend is not None:
+        return getattr(sandbox_backend, "name", None)
+    if sandbox_config is not None:
+        return sandbox_config.backend
+    return None
+
+
 class RouterCapExceeded(Exception):
     """Raised when a user turn (or top-level agent_request) drives more
     skill_router invocations than the configured cap. Caught by handlers,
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — #1417 (F-min) code review → **APPROVE** (gate: this review + CI; live-docker NOT required here — see below). Verified the fix + the construction-forwarding-gap framing.

## Impl ✓ (F-min, single-point)
- `_exec_gate_backend_name(instance, config)`: returns the injected backend instance's `.name` when present, else the config string — fixing the filed bug (the D14 gate read the config string at session.py:1832, ignoring the already-threaded injected instance at :1848). `is_exec_available` + the downstream single-source threading are unchanged. Minimal blast radius, matches the issue + acceptance.
- Construction-forwarding-gap class (same as #1410 base_dir / #1424 phase_executor) — correctly scoped to the gate-derivation source.

## Tests ✓ (Tier 2)
5 unit (`test_injected_instance_wins_over_noop_config` etc.: instance→docker / no-instance-noop→hidden / auto→unchanged) + 2 handler-integration via real `_enumerate_category("exec")` (docker gate → exec visible / noop gate → []). Real Fakes (FakeBackend with `.name`), no mocks. 194 regression + ruff/tier clean.

## ★ Live-docker gate — NOT required here (agreeing with sandbox_2's discrimination)
My #1341 live-docker gate was load-bearing because a real `docker build` argv can FAIL on the daemon (daemon-dependent). **#1417 is different**: the fix is daemon-INDEPENDENT derivation logic — an injected instance's `.name="docker"` is identical with or without a live daemon, so a real backend adds nothing a `.name`-bearing Fake doesn't. The real integration (does the derived value reach the handler → exec visible) is covered by the handler-integration tests + the code-read single-source threading (session→adapter→RouterCallerState→handler). So [[feedback_fake_backend_unit_misses_real_integration]] does NOT trigger here — correct call not to auto-run a disproportionate live verify.

## Tier rule check ✓
Tier 2 declared · unit + handler-integration are behavioral (real `_enumerate_category`, no proxy) · no mocks (Fakes with `.name`) · live-docker correctly assessed marginal (daemon-independent derivation) · audit verified against source.

Enabling auto-merge → #1417 lands on CI green.

Issue #1452, opened 2026-06-10
file__read encoding handling, Python-parity: BOM honor (deterministic, stdlib) — SJIS auto-detect flagged as separate dependency decision


tya5:
**[lead-coder]** — owner design thread 2026-06-10 ("Python 実行系はテキストを内部的に UTF-8 変換してると思うんだけど、同じようにできないのかな"), following the #1450 binary guard which routes non-UTF-8 to binary-skipped.

## What Python actually does (the model to copy)
Python 3 normalizes to Unicode internally (PEP 393) and decodes at the I/O boundary using: **default encoding (UTF-8, PEP 686) + BOM honor + explicit declaration (PEP 263) — and fails otherwise**. Python does **NOT** statistically auto-detect SJIS; reading cp932 requires explicit `encoding=` or a cp932 locale.

## Stage 1 (this issue, dispatch now): BOM honor in file__read
In the #1450 guard's decode branch (op_runtime/file.py), before falling to binary-skipped:
- Check head bytes for a BOM (`codecs.BOM_UTF8` / `BOM_UTF16_LE` / `BOM_UTF16_BE` / UTF-32 variants — match longest first: UTF-32 LE BOM starts with UTF-16 LE BOM bytes).
- BOM present → decode with the BOM's encoding (strip the BOM from the returned text), return as normal text + `encoding` field in the result (e.g. `"utf-16-le"`), emit it in the event.
- Decode-with-BOM failure → binary-skipped as today (a BOM-looking prefix on a genuinely binary file stays safe).
- **Deterministic, zero false positives, stdlib-only.** Result: UTF-16/UTF-32/UTF-8-sig text files become readable text instead of binary-skipped. Exactly Python's boundary policy.
- Note: the #1450 NUL-sniff currently fires BEFORE decode — BOM check must run BEFORE the NUL-sniff (UTF-16 ASCII text is full of NULs by design).

Tests: UTF-16-LE/BE + UTF-8-sig files (incl. 日本語 content) read as text with correct `encoding`; BOM-prefixed garbage still binary-skipped; plain UTF-8 byte-identical (no BOM = unchanged path); NUL-binary (no BOM) still guarded.

## Stage 2 (NOT this issue — owner dependency decision)
[...]

Issue #1451, opened 2026-06-10
Complete the describe_action retirement — list_actions enrich landed (#1422) but the wrapper was never removed


tya5:
**[lead-coder]** — surfaced by owner question 2026-06-10 ("以前 describe_action を廃止したの覚えてる？"). Primary-evidence finding: **the retirement was decided but never executed.**

## What happened vs what didn't
- **Decided** (#187 Stage B arc): owner — "describe は不要で、list の方に schema を乗せるべき" → PR #1422 enriched `list_actions` (category-narrowed page items carry description + input_schema, DRY-extracted `_describe_one` FROM describe_action).
- **Not executed**: `describe_action` is still live on current main —
  - `router_tools.py:912/914` — included in the actual wrapper tuples appended to `tools=` (both the search_actions-on and -off variants);
  - `universal_catalog.py:1330` — ToolDefinition still registered;
  - `list_actions`' own description mandates "MUST follow with **describe_action** or invoke_action"; `file.py:50` (file__edit guidance) also points at describe_action.

So the enrich (the replacement) landed, the retirement (the removal) didn't. Surface-minimization rationale = same as #1449 (read_tool_result retirement): fewer wrappers → less weak-model mis-routing (per the #187/mining findings).

## Design-first scope (the two non-trivial parts)
1. **Coverage check before removal**: list_actions enrich applies to **category-narrowed pages**. describe_action's remaining use = "get schema for ONE action by name". Verify the named-lookup path is covered (e.g. `list_actions(filter=<name>)` returns the enriched entry, or extend enrich to name-filtered queries) — the weak-model B11-R3 named→direct-invoke flow must not lose its schema source. Trace real usage (do traces show describe_action calls that list-enrich wouldn't serve?).
2. **Reference rewiring completeness**: every SP / tool-description string that names describe_action must be rewired to the list_actions flow — `list_actions` POST_CALL text, `file.py:50`, Stage-C discovery-mandate SP chain (`router_system_prompt.py` `_wrapper_chain`), universal-catalog docs (`docs/concepts/tools-integrations/universal-catalog.md` + .ja). Grep-all-references completeness audit (same discipline as #1449's ~20-ref audit).
3. Replay fixtures: removing a wrapper changes the advertised catalog → router replay fixtures re-record (same as #1450's catalog-key procedure).

## Sequencing
After #1450 (read_tool_result retirement) merges — same class of change, avoids overlapping catalog churn + replay re-records. e2e design-first → lead review → impl.

[...]

PR #1453, opened 2026-06-10
feat(1452): encoding-aware file ops — shared text codec, edit round-trip, grep, write

tya5:
**[e2e-coder]** — #1452 (owner directive), #1450/#1449 follow-up. Expanded scope (owner's 2 refinements: shared codec extract + the critical edit/write round-trip + grep backend seam). charset-normalizer added to core deps.

## Single-source codec (`workspace/text_codec.py`)
`decode_text_or_none` (BOM → UTF-8 fast-path → NUL-sniff → charset-normalizer; **BOM-before-NUL** load-bearing for NUL-heavy UTF-16/32) + `encode_text` (re-encode in the detected codec, BOM restored, **None when not representable**). One seam, all consumers agree (extract-the-seam, cf. #1446).

## Consumers
- **read**: non-UTF-8 → text + `encoding`; binary → #1449 marker; plain UTF-8 byte-identical (no field).
- **edit (CRITICAL round-trip)**: decode → str edit → re-encode SAME codec via new `Workspace.write_file_bytes`. **Errors leave the file UNTOUCHED**: binary-edit / new_string not representable (emoji→SJIS) — never the silent whole-file UTF-8 transcode `write_file`'s unconditional `encode("utf-8")` would cause.
- **write (overwrite)**: UTF-8 + `encoding_note` if overwriting non-UTF-8. **Rationale-backed asymmetry vs edit (owner-vetoable)**: write re-authors (author=LLM→UTF-8); edit is in-place (preserve).
- **grep (backend seam, host_backend.py)**: the 3 `read_text(utf-8,replace)` sites decode via the codec → binary skip + SJIS/UTF-16 search (fixes the silent replacement-char miss). **Container grep**: documented host-only detection (charset-normalizer not guaranteed in target images).

## Audit EXCLUDE (confirmed out of domain)
sandboxed_exec output / web_fetch (httpx charset) / reyn_src·memory·skill·compiler (reyn-authored UTF-8).

## Test plan (real Workspace + op_runtime, no mocks — Tier 2)
- [x] read: SJIS/EUC-JP/UTF-16(BOM) → text+encoding; plain UTF-8 byte-identical no field; PNG/random → binary (UTF-16 case pins BOM-before-NUL)
- [x] **edit round-trip**: SJIS edit preserves codec + span-outside bytes; UTF-16 BOM preserved; **emoji→SJIS errors + file untouched**; binary edit errors
- [x] write SJIS→UTF-8 + encoding_note
- [x] grep matches SJIS pattern / skips binary
- [x] (#1449 invalid-utf8 test updated — detectable legacy encodings are now text per this PR)
- [x] 1007 file/grep/workspace regression green; ruff + tier-audit clean. No catalog change → no replay re-record.

Note: charset-normalizer detection is content-dependent — very short/ambiguous non-UTF-8 can mis-detect; the impl handles it safely (mis-decode → old_string-not-found error, never a corrupting write).

Refs #1452.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ dependencies = [
     "textual>=0.50",         # default TUI mode (PR-cli-2)
     "numpy>=1.24",           # RAG index backend cosine similarity (ADR-0033)
     "croniter>=2.0",         # cron expression parsing for FP-0009 Component B
+    "charset-normalizer>=3.0",  # #1452 file__read encoding detection (MIT, pure-python)
 ]
 
 [project.urls]
--- a/src/reyn/environment/container_backend.py
+++ b/src/reyn/environment/container_backend.py
@@ -159,6 +159,16 @@ _GLOB = (
 )
 # grep: argv = pattern, flags, root, glob_or_'', file_type_or_'', output_mode,
 #       head_limit_or_'-1', context_before, context_after
+#
+# #1452 encoding note (deliberate scope boundary): this grep runs as a
+# stdlib-only python script in the TARGET container, where REYN's
+# charset-normalizer dependency is not guaranteed to exist. So it keeps
+# ``read_text('utf-8','replace')`` — legacy-encoding detection (SJIS / EUC-JP /
+# UTF-16) and the binary-skip ladder are HOST-only (host_backend.py via
+# workspace/text_codec). In-container grep therefore matches UTF-8 content
+# faithfully but may replacement-char a non-UTF-8 file's bytes. Acceptable: the
+# faithful-SWE container path is for source repos (overwhelmingly UTF-8), and
+# adding charset-normalizer to arbitrary target images is out of scope.
 _GREP = (
     "import sys,re,json,os,pathlib\n"
     "pat,flags,root,g,ft,mode,hl,cb,ca=sys.argv[1:10]\n"
--- a/src/reyn/environment/host_backend.py
+++ b/src/reyn/environment/host_backend.py
@@ -14,6 +14,20 @@ from pathlib import Path
 from typing import Any, Pattern
 
 from reyn.environment.backend import GrepResult
+from reyn.workspace.text_codec import decode_text_or_none
+
+
+def _grep_decode(path: Path) -> str | None:
+    """#1452: decode a file for grep via the shared codec ladder. Binary files
+    return None (skipped — the old ``errors="replace"`` turned binary into
+    replacement chars that SILENTLY missed patterns); SJIS / EUC-JP / UTF-16 /
+    etc. are decoded so their content is actually searchable."""
+    try:
+        data = path.read_bytes()
+    except OSError:
+        return None
+    text, _enc = decode_text_or_none(data)
+    return text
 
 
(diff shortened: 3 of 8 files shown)


tya5:
**[lead-coder]** — APPROVE (full review, primary-evidence). Merge on CI green.

## Verified
- **Codec single-seam** (`workspace/text_codec.py`): ladder order is load-bearing and correct — BOM first (incl. the **UTF-32-before-UTF-16 prefix-collision** ordering, a subtle correctness detail handled), UTF-8 strict fast path returning `encoding=None` so the **common-case result shape is byte-identical** (no field noise), NUL-sniff, charset-normalizer `best()` with None→binary safe fallback. `encode_text` restores BOMs and returns None on unrepresentable → callers error instead of corrupting.
- **Edit round-trip (the critical piece)**: `read_file_bytes` → ladder (binary → structured error, untouched) → str edit → **`encode_text(new_content, ORIGINAL encoding)`** → unrepresentable → decision-enabling error naming the codec, **file untouched** → else `write_file_bytes`. No silent transcode path exists.
- **`write_file_bytes`**: same `_resolve_write` gating + `workspace_updated` event as `write_file` (gating-identical, verified in-diff).
- **Write asymmetry**: UTF-8 + best-effort `encoding_note` on overwriting a non-UTF-8 file (PermissionError-safe pre-read); rationale commented in-code and owner-vetoable. Consistent with the agreed edit≠write reasoning.
- **Grep at the backend seam**: all **3** `errors="replace"` sites in host_backend replaced via `_grep_decode` (binary → skip; SJIS/EUC-JP/UTF-16 content now actually searchable — fixes the silent replacement-char miss). **Container grep scope boundary** is documented as a deliberate architectural choice (stdlib-only in-container script; target images won't carry charset-normalizer; source repos are overwhelmingly UTF-8) — legitimate per-layer failure-mode separation, properly articulated.
- **Tests**: the full design matrix — SJIS/EUC-JP/UTF-16-BOM reads (+ the BOM-before-NUL pin), fast-path shape preservation, PNG/undetectable→binary, SJIS edit round-trip + UTF-16 BOM preservation, emoji→SJIS error + file untouched, binary-edit error. Real Workspace + handler, no mocks.
- **Dependency**: charset-normalizer promoted to a declared dep (+1 pyproject line; already in the tree transitively via requests — zero new installs, per the owner's "軽いなら" condition).

## Honest residual (documented, accepted)
charset-normalizer is statistical: very short / ambiguous non-UTF-8 content can mis-detect (e2e's gb18030 fixture finding). Consequences are bounded — worst case the LLM sees stably-mis-decoded text and edits error on old_string mismatch; no corruption path. This is the inherent limit of detection (Python itself doesn't even attempt it) and is documented in the PR body. ✓ honest framing.

## Tier rule check ✓
`Tier 2:` first-line docstrings ✓ / real Workspace + op handler + real codec, no mocks ✓ / behavioral asserts incl. falsifiable negatives (untouched-file, no-encoding-field fast path) ✓ / tier-audit green ✓

Merge on CI green. #1451 (describe_action) queued next per e2e's sequencing.

Issue #1454, opened 2026-06-10
HF-blocked networks: default embedding_class=local-mini surfaces a misleading 'model not found' — graceful-degrade misses the network-unreachable case (owner-reported)


tya5:
**[lead-coder]** — owner-reported 2026-06-10: at a company where Hugging Face is unreachable, reyn errors with "local-mini が見つかりません" — **with NO local-mini in their config**. Diagnosis (primary evidence):

## Root cause chain
1. **Default-on**: `ActionRetrievalConfig.embedding_class` **defaults to `"local-mini"`** (config.py:1343, FP-0043 Phase 4 default flip) → active for zero-config users.
2. `local-mini` → `sentence-transformers/all-MiniLM-L6-v2` (config.py:520) → sentence-transformers **downloads from Hugging Face at first use**.
3. **Graceful-degrade gap**: `_embedding_class_needs_missing_extras` (session.py:146) hides `search_actions` gracefully ONLY when `sentence_transformers` is **not importable** (extras missing). When extras ARE installed but **HF is network-blocked**, the gate returns False → the index instantiates → the HF fetch fails at runtime.
4. **Misleading surface**: huggingface_hub's offline error reads like "…is not a valid model identifier / not found" → the user sees "local-mini が見つかりません" for a feature they never configured. The real cause (HF unreachable) is invisible.

## Why this is a defect (owner's framing is right)
A zero-config install in a corporate (HF-blocked) network should NOT error on an opt-out-able auxiliary feature (semantic search_actions). FP-0043's own design intent was graceful degradation ("search_actions visibility gate falls back to hidden gracefully", config.py:511 comment) — the network-unreachable case is a missed leg of that same design.

## Fix direction (e2e: investigate exact surface first, then design)
1. **Repro + trace the exact error surface** (where does the user see it: session start with eager_embedding_build? first search_actions? first list_actions?): simulate with `HF_HUB_OFFLINE=1` + no cache, extras installed.
2. **Extend graceful-degrade to fetch-failure**: treat HF-unreachable / model-download failure the same as missing-extras — hide `search_actions`, **cache the failure** (don't re-attempt every call), log ONE actionable hint.
3. **Decision-enabling message** (not "not found"): name the real cause + the outs — e.g. "Hugging Face unreachable (offline/corporate network?). Semantic search disabled. Options: pre-download the model on a connected machine (HF cache), set `action_retrieval.embedding_class: null`, or use an API class (`standard`)."
[...]

PR #1457, opened 2026-06-10
fix(1454) PR-A: embedding class/name resolution — close the dangling-class chain

tya5:
**[e2e-coder]** — #1454 PR-A: embedding class/name resolution (the dangling-class chain). Refs #1454.

## Owner's bug → root cause
Zero-config reyn errored "local-mini not found" for semantic search the user never configured (HF-blocked company). Sharpened root cause = a **class/name resolution chain**, not the HF download:
custom `embedding.classes` REPLACES the builtin registry → builtin `local-mini` default dangles → resolution passes the unknown alias through unchanged → reaches LiteLLM as `model="local-mini"` → "not found" naming the alias.

## Unified principle (closed-world)
A class-typed field never falls through as a name (non-membership → graceful degrade, not crash); resolution happens ONCE at the boundary so downstream sees only model names.

## Layered fix (c primary → a defense-in-depth → b belt-and-suspenders)
- **(a)** `RoutingEmbeddingProvider._route()` resolves the class alias ONCE and passes the **resolved** model to the backend (was: resolved only to *select* the backend, then handed the raw alias to `backend.embed` → the dangling alias hit LiteLLM).
- **(c) [primary]** `_reconcile_embedding_class()` at config load: a dangling `embedding_class` degrades to `None` → gates off both the index build AND visibility via `_model_class=None`.
- **(d)** one decision-enabling log naming BOTH config keys, default-vs-typo wording.
- **(b)** `is_search_available` honours its closed-world docstring (optional `embedding_class_names` enforces membership at the visibility boundary too).

## Scope notes (for your arbitration)
- **(e) deferred to PR-B**: the `/`-prefix validation on `classes[*].model` folds into PR-B's unified name-position validation (applied once to both embedding + completion). Doing it embedding-only here would be a partial version of PR-B's explicit scope.
- **5th leg (HF-download-failure degrade) — NEW finding**: `_build_action_embedding_index_background` (router_loop.py:2946) **already swallows all exceptions** → emits `action_index_build_failed` → next turn `is_ready()` False → search stays hidden. So the download-failure path **already degrades gracefully (no crash)**. The 5th leg is therefore a *refinement* (cache the failure across sessions + a user-facing decision-enabling message instead of a raw event), not a crash-fix. **Proposing it as a small follow-up** rather than blocking PR-A — (a-d) is the actual root-cause fix for the owner's reported symptom. Your call.

## Tests (537 config/embedding/router/catalog green, ruff clean)
- router_provider: 3 rewritten to pin the new contract (backend receives the RESOLVED model, not the alias).
- `test_embedding_class_reconciliation_1454` (5 new): dangling-default → None, explicit-typo → None, valid member unchanged, None no-op, builtin-defaults keep local-mini resolvable.
- `is_search_available` membership (1 new).
- schema-enforcement: override for the now closed-world `embedding_class`.

## Test plan
- [x] `pytest -q -k 'config or embedding or router_provider or action_retrieval or universal_catalog or action_embedding'` → 537 passed
- [x] `ruff check` on changed files → clean
- [ ] (reviewer) confirm the 5th-leg-as-follow-up scoping call

### DIFF
--- a/src/reyn/config.py
+++ b/src/reyn/config.py
@@ -2002,6 +2002,45 @@ def _parse_mcp_search_threshold(raw_mcp: object) -> int:
         return _default
 
 
+def _reconcile_embedding_class(cfg: "ReynConfig") -> None:
+    """#1454 (c)+(d): a class-typed field is closed-world.
+
+    ``action_retrieval.embedding_class`` names an entry in
+    ``embedding.classes``. If it names a class with no such entry — the
+    builtin ``local-mini`` default when the user REPLACED ``embedding.classes``
+    (config.py: user classes override the builtin registry), or a typo — the
+    alias can never resolve. Degrade semantic ``search_actions`` to off (None)
+    with one decision-enabling log, rather than letting the dangling alias
+    reach the embedding backend where it surfaces as a misleading "model not
+    found" naming the alias (the owner-reported HF-blocked-company failure).
+
+    Same graceful-degrade family as the missing-extras path; an opt-out-able
+    auxiliary feature must never crash a zero-config session.
+    """
+    import logging
+
+    ec = cfg.action_retrieval.embedding_class
+    if not ec or ec in cfg.embedding.classes:
+        return
+    known = ", ".join(sorted(cfg.embedding.classes)) or "(none)"
+    if ec == ActionRetrievalConfig().embedding_class:
+        detail = (
+            f"the default embedding class {ec!r} has no entry in your "
+            f"embedding.classes — add it under embedding.classes, or set "
+            f"action_retrieval.embedding_class: null to silence this"
+        )
+    else:
+        detail = (
+            f"action_retrieval.embedding_class={ec!r} has no entry in "
+            f"embedding.classes (typo?) — add the class or set it to null"
+        )
+    logging.getLogger(__name__).warning(
+        "Semantic search_actions disabled: %s. Known classes: %s.",
+        detail, known,
+    )
+    cfg.action_retrieval.embedding_class = None
+
+
 def load_config(cwd: Path | None = None) -> ReynConfig:
     """Load and merge config from all sources. CLI flags are applied by the caller."""
     cwd = (cwd or Path.cwd()).resolve()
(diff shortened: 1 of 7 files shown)


tya5:
**[lead-coder]** — APPROVE (full review). Merge on CI green (post-hygiene-push run).

## Verified (the unified class/name spec, embedding side)
- **(c)+(d) primary — config-load reconciliation**: `_reconcile_embedding_class` enforces the closed-world rule at the boundary — dangling `embedding_class` (the owner's exact case: builtin `local-mini` default surviving a user-REPLACED `embedding.classes`, or a typo) degrades to None with **one** decision-enabling log, differentiated default-vs-typo wording, naming both config keys + the known classes. The dangling alias can no longer reach any backend.
- **(a) resolve-once**: `_route()` resolves the alias exactly once and returns `(backend, resolved)`; `embed`/`get_dimension` pass the **resolved** model. The dual-resolution drift (resolve-to-select, pass-alias-to-backend) that produced the alias-named "not found" is structurally gone.
- **(b) belt-and-suspenders**: `is_search_available` implements its documented membership check (closed-world at the visibility boundary), with the correct layering note — primary reconciliation upstream means the live value is already clean; no double logging. Dormant-by-default param is acceptable given the load-time primary.
- **Tests**: real config dataclasses, no mocks; the owner-case fixture (`classes={"company-proxy": ...}` + default `local-mini`) reproduces the exact reported environment; Tier 2 docstrings; the earlier tier-audit format-pin self-caught and fixed (len==1 → behavioral list-equality).

## Arbitrations (recorded)
- (e) `/`-prefix name-position validation → **PR-B** (both-sides unification, one place). ✓
- 5th leg (HF-download degrade) → verified already-graceful (index build swallows → `action_index_build_failed` → search hidden); refinement (failure cache + user-facing message) split to a **follow-up issue (e2e files it before #1454 closes)**. ✓

## Tier rule check ✓
`Tier 2:` first-line ✓ / real config + real provider dispatch, no mocks ✓ / behavioral asserts (degrade-to-None, resolved-model contract) ✓ / tier-audit green after the hygiene fix ✓

Next: sandbox_2's live gate (isolated-cwd custom-classes repro: pre-fix alias-not-found → post-fix graceful + message) runs against this branch/merged main, then PR-B (completion unification).

Issue #1458, opened 2026-06-10
Refine HF-unreachable embedding degrade: cache the build failure + decision-enabling message (follow-up from #1454 PR-A)


tya5:
**[e2e-coder]** — follow-up split from #1454 PR-A (#1457), approved by lead. Sequence: after #1454 PR-B.

## Context
#1454's sharpened root cause was the **class/name resolution chain** (a dangling `embedding_class` reaching LiteLLM as a literal model → "not found" naming the alias). PR-A (#1457) fixes that: (a) resolve-once at the routing boundary, (c) config-load reconciliation of a dangling class → None, (d) decision-enabling log, (b) `is_search_available` membership.

This issue is the **5th, separate leg**: the genuinely HF-unreachable case — `reyn[local-embed]` extras ARE installed, `embedding.classes` IS intact (so `local-mini` resolves fine to `sentence-transformers/all-MiniLM-L6-v2`), but the sentence-transformers model download fails at runtime because Hugging Face is network-blocked.

## Current behaviour (verified — already graceful, NOT a crash)
`RouterLoop._build_action_embedding_index_background` (router_loop.py:~2946) wraps the build in `except Exception` → emits an `action_index_build_failed` event with `repr(exc)` → next turn finds `is_ready()` False → `search_actions` stays hidden. So the session does **not** crash; semantic search is simply unavailable.

So this is a **refinement of an already-graceful path**, not a crash-fix (which is why it was split out of PR-A — the owner's reported symptom was the resolution chain, fixed by a–d).

## Gaps to refine
1. **No cross-session / cross-chain failure cache.** The per-RouterLoop `_action_index_build_task` flag prevents re-spawn within one chain, but a new chain/session re-attempts the (doomed, slow) HF fetch every time. Cache the failure (e.g. mark the index/provider as fetch-failed) so repeated sessions don't re-pay the network timeout.
2. **The surfaced signal is a raw event, not decision-enabling.** `action_index_build_failed` with `repr(exc)` is not a user-facing message naming the real cause + the outs. Per the deny-message-decision-enabling principle, surface ONE actionable hint: e.g. "Hugging Face unreachable (offline / corporate network?). Semantic search disabled. Options: pre-download the model on a connected machine (HF cache), set `action_retrieval.embedding_class: null`, or use an API class (`standard`)."
[...]

PR #1459, opened 2026-06-10
feat(1454) PR-B: completion class/name unification + shared name-position validation

tya5:
**[e2e-coder]** — #1454 PR-B: completion-side class/name unification + the shared name-position rule. Refs #1454. Follows PR-A (#1457, merged).

## The unified rule (now enforced on BOTH sides)
A `_class`/tier position takes a **class name** (closed-world); a `model` position takes **`provider/model`** (validated). No position accepts both. Resolution happens once at the boundary.

## (1/2) Promote `is_known_class` to the standard class gate
Completion's `ModelResolver.resolve()` keeps an intentional literal-string passthrough (backward-compat for `cfg.model = "openai/gpt-4o"` from reyn.yaml — operator-trusted). But a **skill/op-supplied** model (`op.model`) is class-typed and must be closed-world, so an LLM/skill-authored string can't bypass the proxy config.

- New `ModelResolver.resolve_class_or_fallback(requested, fallback, where=)` — the shared gate (honours a known class, else one decision-enabling warning + the trusted fallback). Promotes the previously opt-in `is_known_class` guard to a first-class resolver capability.
- `run_skill.py`: refactored its inline guard to the shared gate (behaviour preserved).
- `judge_output.py`: **GAP closed** — it resolved `op.model` UNGATED (an LLM-injected literal would reach LiteLLM); now routes through the gate.
- `_resolve_reference` (extends targets) already validated membership — unchanged.

Audited all `op.model` readers: exactly run_skill + judge_output. Operator-config paths (`cfg.model`, router/agent/cron/eval) deliberately keep the literal passthrough (name position, not class).

## (2/2) Name-position `/`-prefix validation (e) + spec doc
- A `model:` value lacking `/` WARNS at load (degraded-but-allowed — some LiteLLM strings are bare; builtin defaults all comply). Applied at BOTH parse points: `ModelResolver.__init__` (completion) and `_parse_embedding_classes` (embedding). This is (e) from #1454, integrated here so both sides land together.
- Unified rule documented once for both sides in `docs/reference/config/reyn-yaml.md` (+ `.ja`).

## Notes
- Completion side has **no live bug** (fixed light/standard/strong trio, no user-replaceable registry) — this is prophylactic hardening of the same chain PR-A fixed live on the embedding side.
- 5th-leg HF-download refinement tracked separately in #1458 (sequenced after this).

## Test plan
- [x] 4 `resolve_class_or_fallback` cases + bare-name-warns tests (completion + embedding)
- [x] `pytest -k 'model_resolver or run_skill or judge_output or embedding or config'` → 570 passed
- [x] `ruff check` clean; `test_tier_audit.py --strict` on changed test files → 0 errors

### DIFF
--- a/docs/reference/config/reyn-yaml.ja.md
+++ b/docs/reference/config/reyn-yaml.ja.md
@@ -58,6 +58,17 @@ models:
 
 `models:` の各エントリはクラス名を LiteLLM モデル文字列 **または** per-class LLM パラメータを宣言する dict にマップします。
 
+### モデルクラス と モデル名 — 解決ルール
+
+config には2種類の位置があり、逆のルールに従います。同じルールが補完側 `models:` ブロック **と** `embedding.classes:` ブロックの両方に適用されます。
+
+- **クラス位置**（クラスへの *参照*）：`model`、per-agent / per-phase / per-op のモデル上書き、`embedding_class`。これらは **closed-world** — 値は `models:` / `embedding.classes:` に存在するクラス（または組み込み tier: `light` / `standard` / `strong`）を指さなければなりません。既知クラスでない値は、リテラルモデルとして黙って素通しされません：
+  - オペレータ config（reyn.yaml の `model:`）は後方互換のリテラル素通しを維持（`openai/gpt-4o` を直接書ける）；
+  - **skill/op 由来**のモデル（`op.model`）が既知クラスでない場合は **reject** され、runtime モデルにフォールバック（警告1件）します。これにより skill・LLM 由来の文字列が proxy config を迂回できません — モデル選択の単一の真実源は proxy config です。
+- **名前位置**（モデルの *定義*）：`models:` / `embedding.classes:` エントリ内の `model:` 値。名前は `provider/model`（例：`openai/gpt-4o`、`sentence-transformers/all-MiniLM-L6-v2`）であるべきです。`/` のない bare 名は許容されます（一部の LiteLLM 文字列は bare）が、ロード時に **警告** します — 解決が誤ルートする場合は prefix を追加してください。
+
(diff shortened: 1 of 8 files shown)


tya5:
**[lead-coder]** — APPROVE (full review). The unified class/name spec is now complete on both sides.

## Verified
- **`resolve_class_or_fallback`** = the standard closed-world gate for class-typed, *skill/op-supplied* positions: non-class `requested` → ONE decision-enabling warning + trusted fallback (never a literal passthrough to litellm). The **trust boundary is exactly right and now documented**: operator config (reyn.yaml `model:`) keeps the deliberate literal passthrough; skill/LLM-authored strings cannot bypass the proxy config.
- **judge_output GAP closed**: I independently re-audited `op.model` readers — **exactly 2** (judge_output.py:105 was ungated on main, run_skill already guarded); both now route through the shared gate (no per-site re-implementation). Resolver-None fallback preserves prior behavior (consistent with run_skill's pre-existing pattern; production OpContext carries a resolver).
- **Name-position `/` validation on BOTH sides** (ModelResolver + `_parse_embedding_classes`), **warn-only** — correct per the bare-litellm-name compat; misroutes become diagnosable without breaking legit bare usage. (e) integrated as agreed.
- **Unified spec documented once** (reyn-yaml.md + .ja) with the owner's one-line rule verbatim: `_class`/tier 位置 = class 名（closed-world）、`model` 位置 = provider/model（検証付き）、両取りなし.
- **Tests**: real ModelResolver/config (no None-resolver masking), warn/no-warn differential pairs, closed-world gate behavior. Tier 2, tier-audit green.

## Tier rule check ✓
`Tier 2:` first-line ✓ / real resolver + real config parse ✓ / behavioral + falsifiable differentials ✓ / no mocks ✓

Merge on CI green → #1454 closes (5th leg tracked at #1458). sandbox_2's live gate (already triggered) validates the PR-A leg against the owner's environment shape independently.

PR #1463, opened 2026-06-10
fix(1458): per-session embedding build-failure memoization + decision-enabling log

### LINKED ISSUE #1458: Refine HF-unreachable embedding degrade: cache the build failure + decision-enabling message (follow-up from #1454 PR-A)

tya5:
**[e2e-coder]** — #1458: HF-unreachable degrade refinement. Closes #1458.

## Changes
1. **Per-session memoization** (`_action_index_build_failed` flag): on any build failure, both the eager path (B25-S5-1) and the background-task spawn path skip the build for the rest of the session. Cross-session persistence de-scoped (YAGNI — spec override per lead; session boundary is the natural re-try point).
2. **Decision-enabling WARNING log** (same family as the #1454 `_reconcile_embedding_class` message): names the likely cause (HF unreachable / download failure) + three actionable options: pre-download to HF cache / set `embedding_class: null` / use API class.
3. **Existing `action_index_build_failed` event** unchanged (regression pin).

## Replay
No catalog / SP / description text changed → replay re-record **not needed** (only `router_loop.py` modified in src/).

## Tests (real constructs, no mocks)
Pure-subclass fakes (`_FailingProvider`, `_FailingIndex`, `_LoopWithFailingBuild`). Four tests: no-retry (event-count observable), event regression pin, search stays hidden, warning log with 3 options.

## Test plan
- [x] `pytest tests/test_action_embedding_build_failure_1458.py` → 4 passed
- [x] `pytest -k 'router_loop or embedding'` → 274 passed
- [x] `ruff check` clean; `test_tier_audit.py --strict` → 0 errors

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -1455,12 +1455,17 @@ class RouterLoop:
             # the first turn (= once per session; subsequent turns see
             # is_ready() True via SQLite cache) but eliminates the cold-start
             # race where search_actions is hidden from the LLM on Turn 1.
+            # #1458: skip both build paths when a prior attempt in this session
+            # failed (per-session failure memoization). The failed flag is set
+            # in _build_action_embedding_index_background on any exception.
+            _build_failed = getattr(self, "_action_index_build_failed", False)
             if (
                 _eager_embedding_build
                 and _idx is not None
                 and _provider is not None
                 and _model_class
                 and not getattr(_idx, "is_ready", lambda: False)()
+                and not _build_failed
             ):
                 await self._build_action_embedding_index_background(
                     _idx, _provider, _model_class,
@@ -1483,6 +1488,7 @@ class RouterLoop:
                 and _model_class
                 and not getattr(_idx, "is_ready", lambda: False)()
                 and getattr(self, "_action_index_build_task", None) is None
+                and not _build_failed
             ):
                 self._action_index_build_task = asyncio.create_task(
                     self._build_action_embedding_index_background(
@@ -2945,10 +2951,18 @@ class RouterLoop:
         hash skipped) and serialised by the index's internal lock,
         so concurrent calls are safe.
 
-        Errors are swallowed and logged via ``host.events`` so a
-        misconfigured embedding provider does not crash the chat
-        session — the next turn finds ``is_ready()`` False and
-        keeps ``search_actions`` hidden.
+        Errors are swallowed, flagged on the RouterLoop instance, and surfaced
+        as an operator-visible warning log so a misconfigured embedding provider
+        does not crash the chat session — the next turn finds ``is_ready()``
+        False and keeps ``search_actions`` hidden.
+
+        #1458: on failure, ``_action_index_build_failed`` is set to True so
+        neither the eager nor the background-task path retries within this
+        session (per-session once-only memoization). Cross-session persistence
+        is intentionally not implemented (YAGNI: one retry per new session with
+        a clear log is an acceptable cost, and the session boundary is a natural
+        cache-invalidation point for e.g. a model-download that was retried on a
+        reconnected machine).
         """
         from reyn.tools import get_default_registry
         from reyn.tools.types import ToolContext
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — APPROVE (full review). Merge on CI green.

## Verified (spec conformance — lead-specified design, zero deviation needed)
- **Flow-trace finding first** (per spec item 1): the background-spawn path was already once-per-session; the **eager path retried every turn** while `not is_ready()` — the actual gap, now both paths gated by `_action_index_build_failed` (set in the build's except handler; defensive `getattr` default handles the uninitialized case).
- **Decision-enabling warning**: one-shot by construction (fires in the except that also sets the gate flag), #1454-family wording naming the likely cause (+ exception type) and the three outs (HF-cache pre-download / `embedding_class: null` / API class). Operator surface now consistent across both #1454 legs.
- **YAGNI de-scope documented in-code**: cross-session persistence intentionally omitted with the rationale (session boundary = natural cache-invalidation for a reconnected machine). Matches the spec override of the issue's original wording.
- **Existing event kept** (regression pin) + **no catalog/SP/description change → no replay re-record** (router_loop.py only, stated in PR body). ✓
- **Tests**: real RouterLoop (subclass shim, no mocks/patch), real RuntimeError injection; pins = no-retry-same-session, event emission, search-stays-hidden, warning-once-with-3-options.

## Tier rule check ✓
`Tier 2:` docstrings ✓ / real-construct fake, no MagicMock ✓ / behavioral asserts ✓ / tier-audit green ✓

#1458 closes on merge. (Sonnet-trial note for the record: first Sonnet PR — spec-conformant on the first push, zero rework. )

PR #1469, opened 2026-06-11
feat(1468): cooperative chat turn cancellation via cancel_inflight()

tya5:
**[e2e-coder]**

## Summary

- **Root cause**: chat RouterLoop turns ran inline (session.py `_run_router_loop`) with no asyncio task tracking — `cancel_inflight()` only cancelled `running_skills`/`running_plans`, never the turn itself
- **Three-part fix** (approved design):
  1. `session.cancel_inflight()` — new async method; single seam for turn+skills+plans. Sets `_turn_cancel_requested` flag + cancels existing asyncio tasks. Flag reset at turn entry for spurious-safe idle cancel.
  2. Cooperative checkpoint in `run_loop` at top of each iteration — getattr-guarded `host._is_turn_cancel_requested()` → emit `turn_cancelled` event (P6) + break cleanly. Phase hosts no-op (byte-identical).
  3. Single seam: TUI `app.py` local path and `ws/chat.py` remote path both delegate to `session.cancel_inflight()` — deduplicates the inline skill/plan cancel logic that was duplicated in both callers
- `RouterHostAdapter` wires the session flag via `turn_cancel_fn` callback (no session identity leak into adapter)

## V1 boundary (explicit)

Ctrl-C stops the chain **at the next tool-iteration boundary** — a slow tool already in flight (e.g. sandboxed_exec) completes before the cancel fires. Subprocess kill is follow-up scope. This resolves the owner-reported symptom ("chain continues") without the subprocess kill risk.

Idle Ctrl-C (no turn running): `_turn_cancel_requested` is set but cleared at the next `_run_router_loop` entry — spurious-safe by construction.

## Replay fixtures

No SP / tool description changes — replay fixtures unaffected.

## Test plan

- [x] `pytest tests/test_turn_cancel_1468.py` — 9 passed (cooperative cancel, event emission, single seam, spurious-safe idle, phase-host no-op)
- [x] `pytest -q` (full suite) — all passed
- [x] `ruff check .` — no errors

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -1772,6 +1772,15 @@ class RouterLoop:
         _plan_invalid_retries: int = 0
 
         for _iteration in range(self.max_iterations):
+            # #1468: cooperative turn-cancel checkpoint. Checked BEFORE the LLM
+            # call so a cancel_inflight() fired between tool iterations stops the
+            # chain at the next boundary (= after the current tool completes, not
+            # mid-call). getattr-guarded → phase hosts that don't implement
+            # _is_turn_cancel_requested are no-ops (byte-identical).
+            _cancel_fn = getattr(host, "_is_turn_cancel_requested", None)
+            if callable(_cancel_fn) and _cancel_fn():
+                host.events.emit("turn_cancelled", chain_id=self.chain_id)
+                break
             resolved_model = host.resolve_model(self.router_model)
             # #1092 PR-C-5 (2): per-turn phase wall-clock budget enforcement. A phase
             # host implements ``check_phase_budget`` (RAISES PhaseBudgetExceededError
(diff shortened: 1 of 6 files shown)


tya5:
**[lead-coder]** — APPROVE. Architecture verified per the approved design: single seam `session.cancel_inflight()` covering turn+skills+plans (flag owned by session, adapter forwards via `_is_turn_cancel_requested`, getattr-guarded so phase hosts are byte-identical); cooperative checkpoint at the run_loop iteration top emitting `turn_cancelled` (P6) + clean break; both UIs (TUI direct + WS) route through the shared seam — the dual-impl smell removed. V1 boundary documented in the PR body (next-tool-boundary stop; subprocess kill = follow-up). Tier-audit violations fixed properly (unpack-enforcement idiom + public observation method). Tier rule check ✓. Merging.

Issue #1496, opened 2026-06-11
#1495 design: limit-deny → force-close wrap-up (3-axis uniform wiring)


tya5:
**[e2e-coder]** — Design doc for owner review before implementation. Design reviewed by lead-coder 2026-06-12.

## Background

Currently, all \`handle_limit_exceeded\` deny paths emit a canned error and hard-stop the turn. The permission-deny path instead gives the LLM one final tool-less turn to adapt. This asymmetry is the defect to fix.

**Owner directive**: limit-deny should behave like force-close — give the LLM one final tool-less turn (\`tools=[]\`) to summarize what was done, what remains, and why the turn stopped. Error signal preserved. Applies uniformly across chat/plan/phase axes for loop-exhaustion sites.

---

### A. \`ask_timeout_seconds=0.0\` stays as default (no code change)

Infinite wait preserves the exact paused state of an in-flight loop. When the user returns and answers yes, the loop resumes from that point. A finite timeout would destroy the in-flight state — the user returns to find an unresumable partial run. **Continuability > result preservation**. Finite timeout = operator opt-in via \`safety.on_limit.ask_timeout_seconds: N\`. For headless/CI: \`mode: unattended\`.

Force-close wrap-up fires only on REAL deny: user "no" / \`unattended\` / \`no_bus\` / operator-configured finite timeout expiry. The default interactive+0.0 waits for human; on yes → extension → no force-close triggered.

---

## Q1 — Wrap-up SP: cause-neutral (lead-coder decision)

\`wrap_up_system_prompt()\` stays **cause-neutral**: "this turn is ending, no tool calls, consolidate done/outputs/remains". The specific cause (limit X hit) is placed in an **injected context message** added to the message list before the force-close call. Reason: putting cause in SP creates combinatorial explosion + the existing "context approaching size limit" text would be **false** for limit-deny (context is NOT full).

**⚠ Implementation note**: making \`wrap_up_system_prompt()\` cause-neutral changes the text seen by the LLM in existing cumulative-axis (#1092) force-close paths. Replay fixtures that captured the old "context size" framing will need selective re-record. PR body must note: "SP made cause-neutral; cumulative path wrap-up behavior is semantically unchanged."

---

## Q2 — Site C (router_cap): exception to uniform policy (lead-coder decision, confirming with owner)

[...]

Issue #1495, opened 2026-06-11
Wire limit-deny → force-close wrap-up (LLM final say, tool-less) instead of canned error — true symmetry with permission-deny's decision-enabling continuation (owner directive)


tya5:
**[lead-coder]** — owner directive (2026-06-11). Follow-on to the safety wave (#1492/#1494). Resolves the remaining asymmetry between limit-deny and permission-deny found while answering the owner's question.

## The asymmetry (verified on current main)
- **permission-deny**: caught in `dispatch_tool` → `permission_denied` tool_result (decision-enabling what/why/options) appended to history → **loop continues** → the LLM sees it and adapts in-turn.
- **limit-deny** (allow_continue=False, all sites uniform): chat router emits a CANNED `kind="error"` ("Router loop exceeded max iterations…") + returns; OSRuntime/phase raises `PhaseBudgetExceededError`. The LLM never speaks; the user sees a mechanical error.

## Owner's design
On limit-deny: do NOT just emit a canned error. Instead — **give the LLM a final tool-less turn**: tell it an error/limit was hit, advertise NO tools, and let the LLM decide how to respond (summarize what it did, what remains, why it stopped). Then the turn ends and the user decides whether to continue. (= the closest analog to permission's decision-enabling deny, bounded so it can't keep working.)

## Reuse existing machinery — do NOT build new
The force-close wrap-up already exists: `router_loop.py:~2680` `_force_close()` + `_build_force_close_messages()` + the tool-less call (`tools=[]`, continuation suppression, :2721) + `turn_budget/engine.py:wrap_up_system_prompt()`. Today it fires on cumulative-budget overflow (#1092). Wire it to **also** fire on limit-deny.

## Design-first scope (e2e)
1. **Chat/router path**: replace the canned-error-emit on limit-deny with a force-close wrap-up call — inject "limit X hit, no more tools, give your final summary" context + the wrap-up SP, `tools=[]`, emit the LLM's wrap-up as the user-facing message, THEN terminate the turn.
2. **Still signal the error** (owner: "エラーであることは伝えて"): keep an audit error event + make the wrap-up context convey "forced stop due to limit X" (not a normal completion). The user-facing text is the LLM's wrap-up; the machine signal stays for audit.
[...]

Issue #1493, opened 2026-06-11
RouterLoop checkpoint ask: A2A surfacing via A2AInterventionBus.on_dispatch


tya5:
**[e2e-coder]** — follow-up tracking issue for #1492.

## Context

FP-0005 wires `RouterLoop.max_iterations` exhaustion through `handle_limit_exceeded`. In interactive mode, the checkpoint dispatches a `UserIntervention` ask via `bus.request()`. For TUI/chat sessions this works correctly (ChatInterventionBus → TUI render).

For A2A peer sessions, `ChatInterventionBus` is wired but the A2A task cannot surface the ask to the originating agent — there is no `input-required` mirror via `A2AInterventionBus.on_dispatch`. This causes an infinite hang (`ask_timeout_seconds=0.0`).

## Current workaround (PR #1492)

`web/deps.py` `_session_factory` forces `on_limit=OnLimitConfig(mode="unattended")` for all A2A peer sessions. This degrades gracefully: max_iterations exhaustion → decision-enabling error + FP-0005 `partial_data` → A2A task terminates cleanly.

## Required fix

Wire the RouterLoop checkpoint ask through `A2AInterventionBus.on_dispatch` so `input-required` is mirrored to the originating agent's session. The originating agent can then surface the ask to the user via its own TUI/intervention channel.

Implementation sketch:
1. `A2AInterventionBus` dispatches `UserIntervention` as an `input-required` inbox event to the originating agent
2. The originating agent's session receives it and routes to its own `ChatInterventionBus`
3. On user response, the answer propagates back to the A2A peer session

## Acceptance

Once implemented, remove the `on_limit=unattended` override in `web/deps.py` `_session_factory` and re-enable interactive mode for A2A.

PR #1492, opened 2026-06-11
feat(safety): Fix A+B — FP-0005 max_iterations checkpoint + decision-enabling errors

tya5:
**[e2e-coder]** — safety framework wave impl PR (Fix A + Fix B, chat + planner + A2A axes).

## Summary

**Fix A** — wire \`max_iterations\` exhaustion through \`handle_limit_exceeded\` (FP-0005 bypass cluster):

- \`RouterLoop.run_loop()\`: outer \`while True:\` wraps the inner \`for _iteration in range(self.max_iterations):\` loop. \`_loop_cancelled\` sentinel distinguishes cancel-break from exhaustion. After exhaustion: if \`_on_limit is not None\`, calls \`handle_limit_exceeded(bus=make_intervention_bus(), ...)\`. On approval: extends \`max_iterations\` and \`continue\`s. On refusal/no-bus/unattended: \`break\` → decision-enabling error.
- \`RouterLoopHost\` protocol: \`make_intervention_bus() -> Any | None\` (P7-clean; getattr-guarded so old hosts degrade to no-ask).
- \`RouterHostAdapter\`: implements \`make_intervention_bus()\` returning the session's bus factory result or None.
- \`RouterLoopDriver\`: threads \`on_limit=getattr(self._safety, "on_limit", None)\` to \`RouterLoop\`.
- \`planner.py\`: threads \`on_limit=on_limit\` to each step's \`RouterLoop\` so \`step_max_iterations\` exhaustion routes through \`handle_limit_exceeded\`.
- \`web/deps.py\` \`_session_factory\`: A2A peer sessions force \`on_limit=OnLimitConfig(mode="unattended")\` — A2A cannot surface RouterLoop checkpoint asks via \`A2AInterventionBus\` (no input-required mirror path). Graceful degrade: decision-enabling error instead of infinite hang. Follow-up: #1493.

Tests that use \`ChatSession\` + a delegate stub that never returns text must declare \`OnLimitConfig(mode="unattended")\` on their session — the designed CI escape hatch. Two tests in \`test_session_invariants.py\` updated.

**Fix B** — decision-enabling error messages:
- Before: \`"Router loop exceeded max iterations (N)."\`
- After: \`"Router loop exceeded max iterations (N). Configure safety.on_limit.mode=interactive or auto_extend to extend, or increase router_max_iterations."\`

## Owner-veto item (朝の判断用)

[...]

PR #1497, opened 2026-06-11
feat(safety): #1496 chat axis — limit-deny → force-close wrap-up (site H: max_iterations)

tya5:
**[e2e-coder]** — #1496 impl PR-1: chat axis (site H). Design approved at #1496.

## What this PR does

When \`RouterLoop.run_loop()\` exhausts \`max_iterations\` and the limit is denied (extension refused/unattended/no_bus), instead of emitting a flat canned error, give the LLM one final tool-less turn to summarize the work before the turn ends.

### \`wrap_up_system_prompt()\` — cause-neutral (turn_budget/engine.py)

Removes "context approaching its size limit" framing. New text: "You are being asked to WRAP UP the current unit of work." Applies uniformly to both cumulative-axis overflow (existing) AND limit-deny (new). The specific cause is injected as a context message at call time, not baked into the SP.

**Replay fixture impact**: no fixtures pinned the old "approaching its size limit" text. SP tests use \`wrap_up_system_prompt()\` as a callable (not literal), so they auto-pass.

### \`RouterLoop.run_loop()\` — limit-deny path (router_loop.py)

```
limit exhausted → no extension granted
  ├─ _loop_cancelled=True (user-initiated cancel): canned error (unchanged)
  └─ limit-deny path (#1496):
       1. emit limit_denied WAL event (kind, limit, chain_id)
       2. inject cause context message into messages
       3. _force_close_call_with_retry(messages + cause, resolved_model)
       4a. LLM returns text → emit as "agent" (meta.limit_stopped=True, meta.limit_kind="max_iterations") → return
       4b. LLM returns no text / raises → degrade to canned error
```

Error signal: \`limit_denied\` WAL event + \`meta.limit_stopped\` on the outbox message. No competing prose blocks.

## Tests

- \`test_safety_max_iterations_fp0005.py\`: new test 6 — happy-path wrap-up: scripted LLM returns text after exhaustion → agent message with \`meta.limit_stopped\` emitted, no canned error, \`limit_denied\` event in WAL
- \`test_replay_skill_router.py::test_max_iterations_aborts_with_error\`: call count 3→4 (3 loop iterations + 1 force-close wrap-up call; stub returns tool_calls → content=None → fallback canned error)
- \`test_routerloop_convergence_parity_deferrals_1092.py\`: count updated to \`_EFFECTIVE + 1\` (same pattern)

## Staged implementation (per #1496 design)

- **This PR**: chat axis (site H: max_iterations). Pattern established. sandbox_2 verifies chat wrap-up.
- **Next PR**: plan axis (site I: step_max_iterations) + phase axis (sites A: max_act_turns, F: phase_seconds)
- **HOLD**: site C (router_cap) pending owner Q2 confirmation

## Gate

- **Unit**: 34 tests pass (ruff clean, tier audit clean) ✅
- **sandbox_2 live**: max_iterations forced low → LLM wrap-up message visible instead of canned error (pending PR landing)

## Full suite verbatim

```
FAILED tests/test_eval_benchmark_tier1_swebench.py::test_swebench_missing_honest_skip
FAILED tests/test_web_fetch_preview_path_ref.py::test_legacy_no_media_store_path_returns_inline_content
2 failed, 6995 passed, 10 skipped, 3 xfailed, 5 warnings in 271.93s (0:04:31)
```

Both pre-existing on main.

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -2588,7 +2588,59 @@ class RouterLoop:
                 continue  # re-enter inner for loop with extended limit
          break  # no extension granted or no on_limit — exit outer while
 
-        # max_iterations exhausted (or cancelled)
+        # Cancelled path (user-initiated): canned decision-enabling error.
+        if _loop_cancelled:
+            await self.host.put_outbox(
+                kind="error",
+                text=(
+                    f"Router loop exceeded max iterations ({self.max_iterations}). "
+                    f"Configure safety.on_limit.mode=interactive or auto_extend to "
+                    f"extend, or increase safety.loop.max_router_iterations."
+                ),
+                meta={"chain_id": self.chain_id},
+            )
+            return self._total_usage
+
+        # Limit-deny path (#1496): give the LLM one final tool-less turn to
+        # summarize what was accomplished before the turn ends. The cause
+        # is injected as a context message (SP stays cause-neutral). A
+        # structured marker on the outbox message signals forced-stop to
+        # the UI without a competing prose block. Degrades to canned error
+        # if the wrap-up call itself fails.
+        host.events.emit(
+            "limit_denied",
+            kind="max_iterations",
+            limit=self.max_iterations,
+            chain_id=self.chain_id,
+        )
+        _cause_msg = {
+            "role": "user",
+            "content": (
+                f"This turn is ending because the router reached its iteration "
+                f"limit ({self.max_iterations} iterations). "
+                "Do not call any more tools. "
+                "Summarize what has been accomplished, where outputs are, "
+                "what remains, and that this turn stopped due to a limit."
+            ),
+        }
+        try:
+            _wrapup = await self._force_close_call_with_retry(
+                [*messages, _cause_msg],
+                resolved_model=host.resolve_model(self.router_model),
+            )
+            if _wrapup.content:
+                await self.host.put_outbox(
+                    kind="agent",
+                    text=_wrapup.content,
+                    meta={
+                        "chain_id": self.chain_id,
+                        "limit_stopped": True,
+                        "limit_kind": "max_iterations",
+                    },
+                )
+                return self._total_usage
+        except Exception:  # noqa: BLE001 — wrap-up failed; degrade to decision-enabling error
+            pass
         await self.host.put_outbox(
             kind="error",
             text=(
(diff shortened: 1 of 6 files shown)


tya5:
**[lead-coder]** — code APPROVE (chat axis, #1496 site H). Merge precondition = CI green + sandbox_2 live verify (two-layer gate). Verified: SP made cause-neutral (removed the 'context approaching size limit' clause → reused for limit-deny without the false 'context full' framing); cause delivered via injected context message ('this turn is ending because the router reached its iteration limit (N)'); limit_denied WAL event (audit); force-close wrap-up via the existing _force_close_call_with_retry; emit as 'agent' with meta.limit_stopped=True (structured marker — no competing second prose block per the design note); fallback to kind=error when wrap-up returns empty. **No fixture re-record needed — empirically confirmed** (green suite + zero fixture changes = the force-close SP isn't in any replay request key; the cumulative-axis path it's shared with isn't exercised by fixtures). Test call-count N→N+1 updates are legitimate (the wrap-up is one additional LLM call by design). Tier rule check ✓. Holding merge for the live layer.

tya5:
**[lead-coder]** — APPROVE + merging (chat axis, #1496 site H). **Both two-layer gate preconditions met.**

## Live gate (sandbox_2 re-gate #3 @4d14e532) = PASS
4 confirmed limit-hits (unique-file trigger to defeat read-memoization) → ALL produced LLM wrap-up text with **0 GeminiException + 0 canned errors**. Before(pre-fix): limit-hit → GeminiException "config without declarations" → canned. After: limit-hit → LLM wrap-up. The `llm.py` `tool_choice`-omit fix root-treated the true cause.

## The arc this PR's two-layer gate caught (the gate earning its keep)
1. **empty-checkpoint → phase infinite re-entry** (e2e self-caught, content-guard fix);
2. **wrong root-cause** (role-sequence hypothesis) — overturned when sandbox_2 captured the FULL Gemini error body (the prior truncated body had misled all three of us);
3. **the true cause = a pre-existing #1092 latent bug** (tool-less force-close passed `tool_choice` with no `function_declarations`; never live-fired on Gemini until #1495 exposed it) — fixed at the `llm.py` call boundary, so ALL tool-less calls (this path + #1092 cumulative + future) are now Gemini-safe.

## Residual (tracked, non-blocking)
`meta.limit_stopped` IS set on the confirmed-reached wrap-up-success path, but doesn't surface in the A2A /events stream — the same A2A-meta-surface gap as #1492, covered by the existing #1493 A2A-mirroring follow-up.

Tier rule check ✓. Merging. Next: plan(I) / phase(A,F) / router_cap(C, now uniform per owner's decision) — same wrap-up pattern, live-de-risked by this chat-axis proof.

PR #1588, opened 2026-06-14
test(1587): de-flake Esc-Esc-pending timing tests (deterministic clock)

### LINKED ISSUE #1587: Flaky: Esc-Esc-pending timing tests fail intermittently on pytest 3.11
**[lead-coder]** — a recurring **timing-dependent test flake** (testing policy: never pin wall-clock timing).

## Occurrences (2, same class)
1. `tests/cli/test_esc_esc_rewind_1546.py::test_double_esc_within_window_opens` — `assert _conv_has(app, "no checkpoints")` False (seen during #1573 CI)
2. `tests/cli/test_fork_picker_edit_mode_2c.py::test_exit_edit_mode_resets_esc_esc_pending` — `assert app.esc_esc_pending is True` False (seen during #1586 CI)

## Pattern
- Both assert on the **Esc-Esc double-tap pending state** (`esc_esc_pending` / the `_ESC_ESC_WINDOW_S` window).
- Both fail **only on pytest 3.11** (3.12 + ruff + tier-audit pass), **1 of ~7200**, and pass on re-run.
- Both are **unrelated to the PR's diff** (the PRs touched substrate/observer/config, not the TUI Esc-Esc path) → confirmed flake, not regression.

## Root cause (hypothesis)
The Esc-Esc-pending logic uses a wall-clock window (`_ESC_ESC_WINDOW_S`, monotonic ts). The tests depend on real elapsed time between simulated Esc presses; under 3.11's slightly different async/timer scheduling the second tap can fall outside the window → the pending state isn't what the test expects.

## Fix direction
De-flake by removing the wall-clock dependency from the test: inject a deterministic clock / monkeypatch the time source so the Esc-Esc window is exercised deterministically (not via real elapsed time). The production timing logic stays; only the test's reliance on wall-clock is removed. (Testing policy: NEVER pin timing — these tests currently do, implicitly.)

Low-urgency (re-run clears it), but recurring → worth a deterministic-clock fix. TUI domain.

tya5:
**[tui-coder]** — Closes #1587. De-flakes the recurring Esc-Esc-pending timing flake (1-of-~7200, **3.11-only**, re-run passes, PR-diff-untouched). Testing policy: NEVER pin wall-clock timing.

## Root cause
Two wall-clock dependencies in the tests vs the `_ESC_ESC_WINDOW_S` (0.6s, monotonic-ts) window:
1. The real auto-clear `set_timer(_ESC_ESC_WINDOW_S, _clear_esc_esc_hint)` fires after 0.6s and zeroes `_last_clean_esc_ts`. On a slow 3.11 run the test crossed 0.6s **before** its `assert esc_esc_pending is True` → the timer had already cleared it.
2. `test_double_esc_outside_window_rearms` used `asyncio.sleep(_ESC_ESC_WINDOW_S + 0.05)` — the +0.05 margin raced 3.11's timer scheduling.

## Fix — pure test-side (no production change)
The production monotonic-ts window logic is **unchanged**; the tests just stop depending on real elapsed time:
- **`test_esc_esc_rewind_1546.py`**: an autouse fixture widens `_ESC_ESC_WINDOW_S` to `1e6` so the real auto-clear timer fires at ~1e6 s (never during a sub-second test → the pending state isn't zeroed mid-test) and back-to-back Escs are always within-window. `test_double_esc_outside_window_rearms` advances a **faked `_now_monotonic`** past the window (deterministic lapse) instead of sleeping.
- **`test_fork_picker_edit_mode_2c.py::test_exit_edit_mode_resets_esc_esc_pending`**: same window-widen so the timer can't pre-empt the reset-on-exit assertion.

## Test plan
- [x] `pytest tests/cli/test_esc_esc_rewind_1546.py tests/cli/test_fork_picker_edit_mode_2c.py` — 17 passed
- [x] **20× local loop** of the two previously-flaky tests — 0 failures (deterministic by construction: timer at 1e6 s never fires sub-second; re-arm uses a fully faked clock)
- [x] `ruff check` + tier audit `--strict` — clean
- [x] `git status` — **2 test files only, zero production diff** (confirms "production timing logic stays")

**Tier self-check**: Tier 2, public surface (`esc_esc_pending`, `check_action`, conv buffer); the de-flake removes the implicit wall-clock pin (testing-policy "NEVER pin timing"), it doesn't add one. Refs #1546, #1554.

### DIFF
--- a/tests/cli/test_esc_esc_rewind_1546.py
+++ b/tests/cli/test_esc_esc_rewind_1546.py
@@ -14,7 +14,6 @@ Pins (real key path through bindings + check_action, run_test pilot — no mocks
 """
 from __future__ import annotations
 
-import asyncio
 import sys
 from pathlib import Path
 
@@ -24,8 +23,20 @@ _SRC = Path(__file__).parent.parent.parent / "src"
 if str(_SRC) not in sys.path:
     sys.path.insert(0, str(_SRC))
 
-from reyn.chat.tui.app import _ESC_ESC_WINDOW_S, ReynTUIApp
+import reyn.chat.tui.app as app_mod
+from reyn.chat.tui.app import ReynTUIApp
 from reyn.chat.tui.widgets import ConversationView, InputBar
+
+
+@pytest.fixture(autouse=True)
+def _wide_esc_window(monkeypatch):
+    """De-flake (#1587): make the Esc-Esc window huge so the real auto-clear
+    ``set_timer`` never fires during a (sub-second) test — removing the
+    wall-clock race that let a slow 3.11 run zero the pending state mid-test.
+    The production window behaviour is unchanged (same monotonic-ts logic, just
+    a value the test controls); tests that need the window to *lapse* advance a
+    faked ``_now_monotonic`` past it instead of sleeping real time."""
+    monkeypatch.setattr(app_mod, "_ESC_ESC_WINDOW_S", 1_000_000.0)
 
 
 def _make_app() -> ReynTUIApp:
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — de-flake review PASS. Pure test-side: **0 src/ files** (verified — production monotonic-ts window logic untouched, only the tests' wall-clock reliance removed). Root cause correctly diagnosed (the real auto-clear `set_timer(_ESC_ESC_WINDOW_S)` firing mid-test on slow 3.11 + the re-arm test's `sleep(window+margin)`). Fix removes the timing dependency the right way — widen the window so the timer can't fire sub-second + advance a faked `_now_monotonic` instead of real sleep = deterministic by construction (testing policy: removes the implicit timing-pin, doesn't add one). 20× loop / 0 failures proves it. Covers both flaky tests (#1546 + 2c exit_edit_mode). Merging on CI-green → closes #1587 (the recurring Esc-Esc flake class).

PR #1592, opened 2026-06-14
docs(time-travel): cost-tuning how-to (workspace_capture / act_turn_capture)

tya5:
**[docs-maintainer]** — opt-in/out user-doc per owner instruction (via lead-coder dispatch).

## Summary

Adds a `## Tuning time-travel cost` section to `guide/for-users/time-travel.md`. The config reference (`reyn.yaml § time_travel`) and concept (cost section) were already complete; the gap was a user-facing how-to. This fills it.

- **`workspace_capture: false`** (opt-out, real working feature) — runtime-only rewind: agent/conversation state rewinds, repo files don't. When-to-use (large workspace / container / re-run-only), crash-recovery-unaffected note, run-level (startup) note.
- **`act_turn_capture: true`** (opt-in, advanced) — per-step capture with **honest framing**: enabling it today gives *no user-visible rewind benefit* (picker doesn't expose mid-skill-run landing yet); foundational groundwork, default-off, no-op when `workspace_capture: false`.

Links to reference (what each key does) + concept (why it costs). No table duplication.

## Verification note

Premise-verified against current `origin/main` (`b09cf996`) before writing: `time_travel` block keys confirmed in config.py (15 hits), reyn-yaml.md (L275-288), concept cost section (L132). (First pass hit a stale remote-tracking ref from a partial `git fetch origin main`; resolved with full `git fetch origin`.)

## Test plan

- [x] Anchor links resolve: `#time_travel-block` (reyn-yaml) + `#cost-and-the-runtime-only-opt-out` (concept) — both match existing headers
- [x] No history refs in the new section (user doc — grep clean)
- [x] No mermaid added
- [x] No reference-table duplication (guide = how/when, reference = what)
- [x] Page drop: zero (addition only)
- [ ] lead-coder review of act_turn_capture honest framing

### DIFF
--- a/docs/guide/for-users/time-travel.md
+++ b/docs/guide/for-users/time-travel.md
@@ -54,6 +54,40 @@ When using the container environment backend (`--container`), the workspace
 rewind operates on the container filesystem via shadow-git `as-of-N` — the same
 user experience applies; the substrate is container-side.
 
+## Tuning time-travel cost
+
+Time-travel is on by default. Its largest constant cost is the **workspace shadow-git capture** that runs at every checkpoint boundary (every turn and plan-step) so a rewind can restore your repo files. You can tune this in `reyn.yaml` under the `time_travel` block. For the full per-key reference see [`reyn.yaml` § time_travel](../../reference/config/reyn-yaml.md#time_travel-block); for *why* it costs what it does, see [Cost and the runtime-only opt-out](../../concepts/runtime/time-travel.md#cost-and-the-runtime-only-opt-out).
+
+### Opt out of file rewind (`workspace_capture: false`)
+
+```yaml
+# reyn.yaml
+time_travel:
+  workspace_capture: false   # default is true
+```
+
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — review PASS, merging. Accurate + honest user-facing how-to:
- **`workspace_capture: false`** — runtime-only rewind framed correctly (picker/checkout work, working tree untouched); when-worth-it (large workspace / container / re-run-only); **"crash recovery is unaffected — it rides the WAL, not the workspace capture"** correctly reassures (owner's don't-weaken-crash); run-level not mid-session. ✅
- **`act_turn_capture: true`** — the **honest status** is exactly right: *"enabling it today gives no user-visible rewind benefit — the picker does not yet expose a way to land a rewind mid-skill-run, so the extra per-step snapshots are foundational groundwork… not something you can currently navigate to."* No oversell of an unexposed capability (audit_status_not_existence). ✅
- New section: no history-refs; links to the config reference + concept doc by path. ✅

Closes the owner's opt-in/out user-doc directive.

PR #1626, opened 2026-06-14
fix(llm): capture provider response shape on empty-choices (flake observability)

tya5:
**[dogfood-coder]** — flake observability (lead-dispatched from my empty-choices flake trace). The ② A/B oracle hit a ~6/9 empty-choices flake window; my repro **ruled out structural causes** (payload-size / concurrency / burst all clean in direct litellm calls → not a reyn code path), leaving a transient provider/proxy condition. But it's **not diagnosable**: `EmptyLLMResponseError` (llm.py) carried only the model name, so the WHY — Gemini `finish_reason` SAFETY/MAX_TOKENS/RECITATION, `prompt_feedback`, `usage` (vendor-specific fields we can't predict) — is lost on every recurrence.

## Change (observability only — no retry/raise contract change)
- `_empty_response_diag(response)`: a truncated (≤500c) `json.dumps(response.model_dump())`, **best-effort** (falls back to `repr`, never raises — a diag failure must never mask the empty-choices error itself).
- At the raise site: `logging.warning` it **per attempt** (a flake that self-recovers on retry still leaves its WHY in the log) + include it in the `EmptyLLMResponseError` message (the exhaustion path).

→ The next empty-choices recurrence is **diagnosable** (finish_reason / prompt_feedback / safety-block surface) → root-cause-able. This is the enabler I couldn't get during the trace (no live empty reproduced; nothing logs the shape).

## Tests (Tier 2, real Fakes, no mock)
- `test_empty_response_diag_includes_vendor_fields` — diag surfaces `prompt_feedback`/block-reason/`usage` from a `model_dump()`-able response.
- `test_empty_response_diag_best_effort_when_no_model_dump` — degrades to a non-empty `repr` str, never raises.
- `test_empty_choices_error_message_carries_provider_diag` — the raised error message carries the diag (`match="RECITATION"`), sleep no-op'd for speed.
- All 13 in `test_llm_call_retry.py` pass; strict tier-audit 0-errors; existing #187 tests 9/10 unaffected (type-only raises).

## Context
Root-cause verdict (reported to lead): the flake is transient shared-proxy/provider load, **not a structural bug** → no behavioral fix to dispatch; this observability add is the only worthwhile change so a future recurrence is diagnosable. Core llm.py path — careful review welcome.

## Test plan
- [x] `pytest tests/test_llm_call_retry.py -W error::RuntimeWarning` (13 pass)
- [x] strict tier-audit on the test file (0 errors)
- [x] `import reyn.llm.llm` clean
- [ ] (reviewer) confirm the per-attempt `logging.warning` volume is acceptable (fires only on empty-choices, which is rare outside a flake window)

---
PR self-check: docstrings declare Tier; no Tier-4 violations; no invariant weakened (retry/raise contract unchanged — only the error message + a log line added); real-instance Fakes, no mock.

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -532,6 +532,19 @@ class EmptyLLMResponseError(Exception):
     """
 
 
+def _empty_response_diag(response: object) -> str:
+    """Compact provider-response shape for an empty-choices error (flake
+    observability). The block reason — Gemini ``finish_reason``
+    SAFETY/MAX_TOKENS/RECITATION, ``prompt_feedback``, ``usage`` — lives in
+    vendor-specific fields we can't predict, so dump the whole (truncated) response
+    so a recurrence is diagnosable instead of a bare "empty choices". Best-effort:
+    diagnostics must NEVER mask or replace the empty-choices error itself."""
+    try:
+        return json.dumps(response.model_dump(), default=str)[:500]
+    except Exception:  # noqa: BLE001 — never let a diag failure shadow the real error
+        return repr(response)[:300]
+
+
 def _get_retryable_litellm_exceptions() -> tuple:
     """Return the tuple of retryable litellm exceptions, loading litellm lazily.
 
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — review **PASS**, auto-merge enabled. Pure-observability, primary-source verified:
- `_empty_response_diag()` is best-effort: `try model_dump()[:500]` / `except → repr()[:300]`, **never raises** (a diag failure can't shadow the real error), output truncated/bounded.
- Raise site adds a per-attempt warning log + folds the diag into the EmptyLLMResponseError message. **No retry/raise contract change** — the error still raises identically, just with provider-shape context (finish_reason / prompt_feedback / usage).
- Success path untouched (diag runs only in the empty-choices error branch).
- 3 Tier-2 tests, **real Fakes (no mock)**: best-effort-fallback (no model_dump → repr, never raises) + error-message-carries-diag (RECITATION block_reason surfaces). testing-policy conformant.
- Reviewer item (per-attempt warning volume): fires only on empty-choices = bounded to flake windows, which is exactly when the diagnostic is wanted. Acceptable.

Core llm.py but observability-only + well-tested → no co-vet needed (author + lead review). Auto-merge (squash) on CI-green. → the next empty-choices recurrence is diagnosable (finish_reason SAFETY/MAX_TOKENS/RECITATION vs genuine-empty). Good mining-infra add.

Issue #1647, opened 2026-06-15
FP-0034: expose MCP tools as first-class qualified actions (mcp__<server>__<tool>), retire generic call_mcp_tool indirection


tya5:
**[lead-coder]** — deferred architectural follow-up split from #1646 (the empty-MCP-args bug, fixed minimally there by renaming the inner param to kill the double-`args` collision).

## Context
#1646's root cause: MCP tools are reached via a **generic `call_mcp_tool(server, mcp_tool_name, args:{...})` verb**, wrapped by `invoke_action(args:{...})` → **two content-free `args` levels** that a heavy-context LLM collapses → empty tool args. #1646 fixes this minimally by renaming the inner param (distinct names, collision gone).

## The fuller FP-0034 root-fix (this issue)
Expose each MCP tool as a **first-class qualified action** `mcp__<server>__<tool>` resolvable via `universal_dispatch`, whose `describe_action` surfaces the tool's **real inputSchema**, so the LLM emits `invoke_action(action_name="mcp__<server>__<tool>", args={the tool's own params})` — **one args level, the real param schema, no generic content-free `args`, no collision possible by construction**. This is the FP-0034-consistent design (qualified-name dispatch + describe_action + per-resource real schemas, like skill__*/multi_agent__*).

Scope (bigger than #1646): adds **dynamic `mcp__<server>__<tool>` resolution** to universal_dispatch (today only static `mcp__call_tool` etc. verbs resolve) + per-tool inputSchema surfacing via describe_action + enumerate listing. Retire/deprecate the generic call_mcp_tool verb once per-tool actions cover it.

## Why deferred (not in #1646)
#1646's rename is the proportionate structural fix for the confirmed bug; this is a larger architectural change (new resolution capability + schema surfacing across all schemes). **Owner decision** to schedule — it's the ideal MCP abstraction, but disproportionate to bundle with the bug fix. Lead-coder leaning: do it after #1646 lands + verifies, if owner wants the full FP-0034 MCP surface. Not started.

tya5:
**[lead-coder]** — scope input surfaced by #1646's full-flow trace: there are currently **TWO distinct MCP-call handlers with different shapes**, which compounds the LLM confusion:
- `call_mcp_tool` (mcp.py:459) — `{server, mcp_tool_name, args}`, rendered DIRECTLY to the router LLM (router_tools.py:842) with server/mcp_tool_name enums.
[...]

Issue #1646, opened 2026-06-15
MCP tool call may receive empty args (toolcall args not threaded to mcp tool.args)


tya5:
**[lead-coder]** — owner-reported (2026-06-15): the LLM's tool-call args may not reach the MCP tool invocation (`tool.args`). **Framed as a hypothesis pending live-repro confirmation** — code-level map below; the drop point must be confirmed via trace, not asserted from reading.

## Code-level map (canonical MCP path)
- LLM is expected to call `mcp__call_tool` (verb) with `{server, mcp_tool_name, args:{...}}` — there are no individual `mcp__<server>__<tool>` aliases in the universal scheme (`universal_catalog.py:70-73`; `universal_dispatch._OPERATION_RULES` line 268-275).
- `mcp__call_tool` → `_passthrough_args` → `mcp_call_tool` handler.
- Handler `tools/mcp.py:248-254`: `tool_args = dict(args.get("args") or {})` → `host.mcp_call_tool(server, bare_tool, tool_args)`. **Reads the NESTED `args` key strictly — no flat fallback.**

## Hypothesis
If the LLM (or a scheme SP / transformer / salvage rewrite) passes the tool's parameters **flat** (alongside `server`/`mcp_tool_name`) instead of **nested under `args`**, then `args.get("args")` → `{}` and the MCP tool runs with **empty args** — matching the owner's symptom.

## Investigation (confirm-repro-FIRST, then root-cause, then uniform fix)
1. **Repro + trace** a real MCP call via `scripts/dogfood_trace.py --mode llm-payloads`: capture the actual `tool_calls[].function.arguments` the LLM emits for an MCP call (nested vs flat?), and trace the hops router_loop._execute_tool → dispatch → mcp.py args.get("args") → host.mcp_call_tool → MCPIROp → MCPClient. Identify the **exact** drop point with primary data.
2. Check all schemes (universal / enumerate / retrieval / codeact) — do any expose/forward MCP args differently (flat)?
3. Root-cause to the structural seam (not the symptom): is it (a) SP/schema not making nesting unambiguous, (b) a transformer/salvage losing the nesting, or (c) handler should accept flat leniently? Fix the seam **uniformly**; no per-call workaround.
4. Tier-2/3a test driving a real MCP call shape → assert the tool args arrive non-empty at the MCPIROp/MCPClient boundary (round-trip with a **non-default** value so an unwired path can't pass silently). MagicMock-free.

[...]

PR #1648, opened 2026-06-15
fix(1646): MCP tool args empty at call — kill invoke_action↔mcp_verbs `args` collision

tya5:
**[e2e-coder]** — Fixes the owner-observed bug: **MCP tool args are correct in the `tool_call` but EMPTY at the MCP call.**

## Root cause (confirm-repro-first; full-flow gate, not reading)

The bug lives on the `invoke_action`-wrapped path. `mcp_call_tool` (mcp_verbs) advertised its target-tool params under the inner key **`args`**, which collides with `invoke_action`'s OWN outer **`args`**. The LLM collapses the two levels — the target params land flat alongside `tool`, and the inner object is dropped → **empty params at the MCP boundary**.

Isolated schema probes couldn't reproduce it (the single-nest is context-dependent). A **load-bearing full-flow test** on the real wrapped path is what pins it.

## Fix (rename-only; by-construction)

| | change |
|---|---|
| **K3** (the collision-kill) | mcp_verbs `mcp_call_tool` LLM-facing inner key `args`→`tool_args` (schema + read). Distinct from `invoke_action`'s outer `args`, so the nest can't collapse. |
| **K4** (regression WIP 1/2 introduced) | the mcp_verbs→mcp.py delegation passed `"args"` while mcp.py (post WIP 1/2) reads `tool_args` → delegation dropped to `{}`. Now passes `tool_args`. |
| **single-source** | mcp_verbs imports `_MCP_TOOL_ARGS_KEY` from mcp.py so the four inner-key sites across both files can't drift. |

WIP 1/2 (`922a3d6c`, mcp.py inner rename + const) was right but incomplete — it missed both mcp_verbs sites. This PR completes it.

## Verification — load-bearing full-flow gate

`tests/test_mcp_invoke_action_tool_args_1646.py` drives the REAL path:
`invoke_action(action_name="mcp__call_tool", args={tool:"<server>__<tool>", tool_args:{...}})` → mcp_verbs → mcp.py → `host.mcp_call_tool`, asserting a **NON-DEFAULT value round-trips to the MCP boundary non-empty**. A contrast test pins that the pre-fix collapse (flat params, no inner level) yields empty args = the owner-observed failure.

- `pytest -k mcp`: **489 passed, 1 skipped**
- `test_tier_audit --strict` (new test): **0 errors**
- `ruff check .`: **clean**

## Consumer audit (complete)
- code: `mcp.py` + `mcp_verbs.py` (both inner-key sites + delegation)
- describe: registry renders `tool_args` on the LLM-facing schema
- docs: `popular-mcp-servers`, `mcp[.ja]`, `universal-catalog[.ja]` off the stale `{tool, args}`
- tests: `test_mcp_unified` + the new full-flow gate

The two-handler smell (`call_mcp_tool {server,mcp_tool_name}` mcp.py vs `mcp_call_tool {tool}` mcp_verbs — duplicate surfaces, different shapes) is **deferred to #1647** (not consolidated here; this PR = collision-kill + regression-fix).

Refs #1646

## Test plan
- [x] full-flow gate: non-default tool_args round-trips to MCP boundary (2/2 pass)
- [x] `pytest -k mcp` no regression (489 pass)
- [x] tier-audit --strict + ruff clean
- [x] docs consumers updated off stale `{tool, args}`

### DIFF
--- a/docs/concepts/tools-integrations/mcp.ja.md
+++ b/docs/concepts/tools-integrations/mcp.ja.md
@@ -66,7 +66,7 @@ reyn mcp install --source https://github.com/modelcontextprotocol/servers/tree/m
 | `mcp__install_local({name, command, args})`                 | local command (LLM 生成 script 等) を直接 MCP server として登録 |
 | `mcp__list_servers()`                                       | `.reyn/mcp.yaml` に設定された全サーバー名を返す |
 | `mcp__list_tools({server})`                                 | 1 サーバーが露出する tool 一覧を `{name: "<server>__<tool>", description, inputSchema}` 形式で返す |
-| `mcp__call_tool({tool, args})`                              | `<server>__<tool>` ID + tool の declared args で tool を call |
+| `mcp__call_tool({tool, tool_args})`                              | `<server>__<tool>` ID + tool の declared tool_args で tool を call |
 | `mcp__drop_server({server})`                                | install 済サーバーを config から削除 |
 
 LLM router がチャット turn 内で直接これらを呼べます。 初回利用の典型 flow:
--- a/docs/concepts/tools-integrations/mcp.md
+++ b/docs/concepts/tools-integrations/mcp.md
@@ -66,7 +66,7 @@ If you prefer to configure a server manually (or are adding a server not in the 
 | `mcp__install_local({name, command, args})`                 | Register a local command (e.g. LLM-authored script) as an MCP server |
 | `mcp__list_servers()`                                       | Returns the names of all servers configured in `.reyn/mcp.yaml` |
 | `mcp__list_tools({server})`                                 | Returns the tools exposed by one server (each entry has `name="<server>__<tool>"`, `description`, `inputSchema`) |
-| `mcp__call_tool({tool, args})`                              | Call a tool by `<server>__<tool>` identifier (from `mcp__list_tools`) with its declared args |
+| `mcp__call_tool({tool, tool_args})`                              | Call a tool by `<server>__<tool>` identifier (from `mcp__list_tools`) with its declared tool_args |
 | `mcp__drop_server({server})`                                | Remove an installed server from the config |
 
 The LLM router can call these directly during a chat turn. Typical first-time flow:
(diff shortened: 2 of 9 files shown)


tya5:
**[lead-coder]** — review **PASS**, auto-merge enabled. Rigorous re-review this time (I missed the target conflation + the cross-file delegation twice earlier; both now explicitly checked).

## Completeness (my own grep on the branch)
- **K3 collision-kill**: mcp_verbs LLM-facing `args`→`tool_args` (schema:488 + read:533) ✓
- **K4 regression-fix**: mcp_verbs:533 delegation now passes `tool_args` (was `"args"` → mcp.py reads `tool_args` → dropped to {}) ✓, with an explicit regression comment.
- **single-source**: `_MCP_TOOL_ARGS_KEY` defined mcp.py:100, imported mcp_verbs:43 — the 4 keys (mcp.py schema/router-read/phase-read + mcp_verbs schema/read/delegation) can't drift ✓
- invoke_action's OUTER `args` (universal_catalog:1313) correctly **untouched** — collision-kill is precise ✓
- the `args` at mcp_verbs:370/410/444 are the **install_local command-args** (different tool) — correctly left ✓
- docs consumer-complete: mcp[.ja] / universal-catalog[.ja] / popular-mcp-servers off the stale `{tool,args}` ✓

## Full-flow gate (the load-bearing piece)
`test_mcp_invoke_action_tool_args_1646.py` drives the **REAL** wrapped path through real registry handlers (invoke_action→resolve→mcp_verbs→mcp.py→host), only Fake = `_RecordingMCPHost` at the actual MCP boundary (not a mock). Asserts the non-default `query` round-trips intact + a contrast test pins collapsed→{}. **test 1 fails pre-fix** (genuine regression guard, not a tautology). The cross-file delegation is explicitly exercised — exactly the gate that caught the WIP 1/2 conflation.

## Tier rule check ✓
Tier 3a docstrings / no mocks (real handlers + recording fake) / public surface / non-default round-trip value / no Tier-4. ruff + tier-audit green.

## Scope notes
- (i) leniency correctly **skipped** — the contrast test documents that a collapsed shape still yields {} (deliberate non-tolerance, per the no-weak-model-overfitting bar).
- two-handler smell → **#1647** (consolidation), as scoped.

**Behavioral efficacy** (does the distinct key actually stop the LLM collapsing in the owner's live session) is owner-verifiable post-merge — the threading/collision-removal is proven; the LLM-behavior confirmation is the owner re-testing an MCP call. Auto-merge (squash) on pytest-green.

Issue #1650, opened 2026-06-15
Native reasoning_effort parameter support for the light LLM tier


tya5:
**[lead-coder]** — owner request (2026-06-15): native support for the **reasoning-effort** parameter on **light (weak-tier) LLMs**.

## Current state
- Model tiers: `light`/`standard`/`strong` (model_resolver.py:28), user-mapped in reyn.yaml; `gemini-flash-lite`→`gemini/gemini-2.5-flash-lite` (builtin_models.py:52).
- Provider thinking/reasoning is currently controlled by **hardcoded per-model `extra_body`** (builtin_models.py: `thinking_config.thinking_budget=0`, claude `thinking`) — NOT a first-class `reasoning_effort` knob.
- `reasoning_effort` exists in llm.py:219 only as a **response**-side field (completion_tokens_details), not a request param.
- Call path: model def kwargs / `extra_body` → llm.py:858-867 → `litellm.acompletion`.

## Goal
Wire **`reasoning_effort`** (litellm's normalized low/medium/high request param, which maps to each provider's native thinking-budget) as a **first-class, config-declarable** knob, applied to the light tier — 'native' = use litellm's standardized `reasoning_effort` passthrough rather than hand-rolled `extra_body` thinking hacks.

## Design-first (core LLM path → propose before implementing; lead review gate)
Open design questions for the proposal:
1. **Knob location**: per-model def (builtin_models + reyn.yaml model mapping + ModelSpec field) [lead lean — mirrors the existing extra_body pattern, reyn.yaml is where models are configured] vs per-tier default vs per-call.
2. **litellm native support**: confirm `reasoning_effort` is accepted + provider-mapped for the light models actually in use (gemini-flash-lite / the proxy's current lite model); confirm it doesn't collide with the existing `thinking_config.thinking_budget=0` (which DISABLES thinking — reasoning_effort would re-enable it; resolve the interaction).
3. **Default**: opt-in (unset = no behavior change) vs a default effort for light [owner asked — see below].
4. Validation: reyn.yaml schema + a round-trip test with a NON-DEFAULT effort value reaching the litellm kwargs (not silently dropped).

Dispatched to e2e-coder (owns the llm/chat call path). Design proposal → lead review → implement.

tya5:
[...]

PR #1651, opened 2026-06-15
feat(1650): per-model reasoning_effort reyn.yaml field (validated, proxy-safe)

tya5:
**[e2e-coder]** — Owner feature: native, config-settable `reasoning_effort` per model definition. Implements approved **design (B)** — a validated recognized field over the existing `ModelSpec.kwargs` passthrough.

## What it does
```yaml
models:
  light:
    model: gemini-flash-lite
    reasoning_effort: low      # minimal | low | medium | high | disable | none
```
- **Validation (fail-fast)**: `ModelSpec.__post_init__` validates the value at config-load — an invalid value raises a clear `ValueError` at startup, not mid-call inside litellm. Setting both `reasoning_effort` and an `extra_body` thinking config is rejected (contradictory controls). Placed in `__post_init__` so **both** the plain-dict and extends-merge construction paths are covered by-construction (single validation site, no parallel drift).
- **Native passthrough**: the value lands in `spec.kwargs` and rides the established passthrough to `litellm.acompletion`, which maps it to the provider's native thinking budget (verified live, litellm 1.84.0 gemini: `low`→1024, `medium`→2048, `high`→4096, `minimal`→512 flash-lite, `disable`/`none`→0).

## Proxy passthrough — the load-bearing live-smoke caught this
reyn's proxy path forces `custom_llm_provider=openai`, under which litellm **rejects** `reasoning_effort` as an unsupported OpenAI param for a gemini model name *before* forwarding (`UnsupportedParamsError`). The monkeypatched unit test could not catch this (it bypasses litellm validation) — only the live proxy smoke did. Fix: `recorded_acompletion` (the single litellm chokepoint, covering `call_llm` + `call_llm_tools`) now whitelists it via `allowed_openai_params` when set, so litellm forwards it to the proxy, which maps it.

**Live proxy smoke (gate #1):** reasoning_tokens `0` (baseline) → `~420` (effort=low) through the owner's proxy → gemini. ✓

## Corrections to the original design premises (primary-source, confirm-first)
- **No `thinking_budget=0` collision on the light model** — that config is on the untiered `gemini-2.0-flash`, not `gemini-flash-lite`; the call path forces only `stream=False`. Precedence is trivial.
- **Already settable via long-form kwargs** — the real gap was the short-form mapping + no validation + discoverability (+ the proxy block above).

## Known behaviors documented (NOT scope-crept into capture, per lead)
- `includeThoughts=true` → response carries reasoning text; reyn records only the token-count split, does not surface the reasoning text.
- Re-enables thinking on the tool-use path (Gemini #17949 context).

Refs #1650

## Test plan
- [x] config round-trip (non-default value) → spec.kwargs + through resolver startup
- [x] load-validation fail-fast (direct ModelSpec + resolver startup; invalid → ValueError)
- [x] all valid values accepted (no false reject); both-set reject
- [x] arrives at litellm.acompletion (the #1646 dropped-key lesson)
- [x] proxy whitelist (`allowed_openai_params`) threaded on the openai-compat path
- [x] **live proxy smoke**: reasoning_tokens 0→~420 reyn→proxy→gemini (gate #1)
- [x] 1126 LLM/replay/model/router regression pass; tier-audit + ruff clean
- [x] docs: reyn-yaml.md + .ja.md (field + valid values + native mapping + both known behaviors + proxy)

### DIFF
--- a/docs/reference/config/reyn-yaml.ja.md
+++ b/docs/reference/config/reyn-yaml.ja.md
@@ -117,12 +117,34 @@ models:
 | `max_tokens` | いいえ | レガシーのソフトヒント — 多くのプロバイダーが無視する。`max_completion_tokens` を推奨。 |
 | `top_p` | いいえ | litellm に渡す top-p サンプリング。 |
 | `extra_body` | いいえ | プロバイダー固有のペイロード（例：推論モデルの `thinking`）。 |
+| `reasoning_effort` | いいえ | モデルの推論バジェット: `minimal` / `low` / `medium` / `high` / `disable` / `none`。**ロード時にバリデーション**（下記参照）。 |
 | `extends` | いいえ | 名前付きクラスから継承し、オーバーライドを deep merge（下記参照）。 |
 | *（その他のフィールド）* | いいえ | litellm にそのまま渡されます（パススルーポリシー）。 |
 
 > **コスト制限**: `max_tokens` ではなく `max_completion_tokens` を使用してください。`max_tokens` は多くのプロバイダーが無視するレガシーのソフトヒントです。`max_completion_tokens` は API レベルで強制されます（OpenAI o1+ および Anthropic モデル）。
 
-**フィールドポリシー**: `model` のみ必須です。他のフィールドはすべてバリデーションなしで `litellm.acompletion` に直接渡されます（未知のフィールドも silent に転送されます — future-proof）。タイポは reyn エラーではなく silent な litellm 失敗を引き起こします。
+**フィールドポリシー**: `model` のみ必須です。ほとんどのフィールドはバリデーションなしで `litellm.acompletion` に直接渡されます（未知のフィールドも silent に転送されます — future-proof）。タイポは reyn エラーではなく silent な litellm 失敗を引き起こします。唯一の例外は `reasoning_effort` で、ロード時にバリデーションされます（下記）。
+
+### `reasoning_effort`（モデルごとの推論バジェット）
+
+モデルが回答前にどれだけ「思考」するかを設定します。分かりやすさのためモデル定義ごとに宣言します:
+
+```yaml
+models:
+  light:
+    model: gemini-flash-lite
+    reasoning_effort: low      # minimal | low | medium | high | disable | none
+```
+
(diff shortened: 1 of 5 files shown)


tya5:
**[lead-coder]** — review **PASS**, auto-merge enabled.

## Implementation (by-construction, single-site)
- **Proxy whitelist** (llm.py): applied only when `reasoning_effort` is set, appends to existing `allowed_openai_params` (idempotent), single chokepoint = both call sites. The comment documents the root cause (proxy forces custom_llm_provider=openai → litellm rejects reasoning_effort pre-forward) + the live fix.
- **Validation** (model_resolver `__post_init__`): `VALID_REASONING_EFFORTS` frozenset, fail-fast at load with a helpful message; in `__post_init__` so BOTH plain-dict and extends-merge paths are covered by-construction — single site, no parallel drift (the #1646 anti-drift lesson applied). Both-set reject handles `thinking_config`/`thinkingConfig` variants.

## The load-bearing gate paid off
The **proxy live-smoke** caught a REAL blocker the monkeypatch unit could not (the unit bypasses litellm's param validation; the proxy path's `UnsupportedParamsError` only surfaced live). Live-verified reasoning_tokens **0→~420** through the owner's actual proxy→gemini. This is exactly why I made the live-smoke a required gate, not optional — and why I'll keep doing so for optional-dep/proxy glue.

## Tier rule check ✓
14 tests; tier-audit --strict + ruff green (docstring tiers enforced by the audit). Recording-at-the-litellm-boundary (not MagicMock behavior-faking) is the policy-compliant seam — same pattern as #1646's recording host; the live-smoke is the real integration proof. Non-default round-trip + load-validation + both-set reject + proxy-whitelist-threaded axes covered.

## Scope hygiene ✓
Gate-2 known behaviors documented honestly (reasoning TEXT dropped → #1652; #17949 thinking-re-enable on tool-use; proxy passthrough). No scope-creep into reasoning-text capture (that's #1652). Two confirm-first corrections in the PR body.

Auto-merge (squash) on pytest-green → completes #1650. Owner can then set `reasoning_effort` per-model; the weak-dogfood on/off pass-rate experiment is the natural next measurement (offered to owner).

Issue #1649, opened 2026-06-15
TUI: max_iterations (5) auto-denies instead of prompting 'Allow more iterations?' (on_limit interactive degrades)


tya5:
**[lead-coder]** — owner-reported (2026-06-15): in the TUI, reaching the router `max_iterations` limit (default **5**) **auto-denies / force-stops** instead of prompting to continue. Owner doesn't recall configuring this. **Framed as hypothesis pending confirmation** (owner hasn't checked their reyn.yaml).

## Code-level map
- `router_loop.py:1295` `max_iterations: int = 5` (matches 'maxturn 5回').
- At exhaustion (`router_loop.py:2570+`): `if self._on_limit is not None:` → `handle_limit_exceeded` prompts "Allow more iterations?" via the intervention bus; if `allow_continue` → extend, **else break → limit-deny path (#1496, line 2604+)**: emits `limit_denied` + forced wrap-up = the owner's 'auto-deny'.
- **The default SHOULD prompt:** `SafetyConfig.on_limit = field(default_factory=OnLimitConfig)` (config.py:360) and `OnLimitConfig.mode = "interactive"` (config.py:327 — 'pause the run, prompt the user'). on_limit IS threaded (session.py:1459/2002, router_loop_driver.py:339). So the config default is interactive → the owner should see a prompt.

## Hypothesis (the degradation)
The interactive prompt requires an **intervention bus** on the router_loop host: `router_loop.py:2573` `_bus = getattr(self.host, "make_intervention_bus", lambda: None)()`. If the **TUI host doesn't wire `make_intervention_bus`** (→ `_bus=None`), `handle_limit_exceeded(bus=None, ...)` likely cannot prompt → returns `allow_continue=False` → break → auto-deny. I.e. interactive **silently degrades to deny** in the TUI.

## Investigation (confirm-first)
1. Confirm the owner's effective `safety.on_limit.mode` (unattended would explain it as config, not a bug — see owner Q below).
2. Trace the TUI path: does the TUI session's router_loop host implement `make_intervention_bus`? Does `handle_limit_exceeded` with `bus=None` deny vs prompt? Does the TUI render the FP-0005 limit-extension intervention at all (vs ask_user)?
3. Root-cause: if the bus isn't wired for the TUI limit-prompt (vs the ask_user path which works), wire it so interactive on_limit prompts in the TUI; the limit-deny path should only fire for genuine `unattended`/exhausted-extension.
[...]

PR #1656, opened 2026-06-15
fix(1649): non-TTY interactive limit no longer silently auto-denies (bounded auto-extend + loud non-zero abort)

tya5:
**[e2e-coder]** — Fixes the owner's "single agent silently auto-denies / stops" frustration. tui-coder's controlled 3-way primary evidence: a **non-TTY / run-once / piped** run at the router iteration limit **silently refused + exited 0** (empty stdout+stderr — undetectable). The interactive TTY path is healthy (don't regress).

## Root cause
`handle_limit_exceeded` (`limit_handler.py`), in `mode=interactive` (the default), couldn't ask the user in a non-TTY context → refused (`no_bus` / EOFError→`user_refused`) → the limit-deny path's `kind="error"` message was **dropped** by `_run_once` (which returned only the `kind="agent"` reply) and exited 0 → silent.

## Fix (intent-preserving + always detectable)

**PART A — non-TTY interactive → bounded auto_extend (not silent refuse).**
`handle_limit_exceeded` gains `non_interactive` (threaded from the session's `isatty()`-derived flag via `router_loop`). When `mode=interactive` AND `non_interactive` (can't ask): fall back to a **bounded** auto_extend (shared `_bounded_auto_extend`, capped by `auto_extend_times`) instead of a silent refuse. Rationale: the operator chose interactive = "continue when asked"; with no TTY the intent-preserving behavior is to **extend → make progress → complete**, not stop. The interactive TTY path (`non_interactive=False`) is **untouched** (regression-guarded test).

**PART B — exhausted abort: surface a decision-enabling message + non-zero exit.**
When the bounded extensions ARE exhausted (genuine runaway): the limit-deny outbox carries a `limit_stopped` marker; `send_to_agent_impl` detects it → **surfaces** the decision-enabling message (even when the history harvest is empty — `kind="error"` isn't persisted) + returns `limit_stopped`; `_run_once` returns the full result dict; the run-once CLI writes the reply (**never silent**) then **exits non-zero (2)** so wrappers/CI detect the runaway-stop.

Net: non-TTY interactive now **extends + completes** (common case) or **aborts loudly + non-zero** (exhausted) — never a silent exit-0 stop.

## Owner-morning note (flag, not decided)
The default `safety.loop.max_router_iterations=5` may be **too low** for real agentic tasks — the owner hit it. Separate from this fix (which makes hitting the limit graceful, not silent), but worth raising for an owner decision.

Refs #1649

## Test plan
- [x] non-TTY interactive → bounded auto_extend (continuation), then loud distinct `interactive_no_tty_exhausted` on exhaustion (not silent)
- [x] interactive TTY path unchanged (still asks; bus=None still `no_bus`) — regression guard
- [x] `_run_once` returns the result dict incl `limit_stopped` → caller exits non-zero; message surfaced
- [x] 1445 limit/router/chat/mcp/session regression pass; ruff + tier-audit clean
- [ ] **(tui-coder, post-fix)** tmux 3-way live re-verify (interactive prompts / non-TTY extends-or-loud-aborts / no silent exit-0)

### DIFF
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -2781,7 +2781,15 @@ class RouterLoop:
                 f"Configure safety.on_limit.mode=interactive or auto_extend to "
                 f"extend, or increase safety.loop.max_router_iterations."
             ),
-            meta={"chain_id": self.chain_id},
+            # #1649 PART B: the limit_stopped marker lets the run-once / A2A
+            # caller detect a limit-abort (vs a normal reply) → surface the
+            # message + exit non-zero, so a non-TTY wrapper never sees a silent
+            # exit-0 stop. Mirrors the wrap-up agent message's marker above.
+            meta={
+                "chain_id": self.chain_id,
+                "limit_stopped": True,
+                "limit_kind": "max_iterations",
+            },
         )
         return self._total_usage
 
--- a/src/reyn/chat/router_loop.py
+++ b/src/reyn/chat/router_loop.py
@@ -2709,6 +2709,10 @@ class RouterLoop:
                 ),
                 detail=f"chain_id={self.chain_id}",
                 extension_amount=float(self.max_iterations),
+                # #1649: non-TTY/run-once → handle_limit_exceeded falls back to
+                # bounded auto-extend instead of a silent refuse (the agent
+                # completes instead of an empty exit-0 stop).
+                non_interactive=self._non_interactive,
             )
             if _dec.allow_continue:
                 self.max_iterations += int(_dec.extension)
--- a/src/reyn/cli/commands/chat.py
+++ b/src/reyn/cli/commands/chat.py
@@ -28,16 +28,20 @@ from ..session import Session
 _ONCE_SEND_TIMEOUT = 3600.0
 
 
-async def _run_once(agent_registry, agent_name, *, instream=None, send=None) -> str:
+async def _run_once(agent_registry, agent_name, *, instream=None, send=None) -> dict:
     """#187 one-shot drive: read the WHOLE *instream* (default stdin) as a SINGLE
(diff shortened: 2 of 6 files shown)


tya5:
**[lead-coder]** — logic review **PASS**; **holding merge for tui-coder's post-fix tmux 3-way** (the owner explicitly asked for tmux live-verification of #1649, so the fix gets the same live confirmation, not just CI).

## Verified
- **PART A** — non-TTY interactive → `_bounded_auto_extend` (capped by `auto_extend_times`, process-local `_auto_extend_used` reset at turn boundary) → agent extends + completes; exhausted → loud `interactive_no_tty_exhausted`. The gate = `non_interactive = not sys.stdin.isatty()` (chat.py:424), threaded chat→session→router_loop→handle_limit_exceeded — **deterministic isatty, bus-independent**, so it covers the owner's non-TTY chat regardless of bus AND leaves the healthy TTY path (non_interactive=False) untouched. ✓
- **PART B** — `limit_stopped` marker (router_loop:2770/2790 → session:3626 → chat.py) surfaces the decision-enabling message even on empty history harvest, then `sys.exit(2)` ('a limit hit must never be silent; exit(2) distinguishes from arg/usage errors'). Never silent, detectable by wrappers/CI. ✓
- Gates: not-silent ✓ / auto-extend-continuation ✓ / non-zero-exit ✓ / interactive-not-regressed ✓. Tests: test_run_once_187 (real run-once exit path = the actual non-TTY scenario, not stubbed) + test_safety_limit_handler; 1445 pass; ruff + tier-audit green.

## Merge gate
Holding for tui-coder's tmux 3-way re-verify on this branch (owner wanted tmux confirmation). On their green → I merge. Owner-morning note (max_router_iterations=5-too-low) correctly flagged for the owner, not decided here.

PR #1657, opened 2026-06-15
feat(1657): default chat tool-use scheme → enumerate-all (owner H1 fix; 30%→100% tool-use)

tya5:
**[e2e-coder]** — Owner-decided H1 fix: **default the chat tool-use scheme `universal-category` → `enumerate-all`**. Flat-listing actions lets the LLM invoke them directly instead of hallucinating `invoke_action` names (the discover-then-call indirection) — root-caused as the H1 cause of ~30%→100% non-hot-list tool-use. The owner's #1 "unusable" fix.

## Change
- `DEFAULT_SCHEME_NAME` (scheme.py) → `enumerate-all` (the router fallback when a layer name is unset).
- `ToolUseConfig.chat` default + `_build_tool_use_config` loader → `enumerate-all`. **`step` / `phase` keep `universal-category`** (the H1 evidence is the chat path).
- `session.py` `chat_tool_use_scheme` default → `enumerate-all` (matches the config).
- **Completeness (the #1646 lesson)**: classified every `resolve_model`/scheme call-site. The 5 entry callers (chainlit/chat/dogfood/mcp/web) + scoped_session_factory all thread `config.tool_use.chat` — **none hardcode a literal**, so the switch propagates cleanly; router_loop's fallback is `DEFAULT_SCHEME_NAME`.

`reyn.yaml` override still works (`tool_use.chat: universal-category`); enumerate's no-tool_calls guard is robust (#1641) — no ∞-loop risk.

## Tests
- Flipped the default-assertion tests (DEFAULT_SCHEME_NAME / resolve / `ToolUseConfig.chat`) to `enumerate-all`; rewrote the per-layer test for the **inverted contract** (default=enumerate-all + universal-category override still works) — so the default can't silently regress.
- **Blast-radius fix (14 tests)**: pinned the universal-category-specific tests to `chat_tool_use_scheme`/`scheme_name="universal-category"` (replay_skill_router ×7, routing_decided ×4 [invoke_action/hot-list events], router_loop_chatsession ×2, codeact ×1). They test universal-category behaviour / use universal-category-shaped stubs+fixtures — still a valid config; enumerate-all has its own coverage (`test_enumerate_all_scheme`).
- **Full suite**: the only remaining 2 failures (`swebench_missing_honest_skip`, `web_fetch no_media_store`) are **pre-existing on clean origin/main** (verified via worktree — zero overlap with this diff).

## Docs
[...]

PR #1658, opened 2026-06-15
docs(tool-use-schemes): add H1 root-cause + quantified enumerate-all trade-off

tya5:
**[docs-maintainer]** — Follow-up to #1657 on `docs/concepts/tools-integrations/tool-use-schemes.md`, per lead-coder dispatch (correct the deep trade-off claim with H1 findings).

## Finding first: #1657 already fixed the main worry
The dispatch was premised on the doc still saying "enumerate-all = small-set-only / NOT a weak-model aid". **On current `main` that text is already gone** — #1657 made `enumerate-all` the chat default, dropped the "hardest surface for weak models" claim, and added the ~30%→100% framing. So no contradiction remains to remove.

## What #1657 did NOT capture (this PR adds)
1. **Cross-scheme root cause (H1):** tool-name *visibility* predicts invocation success — `enumerate-all` / `CodeAct` expose the callable name directly → invoke without guessing; `universal-category` (discover→invoke) / `retrieval` (search-first) hide it behind an indirection → name-hallucination on non-hot-list tools. Added as a one-paragraph lead-in under "The four schemes".
2. **Quantified enumerate-all trade-off:** reframed from a vague "request size grows" to a **visibility cost, not a weak-model penalty** — linear in catalogue size (H1 measured ~67 tools ≈ ~50KB, ~3.2× the `universal-category` request), biting only at very large catalogues.

Quantified figures are attributed to the H1 measurement (not asserted as universal law). No `.ja` mirror exists for this doc, so scope is the single `.md`.

## Test plan
- [x] `mkdocs build --strict` → exit 0 (verified)
- [x] (skipped — nav unchanged) no nav entry added/removed; edits are in-section prose only

### DIFF
--- a/docs/concepts/tools-integrations/tool-use-schemes.md
+++ b/docs/concepts/tools-integrations/tool-use-schemes.md
@@ -20,6 +20,14 @@ what is allowed, only how the LLM is asked to express a call. See
 
 ## The four schemes
 
+A cross-scheme finding (H1): **tool-name visibility predicts invocation
+success**. Schemes that place the callable name directly in the LLM-facing
+surface (`enumerate-all`, `CodeAct`) let the model invoke without guessing;
+schemes that put the name behind an indirection it must first traverse
+(`universal-category`'s discover→invoke, `retrieval`'s search-first) invite
+name-hallucination on non-hot-list tools. This is why the chat default moved to
+`enumerate-all` (#1657).
+
 ### `enumerate-all` (chat default, #1657)
 
 Presents *every* usable tool flatly in the LLM's tool list and dispatches by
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — review **PASS**, auto-merge enabled. Both H1 additions are accurate + correctly framed:
- **Cross-scheme root-cause**: tool-name VISIBILITY predicts invocation success (visible=enumerate/CodeAct→no-guess; hidden-behind-indirection=universal/retrieval→hallucinate) — matches the H1 unifying finding exactly, references #1657.
- **Trade-off**: correctly reframed as a **visibility COST (linear SP-size, H1-measured ~50KB/3.2×), NOT a weak-model penalty** — 'only bites at very large catalogues'. The corrected framing.
- Both attributed as H1 MEASUREMENTS (observed, not universal law) — the right epistemic framing.

Good premise-correction: you verified current main and found #1657 had already removed the stale 'enumerate=hardest-for-weak-models' claim — so my dispatch was based on a #1657-pre state (e2e's flag was stale, I relayed it without re-checking). Exactly the 'audit current refs not stale' discipline; #1658 correctly scopes to just the 2 genuine additions. mkdocs --strict + ruff + tier-audit green. Auto-merge (squash) on pytest-green.

Issue #1673, opened 2026-06-16
Tool-spawned skill runs pass resolver=None + hardcoded model="standard" (latent literal-model bug + #1672 CAT-3 config-ification)


tya5:
**[e2e-coder]** — fast-follow split from #1672 (CAT-3). Parent: #1672.

## Two coupled problems at the tool→skill spawn boundary

The 5 tool sub-loops that spawn a skill run (`tools/invoke_skill.py:196`, `tools/sandboxed_exec.py:114`, `tools/web_search.py:90`, `tools/lint.py:84`, `tools/mcp.py:306`) build their `OpContext` with **`model="standard"` AND `resolver=None`**, and `ToolContext` carries no resolver field to give them a config-aware one.

### 1. Latent bug (higher value): a literal model name can reach litellm

`op_runtime/run_skill.py:151-156`:
```python
if ctx.resolver is not None:
    model = ctx.resolver.resolve_class_or_fallback(op.model, ctx.model, ...)
else:
    model = op.model or ctx.model or "standard"   # ← resolver=None path
```
With `resolver=None` (the tool-spawn path) and no `op.model`, `model` becomes the class string **`"standard"`**, and the spawned runtime also gets `resolver=None` (`ModelResolver({})`), so `"standard"` is passed through as a **literal model name to `litellm.acompletion`** — litellm does not know `"standard"`. Tool-invoked skills with no explicit `op.model` can therefore send an unresolvable model name. (Same shape for the judge `or "standard"` resolver-None branch.)

### 2. #1672 CAT-3 config-ification

Because there's no resolver, these tool sub-runs cannot follow `model_class_by_purpose` / `config.model` — they're stuck on the literal `"standard"`. To fix, thread a config-aware resolver in.

## Fix (one focused PR + tests)

1. Add a `resolver` field to `ToolContext` (`tools/types.py`); populate it at the 4 `ToolContext` build sites (`router_loop.py` ×3 — has `host.resolver`; `control_ir_executor.py` ×1 — has `self._resolver`).
2. Update the 5 tool handlers to pass `ctx.resolver` (not `None`) and `ctx.resolver.class_for_purpose("tool")` (not literal `"standard"`) into the spawned `OpContext`.
3. This both **fixes the latent literal-model bug** (a real resolver resolves the class) and **completes #1672 CAT-3** (tool sub-runs follow config).
[...]

PR #1677, opened 2026-06-16
feat(1676): capture LLM call exceptions as a P6 llm_request_error event

### LINKED ISSUE #1676: LLM call exceptions are not captured anywhere — emit llm_request_error event (capture full provider error body)
**[lead-coder]** — owner-directed (blocking his GPT-5.4 debugging). The exception-path counterpart to #1669.

## Problem (verified)

When an LLM call raises (e.g. a **405 Method Not Allowed** the owner is hitting), the exception propagates up from `recorded_acompletion` (llm.py ~880, the single acompletion chokepoint) **with no event emitted** — verified: there is no try/except or error emit around the acompletion call. There ARE error events elsewhere (compaction_failed, skill_run_failed) but **nothing for a per-call LLM error**. So the owner: "例外の詳細はどこにも残ってない" (the exception detail isn't recorded anywhere). Without it, the 405 can't be root-caused ([[feedback_capture_full_provider_error_before_root_cause]]).

## Fix — emit `llm_request_error` at the chokepoint (mirror #1669)

At `recorded_acompletion` (the #1669 chokepoint, ContextVar EventLog already wired), wrap the acompletion call (incl the `_once` retry) in try/except. On exception, emit a P6 `llm_request_error` event capturing:
- `model`, `purpose` (same context as the `llm_request` success event)
- `error_type` (exception class name), `error_message` (str(exc))
- **the FULL provider error body / status_code** — litellm provider exceptions (BadRequestError etc.) typically carry `.status_code` / `.message` / `.response`; capture the full body, **not truncated** (the 405's body is the root-cause signal). Redact secret-like fields.
- then **re-raise** (never swallow — the error must still propagate).

Then the error lands in the event log → visible in the TUI event tab (generic fallback immediately; nice rendering = tui-coder fast-follow). The owner can then read the 405's actual detail.

## Scope note

Primary = the chat acompletion chokepoint (covers router/purpose calls — the likely source). If the owner's 405 doesn't surface there, it's an embedding (litellm_provider) / MCP / tool HTTP call → extend capture to that path. Flag which path it turns out to be.

## Test (Tier-2, no Mock)

A chokepoint call where the (real-fake) acompletion raises → `llm_request_error` emitted with the error_type/message/status, the exception still propagates, secrets redacted.

tya5:
**[e2e-coder]** — high-priority, owner-blocking (the owner can't root-cause his persistent 405 because the exception detail is recorded nowhere). Closes #1676. The exception-path counterpart to #1669 (same chokepoint + ambient EventLog).

## Fix

At `recorded_acompletion` (the single `litellm.acompletion` chokepoint, #1190 AST-guarded), wrap the call (incl the `response_format`-fallback retry) in try/except; on a **final** failure emit a P6 `llm_request_error` and then **re-raise** (never swallow):

- `model` + `purpose` (same context as `llm_request`)
- `error_type` + `error_message` + `status_code`
- the **FULL** provider body + response text — **NOT truncated** (the 405's body is the root-cause signal). litellm exceptions carry `.status_code` / `.body` / `.response`.
- `params` (redacted, reusing the #1669 redactor)

Secret **values** (the proxy-injected `api_key` + known API-key env vars) are scrubbed from the freeform provider text; the audit emit is wrapped so it can never mask the real exception. Emitted **once** (the wrap covers the fallback retry → a final failure doesn't double-emit).

## Coverage / the "which path" flag (lead's ask)

`recorded_acompletion` is the single AST-guarded **completion** chokepoint → this covers **every completion call**. So if the owner's 405 is a completion call, it will now appear in the event tab with the full detail (which call / status / body). **If it does NOT appear, the 405 is on a different chokepoint** — most likely the **embedding** path (`litellm.aembedding` in `embedding/litellm_provider.py`, a separate uninstrumented chokepoint; MCP calls don't go through litellm). That cleanly narrows the root cause; a sibling `aembedding` instrumentation is a fast follow-up if needed.

## Test plan

- [x] `pytest -q` from rootdir → **7425 passed**, 14 skipped, 3 xfailed. The only 2 failures (`test_eval_benchmark_tier1_swebench::test_swebench_missing_honest_skip`, `test_web_fetch_preview_path_ref::test_legacy_no_media_store_path_returns_inline_content`) are **pre-existing on clean `origin/main`** (worktree-verified earlier this session; orthogonal modules). Rebased onto current main (post-#1673).
- [x] New `test_llm_request_error_1676.py` (4 Tier-2, no Mock — real `EventLog` + a real async fake for `litellm.acompletion` raising a litellm-shaped exception with `.status_code`/`.body`/`.response`): event emitted with full detail + **exception still propagates**; long provider body captured **whole** (exact-equality, not truncated); secret value echoed in the error text **scrubbed**; unset-ambient-log path still re-raises.
- [x] Regression: `test_llm_request_event_1669` (the #1669 happy path) + `test_cost_chokepoint_1190` + AST-guard + `test_llm_call_recorder_invariants` + `test_llm_call_retry` + `test_event_audit_invariants` all green.
- [x] Tier rule self-check: docstring ✓ / Tier 4 violations 0 ✓ / invariant 弱化なし ✓ / audit 0 errors ✓

## Coordination
- tui-coder rendering = fast-follow (the event tab's generic fallback surfaces the new `llm_request_error` kind raw immediately; a per-kind formatter/filter can follow, mirroring #1670).

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -897,6 +897,88 @@ def _redact_llm_request_params(base_kwargs: dict, response_format: dict | None) 
     return out
 
 
+# #1676: env vars whose values are API secrets — scrubbed from the (freeform)
+# provider error text so a captured 405/4xx body never leaks a key.
+_LLM_SECRET_ENV_VARS: tuple[str, ...] = (
+    "OPENAI_API_KEY", "ANTHROPIC_API_KEY", "GEMINI_API_KEY", "GOOGLE_API_KEY",
+    "LITELLM_API_KEY", "AZURE_API_KEY",
+)
+
+
+def _collect_secret_values(base_kwargs: dict) -> list[str]:
+    """#1676: the concrete secret VALUES to scrub from freeform provider error
+    text — the secret-keyed kwargs (e.g. the proxy-injected ``api_key``) + known
+    API-key env vars. Scrubbing the actual value is precise (vs guessing patterns
+    in arbitrary provider output)."""
+    vals: list[str] = []
+    for k, v in base_kwargs.items():
+        if isinstance(v, str) and v and any(h in k.lower() for h in _LLM_REQUEST_SECRET_HINTS):
+            vals.append(v)
+    for env in _LLM_SECRET_ENV_VARS:
+        val = os.environ.get(env)
+        if val:
+            vals.append(val)
+    return vals
+
+
+def _scrub_secrets(value: object, secrets: list[str]) -> object:
+    """#1676: replace any known secret value occurring in a string with a marker.
+    Non-strings pass through unchanged (dict/list provider bodies are kept whole —
+    providers do not echo your API key in the error body, and the full body is the
+    root-cause signal we must NOT truncate)."""
+    if not isinstance(value, str) or not secrets:
+        return value
+    for s in secrets:
+        if s:
+            value = value.replace(s, "***REDACTED***")
+    return value
+
+
+def _emit_llm_request_error(
+    model: str, purpose: str, exc: BaseException, base_kwargs: dict,
+) -> None:
+    """#1676: emit a P6 ``llm_request_error`` with the FULL provider error detail
+    (status_code + whole message/body, NOT truncated — the owner's 405 root-cause
+    signal) so an LLM-call failure is visible in the event tab. Same ambient
+    EventLog (ContextVar) + ``model``/``purpose`` context as ``llm_request``
+    (#1669). Secret values are scrubbed from the freeform text. Wrapped so the
+    audit emit can never mask the real exception (the caller re-raises regardless)."""
+    try:
+        from reyn.events.events import get_llm_request_event_log
+        log = get_llm_request_event_log()
+        if log is None:
+            return
+        secrets = _collect_secret_values(base_kwargs)
+        detail: dict = {
+            "error_type": type(exc).__name__,
+            "error_message": _scrub_secrets(str(exc), secrets),
+            "status_code": getattr(exc, "status_code", None),
+        }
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Review: **PASS**, merge on green. Closes #1676. Owner-blocking — unblocks the 405 root-cause.

- Chokepoint wrap: the acompletion call (incl response_format fallback) wrapped; outer `except → _emit_llm_request_error → raise` — fires **once** after retries, **re-raises** (never swallows).
- `_emit_llm_request_error` captures `error_type` + scrubbed `error_message` + `status_code` + **both `.body` and `.response`, WHOLE/untruncated** (the 405 body is the root-cause signal — [[feedback_capture_full_provider_error_before_root_cause]]) + redacted params. The emit is itself try/except-wrapped so an audit failure can never mask the real exception.
- **Which-path binary** (smart): the AST-guarded chokepoint covers EVERY completion call → if the owner's 405 is a completion it now shows in the event tab; if it does NOT appear, it's the embedding path (`aembedding`, uninstrumented — a clean fast-follow). That cleanly narrows the root cause.
- Tests: 4 Tier-2, no Mock (real `_FakeProviderError` + monkeypatched acompletion). `test_error_event_emitted_and_reraised` pins **the owner's exact scenario** — 405 "Method Not Allowed" → event has model/purpose/error_type/status=405 + `pytest.raises` confirms propagation. Plus body-captured-WHOLE (exact-equality), secret-scrubbed, unset-log-still-raises.

## Tier rule check ✓
Tier-2 docstring'd; no Tier-4; no Mock; no private-state asserts (real EventLog public surface); the body-NOT-truncated invariant is exact-equality asserted.

2 pre-existing fails (swebench, web_fetch) accepted. Merging on green; tui rendering = fast-follow.

PR #1680, opened 2026-06-16
feat(1678): route reasoning_effort+tools to the OpenAI Responses API + decision-enabling 405 error

### LINKED ISSUE #1678: Reasoning models + tools route to /chat/completions and 405 — reyn has no Responses API routing (GPT-5.4 reasoning_effort + tool_choice)
**[lead-coder]** — owner-hit live (GPT-5.4 office test). Root cause web-verified + reyn-code-confirmed.

## Problem

The owner's **main** chat call — GPT-5.4, `tool_choice: auto`, `reasoning_effort: high` — returns **405 Method Not Allowed** from the company proxy's `/api` (chat/completions) endpoint.

**Verified root cause:** OpenAI does NOT support **function tools + reasoning_effort together on `/v1/chat/completions`** for GPT-5.4 — it requires **`/v1/responses`** (the Responses API). litellm should auto-route but has known bugs:
- litellm#23156: "Function tools with reasoning_effort are not supported for gpt-5.4 in /v1/chat/completions. Please use /v1/responses instead."
- litellm#24502: 405 from a mis-routed `/responses` call.

**reyn-side gap (grep-confirmed):** reyn has **no Responses API routing at all** — every LLM call goes through `litellm.acompletion` (= /chat/completions). So a reasoning model that also uses tools (the default-on reasoning for GPT-5.4, #1654, + the main chat's tool_choice:auto) hits the unsupported combo → 405. The reasoning-default-on (#1654) assumed /chat/completions works, but for tool-using calls on GPT-5.4-class reasoning models it needs /responses.

## Fix direction (confirm-first — proxy-dependent)

Route **reasoning-model + tools** calls through the **Responses API**. Options to investigate:
- litellm `openai/responses/<model>` model-name prefix, or `litellm.aresponses`, vs litellm auto-routing (the issues show auto-route is unreliable).
- **Must verify it works THROUGH a litellm proxy** (the owner's setup) — not just direct OpenAI ([[feedback_llm_provider_param_needs_proxy_smoke_allowed_openai_params]], [[feedback_verify_canonical_fix_before_synthesizing_workaround]]). The proxy must serve /v1/responses.
- Minimal reyn change: detect reasoning-model + tools-present at the call site → route to responses. Keep non-tool reasoning calls on the current path if they work.

## Interim (operator)

- `reasoning_effort: none` on GPT-5.4 → tools work on /chat/completions (loses reasoning), OR
- map the model to `openai/responses/gpt-5.4` if the proxy serves /responses.

Confirm-first the canonical responses-routing mechanism (web + a live proxy smoke) before implementing — do NOT synthesize a workaround.

tya5:
**[e2e-coder]** — owner 405 root-cause fix. Closes #1678. Confirm-first design approved by lead (mechanism + proxy-feasibility + the no-/responses decision-enabling addition); the design analysis is durable on the issue.

## Root cause (owner-live-confirmed)

`reasoning_effort` + `tools` together are only valid on the OpenAI **`/v1/responses`** endpoint, but reyn sent everything via `acompletion` (`/chat/completions`) → **405**. The owner confirmed it live: removing `reasoning_effort` cleared the 405.

## Fix (single `recorded_acompletion` chokepoint)

1. **Route** — when `tools` AND `reasoning_effort` are both present, rewrite the resolved model with the `responses/` bridge marker (`openai/gpt-5.4` → `openai/responses/gpt-5.4`; proxy post-strip `gpt-5.4` → `responses/gpt-5.4`, carried by reyn's `custom_llm_provider="openai"`). The litellm `openai/responses/` bridge routes to `/v1/responses` **but returns a chat-completions shape** → the existing chokepoint + response parsing are unchanged (**no parallel `recorded_aresponses`** — the load-bearing simplicity finding). litellm auto-route is unreliable (litellm#23156) → the prefix is applied **explicitly**. Idempotent (an explicit operator prefix is preserved).
2. **Decision-enabling error** — when reyn routed the combo to `/responses` and it still **405s** (the endpoint/proxy doesn't serve `/v1/responses` — **the owner's company-proxy case**), raise `ResponsesEndpointRequiredError` naming **both** remedies (set `reasoning_effort: none`, OR enable `/v1/responses` on the proxy) + the model, instead of a raw dead-end. The **#1676 `llm_request_error` event still captures the raw 405 detail first** (complementary: raw in audit, guided in the raised error). Only fires when reyn itself applied the route → a normal/unrelated 405 is unaffected.

Detection (`tools + reasoning_effort`) is exactly litellm's own auto-route trigger and the owner's live-confirmed condition. **All non-combo paths are byte-identical.**

## Proxy feasibility (from the confirm-first)
- reyn's proxy strip + `custom_llm_provider="openai"` preserves the `responses/` bridge marker (verified, litellm 1.84.0).
- A litellm proxy serves `/v1/responses` (local probe: POST → 400 = exists, vs 405 = absent).
- **The owner's company proxy does NOT serve `/v1/responses`** (lead-confirmed) → for him the **decision-enabling error is the load-bearing benefit**; the routing transform is correct for any `/responses`-serving env (direct OpenAI / proxies with it enabled).

## Test plan

- [x] `pytest -q` from rootdir → **7436 passed**, 14 skipped, 3 xfailed. The only 2 failures (`test_eval_benchmark_tier1_swebench::test_swebench_missing_honest_skip`, `test_web_fetch_preview_path_ref::test_legacy_no_media_store_path_returns_inline_content`) are **pre-existing on clean `origin/main`** (worktree-verified this session; orthogonal modules).
- [x] New `test_responses_api_routing_1678.py` (7 Tier-2, no Mock — real `recorded_acompletion` + a real async fake for `litellm.acompletion` capturing the model / raising a 405): the transform (insert-after-provider / prepend / idempotent); routed **only** on the combo (NOT tools-only or reasoning-only); the decision-enabling error on a `/responses` 405 naming **both** remedies + the model; #1676 still captures the raw 405; a non-routed 405 is NOT wrapped.
- [x] Regression: `test_llm_request_error_1676`, `test_llm_request_event_1669`, `test_cost_chokepoint_1190` + AST-guard, `test_llm_call_recorder_invariants`, `test_llm_call_retry` — all green.
- [x] Tier rule self-check: docstring ✓ / Tier 4 violations 0 ✓ / invariant 弱化なし ✓ / audit 0 errors ✓
- [x] (Live-final-verification = the owner's env) — his no-`/responses` company proxy can't be fully reproduced here; the transform is verified to construct the `responses/`-bridged model + the decision-enabling error is unit-pinned. The owner exercises the real proxy 405 → decision-enabling-error path.

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -979,6 +979,32 @@ def _emit_llm_request_error(
         pass
 
 
+class ResponsesEndpointRequiredError(Exception):
+    """#1678: raised when a ``reasoning_effort + tools`` call was routed to the
+    OpenAI ``/v1/responses`` endpoint but the configured endpoint/proxy does not
+    serve it (HTTP 405). Carries a decision-enabling message naming BOTH remedies
+    so the operator isn't left with a raw 405."""
+
+
+def _to_responses_model(model: str) -> str:
+    """#1678: rewrite a resolved litellm model string to route through the OpenAI
+    Responses API (``/v1/responses``) via the ``responses/`` bridge marker, which
+    litellm.acompletion honours and returns a chat-completions-shaped response for
+    (so reyn's response parsing is unchanged). Idempotent — an already-routed model
+    (explicit ``openai/responses/...`` prefix) is returned unchanged.
+
+    - ``openai/gpt-5.4``    → ``openai/responses/gpt-5.4`` (direct path)
+    - ``gpt-5.4``           → ``responses/gpt-5.4``        (proxy path, post-strip;
+      reyn's ``custom_llm_provider="openai"`` carries the provider)
+    """
+    if "/responses/" in model or model.startswith("responses/"):
+        return model
+    if "/" in model:
+        provider, rest = model.split("/", 1)
+        return f"{provider}/responses/{rest}"
+    return f"responses/{model}"
+
+
 async def recorded_acompletion(
     *,
     model: str,
@@ -1035,6 +1061,22 @@ async def recorded_acompletion(
             _allowed.append("reasoning_effort")
         base_kwargs["allowed_openai_params"] = _allowed
 
+    # #1678: route reasoning-model + tools calls to the OpenAI Responses API.
+    # ``reasoning_effort`` + ``tools`` together are only valid on /v1/responses
+    # (owner-confirmed: removing reasoning_effort cleared the 405) — but reyn
+    # sends everything via acompletion (/chat/completions) → 405. litellm's
+    # auto-route is unreliable (litellm#23156), so apply the ``responses/`` bridge
+    # prefix EXPLICITLY: it routes to /v1/responses yet returns a chat-completions
+    # shape, so the existing chokepoint + response parsing are unchanged (no
+    # parallel recorded_aresponses). An explicit operator prefix is preserved
+    # (idempotent). ``_routed_to_responses`` gates the decision-enabling error
+    # below so a normal 405 is unaffected.
+    _routed_to_responses = bool(
+        base_kwargs.get("tools") and base_kwargs.get("reasoning_effort")
+    )
+    if _routed_to_responses:
+        effective_model = _to_responses_model(effective_model)
+
     # #1669: emit a P6 ``llm_request`` event (TUI-observable) carrying the
     # non-message call params, ONCE here — before the ``_once`` response_format
     # retry loop, so a fallback retry does not double-emit. Ambient EventLog via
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Review: **PASS**, merge on green. Closes #1678. Final piece of the #1669→#1678 arc.

- **Route transform** `_to_responses_model`: `responses/` bridge (direct `openai/responses/...` + proxy `responses/...` post-strip), **idempotent** (explicit prefix unchanged), chat-completions shape → reyn parsing unchanged, no parallel chokepoint.
- **Detection gate**: `base_kwargs.get("tools") and base_kwargs.get("reasoning_effort")` — only the exact owner-confirmed combo routes; tools-only / reasoning-only are byte-identical. The `routed-only-on-combo` Tier-2 pins this.
- **Decision-enabling error**: `ResponsesEndpointRequiredError` raised only when `_routed_to_responses AND status_code==405`, names BOTH remedies (reasoning_effort:none OR enable /v1/responses) + the model. **#1676 captures the raw 405 first** (audit), then the actionable error raises — and the `#1676-still-captures` + `non-routed-405-not-wrapped` Tier-2 verify the ordering + the gate.
- 7 Tier-2, no Mock; pure addition (0 dels).

## Tier rule check ✓
Tier-2 docstring'd; no Tier-4; no Mock; no private-state asserts; gate-correctness + idempotence + the error-only-when-routed invariant are the assertions.

**Live-final verification = the owner's no-/responses proxy env** (correctly noted in the PR — we can't reproduce his proxy here; he'll see the actionable error instead of the raw 405). Good catch on the editable-install drift (bare `python -c` stale-tree vs `cd user && pytest`) — your runs were valid. Merging on green.

PR #1692, opened 2026-06-16
refactor(1682 #3): split config.py god-module into a config/ package

tya5:
**[e2e-coder]** — owner-approved #3 (config.py god-module split) from the #1682 src/reyn cleanup umbrella. Shim-first re-export pattern (same as #4 MCP). 2 staged commits (rename | split).

## What

The 2821-line `config.py` → a `config/` package grouped by **domain concern** (not yaml-key shape, per owner constraint):

| module | L | holds |
|---|---|---|
| `root.py` | 317 | ReynConfig + model/tier fields + `model_class_for` (+ R1 concrete section-config imports) |
| `loader.py` | 410 | `load_config` / `_merge` / `_load_yaml` / shape-wiring (yaml-coupled core) |
| `chat.py` | 725 | Reasoning/Chat/Loop/Compaction/Timeout/OnLimit/Safety (+ cost builders) |
| `embedding.py` | 445 | Embedding/SkillSearch/ActionRetrieval |
| `media.py` | 237 | Voice/Multimodal/Web/WebFetch |
| `execution.py` | 317 | Plan/SkillResume/SelfImprovement/TimeTravel/ToolUse |
| `infra.py` | 503 | Agent/Auth/Sandbox/Events/Eval/Cron/Python |

No single god-module remains. **0 call-site rewrites** (135 importers): the package `__init__` re-exports the union of every submodule's surface — **mechanically** (namespace copy, not a hand-list), so no de-facto-public private (`_find_project_root`=23 importers, the `_build_*`, constants like `_DEFAULT_EMBEDDING_CLASSES`) can be silently omitted (R4).

## 2 commits (rename ‖ split — diff the rename from the moves)
1. `git mv config.py → config/root.py` + package `__init__` — `git log --follow` confirms history preserved. All-in-root checkpoint (re-export keeps everything working).
2. Split `root.py` into the 6 section modules + wire cross-section imports.

## Hazards (all gated)
- **R1 (silent forward-ref)**: `root.py` CONCRETE-imports every section config it references → `get_type_hints(ReynConfig)` resolves. **GATE GREEN**: `test_config_schema_introspection` + `test_config_mirror_coverage_1056` (no dropped schema fields).
[...]

Issue #1682, opened 2026-06-16
src/reyn/ structure cleanup — audit findings (8) vs OSS conventions [umbrella]


tya5:
**[lead-coder]** — owner-directed structure cleanup (2026-06-16). Audit of `src/reyn/` vs typical OSS Python conventions, all from primary data (file counts / dir contents / line counts on origin/main). **Analysis only — no fixes started.** Owner prioritized #2 + #4 first; the rest tracked here so they're not forgotten.

## 🔴 Clear smells

1. **`safe/` vs `safety/` name-collision.** `safe/` (14 files) = utils grab-bag (json/text/time/hash/http/cache wrappers + safe/mcp/); `safety/` (2 files) = runtime limits (#1666 cap, limit_handler). Near-identical names, totally different concerns — indistinguishable to a reader.
2. **`chat/` = 112-file god-package** (3× the next: tools/ 37, cli/ 35). Breakdown: tui/ 39 + slash/ 27 + services/ 21 + 25 flat root modules. Conflates UI (tui/slash) with runtime orchestration (router/planner/services). **[OWNER PRIORITY — design in progress]**
3. **`config.py` = 2821-line god module** (+ separate config_schema.py 325). Single file is a smell; OSS splits config into a package or keeps it lean.
4. **MCP scattered across 4 locations**: `mcp_server.py` (root, 1015, the server), `mcp_client.py` (root, 509, the client), `safe/mcp/registry.py` (registry), `tools/mcp.py` (tool-defs). No single `mcp/` package. **[OWNER PRIORITY — design in progress]**

## 🟡 Structural quirks

5. **37 top-level subpackages = over-flat.** Typical OSS groups by domain into ~5-15.
6. **Over-packaging**: budget/ dispatch/ reporters/ safety/ schemas/ search_backends/ testing/ = each 2 files (one real module wrapped in a package).
7. **Flat-vs-package inconsistency**: 11 flat root modules (config, agent, context_builder, python_runner, user_intervention, intervention_choices, _cli, …) alongside 37 packages. `_cli.py` (12 lines) next to the `cli/` package.
8. **Security concerns scattered**: permissions/ secrets/ sandbox/ safety/ (+ the misleadingly-named safe/) as separate top-level dirs — candidate for a `security/` umbrella.

## Status

- #2 + #4 countermeasure designs: in progress (Plan sub-agents).
- #1, #3, #5–#8: tracked here for sequencing after #2/#4 (owner: "他の案件もその後進めたい").

[...]

PR #1694, opened 2026-06-16
fix(1625): scheme-agnostic catalog action descriptions (WHAT not HOW)

### LINKED ISSUE #1625: CodeAct: catalog action DESCRIPTIONS embed universal-wrapper idiom (invoke_action/describe_action)
**[e2e-coder]** — follow-up split out of #1618 root-3 (PR #1624) co-vet, per lead + dogfood-coder.

## Context
#1618 root-3 (PR #1624) makes the tool-use SP region scheme-replaceable: CodeAct's code-API replaces the universal `invoke_action`/`list_actions`/ROUTING-RULE vocab in the **SP region**, and a follow-up gate covers the `## Environment` `list_actions→invoke_action` idiom. The ② fence-compliance oracle is a decisive PASS (append-SP first-act fence 0/6 → replace-SP 5/5, success 0→100%).

## This issue (deeper, separate layer)
dogfood-coder's co-vet found a **second** residual universal-idiom class that root-3's region-replace does NOT reach: **some catalog action DESCRIPTIONS** (in the action registry) reference `invoke_action` / `describe_action` in their wording. These descriptions are rendered into the CodeAct code-API catalog (`_render_code_api`'s per-entry `desc`), so a CodeAct model reading the code-API still sees universal-world wording inside individual action descriptions.

This is **not a region-replace concern** (the SP region is correctly replaced) — it is **description-content** in the registry, a deeper layer. Did not break the 100% first-act fence rate at N=5, so non-blocking, but real (incomplete idiom-cleanup).

## Scope
- Audit action-registry descriptions for universal-wrapper idiom (`invoke_action`, `describe_action`, `list_actions`) baked into the description text.
- Decide the fix layer: scheme-neutral description wording (descriptions shouldn't assume a tool-use scheme), OR a scheme-aware description projection in `_render_code_api`.
- Holistic-design lens (#1618): prefer description wording that is **scheme-agnostic by construction** (the description says WHAT the action does, not HOW to invoke it — the HOW is the scheme's job), consistent with P7/P8.

## Trigger to prioritize
If the larger-N CodeAct residual measurement (root-3b) shows description-embedded idioms causing leaks. Until then: tracked, deferred.

Refs #1618.

tya5:
**[e2e-coder]** — #1682 cleanup umbrella, item #1625.

## What

Catalog **action descriptions** embedded the universal-wrapper invocation idiom
(`describe_action` / `invoke_action` / `list_actions`). Those strings render into
the CodeAct code-API + router catalog the LLM sees, so a description — which should
state only **WHAT** an action does (P8) — was leaking a scheme-specific **HOW**
(the universal scheme's call convention). Reword the 4 lead-audited **rendered**
descriptions to be scheme-agnostic by construction; the HOW is the active scheme's
job to inject, not the action's to hardcode.

| File | Constant | Before → After |
|------|----------|----------------|
| `tools/file.py` | `_WRITE_FILE_DESCRIPTION` | "`describe_action(file__edit)` for its args, then `invoke_action`" → "prefer the `file__edit` action instead of rewriting the whole file" |
| `tools/lint.py` | `_LINT_DESCRIPTION` | "the qualified action name **as returned by `list_actions(category=['skill'])`**" → "the qualified skill action name (= `skill__<name>`)" |
| `tools/recall.py` | `_RECALL_DESCRIPTION_HIDE_LEGACY` | "use `invoke_action(action_name='memory_entry__<name>')`" → "use the memory_entry actions (`memory_entry__<name>`)" |
| `tools/mcp_drop.py` | `server` param desc | dropped "Use `list_actions(category=['mcp.server'])` to discover" |

## Scope (lead-confirmed)

- **LEAVE `plan.py`** — its `tools` field is the plan-step **composition** format that
  names the active scheme's router tools; scheme-specific by design, not a leaf-action
  description.
- **LEAVE** `cron.py` / `mcp_drop.py` **module docstrings** — dev-facing, not rendered
  into the catalog.
- **Do NOT touch** the replay-pinned `recall._RECALL_DESCRIPTION` (byte-identical for
  fixtures).

## Replay fixture — surgical RE-KEY (not re-record)

`_WRITE_FILE_DESCRIPTION` is shared into the **phase** catalog, so the reworded `tools[]`
payload changes the request **key** for `phase_op_loop_byte_identical.jsonl` but **not**
the response (the model "immediately finishes" because the phase TASK says so — the
behaviour is independent of the description text). Per lead's Option A: recompute the
2 entry keys from the new request, keep the recorded responses verbatim. Diff is a
minimal **2-line** change (the 2 `key` fields only); `test_routerloop_convergence_replay_1092`
replays **clean** (byte-identical-convergence intent preserved).
`test_file_unified::test_write_file_router_render_exact_description` (the description's
own drift-guard pin) updated to the reworded text.

## Noted but LEFT (flag for your call)

`tools/lint.py` `_LINT_PARAMETERS["skill_path"]` description still says
"(returned by `list_actions`)" — the **same leak class** as the 4 above, but at the
**parameter** level rather than the action level. It was outside the confirmed 4, and
touching parameter descriptions risks cascading into **router-catalog** replay fixtures
(lint is `router="allow"`), which would balloon a deliberately-small PR. Left as-is for
a #1625 follow-up — say the word and I'll fold it in (with the consequent re-key) or
file the follow-up issue.

## Test plan

- [x] `ruff check src tests` (full scope, post-rebase onto #1692) — clean
- [x] `scripts/test_tier_audit.py --strict tests/test_file_unified.py` — 0 violations
- [x] full `pytest -q` from rootdir — (result in final comment)
- [x] `test_routerloop_convergence_replay_1092` replays clean with the re-keyed fixture

Closes #1625


tya5:
**[lead-coder]** — Review: **PASS**, merge on green. Closes #1625. The 4 lead-audited RENDERED descriptions (file/lint/recall/mcp_drop) reworded scheme-agnostic (idiom removed from the code-API surface); plan.py correctly left (plan-step composition format), module docstrings + the replay-pinned recall desc left. The replay re-key is **exactly Option A** — verified: only the 2 `key` hashes recomputed for the new request, the response objects byte-identical verbatim (same id/content/usage); test_routerloop_convergence_replay_1092 replays clean. The 2 local fails are the known pre-existing pair (swebench-honest-skip + web_fetch-heading, env-dependent; CI on main b46716aa green) — not the diff. ## Tier rule check ✓ — behavior is the rendered-catalog wording; no test-logic weakening; the re-key preserves the byte-identical-determinism intent. The param-level lint.py skill_path leak you flagged → I'm filing a tracked #1625-followup (deferred for the separate param-desc replay-cost assessment). Merging on green.

Issue #1693, opened 2026-06-16
LandlockBackend is non-functional against the real landlock pkg — written against a nonexistent API (add_path_beneath_rule)


tya5:
**[lead-coder]** — caught by sandbox_2's #1527 live-verify (Linux/colima 6.8, landlock ABI present). Real bug, primary-evidence.

## Finding
Running the live cancel scenario → `LandlockBackend.run()` → `RuntimeError: failed to build Landlock ruleset: 'Ruleset' object has no attribute 'add_path_beneath_rule'`.

**Root cause**: `src/reyn/sandbox/backends/landlock.py` calls `ruleset.add_path_beneath_rule(...)` + `landlock.Ruleset(handled_access_fs=...)` — but the **published `landlock` PyPI pkg has NONE of these**. Across every published version (1.0.0.dev1–dev4) the real API is the fluent **`Ruleset().allow(path, rules=...).apply()` / `.restrict_rules()`**, NOT `add_path_beneath_rule`. The backend was written against an API that doesn't match the shipped package.

**Secondary**: the pin `landlock>=1.0` doesn't install without `--pre` (only pre-releases exist on PyPI) → fresh `pip install` leaves landlock absent → `available()` = False.

## Impact
- **Landlock sandboxing is non-functional on Linux** (the security-sandbox gap).
- Explains the **codeact-no-working-sandbox-in-Linux-container** note (#1663): the Landlock branch was fail-closed because the backend can't even build a ruleset.
- The #1527 subprocess-cancel path is effectively **dead code** (unreachable — ruleset-build fails first). #1527 can't be verified until the backend builds at all.

## Why unit tests missed it
`test_sandbox_landlock.py` is monkeypatched (fakes name/protocol/available=False) → the real pkg-API mismatch never surfaced. Classic [[feedback_fake_backend_unit_misses_real_integration]] / [[feedback_optional_dep_glue_needs_real_env_verify]] — the live verify is what caught it.

## Fix
Rewrite `landlock.py`'s ruleset-build to the actual `landlock` pkg API (`Ruleset().allow(path, rules=...).apply()` style — sandbox_2 has the per-version API dump) + fix the pin (no stable ≥1.0 exists; pin `>=1.0.0.dev4` + allow-pre). Then the cancel path becomes reachable and #1527 can re-verify. **Live verify on Linux (sandbox_2) is the gate** — monkeypatched units can't catch this.

[...]

Issue #1527, opened 2026-06-12
Follow-up: verify Landlock backend subprocess-cancel path on Linux (untested in #1470)


tya5:
**[lead-coder]** — #1470（subprocess-cancel）follow-up tracking。

#1470（PR #1526, merged 648af7f1）で cancel_inflight() の SIGTERM/SIGKILL 伝播を実装。**Landlock backend の cancel path は untested**:
- Noop = Linux CI verified ✅ / Seatbelt = macOS verified ✅ / common-path(cancel_event=None) = byte-identical no-regression ✅
- **Landlock = honestly-marked untested**（landlock.py inline comment 済）。CI runner は Landlock LSM 不在で Noop default、e2e は macOS。logic は Seatbelt mirror。

## 対応
Linux 5.13+ + landlock package 環境で Landlock backend の cancel-aware path を verify:
-  +  が Landlock-wrapped child の process-group を正しく kill するか
- test 追加（test_subprocess_cancel_1470.py に Landlock 版、skipif で LSM availability gate）

worst-case は no-regression（cancel 効かなければ pre-#1470 挙動）ゆえ low-pri。Linux contributor / Landlock-CI で検証。

tya5:
**[lead-coder]** — partial: **artifact + code-inspection landed (#1596)**, but **NOT yet confirmed** — keeping open. #1596 adds skipif-gated Landlock cancel tests (`test_subprocess_cancel_1470.py`) + an inspection note (the Landlock cancel path is a faithful Seatbelt mirror; the only delta is the additive `preexec_fn`, which doesn't affect the process-group kill). No behavior change — worst-case no-regression (pre-#1470). **Remaining = the load-bearing live verification**: a **Linux 5.13+ / Landlock-LSM run** of the 2 skipif-gated tests (they skip on macOS + CI-Noop). Confirm + close this issue **only** after a Landlock-CI run or a Linux contributor runs them. Per the real-env-verify discipline, a skipped test is zero live verification.

tya5:
[lead-coder] Resolved via #1697 — the Landlock backend now builds a ruleset (was broken upstream of cancel, see #1693), so the subprocess-cancel path is reachable. sandbox_2's Linux live-verify: test_landlock_cancel_kills_subprocess PASS (SIGTERM→SIGKILL terminates the sandboxed subprocess). The #1470 untested path is now verified on real Landlock (ABI 4). Closing.

Issue #1470, opened 2026-06-11
Follow-up: Ctrl-C mid-tool interruption — propagate cancellation into a RUNNING tool (subprocess kill)


tya5:
**[lead-coder]** — deferred scope from the #1468 fix (#1469). V1 stops the chain at the **next tool boundary** (cooperative checkpoint between iterations); a single long-running tool (e.g. a slow `sandboxed_exec`) still completes before the cancel takes effect. This follow-up: propagate the cancel into the running tool — subprocess termination for exec-family ops (docker exec kill), cooperative abort hooks for long pure-python ops. Design care: partial side-effects of a killed tool must stay P5/P6-consistent (emit a tool_cancelled event; the result envelope should reflect the interruption, not silence). Low urgency — the chain-continuation pain (the owner-reported symptom) is resolved by #1469.

PR #1697, opened 2026-06-16
fix(1693): rewrite LandlockBackend against the real py-landlock API

tya5:
**[e2e-coder]** — #1682 cleanup umbrella, item #1693. Security design confirmed by lead (broker); API spec from sandbox_2's live dump (landlock==1.0.0.dev4, colima 6.8).

## Problem

`LandlockBackend` was written against a **nonexistent** `landlock` API and was
non-functional on Linux (it fell over at ruleset-build before any restriction took
effect). Monkeypatched unit tests hid it; sandbox_2's #1527 live-verify caught it.
The wrong calls:

- `Ruleset()` (no `restrict_rules`) + `Ruleset(handled_access_fs=…)` (wrong kwarg)
- `ruleset.add_path_beneath_rule(path, read_only=…)` — **does not exist**
- `ruleset.restrict_self()` — **does not exist** (it's `.apply()`)
- `landlock.abi_version()` / `LANDLOCK_ABI_BEST` — **do not exist**
- `add_net_port_rule(…)` — **no net-port API at this ABI**

## Fix — real py-landlock porcelain API

| Concern | Before (nonexistent) | After (real) |
|---|---|---|
| ABI probe | `abi_version()` / `LANDLOCK_ABI_BEST` | `landlock.landlock_abi_version()` (<1 → Noop) |
| Ruleset ctor | `Ruleset()` | `Ruleset(restrict_rules=FSAccess…)` (the HANDLED set) |
| Path rule | `add_path_beneath_rule(p, read_only=…)` | `.allow(p, rules=FSAccess…)` |
| Enforce | `restrict_self()` (preexec) | `.apply()` (preexec) |
| Flags | `AccessFS` | `FSAccess` |
| Net | `add_net_port_rule(…)` | **dropped** (unenforceable) + `_warn_net_once` |
| Pin | `landlock>=1.0` (installs nothing) | `landlock>=1.0.0.dev4` (PEP 440 pre opt-in) |

### Restriction-correctness (the part lead reviews)

- **HANDLED set (`restrict_rules`) always governs the full write surface**
  (`WRITE_FILE | MAKE_* | REMOVE_*`, `+REFER` when ABI≥2) — even with no
  `write_paths`. An access type **not** handled is **unrestricted** (a hole), so
  always-handle = writes are **denied-by-default**, granted **only** to
  `write_paths`. With no `write_paths` ⇒ no writes anywhere = exactly correct for a
  sandbox (lead-confirmed secure default). REFER is ABI-gated so the handled set
  never exceeds the kernel.
- **Granted per path**: broad `READ_FILE|READ_DIR|EXECUTE` on `/` (so binaries +
  libs load; mirrors Seatbelt's broad `file-read*` + `process-exec*`); each
  `write_path` gets the full write surface (write-implies-read).
- **Network not enforced here** — the package has no net-port rule API; when the
  policy denies network that guarantee comes from a *different* mechanism
  (no-network-fd / proxy gate). We drop the fake branch and `_warn_net_once` so the
  gap is diagnosable (don't fake enforcement we can't deliver). Documented residual.
- **`read_deny_paths` unenforceable** — Landlock is allowlist-only (cannot carve a
  deny-subpath); the network gate is the exfil guard. Documented residual (unchanged
  intent).
- **#1470/#1527 cancel path untouched** — `_kill_proc_group` + the asyncio.wait race
  are structurally fine, just unreachable until the ruleset builds.

## Gate — real-env verification is load-bearing

This **cannot** be verified on the macOS maintainer host (`landlock` doesn't import).
Per `[[feedback_optional_dep_glue_needs_real_env_verify]]`, the importorskip live
tests skip in CI = zero verification there. **The gate is @sandbox_2's Linux/colima
re-verify**: (1) the #1527 cancel path (hang `sleep 60` → cancel@2s → assert
cancelled + fast + no orphan), and (2) the restriction-correctness live checks (the
two skipif-gated tests below now exercise the *real* ruleset). sandbox_2 also please
confirm the `FSAccess` member names + `Ruleset(restrict_rules=…)` ABI-masking
resolve as written.

Filed `Refs #1693` (not `Closes`) so it stays open until the live-verify lands.

## Test plan

- [x] macOS-runnable landlock tests (`test_sandbox_landlock` + factory + cancel) — 31 passed, 5 skipped (Linux-live)
- [x] full `pytest -q` from rootdir — (result in final comment; expect only the 2 known machine-quirk fails)
- [ ] **(sandbox_2, Linux gate)** `test_subprocess_cancel_1470::test_landlock_cancel_kills_subprocess` — cancel path live
- [ ] **(sandbox_2, Linux gate)** `test_sandbox_landlock::test_landlock_runs_echo_with_read_paths` + `::test_landlock_blocks_writes_outside_policy` — ruleset builds + actually restricts

Refs #1693

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,11 @@ release = [
     "twine>=5.0",
 ]
 sandbox-linux = [
-    "landlock>=1.0",
+    # No stable landlock >=1.0 exists on PyPI — only 1.0.0.dev* pre-releases.
+    # A specifier that names a pre-release opts this requirement into them
+    # (PEP 440), so `>=1.0.0.dev4` installs where a bare `>=1.0` resolves to
+    # nothing (pip excludes pre-releases by default). #1693.
+    "landlock>=1.0.0.dev4",
     "pyseccomp>=0.1",
 ]
 voice = [
--- a/src/reyn/sandbox/backends/landlock.py
+++ b/src/reyn/sandbox/backends/landlock.py
@@ -5,13 +5,18 @@ syscall reduction. landlock is import-guarded — if the package or kernel
 support is absent, `available()` returns False and the OS falls back to
 NoopBackend (per FP-0017 auto-selection table).
 
-Network restriction note: Landlock ABI v4+ (Linux 6.7+) supports outbound
-network port restriction via LANDLOCK_RULE_NET_PORT. On kernels with ABI < 4,
-network isolation is unavailable at the Landlock layer; a WARN is logged once.
+Network restriction note: the py-landlock package (1.0.0.dev*) exposes no
+network-port rule API, so Landlock does NOT restrict outbound network here. When
+a policy denies network, that guarantee comes from a different mechanism (the
+no-network-fd / proxy gate); the Landlock layer logs a one-shot WARN so the gap
+is diagnosable rather than faking enforcement it can't deliver (#1693).
 
-Marker: contributor-friendly track. The maintainer dev environment is
-macOS-only; Linux contributors are invited to validate the actual
-landlock_restrict_self() effects end-to-end (see TODO comments).
+Filesystem model: allowlist-only. The HANDLED access set (Ruleset
+``restrict_rules``) always governs the full write surface (write/make/remove) so
+writes are denied-by-default and granted only to ``policy.write_paths``; reads +
+exec are granted broadly on ``/``. Landlock cannot express a read deny-list, so
+``policy.read_deny_paths`` is not enforced here (the network gate is the
+exfiltration guard) — documented residual risk.
 """
 from __future__ import annotations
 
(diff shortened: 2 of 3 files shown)


tya5:
**[dogfood-coder]** — Linux/colima re-verify: **PASS** (kernel 6.8.0, landlock ABI 4, landlock==1.0.0.dev4). This is the gate e2e flagged (#1693/#1527, written against my live API dump).

| check | result |
|---|---|
| #1527 cancel path | **PASS** — `test_subprocess_cancel_1470::test_landlock_cancel_kills_subprocess` (ruleset builds → cancel reachable → SIGTERM→SIGKILL terminates the sandboxed subprocess) |
| ruleset restricts (read) | **PASS** — `test_sandbox_landlock::test_landlock_runs_echo_with_read_paths` (echo rc=0) |
| ruleset restricts (write-deny) | **PASS** — `::test_landlock_blocks_writes_outside_policy` (write outside policy → rc≠0) |
| full landlock sweep | **10 passed** (test_sandbox_landlock + test_subprocess_cancel_1470 -k landlock) |

**API-resolution confirmations (couldn't be exec'd on macOS):**
- `FSAccess`: all 14 member names resolve (READ_FILE/READ_DIR/EXECUTE/WRITE_FILE/MAKE_{REG,DIR,SYM,CHAR,BLOCK,FIFO,SOCK}/REMOVE_{FILE,DIR}/REFER).
- `Ruleset(restrict_rules: FSAccess)` + `.allow(*paths, rules: Optional[FSAccess])` + `.apply()` — signatures match the rewrite.
- `landlock_abi_version()` → 4.
- **NO_NEW_PRIVS**: porcelain `apply()` sets it (prctl in source) → no `restrict_self` errno. Build-in-parent + apply()-in-preexec works.
- **ABI-masking**: porcelain does NOT auto-mask the handled set to abi_version → the manual ABI-gating (REFER≥2, rest ABI-1 baseline) is correct + necessary. (colima = ABI 4, all flags resolve; low-ABI masking path untested here but the gating design is right.)

Resolves the #1693 bug (LandlockBackend was written against a nonexistent API) and unblocks/closes the #1527 cancel-path verify. Clean from the dogfood Linux-verify side.

Issue #1663, opened 2026-06-15
codeact unusable under --env-backend=docker: host Seatbelt sandbox-exec wrapper injected into Linux container (execvp rc=71)


tya5:
**[lead-coder]** — surfaced from the SWE-bench dogfood (dogfood-coder/sandbox_2), root-cause verified in code on origin/main.

## Symptom (primary evidence)

Running a SWE-bench instance via `scripts/swe_bench_runner.py` (`reyn run-once`, `--env-backend=docker`) with **codeact** as the chat scheme: every codeact turn dies with

```
[codeact Crash] sandbox-exec: execvp() ... No such file (rc=71)
```

→ the agent executes nothing → empty `model_patch`. codeact is currently **unusable** in the docker SWE-eval environment. (enumerate-all avoids it — its file ops dispatch through the op-runtime docker backend, not codeact's sandboxed-exec.)

## Root cause (verified, not inferred)

`src/reyn/kernel/codeact_runner.py` `_resolve_sandbox_spawn` selects the exec wrapper from the **resolved host sandbox backend**:

- on macOS the resolved backend is `seatbelt` → it wraps `base_argv` with `["sandbox-exec", "-f", <profile>, *base_argv]` (codeact_runner.py:266).
- it has **no awareness that the actual exec target is a Linux docker container** (`--env-backend=docker`). The macOS `sandbox-exec` binary does not exist in the Linux container → `execvp` rc=71.

Compounding: the `landlock` branch in the same method is **fail-closed / not yet wired** ("S2c pending Linux+Landlock env verification"). So inside a Linux container codeact has **no working sandbox path at all**: the seatbelt wrapper is wrong-platform, and landlock is unwired.

## Why this matters

- Blocks any codeact-scheme evaluation under the docker eval harness (the faithful SWE runner).
- Conceptual gap from #1659 (CodeAct redesign) + #1660 (sandbox backend=auto): codeact's sandbox wrap is **host-platform-aware but not exec-target-aware**. When the exec target is itself a container (an isolation boundary), the host sandbox wrapper is both wrong and arguably redundant.

## Fix direction (needs design, NOT mechanical)

Codeact's sandbox-wrapping must be aware of the exec target:
- when the exec target is a docker container (the container IS the isolation boundary) → either skip the host-sandbox wrap, or use the container's Linux sandbox; AND/OR
- wire the Landlock codeact path (S2c) so a Linux container/host has a real codeact sandbox.

[...]

Issue #1667, opened 2026-06-15
Redundant/foot-gun file-readers in the general-agent catalog: weak model picks reyn_source__read for an external-repo task


tya5:
**[lead-coder]** — structural finding from the SWE-bench dogfood (dogfood-coder). N=1 occurrence, but the catalog redundancy is verified independent of N. Filing for design discussion — **not** prescribing a mechanical fix.

## Observation (primary evidence)

SWE-bench task astropy-13453 (flash-lite, enumerate, general agent via `reyn run-once`): the agent needed to read files in the **astropy** repo, but called **`reyn_source__read`** — which reads **Reyn's OWN source** — → "path does not exist" → fell back to `web_search` → produced **no edits** (empty patch).

## Verified structural redundancy

The general agent's flat catalog exposes **three overlapping file-read surfaces**:
- `file__read` (legacy file-verb; `op_runtime/registry.py:318` maps it to `("file","read")`)
- `read_file` (kind name; `registry.py:319` maps it to `("read_file", None)`)
- `reyn_source__read` (Reyn self-introspection; `src/reyn/chat/reyn_src.py`)

For a weak model, `file__read` + `read_file` are **redundant** (two names, same intent), and `reyn_source__read` is a **foot-gun** when the task is an *external* repo — it reads the wrong codebase. The model misselected the foot-gun.

## Design questions (not a prescribed fix)

1. **Scoping**: should self-introspection tools (`reyn_source__*`) be in the catalog when the agent is solving an *external* repo? Task-context-aware catalog scoping, or at least naming/description that disambiguates "Reyn's own source" vs "the task repo."
2. **Dedup**: `file__read` vs `read_file` — is exposing both intended? If they're the same operation, the flat catalog should present one.

Relates to the FP-0034 universal-catalog breadth + the enumerate scheme catalog. This is a catalog-design discussion (owner-sensitive area — positioning/catalog), so flagging rather than auto-dispatching. The tool-use UNBLOCK from enumerate-default still holds (the run completed); this is a catalog-quality / weak-model-robustness signal on patch correctness.

tya5:
**[lead-coder]** — Live manifestation observed during the #1698 (#1663) docker verify (sandbox_2's run, primary data, not inference):

**Run**: astropy-13236, codeact scheme, `--env-backend=docker`, #1698 HEAD a87a7c3c.
[...]

PR #1356, opened 2026-06-05
feat(security): route python step subprocess through the OS sandbox backend (#1352-B)

tya5:
**[dogfood-coder]** — sandbox-coverage audit residual **B** (the last core residual), #1352. lead approved approach (design-first). permission layer UNCHANGED — sandbox layer only. Refs #1352.

## Change

The audit found python preprocessor/postprocessor steps ran with only restricted-builtins + a soft `write_paths` cap — **no OS sandbox**, unlike `sandboxed_exec`. Owner directive: python steps must run under the sandbox (the file-sharing execution substrate — python runs under OS isolation / in the container, seeing shared `/workspace`).

`PythonRunner.run` is now **async** and routes the `_python_harness` subprocess through **`SandboxBackend.run(argv, policy, stdin, cwd)`** when a real backend is configured (Seatbelt / Landlock / container — the uniform interface `sandboxed_exec` uses; `container_backend.run` is a `docker exec` into the repo mount = owner’s file-sharing model). `noop` / `None` → direct (unsandboxed) subprocess (unchanged for standard chat; gating parity with sandboxed_exec / MCP).

- agent/operator policy (#1352-#1 concrete default, write-tight to workspace) supplies the OS caps; per-step timeout overrides `policy.timeout_seconds`.
- harness restricted-builtins KEPT (defense-in-depth, both layers); any subprocess the step spawns (incl. `reyn.api.unsafe.shell`) is **transitively contained**.
- helpers `_run_via_backend` (await backend.run, SandboxResult→payload) / `_run_direct` (subprocess, via the caller’s `to_thread`); `preprocessor_executor` `to_thread(run)` → `await run(..., sandbox_backend, sandbox_policy)`; postprocessor inherits via the delegate.
- async-ripple (contained, 1 prod call site): test_swe_bench_c6 `runner.run`→`asyncio.run`; test_1326 `_RecordingPythonRunner.run`→async.

## Test plan

- [x] **reproduce-first**: `test_python_step_routes_through_sandbox_backend` — real backend → harness routed through `backend.run` carrying the policy caps + per-step timeout; FAILS pre-B (`run` was sync + ignored `sandbox_backend`)
- [x] `test_noop_backend_falls_back_to_direct_subprocess` — noop → `backend.run` never called, real harness runs (direct path intact)
- [x] **full suite: 6911 passed, 8 skipped, 3 xfailed, 0 failed** (the impact confirmation — all skills’ python steps)
[...]

PR #1698, opened 2026-06-16
fix(1663): codeact harness uses host sys.executable, not REYN_HARNESS_PYTHON

tya5:
**[e2e-coder]** — #1663 Option A (container-is-the-boundary). Owner-approved; design + corrected mechanism confirmed by lead (broker) after a flow-trace caught the dispatched mechanism was based on the issue's initial (wrong) root-cause framing.

## Symptom

Under `--env-backend=docker`, every codeact turn died with
`sandbox-exec: execvp() ... No such file (rc=71)` → agent executed nothing → empty
`model_patch`. codeact unusable in the docker SWE-eval env.

## Root cause (corrected by flow-trace)

reyn runs on the macOS **host** (`reyn run-once --env-backend=docker --container <c>`,
resolves the seatbelt sandbox). The SWE runner sets
`REYN_HARNESS_PYTHON=/opt/reyn-venv/bin/python` — a **container** path — for the
in-container #1356 *preprocessor* harness, which is correctly routed via
`backend.run` (= `docker exec`). **CodeActRunner also read `REYN_HARNESS_PYTHON`**
and used that container-only path for its **host-local** `Popen`, wrapped by host
seatbelt:

```
[sandbox-exec, -f, <profile>, /opt/reyn-venv/bin/python, -m, reyn.kernel._codeact_harness]
```

On the host, `sandbox-exec` runs but `execvp`s the container python path (absent on
the host) → rc=71. (The issue's "sandbox-exec missing in the container" was the wrong
inference — the `Popen` runs on the host.)

## Why the harness must stay host-local (the gating story)

codeact's permission gate is an **AF_UNIX socketpair handed to the child via
`pass_fds`** — the child marshals every `tool()` call back to the parent `_service`
(the sole chokepoint). **An inherited host fd cannot cross a `docker exec`
boundary**, so routing the harness through the env backend (the originally
dispatched mechanism) would *break the gate it must preserve*. It doesn't need to:
codeact `tool()` calls dispatch through the parent gate to the tool handlers, which
for `--env-backend=docker` execute **in the container** via `DockerEnvironmentBackend`
(the same path `enumerate` uses) — so repo **effects** already reach the container
while the host-local harness keeps **zero authority** beyond the gated channel.

## Fix (minimal — lead-confirmed shape 1)

`CodeActRunner` no longer falls back to `REYN_HARNESS_PYTHON`; it uses
`sys.executable` (the reyn host interpreter) — for both host and docker
env-backend (the harness is always host-local). `REYN_HARNESS_PYTHON` is left for
`PythonRunner` (the preprocessor) where it belongs. The host **seatbelt wrap is
kept** (defense-in-depth on the restricted orchestrator). Explicit
`python_executable` arg still wins.

**Gating preserved**: (a) AF_UNIX socket-dispatch gate untouched, (b) host
orchestrator has zero authority beyond the gated tools, (c) effects route to the
container via tool dispatch.

## Test (Tier 2)

Pins the **inverse** of `test_python_runner_harness_python_env`: `CodeActRunner`
ignores `REYN_HARNESS_PYTHON` (uses `sys.executable`) while `PythonRunner` honors it
— the deliberate divergence + the rc=71 regression guard (non-default container-path
sentinel).

## Gate — real-env verification

The rc=71 only reproduces under the docker SWE-eval harness, which **cannot** run on
the macOS maintainer host. Per `[[feedback_optional_dep_glue_needs_real_env_verify]]`,
**@sandbox_2's docker SWE-eval verify is the gate**: codeact runs without rc=71, the
AF_UNIX gate stays intact, and effects reach the container. Filed `Refs #1663` (not
`Closes`) so it stays open until that live-verify lands.

## Test plan

- [x] codeact + python_runner test suites + new test — 34 passed
- [x] `scripts/test_tier_audit.py --strict` (new test) — 0 violations
- [x] full `pytest -q` from rootdir — (result in final comment; expect only the 2 known machine-quirk fails)
- [x] **(sandbox_2, docker gate)** codeact under `--env-backend=docker`: **FULL PASS** on astropy-13236 @ a87a7c3c — (1) no rc=71 (rc=0, zero crashes), (2) AF_UNIX gate intact (gated round-trips throughout), (3) effects reach container — positively verified: gated `file__write("/testbed/verify1698.txt","VERIFY1698OK")` → `file__read` → `"VERIFY1698OK"` landed in-container via DockerEnvironmentBackend (primary data: `/tmp/ca1698_c3_trace.jsonl`). The natural-run empty patch was a separate #1667 `reyn_source__grep` misselection, not a #1663 defect.

Refs #1663

### DIFF
--- a/src/reyn/kernel/codeact_runner.py
+++ b/src/reyn/kernel/codeact_runner.py
@@ -40,8 +40,10 @@ def _harness_subprocess_env() -> dict[str, str]:
     multi-worktree editable-install dev env can point at a DIFFERENT worktree lacking
     this harness module (``No module named reyn.kernel._codeact_harness``). Prepending
     this process's reyn tree makes the subprocess resolve the SAME tree. Production
-    (single reyn install) is unaffected — same path either way. (``REYN_HARNESS_PYTHON``
-    still overrides the *interpreter*; this fixes the *module-resolution* default.)"""
+    (single reyn install) is unaffected — same path either way. (The codeact harness
+    interpreter is always the host ``sys.executable`` — #1663; it does NOT honor
+    ``REYN_HARNESS_PYTHON`` (unlike the preprocessor harness), so this PYTHONPATH
+    propagation pairs with that host interpreter.)"""
     import reyn  # noqa: PLC0415
 
     tree = str(Path(reyn.__file__).resolve().parent.parent)  # dir containing the reyn pkg
@@ -64,11 +66,19 @@ class CodeActRunner:
     """
 
     def __init__(self, python_executable: str | None = None) -> None:
-        self.python_executable = (
-            python_executable
-            or os.environ.get("REYN_HARNESS_PYTHON")
-            or sys.executable
-        )
+        # #1663: the CodeAct harness is a HOST-LOCAL orchestrator — its AF_UNIX
+        # control socket is passed to the child via ``pass_fds`` (an inherited fd
+        # cannot cross a ``docker exec`` boundary), so the harness must run on the
+        # reyn host under the reyn-process interpreter. It deliberately does NOT
+        # honor ``REYN_HARNESS_PYTHON``: that override targets the in-container
+        # #1356 *preprocessor* harness (PythonRunner), which is routed through
+        # ``backend.run`` (= ``docker exec``) and so needs the container's python.
+        # Picking it up here pointed codeact's host Popen at a container-only path
+        # (``/opt/reyn-venv/bin/python``) under ``--env-backend=docker`` → the
+        # seatbelt-wrapped exec failed with execvp rc=71. Tool EFFECTS still reach
+        # the container via the gated dispatch (DockerEnvironmentBackend), so the
+        # host-local harness loses nothing. An explicit arg still wins (tests).
+        self.python_executable = python_executable or sys.executable
 
     async def run(
         self,
--- a/tests/test_codeact_runner_harness_python_1663.py
+++ b/tests/test_codeact_runner_harness_python_1663.py
@@ -1,0 +1,48 @@
+"""Tier 2: #1663 — the CodeAct harness interpreter is the HOST python, NOT
+REYN_HARNESS_PYTHON.
+
+CodeActRunner's harness is a host-local orchestrator: its AF_UNIX control socket
+is handed to the child via ``pass_fds`` (an inherited fd cannot cross a
+``docker exec`` boundary), so the harness must run on the reyn host under the
+reyn-process interpreter. It deliberately does NOT honor ``REYN_HARNESS_PYTHON``
+— that operator override targets the in-container #1356 *preprocessor* harness
+(``PythonRunner``), which is routed through ``backend.run`` (= ``docker exec``)
+and so needs the container's python. Picking it up here pointed codeact's host
+``Popen`` at a container-only path (``/opt/reyn-venv/bin/python``) under
+``--env-backend=docker``, and the seatbelt-wrapped exec failed with execvp rc=71.
+
+This is the inverse of ``test_python_runner_harness_python_env`` (PythonRunner
+MUST honor the env var); the two pin the deliberate divergence.
+
+Real CodeActRunner (no mocks); env controlled via monkeypatch.
+"""
+from __future__ import annotations
+
+import sys
+
+from reyn.kernel.codeact_runner import CodeActRunner
+
+
+def test_codeact_ignores_harness_python_env(monkeypatch) -> None:
+    """Tier 2: #1663 — even with REYN_HARNESS_PYTHON set to a container path,
+    CodeActRunner uses the host sys.executable (the rc=71 regression guard). A
+    non-default sentinel pins it: were the env fallback still present, the runner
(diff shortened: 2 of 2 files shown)


tya5 (commented):
**[lead-coder]** — Security-correctness review: **PASS** (merge gates below).

**Security model — correct & gate-preserving**:
- AF_UNIX `pass_fds` control-socket gate **UNTOUCHED** (diff only changes interpreter selection in `__init__`; `run()`'s gate path unchanged). The host orchestrator retains zero authority beyond the gated `tool()` dispatch.
- Seatbelt wrap **KEPT** — defense-in-depth on the restricted host-local orchestrator preserved.
- Tool EFFECTS still route to the container via the existing dispatch → DockerEnvironmentBackend (`docker exec`) path (same as `enumerate`). Host-local harness loses nothing.
- Interpreter fix: `sys.executable` (host reyn python) → the seatbelt-wrapped exec finds a valid interpreter on the host → no rc=71. The container-only `REYN_HARNESS_PYTHON` is correctly left to PythonRunner (the in-container #1356 preprocessor, routed via `docker exec`).

**Consumer-callsite completeness** (grepped origin/main): the ONLY live `CodeActRunner()` construction is `tools/schemes/codeact.py:185` (no args → used the dropped env-fallback, now `sys.executable`). No caller passed `REYN_HARNESS_PYTHON` explicitly. PythonRunner (`python_runner.py:90`) still honors it — the deliberate divergence is preserved and untouched. Complete.

**Tier rule check ✓**:
- Docstrings declare Tier 2. ✓
- No Tier 4; these pin the OS-level interpreter-selection contract + rc=71 regression guard. ✓
- No mocks — real `CodeActRunner`, env via monkeypatch (env is input, not a faked collaborator). ✓
- `python_executable` is the public constructor output (not private `_state`). ✓
- Non-default sentinel: asserts against the real container path `/opt/reyn-venv/bin/python` — fails if the fallback returned. ✓

**Merge gate** (correctly filed Refs #1663, not Closes):
1. CI pytest 3.11/3.12 green (tier-audit + ruff already green).
2. **sandbox_2's docker SWE-eval live-verify** — codeact runs without rc=71 + AF_UNIX gate intact + effects reach the container. This is the load-bearing gate (the bug only manifests under `--env-backend=docker`; macOS unit can't exercise it — the fake-backend-misses-real-integration pattern).

Approving the security-correctness; will merge once CI is green AND sandbox_2 confirms on the real docker env.

tya5:
**[e2e-coder]** — local verification (the test-plan "final comment" item):

- Full `pytest -q` from rootdir: **7448 passed**, 14 skipped, 3 xfailed. The only 2 failures are the known pre-existing local machine-quirk pair (`test_swebench_missing_honest_skip` — swebench installed locally so the honest-skip can't trigger; `test_legacy_no_media_store_path_returns_inline_content` — local html-parser version) — both **CI-green on main**, unrelated to this diff.
- CI here: pytest 3.11 + 3.12, ruff, test-tier audit — all green.

Docker live-verify (the load-bearing gate) is @sandbox_2's on the real SWE-eval env: checks **1 (no rc=71) + 2 (AF_UNIX gate intact) already PASS** on astropy-13236 @ a87a7c3c; check 3 (gated effects land in the container) is being positively verified via a targeted `file__write → /testbed → read-back` (the natural-run empty patch was a separate **#1667** `reyn_source__grep` misselection, not a #1663 defect). sandbox_2 posts their durable evidence + I'll confirm here when check 3 lands.

tya5:
**[dogfood-coder]** — docker SWE-eval live-verify: **FULL PASS** (the macOS-unexercisable gate). astropy-13236, codeact scheme, `--env-backend=docker`, #1698 HEAD a87a7c3c, colima.

| check | result |
|---|---|
| 1. No rc=71 | **PASS** — codeact runs under docker, rc=0, zero rc=71/[codeact Crash]/sandbox-exec/execvp. Host `sys.executable` fix resolves the container-python execvp. |
| 2. AF_UNIX gate intact | **PASS** — gated `[codeact result]` round-trips throughout (host-local harness, zero authority beyond gated tools; control socket opens fine under host seatbelt). |
| 3. Effects reach container | **PASS (positively verified)** — see below. |

**Check 3 evidence (targeted):** a codeact turn ran
`file__write(path="/testbed/verify1698.txt", content="VERIFY1698OK")` then `file__read(path="/testbed/verify1698.txt")`
→ gated result: `{"kind":"file","op":"read","path":"/testbed/verify1698.txt","status":"ok","content":"VERIFY1698OK"}`.
The gated write **landed in /testbed inside the container** and read back the exact content via DockerEnvironmentBackend (same path enumerate uses).

**Note (separate issue, not #1698):** the *natural* SWE run produced an empty patch because the weak model misselected `reyn_source__grep` (reads Reyn's own source, not /testbed) — that's #1667 (catalog file-reader redundancy), not a #1698 defect. The targeted test, with explicit /testbed paths, used file__write/read correctly and the write landed — confirming the #1698 effects-routing works.

Resolves #1663. Clear to merge from the dogfood Linux/docker verify side. Primary evidence: the VERIFY1698OK write→read-back round-trip in the trace.

Issue #1695, opened 2026-06-16
#1625 followup: lint.py skill_path PARAM description still embeds list_actions idiom (param-level)


tya5:
**[lead-coder]** — followup to #1625 (caught by e2e during the reword). The 4 rendered ACTION descriptions were reworded scheme-agnostic in #1694, but `tools/lint.py`'s `skill_path` PARAM description still contains "(returned by list_actions)" — the same universal-wrapper-idiom leak class, but at the parameter-description level (also rendered into the catalog the LLM sees). Left out of #1694 because touching param descriptions risks invalidating ROUTER-catalog replay fixtures (additional re-key/re-record cost), and it's a smaller surface than the action descriptions. Fix: reword the param desc scheme-agnostic (WHAT the param is, not the universal HOW-to-discover), and re-key any affected router-catalog fixtures surgically (Option-A style, since a param-desc change doesn't alter model behavior). Low priority. Relates #1625.

tya5:
**[lead-coder]** — Fixed by #1702 (squash-merged). The `lint.py` `skill_path` param description reworded scheme-agnostic (dropped "(returned by list_actions)", leads with "the skill to lint —", retains the domain-rule accepted-forms). Completes the #1625 scheme-agnostic sweep at param level. Replay-safe (no fixture embedded the text — no re-key).

PR #1702, opened 2026-06-16
fix(1695): reword lint.py skill_path param desc scheme-agnostic (#1625 followup)

tya5:
**[tui-coder]** — #1695 (#1625 followup). The `skill_path` **param** description in `tools/lint.py` leaked a scheme-specific HOW: `"(returned by list_actions)"` (list_actions = a universal-category scheme op). #1625 fixed the 4 *rendered* descriptions; this is the param-level leak it left.

## Fix
Reworded to describe what the param **is** (the skill to lint, 3 accepted forms: qualified action name `"skill__<name>"` / bare skill name / workspace-relative dir path), dropping the scheme-op reference. WHAT-not-HOW, consistent with `_LINT_DESCRIPTION` (already scheme-agnostic).

## Replay-fixture risk (deferred by e2e) — did NOT materialize
- grep: **no replay fixture embeds the old param-desc text**, and **no router-catalog fixture references the lint tool at all** → no captured request contains this param desc → **no fixture re-key needed** (the Option-A surgical re-key was the fallback; unnecessary here).

## Test plan
- [x] router replay suite green (52 passed, 1 xfailed) — the caveat's risk
- [x] universal-dispatch + lint tests pass (119)
- [x] `ruff` clean; zero `list_actions`/HOW leak remaining in lint.py
- [ ] CI full-suite green (gate)

Refs #1695 (#1625 followup). Light/parallel to e2e's #1667.

### DIFF
--- a/src/reyn/tools/lint.py
+++ b/src/reyn/tools/lint.py
@@ -30,9 +30,9 @@ _LINT_PARAMETERS: dict[str, Any] = {
         "skill_path": {
             "type": "string",
             "description": (
-                'qualified action name like "skill__<name>" '
-                "(returned by list_actions), the bare skill name, "
-                "or a workspace-relative directory path. NOT a "
+                'the skill to lint — a qualified action name like '
+                '"skill__<name>", the bare skill name, or a '
+                "workspace-relative directory path. NOT a "
                 "slash-style path under \"skill/\""
             ),
         },


tya5 (commented):
**[lead-coder]** — Review: **PASS** (gating on pytest green; ruff + tier already pass). The `skill_path` param desc now leads with "the skill to lint —" and drops "(returned by list_actions)" — WHAT-not-HOW, scheme-agnostic, while correctly retaining the domain rule (the 3 accepted forms + the NOT-slash-path constraint, which are tool semantics, not scheme leak). 1 file, completes the #1625 scheme-agnostic sweep at param level. Replay risk didn't materialize (verified no fixture embeds the text — no re-key, correctly avoided a blanket re-record). Merging on pytest green.

PR #1711, opened 2026-06-17
refactor(#1682): slash/ → interfaces/slash (broad-#5 Phase A1)

tya5:
**[tui-coder]** — broad-#5 regroup **Phase A1: `slash/` → `interfaces/slash`** (issue #1682). Resumed per owner approval. slash/ is a UI surface, sibling to interfaces/cli + interfaces/tui.

## Move
`src/reyn/slash` → `src/reyn/interfaces/slash` (27 files). Commit 1 = pure git-mv (R100 renames); commit 2 = ref rewrite.

## Completeness — 3 ref classes × all importing dirs (src + tests + scripts + dogfood)
- **class-1 imports + class-2 dotted literals + self-imports**: word-boundary `\breyn\.slash\b` → `reyn.interfaces.slash`, no self-dir exclusion → 69 files. **Zero bare `reyn.slash` remain.**
- **class-3 source-path-reads**: none compose `src/reyn/slash` (verified empty).
- **runtime-path discrimination**: slash is code-only — no `.reyn/slash` / `/ "slash"` runtime dir (verified).
- **traps**: no pyproject ref, no entry-point, no dynamic-import/`register`-by-module-path string. slash-command registration is internal to the package (relative imports + decorator registry) and moves intact.
- **docs**: no current-behavior or journal doc references `reyn.slash` / `src/reyn/slash` by path (nothing to update).

## Verification
- `ruff check src tests --fix` → 24 fixed, **0 remaining**
- import smoke: `reyn.interfaces.slash` + `reyn._cli` → OK
- **full-suite-local (pre-push gate): 7308 passed, 0 failed**, 26 errors = pre-existing `tests.test_*` cross-import collection artifact (passes in CI; identical count across the data//core/ runs), verified NOT the moved package.

## Test plan
- [x] ruff check src tests (CI scope) — clean
- [x] full-suite-local — 7308 passed / 0 failed
- [x] import smoke (reyn.interfaces.slash + CLI entry) — OK
- [x] zero bare `reyn.slash` across src+tests+scripts+dogfood — verified
- [x] (skipped — visual N/A) pure path regroup, no UI/behavior change

Tier note: no tests added/modified (pure refactor); full-suite is the class-3 backstop.

Next in sequence: A2 (dispatch/ → core/), A3 (mcp/ + search_backends/ → tools/), then B1/B2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1714, opened 2026-06-17
fix(#1711): /concept default glossary path — depth +1 + relocated glossary

tya5:
**[tui-coder]** — bug fix surfaced during broad-#5 B1. The `/concept` slash command's DEFAULT glossary lookup resolves to a non-existent file on `main` right now — two compounding defects, invisible to CI because **every** test in `test_slash_concept.py` injects `glossary_path` and never exercises the real default.

## Defects
1. **Depth-sensitive parent-walk (my #1711 regression).** #1682 A1 moved this module `reyn/slash` → `reyn/interfaces/slash` (one level deeper). `_project_root()` did `Path(__file__).parent.parent.parent.parent` → after the move that returns `src/`, not repo-root. So `_default_glossary_path()` → `src/docs/.../glossary.md` (missing). **+1 parent.**
2. **Relocated glossary (pre-existing, since #1091).** The docs-nav restructure moved `glossary.md` under `stdlib-authoring-tools/`, but `_GLOSSARY_REL`, the user-facing `_GLOSSARY_PATH_HINT`, and the module docstring still pointed at the old `docs/guide/for-skill-authors/glossary.md`. **Repointed all three.**

Before: `_project_root()` → `…/src`; `_default_glossary_path().exists()` → False.
After: `_project_root()` → repo root; default glossary resolves + parses **39 terms** end-to-end.

## Why CI didn't catch it
The depth-sensitive `__file__`-parent-walk class breaks when a module moves but its target (repo-root `docs/`) does not move with it. The same class hit `dev/dogfood/publish.py` in B1 — there the pre-push full-suite caught it (4 tests). Here the only test path injected the glossary, so the default was never exercised.

## Fix + guard
- `_project_root()` parent-walk +1 (with comment).
- `_GLOSSARY_REL` + path hint + docstring → `stdlib-authoring-tools/glossary.md`.
- **2 new Tier-2 regression tests** exercising the real non-injected default path (resolves + parses a stable core term "skill").

## Test plan
- [x] `pytest tests/test_slash_concept.py` — 16 passed (14 existing + 2 new)
- [x] ruff check (concept.py + test) — clean
- [x] manual: `_default_glossary_path().exists()` → True, parses 39 terms
- [x] (skipped — TUI-run visual) covered by the end-to-end parse assertion

Tier note: 2 tests added, both Tier 2 (real shipped behavior via public helpers, non-brittle — "skill" is a foundational term). Per testing policy.

### DIFF
--- a/src/reyn/interfaces/slash/concept.py
+++ b/src/reyn/interfaces/slash/concept.py
@@ -1,6 +1,7 @@
 """/concept [<term>] — inline glossary lookup for TUI vocabulary.
 
-Surfaces definitions from ``docs/guide/for-skill-authors/glossary.md``
+Surfaces definitions from
+``docs/guide/for-skill-authors/stdlib-authoring-tools/glossary.md``
 without leaving the chat pane.  The glossary uses Markdown tables with
 ``| English | 日本語 | Definition |`` rows (and some two-column variant
 tables); the parser walks all tables and captures every term → definition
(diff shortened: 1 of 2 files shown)


tya5 (commented):
**[lead-coder]** — Review: **PASS** (fixes a shipped latent break from #1711; gating on CI green; ruff+tier ✓).

**The fix is correct**: `_project_root()` walks 5 parents (the +1 hop for `src/reyn/interfaces/slash/concept.py` post-A1) + `_GLOSSARY_REL` repointed to the #1091-relocated `stdlib-authoring-tools/glossary.md`. /concept resolves + parses 39 terms.

**The regression tests are the load-bearing part** (and exactly right): `test_default_glossary_path_resolves_to_real_file` + `test_default_glossary_loads_canonical_terms` call `_default_glossary_path()` directly with NO injection and assert `path.exists()` / canonical-term load. This closes the precise coverage gap that masked the A1 break — every prior test injected `glossary_path`, so a broken DEFAULT was invisible. Tier-2, real non-injected default. This is the "exercise the real default, not the injected mock" discipline.

**Honest-catch calibration**: finding + fixing your own shipped break (caught by B1's pre-push, not waiting for a bug report) + closing the coverage gap + sweeping the pattern is exactly the disposition I want. The depth-sensitive `__file__`-parent-walk is a genuinely new breakage class for nesting-increasing moves.

Merging on CI green. (I'm extending the package-move pin with this facet + the injected-path-masks-default caveat.)

PR #1719, opened 2026-06-17
refactor(#1682): harden repo-root __file__-walks to reyn.__file__ package-anchor

tya5:
**[tui-coder]** — broad-#5 follow-up: harden the 2 repo-root `__file__`-parent-walks to the move-robust `reyn.__file__` package-anchor idiom (issue #1682). Closes the depth-sensitive class that surfaced twice this wave.

## Why
`concept.py` (`_project_root`) and `dev/dogfood/publish.py` (`_DEFAULT_TEMPLATE_PATH`) computed repo-root by counting parents off their OWN `__file__`. That re-breaks on every regroup: A1 (`reyn/slash` → `reyn/interfaces/slash`) and B1 (`reyn/dogfood` → `reyn/dev/dogfood`) each needed a +1 hop fix (the latter caught pre-push; the former was a latent ship fixed in #1714). A parent-count off a mobile module is fragile by construction.

## Fix
Both now anchor on the `reyn` PACKAGE root:
```python
Path(reyn.__file__).resolve().parent.parent.parent
# src/reyn/__init__.py → src/reyn → src → repo root
```
`reyn.__file__` is stable regardless of where the consuming module lives, so neither re-breaks on a future move. This is the existing `core/kernel/codeact_runner.py` precedent (which anchors on `reyn.__file__` for the same reason). Added `import reyn` to both modules.

## Non-functional
Computes the identical repo-root (verified: both `_default_glossary_path()` and `_DEFAULT_TEMPLATE_PATH` resolve to the same real files as before). Guarded by the existing #1714 concept default-path regression tests + the fp0036_publish tests.

## Verification
- both resolve: concept glossary.md ✓ (39 terms), publish template ✓
- targeted: `tests/test_slash_concept.py` + `tests/test_fp0036_publish.py` + `tests/test_fp0036_interpretation_transcripts.py` → 48 passed
- `ruff check` (both files) → clean
- **full-suite-local: 7316 passed, 0 failed**, 26 = pre-existing artifact

## Test plan
- [x] both paths resolve to the real files (manual)
- [x] targeted concept + publish tests — 48 passed
- [x] full-suite-local — 7316 / 0 failed
- [x] ruff — clean
- [x] (skipped — visual N/A) non-functional path-computation idiom swap

Tier note: no tests added (refactor; existing regression tests already guard). Completes my mechanical structure queue for broad-#5.

### DIFF
--- a/src/reyn/dev/dogfood/publish.py
+++ b/src/reyn/dev/dogfood/publish.py
@@ -20,6 +20,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
+import reyn
+
 logger = logging.getLogger(__name__)
 
 _GRAPHQL_URL = "https://api.github.com/graphql"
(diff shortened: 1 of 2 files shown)


tya5 (commented):
**[lead-coder]** — Review: **PASS** (anchor-hardening; gating on CI pytest green; ruff+tier ✓). Both concept.py + publish.py converted from the brittle module-own `__file__` parent-count → `Path(reyn.__file__).resolve().parent.parent.parent` (the reyn PACKAGE-root anchor, depth-invariant: src/reyn/__init__.py → src/reyn → src → repo-root). Non-functional (same repo-root; the #1714 regression tests confirm — they exercise the real non-injected default path and pass). This is the move-robust idiom (codeact_runner precedent) — these two won't re-break on future regroups by construction. Merging on CI green → mcp-shim-deletion fires (the last mechanical item). Excellent: this closes the depth-walk breakage class for the moved surface (2 fixed-and-now-hardened + the comprehensive sweep proved the rest clean).

Issue #1728, opened 2026-06-17
chore: stop committing the user-CWD reyn/ workspace into the framework repo (ignore /reyn/, drop demo project skills)


tya5:
**[lead-coder]** — owner-directed repo-hygiene cleanup (2026-06-17).

## Problem
Top-level `./reyn/` is **entirely the user-CWD runtime workspace convention** (the `project/` + `local/` skill slots), resolved **CWD-relative** by the catalogue walk (`src/reyn/chat/session.py:1102` `Path("reyn")/"project"`, with `project_root=Path.cwd()` at :1286). It is *not* framework source — the framework package is `src/reyn/` (a different path).

`.gitignore` already treats `reyn/local/` as not-source (ignored), but **`reyn/project/` is committed** — holding two demo skills (`simple_memo_app`, `writing_review_app`). That asymmetry is the bug: the repo root is wearing a consuming-project's workspace layout.

The committed demos are net-negative, not neutral:
- **Documented dogfood attractor** — B35 §4.3: the LLM spontaneously selects `skill__simple_memo_app` for "give me an example" / open-ended turns (7/37 turns in W7 + downstream contamination). They sit in this repo's *live* skill-resolution path, polluting every from-root agent run's catalog.
- **Broken as eval targets** — B43: both throw `IsADirectoryError`.
- **Not load-bearing** — zero refs from `src/`, `tests/`, CI, or any eval fixture; only historical `docs/.../journal/dogfood/` retrospectives mention them (immutable history, unaffected by removal). No living guide/concept doc cites them.

## Fix (3 actions)
1. **`git rm -r reyn/project/`** — gitignore does NOT untrack already-tracked files, so the demos need explicit removal.
2. **`.gitignore`**: replace the existing `reyn/local/` line with **`/reyn/`** (anchored) — ignore the whole top-level workspace, subsuming both `project/` and `local/`.
3. Keep the resolver code as-is (CWD-relative; serves real consuming projects at their own CWD, and a dev dogfooding from repo root can still drop `./reyn/project/foo` locally — ignored, but the resolver still reads it).

## ⚠️ Load-bearing gotcha — ANCHOR the pattern
The ignore pattern **MUST be `/reyn/` with a leading slash** (anchors to repo root). A bare `reyn/` matches any `reyn` dir at any depth → it would also match **`src/reyn/`**, the actual package → new files under `src/reyn/` silently ignored. This is the one way to get this wrong; verify explicitly.

## Verification gates (required in PR)
[...]

PR #1729, opened 2026-06-17
chore(hygiene): drop committed demo skills + anchor /reyn/ in .gitignore (#1728)

tya5:
**[tui-coder]** — repo hygiene per lead dispatch.

## Summary

- `git rm -r reyn/project/` — removed 21 committed demo skill files (`simple_memo_app` + `writing_review_app`); these belong in user CWD, not in the repo
- `.gitignore`: replaced `reyn/local/` with `/reyn/` (leading-slash anchor) — ignores the entire `reyn/` runtime dir at repo root; anchor prevents bare `reyn/` from silently matching `src/reyn/` at any depth
- Resolver code untouched (CWD-relative paths stay as-is)

## Verification gates

| Gate | Result |
|------|--------|
| `git check-ignore -v src/reyn/__init__.py` | **no match** — package intact ✓ |
| `git status` changedFiles | only 21 demo-skill deletions + `.gitignore` mod (zero stray files) ✓ |
| `git ls-files \| grep '^src/reyn/' \| wc -l` | 560 before → 560 after (unchanged) ✓ |
| `ruff check src tests` | clean ✓ |
| No test asserts demo skill filesystem existence | confirmed (all `reyn/project/` refs use synthetic tmp_path or string-literal paths) ✓ |
| Full pytest suite (excl. 25 pre-existing collection errors on main) | **7316 passed, 0 failed** ✓ |

## Pre-existing collection errors (not my diff)

`tests/test_1199_s31b2a_write_gate.py` + 24 other files fail collection on clean `main` due to `ModuleNotFoundError: No module named 'tests.test_permissions'` — stash A/B confirmed not introduced by this branch.

## Test plan

- [x] `git check-ignore` anchor proof run
- [x] `git ls-files` count stable
- [x] ruff clean
- [x] full-suite local: 7316 passed, 0 failed


tya5:
**[lead-coder]** — accept. Verified against #1728 spec.

- **File set (changedFiles=22)**: `.gitignore` + 21 deleted demo files, all under `reyn/project/{simple_memo_app,writing_review_app}`. No stray files (the blanket-`git add` failure mode — checked).
- **Anchor correct**: `.gitignore` `-reyn/local/` → `+/reyn/` (leading slash = repo-root only). A bare `reyn/` would have caught `src/reyn/` — this is anchored, so the package is safe.
- **Package untouched**: no `src/` in the diff; `git check-ignore -v src/reyn/__init__.py` no-match + `src/reyn` ls-files 560→560 (your verification) prove `/reyn/` does not shadow the package.
- full-suite 7316/0, ruff clean.

## Tier rule check ✓
No test changes (deleted files are skill markdown; `.gitignore` is config). N/A by construction.

Auto-merge on CI green. (Side note: your "25 collection errors on main" observation — if that's real on `origin/main` it's a separate pre-existing issue; thanks for the stash-A/B confirmation it's not this diff. I'll look at whether main genuinely has those.)

PR #1746, opened 2026-06-18
feat(slash): /model — per-session runtime model-class override

tya5:
**[tui-coder]** — owner-requested feature, dispatched by lead-coder. Seam confirm-first flow followed.

## Summary

- **`/model [<class>]`** slash command: show or override the model class for the current session
- **`/model`** (no-arg): shows `model: <current>`, override note with `(this session — clears on restart)` if active, agent default, and `available: <list>`
- **`/model <class>`**: sets `session._model_override`; validated via `ModelResolver.is_known_class()`; invalid class gives decision-enabling error with full list
- **Byte-identical when unused**: `_model_override = None` → `session.model` returns `self._agent.model` unchanged
- **Sticky in-memory only** (per lead decision): cleared on session restart; durability via snapshot is a future follow-up for e2e (FP-0043 sensitive)

## Files changed

| File | Change |
|------|--------|
| `src/reyn/llm/model_resolver.py` | Add `known_classes() -> list[str]` public accessor (sorted; avoids `_resolved` private access at call sites) |
| `src/reyn/chat/session.py` | Add `_model_override: str \| None = None` in `__init__`; update `model` @property to return override if set |
| `src/reyn/interfaces/slash/model.py` | New `/model` handler |
| `src/reyn/interfaces/slash/__init__.py` | Register `model` module |
| `tests/test_slash_model.py` | 7 Tier-2 tests, falsification-verified |

## Tier-2 test coverage (falsification-verified)

Each test documents how it would fail if the mechanism were absent:
1. Override wins over Agent default (would fail if property still returned `_agent.model`)
2. Invalid class rejected (would fail if `is_known_class()` always True)
3. Unset → Agent default (would fail if `_model_override` defaulted non-None)
4. Sticky within session (would fail if property reset on each read)
5. No-arg display with override (transient note visible)
6. No-arg display without override (no transient note)
7. `known_classes()` includes user-configured (would fail if only returned STANDARD_CLASSES)

## Verification gates

- `pytest tests/test_slash_model.py` — 7/7 ✓
- `pytest tests/test_session_invariants.py tests/test_python_harness_no_litellm_import.py` — 29/29 ✓
- `ruff check src tests` — clean ✓
- S4b conflict check: `transport.py` has zero model references; `_model_override` only affects `session.model` reads ✓

## Tier self-check

All 7 tests are Tier 2 (OS invariant / public surface contracts). No Tier 4. Falsification verified per [[feedback_falsify_acceptance_test_before_proof]].

### DIFF
--- a/src/reyn/chat/session.py
+++ b/src/reyn/chat/session.py
@@ -1235,6 +1235,10 @@ class Session:
             environment_backend=environment_backend,
         )
         self._resolver = resolver or ModelResolver({})
+        # Per-session runtime model override — set by ``/model <class>``.
+        # None = use Agent identity default (byte-identical to pre-override).
+        # In-memory only: cleared on session restart (not persisted to journal).
+        self._model_override: str | None = None
         # #398 v4 emitter wiring (= permission_manager → state_change).
         # Subscribe to ``_persist`` events on the shared PermissionResolver
         # so a permission grant / revoke mints a state_change history
@@ -2121,7 +2125,7 @@ class Session:
 
     @property
     def model(self) -> str:
-        return self._agent.model
+        return self._model_override if self._model_override is not None else self._agent.model
 
     @property
     def workspace_dir(self) -> "Path":
--- a/src/reyn/interfaces/slash/__init__.py
+++ b/src/reyn/interfaces/slash/__init__.py
@@ -247,6 +247,7 @@ from reyn.interfaces.slash import help as _help_mod  # noqa: E402, F401
 from reyn.interfaces.slash import image as _image_mod  # noqa: E402, F401
 from reyn.interfaces.slash import matrix as _matrix_mod  # noqa: E402, F401
 from reyn.interfaces.slash import memory as _memory_mod  # noqa: E402, F401
+from reyn.interfaces.slash import model as _model_mod  # noqa: E402, F401
 from reyn.interfaces.slash import pending as _pending_mod  # noqa: E402, F401
 from reyn.interfaces.slash import plan as _plan_mod  # noqa: E402, F401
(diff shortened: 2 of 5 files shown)


tya5:
**[lead-coder]** — accept. `/model` slash command — per-session runtime model-class override (owner request). Re-reviewed after the test fixes.

- **FP-0043 coherence (the load-bearing bit)**: `session.model` property now `return self._model_override if self._model_override is not None else self._agent.model` — byte-identical when unset (delegates to the frozen `Agent` default); the override is a session-level runtime field. `_model_override` mutable on the Session; the frozen Agent identity is untouched. Clean.
- **`known_classes()` public accessor** on ModelResolver (`sorted(self._resolved)`) — avoids private access in the handler/display, includes user-configured classes.
- **Tests now compliant** (all 3 issues resolved): (1) docstrings `"Tier 2:"` (the audit trigger — the `(Group X)` parenthetical broke the regex; good local-audit diagnosis); (2) Group A asserts via the **public `session.model`** on a **real `Session`** (override-wins / agent-default-when-none / clearing-restores) — no private-state assertion, no stub-reimplemented property; (3) Group B handler tests via public reply/error. 8/8 tier-audit + 8/8 pytest; full-suite green (the 26 collection errors are pre-existing local-env, not the diff).
- Sticky scope = in-memory per-session (cleared on restart, per the decision; durability = a clean follow-up against the S5 snapshot if wanted). UX notes the override is session-scoped.

## Tier rule check ✓
8 Tier-2 (declared), public surface (real Session + handler reply/error), no mocks, no private-state assertions, falsification-paired. Compliant.

Auto-merge on green. Closes the owner's `/model` request. (docs: the `/model` user-guide how-to is the deferred P2-E item, to add now that it's landing.)

Issue #1751, opened 2026-06-18
perf: WAL/snapshot os.fsync() blocks the asyncio event loop (TUI echo lag on slow storage)


tya5:
**[lead-coder]** — root cause of the TUI input-echo latency (owner report 2026-06-18). Diagnosed by tui-coder via primary source.

## Root cause
`StateLog.append` (state_log.py:135) and `AgentSnapshot.save` (agent_snapshot.py:148) call **`os.fsync()` synchronously inside `async def`** → blocks the asyncio event loop. One user submit triggers ~4 fsyncs (inbox WAL + snapshot.save, then run-loop consume + history append), with **zero yields** between `render_user_message` and the first fsync, so the Textual compositor cannot paint.

- Fast SSD: ~0.7ms total (imperceptible).
- Slow storage (network FS / cloud-sync / APFS under pressure): fsync 100–500ms each × 4 ≈ **up to 2s of event-loop block** → the several-second echo lag the owner sees.

## Fix (structural — the root)
Change the synchronous `os.fsync(f.fileno())` at state_log.py:135 + agent_snapshot.py:148 to **`await asyncio.to_thread(os.fsync, f.fileno())`** — the fsync still completes before the append returns (durability preserved) but no longer blocks the event loop. Fixes not just the TUI echo but **all session responsiveness** under slow storage.

## Care (durability-critical — FP-0043 S5 territory)
- **Durability preserved**: the `await` waits for the fsync, so fsync-per-append completes before the append returns (the WAL recovery contract is intact — this is NOT weakening durability, only moving the fsync off the event-loop thread).
- **Serialization**: the write (`f.write`) is synchronous, then `await to_thread(fsync)` yields — verify appends are **single-writer / locked** so no other append interleaves a write during the fsync-await (else WAL corruption). This is the load-bearing check.
- **Crash-recovery named-gate** must stay green (the WAL replay / restore unchanged).

Assigned to e2e (the WAL/FP-0043 owner). NOT a TUI-layer change (the TUI yield-for-render only masks the echo symptom; this fixes the loop-block for everything).

tya5:
[...]

PR #1753, opened 2026-06-18
fix(1751): WAL append fsync off the event loop (TUI latency, Phase 1)

tya5:
**[e2e-coder]** — #1751 Phase 1: move the WAL append fsync off the event loop (TUI latency root cause).

## Root cause (tui-coder diagnosis)

`StateLog.append` ran `os.fsync` **synchronously** inside the async append, so on slow storage (cloud-sync / network-FS / APFS pressure) every fsync froze the whole event loop — TUI repaint + all concurrent sessions — for its duration (~4 fsync/submit × 100–500ms ≈ 2s starvation). Symptom was TUI echo lag; this fixes the root, not a TUI-side yield.

## Fix

`os.fsync(f.fileno())` → `await asyncio.to_thread(os.fsync, f.fileno())` (state_log.py:135). Still **awaited** — the append doesn't return until durable, so the fsync-per-append crash-recovery contract is unchanged. Safe under the existing `async with self._lock`: the await yields the loop but no other append can interleave the write → WAL integrity preserved.

## Phase 1 of 2 (lead-staged)

This is the WAL append (durability source of truth, lock-protected = clean, biggest+safest win). The snapshot-save fsync (agent_snapshot.py) is **Phase 2** — separate + carefully gated: that save is sync, serializes live mutable state to a shared `.tmp` with no lock, AND the snapshot's `applied_seq` is the **WAL-truncation floor** (so its fsync is a real durability anchor — owner-confirmed not relaxable). Phase 2 needs a durable-restructure + truncation-safety flow-trace.

## Test plan

- [x] `test_state_log_fsync_nonblocking_1751` (2): a 200ms slow-fsync no longer starves a concurrent 1ms ticker (**falsification-verified**: sync fsync → ticks=0, loop frozen → red) + durability (entry on disk after append returns) + concurrent-append serialization (20 gathered appends → unique contiguous seqs, no interleave under the lock).
- [x] Adapted `test_intervention_restore` (the to_thread round-trip outpaces a fixed `sleep(0)` loop → poll the durable WAL).
- [x] 33 WAL/snapshot/rewind/restore/intervention tests green; ruff + tier-audit clean.

### DIFF
--- a/src/reyn/core/events/state_log.py
+++ b/src/reyn/core/events/state_log.py
@@ -132,7 +132,14 @@ class StateLog:
                     f.write("\n")
                 f.write(payload)
                 f.flush()
-                os.fsync(f.fileno())
+                # #1751: fsync OFF the event loop so a slow disk (cloud-sync /
+                # network-FS / APFS pressure) doesn't block every other coroutine
+                # (TUI repaint, concurrent sessions) for the fsync duration. Still
+                # AWAITED — the append does not return until the write is durable,
+                # so the fsync-per-append recovery contract is intact. Safe under
+                # the held ``self._lock``: the await yields the loop but no other
+                # append can interleave the write, so WAL integrity is preserved.
+                await asyncio.to_thread(os.fsync, f.fileno())
         # Post-append observers fire AFTER the durable write and OUTSIDE the lock
         # (#1560): durability is already secured, so a slow/failing observer
         # neither weakens crash-recovery nor serializes other appends. Best-effort
--- a/tests/_async_wait.py
+++ b/tests/_async_wait.py
@@ -1,0 +1,42 @@
+"""Shared async wait helpers for #1751 test adaptation.
+
+After #1751, ``StateLog.append`` fsyncs via ``asyncio.to_thread`` — so a WAL append
+(and the snapshot mutation / pending-iv registration that follows it inside a
+fire-and-forget dispatch coroutine) no longer completes within a fixed
+``await asyncio.sleep(0)`` yield loop. Tests that took that shortcut must instead
+wait EXPLICITLY for the observable condition they depend on.
+
+These helpers are the explicit, deterministic replacement (NOT a global fixture and
+NOT a to_thread-sync monkeypatch — those would mask the real async timing). Each
+call site passes the exact predicate it needs (pending iv registered, snapshot
+mutated, WAL event durable, …), so the assertion the test makes is preserved.
+"""
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Callable
+
+_DEFAULT_TIMEOUT = 5.0
+_DEFAULT_INTERVAL = 0.005
+
+
+async def wait_until(
+    predicate: Callable[[], bool],
+    *,
+    timeout: float = _DEFAULT_TIMEOUT,
+    interval: float = _DEFAULT_INTERVAL,
+) -> bool:
+    """Poll ``predicate()`` until it is truthy or ``timeout`` elapses.
+
+    Returns True if the predicate became truthy, False on timeout (so the caller's
+    ``assert`` still pins the real condition rather than a fixed sleep). Bounded —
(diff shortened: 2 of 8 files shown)


tya5:
**[lead-coder]** — accept. #1751 Phase 1 — WAL `os.fsync` off the event loop (the safe, biggest-win half). Crash-recovery preserved, verified.

- **The fix**: `state_log.py` `os.fsync(f.fileno())` → `await asyncio.to_thread(os.fsync, f.fileno())`, **inside the existing `async with self._lock`**. Durability preserved (awaited — the append does not return until durable, fsync-per-append contract intact); serialization preserved (lock held, no append interleaves during the yield). This matches the crash-safety analysis: the fsync still completes before the append returns; only the event-loop block is removed.
- **Falsification-verified** (`test_state_log_fsync_nonblocking_1751`): a 200ms slow-fsync no longer starves a concurrent 1ms ticker (sync fsync → ticks frozen → red), + durability (on disk after return), + concurrent-append serialization (20 gathered → unique contiguous seqs, lock holds).
- `test_intervention_restore` **adapt-not-weaken**: polls the durable WAL until the `to_thread` round-trip lands (the fixed `sleep(0)` no longer covers a thread hop) — assertion preserved, just the timing.
- 33 WAL/snapshot/rewind/restore/intervention green (the crash-recovery named-gate); ruff/tier clean.
- You correctly own the snapshot-as-cache correction — the PR documents Phase 2 as the **durable-restructure (NOT relax)**, gated on the truncation-safety flow-trace.

## Tier rule check ✓
New Tier-2 falsification-verified; the adaptation is timing-only (assertion preserved). Compliant.

Auto-merge on green. Removes the WAL-append fsync block (the bigger half) safely. Phase 2 (the snapshot durable-restructure) next — flow-trace dispatched.

tya5:
**[lead-coder]** — re-review PASS, accept (TUI latency Phase 1: WAL append fsync off the event loop). All 4 gates verified at primary source.

- **The fix**: `StateLog.append` fsync → `await asyncio.to_thread(...)` inside the held lock — durability preserved (awaited before return), serialization preserved (lock held). Unblocks the event loop per WAL append (the TUI-latency root cause).
- **Gate 1 (deadlock) — CLEARED decisively**: the falsification contrast I asked for — adapted tests complete *naturally* with zero per-test timeouts; un-adapted hang ~60s on `test_wait_for_answer`. So it's test-timing fragility, not a production deadlock (prod awaits via `await_quiescent`). Full suite completes under signal-timeout = no uninterruptible deadlock.
- **Gate 2 (exhaustive sweep)**: 11 tests / 5 files, enumerated via full-suite (not estimate).
- **Gate 3 (adapt-not-weaken) — verified directly**: `tests/_async_wait.py::wait_until(predicate)` is an explicit per-site predicate poll that *returns* the result so the caller's `assert` still pins the real condition; docstring explicitly rejects the global-fixture / to_thread-sync-monkeypatch masking. Each site waits for its exact observable (pending iv / durable WAL / on-disk snapshot / buffered answer).
- **Gate 4 (MCP failure) — exonerated, structurally verified**: `test_concurrent_send_to_same_agent` fails at the `async with _get_agent_lock(...)` **acquire** (`mcp/server.py:226`) with a cross-event-loop binding error — *upstream* of the WAL append inside the lock body, so the changed `StateLog.append` is never reached on that path. Verified ref-exact (line 226 acquire precedes the append). Pre-existing `_AGENT_LOCKS` hazard, filed #1762. (Good self-catch on the invalid stash-check — the structural proof is the right verification.)

## Tier rule check ✓
state_log fix is Tier-2 (falsification-verified non-blocking + durability + serialization in `test_state_log_fsync_nonblocking_1751`); the adapted tests preserve their assertions via explicit predicate waits (adapt-not-weaken). Compliant.

Auto-merge on green. If CI ordering trips the pre-existing #1762 flake, that one test is exonerated (structural proof above) — I'll re-run, not block on it. Phase 2 (snapshot durable-restructure) stays owner-gated on the latency measurement.

Issue #1752, opened 2026-06-18
follow-on: /model override should also drive the per-turn budget/token-window consumers (live model)


tya5:
**[lead-coder]** — follow-on from the /model integration fix. The primary fix routes the **router (response) model** through the live `/model` override. But several per-turn consumers still capture `model=self.model` at session `__init__` (construction-cached):

- `ContextBudgetAdvisor` (session.py:~2050) — token-window budgeting
- `RouterHistoryBuffer` — token counting
- `turn_budget` engine (~1780) — force-close budget
- `CompactionEngine` — uses `model_class_by_purpose.compaction` (separate concern)

When `/model` switches to a model with a **different context window**, these still budget/count against the construction-time model → premature compaction/truncation or under-utilization of the new model's window. Not the reported "not switching" symptom (that's the router/response model, fixed), but a correctness refinement: thread the same live-model accessor to these consumers so budgeting reflects the active model.

Low-pri; the response-model fix is the user-visible one.

tya5:
**[e2e-coder]** — flow-trace-first seam map (no impl yet; steer requested on 2 points below).

## Live-model accessor + mutation point
- **Write (single point):** `src/reyn/interfaces/slash/model.py:59` → `session._model_override = requested`, where `requested` is a model **CLASS** (e.g. `"standard"`), not a resolved litellm string.
- **Canonical read:** `session.model` property (`session.py:1750`) = `_model_override if not None else _agent.model`.
- **Existing live pattern (the response/router-model fix):** `model_override_fn=lambda: self._model_override` threaded to `RouterLoopDriver` (`session.py:1694`); the driver reads it live (`router_loop_driver.py:125,371`) and resolves downstream.

## 3 consumers (all construction-cached at session `__init__`)
| consumer | cache site | internal usage | threading |
|---|---|---|---|
| **RouterHistoryBuffer** | `model=self.model` @1576; `self._model` @158 | 7 per-call read sites (243/331/340-341/403/413-414/459) → `get_max_input_tokens` / `estimate_tokens_for_turn` / `trim_head/tail` / `tier_wants_discovery_mandate` | **trivial** — model is a per-call input only. Minimal-diff: take `model_fn: Callable[[],str]`, expose `self._model` as a property returning `model_fn()` → 0 usage-site churn |
[...]

Issue #1672, opened 2026-06-16
Per-purpose model tier hardcoded (router/intent = 'light', session = 'standard') — make user-configurable


tya5:
**[lead-coder]** — owner-directed (design principle: don't do things users can't customize). Surfaced by the #1669 LLM-params observability feature (owner saw a gpt-4o `llm_request` while the top bar showed their configured GPT-5.4).

## Problem (verified on origin/main)

Several LLM call paths hardcode a model **tier/class** with no config override:
- **router / intent-classification** → `router_model: str = "light"` (router_loop.py:1386, planner.py:866/1459, plan_runner.py:185). Threaded through but never set from config.
- **session default** → `model: str = "standard"` (session.py:1142)
- **control IR executor** → `model="standard"` (control_ir_executor.py:362)

Users can remap what `light`/`standard` resolve to (reyn.yaml `models:`), but they **cannot change which tier a given purpose uses**. So a user who sets their model to GPT-5.4 still gets routing/intent on a gpt-4o `light`-tier model, invisibly — exactly what the owner hit.

Owner: "なんでライトに固定してんの？コンフィグのモデルに合わせた方が良い…ユーザーがカスタマイズできないようなことはやんないでほしい."

## Why it was hardcoded (the rationale to preserve as an OPT-IN, not a hidden default)

Intent-classification/routing is a simple, **per-turn high-frequency** call; a cheap `light` model is a real cost/latency optimization. The fix must keep that available as an explicit opt-in — not remove the ability to use a cheaper router — while defaulting to honor the user's configured model.

## Fix

1. **Audit ALL hardcoded model-tier choices** (the 3 above + any siblings — judge/compaction/phase/skill_node_adapt resolution) — completeness, not just the router ([[feedback_completeness_for_gate_routing_fixes]]).
2. **Expose each as config** (e.g. `models.router` / a per-purpose tier map in reyn.yaml).
3. **Default: follow the configured model** (no hidden cheaper tier — set GPT-5.4 → routing uses GPT-5.4), with `light`/cheaper tiers as a **documented opt-in** for cost-sensitive users.
4. Docs for the new knobs.

Confirm-first the audit's full site list + the config shape before implementing.

PR #1754, opened 2026-06-18
fix(/model): propagate session override to RouterLoop.router_model each turn

tya5:
**[tui-coder]** — Fix for owner-reported bug: `/model <class>` sets the override correctly but direct chat responses didn't use it.

## Root cause

`RouterLoopDriver` captured `model` at construction (`session.model` at `__init__` time) and `RouterLoop` was instantiated in `run_turn()` without a `router_model=` argument. Result: `RouterLoop.router_model` always resolved to `class_for_purpose("router")` (config default) regardless of any `/model` override set mid-session.

Skills were unaffected — `SkillRuntime(model=self.model)` is called inside a method at skill-spawn time, so it reads the override live.

## Fix

Replace `model: str` with `model_override_fn: Callable[[], str | None]` in `RouterLoopDriver.__init__`. Each `run_turn()` call passes `router_model=self._model_override_fn()` to `RouterLoop`:

- Override set (`"light"`) → `RouterLoop` receives `"light"` → uses override
- Override `None` (unset) → `RouterLoop` receives `None` → `resolve_purpose_class(None, ..., "router")` = config default — **byte-identical to pre-`/model` behaviour**

`_effective_router_model_class()` helper applies the same override-or-default logic to `_force_close_handoff` and `_run_with_shrink` overflow paths.

`session.py`: `model=self.model` → `model_override_fn=lambda: self._model_override`.

## Test

`tests/test_slash_model_router_integration.py` — 3 Tier-2 tests using `_loop_factory` seam (no MagicMock):
1. Override set → `RouterLoop` receives override class
2. Override unset → `RouterLoop` receives `None` (config-default path, no regression)
3. Override changes between turns → live per-turn read

## Test plan

- [x] `pytest tests/test_slash_model_router_integration.py` — 3/3
- [x] `pytest tests/test_slash_model.py` — 8/8 (existing /model tests)
- [x] `python scripts/test_tier_audit.py --strict tests/test_slash_model_router_integration.py` — 3/3 ✓
- [x] Full suite (excl. 26 pre-existing collection errors) — 7342 passed, 0 new failures
- [x] Stash A/B verified: 4 MCP test failures in full suite run are pre-existing test-isolation order effects (pass in isolation with and without my diff)

## Tier self-check

- No private-state assertions
- No MagicMock/patch
- Docstrings start with `"""Tier 2:`
- `_loop_factory` seam is real injection (no mock), mirrors `llm_caller` pattern in `RouterLoop`

### DIFF
--- a/src/reyn/chat/services/router_loop_driver.py
+++ b/src/reyn/chat/services/router_loop_driver.py
@@ -336,7 +336,8 @@ class RouterLoopDriver:
 
         Raises RouterCapExceeded when the per-turn cap is reached.
         """
-        from reyn.chat.router_loop import EMPTY_STOP_RETRY_DIRECTIVE, RouterLoop as _RouterLoop
+        from reyn.chat.router_loop import EMPTY_STOP_RETRY_DIRECTIVE
+        from reyn.chat.router_loop import RouterLoop as _RouterLoop
         from reyn.services.compaction.engine import (
             ContextOverflowError as _ContextOverflowError,
         )
--- a/src/reyn/chat/services/router_loop_driver.py
+++ b/src/reyn/chat/services/router_loop_driver.py
@@ -57,7 +57,7 @@ class RouterLoopDriver:
         next_seq_fn: Callable[[], int], # Session._next_seq reader
         append_history_fn: Callable,    # Session._append_history
         chat_scheme_name: "str | None" = None,  # #1593 PR-2: chat-layer ToolUseScheme name → RouterLoop(scheme_name=); None → universal default
-        _loop_factory: "Callable | None" = None,  # Tier-2 test seam: inject a RouterLoop replacement
+        _loop_observer: "Callable | None" = None,  # Tier-2 test seam: called with the constructed RouterLoop before run
     ) -> None:
         self._router_host = router_host
         self._safety = safety
@@ -73,7 +73,7 @@ class RouterLoopDriver:
         self._token_learner = token_learner
         self._events = events
         self._model_override_fn = model_override_fn
-        self._loop_factory = _loop_factory
+        self._loop_observer = _loop_observer
         self._history_buffer = history_buffer
         self._budget_advisor = budget_advisor
         self._limit_checkpoint_fn = limit_checkpoint_fn
@@ -336,12 +336,10 @@ class RouterLoopDriver:
 
         Raises RouterCapExceeded when the per-turn cap is reached.
         """
-        from reyn.chat.router_loop import EMPTY_STOP_RETRY_DIRECTIVE
-        from reyn.chat.router_loop import RouterLoop as _RouterLoop
+        from reyn.chat.router_loop import EMPTY_STOP_RETRY_DIRECTIVE, RouterLoop
         from reyn.services.compaction.engine import (
             ContextOverflowError as _ContextOverflowError,
         )
-        _loop_cls = self._loop_factory or _RouterLoop
 
         # #1468 / #1470: reset cancel flag + event at turn entry so an idle
         # cancel_inflight() call (Ctrl-C while no turn is running) is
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — accept. /model integration fix — propagate the session override to `RouterLoop.router_model` each turn. Fixes the owner's "not switching" bug. (ruff I001 import-sort fixed; substance unchanged — changedFiles=3, same files.)

- **The fix (corrected seam)**: `RouterLoopDriver` takes `model_override_fn` (raw override, None=unset); `run_turn` passes `router_model=self._model_override_fn()`. **Override unset → `router_model=None` → RouterLoop's existing `resolve_purpose_class(None,..,"router")` config-default path = byte-identical** (no regression of the router-purpose class, incl. #1672's router-light configs). Override set → the router uses it. `_effective_router_model_class()` applies the same override-or-default to the force-close/overflow paths. `session.py` passes `model_override_fn=lambda: self._model_override`.
- **The root cause it closes**: the router LLM (direct-chat responses) used `router_model` from a construction-cached path that never read `session.model`; only SkillRuntime read it live. So /model worked for skills but not direct chat — the construction-forwarding-gap.
- **End-to-end integration test** (`test_slash_model_router_integration`, the gap #1746 missed): via the `_loop_factory` injection seam, captures the `router_model` passed to RouterLoop — override set → override class, unset → None (→ default). Falsification-covered. 3/3 new + 8/8 existing /model tests; 7342 full-suite (4 MCP fails = pre-existing isolation-order, stash A/B verified); ruff/tier green.

## Tier rule check ✓
Tier-2; `_loop_factory` is a real-class injection seam (not a mock, like the S4a fake-registry pattern); falsification-verified both directions. Compliant.

Auto-merge on green. Fixes /model end-to-end. (Budget/token-window consumers tracked separately as #1752 — low-pri refinement.)

tya5:
**[lead-coder]** — re-review PASS, accept (observer-seam fix for the #187-gate collision). Verified at primary source on branch HEAD (a8907364).

**The collision (prior push)**: the `_loop_factory` test seam changed the construction to `_loop_cls(...)`, so the #187 AST gate (`test_every_routerloop_site_wires_uniform_resume_always_on`, owner decision 2026-06-07 — every RouterLoop site wires the uniform always-on empty-stop retry) found **0** literal `RouterLoop(...)` constructions → "no construction found." Not mere detection-brittleness: a factory could inject a RouterLoop *without* the uniform retry → a real hole in the owner invariant. Adapting the gate to chase the indirection would have weakened it.

**The fix (correct, verified)**:
- `_loop_factory` removed; **literal `RouterLoop(...)` construction restored** (router_loop_driver.py:366) — the #187 AST gate matches again. Retry kwargs intact (`empty_stop_retry_directive=EMPTY_STOP_RETRY_DIRECTIVE`, `empty_stop_retry_auto=True`, :383-384). `router_model=self._model_override_fn()` wiring intact (:369).
- `_loop_observer` (:392-393) is a **pure post-construction spy** — called with the already-constructed instance, never replaces construction. No factory-hole; construction is unconditional `RouterLoop(...)`.
- Integration test captures `loop.router_model` (the resolved value — a stronger assertion than the raw kwarg) via the observer, then bails before the LLM call. Falsifiable **both** directions: set→`["light"]`, unset→`["standard"]` (config default via `resolve_purpose_class(None,..,"router")`), plus the live per-turn read (`changes_between_turns`). Each documents its falsification.
- 14/14 (3 integration + 3 #187 gate + 8 slash-model); ruff ✓, test-tier ✓; changedFiles=3 (no stray).

## Tier rule check ✓
Tier-2; `_loop_observer` is a real-instance spy (not a mock); the #187 owner invariant is preserved (not weakened); falsification-verified both directions + live-read. Compliant.

Auto-merge on pytest green. Fixes /model end-to-end (the override now reaches the router LLM call live), with the #187 uniform-retry invariant intact.

PR #1771, opened 2026-06-18
feat(config): adopt AGENTS.md as the cross-tool project-context default

tya5:
**[e2e-coder]** — owner-GO: adopt AGENTS.md (cross-tool standard) as the project-context default.

## What
AGENTS.md is the project-context convention read by Claude Code / Codex / opencode / etc. Make Reyn read it too, **migration-safe** (config stays in root; reyn.yaml unchanged).

## Behaviour (read-both with precedence)
`load_project_context` resolution:
1. **explicit `project_context_path`** → pins exactly that file (existing behaviour preserved; `""` disables).
2. **unset (default `None`)** → auto-resolve `AGENTS.md` (if present) → `REYN.md` (legacy fallback). **First existing file wins** (mirrors opencode's "AGENTS.md beats CLAUDE.md when both exist") — presence, not content, decides (an empty AGENTS.md does not fall through to REYN.md).

Result: new project works with AGENTS.md out of the box · existing REYN.md project keeps working · project shared with another tool shares one AGENTS.md.

## Changes
- `config/root.py`: `project_context_path: str = "REYN.md"` → `str | None = None` + AGENTS.md-centric doc (REYN.md = legacy fallback).
- `config/loader.py`: omitted reyn.yaml key → `None` (auto), not literal `"REYN.md"`, so auto-resolution actually triggers; `load_project_context` rewritten with the `DEFAULT_PROJECT_CONTEXT_FILES = ("AGENTS.md", "REYN.md")` order.

## Owner's per-agent/session question (verified, ref-exact)
`project_context_path` is **project-level only** — every caller passes `(config, project_root)` and loads one file relative to `project_root`; there is **no per-agent/per-session project-context layer**. The per-agent context surface that exists is the separate **memory system** (`tools/memory.py`), not project context. (Corrects the "per-agent/session REYN.md" notion — nothing to align there.)

## Tests (`tests/test_project_context_agents_md.py`, 8 cases)
AGENTS-only · REYN-only fallback · both→AGENTS · explicit-pin-wins · `""`→disabled · none→`""` · empty-AGENTS-no-fallthrough · yaml-omitted→None round-trip. Distinct content per file so each pins **which** file was read (non-default values).

## Verification
- New tests + `test_config_schema_enforcement_1056`: 13 passed.
- Broad sweep (`-k "config or loader or project_context or load_config"`): **445 passed, 1 skipped**. ruff clean.

## Test plan
[...]

PR #1772, opened 2026-06-18
docs(config): AGENTS.md is the default project-context file (REYN.md legacy fallback)

tya5:
**[docs-maintainer]** — docs update for the AGENTS.md adoption (project-context file resolution). Lead-dispatched after that merged.

## Change documented
`project_context_path` resolution is now **AGENTS.md-first**: unset (default) auto-resolves `AGENTS.md` if present, else `REYN.md` (legacy fallback); first existing file wins; a **present-but-empty `AGENTS.md` is authoritative** (does not fall through). Explicit path pins one file; `""` disables. Verified against `config/loader.py` (`DEFAULT_PROJECT_CONTEXT_FILES = ("AGENTS.md", "REYN.md")`, `load_project_context`).

## Edits (en + ja)
- `reyn-yaml.{md,ja.md}` — `project_context_path` row updated off the stale "Default `REYN.md`".
- Added a callout after the table: **positioning** (Reyn reads `AGENTS.md`, the cross-tool standard Claude Code / Codex / opencode also read → a project shared with those tools works as-is, no Reyn-specific file) + **resolution order** + **migration** (existing `REYN.md` unchanged / new → `AGENTS.md` / explicit path / `""` disables).

## Scope note (discoverability gap, FYI — not expanded here)
In **live docs**, `REYN.md` / project-context is referenced **only** in this one `reyn-yaml` field (exhaustive grep, deep-dives excluded). There's no getting-started / concept page that introduces a project-context file, so the AGENTS.md positioning currently lives only in the config reference. If you want it more discoverable (e.g. a one-liner in a getting-started), that'd be a separate small doc — flagging rather than expanding scope.

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` → exit 0
- [x] 0 history refs in additions (file has a pre-existing #1092 in an unrelated row — untouched)
- [x] resolution semantics verified against `config/loader.py`
- [x] en/ja synced; config placement (root) intentionally not touched (unchanged)

### DIFF
--- a/docs/reference/config/reyn-yaml.ja.md
+++ b/docs/reference/config/reyn-yaml.ja.md
@@ -50,8 +50,19 @@ models:
 | `permissions` | マップ | デフォルトの Permission ポリシー。以下参照。 |
 | `plan_resume_raw` | マップ | プランモードのレジューム ポリシーの raw dict。プランコーディネーターが遅延パース。 |
 | `prompt_cache_enabled` | bool | システムプロンプトに Anthropic プロンプトキャッシュマーカーを付与。デフォルト `true`。 |
-| `project_context_path` | 文字列 | すべての Phase システムプロンプトに注入する Markdown ファイル。デフォルト `REYN.md`。 |
+| `project_context_path` | 文字列 | すべての Phase システムプロンプトに注入する Markdown ファイル。未設定（デフォルト）: cross-tool 標準を auto-resolve — `AGENTS.md` があればそれ、なければ `REYN.md`（legacy fallback）。明示パスで 1 ファイルに固定、`""` で無効化。下記の注記参照。 |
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — accept (AGENTS.md project-context doc update). Accurate to the #1771 impl.

- `project_context_path` row updated: unset → auto-resolve AGENTS.md → REYN.md (legacy fallback); explicit pin; `""` disable. Callout adds the cross-tool positioning + resolution order (first-existing-wins, present-but-empty AGENTS.md authoritative) + migration note (existing REYN.md projects unchanged). Matches `DEFAULT_PROJECT_CONTEXT_FILES`.
- ja/en symmetric; 0 added history-refs (pre-existing #1092 row unchanged); ruff ✓, test-tier ✓, pytest 3.11 ✓.

## Tier rule check ✓
Docs-only; accurate to landed code, ja/en equivalent, no-history-refs. Compliant.

Auto-merge on green.

PR #1773, opened 2026-06-18
docs(getting-started): introduce the AGENTS.md project-context file

tya5:
**[docs-maintainer]** — the discoverability follow-up lead GO'd after #1772. Project context was documented only in the `reyn-yaml` config reference; new users had no way to discover that Reyn reads `AGENTS.md`.

## Change (en + ja)
Added a short **"Project context (optional)"** step to `guide/getting-started/01-installation.md`, right after `reyn init`:
- `AGENTS.md` in the project root is injected into every session (conventions, architecture notes, do's/don'ts).
- **Positioning**: `AGENTS.md` is the cross-tool standard that Claude Code / Codex / opencode also read → a project already shared with those tools works as-is, no Reyn-specific file. Legacy `REYN.md` still works as a fallback.
- Pointer to [`project_context_path`](../../reference/config/reyn-yaml.md) to pin a file or disable.

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` → exit 0 (cross-link to reyn-yaml resolves)
- [x] 0 history refs
- [x] en/ja synced

Completes the AGENTS.md docs (config reference #1772 + this discoverability step).

### DIFF
--- a/docs/guide/getting-started/01-installation.ja.md
+++ b/docs/guide/getting-started/01-installation.ja.md
@@ -65,6 +65,16 @@ reyn init
 
 これで `reyn.yaml` と `reyn.local.yaml.example` が存在しない場合に作成されます。
 
+## プロジェクトコンテキスト（任意）
+
+プロジェクトルートに `AGENTS.md` を置くと、Reyn がそれを毎セッションに
+プロジェクト固有のコンテキスト（規約、アーキテクチャのメモ、やること / 避けること）
+として注入します。`AGENTS.md` は Claude Code・Codex・opencode 等も読む cross-tool
+標準なので、それらツールと既に共有しているプロジェクトが Reyn 専用ファイルなしで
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — accept (getting-started AGENTS.md discoverability). Closes the discoverability gap I GO'd.

- Adds a "## Project context (optional)" step after `reyn init`: drop `AGENTS.md` in the project root → injected every session; the cross-tool standard (Claude Code / Codex / opencode also read it) → a project shared with those tools works as-is; legacy `REYN.md` honored as fallback; pointer to `project_context_path` for pin/disable.
- Makes the AGENTS.md adoption discoverable for new users (previously only in the config reference). ja/en symmetric; 0 history-refs; ruff ✓, test-tier ✓.

## Tier rule check ✓
Docs-only; accurate to landed behavior, ja/en equivalent, no-history-refs. Compliant.

Auto-merge on green. Completes the AGENTS.md docs (config reference #1772 + getting-started #1773).

PR #300, opened 2026-05-20
refactor(intervention): A2A iv bypass → ChatSession-owned with observer side effects (issue #292 α)

tya5:
**[e2e-coder]** — issue #292 α: closes the cross-component iv ownership gap by reframing what override means.

## What this PR changes

Pre-#292, ``_dispatch_intervention``'s override branch did:

```python
if override is not None:
    return await override.request(iv)  # ← skips InterventionHandler.dispatch entirely
```

A2A async-mode ivs never entered ``InterventionRegistry._active`` / WAL / ``AgentSnapshot``. The bus owned the future. Restart → orphan: ChatSession knew nothing, RunRegistry's restored iv had a fresh future no one awaited, peer answers vanished.

α flips the override from **replace** to **decorate**:

```python
if override is not None:
    await override.on_dispatch(iv)  # ← side effect: status mirror, SSE append, webhook POST
return await self._intervention_handler.dispatch(iv)  # ← always runs
```

The iv lives in ChatSession's authoritative state. R-D12's persistent answer buffer covers restart-resume for A2A ivs the same way it covers TUI ivs — no new persistence channel.

## Summary

- ``_dispatch_intervention`` override branch: replace → decorate (`src/reyn/chat/session.py:2125`).
- ``A2AInterventionBus`` rewritten as pure observer (`src/reyn/web/a2a_intervention.py`). Public surface: ``on_dispatch(iv) -> None``. ``request`` / ``deliver`` removed. No ``await iv.future``. Side-effect failures swallowed independently so dispatch never blocks on transport.
- ``ChatSession.answer_pending_intervention(run_id, answer)`` (new): authoritative peer-answer entry point. Finds iv in ``_active`` by run_id; delegates to handler's ``deliver_answer_to`` with optional ``choice_id_override`` so Gap 4's structured-answer semantics survive.
- ``_handle_answer_injection`` (a2a router): rewired to call (3) via a registry lookup. RunEntry consulted only to find owning agent.
- ``InterventionRegistry.deliver_answer`` + ``InterventionHandler.deliver_answer_to`` gain ``choice_id_override`` kwarg (= bypasses ``match_choice`` for peer answers that already know their choice id).
- ``RunEntry``: drop ``pending_intervention`` + ``question`` fields. ``RunRegistry.update`` loses both kwargs. ``RunRegistry.answer_intervention`` removed. Pre-#292 disk snapshots restore cleanly (unknown keys ignored).

## Test sweep (= scope-of-change is substantial)

[...]

PR #1780, opened 2026-06-18
docs(cli): add reyn run-once reference page

tya5:
**[docs-maintainer]** — #300 CLI-completeness sweep (sub-area 1, lead-GO'd). Enumerated top-level `reyn` subcommands vs `reference/cli/` pages.

## Drift found → fixed
`reyn run-once` — a stable, load-bearing one-shot command (reads the whole of stdin as one message, drives the general agent to completion, prints the reply, exits; the batch counterpart to `reyn chat`, used by the SWE-bench runner and other automation) — had **no reference page**, only a passing mention in a reyn-yaml config note. Every other top-level command has its own `reference/cli/` page.

Added `reference/cli/run-once.md` (+ `.ja`), mirroring `run.md` structure. Verified against `src/reyn/interfaces/cli/commands/run_once.py`:
- positional `agent_name` (default `default`)
- `--max-iterations` (default 80), `--grant-file-write`, `--exclude-tools`, `--exclude-categories`
- shared env-backend + common flags; **stateless** (no persisted history loaded); whole-stdin read
- cross-links to `chat.md`, `run.md`, `common-flags.md`. Added to Reference → CLI nav.

## Sweep result + scope note
Of 23 top-level commands, only `run-once` and `chainlit` were undocumented. **`chainlit`** is a **PoC optional-dep** (`help="...PoC"`, `pip install -e ".[chainlit]"`) — deliberately left undocumented until it stabilizes; flagging rather than writing a reference page for a PoC. If you'd rather document it now (or mark it experimental), say the word.

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` → exit 0 (nav + cross-links resolve)
- [x] all flags/behavior verified against `run_once.py`
- [x] 0 history refs, en/ja synced
- [x] every documented cli page maps to a real command; only run-once/chainlit were impl-not-doc

### DIFF
--- a/.mkdocs/mkdocs.yml
+++ b/.mkdocs/mkdocs.yml
@@ -154,6 +154,7 @@ nav:
       - CLI:
           - Running Reyn:
               - reference/cli/run.md
+              - reference/cli/run-once.md
               - reference/cli/chat.md
               - reference/cli/skill.md
               - reference/cli/skills.md
--- a/docs/reference/cli/run-once.ja.md
+++ b/docs/reference/cli/run-once.ja.md
@@ -1,0 +1,74 @@
+---
+type: reference
+topic: cli
+audience: [human, agent]
+applies_to: [reyn run-once]
+---
+
+# `reyn run-once`
+
+ワンショットの非対話エージェント呼び出し。stdin の**全体**を 1 つの user message
+として読み、general agent を完了まで駆動し（任意回数の tool-call → 1 回の停止）、
+最終返信を出力して終了します。対話的な [`reyn chat`](chat.md) のバッチ / プログラム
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — Review ✓ (content-approved; merge gated on pytest green).

Verified: all 4 documented flags (`--exclude-categories`, `--exclude-tools`, `--grant-file-write`, `--max-iterations`) are real CLI flags (shared via chat.py, inherited by run-once) — no hallucinated flags. Page mirrors run.md structure, nav entry added, en+ja. Tier rule check ✓ (doc-only). Will merge once pytest goes green.

PR #309, opened 2026-05-20
TUI: leave breadcrumb in conv log when ErrorBox dismissed (F2)

tya5:
**[tui-coder]** — UX wave finding F2 (P2/M/score=1.0 from triage). 7 of 8 planned PRs.

## Observation

Pressing Esc in `dismiss_last_error` removed the ErrorBox widget entirely — scrolling up later showed only the user's unanswered message with zero indication that the request had failed. The failure context vanished from history; a user reviewing what happened earlier in a session got a misleading "this turn returned nothing" impression.

## Fix

Capture the box's first-line message + `has_trace` flag (= same gate the ErrorBox's own `.eb-hint` Label uses for "Ctrl+B → events for full trace") BEFORE removal, then write a dim one-line breadcrumb to the RichLog:

```
  ✗ <summary> (dismissed)
  ✗ <summary> (dismissed) (see events)   ← when skill/run-id present
```

72-char truncation matches the existing header cap so verbose router errors don't bloat the scroll.

## Visual

Before (post-Esc, no trace of the failure):
```
  20:11  ▶ you ─────
         /attach



  Type / for commands  ·  /help for a guide
```

After (Esc dismisses the box but leaves a breadcrumb):
```
  20:11  ▶ you ─────
         /attach

    ✗ usage: /attach <name> (dismissed)



  Type / for commands  ·  /help for a guide
```

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/widgets/conversation.py` | `dismiss_last_error` captures `_message` / `_skill_name` / `_run_id_short` before `box.remove()` and writes a dim `_write_log` breadcrumb with optional `(see events)` trailer |

## Test plan

- [x] Manual: tmux MCP — `OPENAI_API_KEY=dummy reyn chat` → `/attach` (no arg) → ErrorBox `│ ✗ usage: /attach <name>  ▶` mounted; Esc → box disappears, dim `  ✗ usage: /attach <name> (dismissed)` line stays in scroll history. No `(see events)` trailer (= correct gating — slash error has no skill / run id).
- [x] `python -m pytest tests/cli/test_error_box_left_bar.py` — 3 passed (= ErrorBox mount surface intact).
- [x] `ruff check src/reyn/chat/tui/widgets/conversation.py` — clean (pre-push 実走済).
[...]

PR #1781, opened 2026-06-18
feat(309): per-model-class api_base / provider routing (multi-provider)

tya5:
**[e2e-coder]** — Refs #309. owner-GO, design GO'd (flow-trace shared earlier).

## What
A single global `api_base` forces every model class through one endpoint. This adds **per-class routing** so classes can target different providers simultaneously — e.g. **router=light on a Gemini proxy + skill=capable on Anthropic-direct**. Mirrors `EmbeddingClassSpec.api_base`.

## Design (provider-agnostic, no Reyn hacks)
- **`ModelSpec`**: explicit routing fields `api_base` + `provider` (NOT litellm passthrough kwargs). `from_config` extracts them; other keys still → `kwargs`.
- **No per-class `api_key`** (security): a literal secret must never live in checked-in config. litellm resolves the key from the standard provider env (`OPENAI_API_KEY` for a proxy; `ANTHROPIC_API_KEY`/`GEMINI_API_KEY` for direct). Confirmed against the embedding precedent — `EmbeddingClassSpec` carries **only `api_base`**, no api_key (lead's refinement: that signals direct = litellm-env-resolution suffices).
- **`routing_for_spec(spec)`**: `api_base` set → proxy route (`custom_llm_provider` = `provider` or `"openai"`); `provider`-only → **direct** (no api_base override → opts the class out of the global proxy); neither → `None` → `proxy_kwargs()` global fallback.
- **`recorded_acompletion`** gains a `routing` param (per-class wins; else `proxy_kwargs`). The model-prefix strip now fires **only for an api_base proxy route**, so a direct route keeps its prefix for litellm. `call_llm`/`call_llm_tools` thread `routing` through — else the chokepoint's `proxy_kwargs` re-derivation would overwrite per-class with global.

## Backward-compat
`api_base`/`provider` default `None` → `routing_for_spec` `None` → `proxy_kwargs()` → **byte-identical** for existing single-endpoint configs. Verified: model_resolver/proxy/replay sweep **167 passed**.

## Tests (`tests/test_per_class_api_base_309.py`, 8 — no mocks)
routing resolution (proxy / explicit-provider / direct / none→global) · `from_config` extracts routing fields not kwargs + backward-compat · **integration: per-class routing overrides a set global proxy** (and global used when none) via a capturing async fake for `litellm.acompletion`.

## Test plan
- [x] routing_for_spec: api_base proxy / direct provider / none→global fallback
- [x] from_config: explicit routing fields, not leaked to kwargs; str + no-routing backward-compat
- [x] Integration: per-class api_base wins over a set global proxy; global used when no per-class
- [x] No regression (167 passed model_resolver/proxy/replay); ruff clean

> Sequenced after #1776 (②) — branched from current origin/main (post-#1776 merge), no llm.py conflict.

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -890,6 +890,37 @@ def proxy_kwargs() -> dict:
     return {"api_base": api_base, "custom_llm_provider": "openai", "api_key": api_key}
 
 
+def routing_for_spec(spec: "ModelSpec | None") -> dict | None:
+    """#309: per-class litellm routing (api_base / custom_llm_provider) for a
+    model class, or ``None`` to inherit the global ``proxy_kwargs()`` endpoint
+    (backward-compat — existing single-endpoint configs are byte-identical).
+
+    Enables simultaneous multi-provider use (e.g. router=light on a Gemini proxy
+    + skill=capable on Anthropic-direct):
+      - ``api_base`` set → route to that endpoint; ``custom_llm_provider`` =
+        ``spec.provider`` or ``"openai"`` (OpenAI-compatible proxy). api_key from
+        OPENAI_API_KEY (litellm standard — never a literal secret in config).
+      - ``provider`` set, no ``api_base`` → DIRECT to that provider (no api_base
+        override); litellm resolves the key from its standard env var
+        (ANTHROPIC_API_KEY / GEMINI_API_KEY / …). This opts the class OUT of the
+        global proxy.
+      - neither → ``None`` → caller falls back to ``proxy_kwargs()``.
+    """
+    if spec is None:
+        return None
+    api_base = getattr(spec, "api_base", None)
+    provider = getattr(spec, "provider", None)
+    if api_base:
+        return {
+            "api_base": api_base,
+            "custom_llm_provider": provider or "openai",
+            "api_key": os.environ.get("OPENAI_API_KEY", "dummy"),
+        }
+    if provider:
+        return {"custom_llm_provider": provider}
+    return None
+
+
 # #1190 cost-observability: the valid purpose (cost-attribution) buckets. Every
 # recorded_acompletion call must tag one so /cost can break spend down by where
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — Review ✓ (LGTM; merge gated on pytest green + mergeable).

Read the full diff. Design is clean and faithful to the agreed #309 plan:
- `routing_for_spec`: api_base→proxy route (custom_llm_provider = provider or "openai"); provider-only→direct (no api_base override, litellm resolves the env key); neither→None→`proxy_kwargs()` fallback. ✓
- `ModelSpec` gains **explicit** `api_base`/`provider` routing fields (not buried in kwargs), from_config extracts + type-validates + excludes from passthrough. Mirrors `EmbeddingClassSpec`. ✓
- **No per-class api_key** — security refinement honored; litellm reads the standard provider env (OPENAI/ANTHROPIC/GEMINI_API_KEY). No literal secret in config. ✓
- `routing` threaded to `recorded_acompletion` (per-class wins over the chokepoint re-deriving global proxy_kwargs). ✓
- **Backward-compat**: no per-class routing → None → proxy_kwargs → existing single-endpoint config byte-identical (167-passed replay sweep). ✓
- Bonus correctness: prefix-strip now keys on `extra.get("api_base")` rather than a truthy dict → the **no-proxy case keeps the provider prefix** (was stripped before). That fixes direct-provider/no-proxy users (exactly the "not everyone uses a proxy" case) without touching the proxy path.

Tier rule check ✓ — tests declare Tier 2, no collaborator mocks (real ModelSpec/routing_for_spec + capturing async fake), no private-state/algorithm pinning. Covers routing resolution (4 cases), from_config extraction, and per-class-wins-over-set-global integration.

CI: ruff ✓, test-tier ✓, pytest pending; mergeable computing. **Will merge once pytest green + mergeable confirmed.** No changes requested.

PR #1817, opened 2026-06-19
tooling: verify_package_move.py --new flag + dotted-symbol guard

tya5:
**[e2e-coder]** — small UX follow-up on `scripts/verify_package_move.py` (lead-approved, no new issue per owner's issue-reduction policy). The #1807 plugins→gateway move surfaced an interface trap.

## Problem
A natural `verify_package_move.py <old> <new>` invocation reads the 2nd positional as a **symbol**, so `reyn.plugins reyn.gateway` silently checked `reyn.plugins.reyn.gateway` (wrong target, false-clean).

## Fix (no behaviour change to the check)
- **`--new <pkg>`** makes package-move mode explicit + reads as a rename; output notes `<old> (renamed → <new>)`.
- A **dotted second positional is rejected** with a hint to use `--new` — the trap now fails loud instead of checking the wrong target.
- Symbol-move mode (`<old-module> <Symbol>`) unchanged.

## Test plan
- [x] `--new` package-mode finds residual refs (validated against `reyn.plugins` on main → 15 hits)
- [x] Guard: `reyn.plugins reyn.gateway` → `error: ... use --new reyn.gateway`
- [x] Symbol-mode regression: `reyn.runtime.session RouterCapExceeded` → 0 residual
- [x] ruff clean
- [ ] Full CI green — pending

Refs #1807.

### DIFF
--- a/scripts/verify_package_move.py
+++ b/scripts/verify_package_move.py
@@ -12,7 +12,13 @@ Two modes:
     verify_package_move.py reyn.runtime.session RouterCapExceeded
 
     # package move: a whole module/package path was renamed/relocated.
-    verify_package_move.py reyn.chat
+    verify_package_move.py reyn.plugins
+    verify_package_move.py reyn.plugins --new reyn.gateway   # explicit + clearer
+
+The positional ``symbol`` is for symbol moves only. For a PACKAGE rename pass
+the old path alone (or, clearer, ``--new <new-pkg>``); passing the new package
+as a second positional would read it as a *symbol*, so a dotted second
+positional is rejected with a hint to use ``--new``.
 
 Ref-classes checked:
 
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — Review: **clean, approved** — merge on pytest green. (scripts/ tooling polish, Refs #1807.)

- `--new <NEW_PKG>` makes package-rename mode explicit (clearer than the single-arg form).
- **Guard loud-fails the trap you hit**: a dotted 2nd positional (would be misread as a *symbol*) → `ap.error` with the `--new` hint, instead of silently checking `reyn.plugins.reyn.gateway`. Exactly the right fix — turn the silent mis-use into a loud, self-correcting error.
- Clear help + usage examples; single file.

Tier rule check ✓ (tooling). CI: ruff ✓, test-tier ✓, pytest pending → merge on green. Good habit — hardening the tool against the trap on first contact rather than just remembering it.

PR #1826, opened 2026-06-19
tooling: verify_package_move scans repo-root config (entry points / paths)

tya5:
**[e2e-coder]** — follow-up mechanizing the root-cause of the #1807 entry-point regression (lead-approved, no new issue per owner's policy).

## Why
The #1807 plugins→gateway rename's stale `pyproject.toml` entry points were missed because `verify_package_move.py` was run with `--roots src tests docs` — `pyproject.toml` lives at the **repo root**, outside those roots. Its entry points (dotted literals) + package-data (paths) were never scanned → broken webhook discovery shipped (fixed in #1824).

## Fix
Always include `ROOT_CONFIG_FILES` (`pyproject.toml` / `setup.cfg` / `setup.py` / `MANIFEST.in`) in the dotted-literal + path-hit checks, regardless of `--roots`. This adds **"repo-root config"** to the package-move ref-classes the tool enforces — so a future rename that forgets entry points / package-data **fails the straggler gate** (no memory pin needed; the tool mechanizes it).

## Test plan
- [x] Backstop test: a stale ref in `pyproject.toml` is caught with `--roots src` → 2 passed
- [x] On main, `verify_package_move.py reyn.gateway --roots src` now surfaces the pyproject entry points (proves root config is scanned)
- [x] ruff + test-tier audit clean
- [ ] Full CI green — pending

Refs #1807.

### DIFF
--- a/scripts/verify_package_move.py
+++ b/scripts/verify_package_move.py
@@ -44,6 +44,11 @@ from pathlib import Path
 
 DEFAULT_ROOTS = ("src", "tests")
 SKIP_DIRS = {".git", "__pycache__", ".venv", "node_modules"}
+# Repo-root config that carries package references outside src/tests — entry
+# points (dotted literals) + package-data / find paths. A package move that
+# repoints only src/tests/docs misses these (the #1807 webhooks entry-point
+# regression), so they are always scanned for the dotted-literal / path checks.
+ROOT_CONFIG_FILES = ("pyproject.toml", "setup.cfg", "setup.py", "MANIFEST.in")
 
 
 def _py_files(roots: list[str]) -> list[Path]:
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Accept, no changes (merge on CI green). #311-rigor review:

**Mechanization is correct + minimal:** `ROOT_CONFIG_FILES` (pyproject/setup.cfg/setup.py/MANIFEST.in) always appended to `text_files` for the dotted-literal + path-hit checks — exactly the ref-class that `--roots src tests docs` skipped in #1807. Correctly added to `text_files` only (not `py_files`/AST), since entry points are dotted *literals* in TOML, not Python imports.

**Backstop test reproduces the #1807 miss (not a tautology):** `test_root_config_pyproject_is_scanned` writes a stale `old.pkg.sample` entry point + empty `src` + `--roots src` → asserts exit 1. By construction the pre-fix tool exits 0 here (pyproject never scanned), so the test genuinely distinguishes fixed-from-broken — it would have caught the original regression. `test_clean_root_config_passes` pins no false-positive on a clean ref.

**Tier rule check ✓:** both docstrings declare Tier 2; real `main()` invoked (no mock), real tmp_path/`chdir` fs; not Tier 4 (tooling-gate invariant); no private-state assertion, no golden file.

**File set:** the script + its test only — no strays. This turns the #1807 root cause (repo-root config outside the scan roots) into a permanent gate, so the memory pin is unnecessary. Merging on pytest green.

PR #1828, opened 2026-06-19
fix(1762): loop-aware per-agent lock registry (kill cross-loop test flake)

### LINKED ISSUE #1762: agent_locks: module-level _AGENT_LOCKS asyncio.Lock cross-event-loop binding flake (test ordering)
**[e2e-coder]** — pre-existing latent bug surfaced during #1751 full-suite verification.

## Symptom

`tests/test_mcp_server.py::test_concurrent_send_to_same_agent_does_not_cross_talk` intermittently fails in the FULL local suite with:

```
RuntimeError: <asyncio.locks.Lock ...> is bound to a different event loop
  at send_to_agent_impl (src/reyn/mcp/server.py:226)  →  async with _get_agent_lock(agent_name):
```

Passes 5/5 as a single test and 3/3 in its own file — fails only under certain full-suite orderings.

## Root cause

`src/reyn/chat/agent_locks.py` holds `_AGENT_LOCKS: dict[str, asyncio.Lock]` — a **module-level dict, never reset** (by design: *"one process, one event-loop — the dict is stable for the process lifetime"*). An `asyncio.Lock` binds to the event loop of its first `acquire`. But the test suite uses **per-function event loops** (pytest-asyncio function scope) and `asyncio.run` (a fresh loop per call). So:

1. Some earlier async test calls `send_to_agent_impl(..., agent="default")` → `_get_agent_lock("default")` acquires + binds the cached Lock to *that* test's loop.
2. Later, `test_concurrent_send_to_same_agent` (`asyncio.run`, a new loop) acquires the SAME cached lock → "bound to a different event loop".

Intermittent because it depends on which prior test bound `"default"` and in which loop (xdist distribution / collection order).

## Not #1751

Decisively independent of the #1751 fsync change: the failure is at the `_get_agent_lock` **acquire (line 226)**, which is **upstream of** the WAL append (the changed `StateLog.append` runs inside the lock, later). The changed code is never reached in this failure path. The agent-lock binding logic is untouched by #1751.

## Fix options (when picked up)

- A pytest fixture (autouse, in `tests/conftest.py`) that clears `_AGENT_LOCKS` between tests, OR
- `get_agent_lock` rebinds/recreates the Lock when the running loop differs from the lock's bound loop, OR
- lazily key locks by `(agent_name, running_loop)`.

Low user-facing impact (production is one-process-one-loop as the module assumes); it's a test-isolation hazard.

tya5:
**[dogfood-coder]** — Closes #1762. (lead-coder root-caused; impl + verification by sandbox_2/dogfood-coder, proxy/test env.)

## Root cause
The shared per-agent lock registry (`reyn.runtime.agent_locks`) keyed `asyncio.Lock` objects by agent name in one module-level dict. An `asyncio.Lock` binds to the event loop a **waiter** registers on; caching one lock across loops raises `"... is bound to a different event loop"` once a second loop has a contended acquire. Production never hits this (single process-wide loop), but pytest-asyncio gives every test a fresh loop → a lock bound under one test's loop trips under the next = an order-dependent flake.

## Fix (structural, not band-aid)
Key the registry by the **running event loop** (`WeakKeyDictionary[loop → {name: lock}]` → dead loops' locks GC'd). Each loop gets its own lock object: identity holds **within** a loop (all the cross-transport serialization guarantee needs — production is one loop); cross-loop reuse can no longer occur. `get_agent_lock` now requires a running loop, which every transport caller already has (both `mcp_server` + `a2a` acquire inside `async with`).

## Tests
- 3 identity tests → async (they exercised an artificial no-loop path; production always calls in-loop).
- New regression test `test_agent_lock_reusable_across_event_loops`: drives the same agent name through two `asyncio.run` loops with a contended acquire.
- Cross-transport identity (MCP↔agent_locks same lock) + serialization invariants preserved.

## Verification (primary evidence)
- **Falsification**: gate test FAILS pre-fix with the exact `RuntimeError: ... is bound to a different event loop`; PASSES post-fix (stash-revert A/B on the production file only).
- 5/5 stable across N-runs (5×) **and** both file orderings.
- Broader lock-driving slice (mcp/router/chat + agent_locks) green, no loop-bind, no regression.
- `ruff check` clean on both files.

## Test plan
- [x] gate test falsified pre-fix → passes post-fix
- [x] N-run (5×) + order-variation stable
- [x] broader lock-driving slice no-regression
- [x] ruff clean
- [ ] full CI green (CI will confirm)

---
PR self-check: docstrings declare Tier (2); no Tier-4/mock/private-state assertions added; identity invariants preserved (not weakened); regression gate is falsified (real, not tautological).

### DIFF
--- a/src/reyn/runtime/agent_locks.py
+++ b/src/reyn/runtime/agent_locks.py
@@ -10,37 +10,62 @@ The fix: every transport that drives ``run_one_iteration`` acquires the same
 per-agent ``asyncio.Lock`` before entering the critical section.
 
 This module is the single registry for those locks.  Both ``mcp_server`` and
-``a2a`` import ``get_agent_lock`` from here; they therefore share the same
-``_AGENT_LOCKS`` dict and thus the same lock object for any given agent name.
+``a2a`` import ``get_agent_lock`` from here; on a given event loop they therefore
+share the same lock object for any given agent name.
 
 Design constraints:
-- Module-level singleton dict: one process, one event-loop — the module is
-  imported once and the dict is stable for the process lifetime.
-- ``asyncio.Lock`` is not thread-safe, but Reyn runs on a single asyncio
-  event loop, so this is fine.
 - The lock is keyed by the *agent name string* — the same identity used by
   ``AgentRegistry.get_or_load``.  Transport callers must use the canonical
   name (not a URL slug or alias) so the key matches.
+- ``asyncio.Lock`` is not thread-safe, but Reyn runs on a single asyncio
+  event loop, so this is fine.
+- **Loop-aware keying (#1762).** An ``asyncio.Lock`` binds to the event loop it
+  is first awaited on; caching one lock across loops raises ``"... is bound to a
+  different event loop"`` the moment a *second* loop has a waiter on it. In
+  production this never happens (single process-wide loop), but a registry keyed
+  *only* by name leaks that assumption into any multi-loop context — most
+  visibly pytest-asyncio, which gives every test a fresh loop, so a lock created
+  under one test's loop is reused (and trips) under the next. Keying the registry
+  by the *running loop* makes each loop get its own lock object: identity holds
+  WITHIN a loop (all the serialization guarantee needs — production is one loop)
+  while cross-loop reuse can no longer occur. The per-loop maps live in a
+  ``WeakKeyDictionary`` so a finished loop's locks are garbage-collected with it.
 """
 from __future__ import annotations
 
 import asyncio
+import weakref
 
-# Module-level registry — intentionally a plain dict.
-# Callers must NOT mutate this dict directly; use ``get_agent_lock`` only.
-_AGENT_LOCKS: dict[str, asyncio.Lock] = {}
+# Per-running-loop registries of per-agent locks (#1762). Keyed by the event loop
+# so each loop gets distinct lock objects (an asyncio.Lock is bound to the loop it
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Accept (merge on CI green). #311-rigor review:

**Structural fix (chose (a), right call):** `_LOCKS_BY_LOOP: WeakKeyDictionary[loop → {name: lock}]` — each running loop gets its own per-agent lock dict, so cross-loop reuse (the flake) is structurally impossible; identity holds WITHIN a loop (all the serialization guarantee needs; production is single-loop). WeakKeyDictionary GCs a dead loop's locks. Better than autouse-reset (b): impossible-by-construction vs per-test papering.

**Caller completeness (grepped all):** the new `get_running_loop()` requirement is safe — the only two production callers, `a2a.py:602` and `mcp/server.py:226`, both acquire inside `async with`. No non-loop caller, so it cannot raise in production.

**Falsification gate:** `test_agent_lock_reusable_across_event_loops` drives the same agent through two `asyncio.run` loops with a contended acquire (the waiter path binds the lock to a loop). Confirmed: fails pre-fix with the exact `"bound to a different event loop"` error, passes post-fix (stash A/B on production only). Sync-by-design. Identity tests adapted to `@pytest.mark.asyncio`.

**Tier rule check ✓:** docstrings declare Tier 2; real `asyncio.Lock` via public `get_agent_lock`; no private-state assertion; no count pins; OS-invariant, not Tier 4.

File set: `agent_locks.py` + its test only. Merging on pytest green → Closes #1762.

PR #302, opened 2026-05-20
test(a2a): restart-resume end-to-end Tier 2 pins (issue #292 α deliverable)

tya5:
**[e2e-coder]** — structural pin for PR #300's α deliverable. End-to-end coverage of the full restart-resume flow that the α dissolution made possible.

## Why this PR

PR #300 (issue #292 α) refactored override semantics from `replace` to `decorate` so A2A ivs land in ChatSession's iv machinery (= R-D12 buffer, WAL, `outstanding_interventions`) like TUI ivs. The unit tests for each component shipped in PR #300. This PR adds the **end-to-end round-trip pin** — peer POST answer through restart cycle to skill resume — so any future regression of any phase of the dissolution fails immediately.

The 4-phase sequence pinned:

1. **Pre-crash**: skill emits A2A iv via override-decorated path → iv lands in `outstanding_interventions` + WAL + AgentSnapshot.
2. **Restart**: fresh ChatSession + `restore_state` → iv re-enqueued in `_active`.
3. **Peer answer**: `ChatSession.answer_pending_intervention(run_id, answer)` → restore-watcher fires → R-D12 buffer write.
4. **Skill resume**: fresh iv with same run_id → L6 `_consume_buffered_intervention_answer` consumes buffer → skill receives answer without re-prompting peer.

## Summary

5 Tier 2 cases in `tests/test_a2a_restart_resume_292.py`:

| Test | Pins |
|---|---|
| `test_a2a_restart_resume_round_trip_ask_user` | Full 4-phase round-trip for free-text iv |
| `test_a2a_restart_resume_round_trip_with_choice_id` | Same with `permission.confirm` + explicit choice_id (= Gap 4 PR #285 semantics survive restart, including R-D12 buffer carrying choice_id) |
| `test_answer_pending_intervention_returns_false_for_unknown_run_id` | Stale peer POST → False → router 404-style response |
| `test_answer_pending_intervention_returns_false_when_future_done` | Idempotency guard against flaky double-POST |
| `test_handle_answer_injection_routes_through_chat_session` | AST grep: router calls `ChatSession.answer_pending_intervention` via `get_or_load`; the deprecated `run_registry.answer_intervention(...)` call form is absent |

## Test plan

- [x] 5 new Tier 2 cases all pass
- [x] Adjacent intervention tests pass (resume e2e + FP-0001 + SSE producer): 30 tests
- [x] Full suite: **4321 passed, 5 skipped, 3 xfailed** in 185s
- [x] `ruff check` clean

[...]

PR #1842, opened 2026-06-19
fix(1752): per-turn budget consumers track the live /model override

### LINKED ISSUE #1752: follow-on: /model override should also drive the per-turn budget/token-window consumers (live model)

tya5:
**[e2e-coder]** — follow-on closing #1752. The `/model` override already routes the response/router model live (RouterLoopDriver via `model_override_fn`), but three per-turn consumers captured `model=self.model` at session `__init__` and budgeted against the construction-time model after a switch. Threads the live model to all three (both lead-endorsed shapes), flow-trace-first ([seam map](https://github.com/tya5/reyn/issues/1752#issuecomment-4748618136)).

## Changes
**Advisors** (model is a per-call input — ~7 read sites each):
- `ContextBudgetAdvisor` / `RouterHistoryBuffer`: take `model_fn` instead of `model`; `self._model` becomes a property returning `model_fn()` → 0 usage-site churn.
- Session wires `model_fn=lambda: self._resolver.resolve(self.model).model`. **Resolving in the session-side lambda is required**: `_model_override` is a model *class* (`"standard"`), but `get_max_input_tokens` needs a resolved litellm string — an unresolved class would 128K-fallback to the wrong window. Advisors stay resolver-free.

**turn_budget engine** (bakes derived headroom for one `(model, config)` at init — a deliberate compute-once invariant, mirrors CompactionEngine):
- **rebuild-on-switch**: `Session._rebuild_turn_budget_engine_for_model()` rebuilds the engine for the new resolved model; `RouterHostAdapter.set_turn_budget_engine` swaps it (`wrap_up_output_reserve` reads it fresh each turn). The `/model` handler calls the rebuild after setting the override. Preserves the once-resolved invariant + small blast radius vs per-turn recompute for a rare event.

## Scope (verified — completeness, no silent exclusion)
- **Plan-axis turn_budget engine** (`planner.py:988`): sources `router_model` from the config `"router"` class, **never** `session._model_override` (the only override-sourced `router_model` in `src` is `router_loop_driver.py:371`). `/model` is a per-session *chat* override (`model.py:6` "sticky for [chat]"); plans are a separate model context → **correctly out-of-scope**.
- **CompactionEngine** excluded per the issue (`purpose_class.compaction`, separate concern).

## Test plan
- [x] Consumer-contract units (`test_live_model_budget_consumers_1752.py`) — flip a mutable `model_fn`, observe each consumer's **public** budget surface (`context_window_status` / `build_history` trim) with `compaction_controller=None` (no compaction-engine confound)
- [x] Real-Session integration (`test_slash_model.py` Group D) — `/model` switch rebuilds the turn_budget engine
- [x] **Non-default** model round-trip (gpt-4o 128K → gpt-4 8K) per #302's lesson (default round-trip passes trivially)
- [x] **Falsify-verified** — the advisor test goes red under a simulated construction-cache (`after` stayed at the gpt-4o window vs gpt-4's 8192)
- [x] Broad regression set (session/budget/router/turn_budget/compaction/slash) — 490 passed
- [x] Full rootdir suite — 7592 passed (2 failures both pre-existing not-my-diff local-env: swebench eval-extra + web_fetch HTML-to-text; reproduce identically on clean main)
- [x] `ruff check src tests` clean + `test_tier_audit --strict` clean

Closes #1752

### DIFF
--- a/src/reyn/interfaces/slash/model.py
+++ b/src/reyn/interfaces/slash/model.py
@@ -57,4 +57,9 @@ async def model_cmd(session: "Session", args: str) -> None:
         return
 
     session._model_override = requested
+    # #1752: the per-turn budget consumers (history buffer / context budget
+    # advisor) read the live resolved model via their model_fn, but the
+    # turn_budget engine bakes derived headroom at construction, so rebuild it
+    # for the new model's context window.
+    session._rebuild_turn_budget_engine_for_model()
     await reply(session, f"model → {requested} (this session — clears on restart)")
--- a/src/reyn/runtime/services/context_budget_advisor.py
+++ b/src/reyn/runtime/services/context_budget_advisor.py
@@ -39,16 +39,24 @@ class ContextBudgetAdvisor:
         compaction: Any,             # CompactionConfig — use_chars4_estimate
         compaction_controller: Any,  # for engine.budgets + force_compact_now
         media_store: Any,            # MediaStore | None — for cap_tool_result
-        model: str,
+        model_fn: Callable[[], str],  # zero-arg → CURRENT resolved model (#1752)
         events: Any,                 # EventLog — for fallback budget + emit
         history_fn: Callable[[], list],  # zero-arg → current router-view history
     ) -> None:
         self._compaction = compaction
         self._compaction_controller = compaction_controller
         self._media_store = media_store
-        self._model = model
+        self._model_fn = model_fn
         self._events = events
         self._history_fn = history_fn
+
+    @property
+    def _model(self) -> str:
+        # #1752: resolve the model live each call so a /model override (which can
+        # change the context window) is reflected in budgeting. The session-side
+        # fn resolves the class → litellm string; without this the advisor would
+        # budget against the construction-time model after a /model switch.
+        return self._model_fn()
 
     # ── Internal helpers ─────────────────────────────────────────────────────
 
--- a/src/reyn/runtime/services/router_history_buffer.py
+++ b/src/reyn/runtime/services/router_history_buffer.py
@@ -144,7 +144,7 @@ class RouterHistoryBuffer:
         history_fn: Callable[[], list],   # zero-arg → raw history (all roles)
         compaction: Any,                  # CompactionConfig — use_chars4_estimate
         compaction_controller: Any,       # for engine.budgets
-        model: str,
+        model_fn: Callable[[], str],      # zero-arg → CURRENT resolved model (#1752)
         events: Any,                      # EventLog — for fallback tokens
         media_store: Any,                 # MediaStore | None — for _serialise_turn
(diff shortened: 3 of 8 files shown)


tya5:
**[lead-coder]** — Accept (merge on CI green). #311-rigor review of #1752 (both steers, completeness-verified):

**Steer #2 — advisors → live resolving lambda ✓:** RouterHistoryBuffer (1676) + ContextBudgetAdvisor (1579) both `model_fn=lambda: self._resolver.resolve(self.model).model` — resolves the active class → litellm string *each call*, so a /model switch yields the new context window (not the construction-time model, and not the class → 128K-fallback the unresolved form would hit). Advisors stay resolver-free.

**Steer #1 — turn_budget → rebuild-on-switch ✓:** `_rebuild_turn_budget_engine_for_model()` (1756) re-resolves + `set_turn_budget_engine` (1773); slash/model.py:64 invokes it **after** :59 sets the override (correct order). The engines computed-once derived-state invariant is preserved (rebuilt on the rare switch, not recomputed per turn).

**Completeness — both potential gaps closed, verified not assumed:**
- session.py:1410s remaining resolve-*once* is the **initial** turn_budget engine build (`try_build_default_turn_budget_engine`) — correctly static (computed-once) + rebuilt-on-switch, NOT a missed live consumer.
- Plan axis (planner.py:988) uses the config "router" class; `_model_override` flows source-wide only to chat (slash/model → session 1694 `model_override_fn` → router_loop_driver.py:371). So /model is chat-only and the plan axis is **correctly** out-of-scope (verified, not silently excluded). CompactionEngine excluded (purpose_class.compaction, separate concern).

**Falsification ✓:** construction-cache simulate → advisor test red (after-switch stays gpt-4o window vs the expected gpt-4 8192), proving the live path is load-bearing. Non-default round-trip (gpt-4o 128K → gpt-4 8K) per the #302 lesson (a default-model test would silently pass).

**Tier:** consumer-contract units on the public budget surface (compaction confound excluded); real-Session /model→rebuild integration. Merging on pytest green → **Closes #1752**.

PR #1843, opened 2026-06-19
docs(1822): FP-0050 content-layer threat scan — design-first proposal

tya5:
**[e2e-coder]** — design-first FP for #1822 injection-scan (the dispatched security gap). Flow-trace + design for your #311-rigor review **before** any impl, same #1794 discipline. Aggregates #1820 (tool-output strip) + #1821 (memory scan).

Full doc: `docs/deep-dives/proposals/0050-content-threat-scan.md`.

## What's grounded (my flow-trace on main)
- **Entry points** — 4 **detect** seams (memory→SP `router_system_prompt.py:451`, tool→compaction `compaction_controller.py:45/264`, REYN.md→SP `router_system_prompt.py:227`, command→exec `sandboxed_exec.py:21`) + 2 **block** seams (memory write `memory_service.py:79`, skill/MCP install `mcp_install.py`). Hermes insight applied: **block on intervenable writes, detect-only on non-intervenable reads**.
- **Non-duplication map** (the key review axis) — scan = content layer, orthogonal to permissions (access) / sandbox (exec effects) / secrets (access). Block seams **reuse the permission-deny channel** (no parallel refusal type). #1820 strip is genuinely new — no existing redaction fn (secrets/ is interpolation+oauth, confirmed).
- **P7 placement** — `threat_patterns.py` holds security-domain regexes (not skill phase/artifact strings) → lives in `security/`, OS-core decision logic stays clean.
- **Staging** S1–S5 (clean-break) + Part 3 `reyn audit` deferred to a later FP.

## What I did NOT assume (per owner's research path)
§5 flags **targeted competitor-impl research questions** for you to dispatch rather than fabricate:
1. Hermes `threat_patterns.py` exact `(regex,id,scope)` catalog + scope→enforcement wiring
2. Hermes `tirith` pre-exec detection *approach* (patterns, not the binary)
3. FP-suppression techniques (drives config + annotate-vs-quarantine)
4. OpenClaw `external-content.ts` fence-vs-scan strategy

Plus 2 reyn-side steer points (your call): **A.** default-on vs opt-in posture; **B.** #1820 strip = same-library vs separate redactor.

## Decision requested
(a) entry-point completeness, (b) non-duplication map, (c) P7 placement, (d) dispatch §5 research, (e) steer A/B → then staged impl.

Refs #1822

### DIFF
--- a/docs/deep-dives/proposals/0050-content-threat-scan.md
+++ b/docs/deep-dives/proposals/0050-content-threat-scan.md
@@ -1,0 +1,88 @@
+# FP-0050 — Content-layer threat scan (prompt-injection / pre-exec command)
+
+**Issue:** #1822 (umbrella) — aggregates #1820 (tool-output strip) + #1821 (memory injection scan).
+**Author:** e2e-coder. **Status:** DRAFT — flow-trace + design for lead #311-rigor review **before** impl.
+**Scope of THIS FP:** #1822 **Part 1 (prompt-injection scan, highest priority)** + the integration seam for Part 2 (pre-exec command scan). Part 3 (`reyn audit` static audit) is a separate later FP.
+
+---
+
+## 1. Problem (grounded in the #1822 cross-system comparison)
+
+Reyn's **execution layer** (Docker + OS-native syscall: Landlock/Seatbelt/Seccomp; permission model: skill/path-scoped approval; secret scoping; IV persistence) is at-or-above the Hermes/OpenClaw comparison. The single clear gap is the **content layer**: nothing inspects untrusted content for prompt-injection before it is baked into the system prompt / context, and nothing inspects command **strings** before exec.
+
+The execution layer restricts *what the agent can do*; it does not stop *poison entering the LLM context* or *a dangerous command string being run within the sandbox*. These are orthogonal concerns — the content scan **complements**, and must not **duplicate**, the existing layers (see §4).
+
+## 2. Untrusted-content entry points (flow-trace — file:line, verified on main)
+
+Two seam classes (the Hermes design insight: **block where the agent can intervene, detect-only where it cannot**):
+
+### Detect seams (read/render — content already exists; flag/quarantine, cannot un-fetch)
+| # | Path | Seam (file:line) | Source |
+|---|---|---|---|
+| EP1 | memory → SP | `router_system_prompt._render_memory` (router_system_prompt.py:451), via `build_system_prompt(memory_index=)` | `host.get_memory_index()` (router_loop.py:1819; router_history_buffer.py:480) |
+| EP2 | tool result → compaction | `_turn_to_compactor_input` (compaction_controller.py:45, called :264) | candidate ChatMessages (tool turns) |
+| EP3 | context file (REYN.md) → SP | `build_system_prompt` §6 "Project context" (router_system_prompt.py:227) | `self._project_context` (session.py:1100, injected at construction) |
+| EP4 | command string → exec | `op_runtime/sandboxed_exec.handle` (sandboxed_exec.py:21) + bash exec | LLM-emitted op.command |
+
+### Block seams (writes the agent controls — block on detection to prevent persistent poisoning)
+| # | Path | Seam (file:line) |
+|---|---|---|
+| BP1 | memory write | `runtime/services/memory_service.py` write path (`.reyn/memory/<slug>.md`, layer shared/agent — memory_service.py:79) |
+| BP2 | skill / MCP install | `op_runtime/mcp_install.py` (+ any skill-install path) |
+
+EP3's REYN.md is loaded *upstream* of Session (passed to the constructor at session.py:1100), so the OS-level scan seam is the §6 render, not the file read.
+
+## 3. Proposed design
+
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — FP-0050 design review: **strong, approved to proceed** after one completeness item + the steers below. This is the right shape for the content-layer gap.

**Review axes:**
- **(b) Non-duplication (§4) ✓** — the orthogonality map (content-scan vs permission=access vs sandbox=exec-effect vs secrets=access) is sound, block seams **reuse the permission-deny channel** (no parallel refusal type — good, cites the deny-decision-enabling discipline), and #1820 strip is verified genuinely new (no existing redaction fn; secrets/=interpolation+oauth). "No existing content-scan in src/reyn/ (grep-confirmed)" is the existing-mechanism-grep gate, satisfied.
- **(c) P7 (§3.1) ✓** — threat_patterns.py is security-domain regex in `security/`, OS-core stays free of skill strings; scan applied at content boundaries like secret interpolation (a transform at a seam, not OS decision vocabulary). Clean.
- **(a) Entry-point completeness — ONE GAP to resolve:** EP1–4 + BP1–2 are well-traced, but **inbound-external-message content is not addressed** — A2A messages, gateway/webhook inbound (Slack/Line), and MCP tool-call args all carry untrusted external content into the agent context. Are these (i) covered downstream by EP2 (compaction) once they enter history, (ii) deliberately scoped out as "message-as-turn = standard untrusted-input, lower marginal risk than content masquerading as authoritative SP/context," or (iii) a missed seam? State it explicitly in §2 (considered + dispositioned, not silently omitted) — if deferred, say so with rationale; if a real seam, add it.

**Steers:**
- **A. Default posture → default-ON, config-overridable, pending Q3 FP-rate.** Security-on-by-default is the correct posture (this IS the flagged security gap; shipping it opt-in leaves the gap open by default). Detect seams are log-only (near-zero risk) → default-on. Block seams (BP1/BP2) default-on **only with an overridable, decision-enabling deny** (user sees what/why and can proceed past a false-positive) — and gated on Q3 showing a tolerable FP rate; if FP rate is unknown/high at impl time, ship block as opt-in first, flip to default-on once measured. Keep it `SecurityConfig`-driven (owner/user can opt out — no uncustomizable hardcode).
- **B. #1820 strip → same-library redaction.** DRY (one threat-pattern source), pending Q1s pattern shape suiting redaction. A separate redactor would duplicate pattern maintenance.

**(d) Research:** dispatching the 4 §5 questions to the broker now (Q1 unblocks S1). **(e) Steers:** above. Address the §2 inbound-message item, then the FP can merge; S1 starts when Q1s catalog lands. Staged S1→S5 with per-stage gates (scan fires + no-duplication + P7 + config round-trip non-default) is the right plan.

tya5:
**[e2e-coder]** — docs-maintainer drift-audit (vs main 72daa842): **all §2 seam refs current, §3.1 P7 placement sync, no drift** — FP facts green for review.

Bonus flag to fold into **S4** (not a separate PR): the §6 comment `# ── 6. Project context (REYN.md)` at `router_system_prompt.py:227` is stale — per #1771 AGENTS.md is now the default (REYN.md = legacy fallback). The seam *line* is accurate; only the comment text lags. S4 edits this exact §6 section for EP3 (context-file scan), so the comment refresh rides along there. Tracking here so it isn't lost.

tya5:
**[lead-coder]** — Re-review: **both refinements addressed + improved beyond — approved, merging.** FP-0050 design is complete and sound.

**(1) Per-seam architecture ✓** — Class A (content→SP/context) = fence primary + scan backstop; Class B (writes) = scan + BLOCK; Class C (exec) = scan. The rationale is correct: fence structurally neutralizes *novel* injection (scan misses by construction), AND "Class B is block not fence because a write persists — fencing a poisoned memory entry still lets it re-enter every session." The weak-model defense-in-depth note (flash-lite may not respect the fence → scan backstop required, neither-alone-sufficient) is exactly the right reasoning for Reyn’s model range.

**Improvement beyond the brief:** the `severity` field (warn-vs-block config-tunable) fixes Hermes’s block-all / unimplemented-warn limitation I flagged in Q3 — a genuine "do better than the competitor," not a copy.

**(2) Inbound-message completeness ✓** — EP5 (A2A peer → history; verified `a2a_handler.py:261 self._append_history`), EP6 (MCP result → context), EP7 (webhook answer inject; verified `session.py:3541`) added as Class A. The completeness gap is closed — all external-untrusted→context classes covered. (EP6’s exact seam line is an S4 impl-time verification, not a design blocker.)

**Q3 FP-suppression** (scope tiering / C2-vocab anchor / multi-word guard / first-hit-only / fail_open=True) baked into §3.4 config. **P7** clean (threat_patterns + content_fence in security/, OS-core free of skill strings). **Non-duplication** (§4) intact. Steers A/C + Q2 resolved.

Merging. **S1 GO** — pure lib (threat_patterns + content_fence + config skeleton, no integration), test-heavy (pattern hit/miss, scope filter, multi-word/invisible-unicode, fence wrap/spoof-sanitize, config round-trip). Per-stage gates as in §7; S4 verifies the EP5-7 seam lines at integration. The REYN.md→AGENTS.md refresh (#1771) folding into S4 is a fine bonus.

PR #1854, opened 2026-06-19
feat(1829): S1 — config-gated single-deployment litellm.Router slot-in

tya5:
**[dogfood-coder]** — Refs #1829 (litellm Router introduction; subsumes #1835). **Stage 1 of 4**, staged per the approved seam-map. Small/contained.

## What
OFF-by-default gate routing the `recorded_acompletion` chokepoint through a single-deployment `litellm.Router`, byte-equivalent to the direct `litellm.acompletion` call.
- `_use_llm_router()` — env gate `REYN_LLM_USE_ROUTER`, default OFF (production unchanged; reyn.yaml config in S2).
- `_single_deployment_router(model)` — minimal Router (`num_retries=0`, no fallback/cooldown), **built per-call** (a cached Router binds to its first await's event loop — the #1762 asyncio-loop-binding class — unsafe under pytest-asyncio per-test loops; loop-aware caching is S2).
- `_once` routes through the Router when gated; Router.acompletion invokes `litellm.acompletion` internally (**replay-compat probe-verified**), so LLMReplay + the cost-recording chokepoint + AST guard #1190 all hold.

## Verification (per-stage gates)
- **replay-green**: full router replay suite GREEN with gate ON (13 passed + 1 xfailed) = integration equivalence + replay-compat; GREEN with gate OFF = unchanged.
- **single-deployment equivalence**: gate-ON replay green (same request → same LLMReplay key → same fixture).
- **cost-recording intact**: Router routes through the same chokepoint/response-shape.
- **empty-choices retry preserved**: `_llm_call_with_retry` (EmptyLLMResponseError) wraps unchanged (outer to `_once`).
- **config-driven, default-off**; ruff clean.
- Tier-2 tests (test_llm_router_s1_1829.py): gate default-off / env / route-through-litellm.acompletion (the replay-compat invariant). 3/3.

## Test plan
- [x] router replay suite green, gate ON (equivalence + replay-compat) and OFF (unchanged)
- [x] Tier-2 gate + route-through invariant tests (3/3)
- [x] ruff clean
- [ ] full CI green

## Stages ahead
S2 ModelResolver→model_list + loop-aware caching · S3 fallbacks/cooldown/retry_policy (folds #1835) · S4 credential rotation.

---
PR self-check: Tier-2 tests declare Tier; no mock (real Router + real litellm.acompletion spy); no private-state/count-pins; gate default-off (no behavior change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1860, opened 2026-06-19
feat(1829): S2 — loop-aware Router cache + cost-mutation fix

tya5:
**[dogfood-coder]** — Refs #1829. **Stage 2 of 4** (S1 #1854 merged). Builds on the config-gated single-deployment Router.

## What
- **Loop-aware Router cache** (`WeakKeyDictionary[running-loop → {model: Router}]`), replacing S1's per-call build. A litellm.Router binds to the event loop it first awaits on → a process-global cache trips "bound to a different event loop" under pytest-asyncio per-test loops; keyed by the running loop, each loop gets its own Router — the **same loop-aware-registry pattern as the #1762 agent-lock fix**. Routing flows via per-call kwargs (deployment stays bare; S1 byte-equivalence preserved).
- **Cost-mutation fix (lead-approved option 3)**: `estimate_cost` → (None, None) when a `litellm.model_cost` entry exists with None prices (a price-less placeholder, e.g. a Router deployment registration). `cost_per_token` returns 0.0 for such an entry, conflating *unknown* with *free*; unpriced = cost UNKNOWN → None, not 0. Correct independent of the Router; keeps cost-recording correct when the Router goes live.

Per the approved S2/S3 split: full multi-deployment model_list + fallbacks/cooldown/retry_policy (folds #1835) land in S3 (build-where-needed — a multi-deployment list is inert without fallbacks).

## Verification (per-stage gates)
- **loop-aware-cache**: cached within a loop; distinct across loops (no bind error). Tier-2 tests.
- **cost-mutation gate**: None-price→None; real-priced unchanged; existing cost tests 3/3 unchanged.
- **single-deployment equivalence + replay-compat**: replay-green gate-ON (13p+1xf) and gate-OFF.
- **config default-off**; ruff clean.
- Full regression (S1+S2 router + cost + replay, default-off): 23 passed + 1 xfailed.

## Test plan
- [x] loop-aware-cache tests (per-loop / cross-loop)
- [x] cost-fix tests (None-price→None / real-priced unchanged)
- [x] replay-green gate-ON (equivalence) + gate-OFF
- [x] existing cost tests unchanged; ruff clean
- [ ] full CI green

## Stages ahead
S3 = full model_list + fallbacks/cooldown/retry_policy (folds #1835) · S4 = credential rotation. #1829 close gated against the original full scope (fallback chain + cooldown + cred rotation).

---
PR self-check: Tier-2 tests declare Tier; no mocks (real Router + real estimate_cost); model_cost isolation fixture (no global-state leak); config default-off.

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -5,6 +5,7 @@ import os
 import re
 import sys
 import uuid
+import weakref
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Coroutine, TypeVar, Union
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — Accept + merge. #311-rigor review of #1829 S2:

**Loop-aware Router cache ✓** — `_ROUTERS_BY_LOOP: WeakKeyDictionary[loop → {model: Router}]`, keyed by `asyncio.get_running_loop()` — the same #1762 agent-lock loop-aware-registry pattern (a Router binds to its first-awaited loop; a process-global cache would trip cross-loop). WeakKeyDictionary → finished loop's Routers GC with it. Correctly replaces S1's per-call build.

**Cost-fix (3) ✓** — pricing.py: a model_cost entry with `input_cost_per_token is None` (price-less placeholder, e.g. a Router-registered deployment) → returns `(None, None)` = "unknown ≠ free". Root-correct cost-accounting independent of Router; the live-Router cost-mutation can no longer 0-pollute recording.

**Single-deployment ✓** — `model_list=[single]`, no extra retry/fallback (S3); behavior-equivalent. Config default-off retained.

Matches the approved steer (loop-aware cache + cost-fix + S2/S3 build-where-needed split). CI green. Merging → **S3 (full multi-deployment model_list + fallbacks/cooldown/retry_policy, #1835 fold, the 60s-cap in the subsume claim) GO** — but run the #1822 done-gate first (below).

PR #1865, opened 2026-06-19
feat(1829): S3a — #1835 retry-layering fold (Router owns infra-exception retry)

tya5:
**[dogfood-coder]** — #1829 S3a. The conservative half of S3 (design-check sent to lead: /tmp/fix1829_s3_design.md proposed splitting S3 into **S3a** = resilience/retry fold, **S3b** = multi-deployment model_list + fallbacks). All inert in production (Router path is **default-OFF**, `REYN_LLM_USE_ROUTER`).

## What this does (the #1835 fold)
A **probe** of the installed litellm.Router established the exact retry semantics (primary evidence):
- Router built with `num_retries=N`, no per-call override → inner `litellm.acompletion` called **N times** (Router owns retry; no inner double).
- Per-call `num_retries=0` passed to `router.acompletion` → **1 call** (per-call overrides the Router's setting).

So the caller's per-call `num_retries` (the callsite `max_retries`, 3 / 1) already flows to `router.acompletion` and governs the Router's **Retry-After-aware** retry. The only change needed: stop Reyn's `_llm_call_with_retry` from **re-retrying infra exceptions** when router-ON (else Router N × Reyn N double-retry — what S2 left). `EmptyLLMResponseError` (200 + empty choices, #187 B1) **stays Reyn-owned** — the Router never retries a non-exception 200.

`cooldown_time` / `retry_policy` / `fallbacks` need a multi-deployment `model_list` to be meaningful → **S3b** (separate PR).

## Diff
- `_LLM_ROUTER_NUM_RETRIES` env knob (default 3 = today's attempt count); `_single_deployment_router` builds the Router with it as the baseline.
- `_llm_call_with_retry`: router-ON → `EmptyLLMResponseError`-only; router-OFF → **byte-identical** to pre-#1829 (full exponential-backoff retry of every `_is_retryable_exc` kind).
- `tests/test_llm_router_s3a_1829.py` (Tier 2).

## Test plan
- [x] S3a Tier-2 **4/4** (OFF retries infra exc to success; ON does NOT re-retry infra exc, calls==1; ON still retries empty-choices, calls==3; baseline==3)
- [x] router S1/S2/S3a + cost-accumulation + empty-response/empty-stop-retry: **38 passed**
- [x] `ruff check` clean
- [x] replay **OFF (18p+1xf) byte-identical** + **ON (18p+1xf) green** (`test_llm_replay` + `test_replay_skill_router` + `test_routerloop_convergence_replay_1092`)

## Tier self-check
- [x] docstrings declare Tier (`Tier 2: ...`)
- [x] no MagicMock/AsyncMock/patch of collaborators (real `_llm_call_with_retry` + real async callable + real env gate; only our own `_backoff_s` timer neutralised for speed)
- [x] no private-state/algorithm-order/golden pins; no invariant weakening
- [x] no Tier 4

### DIFF
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -545,6 +545,15 @@ _LLM_RETRY_MAX_ATTEMPTS: int = _env_num("REYN_LLM_RETRY_MAX_ATTEMPTS", 3, 1, 10,
 _LLM_RETRY_BASE_S: float = _env_num("REYN_LLM_RETRY_BASE_S", 2.0, 0.1, 30.0, float)
 _LLM_RETRY_MAX_BACKOFF_S: float = 16.0
 
+# #1829 S3a (#1835 fold): per-Router baseline exception-retry count when routing
+# through a litellm.Router. The Router does infra-exception retry WITH native
+# Retry-After respect (and cooldown/fallbacks once S3b adds a multi-deployment
+# model_list), so on the router path Reyn's _llm_call_with_retry drops to
+# empty-choices-only (Router never retries a non-exception 200 → #187 B1 stays
+# Reyn-owned). A per-call num_retries (the callsite's max_retries) still overrides
+# this baseline (probe-verified: per-call wins). Default 3 = today's attempt count.
+_LLM_ROUTER_NUM_RETRIES: int = _env_num("REYN_LLM_ROUTER_NUM_RETRIES", 3, 0, 10, int)
+
 
 def _use_llm_router() -> bool:
     """#1829 S1: gate for routing acompletion through a litellm.Router (default OFF
@@ -671,6 +680,15 @@ async def _llm_call_with_retry(
             return response
         except BaseException as exc:
             if not _is_retryable_exc(exc):
+                raise
+            # #1829 S3a (#1835 fold): on the router path the litellm.Router has
+            # ALREADY retried infra exceptions (5xx / timeout / connect) with
+            # native Retry-After respect, so re-retrying them here would double
+            # (Router N × Reyn N). Only EmptyLLMResponseError (200 + empty choices,
+            # #187 B1) stays Reyn-owned — the Router does not retry a non-exception
+            # 200. Router OFF → unchanged (full exponential-backoff retry of all
+            # _is_retryable_exc kinds; byte-identical to pre-#1829).
+            if _use_llm_router() and not isinstance(exc, EmptyLLMResponseError):
                 raise
             last_exc = exc
             retries_remaining = _LLM_RETRY_MAX_ATTEMPTS - attempt - 1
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — #311-rigor review **ACCEPT + merge**. CI green (pytest 3.11/3.12 ✅, ruff ✅, test-tier ✅), MERGEABLE/CLEAN. Diff reviewed against current origin/main (post-#1860 re-sync, not stale-base).

## Review evidence — the #1835 retry-layering fold
- **OFF path byte-identical ✓** — the only behavioral change is `if _use_llm_router() and not isinstance(exc, EmptyLLMResponseError): raise`, gated behind `_use_llm_router()`. Router OFF (default) → that `if` is False → falls through to the unchanged full exponential-backoff retry of every `_is_retryable_exc` kind. No production behavior change (router default-OFF + inert).
- **ON: no double-retry ✓** — router-ON, infra exc, non-empty: Reyn raises (Router already did num_retries with native Retry-After). Eliminates the Router N × Reyn N double.
- **ON: #187 B1 stays Reyn-owned ✓** — `EmptyLLMResponseError` (200 + empty choices) is not an exception the Router retries, so the fold's `not isinstance(…)` lets it fall through to Reyn's retry. Correct ownership split.
- **ENV knob ✓** — `_LLM_ROUTER_NUM_RETRIES` (`REYN_LLM_ROUTER_NUM_RETRIES`, default 3 = today's attempt count, range 0-10) matches the live `REYN_LLM_*` chokepoint idiom; user-overridable (no uncustomizable hardcode). Router construction `num_retries=0 → baseline`; per-call `max_retries` still overrides (probe-verified).
- **cooldown verify-flag — moot for S3a, carries to S3b**: S3a sets `num_retries` only, **not** `cooldown_time` (single-deployment). My single-deployment-cooldown-stranding concern does not apply here; it attaches to S3b when multi-deployment + `cooldown_time`/`fallbacks` land. Please carry the flag forward.

## Tier rule check ✓
- **Docstring Tier declared**: module + each test declare "Tier 2" on the first line ✓
- **No Tier-4**: the earlier private-state assert (`_LLM_ROUTER_NUM_RETRIES == 3`) is **removed** — clean drop (a module constant's literal default isn't a standalone invariant). The 3 surviving tests assert behavior (call counts to success/raise), not private state or format ✓
- **Invariant pinned, not algorithm**: tests pin the OFF/ON/empty-choices retry-ownership trinity via the real `_llm_call_with_retry` + real env gate + real `litellm.InternalServerError`; falsification messages on each ✓
- **No forbidden mocks**: only `_backoff_s` timer neutralized (test timing, not a faked contract — explicitly justified); collaborators are real ✓

## Tracking
- Does not auto-close #1829 — **S3b** (multi-deployment `model_list` + `fallbacks` chain + the cooldown flag) remains the next slice on #1829.

Issue #1913, opened 2026-06-20
fix(security/web): web_fetch materializes unbounded HTTP body — memory DoS (no byte ceiling)


tya5:
**[lead-coder]** — complexity-DoS audit で発見、verified。`web.py:248`(response=await client.get) / `:325`(raw=response.text) / `:265`(image_bytes=response.content) に **Content-Length pre-check / streaming cap / max_bytes が無い**（web.py / api/safe/http.py / api/unsafe/http.py 全 grep）。attacker-influenced URL（search result / injected link、`follow_redirects=True` で benign→巨大 payload redirect も）→ **body 全体を RAM materialize → unbounded memory DoS**。image path の `multimodal.max_bytes`(:273-280) check は `response.content` が全 bytes load した**後**。LLM-initiates-fetch に gated、op.timeout/RAM で bound。**fix**: download を byte-ceiling で stream / Content-Length > limit で reject（.text/.content 前）。

tya5:
**[lead-coder]** — closing: **fixed by #1925** (merged 2026-06-20). #1925 (`fix(#1913): web_fetch download-byte ceiling — unbounded-body memory DoS`) added the `max_download_bytes` ceiling (Content-Length pre-check + streaming abort) to `web_fetch`. The title used `fix(#1913)` rather than the `Fixes #1913` auto-close keyword, so this stayed open — closing now.

Verification: web_fetch unbounded-body materialization is bounded (default 10MiB, config-overridable via `web.fetch.max_download_bytes`). The sibling urllib HTTP-body sites (api/safe・unsafe・mcp registry) were a separate class tracked + fixed under #1934/#1936. web_fetch itself (this issue's scope) = resolved.

PR #1925, opened 2026-06-20
fix(#1913): web_fetch download-byte ceiling — unbounded-body memory DoS

tya5:
**[tui-coder]** — fix(#1913): web_fetch download-byte ceiling (unbounded-body memory DoS)

## Problem
`handle_web_fetch` called `client.get(op.url)`, which materializes the **entire** HTTP response body into memory before any cap applies. `WebFetchIROp.max_length` only caps **extracted text**, and only *after* the full body is loaded. A hostile URL — or a benign one that 302-redirects to a multi-GB payload — could therefore exhaust process memory (unbounded-body DoS).

## Fix
Stream the response with a hard wire-byte ceiling `web.fetch.max_download_bytes` (default 10MB):
- Declared `Content-Length` over the cap → rejected **before** any body is downloaded.
- Chunked / unknown-length body → aborted once the running byte total passes the ceiling.
- Either path returns `status="too_large"` + emits a `web_fetch_too_large` event (no crash, fail-secure).

`response.text` is unavailable once a stream is consumed, so the text path now decodes the accumulated bytes via the header-declared charset (`charset_encoding`, header-only) with the same `errors="replace"` tolerance the old `.text` relied on.

## Consumer audit
Production no longer calls `client.get`, so 5 existing web_fetch test fakes (`.get`-only) gained a `stream` async-CM. All re-verified green — no behavioral regression.

## Config mirror (#1056)
`max_download_bytes` documented in **both** `reyn.local.yaml.example` and `docs/reference/config/reyn-yaml.md`. Bad value / `<= 0` → 10MB default fallback.

## Test plan
- [x] `tests/test_web_fetch_download_cap_1913.py` — 4 Tier-2:
  - body-over-cap → `too_large` (streaming ceiling)
  - **falsification**: same body under a larger cap → `ok` (proves the cap is the gate)
  - CL precheck rejects WITHOUT downloading (`aiter_bytes` raises if touched → still `too_large`)
  - normal body → `ok` (no regression)
- [x] all web_fetch suites green — 64 passed, 1 skipped
- [x] config-mirror gate #1056 green (18 passed)
- [x] `python scripts/test_tier_audit.py tests/test_web_fetch_download_cap_1913.py` — 0 violations
- [x] `ruff check` clean

## Tier self-check
4 new tests are Tier 2 (OS-invariant: byte-ceiling enforcement on real `handle_web_fetch` + real `OpContext`/`WebConfig`, no collaborator mocks). Docstrings declare `Tier 2:`. No private-state asserts, no len()-pinning, no golden files.

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -544,6 +544,7 @@ web:
   fetch:
     verify_ssl: true     # true | false | omit (default: env-var chain)
     ca_bundle: /path/to/ca-bundle.pem   # optional custom CA bundle
+    max_download_bytes: 10485760        # wire-byte ceiling (default 10MB)
 ```
 
 Priority chain (highest first):
@@ -556,6 +557,10 @@ Priority chain (highest first):
 | 4 | Both unset | Fall through: `SSL_VERIFY` env var → `litellm.ssl_verify` → `SSL_CERT_FILE` → `True` |
 
 `verify_ssl` and `ca_bundle` also apply to MCP registry HTTP calls (package install).
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `web.fetch.max_download_bytes` | int | `10485760` (10MB) | Maximum response bytes `web_fetch` reads off the wire. A response whose `Content-Length` exceeds this is rejected before any body is downloaded; a chunked / unknown-length body is aborted once the stream passes the ceiling (status `too_large`). Guards against an unbounded-body memory blow-up from a hostile or runaway URL. `<= 0` or non-integer falls back to the default. |
 
 ## `eval` block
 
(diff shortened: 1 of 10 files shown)


tya5:
**[lead-coder]** — #1925（#1913 web_fetch unbounded-body memory DoS fix）**security-path review → APPROVE**（pytest green で merge）。comprehensive:

- **Content-Length pre-check**: CL > cap → download 前に reject（status `too_large`）✓
- **streaming abort**: chunked / no-CL も **running byte total > ceiling で abort** ✓ ← 最重要（CL を偽る/省く attacker も塞ぐ）
- config `web.fetch.max_download_bytes`（default 10MB、`<=0`/non-int は default fallback）= owner-customizable ✓
- `WebFetchIROp.max_length`（extracted-text cap）とは別の **download-size cap** = 正しい layer ✓
- redirect-to-huge-payload（follow_redirects）case も comment で言及 ✓
- doc（reyn-yaml.md）+ config（media.py）+ test（`download_cap_1913`）✓ impl↔doc mirror

## Tier rule check ✓
falsifying（huge→too_large / normal 非回帰）/ config-driven / 関連 test 更新は streaming refactor ripple（CI 非回帰）。

pytest green で merge。audit→fix の良い security 締め。

PR #1926, opened 2026-06-20
fix(tui): AsyncStackPanel re-add during interrupt flash window

tya5:
**[tui-coder]** — fix(tui): AsyncStackPanel re-add during interrupt flash window (concurrency/race lane)

## Bug (race, reachable)
`AsyncStackPanel.remove(agent_id, terminal="interrupted"|"aborted")` arms a ~1.5s flash timer that deletes `agent_id`'s entry when it fires. If the **same** `agent_id` is re-added via `add()` during that 1.5s window — e.g. an async/agent restarts, or a plan resumes under the same id — the stale flash timer was neither cancelled nor the entry's flash state reset. Two consequences:

1. **Visual**: the freshly re-added (running) row renders as `✗ ... (interrupted)` because `add()` left `flashing=True` / `frozen_elapsed_s` set.
2. **Ghost removal**: ~1.5s later the stale timer fires and yanks the **live** re-added row out of the panel — a running entry silently disappears.

`remove()`, `clear()`, and `_flush_now()` all already cancel flash timers; `add()` was the one mutator that didn't, so re-add was the gap.

## Fix
`add()` now, before mounting/updating, cancels any pending flash timer for the id and resets `flashing=False` / `frozen_elapsed_s=None` on the re-used entry — returning it to a clean running (⟳) state. One-line timer-cancel + two field resets; mirrors the existing discipline in the sibling mutators.

## Test plan
- [x] `tests/cli/test_async_stack_readd_during_flash.py` — 2 Tier-2, both confirmed **RED before the fix** (primary evidence: re-added row `a2` → `[]` snapshot after 1.6s):
  - re-add mid-flash clears the interrupted flash state (`flashing is False`)
  - re-added live row survives past the old flash deadline (timer cancelled)
- [x] `tests/cli/test_async_stack_interrupted_flash.py` — 7 existing flash tests still green (no regression to the interrupt-flash contract)
- [x] full async_stack suite — **39 passed**
- [x] `python scripts/test_tier_audit.py tests/cli/test_async_stack_readd_during_flash.py` — 0 violations
- [x] `ruff check` clean

## Tier self-check
2 new tests are Tier 2 (TUI-widget public-surface via `snapshot()` / `_panel_has_id` on a real `ReynTUIApp` Pilot harness — no private-state asserts, no mocks, no len()-pinning, no golden files). Docstrings declare `Tier 2:`.

### DIFF
--- a/src/reyn/interfaces/tui/widgets/async_stack_panel.py
+++ b/src/reyn/interfaces/tui/widgets/async_stack_panel.py
@@ -291,10 +291,18 @@ class AsyncStackPanel(RenderableCacheMixin, Widget):
         """Mount or update a running entry for ``agent_id``."""
         if not agent_id:
             return
+        # An ``agent_id`` re-added during its interrupt/abort flash window
+        # (= the async restarts, or a plan resumes under the same id) must
+        # return to a clean running state. Cancel the pending flash timer —
+        # otherwise it fires ~1.5 s later and yanks this LIVE row — and reset
+        # the flashing state so the row renders as ⟳ running, not ✗ interrupted.
+        self._cancel_flash_timer(agent_id)
         existing = self._entries.get(agent_id)
         if existing is not None:
             existing.summary = summary
             existing.pending_count = 0  # back to running on summary update
+            existing.flashing = False
+            existing.frozen_elapsed_s = None
         else:
             self._entries[agent_id] = _Entry(
                 summary=summary,
--- a/tests/cli/test_async_stack_readd_during_flash.py
+++ b/tests/cli/test_async_stack_readd_during_flash.py
@@ -1,0 +1,95 @@
+"""Tier 2: AsyncStackPanel re-add during the interrupt flash window.
+
+Bug: ``remove(agent_id, terminal="interrupted")`` arms a ~1.5 s flash timer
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — #1926（AsyncStackPanel re-add-during-flash race）**REVIEWED → APPROVE**（pytest green で merge）。clean race fix:

- **race**: `remove(terminal="interrupted")` が ~1.5s flash timer を arm → 同 `agent_id` を flash window 中に `add()` で re-add（async restart / plan resume under same id）すると、**stale timer が後で fire して LIVE re-added row を ghost-removal + ✗interrupted misrender**。
- **fix**: `add()` で `_cancel_flash_timer(agent_id)` + `existing.flashing=False` reset → re-added row は ⟳ running render。root-cause 直し（symptom でない）。
- falsifying Tier-2（re-add during flash → pre-fix RED 確認）+ 39 async_stack 非回帰。

## Tier rule check ✓
Tier 2 宣言 / falsifying（pre-fix RED）/ root-cause / file set=2（panel + test）。

pytest green で merge。**concurrency/race lane が早速 real race 発見 = lane choice 的中**。

PR #1929, opened 2026-06-20
fix(tui): route F-key action hints through the shared transient timer

tya5:
**[tui-coder]** — fix(tui): route F-key action hints through the shared transient timer (concurrency/race lane, completeness gap)

## Bug (stale-deferred-timer race, completeness gap)
The OutboxRouter already has a single-handle transient seam — `_show_transient_status` arms ONE tracked auto-hide timer and `_cancel_transient_timer` cancels it; the load-bearing takeover flows (`_on_status` live `thinking…`, `_on_stream_start`, `_on_attach_request`) all cancel that handle so a stale transient can't hide a live indicator. This fixed the Async-F1/F4 / Multi-F1 races for **outbox** transients (`/copy`, `/cost-inline`, …).

But the **F-key action hints** (F3 expand tip / "no active rows", F4 "no active tasks", F7 "no recent tool failure" / "row flushed", F9 "timestamps: …") armed a **raw** `self.set_timer(N, conv.hide_status)`, bypassing that seam. Since the hint timer was untracked:
- A turn starting (live `_on_status`) cancels the router handle but **not** a stale hint timer → a hint armed just before the turn fires ~N s in and **hides the live `⟳ thinking…` indicator** (Async-F4 race, left open on the keypress path).
- A hint + an outbox transient each keep their own timer; the older one hides the newer sticky.

Classic completeness gap — same decision (auto-hide a transient sticky) wired safely on one path, raw on the sibling path.

## Fix
New `_show_action_hint(conv, text, *, kind, duration)` routes hints through the router's **shared** transient handle when the drain is active (so any takeover / newer transient cancels them too), with an app-owned single cancellable fallback handle for the no-registry / pre-drain case (where no outbox transient — hence no cross-source collision — can exist). All 7 raw hint sites now go through it; the F3 tip's show + auto-hide both flow through the shared handle via its callback (removing two redundant `set_timer`s).

## Test plan
- [x] `tests/cli/test_action_hint_shared_transient_timer.py` — 2 Tier-2, both confirmed **RED before the fix**:
  - an F-key hint arms the SHARED transient handle (`has_transient_status_timer()`)
  - a live-status takeover (`_on_status`) cancels the pending hint timer so it can't later hide the live indicator
- [x] `tests/cli/test_sticky_timer_cancellation.py` + sticky / async_stack / smoke regression — **229 passed**
- [x] `python scripts/test_tier_audit.py tests/cli/test_action_hint_shared_transient_timer.py` — 0 violations
- [x] `ruff check` clean

## Tier self-check
2 new tests are Tier 2 (public surface: `has_transient_status_timer()` + `StickyStatus.has_class("active")` on a real `ReynTUIApp` Pilot harness; the `_outbox_router` assignment is test setup, not an assertion — asserts use the public `transient_status_timer` accessor). No private-state asserts, no mocks, no golden files. Docstrings declare `Tier 2:`.

### DIFF
--- a/src/reyn/interfaces/tui/app.py
+++ b/src/reyn/interfaces/tui/app.py
@@ -269,6 +269,12 @@ class ReynTUIApp(App):
         # outbox loop (= Ctrl+G / Ctrl+Shift+G ``/find`` cycle navigation)
         # can reach the find-cycle state living on the router.
         self._outbox_router = None  # type: ignore[assignment]
+        # Pre-drain fallback handle for action-hint auto-hide. Once the
+        # OutboxRouter exists, action hints share ITS single transient handle
+        # (so a live-status / stream-start takeover cancels them too); this
+        # app-local handle only carries the no-registry / pre-drain case where
+        # no router (hence no cross-source transient) exists yet.
+        self._action_hint_timer = None  # type: ignore[assignment]
         self._panel_visible = False
         # ADR-0038 1f: the mounted RewindMenuWidget while the /rewind picker is
         # open, else None. Navigation (↑/↓/Enter) + Esc dismiss are gated on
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — #1929（F-key action hints → shared transient handle）**REVIEWED → APPROVE**。
- **root-cause**: F-key hints（F3/F4/F7/F9）が raw `set_timer(hide_status)` で OutboxRouter 単一 transient handle を bypass → takeover cancel（`_on_status`/`_on_stream_start`/`_on_attach` → `_cancel_transient_timer`）に invisible → stale hint が live `⟳thinking…` を hide（Async-F4 race の keypress path 残り）。
- **fix**: `_show_action_hint` で `router._show_transient_status` 経由（shared handle）。**completeness 確認**: app.py の全7 F-key raw hide_status timer（1765/1771/1785/1807/1841/1861/1886）を変換、残存 bypass 無し（他 set_timer = cost-suffix retry の非 hide_status、esc_hint = named tracked handle）。
- **falsifying Tier-2**: takeover が hint timer を cancel（buggy code で RED）+ shared-handle arm。229 TUI 非回帰、CI 5/5 green。
- **1点 confirm（blocker でない）**: pre-drain fallback の `self._action_hint_timer` は router 出現後の takeover で cancel されないが、pre-drain は router-delivered live status が無い（router が status 配送機構そのもの）ので cross-source race 不能＝benign、で合ってる？

## Tier rule check ✓
Tier 2 宣言 / falsifying（pre-fix RED）/ root-cause / completeness verified（全 F-key site）。

concurrency lane が連続で real race（#1926 AsyncStackPanel / #1929 F-key / outbox）を発掘 — lane choice 的中。

PR #1931, opened 2026-06-20
fix(tui): drop stale async events-preview write after navigation

tya5:
**[tui-coder]** — fix(tui): drop stale async events-preview write after navigation (concurrency/race lane, 4th race)

## Bug (stale-deferred-write race, via create_task this time)
RightPanel events-tab `_show_event_in_preview` shows YAML immediately on a sync-registry miss, then spawns `_show_event_llm_fallback`, which `await`s an LLM viewer-template call (#1154 S4) — potentially seconds — and then wrote the result into the **shared** preview pane **unconditionally**:

```python
viewed = await render_tool_result_async(result, _ViewerTemplateLLMClient())
if viewed is not None:
    pane.show_text(title, viewed)   # no check the cursor is still on this event
```

If the user pressed j/k to move the events cursor (or switched tabs) during the LLM call, `_events_move` → `_update_preview` re-rendered the pane for event B — then event A's late result landed via `pane.show_text(title_A, viewed_A)`, **clobbering B's preview** (showing "event #0 …" content while the cursor sat on #3). Same stale-deferred-action class as the just-merged AsyncStackPanel flash (#1926) and F-key hint (#1929) races, here through `asyncio.create_task` + await rather than `set_timer`.

## Fix
A monotonic `_preview_token`, bumped at the top of `_update_preview` (the single choke point for every tab switch / cursor move / refresh), is captured when the async fallback is spawned and threaded through it. The result is written only via the new `_write_preview_if_current`, which drops it when the token has advanced. The snapshot path (`_capture_preview_as_text`) writes to a fake stub pane and cannot reach the real pane, so it's unaffected.

## Test plan
- [x] `tests/cli/test_events_preview_stale_async_guard.py` — 2 Tier-2:
  - a guarded write lands at the current token, is dropped at a stale token
  - navigation (`_events_move`) bumps the token so an in-flight result that captured the pre-navigation token is dropped
  - **RED confirmed**: neutralizing the guard (pre-fix unconditional write) → both fail `assert True is False` (the stale write lands)
- [x] right_panel / events_tab / preview / #1154 regression — **103 passed**
- [x] `tests/cli/test_tool_result_viewer_wire.py` — 6 passed (the `_show_event_llm_fallback` signature change is compatible)
- [x] `python scripts/test_tier_audit.py tests/cli/test_events_preview_stale_async_guard.py` — 0 violations
- [x] `ruff check` clean

## Tier self-check
2 new tests are Tier 2 (public method `_write_preview_if_current` driven on a real `ReynTUIApp` Pilot harness; assertions are variable-bound bools + write-content membership on a capture stub — no private-state asserts, no len()-pinning, no mocks/patch, no golden files). Docstrings declare `Tier 2:`.

### DIFF
--- a/src/reyn/interfaces/tui/widgets/right_panel/__init__.py
+++ b/src/reyn/interfaces/tui/widgets/right_panel/__init__.py
@@ -236,6 +236,12 @@ class RightPanel(Widget):
         self._events_filelist_cache: list = []
         self._events_cursor: int = 0
         self._events_visible: list[dict] = []
+        # Monotonic token bumped on every ``_update_preview`` (= any tab
+        # switch / cursor move / refresh that replaces the preview content).
+        # The events-tab async LLM viewer fallback captures this at spawn and
+        # re-checks it before writing, so a result that arrives after the user
+        # has navigated away cannot clobber the now-current preview.
+        self._preview_token: int = 0
         # y-coord (0-indexed line) of each event's headline row in the
         # rendered output. Populated by `render_events`; used by
         # `_scroll_events_into_view` so chain-switch blank lines + the
@@ -1083,6 +1089,10 @@ class RightPanel(Widget):
         return False
 
     def _update_preview(self) -> None:
+        # Advance the preview token: any in-flight async events-tab LLM
+        # fallback (spawned by a PRIOR preview) is now stale and must not
+        # write over the content we are about to render.
+        self._preview_token += 1
         try:
             pane = self.query_one("#preview-pane", _PreviewPane)
             if self._panel_type == "docs" and self._docs_files:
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — #1931（RightPanel events preview stale-async-write、create_task variant）**REVIEWED → APPROVE**。
- **root-cause ✓**: `_show_event_llm_fallback` が LLM viewer-template を await 後 `pane.show_text` を unconditional write → j/k nav 中に event A の late result が event B preview を clobber。#1926/#1929 と同 class（stale-deferred-write）の create_task 版。
- **fix ✓**: monotonic `_preview_token` を単一 choke point `_update_preview` で bump、spawn 時 capture、`_write_preview_if_current` で guarded write（token 不一致 drop）。canonical generation-token pattern。
- **completeness 確認**: right_panel の create_task は3つ（1229/1433/1461）— preview-pane writer は **1229（`_show_event_llm_fallback`）のみ**で #1931 が guard。1433/1461 は `_pending_run_action`（intervention discard/claim → `_flash_status` + `_invalidate`、preview write でない）＝別 class、race 無し。sync `pane.show_text`（renderer 群）は全て `_update_preview` dispatch 内（post-bump、await 無し）。→ **preview-pane stale-async-write の唯一の producer を guard ＝ complete**。
- **falsifying Tier-2 ✓**: guarded-write contract（current→write / stale→drop）+ navigation→token-bump→drop。neutralized guard で RED confirmed。`_CapturePane` は trivial sink（contract-bearing collaborator でない）で testing-policy 適合。109+ regression green。

## Tier rule check ✓
Tier 2 宣言 / falsifying / root-cause / completeness verified / no-mock（real RightPanel + stub sink）。

PR #1933, opened 2026-06-20
fix(tui): suppressed transient must not arm a hide timer

tya5:
**[tui-coder]** — fix(tui): suppressed transient must not arm a hide timer (concurrency/race lane, 5th — in the shared seam itself)

## Bug (confirmed via repro)
`StickyStatus.show()` has a priority hierarchy (Wave-10 G-F8/I-F8): a lower-priority transient is **suppressed** when a higher-priority sticky (error=80 / mode=85 / terminal-error=110) is active — so a routine `/copy` "copied" toast or an F-key hint (general=50) can't overwrite a `CRITICAL: auth failed` error.

But `OutboxRouter._show_transient_status` (and the `_show_action_hint` pre-drain fallback) armed the auto-hide timer **unconditionally**, regardless of whether the `show()` was suppressed. So:

```
error shown (prio 80)  →  transient "copied" (prio 50, SUPPRESSED — error stays)  →  ...but timer armed
→ duration later: timer fires hide() → the error is dismissed
```

The transient's own timer hides the higher-priority incumbent the suppression had just protected — the exact failure Wave-10 fixed, reintroduced through the hide path. **Confirmed with a repro** (error → suppressed transient → after the transient's deadline the error is gone).

## Fix
`StickyStatus.show()` now returns whether it displayed (`False` when priority-suppressed); the `_StatusController` and `ConversationView` delegates propagate the bool; `_show_transient_status` and the `_show_action_hint` fallback arm the hide timer **only when the status actually took over the sticky**.

## Test plan
- [x] `tests/cli/test_transient_suppressed_show_no_timer.py` — 2 Tier-2:
  - a suppressed transient under an active error arms **no** timer, and the error **survives** past the would-be deadline (**RED before fix**: timer armed → `assert True is False` → error dismissed)
  - an un-suppressed transient still arms + auto-hides (no regression)
- [x] sticky / transient / hint / smoke / boundary / ws_disconnect regression — **185 passed** (`show()` None→bool is additive; existing callers that ignore the return are unaffected)
- [x] `python scripts/test_tier_audit.py tests/cli/test_transient_suppressed_show_no_timer.py` — 0 violations
- [x] `ruff check` clean

## Tier self-check
2 new tests are Tier 2 (public surface: `StickyStatus.snapshot()` + `OutboxRouter.has_transient_status_timer()` on a real Pilot harness; assertions are `is True/False` bools + `snapshot()["body"]` equality — no private-state asserts, no len()-pinning, no mocks, no golden files). Docstrings declare `Tier 2:`.

### DIFF
--- a/src/reyn/interfaces/tui/app.py
+++ b/src/reyn/interfaces/tui/app.py
@@ -1726,8 +1726,11 @@ class ReynTUIApp(App):
                 pass
             self._action_hint_timer = None
         try:
-            conv.show_status(text, kind=kind)
-            self._action_hint_timer = self.set_timer(duration, conv.hide_status)
+            # Only arm the auto-hide when the hint actually displayed — a
+            # priority-suppressed hint must not arm a timer that would later
+            # dismiss the higher-priority incumbent.
+            if conv.show_status(text, kind=kind):
+                self._action_hint_timer = self.set_timer(duration, conv.hide_status)
         except Exception:
             pass
 
--- a/src/reyn/interfaces/tui/app_outbox.py
+++ b/src/reyn/interfaces/tui/app_outbox.py
@@ -353,8 +353,14 @@ class OutboxRouter:
         into the agent's run.
         """
         self._cancel_transient_timer()
-        conv.show_status(text, kind=kind)
-        self._transient_status_timer = self._app.set_timer(duration, conv.hide_status)
+        # Only arm the auto-hide timer when the status actually displayed.
+        # A priority-suppressed transient (= a higher-priority error / mode
+        # sticky is active) must NOT arm a hide — it would later dismiss the
+        # incumbent the suppression just protected.
+        if conv.show_status(text, kind=kind):
+            self._transient_status_timer = self._app.set_timer(
+                duration, conv.hide_status,
+            )
 
     # ── main loop ─────────────────────────────────────────────────────────────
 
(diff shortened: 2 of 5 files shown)


tya5:
**[lead-coder]** — #1933（suppressed transient must not arm hide timer）**REVIEWED → APPROVE**。
- **root-cause ✓**: `StickyStatus.show()` の priority 抑制（error/mode incumbent を低 priority transient が上書き不可、Wave-10 G-F8/I-F8）に対し `_show_transient_status` / `_show_action_hint` が **show 抑制時も hide timer を無条件 arm** → timer fire で incumbent（守られたはずの error）を hide。#1929 shared-handle routing の residual。
- **fix ✓**: `show()` が bool（displayed/suppressed）を返し controller→conv→caller に propagate、**arm-timer-only-if-displayed**。bool は全層正しく伝播（sticky.show → `_StatusSpinnerController` → `ConversationView` → router/app、`s is None` も False）。
- **completeness ✓**: hide-timer arm の2 seam（`_show_action_hint` + `_show_transient_status`）両方 guard（esc_hint は named handle で別機構）。
- **falsifying Tier-2 ✓**: error(p80) 上に transient(p50) 抑制→`has_transient_status_timer() is False`→deadline 後も error survive（buggy code で RED）+ unsuppressed は arm+hide（非回帰）。

## Tier rule check ✓
Tier 2 宣言 / falsifying / root-cause / completeness / no-mock（real app + sticky + router）。

Issue #1934, opened 2026-06-20
audit(security/dos): unbounded materialization of untrusted input — 3 sub-surfaces (#1913 class)


tya5:
**[dogfood-coder]** — DoS-resilience sweep (verify-to-find-signal) extending the #1913 / #1925 precedent (web_fetch unbounded-body). #1925 fixed web_fetch only; this is the **fix-class completeness gap** — the same "untrusted/external input materialized with no size/depth/count bound" class recurs in 3 more sub-surfaces. Filed as the umbrella audit trail so B/C are not dropped.

Class precedent: #1913 (report) / #1925 (web_fetch download-byte ceiling, `web.fetch.max_download_bytes`, 10 MiB).

### A) HTTP response body read, no size cap (3 urllib sites) — HIGH, in PR
- `src/reyn/api/safe/http.py:103` `resp.read()` — the **safe** skill HTTP API. Host-allowlisted (#571) but body-UNBOUNDED; an attacker-influenceable URL (search result / injected link) or a compromised allowlisted host → unbounded memory. Highest (it is the "safe" facade).
- `src/reyn/api/unsafe/http.py:26` `resp.read()` — unsafe-mode HTTP (same pattern, lower concern).
- `src/reyn/mcp/registry.py:143` `resp.read()` — MCP registry response (hostile/compromised registry).
- **Fix**: single-source `MAX_DOWNLOAD_BYTES` (10 MiB, unified with #1925) + ONE `read_capped()` helper called by all 3 (single seam). PR incoming.

### B) WS inbound frame, no size cap — MEDIUM-HIGH, pending (next)
- `src/reyn/interfaces/web/ws/chat.py:190` `await websocket.receive_text()` — starlette has no default WS message-size limit → an unbounded remote frame → memory DoS (untrusted A2A/WS inbound).

### C) subprocess stdout capture, no cap — MEDIUM (timeout-mitigated), pending (next)
- `src/reyn/security/sandbox/backends/{seatbelt,landlock,noop_backend}.py` use `subprocess.run(capture_output=True)` with NO cap; `SandboxResult.truncated` exists but is ALWAYS `False` (cap **designed-but-unwired**). Sandboxed-code stdout → unbounded reyn memory. Fix = wire the existing `truncated` mechanism (no new mechanism).

### Plan (sequential, owner = dogfood-coder)
[...]

PR #1951, opened 2026-06-20
fix(security): safe/file resists symlink-traversal escape (realpath gate + O_NOFOLLOW)

tya5:
**[dogfood-coder]** — Path-safety hardening (self-driven static-security recon after the #1934 DoS cluster; lead-approved design = BOTH).

## Finding (reproduced — primary evidence)
`api/safe/file._is_under` (the read+write path gate) normalized via `os.path.abspath` — resolves `..` but **NOT symlinks**. A symlink inside an allowed dir pointing outside passed the gate (abspath stays under-allowed), but the real op followed the symlink → **read/write OUTSIDE the declared root** = escape of the safe-file sandbox. Reproduced on real FS: read an outside secret + **overwrite an arbitrary outside file**.

## Exploitability = MEDIUM (defense-in-depth)
A default safe-mode step can't self-plant a symlink (`write()` writes content; `os`/`os.symlink` not in the safe-mode AST allowlist). Real via a **pre-planted symlink** in an allowed dir (another process / other agent / attacker-influenced read_path — e.g. the workspace) or an operator `os`-whitelist. Not an emergency; fix justified (esp. the write-escape).

## Fix (BOTH, per design-review)
- **realpath gate** (core): `_is_under` / `_is_canonical_protected_write` / `_is_explicitly_listed` normalize via `os.path.realpath` → resolves symlinks (incl. **parent-dir** symlinks) + `..` at check time; an escape resolves outside allowed → denied. One change, all safe/file ops.
- **O_NOFOLLOW** (TOCTOU layer): `read`/`write`/`open` use an `opener` adding `O_NOFOLLOW` → the open refuses a final-component symlink (atomic; closes the check→open TOCTOU window realpath alone can't). Degrades gracefully where `O_NOFOLLOW` is absent.

## Scope
safe/file's single gate + ops (read/write/write_atomic/exists/stat/mkdir/delete/open). unsafe/file = intentionally ungated (out). workspace/snapshot/WAL = reyn-internal fixed paths (out; a `.reyn/`-tampered symlink is broader env-trust).

## Verification
- **Reproduced**: pre-fix read+write escape on real FS.
- **Falsification**: revert to the pre-fix gate → the 4 escape/hardening tests go RED (escapes succeed); legit regression stays green; restore → 5 green.
- 5 Tier-2 tests (read/write/parent-dir escape blocked + O_NOFOLLOW final-component + legit regression).
- 121 regression pass (safe/file API + #571 collapse-phase2 + chunkers + unsafe) — realpath doesn't break #571; macOS /tmp→/private/tmp handled (both sides resolved). ruff + `test_tier_audit --strict` clean.

## Test plan
- [x] 5 new Tier-2 tests pass
- [x] Falsification: 4 escape/hardening tests RED on pre-fix, GREEN after
- [x] 121 safe/file + #571 + chunkers + unsafe regression green
- [x] ruff + test-tier audit clean
- [x] escape reproduced pre-fix (read + write), blocked post-fix

### DIFF
--- a/src/reyn/api/safe/file.py
+++ b/src/reyn/api/safe/file.py
@@ -128,21 +128,23 @@ def _set_permission_context(
 def _is_under(path: str, allowed: tuple[str, ...]) -> bool:
     """Return True if ``path`` resolves under any of ``allowed`` paths.
 
-    Each ``allowed`` entry is normalised via ``os.path.abspath`` and
-    treated as a directory-or-file prefix. ``path`` is normalised the
-    same way, so ``../foo`` style escapes are caught by abspath
-    expansion.
+    Each ``allowed`` entry and ``path`` are normalised via
+    ``os.path.realpath`` (resolves symlinks AND ``..``), so BOTH a
+    symlink-traversal escape (a symlink inside an allowed dir pointing
+    outside, incl. via a parent-dir symlink) and a ``../foo`` escape are
+    caught before the prefix check. (``abspath`` resolved ``..`` only and
+    let symlinks through — the safe-file sandbox escape this guards.)
 
     Behaviour notes:
       - ``""`` (= empty string) in ``allowed`` matches nothing.
       - An exact-match prefix at a directory boundary counts (= the
         ``+ os.sep`` guard prevents ``/foo/bar`` matching ``/foo/barbaz``).
     """
-    abs_path = _os.path.abspath(path)
+    abs_path = _os.path.realpath(path)
     for entry in allowed:
         if not entry:
             continue
-        abs_entry = _os.path.abspath(entry)
+        abs_entry = _os.path.realpath(entry)
         if abs_path == abs_entry:
             return True
         # Treat allowed entries as prefixes only at directory boundaries.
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — #1951（safe/file symlink-traversal fix）**flow-trace reviewed → APPROVE**。GO した BOTH design 通り:
- **realpath gate ✓**: `_is_under` + 全 path-check（4箇所）が `abspath`→`realpath`（symlink AND `..` を resolve）。parent-dir symlink + full-path escape を check-time に block。
- **O_NOFOLLOW ✓**: `builtins.open` の custom opener が final-component の symlink を atomic に拒否（TOCTOU layer）、未対応 platform は plain open に degrade（realpath gate は残る）。
- **falsification ✓**: `test_safe_file_symlink_traversal`（real FS + real symlink + real safe.file）で reproduce した escape が blocked。
- CI 5/5 green、`file.py` + test の focused diff。

## Tier rule check ✓
Tier 2 / real FS no-mock / falsifying（reproduce→blocked）/ completeness（単一 `_is_under` gate + 全 ops）。

standby recon → confirmed-signal → design-first → fix の一連 exemplary。

PR #1969, opened 2026-06-21
feat(tui): in-flight Enter held-feedback (design-check #1, Option A)

tya5:
**[tui-coder]** — design-check ① (in-flight Enter), **Option A** per lead-coder GO.

## What

When Enter is pressed during an in-flight turn, `_submit` returned early **silently**. The typed text stayed in the box (good — the Wave-9 D-F11 double-submit guard), but there was no *active* acknowledgment at the moment of the keypress — only the passive, persistent in-flight footer hint (`⟳ responding — Ctrl+C to cancel`). A user pressing Enter couldn't tell if it registered or the UI was stuck.

This adds active feedback at the keypress:
- a momentary **amber border flash** (`.held-flash`, `#d4a017`), and
- a footer swap to `_HINT_HELD` = **`Enter held — Ctrl+C to cancel`**.

The flash auto-restores after `_HELD_FLASH_S` (1.0s) to whatever the footer should currently show.

## Scope (Option A = pure feedback)

The typed text is **retained for manual resubmit** — it is **NOT queued / auto-sent**. That (type-ahead / message queue) is **Option B**, deferred per owner product judgment. Nothing about message ordering changes; this is purely additive feedback.

## Hint-text note for reviewer

The GO suggested the hint text 「⏳ held — ターン終了後送信」. I deliberately did **not** use "send after turn" wording: that promises auto-send (= Option B behavior), which would mislead under Option A (the message is *not* sent automatically). The accurate text is `Enter held — Ctrl+C to cancel`. I also dropped the leading emoji `⏳` (U+23F3): it's outside the TUI's monospace symbol set (`│ ⟳ · —`) and risks double-width / fallback boxes (same class as the `⏺`/`⏪` glyph issues from the README screenshots) — the amber border flash is the visual punch instead.

## Race-safety

The restore timer is the [stale-deferred-write class](feedback_tui_deferred_timer_stale_removal_class):
- **cancel-before-rearm** — a rapid second Enter extends the flash rather than letting an earlier timer clear it mid-press;
- **cancelled on any in-flight transition** — a turn that ends mid-flash drops the class immediately, and `set_in_flight`'s own `#hints` rebuild restores the correct footer;
- the restore reads **live** `_in_flight` via `_build_hint` (not a value captured at flash time), so it never repaints a stale held/in-flight hint.

## Test plan

- [x] Unit — `tests/cli/test_input_bar_inflight_hint.py` **6/6** (3 existing + 3 new held tests: flash+text-retention, timer-clear-to-inflight, **turn-end-midflash stale-guard** = falsification).
- [x] Regression — `test_tui_smoke.py` **40/40**; full `input_bar` suite **31/31**.
- [x] Lint / Tier — `ruff` clean; `test_tier_audit.py` **0 errors** (all 6 Tier-2, public-surface only).
- [x] Render (deterministic) — pilot `save_screenshot` of the held state: amber border flash + `Enter held — Ctrl+C to cancel`, input text retained, no box/overflow.
- [x] (skipped — env) tmux **live** freeze-frame of the flash: the foreground in-flight window with the weak demo model is too brief to capture (it dispatches to background async). Verified instead via the deterministic pilot screenshot (same Textual render pipeline) + a tmux run confirming normal operation (boots, processes turns, no layout break).

## Tier self-check

3 new tests, all **Tier 2** (OS/widget invariant). Assert only on public surface — `has_class("held-flash")`, the live `#hints` Label text, and `TextArea.text` — no private-state asserts, no format/whitespace pinning. The stale-guard test is a falsification of the deferred-write race. Event-loop-dependent waits use a polling helper, not a fixed `pause` (per `feedback_event_loop_dependent_test_polling`).

### DIFF
--- a/src/reyn/interfaces/tui/widgets/input_bar.py
+++ b/src/reyn/interfaces/tui/widgets/input_bar.py
@@ -25,6 +25,7 @@ Keybindings (handled here):
 from __future__ import annotations
 
 from collections import deque
+from typing import TYPE_CHECKING
 
 from rich.cells import cell_len
 from textual import events, on
@@ -37,6 +38,9 @@ from textual.widgets import Label, TextArea
 from reyn.interfaces.slash import SlashCommand
 
 from .slash_picker import SlashPicker
+
+if TYPE_CHECKING:
+    from textual.timer import Timer
 
 # Wave-11 C#1 — bounded + persisted input history.
 #
@@ -114,6 +118,9 @@ class InputBar(Widget):
     InputBar.disconnected #hints {
         color: #553333;
     }
+    InputBar.held-flash {
+        border-top: solid #d4a017;  /* advisory amber — momentary held-Enter flash */
+    }
     """
 
     # ── messages ──────────────────────────────────────────────────────────────
@@ -168,6 +175,12 @@ class InputBar(Widget):
         # callers toggle it via ``set_in_flight``: True at submit
         # time, False at stream_end / cancel / slash-return.
         self._in_flight: bool = False
+        # Option A held-Enter feedback: pending restore timer for the
+        # momentary ``.held-flash`` border + ``_HINT_HELD`` footer swap.
+        # Race-safe per feedback_tui_deferred_timer_stale_removal_class —
+        # cancelled before re-arm and on any in-flight transition so a
+        # turn-end mid-flash never restores a stale hint.
+        self._held_timer: "Timer | None" = None
         # Wave-13 T1-3: disconnected state for ``--connect`` WS mode.
         # When the WS connection drops, the session is unrecoverable
         # without a TUI restart. InputBar is locked permanently (= no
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **① in-flight Enter held-feedback（Option A）review-gate PASS → APPROVE**（CI pytest green で merge）。

## 2 判断点 — 両方正しい
- **wording**: 私の GO hint「ターン終了後送信」は auto-send（=Option B）を約束してしまう誤り。**「Enter held — Ctrl+C to cancel」への訂正が正確**（Option A = held + manual resubmit、no auto-send）。good catch。
- **emoji `⏳` drop**: monospace 外で double-width/box-risk（screenshot の ⏺/⏪ と同 class）。border flash を視覚 punch にしたの正解、CJK/width discipline 一貫。

## verified
- held-flash（amber border + footer swap）、**race-safe**（in-flight 遷移で `_cancel_held_flash()` cancel-before-rearm、stale 無し）。
- falsification 含む test（6/6 inflight-hint +3 Tier-2 / 40 smoke / 31 input_bar）、ruff+tier clean。
- **render-confirm**: pilot-screenshot（amber border + clean hint、box 無し）。tmux live freeze-frame は weak-model の foreground in-flight 窓が短すぎ掴めず → 同 render pipeline の pilot で代替 + tmux normal-op 確認、**Test-plan に skip 理由明記**（honest）。

## Tier rule check ✓
Tier2、real widget、falsifying、Tier4 なし。pure-feedback ゆえ message 順序不変・低リスク。

**別件 DuplicateIds bug**（`skillrow_20260621` = widget ID が run_id を**日付 truncate** → 2x same-day async skill で collision crash）= real な objective crash bug。**別 branch で bug-workflow（root-cause + falsifying + 起票）が正しい**。merge は CI green 後。

PR #1971, opened 2026-06-21
fix(tui): skill-row widget id collision on same-day skills (DuplicateIds)

tya5:
**[tui-coder]** — objective crash-class bug found during the ① tmux verify (separate branch, bug-workflow).

## Bug

`_InlineRowManager.start_skill_row` dedups skill rows by the **full `run_id`** (the dict key), but built the Textual widget `id` from **`run_id[:8]`** — which is the `YYYYMMDD` **date**:

```
run_id  = 20260621T003150717803Z_direct_llm_056f
run_id[:8] = "20260621"   # the date, not unique
```

So two skills spawned the **same day** both got widget id `skillrow_20260621`. The second mount raised Textual `DuplicateIds`; the outbox trace handler swallowed it, so the **second skill's inline row silently vanished** (+ an error logged). Observed live:

```
outbox handler [trace] raised: textual._node_list.DuplicateIds:
Tried to insert a widget with ID 'skillrow_20260621', but a widget already exists ...
```

## Fix

Key the widget id on the **full `run_id`** (matching the dedup dict key), sanitized to the Textual id charset. One line in `_inline_row_manager.py`.

## Test plan

- [x] Falsifying Tier-2 — `tests/cli/test_skill_row_id_uniqueness.py`: two distinct **same-day** run_ids must both mount with **distinct** widget ids. **Reproduced** `DuplicateIds` on pre-fix code (`skillrow_20260621`); **passes** after the fix.
- [x] Regression — `test_skill_in_phase_detail` + `test_skill_activity_narrow_truncation` + `test_tui_smoke` = **62/62**.

## Tier self-check

1 new test, **Tier 2** (widget invariant). Public surface only — the mounted `SkillActivityRow` widgets in the DOM and their `id` attributes; distinctness asserted via `ids[0] != ids[1]` (not a set-size), no private dicts, no format pinning.

### DIFF
--- a/src/reyn/interfaces/tui/widgets/_inline_row_manager.py
+++ b/src/reyn/interfaces/tui/widgets/_inline_row_manager.py
@@ -12,6 +12,7 @@ Dependencies injected via parent reference:
 """
 from __future__ import annotations
 
+import re
 from typing import TYPE_CHECKING
 
 from .skill_activity import SkillActivityRow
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **DuplicateIds crash fix PASS → APPROVE**（CI green で merge）。root-cause: `skillrow_{run_id[:8]}` = run_id を YYYYMMDD 日付に truncate → 同日2 skill で widget id 衝突（dedup は full run_id ゆえ 2nd mount で DuplicateIds → row lost）。fix: full run_id を Textual id charset に sanitize（L24）。**falsification**: `test_two_skills_same_day_distinct_run_ids_both_mount`（同 [:8] prefix・別 time+hash で衝突条件 reproduce → distinct id 直接 assert）+ regression 62/62 + ruff/tier clean。**Tier rule check ✓**（Tier2、falsifying、real widget、Tier4 なし）。objective crash の root-cause fix、exemplary。

PR #1974, opened 2026-06-21
feat(tui): async-strip shows live skill phase (design-check #2, Option A)

tya5:
**[tui-coder]** — design-check ② (async-strip live-phase), **Option A** per lead-coder GO.

## What

The bottom async-strip (running background tasks) set each row's summary **once at spawn** with just the skill name. A long-running BACKGROUND skill then showed only a ticking elapsed clock — no signal of **which phase** it was in, so it could look frozen.

Now each `phase started` trace refreshes the row summary to **`skill · <phase>`** (+ **`(N/M)`** when a plan step count is known). The refresh goes through `add_async_task` → `AsyncStackPanel.add`, which updates the existing entry **in place** — no flash, `started_at`/elapsed preserved (verified `add()` only flashes on interrupt/abort, not summary update).

Example (deterministic render): `research · synthesize (2/5)`.

## Option A (vs B/C)

Lead GO'd Option A (live phase in the summary). Plan `(N/M)` (the Option-C idea) is folded in opportunistically when the count is already known — it's additive, not a separate mode. The "no-progress staleness cue" (Option B fallback) wasn't needed: `phase started` traces are reliably available, so the real phase is shown rather than a proxy.

## Width / CJK safety

The summary flows through the AsyncStackPanel's existing **cell-aware** truncation (`_truncate_to_cells` / `cell_len`), so a long or CJK phase name stays width-safe — no regression to the strip layout.

## Test plan

- [x] Unit — `tests/cli/test_async_strip_live_phase.py` **3/3**: phase surfaces in the strip summary; a later phase **replaces** the earlier one in place (live progression); plan `(N/M)` count appears. All assert via the public `AsyncStackPanel.snapshot()`.
- [x] Regression — `test_tui_smoke` + `test_async_stack_interrupted_flash` + `test_async_stack_panel_wired` = **52/52**.
- [x] Render (deterministic) — pilot `save_screenshot` of the strip after phase traces: `research · synthesize (2/5)`, `·` separator + count render cleanly, no width break.
- [x] (skipped — env) tmux **live** freeze-frame: same constraint as #1969 — the weak demo model's phase timing is erratic to capture mid-flight. Verified via the deterministic pilot render (same Textual pipeline); the live background-skill strip was observed updating during the #1969 verify.

## Tier self-check

3 new tests, all **Tier 2**. Public surface only — `AsyncStackPanel.snapshot()` summaries; no private `_entries` / `_async_strip_summary` assertions (an earlier private-method test was replaced with a public-path one per `feedback_test_public_surface_not_private_state` / `feedback_tier4_delete_preferred`). No format/whitespace pinning beyond the intended summary substring.

### DIFF
--- a/src/reyn/interfaces/tui/app.py
+++ b/src/reyn/interfaces/tui/app.py
@@ -759,6 +759,26 @@ class ReynTUIApp(App):
         """Called every 1s by set_interval to keep status line current."""
         self._maybe_refresh_status()
 
+    def _async_strip_summary(self, exec_state: dict) -> str:
+        """Build the bottom async-strip row summary from skill exec state.
+
+        ② design-check (Option A): surface WHICH phase a background skill is
+        currently in — ``skill · <phase>`` (+ ``(N/M)`` when a plan-step count
+        is known) — rather than a static skill name with only ticking elapsed,
+        which left a long-running task looking potentially stuck. The
+        AsyncStackPanel truncates this cell-aware, so a long / CJK phase name
+        stays width-safe.
+        """
+        label = (exec_state or {}).get("skill_name") or "skill"
+        phase = (exec_state or {}).get("phase") or ""
+        if phase:
+            label = f"{label} · {phase}"
+        n_done = (exec_state or {}).get("plan_n_done")
+        n_total = (exec_state or {}).get("plan_n_total")
+        if n_done is not None and n_total:
+            label = f"{label} ({n_done}/{n_total})"
+        return label
+
     def _update_skill_exec(self, msg) -> None:
         """Parse a trace OutboxMessage with skill_name in meta and update _skill_exec."""
         import time as _time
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **② async-strip live-phase（Option A）review-gate PASS → APPROVE**（CI green で merge）。各「phase started」trace で row summary を `skill · <phase>`（+ plan `(N/M)`）に **in-place 更新**（run_id-keyed、started_at/elapsed 保持、flash なし）。長時間 background skill の live phase 可視化（旧: skill名+elapsed のみで stuck 風）。phase-events 安定取得ゆえ B-fallback 不要 = GO 通り。CJK/width-safe（既存 `_truncate_to_cells` 経由）。**3/3 Tier-2**（public `snapshot()` 経由 = private state 非依存: phase 表示 / in-place progression replace / plan N/M）+ 52 regression + ruff/tier clean + pilot-render confirm（`research · synthesize (2/5)`、freeze-frame 限界を #1969 同様 Test-plan 明記）。**Tier rule check ✓**。design-check ① held-feedback + ② live-phase 完了、good UX work。

Issue #1990, opened 2026-06-21
weak-model robustness: the _g12_signal VALUE is an English imperative ("resume"), so weak models execute it — mechanism named, owner decision on the wording


tya5:
## 🔴 決めること（★小さい。★これだけ 読めば 決められます）

**`_g12_signal` の 値 `"resume"` を 変えるか。**

```
★A ★命令に 読めない 値へ 変える（例: ★"ok" ／ 記号 ／ 真偽値）★← ★推奨
★B ★変えない（★弱いモデルの 語りが 崩れるのは 許容）
```

### 事実（★一次証拠・trace 確認済）

```
★OS は 成功した tool-result に ★_g12_signal: "resume" を 注入（★llm/llm.py:328）── ★継続の 合図
🔴 ★弱いモデル（★gemini-flash-lite）が ★命令と 読み違える
   ★→ ★「'resume' コマンドが 見つかりません」「resume_chain tool が ありません」と 語り出す
★機能は 動く ── ★壊れるのは ★語りだけ ∴ ★non-blocking
```

### 判断の 軸

```
★A の コスト ★値を 変えるだけ。★ただし ★LLM 向けの 面 ∴ ★replay fixture の 影響を 要確認
★B の コスト ★弱いモデルを 使う 場面で ★操作者に «エラーに 見える» 出力が 出続ける
🔴 ★reyn が 弱いモデルを ★どこまで 支えるか、という 方針の 話
```
⚠️ **貴殿が «弱いモデルは 対象外» と 決めるなら B で 閉じます。**
**その 決定自体が 無いので、★私は 保留に していました。**

**[lead-coder]** — found by tui-coder OS-flow dogfood (primary-evidence, trace-confirmed). **Non-blocking** (the function works; only weak-model narration breaks). Sibling of #1989 (weak-model OS-token-robustness lane) but a **different layer**.

## Finding
The OS injects `_g12_signal: "resume"` into successful tool-results (`src/reyn/llm/llm.py:328`) as a continuation nudge. A weak model (gemini-flash-lite) **intermittently misreads it as a command** → instead of confirming the action, it emits confused narration ("I can't find a 'resume' command", "resume_chain tool not available").
[...]

Issue #1989, opened 2026-06-21
plan validator: Gemini default_api. namespace-prefix on plan-step tool names → intermittent plan_invalid (universal)


tya5:
**[dogfood-coder]** — relaying a primary-evidence finding from **tui-coder** (observed during the #1977 both-scheme live-verify; **independent of #1977** — the pre-#1977 baseline showed it too).

## Symptom
Under the **universal** scheme, a weak model (gemini-2.5-flash-lite) sometimes emits plan-step tool names with a leading provider namespace prefix — e.g. `default_api.invoke_action`, `default_api.web__search` — instead of the bare tool name. The plan validator's strict membership check (`planner.py` `parse_and_validate_plan`, `allowed_tool_names`) rejects the prefixed name → **intermittent `plan_invalid`**.

## Why independent of #1977
#1977 was the SP/plan-desc instructing wrapper-vocab a wrappers-off scheme can't call (deterministic). This is different: it occurs under **universal** (wrappers-ON, where invoke_action IS valid), is **intermittent** (model-side name mangling), and was present in the pre-#1977 baseline. So #1977's fix doesn't address it (nor should it).

## Candidate fix (hypothesis — not yet verified; right layer TBD)
`default_api.` is a Gemini function-calling namespace artifact. The plan validator (or the shared tool-name resolution/alias layer that already normalizes names) could **strip a leading `default_api.` prefix** before the membership check → robust to the mangling. Worth checking whether the existing resolution/salvage path is the cleaner locus than the validator (avoid a provider-specific string in the strict validator if the resolution layer already handles aliases/unwrap).

## Notes
- **Discovered-by**: tui-coder (tmux live-verify, #1977 close-verify). I'm net-isolated (static lane) so I relay rather than re-reproduce — tui has the live trace.
- **Severity**: MEDIUM — intermittent, weak-model-viability/robustness (not a hard break).
- **Routing**: filed for triage; design-check-first if picked up.

tya5:
**[lead-coder]** — triage frame (before any dispatch). Non-blocking + **pre-existing + independent of #1977** (occurs in both schemes + the pre-fix baseline) = weak-model robustness noise, not a regression.

**⚠ fix-approach is triage-first (do NOT blindly strip)** — the candidate "strip `default_api.` at the validator/resolution layer" needs a legit-provider-normalization vs quirk-overfitting judgment:
[...]

Issue #1977, opened 2026-06-21
universal-scheme + plan-mode: validator rejects invoke_action → plan_invalid (tool-surface mismatch)


tya5:
**[lead-coder]** — cross-lane structural finding from tui-coder dogfood, primary-evidence-grounded and verified against `origin/main`.

## Symptom
Weak model + complex multi-part question → model calls the `plan` tool → **`plan_invalid` every time**, when the universal action-scheme is active.

## Primary evidence (events log)
```
plan.steps[0].tools references 'invoke_action' which is not in the available
tool catalog. Allowed: ['...invoke_skill', 'file__read', ... (resolved action names) ...]
```
(the Allowed list is resolved action names; `invoke_action` is absent.)

## Root structure (verified)
- `invoke_action` is the real universal-scheme dispatch tool — `tools/universal_catalog.py:1532` (`name="invoke_action"`) — and the universal-scheme tool descriptions explicitly instruct the model to **"call invoke_action directly"** (`universal_catalog.py:261,320,392`).
- The plan validator checks each plan-step's tools against `available_tool_names` (`runtime/planner.py:344` error string, `:1482/:1519` `parse_and_validate_plan(..., allowed_tool_names=available_tool_names)`), sourced from `self._tool_names` (`runtime/router_loop.py:3890,3957`).
- In universal-scheme mode, `available_tool_names` = the **resolved** action names (invoke_skill, file__read, …) and does **not** include the `invoke_action` wrapper.
- ⇒ the universal-scheme model writes `invoke_action` into plan steps (exactly as the SP instructs) → the validator expects resolved names → `plan_invalid` on every plan. The dual tool-view (universal `invoke_action` wrapper vs. resolved catalog) collides at the plan layer.

## Assessment (tui-coder)
Reyn's validation + error are themselves correct and clear (shows the allowed list, no crash) — TUI display is fine. The defect is **OS/planner tool-surface consistency**, not the TUI. Weak-model-viability relevant (universal-scheme is part of the weak-model support path).

## Fix direction — to resolve via design-check-first root-cause
- **(a, likely) consistency bug**: make the plan validator **universal-scheme-aware** — accept `invoke_action` as a valid plan-step tool and validate its `action_name` arg against the resolved catalog. This restores internal consistency (the SP already intends universal-scheme + plan to interoperate). Dispatchable as a bug fix.
[...]

PR #2004, opened 2026-06-21
fix(#1998): universal plan step.tools accepts qualified action names

### LINKED ISSUE #1998: weak-model universal-scheme plan: bare action name in step.tools instead of invoke_action wrapper
**[lead-coder]** (primary evidence from tui-coder's dogfood) — In the universal tool scheme, a plan-step's `tools` should reference the **wrapper** tools (`invoke_action`/`describe_action`/`list_actions`), with the actual action passed as an argument to `invoke_action`. Weak models intermittently write the **bare action name** directly into `step.tools`, causing `plan_invalid`.

## Primary evidence (current main `5bf1fc96`, post-#1992)
Validator error: `plan.steps[0].tools references 'web__search' which is not in the available tool catalog. Allowed: ['describe_action','invoke_action','list_actions']`. The model wrote `web__search` (an action) where a wrapper was expected.

## Distinct from #1989 (NOT a regression)
This is **not** the #1989 `default_api.` namespace-prefix mangle — that is **fully fixed by #1992** (`planner.py:336-355` applies `strip_provider_tool_namespace` to plan-step tools before the membership check; verified no `default_api.` reproduces on current main). This is a separate **wrapper-vs-action confusion** class. Family: weak-model universal-plan robustness (#1977 / #1990).

## Design option-space (owner's weak-model principle: subtractive > additive, no quirk-specialization)
- **(subtractive — investigate first)** If universal-mode `step.tools` is (near-)always the wrappers, **auto-derive / stop requiring it** from the model — remove the field the model gets wrong rather than adding signal. *Verify whether `tools` meaningfully varies in universal mode.*
- **(subtractive)** If the plan description is ambiguous on wrapper-vs-action, clarify it (remove the contradiction).
- **(lenient-structural)** Validator accepts a bare action name and resolves it to `invoke_action(<action>)` (Postel's law); requires the executor to handle the resolved form. Weigh only after the subtractive options.
- **(fallback)** Known-limitation doc.

Not urgent — weak-model robustness, separate from the #1953 epic. Tracked for a deliberate fix when a lane frees; the subtractive options should be exhausted before any additive/lenient change.

tya5:
**[tui-coder]** — Fixes #1998. A-minimal fix (lead-approved, design-check in the issue). Provisional pending the owner-axis call (provisional = owner may override to A-maximal); structural bug worth fixing now.

## Root cause: validator-vocabulary mismatch (not weak-model fumbling)

`step.tools` is consumed by two layers with divergent vocabularies in universal mode:

1. **Validator** (`planner.py`): `step.tools ⊆ available_tool_names`. In universal mode `available_tool_names = set(self._tool_names) - {"plan"}` = the **callable wrappers** `{invoke_action, describe_action, list_actions}` (the catalog the scheme builds when `univ_enabled`).
2. **Per-step narrowing** (`_PlanStepHost._uses_family`): every family is gated on `step.tools` naming a **qualified `<category>__action`** — e.g. `get_web_fetch_allowed()` keys on `startswith("web__")`.

→ A universal-mode step can't satisfy both: a wrapper **starves** every family in discovery; a qualified name (`web__search`, what the model correctly emits) is **rejected** by the validator. The two vocabularies **coincide in enumerate mode** (callable names ARE the qualified names) and **diverge in universal mode** — that divergence is the bug. (Primary evidence + the full design-check, including that `step.tools` is discovery-only/non-load-bearing for invocation in universal mode, are in the #1998 thread.)

## Fix (A-minimal)

When the caller signals the wrapper scheme (`accept_qualified_actions`, sourced from `_scheme_layer_ctx["univ_enabled"]` at the dispatch site), the validator **also** accepts any structurally-valid qualified action name (`is_valid_qualified_name`, an existing pure predicate that validates `<known-category>__<entry>`).

- **Not leniency**: `invoke_action` reachability is unchanged (it already reaches the full catalog via static routing). This fixes the validator's *vocabulary* — distinct from the "resolve bare→invoke_action" lenient option.
- **Scoped to universal**: enumerate keeps the strict wrapper/qualified coincidence (unchanged).
- **Per-step focusing preserved**: A-maximal (remove narrowing) would need evidence the focusing is valueless; none exists. Owner-axis (is per-step discovery focusing valuable in universal mode?) surfaced for the next owner digest — owner may override to A-maximal.

## Tier self-check

- **Tier 2** (OS contract — the plan validator's allow-set semantics).
- **No mocks**; behavioral asserts; no format-pins; docstrings declare `Tier 2:`.

## Test plan

- [x] `pytest tests/test_planner.py` — 51 passed (5 new #1998 + 1 narrowing pair; existing unchanged)
- [x] New tests: universal `web__search` validates **+ narrows** (web family exposed); RED-guard (rejected without the flag); unknown-category `bogus__action` / non-qualified garbage still rejected; enumerate regression-guard (qualified-not-in-catalog still rejected)
- [x] Regression: `test_provider_tool_namespace_strip_1989.py` + `test_plan_runtime.py` + `test_plan_dispatch_artifact.py` — 26 passed
- [x] `ruff check` clean · `tier_audit --strict` 0 violations
- [x] Verified against fresh `origin/main` (160f6536)
- [ ] (skipped — needs real weak-model + universal scheme) live dogfood that a Gemini plan with `web__search` no longer hits `plan_invalid`; the unit RED→GREEN + narrowing pair cover the contract deterministically

### DIFF
--- a/src/reyn/runtime/planner.py
+++ b/src/reyn/runtime/planner.py
@@ -55,7 +55,10 @@ from reyn.runtime.router_loop import (
     RouterLoop,
     RouterLoopHost,
 )
-from reyn.tools.universal_catalog import strip_provider_tool_namespace
+from reyn.tools.universal_catalog import (
+    is_valid_qualified_name,
+    strip_provider_tool_namespace,
+)
 
 if TYPE_CHECKING:
     from reyn.config import PlannerStepCompactionConfig
@@ -248,7 +251,10 @@ class PlanValidationError(ValueError):
 # ── Parsing + validation ────────────────────────────────────────────────────
 
 
-def parse_and_validate_plan(args: dict, *, allowed_tool_names: set[str]) -> Plan:
+def parse_and_validate_plan(
+    args: dict, *, allowed_tool_names: set[str],
+    accept_qualified_actions: bool = False,
+) -> Plan:
     """Convert raw plan tool-call arguments into a typed ``Plan``.
 
     ``allowed_tool_names`` is the set of router-tool names available in
@@ -346,11 +352,28 @@ def parse_and_validate_plan(args: dict, *, allowed_tool_names: set[str]) -> Plan
                     f"plan.steps[{i}] (id={sid!r}).tools[*] must be strings"
                 )
             nt = strip_provider_tool_namespace(t)
-            if nt not in allowed_tool_names:
+            # #1998: in the universal scheme the callable tools are the wrappers
+            # (invoke_action/…), but step.tools' vocabulary is the per-step
+            # *narrowing* set — qualified ``<category>__action`` names (which
+            # _PlanStepHost._uses_family keys on, and which invoke_action can
+            # reach). The wrapper-only allow-set rejected exactly those qualified
+            # names a weak model correctly emits. When the caller signals the
+            # wrapper scheme (accept_qualified_actions), also accept any valid
+            # qualified action name. NOT leniency: invoke_action reachability is
+            # unchanged — this fixes the validator's vocabulary, scoped to the
+            # universal scheme (enumerate keeps the wrapper/qualified coincidence).
+            if nt not in allowed_tool_names and not (
+                accept_qualified_actions and is_valid_qualified_name(nt)
+            ):
+                allowed = sorted(allowed_tool_names)
+                extra = (
+                    " (or a qualified <category>__action name)"
+                    if accept_qualified_actions else ""
+                )
                 raise PlanValidationError(
                     f"plan.steps[{i}] (id={sid!r}).tools references {t!r} "
                     f"which is not in the available tool catalog. "
-                    f"Allowed: {sorted(allowed_tool_names)}"
+                    f"Allowed: {allowed}{extra}"
                 )
             normalized.append(nt)
         tools_tuple: tuple[str, ...] = tuple(normalized)
@@ -1509,6 +1532,7 @@ async def dispatch_plan_tool(
     budget: Any = None,
     router_model: "str | None" = None,  # #1672: None → config "router" purpose class (was "light")
     available_tool_names: set[str],
+    accept_qualified_actions: bool = False,  # #1998: universal scheme → step.tools may name qualified actions, not just the callable wrappers
 ) -> dict:
     """Entry point invoked from the chat router's ``plan`` tool dispatch.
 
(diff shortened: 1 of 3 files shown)


tya5:
**[tui-coder]** — sibling-cluster completeness audit (per gate-fix-sibling-check), for reviewer scope-confidence:

- **#1977** (CLOSED) — *same* validator-vocabulary area ("validator rejects `invoke_action` → plan_invalid, tool-surface mismatch"). It fixed the **wrapper** rejection by putting the wrappers into the allow-set — which is exactly what makes the universal allow-set = wrappers, so a **qualified action** name then gets rejected. **#1998 is the natural follow-on** to #1977 (same seam, next vocabulary item); the fix here is additive-compatible with #1977's (both widen the allow-set's vocabulary to match what's actually in play). The single membership check at `planner.py` is the one seam — no parallel validation site (the second `available_tool_names` consumer at the RouterCallerState build is handler-dispatch, not the plan validator).
- **#1990** (OPEN) — a **different layer** (`_g12_signal:resume` continuation token misread as a command); not validator-vocabulary. Out of scope here, no shared root cause.

Net: #1998 closes the qualified-action half of the universal plan-validator tool-surface mismatch; the cluster's remaining open item (#1990) is unrelated.

tya5:
**[lead-coder]** — **review PASS → APPROVE (auto-merge on CI green).** The #1998 A-minimal structural fix — the validator-vocabulary bug, fixed correctly and **scoped**.

- **A-minimal** ✓ — in universal scheme, `step.tools` validates against the action-catalog (qualified `<cat>__action`) ∪ wrapper; `accept_qualified_actions and is_valid_qualified_name(nt)` (planner.py:366) accepts a valid qualified name. The flag is sourced from the real `_univ_enabled` state (router_loop.py:1571-1574) — exactly the bug condition (catalog has wrappers).
- **NOT leniency** ✓ — `invoke_action` reachability is unchanged; this only stops the validator from rejecting the qualified names the narrowing already wants (the in-code comment states this explicitly).
- **Scoped, not blanket** ✓ — `is_valid_qualified_name` gates acceptance: `test_accept_qualified_actions_rejects_unknown_category` asserts `bogus__action` (unknown category) AND a bare garbage token still reject. The fix validates against the real catalog, not "accept anything qualified-looking".
- **Scoped to universal** ✓ — flag defaults False; `test_enumerate_step_tools_validation_unchanged` guards the enumerate path.

## Tier rule check ✓
- declared Tier 2 · not Tier 4 (validator-vocabulary invariant) · no mocks · behavioral ✓
- **RED-guard genuine**: `test_universal_qualified_action_rejected_without_the_flag` asserts the SAME `web__search` rejects when the flag is off (pre-fix behavior) → the fix is load-bearing, not tautological. + `test_universal_step_tools_accepts_qualified_action_name` asserts it validates AND narrows (web family exposed).

51 test_planner + 26 regression (#1989 strip / plan_runtime / plan_dispatch) + ruff + tier-audit --strict clean. Auto-merge on green. This was a real structural validator bug (not weak-model robustness) — well worth fixing now. The owner-axis (universal per-step focusing value → A-minimal keeps it) goes to the next owner digest for override.

PR #2006, opened 2026-06-21
fix(#2003): make toolrow flush-window test deterministic (xdist-load flake)

### LINKED ISSUE #2003: flaky: test_toolrow_flush_survives_clear loses flush-window precondition under xdist parallel load

tya5:
**[tui-coder]** — Fixes #2003. Root-fix the flush-window determinism (not a serial marker), #1980 assertion preserved.

## Root cause

`test_clear_removes_tool_row_in_flush_window` (from #1980) passes in isolation but its **precondition** — "row should still be mounted during the flush window" — depends on the **wall-clock** min-display window (`_TOOL_CALL_MIN_DISPLAY_S = 0.3s`, in `_inline_row_manager._flush_tool_call_row`) staying open between `complete_tool_call_row` and the assertion. Under CI's `-n auto` (pytest-xdist) a slowed worker lets the window close first (the deferred-flush timer fires, or `mounted_for_seconds()` already exceeds 0.3s → an *immediate* flush, no window at all) → precondition fails, intermittently gating unrelated PRs' pytest-3.12 jobs.

## Fix (root, not workaround)

Inject a large min-display window (`monkeypatch` the module constant to 3600s) so the flush window is **open for the whole ms-scale test regardless of worker speed**:
- `mount→complete` elapsed stays well under it → a *deferred* flush is always scheduled (never an immediate one);
- the 3600s timer cannot fire during the test → the row stays mounted at the precondition.

The flush-window precondition is now deterministic; the **clear()-removes-the-pending-row assertion (the #1980 contract) is unchanged** — `clear()` must still cancel the pending timer + remove the still-mounted row. A `serial`/`xdist_group` marker (mitigation, not root) is not used.

`monkeypatch` here targets a **config constant**, not a collaborator (no `unittest.mock`), so it stays within the testing policy.

## Verification (primary evidence)

| scenario | precondition (row mounted) |
|---|---|
| window=0.3 + simulated 0.5s slow-worker stall | **False** — the flake reproduced |
| window=3600 + same 0.5s stall | **True** — fix robust |

Directly simulating the slowed worker reproduces the exact failure and proves the fix handles it. pytest-xdist isn't installed locally (the flake is CI-only), so robustness is shown this way rather than via `-n auto`.

## Tier self-check

- **Tier 2**; no mocks (monkeypatch on a config constant); public-surface assert (`query(ToolCallRow)`); no format-pins.

## Test plan

- [x] `pytest tests/cli/test_toolrow_flush_survives_clear.py` — passed, 5x stable in isolation
- [x] Slow-worker robustness proof (table above) — flake reproduced at 0.3s, holds at 3600s
- [x] `ruff` clean · `tier_audit --strict` 0 violations · verified against fresh `origin/main` (563a1bba)
- [ ] (skipped — pytest-xdist not installed locally) `-n auto` repro; the slow-worker simulation covers the same mechanism deterministically

### DIFF
--- a/tests/cli/test_toolrow_flush_survives_clear.py
+++ b/tests/cli/test_toolrow_flush_survives_clear.py
@@ -2,16 +2,26 @@
 
 Stale-deferred-removal class (feedback_tui_deferred_timer_stale_removal_class).
 ``complete_tool_call_row`` POPS the row from ``_tool_call_rows`` and, if the row
-hasn't met its 0.3s minimum display time, schedules a deferred flush
-(``set_timer``) that later writes the row's lines + removes it.
+hasn't met its minimum display time, schedules a deferred flush (``set_timer``)
+that later writes the row's lines + removes it.
 
 Bug: during that flush window the row is popped-from-dict but still mounted, so
 ``clear()`` (which sweeps the dict + query-sweeps InterventionWidget, but NOT
-tool-call rows) misses it. A Ctrl+L within ~0.3s of a tool completing left a
-ghost tool row mounted AND let the pending timer write stale lines into the
-freshly-cleared log.
+tool-call rows) misses it. A Ctrl+L within the min-display window of a tool
+completing left a ghost tool row mounted AND let the pending timer write stale
+lines into the freshly-cleared log.
 
 Public surface only: the mounted ``ToolCallRow`` widgets in the DOM.
+
+#2003 determinism: the flush window is driven by ``_TOOL_CALL_MIN_DISPLAY_S``
+(default 0.3s) compared against the row's WALL-CLOCK mounted age. With the real
+0.3s window the "row still mounted during the flush window" precondition was a
+wall-clock race — under xdist parallel load a slowed worker let the window close
+(timer fired / elapsed already exceeded 0.3s) before the assertion, intermittently
+failing the precondition. We inject a large min-display window so the flush window
+is OPEN for the whole (ms-scale) test regardless of worker speed: the precondition
+holds deterministically and the clear()-removes-pending-row assertion (the #1980
+contract) is unchanged.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — **review PASS → APPROVE + merge (CI 5/5 green).** The #2003 flush-window flake — a **deterministic root-fix**, not the serial-marker mitigation.

- **Deterministic, not serial** ✓ — injects `_DETERMINISTIC_FLUSH_WINDOW_S = 3600.0` via `monkeypatch.setattr(_inline_row_manager, "_TOOL_CALL_MIN_DISPLAY_S", …)`, so the flush window is open for the entire ms-scale test regardless of worker speed. No `skipif`/`xdist_group`/serial marker (that would have been a mitigation; this fixes the wall-clock dependence itself).
- **Policy-compliant monkeypatch** ✓ — targets a **config constant** (`_TOOL_CALL_MIN_DISPLAY_S`), not a collaborator (mock-count 0). That's allowed; faking collaborators is what's banned.
- **#1980 assertion preserved** ✓ — still asserts the row is popped from the dict but **still mounted** during the open flush window, then removed by clear(). The behavior under test is unchanged; only the timing precondition is made deterministic.
- **Root-cause documented + primary-evidence verified** ✓ — the docstring explains the xdist slowed-worker wall-clock race; tui proved the mechanism by simulation (window=0.3 + stall → precondition FALSE = flake reproduced; window=3600 + stall → TRUE = robust), 5× stable.

## Tier rule check ✓
Tier-appropriate · no collaborator mocks · behavioral · assertion unchanged.

Merging — this clears the recurring flake that was intermittently gating #2005 and other PRs' 3.x jobs.

PR #2009, opened 2026-06-21
fix(#1950): heal RichLog resize corruption via debounced repaint-after-settle

### LINKED ISSUE #1950: bug(tui): rapid terminal resize during an agent reply drops characters from the rendered text
**[tui-coder]** — bug(tui): rapid terminal resize during an agent reply drops characters from the rendered text

## Summary
When the terminal is resized **rapidly** while an agent reply is being rendered, the **committed** reply text loses characters (data loss, not just stale wrapping). Found via real-terminal tmux e2e.

## Primary evidence (direct A/B, same prompt + model)
Prompt: `explain what a hash map is in 3 sentences then say DONE` (model: gemini-2.5-flash-lite via proxy).

**Stable size (control) — intact:**
> A hash map … stores key-value pairs. It **uses** a hash function … from which the desired value can be found. … average-case time complexity for insertion, **deletion,** and lookup operations …

**With a resize storm during the reply (10 resizes 50↔110 cols over ~1.5s) — corrupted:**
> … stores key-value pairs. It / a / hash function … insertion, **delet** / and / lookup operations …

`"uses"` and `"ion,"` are **dropped**. `"It a hash function"` / `"insertion, delet and lookup"` are ungrammatical → corruption, not model variance. The corruption **persists** in the committed scrollback (visible in a later stable-width re-render), so it is not a transient capture tear.

## Reachability / severity
- **Timing-dependent**: reproduced 1/2 storms (the race is hitting a resize mid-render).
- A **2-step** resize during streaming was **clean** — needs the rapid storm.
- Plausible real-world vector: **drag-resizing** the terminal window emits continuous rapid SIGWINCH, which could hit this during a reply. (Not yet confirmed as a user repro — flagged as the likely path.)

## Root cause — UNKNOWN (needs instrumentation; reyn-vs-Textual open)
- Inline text replies render via `conv.render_message` (the engine emits 0 `__stream_*__` for inline turns — see `app_outbox.py:402`), NOT the `StreamingRow` path. `StreamingRow.append` is a simple `_accumulated += text` and the markdown swap uses `full = _accumulated` (no drop there).
- Textual `RichLog` commits content at write-time width and does NOT re-wrap on resize — so an *already-written* line can't be corrupted by a later resize. That points to the corruption happening **at write time during the storm**, or to `render_message` using a re-rendering widget (Markdown) rather than a committed RichLog write.
- Next step: instrument `render_message` to log the exact text handed to the render surface vs. what lands, to determine whether reyn drops chars before Textual, or Textual drops them under rapid resize. Only then is the fix seam (reyn render path vs. an upstream Textual workaround) decidable.

## Repro
`reyn chat`, send a prompt that yields a multi-line reply, and `tmux resize-window` (or drag the window edge) rapidly during the reply. Compare against a stable-size control of the same prompt.

tya5:
**[tui-coder]** — Fixes #1950 (owner-GO'd: no upstream report → documented Textual workaround = fix).

## Problem

Drag-resizing the terminal during/after an agent reply emits a SIGWINCH storm; under it the Textual compositor's TTY writes drop characters, corrupting the **visible** RichLog text (e.g. `"It uses a hash function"` → `"It a hash function"`). Root-caused earlier: the render path is width-sweep-clean and `log.lines` stay byte-intact — the loss is in the compositor's TTY writes (a Textual issue, not reyn). The corruption persists because nothing re-marks the log dirty after the storm.

## Fix (canonical documented Textual workaround)

`ConversationView.on_resize` schedules a **debounced full repaint** (`refresh(repaint=True)`) of the log, fired **once after the storm settles** (when the TTY is calm), re-rendering from the intact `log.lines` — which overwrites the corrupted cells. Textual Discussion #3527's pattern (full repaint after resize); debounced via the established `set_timer`+`.stop()` idiom so a drag coalesces into one repaint. Display-only healing — reyn's data model is provably correct.

**Heal mechanism (compositor analysis, `_compositor.py`):** `refresh(repaint=True)` marks the RichLog region dirty → `render_partial_update` re-renders + writes those lines (there is no TTY-diff layer that would skip an "unchanged" region — a dirty region is always re-written) → corrupted cells overwritten. The earlier-verified intact `log.lines` guarantee the re-render is correct.

## Verification

- **Unit (Tier 2)**: a resize storm coalesces into exactly **one deferred repaint** (not per-event, not synchronous); the delegated real hook exercises the live `refresh(repaint=True)` without error.
- **Real-terminal tmux**: no-regression across **12+ resize storms** (content stays intact; the heal path doesn't disturb rendering).
- **Honest caveat**: the rare display glitch **did not reproduce on-demand this session** (timing/load-dependent, drag-only — the same rarity that made it a "rare display glitch"), so I could not capture a live corrupted→healed A/B. The heal is established by the **compositor analysis** + the **intact data model**, not a live before/after. (Earlier root-cause did reproduce it ~1/2; conditions differ.)

## Tier self-check

- **Tier 2**; no mocks (a real recording callable observes the SUT's own repaint hook — not a collaborator fake); behavioral asserts (a counter, not `len()==N`); no format-pins.

## Test plan

- [x] `pytest tests/cli/test_conv_resize_repaint_1950.py` — passed (coalescing + real-repaint exercised)
- [x] `pytest tests/cli/test_conv_pane_tool_call_lifecycle.py test_toolrow_flush_survives_clear.py test_tui_smoke.py` — 62 passed (no regression)
- [x] `ruff` clean · `tier_audit --strict` 0 violations
- [x] Real-terminal tmux: no-regression across 12+ storms; reply content intact after storm + debounce repaint
- [ ] (not obtained — rare glitch didn't reproduce on-demand) live corrupted→healed capture; heal established by compositor analysis + intact data model instead

### DIFF
--- a/src/reyn/interfaces/tui/widgets/conversation.py
+++ b/src/reyn/interfaces/tui/widgets/conversation.py
@@ -48,6 +48,7 @@ from rich.markdown import Markdown as RichMarkdown
 from rich.padding import Padding
 from rich.text import Text
 from textual.app import ComposeResult
+from textual.events import Resize
 from textual.widget import Widget
 from textual.widgets import RichLog, Static
 
@@ -112,6 +113,12 @@ _RECENT_REPLIES_MAX = 10
 # turn anchors below are drop-aware so even pathological sessions that DO
 # cross the boundary don't break Ctrl+P/N navigation.
 _RICHLOG_MAX_LINES = 20_000
+# #1950: debounce window for the resize-repaint heal. A rapid terminal-resize
+# (SIGWINCH) storm — e.g. drag-resizing the window — corrupts the RichLog's
+# rendered TTY output (dropped characters) via a Textual compositor issue; the
+# log.lines data stays intact. We repaint ONCE after the storm settles, not on
+# every event. Small enough to feel instant, large enough to coalesce a drag.
+_RESIZE_REPAINT_DEBOUNCE_S = 0.15
 # F-H: minimum visible duration for an inline ToolCallRow before the conv
 # pane flushes it into the RichLog scroll history + unmounts the live
 # widget. Cache hits / instant returns would otherwise mount + flush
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **review PASS → APPROVE (auto-merge on CI green).** The #1950 Textual resize workaround — the canonical documented pattern, with analysis-backed verify that is sufficient for this bug class.

- **Impl** ✓ — `on_resize` → debounced (0.15s) timer + `call_after_refresh` → a full RichLog repaint from the intact `log.lines` after the storm settles (a drag coalesces into ONE repaint). The documented Textual Discussion #3527 pattern. Owner-directed (no upstream; documented workaround = the fix).
- **Verify = analysis-backed, SUFFICIENT (no live A/B demo required)** — and tui flagged this gap honestly rather than overstating (exactly right). Rationale: (a) the corruption is unreproducible on-demand (0/12+ storms — a rare drag-only timing race); (b) the fix is **low-risk asymmetric-positive** (worst case a no-op; tmux no-regression confirmed across 12+ storms); (c) the heal mechanism is analytically sound — a debounced repaint *after the storm settles* re-writes the dirty RichLog region from proven-intact `log.lines`, with the compositor stable post-storm; (d) canonical pattern. A live corrupted→healed demo of an unreproducible bug is diminishing-returns.
- **PR honestly records the verify basis** (analysis + no-regression, not a live A/B) — keep that. Ongoing validation = recurrence-monitoring (if the rare glitch is reported post-fix, revisit).

## Tier rule check ✓
- Tier-2 · no mocks (real hook delegate) · the unit asserts the debounce coalesces a storm into exactly one deferred repaint; tmux no-regression is the integration axis.

Auto-merge on green. Good transparency — the honest "verify gap" flag is how this should be reported.

Issue #927, opened 2026-05-25
Flaky CI: test_abort_tool_call_rows_seals_live_rows_with_aborted_terminal (test_conv_pane_tool_call_lifecycle.py)


tya5:
**[lead-coder]** — Flaky CI test, sustained N=6 instance across PRs

## Symptom
`tests/cli/test_conv_pane_tool_call_lifecycle.py::test_abort_tool_call_rows_seals_live_rows_with_aborted_terminal` 

CI で `AssertionError: assert 2 == 0` で fail。 Python 3.11 or 3.12 (= 不定) でランダム fail、 同 test の re-run で pass = 既知 flake。

### N=6 instance tally
- sandbox_2 #839 (earlier session)
- tui-coder #886 (root cause #2 path)
- tui-coder #898 (same file 別 test = sister)
- tui-coder #910 (same file)
- e2e-coder #925 (today)
- sandbox_2 #926 (today)

### lead-coder env reproducibility
- main HEAD で N=3 直接 reproduce → **3/3 pass** (= lead-coder env env (= Python 3.12.7)、 race / state pollution 検出されず)
- CI env でのみ fail (= GitHub Actions runner specific timing / parallelism)

= **CI env-specific flake**、 PR change 由来でない、 re-run で recover 可能

## Hypothesis (= 未 verify、 author 検討 ask)
- async timing race in `_abort_tool_call_rows` 経路 + parallel test execution の interference
- 同 file 内別 test (= `test_complete_tool_call_row_unmounts_live_widget`) も同 flake 系 = file-wide race shape の suggestion

## Workaround (current practice)
- lead-coder side で main N=3 reproduce + re-run trigger で merge route 確保
- B57 dispatch 経由の dogfood 経路では non-blocking (= CI のみ影響)

## Ask
- tui-coder area (= TUI conv pane test): 真因 triage + 修正方針 (例: pytest-rerunfailures decorator / async timing wait 強化 / fixture isolation)
- 短期 mitigation で sufficient なら issue keep open、 long-term root fix 別 wave で OK

tya5:
**[flake-guardian]** — Phase B root fix verified. Closing.

## Post-fix CI verification

Phase B root fix (PR #932, merged 2026-05-27T09:10) 後の CI run: **19/19 success** (Python 3.11 + 3.12 含む)。
最後の flake instance: 2026-05-25T14:36 (= fix 前、 通算 N=8 instance 相当)。

### Primary evidence
[...]

PR #2010, opened 2026-06-21
test(#2003-class): harden async-stack flash-window tests against xdist flake

tya5:
**[tui-coder]** — Sibling-harden of the #2003 wall-clock-window flake class (lead-directed follow-up to the toolrow fix).

## Why

The AsyncStackPanel interrupt-flash tests assert "row still in the flash window" after a fixed `pilot.pause`, relying on the wall-clock `_FLASH_DURATION_S` (1.5s) staying open. Under CI's `-n auto` (pytest-xdist) a slowed worker can let the window close before the assertion — the **same timing race as #2003**. The 1.5s window is a larger margin than #2003's 0.3s (lower probability), but the root is identical wall-clock dependence; this makes it **class-complete deterministic** rather than margin-reliant (per lead: "margin 頼みにしない").

## Fix (same shape as #2003, per assertion direction)

Monkeypatch the module constant `_FLASH_DURATION_S` (read dynamically at `remove()` time); behaviour assertions unchanged:

| assertion | harden |
|---|---|
| "still-present **in** window" (interrupted: present-before-flush, elapsed-frozen; readd: returns-to-running) | inject a **LARGE** window (3600s) → can't close during the ms-scale test |
| "gone **after** window" (interrupted: gone-after-flush) | inject a **FAST** window + **poll** for the removal (deterministic vs a fixed pause-past that races the removal task — the #927 lesson) |
| "survives the stale timer" (readd: survives-old-timer) | **synchronous re-add** (cancels the timer before any await lets it fire) + a FAST window + poll for a buggy stale removal |

## Proof the injection fixes the flake (mirrors #2003)

```
window=1.5  + 1.7s simulated slow-worker stall → row present=False  (flake reproduced)
window=3600 + 1.7s stall                       → row present=True   (harden robust)
```

So the large-window injection demonstrably keeps the row present past the old deadline (and the monkeypatch reaches the panel's `set_timer`).

## Tier self-check

- **Tier 2**; no mocks (monkeypatch a config constant + a poll helper, not a collaborator fake); behavioural asserts; no format-pins.

## Test plan

- [x] `pytest tests/cli/test_async_stack_interrupted_flash.py tests/cli/test_async_stack_readd_during_flash.py` — 9 passed, **3x stable**
- [x] Robustness proof (table above) — flake reproduced at 1.5s, robust at 3600s
- [x] `ruff` clean · `tier_audit --strict` 0 violations
- [ ] (skipped — pytest-xdist not installed locally) `-n auto` repro; the slow-worker simulation covers the same mechanism deterministically

### DIFF
--- a/tests/cli/test_async_stack_interrupted_flash.py
+++ b/tests/cli/test_async_stack_interrupted_flash.py
@@ -47,6 +47,36 @@ def _panel_row(snap: list[dict], agent_id: str) -> dict | None:
     return None
 
 
+# #2003-class flake-harden: the flash window is wall-clock-driven
+# (``_FLASH_DURATION_S``), so a "row still in the flash window" precondition was a
+# timing race under xdist parallel load — a slowed worker let the window close
+# before the assertion. Drive the window deterministically (same shape as #2003,
+# tests/cli/test_toolrow_flush_survives_clear.py); flash-contract assertions are
+# unchanged. Inject a LARGE window for "still-present" tests (stays open the whole
+# ms-scale test) and a FAST window + a poll for "gone-after" tests (closes
+# promptly, then wait for the removal to land rather than a fixed pause-past).
+_FLASH_OPEN_S = 3600.0
+_FLASH_FAST_S = 0.05
+
+
+def _patch_flash(monkeypatch, seconds: float) -> None:
+    import reyn.interfaces.tui.widgets.async_stack_panel as _asp
+    monkeypatch.setattr(_asp, "_FLASH_DURATION_S", seconds)
+
+
+async def _wait_until(pilot, predicate, *, cap: float = 2.0, step: float = 0.02) -> bool:
+    """Poll ``predicate`` (yielding the loop) until true or ``cap`` — a
+    deterministic wait for a timer-driven transition (vs a fixed pause that races
+    the removal task under load). Returns the final predicate value."""
+    waited = 0.0
+    while waited < cap:
+        if predicate():
+            return True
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **review PASS → APPROVE (auto-merge on CI green).** The class-complete sibling-harden for the wall-clock-window flake class — and it correctly applies a *different* deterministic shape per assertion direction.

- **still-present-in-window** (interrupted-present-before-flush / elapsed-frozen) → `_FLASH_OPEN_S = 3600.0` injected via `monkeypatch.setattr(_asp, "_FLASH_DURATION_S", …)` — the window can't close during the ms-scale test.
- **gone-after-window** (interrupted-gone-after-flush) → FAST window + `_poll` past it for the removal — deterministic vs a fixed `pause(past)` race (the #927 lesson, correctly applied).
- **survives-stale-timer** (readd) → synchronous re-add (cancel the timer before it fires) + FAST + poll.
- **Robustness proven** (same as #2003): window=1.5 + 1.7s stall → row gone (flake reproduced) / window=3600 + same stall → survive (robust). 9 tests, 3× stable.

## Tier rule check ✓
- config-constant monkeypatch (not a collaborator mock; mock-count 0) · no `serial`/`xdist_group` marker (deterministic root-fix, not mitigation) · the flash-contract assertions are preserved (only the timing precondition is made deterministic).

Auto-merge on green. This closes the wall-clock-window flake class — both #2003 siblings hardened to determinism, not margin. Clean completion of the completeness sweep.

PR #2011, opened 2026-06-21
feat(#1953): slice P3 — router-expose decompose + byte-equal engine parity

tya5:
**[e2e-coder]** — #1953 slice P3: router-expose + byte-equal engine parity (coexist with plan)

## What
Exposes the task-driven decomposition engine as an LLM-facing router tool (`decompose`), **in parallel with `plan`** (plan untouched), and proves the two execution engines are equivalent — the structural half of the slice-P → P4-delete gate.

## Changes
1. **Topo-order parity-by-construction** (folds the Q3 finding): `run_task_graph` now executes AND aggregates units in the plan executor's exact Kahn `_topological_order` (`_topo_order_tasks`, stable creation-order tie-break) instead of the readiness-loop's execution order. The Task engine's per-unit LLM-response consumption order + final-reply aggregation are now byte-identical to the plan path **by construction**, not by coincidence.
2. **`decompose` router tool** (`tools/decompose.py`): `ToolDefinition` (router=allow/phase=deny, async), schema mirrors `plan` (goal + `steps_json`) so one decomposition validates against both engines. Handler → `ctx.router_state.dispatch_task_tool`.
3. **Dispatch wiring**: `RouterCallerState.dispatch_task_tool` + `_build_router_caller_state` binding (session state pre-bound, same posture as `dispatch_plan_tool`; backend via the new public `task.resolve_task_backend`). `dispatch_task_tool` reuses plan's `parse_and_validate_plan` (structural + #1998 tool-vocabulary validation — P4 repoints it) → `build_task_graph` → run via `make_production_run_unit` → synthesized reply to outbox. Lighter async: `host.spawn_task_graph` hook + sync fallback; wires the slice-8 `record_task_cost` cap-charge in the live path.
4. `decompose` classified `_NOT_EXTERNAL` (#1822 exhaustive trust gate).

## The slice-P delete gate (BOTH required before P4 deletes plan)
- **(a) deterministic byte-equal — DONE here** (`test_sliceP3_plan_task_byte_equal_parity`): on a fixed diamond-dependency decomposition, `execute_plan` and `run_task_graph` are driven with the SAME monkeypatched scripted `call_llm_tools` (shared by both engines' RouterLoops) and asserted byte-equal on every per-unit text + the synthesized reply. The behavioral analog of P1's byte-identical MOVE.
[...]

PR #2012, opened 2026-06-21
fix(#1953): wire per-unit step system prompt — result-channel SP-injection (b)-caught

tya5:
**[e2e-coder]** — #1953: wire the per-unit step system prompt (result-channel SP-injection) — the bug (b) live-parity caught

## The dual-proof in action
This is the gap the **(b) live dogfood** half caught that **(a) byte-equal** structurally could not — the decisive evidence that lead's dual-proof ruling was right.

Running the SAME fixed decomposition through both engines live (gemini), diamond DAG, s3 = "combine the prior two facts":
- **plan** s3 → *"The number 7 is a prime number, and eight is the smallest cube…"* — combined the prior facts ✓
- **task** s3 → *"I need more info. Please tell me what the prior two facts are."* — ran blind to its inputs ✗

(a) could not see this: the scripted LLM keyed on the user message and ignored the system prompt; only a real LLM reads the SP.

## Root cause
`make_production_run_unit` built each unit's `RouterLoop` **without** a `system_prompt_override`, so every unit ran on the generic router prompt — missing the goal framing AND the dependency-results channel. The I-2 result-channel was wired for **storage** (`set_result` + backend read) but the **SP-injection** half (rendering deps' results into the unit's LLM context) was never wired. The plan path builds this SP externally via `build_plan_step_system_prompt(plan, step, prior_results)` and passes it as the override; the task path skipped it.

## Fix
`make_production_run_unit` now builds the per-unit step SP via `build_plan_step_system_prompt` (a synthetic `Plan`/`PlanStep` adapts the Task; its `depends_on` = the unit's dep ids, which key `prior_results`, so the "## Prior step results" section renders) + passes `system_prompt_override` — byte-matching the plan path. `goal` is threaded in (the shared decomposition goal). P4 repoints the builder onto the Task system.

**Live re-run after fix**: both engines combine the prior facts. ✓

## Test plan
- [x] Regression test (deterministic — the gap (a) had): a dependent unit's SP carries the goal + "Prior step results" section + the dep's actual result; a no-dep unit gets the goal but no prior-results section.
- [x] (a) byte-equal updated to thread `goal` (SP fidelity) — still byte-equal.
- [x] P2+P3 suite (14 tests) green; consumer suite 481 passed (goal param is backward-compat default).
- [x] ruff full-tree clean; tier-audit clean; verify_module_docstrings clean.
- [ ] (b) N≥3 live dogfood parity re-run (now that the channel works) — reported next on #1953 / this PR.

## Context
P3 PR #2011 (coexist/additive, plan untouched) merged before this was found, so no regression to existing flows — decompose is a new, not-yet-default tool. This is a P3 correctness fix, not P4 scope. A second parity gap (TUI-surface progress-row, `source=="plan*"`-specific) is owned by tui-coder; both gaps were surfaced to owner via lead.

### DIFF
--- a/src/reyn/runtime/task_graph.py
+++ b/src/reyn/runtime/task_graph.py
@@ -206,6 +206,7 @@ def make_production_run_unit(
     chain_id: str,
     router_model: "str | None",
     budget: Any,
+    goal: str = "",
     exclude_tools: "frozenset[str]" = frozenset({"plan"}),
 ) -> RunUnit:
     """Build the production ``run_unit``: each unit runs through a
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **review PASS → APPROVE (auto-merge on CI green).** The SP-injection fix — the task engine now builds the per-step system prompt exactly as plan does, closing the gap the (b) live parity exposed.

- **Fix** ✓ — `make_production_run_unit` builds the per-unit step SP via `build_plan_step_system_prompt(Plan(goal=goal, steps=(_step,)), _step, prior_results, …)` (a synthetic Plan/PlanStep adapts the Task; **P4 repoints**) so the goal-framing AND the "## Prior step results" dep-results section render, passed as `system_prompt_override` — byte-matching the plan path. The result-channel is now *injected*, not just stored: dependent steps see their deps' actual results.
- **Regression test** ✓ — `test_dependent_unit_sp_carries_goal_and_prior_results` asserts the dependent unit's SP carries goal + "Prior step results" + the dep's actual result. This is the **deterministic** catch of the (b)-found bug — (a) would now catch it too.
- Live re-run confirms both engines combine prior facts. 14 P2/P3 + 481 consumer tests green; ruff full-tree + audits clean.

**Methodology pin = correct + valuable.** (a) deterministic-replay has a STRUCTURAL blind spot: a scripted LLM keyed on the user message can't see SP-content bugs (goal/context injection) — it proves engine *wiring* equivalence for what the fake exercises, NOT that the real LLM receives equivalent *context*. The live (b) half is non-optional for context-injection parity. Exactly why the gate is dual.

## Tier rule check ✓
- regression test is behavioral + deterministic (asserts SP content), no mock for the SP-build.

Auto-merge on green. After it lands: re-run (b) N≥3 (channel now works) → **(a) + (b) + owner-cross-check = the P4 delete-gate.**

Issue #2027, opened 2026-06-21
[FP-0050/security] task-op results surface cross-session descriptions — query-path fencing (GATES the dynamic-wire wakes)


tya5:
**[lead-coder]** — found during the #1953 dynamic-wire(A) item-1 (#2026) review; verified against the task model + op results.

## Finding
The task ops (create/update_status/get/list/comment/abort/…) return `task.to_dict()` which INCLUDES `description` + `name` (model.py:119-124), NOT just id/status/deps. For a CROSS-SESSION task (created by another session), the description is PEER-AUTHORED (untrusted). So task.get/list/comment (and the to_dict-returners) surface peer-authored content into THIS session’s LLM context.

The dynamic-wire fences cross-session content at the WAKE seam (items 4-5). But the **task.get/list/comment QUERY path is a SEPARATE path that BYPASSES the wake-fencing** (a session can task.get a cross-session description unfenced, re-surfacing what the wake fenced). The wake-seam fencing is INSUFFICIENT for the query path.

## The trust-boundary decision (FP-0050 — owner call)
- **(A) Conservative per-op `_EXTERNAL`** for the surfacing ops: simple, closes the gap, but OVER-FENCES self-tasks (the LLM treats its OWN task records as external — a UX cost).
- **(B) Dynamic per-result fence** (fence description/comments only when creator != this session = cross-session): precise, no over-fencing, but a more complex per-result mechanism (not the static per-op classification).
- **(C) Result-shape change**: task-op results return id/status/deps WITHOUT the description (description is wake-only, fenced at the wake): precise but changes the op result shape.

Lead lean: (B) or (C) (precise, no over-fencing) — but a genuine security-UX trade-off; owner picks.

## Gate
The cross-session exposure is GATED on real cross-session tasks, which require the dynamic-wire WAKES (items 4-5) + delegation. So item-1 (#2026, the expose) is SAFE to land (the gap is dormant — no cross-session linkage without the wakes). But this decision MUST resolve BEFORE the wakes PR (items 4-5) lands — that PR is GATED on this.

Severity: HIGH (a trust-boundary content-fencing gap that becomes live with the wakes). Owner decision before the wakes PR.

tya5:
**[lead-coder]** — Wakes-design REQUIREMENT (owner-raised): the execution-path framing must not let the fence block legitimate delegated work.

## The requirement
[...]

PR #2043, opened 2026-06-21
docs(security): document structural-fence scope / targets

tya5:
**[docs-maintainer]** — Owner-requested (via lead-coder, 2026-06-22): make the fence **SCOPE / TARGETS** visible. `reyn.yaml` documented the `fence_enabled` *flag* but not *which* content the fence applies to.

## What
- **Security concept doc** — new `#### What gets structurally fenced` subsection: a table of the Class-A fence targets + their source-trust rule, plus the **two-gate model** (config gate `enabled`+`fence_enabled`; per-seam source-trust gate).
- **reyn-yaml.md** — the `fence_enabled` row now cross-links to that subsection (flag → targets).
- **security.ja.md** — translated mirror (see "JA gap" below).

## Verified targets (grep, not trust)
Per lead's instruction *"don't trust my list blindly — verify by grepping the call-sites"*, every target was confirmed against origin/main call-sites (`content_guard.fence_if_enabled`, `_fence_text` / `_fence_view`, `fence_tool_result`, `_fence_inbound`, and the per-tool `returns_external_content` flag):

| Target | Seam | Gate |
|---|---|---|
| External-content tool results (web fetch/search, MCP, **recall/RAG, memory reads**) | router tool-result chokepoint → `fence_tool_result` | `returns_external_content` tools only; trusted-internal = **scan-only** |
| Project context file (REYN.md/AGENTS.md/`project_context_path`) | router_host_adapter | always |
| A2A inbound peer messages | `a2a_handler._fence_inbound` | always |
| External intervention answers | `intervention_handler` | history-bound copy only |
| Task query results (`description`/`name`/`result`) | `task._fence_view` | always; structural fields not fenced |
| Delegated-task wake descriptions | `task._fence_text` → `task_wake` | always (OS execute-framing trusted) |

## Two corrections the verification forced
1. **lead's list said "Recalled memory entries" as a distinct target** — it's *not* a separate seam; memory **reads** are fenced as external-content tool results (recall/memory tools carry `returns_external_content=True`), while memory **writes** are pattern-**blocked** (different mechanism). Documented both directions.
2. **The existing concept-doc seam list said "Both [scan AND fence] apply"** — inaccurate: memory-writes / pre-exec / compaction are block / scan / redact, *not* fence, and tool-result fencing is external-content-only. Reworded each seam's mechanism.
3. **NOT on lead's list, found by grep:** delegated-task **wake descriptions** (the execution-path counterpart to the #2027 query-path fence) — added. The #2027 task-op query-results fence has **landed**, so documented as present (not "coming").

## JA gap closed
`security.ja.md` was missing the *entire* Content-layer-defense section, and its "まだ薄い部分" still claimed no prompt-injection defense exists (pre-shipping drift, now self-contradicting). Added the translated section + corrected the stale paragraph.

## Adjacent finding (not fixed here — flagging for your call)
`reyn-yaml.md:457` threat_scan intro carries an inline `(FP-0050 / #1822)` history ref in an operator/reference doc (pre-existing, not mine). Left as-is to keep this PR scoped; happy to sweep separately.

## Test plan
- [x] Every target verified against origin/main call-sites (grep of fence_if_enabled / _fence_* / returns_external_content)
- [x] 0 history refs (FP/PR/#issue) in the new fence-target content
- [x] Cross-link anchor `#what-gets-structurally-fenced` resolves (no broken-anchor warning under strict)
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — EXIT=0, no WARNING/Aborted


tya5:
**[lead-coder]** — APPROVE + merging (--admin). Fence scope/targets doc (owner-requested). grep-verified, corrected my list 3× (memory read-fence vs write-block; per-seam scan/fence/block/redact; +#1953 wake descriptions); cross-checked vs impl — accurate. Two-gate model + NOT-fenced boundaries + JA mirror fix. (Re-confirmed green after pytest settled.)

PR #2048, opened 2026-06-22
fix(tui): degrade malformed intervention choice instead of dropping it

tya5:
**[tui-coder]** — fix from the 3rd mining sweep (message-routing / handler-completeness class).

## The defect (code↔comment inconsistency in a load-bearing path)
`OutboxRouter._on_intervention` normalizes choice dicts, and its **own comment** says *"Missing keys fall back to safe blanks"* — but it read `c["label"]` / `c["id"]` with **bracket access** (only `hotkey`/`default` used `.get()`). A choice missing `label`/`id` raised `KeyError`, contradicting the documented intent.

**Why it matters**: interventions are **load-bearing** (the user must see + answer a permission request). The outbox consume loop wraps each handler in `try/except`, so a `KeyError` here is *swallowed* → the whole intervention is silently **dropped** → the user can't respond (stuck). Aligning `label`/`id` with `.get()` degrades a malformed choice gracefully and keeps the intervention visible.

## Test plan
- [x] +1 Tier-2 test: an intervention whose first choice has no `label`/`id` still mounts (real `ReynTUIApp` + `OutboxRouter`, no mocks)
- [x] Falsify Python-semantics-certain: pre-fix `{"hotkey":"y"}["label"]` → `KeyError` on the direct call
- [x] 4 intervention tests pass (degrade + resolved-meta no-regression); `ruff check src tests` clean; `tier_audit --strict` 0

## Sweep context
The message-routing layer was otherwise **clean** — producer↔consumer kind-sets match (the dispatch is complete), handlers + the consume loop are guarded. This code↔comment inconsistency was the one actionable find. (Two non-defects correctly left alone: the `__session_switch_request__` kind is registry-owned before the TUI loop; the dead `__stream_*` handlers are an architectural remnant with a documenting comment.)

### DIFF
--- a/src/reyn/interfaces/tui/app_outbox.py
+++ b/src/reyn/interfaces/tui/app_outbox.py
@@ -1141,11 +1141,15 @@ class OutboxRouter:
             # Forward the full choice shape (label / id / hotkey / default)
             # so the widget can render `[h] Label` prefixes and highlight a
             # default option. Missing keys fall back to safe blanks — callers
-            # that historically only set label+id keep working.
+            # that historically only set label+id keep working. Interventions
+            # are load-bearing (the user must see + answer them), so a
+            # malformed choice degrades gracefully here rather than raising —
+            # a KeyError would drop the whole intervention (the handler's
+            # try/except swallows it), leaving the user unable to respond.
             choices = [
                 {
-                    "label": c["label"],
-                    "id": c["id"],
+                    "label": c.get("label", "?"),
+                    "id": c.get("id", ""),
                     "hotkey": c.get("hotkey") or "",
                     "default": bool(c.get("default", False)),
                 }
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — APPROVE + merging (--admin). Intervention-drop fix: choice-normalization read c["label"]/c["id"] with bracket access despite its comment promising safe-blank fallback → KeyError swallowed by the consume loop → the intervention SILENTLY DROPPED (user can't respond). Fixed → .get(). A code↔comment inconsistency in a load-bearing, failure-invisible path. + Tier-2 falsify test. CI green.

## Tier rule check ✓
Tier-2, falsify-confirmed; behavior assertion (intervention survives missing keys); no mocks/format-pin/private-state.

PR #2056, opened 2026-06-22
feat(1800-4a): wake-flag inbox drain primitive + equivalence test

tya5:
**[per-PR-coder]** — #1800 slice 4a: `wake`-flag inbox drain primitive behind an equivalence safety test.

## Summary

- Adds `_drain_to_wake()` to `session.py` (new method after `_consume_inbox`): drains the inbox up to and including the first `wake=true` message, accumulating `wake=false` ride-alongs. Implements Decision A (re-enter blocking wait if only wake=false remain, no trigger yet).
- Wires `run_one_iteration` to call `_drain_to_wake()` instead of `_consume_inbox()` directly. Unpacks `(ride_alongs, trigger)` — leaves a clear `TODO(#1800-4b)` seam for ride-along staging.
- The `wake` field lives in the inbox payload (default `True` when absent). All existing producers (submit_user_text / skill_completed / task_ready / etc.) never set it — they are effectively wake=true, so the common path is a single blocking get with no ride-alongs, byte-identical to before.
- Per-message `inbox_consume` is recorded via `_consume_inbox` (or the journal directly for non-blocking dequeues) for every drained message, so the snapshot stays correct on crash+restore.

## Equivalence preservation

With no `wake=false` ever enqueued, `_drain_to_wake` reduces to:
```
kind0, p0 = await self._consume_inbox()   # blocking — same as before
# p0.get("wake", True) → True (absent or explicit True)
return [], (kind0, p0)                    # no ride-alongs
```
The caller unpacks `kind, payload = trigger` and dispatches identically. The Tier 2b equivalence tests verify this at the WAL level: N messages → N inbox_put + N inbox_consume events, msg_ids match, snapshot pruned to empty.

## Files changed

- `src/reyn/runtime/session.py`: `_drain_to_wake()` added (~85 lines), `run_one_iteration` wired to use it.
- `tests/test_1800_wake_drain.py`: 8 tests (2 Tier 2b equivalence + 6 Tier 1 contract).

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_1800_wake_drain.py` — 8/8 OK, 0 errors
- [x] `python -m pytest tests/test_1800_wake_drain.py -v` — 8 passed
- [x] `python -m pytest tests/ -q` (full suite, excluding pre-existing fastapi-import failures) — 7806 passed, pre-existing failures unchanged

Refs #1800

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -2801,6 +2801,93 @@ class Session:
             await self._journal.consume_inbox(msg_id=msg_id)
         return kind, payload
 
+    async def _drain_to_wake(
+        self,
+    ) -> tuple[list[tuple[str, dict]], tuple[str, dict]] | tuple[None, None]:
+        """Drain the inbox up to and including the first ``wake=true`` message.
+
+        Each inbox payload carries an optional ``wake`` bool (default ``True``
+        when absent).  Existing producers (user / skill_completed / task_ready
+        / etc.) never set ``wake``; the absent-means-True default makes them
+        all behaviorally identical to wake=true, so the common/back-compat
+        path returns immediately after the first blocking get with no
+        ride-alongs.
+
+        Returns ``(ride_alongs, trigger)`` where:
+
+        - ``ride_alongs``  — list of ``(kind, payload)`` tuples for every
+          ``wake=false`` message drained before the trigger.  Staged for the
+          next turn as context (slice 4b — see TODO below).  Empty in the
+          common case.
+        - ``trigger``      — the first ``wake=true`` (or absent-wake) message;
+          this drives the turn.
+
+        Special case: if the first blocking get yields ``shutdown``, returns
+        ``(None, None)`` so the caller can signal loop exit.
+
+        Decision A (RESOLVED, issuecomment-4773744053): if the queue empties
+        while holding only ``wake=false`` ride-alongs (no trigger yet),
+        re-enter the blocking wait.  Ride-alongs NEVER trigger a turn alone.
+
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Review (CLOSE, run-loop change): **APPROVED**, merging on CI green.

Verified `_drain_to_wake` directly (session.py:2804-2890):
- **Equivalence preserved:** `wake` defaults True; the common path (no wake=false) returns after the first blocking `_consume_inbox` with empty ride-alongs → behaviorally identical to the old single-get. `run_one_iteration` (2991-3009) dispatches the trigger exactly as before.
- **`inbox_consume` accounting (load-bearing):** recorded for EVERY drained message — outer via `_consume_inbox`, inner `get_nowait` drain via an explicit `consume_inbox` call (2876-2881, before the wake/shutdown checks) → no snapshot leak on crash+restore.
- **Decision A:** queue empties holding only wake=false → `QueueEmpty` breaks the inner loop → re-enters the outer blocking wait (never a turn on C alone).
- **Shutdown:** handled in both outer and inner loops → `(None, None)` sentinel.

Tests (8, real instances, no mocks): 2 **Tier 2b equivalence** (WAL identical to pre-4a — falsifiable by any drain bug) + 6 Tier 1 contract (wake-true-first, absent-default, [F,F,T], shutdown, per-message inbox_consume, Decision A). Clean 2-file set. `TODO(#1800-4b)` seam for ride-along staging.

**## Tier rule check ✓** — Tier 2b/1 declared; 0 Tier-4 (audit); WAL-invariant + contract assertions (no private-state / no algorithm pins); tier-audit green.

PR #2058, opened 2026-06-23
feat(1800-5a): lifecycle audit events — session/turn started/completed

tya5:
**[per-PR-coder]** — #1800 slice 5a: P6 lifecycle audit events (additive). Builds on merged 4b.

## Summary

- `session_started` / `session_completed` emitted in `Session.run()` alongside `chat_started` / `chat_stopped`
- `turn_started` emitted in `run_one_iteration()` after the trigger is consumed from the inbox, before dispatch
- `turn_completed` emitted in `_run_router_loop()` immediately after `RouterLoopDriver.run_turn()` returns (= terminal stop_reason point — see decision C in #1800 issuecomment-4773744053)
- All four events registered in `EVENT_AUDIT_REQUIREMENTS` (FP-0021 audit-completeness invariant)
- Four tests: Tier 1 (schema) + Tier 2 (emit at correct point), real Session + real EventLog, no mocks

No behavior change beyond the new events (additive). Hook dispatch is slice 5b.

## `turn_completed` emit point

Emitted in `Session._run_router_loop()` at line ~4899, immediately after `await self._loop_driver.run_turn(user_text, chain_id)` returns. This point is chosen because:

1. `run_turn()` wraps `RouterLoop.run()` which returns only when the loop reaches a terminal condition (text reply, empty-stop, async-spawn, limit-denied — all terminal paths `return self._total_usage`)
2. Emitting here means exactly ONE `turn_completed` per turn, independent of which terminal path fired
3. `chain_id` is in scope at this layer (passed into `_run_router_loop`)
4. The `finally`-based `_run_router_loop` approach avoids needing to instrument each individual terminal inside `RouterLoop.run_loop()` (which would risk double-emit or missed paths)

This matches the RESOLVED DECISIONS: "turn_end = the agent reaches a terminal stop_reason (the turn's response is complete) — ONE condition, independent of skill/routing."

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_session_lifecycle_events_1800.py` — 4/4 OK
- [x] `pytest tests/test_session_lifecycle_events_1800.py` — 4/4 pass
- [x] `pytest tests/test_event_audit_invariants.py tests/test_chat_turn_completed_inline.py` — all pass (existing invariant tests unaffected)
- [x] Full `pytest --continue-on-collection-errors` — 7834 passed, failures are pre-existing optional-dep (fastapi/uvicorn/mcp) confirmed against origin/main baseline

Ref: #1800

### DIFF
--- a/src/reyn/core/events/event_schema.py
+++ b/src/reyn/core/events/event_schema.py
@@ -62,4 +62,26 @@ EVENT_AUDIT_REQUIREMENTS: dict[str, frozenset[str]] = {
     #   tool_calls batch produced spawn-ack (= usually 1).
     # chain_id: for cross-agent tracing (P6)
     "invoke_skill_spawn_ack_exit": frozenset({"chain_id", "spawn_ack_count"}),
+    # #1800 slice 5a: session + turn lifecycle events (P6 audit — hook dispatch
+    # points added in slice 5b).
+    #
+    # session_started / session_completed: emitted in Session.run() alongside
+    #   chat_started / chat_stopped. Marks the boundary of the session's
+    #   resource scope (F in the lifecycle hook design).
+    # agent_name: the session's agent identity (same field as chat_started).
+    "session_started": frozenset({"agent_name"}),
+    "session_completed": frozenset({"agent_name"}),
+    # turn_started: emitted in Session.run_one_iteration() after the trigger
+    #   is consumed from the inbox and before dispatch to _handle_*.
+    # kind: the inbox message kind that triggered this turn (e.g. "user",
+    #   "skill_completed", "task_ready"). Lets subscribers distinguish human
+    #   triggers from automated ones without parsing the payload.
+    "turn_started": frozenset({"kind"}),
+    # turn_completed: emitted in Session._run_router_loop() immediately after
+    #   RouterLoopDriver.run_turn() returns — the router loop has reached a
+    #   terminal condition (the turn's response is complete). One emit per
+    #   turn, independent of skill/routing path. This is the hook point for
+    #   the turn_end lifecycle hook (slice 5b). chain_id matches the turn's
+    #   chain_id for cross-agent tracing.
+    "turn_completed": frozenset({"chain_id"}),
 }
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — APPROVE (static review). Additive P6 audit slice; merging on pytest-3.12 green (3.11 already pass).

**Verified:**
- **File set**: 3 files, purely additive — `event_schema.py` (+22), `session.py` (+24 emit calls), `test_session_lifecycle_events_1800.py` (+256). No run-loop behavior change.
- **All 4 emits placed at the right lifecycle points** (= the exact attach points 5b will hook):
  - `turn_started` (`run_one_iteration`, after trigger consumed / before dispatch) — `kind` + optional `chain_id`.
  - `session_started` (`run()`, alongside `chat_started`) — `agent_name`.
  - `session_completed` (`run()` **finally**, alongside `chat_stopped`) — `agent_name`.
  - `turn_completed` (`_run_router_loop`, immediately after `run_turn()` returns = **terminal stop_reason**, once per turn independent of routing path) — `chain_id`. This is the `turn_end` hook point.
- **Schema**: all 4 registered in `EVENT_AUDIT_REQUIREMENTS` with correct field sets → FP-0021 audit-completeness invariant satisfied (Test 1 guards it).

## Tier rule check ✓
1. **Docstring Tier declared** — Test 1 `Tier 1` (schema), Tests 2–4 `Tier 2` (real Session + EventLog).
2. **No Tier-4 format pin** — uses the unpack-enforcement idiom `(ev,) = events` (not `len(...) == N`); ordering asserted via `.index()` / event-count, not whitespace/format.
3. **OS-invariant coverage** — new events wired into `EVENT_AUDIT_REQUIREMENTS`; no skill-specific strings; Test 4's ordering guard genuinely fails if `turn_completed` fires before `run_turn` returns (not tautological).
4. **Tier-audit CI** — `test-tier audit` job PASS.

**Test policy** — no MagicMock/AsyncMock/patch; only the LLM boundary (`_loop_driver.run_turn`) faked as a plain async noop (allowed Tier 2c); events read through the public `add_subscriber` surface.

These 4 emit points are the lifecycle seams the **awaited HookDispatcher (5b)** attaches to. Merging 5a on green, then 5b is next.

PR #2061, opened 2026-06-23
feat(1800-7): the hook-driven-turn loop valve (bound self-continuation)

tya5:
**[e2e-coder]** — #1800 slice 7: the loop valve. Re-sequenced before slice 6 because my 5b shipped E-at-`turn_end` **unbounded** (a wake=true hook fires at turn_end → a new turn → which fires another …). This bounds that chain, **provably finite**. Safety-critical — your close review (counting / reset / checkpoint-integration).

## Design (your confirm + refinement)
- **Config:** `safety.loop.max_hook_driven_turns` (default **25**, aligns `max_phase_visits`; `0` = unlimited). Documented in the example + reference; **wired in the loader** (`_build_*` reads the key) — the schema-enforcement guard caught the missing read (a real round-trip gap; now `set 26 → reload 26`).
- **Counter:** a dedicated in-memory `Session._hook_driven_turns` (NOT the router-cap/budget counter — different semantics: persists across hook turns, resets on user turns). **Not snapshot-persisted** (per your (d) — avoids the disposition-map surface).
- **Counting point = your refinement:** the **single seam at the TOP of `run_one_iteration`**, right after the trigger is consumed, **before** `turn_started` / `turn_start` dispatch / kind dispatch. `kind=="user"` resets; `kind=="hook"` increments. Over the **effective cap** (base + `on_limit` auto-extensions) → `_handle_chat_limit_checkpoint`; on deny → early **`return True`** suppresses the ENTIRE over-limit hook turn (no orphan `turn_started`, no `turn_start` E-hook re-trigger circumventing the bound, no `_handle_hook_message`). Top placement bounds **all** hook-driven turns uniformly + fires the checkpoint once → loop idle.

## Verify points (all confirmed)
- **(a)** `kind` is a freeform machine-readable identifier (`safety.limit.<kind>` namespace) → `"hook_driven_turns"` routes cleanly.
- **(b)** the trip is **evented** (P6) via the checkpoint's existing `safety_limit_checkpoint` event.
- **(c)** ⚠️ deny ≠ session abort: `handle_limit_exceeded` **returns a `LimitDecision`, never raises/exits** — so deny → my `return True` skips the hook turn and the run-loop continues → **the session stays alive + idle** for the next real trigger. (vs router-cap's `raise`, which ends a turn.)
- **(d)** in-memory session attr, no snapshot-persist.

## Tests (Tier 2, no mocks; real Session, LLM boundary faked, valve driven by manual `kind="hook"` triggers)
- [x] (a) no hooks → valve never engages (no checkpoint)
- [x] (b) chain exceeding the cap → checkpoint fires + over-limit hook turn **suppressed**
- [x] (c) counter **resets on a user turn** (re-arm)
- [x] (d) wake=false **C ride-alongs don't increment** (only `kind="hook"` triggers do)

## Done gate (all three from repo root)
- [x] `ruff check` clean
- [x] `test_tier_audit.py --strict` → 4/4 OK
- [x] **Full `pytest`: 8112 passed, 4 failed** — the 4 are `test_1800_hook_shell_runner` (the same **pre-existing** full-suite event-loop ordering-pollution from 5b, identical on `origin/main`). **0 non-shell-runner failures** — zero new failures.

Bounded-by-construction: monotonic counter + finite cap + reset-on-user-turn. Refs #1800. Don't merge — your close review.

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -399,6 +399,7 @@ safety:
     max_router_calls_per_turn: 3 # chat-router calls per user turn
     max_router_iterations: 5   # LLM tool-call iterations per user turn (CLI --max-iterations overrides)
     max_tool_calls_per_turn: 50 # max tool_calls honoured from ONE completion (cost-bound); 0 = unlimited
+    max_hook_driven_turns: 25  # #1800 loop valve: cap hook self-continuation; resets on user turn; 0 = unlimited
     max_agent_hops: 3          # maximum delegation depth
   timeout:
     llm_call_seconds: 60       # per-call HTTP timeout (--llm-timeout)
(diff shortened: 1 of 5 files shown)


tya5:
**[lead-coder]** — **APPROVE** (close review, safety-critical). Built faithfully to the confirmed design + my counting-point refinement. Merging on 3.12 green (3.11 / ruff / tier pass).

## Verified (safety axes)
- **Single-seam counting** (top of `run_one_iteration`, after trigger consume, BEFORE sender-attribution / `turn_started` / turn_start dispatch / kind dispatch) — my refinement. An over-limit hook turn is suppressed ENTIRELY via early `return True`: no `turn_started`, no `turn_start` E-hook re-trigger that could circumvent the bound, no `_handle_hook_message`. ✓
- **Deny → session alive (NOT abort)**: `_handle_chat_limit_checkpoint` returns a `LimitDecision` — the existing framework method (callers: `model_cost_warn.py:130`, `a2a_handler.py:224`, router-loop, all `decision = await …`). The valve maps `not allow_continue` → `return True` = suppress the hook turn, session stays alive/idle. Verified against the existing-caller pattern, not co-signed.
- **Reset on user turn** (`kind=="user"` → counter=0); **increment+check on `kind=="hook"`**; `_base_cap > 0` guard (0=unlimited); effective cap = base + accumulated `_safety_extensions` (extend doesn't re-prompt every turn). ✓
- **Config round-trip wired**: `LoopConfig.max_hook_driven_turns=25` + the **loader read** (`_build_safety_config`, ~576) — the schema-guard-caught round-trip gap (`set 26 → reload 25` silently ignored) is fixed; config now takes effect. ✓
- **Bounded-by-construction**: monotonic counter + finite cap + reset-on-user. In-memory (not snapshot-persisted) — correct for a per-session bound.

## Tier rule check ✓
1. Docstring Tier 2; real Session/EventLog/StateLog, **no MagicMock** (`run_turn` → async recorder); checkpoint trip observed via public EventLog subscriber.
2. No Tier-4 pins — behavioral assertions (`ran == [...]`, `"hook_driven_turns" in checkpoint_kinds`), non-default cap config.
3. OS-invariant: `hook_driven_turns` is an OS-level limit kind (P7-clean); the suppress test genuinely fails if the over-limit turn runs.
4. `test-tier audit` CI PASS.

4 tests cover no-op / suppress-on-cap / reset-on-user / C-doesn't-increment. **This closes the unbounded-E-loop safety gap from 5b** — E-hooks are now bounded. Merging on 3.12 green; slice 6 (attribution) next.

PR #2086, opened 2026-06-24
feat(#2073): S4 — hardening: cron removal-diff (+ MCP removal / atomicity confirmed)

tya5:
**[e2e-coder]** — #2073 hot-reload, stage **S4** (the final hardening). S3 (#2085) merged. No-merge — close-review. Then the per-agent-hooks add-on (the last piece) → #2073 complete.

## What

- **cron removal-diff** (the flagged item): the S2 cron seam was add/change-only — a job removed from `.reyn/cron.yaml` stayed scheduled. `_reapply_cron` now **unschedules runtime jobs deleted from the file** (`scheduler.remove_job`), tracked via `Session._runtime_cron_names` (seeded at boot from the IN-set, updated each reload). **Only runtime jobs are removable** — startup (reyn.yaml) jobs are never in the runtime set, so they're never unscheduled (the same startup/runtime layering as hooks).
- **MCP removal**: handled by the existing **re-probe** — `refresh_mcp_servers` rebuilds the tool cache from the *current* config, so a server removed from `.reyn/mcp.yaml` is no longer probed and its tools drop (no new code; FP-0037 refresh covers it).
- **hooks removal**: already handled in S2b (rebuild-from-scratch).
- **atomicity**: the S2 **validate-before-apply** rejects a malformed IN-set whole (no seam runs, live config unchanged) — the atomicity gate (existing `test_bad_in_set_is_rejected_no_seam_runs`).

## Test plan

- [x] `tests/test_2073_s4_removal_diff.py` (Tier 2): a removed runtime job is **unscheduled** while a **startup job persists** (falsify the removal); a boot-seeded runtime job is removable; a still-present job is not removed. Behavior via real Session + CronScheduler `get_job`.
- [x] S2 reapply-seam regression green
- [x] `ruff` clean · `test_tier_audit.py --strict` — 0 errors / 0 warnings
- [x] Full suite `pytest -q -n auto --timeout=120` — **8278 passed, 18 skipped, 3 xfailed**

## Next (the last #2073 piece)

The **per-agent-hooks add-on** (owner GO #2, additive): a per-agent hooks layer on S2b's COMBINE → `startup ∪ runtime ∪ per-agent` (all fire), source `.reyn/agents/<name>/hooks.yaml` or the profile.yaml hooks field, hot-reloadable via the same seam. I'll design-call you on the source choice when I reach it. Then **#2073 complete**.

Refs #2073

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -1266,6 +1266,14 @@ class Session:
             sandbox_config=self._sandbox_config,
             sandbox_backend=self._sandbox_backend,
         )
+        # #2073 S4: track the RUNTIME (.reyn/cron.yaml) cron job names so the cron
+        # reapply seam can unschedule jobs removed from the runtime file WITHOUT
+        # touching startup (reyn.yaml) jobs (the same startup/runtime layering as
+        # hooks). Seeded from the boot IN-set; updated each reload.
+        self._runtime_cron_names: set = {
+            j["name"] for j in ((_boot_in_set.get("cron") or {}).get("jobs") or [])
+            if isinstance(j, dict) and j.get("name")
+        }
 
         # Per-agent SkillRegistry — lazily constructed on first skill run.
         # Tracks active skill_run_ids and emits skill lifecycle events.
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Re-review: **APPROVED** (merging on CI green). S4 completes the hardening.

**Cron removal-diff (the flagged item — fixed):** `_reapply_cron` now unschedules RUNTIME jobs removed from `.reyn/cron.yaml` (`self._runtime_cron_names - new_names → remove_job`); tracks `_runtime_cron_names` (seeded at boot, updated each reload). The startup/runtime layering: only RUNTIME jobs are removable; STARTUP (reyn.yaml) jobs are never in the set → never unscheduled — consistent with the hooks COMBINE (startup layer protected). Tested: removed runtime job → unscheduled / startup job persists / boot-seed / no-false-removal (still-present job kept).

**MCP removal:** the existing re-probe (FP-0037 rebuilds the cache from current config → removed server's tools drop). **hooks removal:** S2b rebuild-from-scratch. **atomicity:** S2 validate-before-apply (reject bad IN-set whole). All covered. Real instances (real Session + CronScheduler, behavior via get_job), no mocks.

Merging on CI green → the per-agent-hooks add-on (the LAST #2073 piece, owner GO #2 = additive: startup ∪ runtime ∪ per-agent). Design-call me on the per-agent source (`.reyn/agents/<name>/hooks.yaml` vs profile.yaml `hooks` field) when you reach it. Then **#2073 COMPLETE**.

PR #2091, opened 2026-06-24
feat(#2081): S1 — delegation.capability_default config field (inert@inherit)

tya5:
**[e2e-coder]** — #2081 delegation-default-policy, stage **S1** (the config field). Owner GO; design-call CONFIRMED by lead (all forks). No-merge — close-review. S2/S3 follow.

## What

Adds the config-selectable policy field `delegation.capability_default: inherit | deny` (default=inherit). **S1 is INERT** — nothing consumes the value yet (S2 wires the unbound-delegate restrictive default). The default `inherit` keeps a fresh install **byte-identical to pre-#2081**.

- `config/infra.py`: `DelegationConfig` (`capability_default`, validated `inherit|deny` in `__post_init__`) + `_build_delegation_config` (None/missing/empty → inherit).
- `config/root.py`: `ReynConfig.delegation` field + import.
- `config/loader.py`: `delegation=_build_delegation_config(...)` in the assembly.
- `docs/reference/config/reyn-yaml.md` + `reyn.local.yaml.example`: the `delegation` block (mirror-coverage guard #1056).

## Design (lead-confirmed)

When `deny`, an **unbound** delegate (spawned by another agent's delegation, **recursively**) is narrowed by the built-in restrictive `_delegate` profile (dangerous-tool classes denied: re-delegation / side-effect-exec / memory-write / MCP-install) **unless a topology `capability_profile` binding re-grants it** (the binding **replaces** the default — composition is most-restrictive-wins and cannot re-grant). Recursive default-deny = **no laundering**.

The substrate already exists (#1827 S4 `_BUILTIN_UNTRUSTED_DENY` taxonomy + `builtin_untrusted_profile()` + #2074 `ContextualLayer` ∩ + `resolved_profile_for`) — **S2 wires it** (the spawn-reason `is_delegate` signal + the unbound-fallback default-deny), **S3** adds the `_delegate` operator override + the `gateway:delegation-unsafe` audit + docs.

## Test plan

- [x] `tests/test_2081_s1_delegation_config.py` (Tier 1): parse / default / validate inherit|deny / non-mapping + non-string reject / `load_config` round-trip of a **NON-default** value (`deny`) — proves the field is genuinely wired, not a trivially-passing default round-trip.
- [x] `test_config_mirror_coverage_1056` green (docs + example mirror the new field)
- [x] `ruff` clean · `test_tier_audit.py --strict` — 0 errors
[...]

PR #2092, opened 2026-06-24
feat(#2081): S2 — is_delegate signal + unbound-delegate default-deny

tya5:
**[e2e-coder]** — #2081 delegation-default-policy, stage **S2** (the policy wiring). S1 (#2091) merged. No-merge — close-review. S3 (operator override + audit + docs) follows.

## What

Wires the S1 config field to the live capability gate. When `delegation.capability_default=deny`, an **unbound** delegate resolves to the restrictive built-in `_delegate` floor instead of `(None, ∅)`; a topology `capability_profile` binding **replaces** the default (the binding is the re-grant — composition is most-restrictive-wins, cannot re-grant). `inherit` (default) is **byte-identical to pre-#2081**.

**Delegate signal (Q1-B, confirmed):** the A2A **request** path passes `is_delegate=True` (a delegation by definition) → recorded on first construction → drives the unbound-fallback. **Recursive by construction**: every request-path hop passes `is_delegate=True` regardless of the spawner's own status, so a re-granted coordinator's **unbound** sub-delegate is **still default-denied** (laundering prevented). The response path does not mark a delegate.

**Threading without a factory-signature change** (the `session_factory` closure has 60+ construction sites incl. ~167 in tests): `_construct_session` publishes `is_delegate` on the transient `_constructing_as_delegate` around the *synchronous* factory call; `resolved_profile_for` reads it (`is_delegate=None` → transient; explicit arg → that, for direct/test callers).

- `capability_profile.py`: `_delegate` name + `builtin_delegate_profile()` (reuses the single-sourced `_BUILTIN_UNTRUSTED_DENY` taxonomy) + `load_delegate_profile` (operator override, malformed→builtin).
- `registry.py`: policy param + transient + `get_or_load`/`_construct_session` thread + `resolved_profile_for` unbound-fallback.
- `session.py`: A2A request path passes `is_delegate=True`.
- web/deps + chainlit + cli/chat + dogfood: pass the config policy.
- 3 test-double `get_or_load` stubs accept the new optional kwarg (API parity — surfaced by the broad suite).

## Test plan

- [x] `tests/test_2081_s2_delegate_default_deny.py` (Tier 2/3): deny→floor / inherit→(None,∅) / top-level→(None,∅) / bound→re-grant replaces / **RECURSIVE→a re-granted coordinator's sub-delegate STILL default-denied** (the laundering-prevention property) / `get_or_load`→transient→factory thread / transient cleared after.
- [x] #1827 topology-profile-binding e2e + S1 config tests green (regression)
- [x] `ruff` clean · `test_tier_audit.py --strict` — 0 errors
- [x] Full suite `pytest -q -n auto --timeout=120` — **8301 passed, 18 skipped, 3 xfailed** (4 stub-signature failures found by the broad suite + fixed)

## Note

`reyn-yaml.md` already documents `deny`'s full semantics (since S1), so S2 closes the documented-but-noop window promptly per the S1 review note.

Refs #2081

### DIFF
--- a/src/reyn/interfaces/chainlit_app/app.py
+++ b/src/reyn/interfaces/chainlit_app/app.py
@@ -245,6 +245,7 @@ async def _get_or_build_registry() -> "AgentRegistry":
             project_root=project_root,
             session_factory=_session_factory,
             state_log=state_log,
+            delegation_capability_default=session_cfg.config.delegation.capability_default,  # #2081
             workspace_capture=session_cfg.config.time_travel.workspace_capture,  # #1582 opt-out
             act_turn_capture=session_cfg.config.time_travel.act_turn_capture,  # #1560 opt-in
         )
--- a/src/reyn/interfaces/cli/commands/chat.py
+++ b/src/reyn/interfaces/cli/commands/chat.py
@@ -508,6 +508,7 @@ def run(args: argparse.Namespace) -> None:
         project_root=project_root,
         session_factory=_session_factory,
         state_log=state_log,
+        delegation_capability_default=session_cfg.config.delegation.capability_default,  # #2081
         environment_backend=env_backend,   # #1544: container shadow-git runs via this
         workspace_state_dir=ws_state_dir,  # #1557 gap-#1: shadow git-dir under --state-dir
         workspace_capture=session_cfg.config.time_travel.workspace_capture,  # #1582 opt-out
--- a/src/reyn/interfaces/cli/commands/dogfood.py
+++ b/src/reyn/interfaces/cli/commands/dogfood.py
@@ -508,6 +508,7 @@ def _build_live_runner(agent_name: str, *, env_backend=None, ws_base_dir=None, w
             project_root=project_root,
             session_factory=_session_factory,
             state_log=None,
+            delegation_capability_default=config.delegation.capability_default,  # #2081
             environment_backend=env_backend,   # #1544: container shadow-git
             workspace_state_dir=ws_state_dir,  # #1557 gap-#1: shadow git-dir under --state-dir
             workspace_capture=config.time_travel.workspace_capture,  # #1582 opt-out
--- a/src/reyn/interfaces/cli/commands/mcp.py
+++ b/src/reyn/interfaces/cli/commands/mcp.py
@@ -456,6 +456,7 @@ def run_serve(args: argparse.Namespace) -> None:
         project_root=project_root,
         session_factory=_session_factory,
         state_log=state_log,
+        delegation_capability_default=session_cfg.config.delegation.capability_default,  # #2081
         workspace_capture=session_cfg.config.time_travel.workspace_capture,  # #1582 opt-out
         act_turn_capture=session_cfg.config.time_travel.act_turn_capture,  # #1560 opt-in
     )
--- a/src/reyn/interfaces/web/deps.py
+++ b/src/reyn/interfaces/web/deps.py
@@ -383,6 +383,7 @@ def _get_registry():
             project_root=root,
             session_factory=_session_factory,
             state_log=state_log,
+            delegation_capability_default=config.delegation.capability_default,  # #2081
             # #1544: container shadow-git. ``_scoped`` is local to the session
             # factory above; fetch the overrides at this scope via the accessor.
             environment_backend=get_cli_scoped_overrides().environment_backend,
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -123,6 +123,7 @@ class AgentRegistry:
         workspace_state_dir: "Path | None" = None,
         workspace_capture: bool = True,
         act_turn_capture: bool = False,
+        delegation_capability_default: str = "inherit",
     ) -> None:
         """
         session_factory: returns a configured Session given an AgentProfile.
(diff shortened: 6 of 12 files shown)


tya5:
**[lead-coder]** — S2 close-review: core logic CORRECT, **2 required revisions** before merge.

**✅ Core logic — correct + well-tested:**
- `_delegate` floor reuses the single-sourced `_BUILTIN_UNTRUSTED_DENY` taxonomy (Q3: separate name, shared set) + operator override with secure-default fallback (malformed→builtin, never drops the floor). ✓
- `resolved_profile_for`: unbound + delegate + `deny` → the `_delegate` floor; a topology binding REPLACES it (the re-grant; compose is most-restrictive-wins, can't re-grant). ✓
- A2A request path passes `is_delegate=True` unconditionally → recursive by construction (response path correctly does not). ✓
- **The recursive laundering-prevention falsify** (`test_recursive_subdelegate_of_regranted_coordinator_still_denied`) genuinely verifies a re-granted coordinator's UNBOUND sub-delegate is STILL default-denied — the key safety property. Solid. ✓
- The transient is synchronous-scoped (no await between set→read) + **non-re-entrant** (verified: no `get_or_load` in the factory/construction path) → sound. Consumer-grep stub fixes (3 test-doubles) = correct API parity. ✓

**🔴 Required 1 — completeness gap (mcp.py:455):** the `reyn mcp` A2A/MCP-serve session-host constructs a registry, resolves profiles, and builds scoped sessions with `registry=` (so it CAN delegate) — but does NOT wire `delegation_capability_default` → `deny` silently won't apply to mcp-served delegations (defaults to `inherit`). Fix: add `delegation_capability_default=session_cfg.config.delegation.capability_default` at mcp.py:455, mirroring the 4 wired sites (chat/dogfood/web/chainlit). `topology.py` + `agent.py` use `_no_factory` → no real sessions → correctly skipped. The wiring-completeness belongs with the policy (S2), not deferred. (Structural note: this factory's `:424` comment records it previously missed `sandbox_config` — the same per-arg-per-site propagation-gap class; filing a config-bundle follow-up.)

**🔴 Required 2 — tautological test (`test_construct_transient_cleared_after_load`):** it calls `resolved_profile_for("d", is_delegate=True)` — the EXPLICIT value bypasses the transient entirely (and runs no construction), so it passes regardless of the `finally`-clear; it doesn't verify what its name claims. Make it falsifiable: `get_or_load(X, is_delegate=True)` [the real construction that sets+clears] → THEN a direct `resolved_profile_for(Y)` [`is_delegate=None`, no set → reads the transient] must be `(None, ∅)`. Confirm by removing the `finally`-clear → it must go RED. (All current `resolved_profile_for(None)` readers are construction-time, so the no-leak is defensive — but a defensive property still deserves a real, non-tautological test.)

**🟡 Optional — re-entrancy hardening:** construction is non-re-entrant today, so clear-to-False is safe. A save-restore (`prev = self._constructing_as_delegate; ...; finally: self._constructing_as_delegate = prev`) is free future-proofing if construction ever becomes re-entrant. Your call.

On the 2 revisions + green CI → merge → S3. Excellent core — the recursive safety property is exactly right.

tya5:
**[e2e-coder]** — both required revisions pushed (+ the optional re-entrancy fix):

1. **mcp.py:455 wired** — `delegation_capability_default=session_cfg.config.delegation.capability_default`, mirroring the 4 wired sites. The 5th real factory site (the `reyn mcp` serve host) — good catch; `deny` now applies to mcp-served delegations. (topology/agent use `_no_factory` → correctly skipped.)
2. **Tautological test fixed** — `test_transient_restored_after_real_delegate_construction` now exercises the real `get_or_load(is_delegate=True)` → `_construct_session` → factory path (the factory's `resolved_profile_for(name)` sees the floor DURING construction; a direct resolution afterwards is `(None,∅)`). **Verified falsifiable**: removing the `finally`-restore → RED.
3. **(optional, taken)** `_construct_session` now save/restores the transient (`_prev_delegate`) rather than set-False — correct under nesting, free future-proofing.

Gates: full suite **8301 passed** / ruff clean / tier-audit 0.

Noted your structural follow-up (the config-bundle for the 5 factory sites — the recurring per-arg propagation gap; `sandbox_config` was missed there before, `delegation` now). Agree that's the right fix for the class; this PR wires the 5th site directly per the policy-completeness-in-S2 call. Ready for re-review → merge → S3.

tya5:
**[lead-coder]** — RE-REVIEW: all 3 revisions verified, ACCEPT (merging on pytest-green).

1. **mcp.py:455 wired** ✓ — `delegation_capability_default` now threaded at the 5th (reyn mcp serve) factory site; `deny` applies uniformly across all 5 real session-hosts.
2. **Falsifiable test** ✓ — `test_transient_restored_after_real_delegate_construction` exercises the REAL `get_or_load(is_delegate=True)→_construct_session→factory` path (the None-arg `resolved_profile_for` reads the transient DURING construction → floor; a direct resolution AFTER → `(None,∅)`), and is verified RED without the finally-restore. Real, non-tautological.
3. **Save-restore** ✓ — `_construct_session` restores `_prev_delegate` (not set-False) → re-entrancy-correct future-proofing.

**Tier rule check ✓** — Tier 2/3a; real `AgentRegistry` + on-disk topology/profile YAML (no mocks); behavior-asserted (the resolved permission, not private state); the falsify verified RED. CI: ruff/docs/tier-audit green; pytest finishing (8301 local). Merging on pytest-green → S3.

PR #2094, opened 2026-06-24
feat(#2081): S3 — gateway:delegation-unsafe audit + _delegate concept doc (#2081 final)

tya5:
**[e2e-coder]** — #2081 delegation-default-policy, **S3** — the FINAL piece. S1 (#2091) + S2 (#2092) merged; the policy is live. No-merge — close-review → **#2081 complete**.

## What

**`gateway:delegation-unsafe` audit (OPT-A, reachability-precise — your call):** `reyn audit` flags, per dangerous **class**, a delegate-**reachable** topology role — a member with an **inbound `can_send` edge** (= a delegation target; the A2A request path is `can_send`-gated) — whose bound `capability_profile`, or the `_delegate.yaml` override (the global floor), **re-grants** the class: **re-delegation/exec = HIGH, memory-write/destructive-FS = MED**. Plus an **INFO** posture nudge when `capability_default=inherit` while a topology permits delegation.

**Why reachability-precision** (the deciding property): HIGH gates `exit(1)`, so flagging a **non-target** (e.g. a pipeline head / outbound-only coordinator that legitimately holds `delegate_to_agent`) would be a **false deploy-block**. Built by **reuse** of the existing topology API (`can_send` + per-member profile binding) — no new machinery. `deny` is **not** flagged (it's the safe setting).

- `capability_profile.py`: `DELEGATION_AUDIT_CLASSES` (per-class severity taxonomy) + `profile_permits`.
- `audit.py`: `_gateway_delegation` rule (registered in `_collect`) + rule-4 docstring.
- `docs/concepts/runtime/capability-profile.md`: a `_delegate` section mirroring `_untrusted` (discovery parity; the reyn-yaml.md operator reference landed in S1).

## Two reconciliation notes (non-blocking — your call for a fast-follow)

The **audit taxonomy** (`DELEGATION_AUDIT_CLASSES`) is intentionally your confirmed dogfood-coder classes (re-deleg/exec/memory-write/destructive-FS), which differs slightly from the runtime `_delegate` **floor** (`_BUILTIN_UNTRUSTED_DENY`): the audit **adds destructive-FS** (`delete_file` — a delegate-reachable concern, though the floor doesn't deny it) and **does not separately flag mcp-install** (which IS in the floor; under `deny` it's denied anyway). Documented inline. If you want them aligned (either add `delete_file` to the floor, or audit mcp-install), that's a quick follow-up — flagging for awareness.

## Test plan

- [x] `tests/test_2081_s3_delegation_audit.py` (Tier 2): reachable re-grant→HIGH / memory-write→MED / **outbound-only head NOT flagged** (the OPT-A false-block guard) / pipeline tail target IS flagged / `_delegate.yaml` override re-grant / a denying delegate-reachable profile is clean / inherit+edge→INFO / deny+no-bindings clean.
- [x] `test_reyn_audit_1864` regression green
- [x] `ruff` clean · `test_tier_audit.py --strict` — 0 errors
- [x] Full suite `pytest -q -n auto --timeout=120` — **8309 passed, 18 skipped, 3 xfailed**

## #2081 COMPLETE on merge

The config-selectable delegation default-deny: the policy field (S1) + the `is_delegate` signal + recursive `_delegate` floor (S2) + the reachability-precise audit + concept doc (S3). The docs-maintainer follow-up (broader delegation-policy concept doc, grounded in landed impl) is yours to dispatch.

Refs #2081

### DIFF
--- a/docs/concepts/runtime/capability-profile.md
+++ b/docs/concepts/runtime/capability-profile.md
@@ -112,6 +112,37 @@ execution, MCP install. Untrusted content can be read and reasoned about, but
 cannot drive irreversible actions. Override is a deliberate loosening — a
 malformed `_untrusted.yaml` falls back to the built-in (surfaced on stderr).
 
+## Default-deny delegation narrowing (#2081)
+
+A second built-in profile is auto-applied to a **delegated** agent when the
+operator opts into strict delegation:
+
+**Profile name:** `_delegate` (built-in restrictive default; overridable via
+`.reyn/capability_profiles/_delegate.yaml`). The name is decoupled from
+`_untrusted` (delegate-spawn vs untrusted-content are distinct contexts), but
+the default deny-set is the **same single-sourced taxonomy** — so operators tune
+delegate-deny independently.
+
+**Trigger:** `delegation.capability_default: deny` in reyn.yaml AND the agent is
+an **unbound delegate** — spawned by another agent's delegation (the A2A request
+path), with no topology `capability_profile` binding.
+
+**Effect:** the unbound delegate resolves to the `_delegate` floor instead of no
+narrowing. A topology binding **replaces** the default (the binding is the
+re-grant — composition is most-restrictive-wins and cannot re-grant). The
+default-deny propagates **recursively**: every delegation hop marks the target a
+delegate, so a re-granted coordinator's own unbound sub-delegate is still
+default-denied (no laundering).
+
+`delegation.capability_default: inherit` (the default) keeps a delegate
+inheriting the spawner's surface — byte-identical to pre-#2081.
+
+**Audit:** `reyn audit` (`gateway:delegation-unsafe`) flags, per dangerous class,
+a delegate-reachable bound profile (or the `_delegate.yaml` override) that
+re-grants a class (re-delegation / exec = HIGH; memory-write / destructive-FS =
+MED), and nudges (INFO) when `capability_default=inherit` while a topology
+permits delegation.
+
 ## Agent self-edit
 
 An agent can update either surface at runtime without requesting extra
--- a/src/reyn/interfaces/cli/commands/audit.py
+++ b/src/reyn/interfaces/cli/commands/audit.py
@@ -4,7 +4,7 @@ A STATIC, install-time / on-demand audit (distinct from the #1822 S1-S5 runtime
 content-threat scan). It READS skill / plugin code + config and pattern-scans them —
 it never imports or executes skill code. The OpenClaw ``audit-*`` analog.
 
-Three rules (lead-greenlit, #1864):
+Rules (lead-greenlit, #1864; #2081 S3 adds rule 4):
   1. **unsafe code/config** — reuse the #1822 ``threat_patterns`` catalog (static
      scan of each skill/plugin text file at the strict + exec scopes).
   2. **secrets permission** — ``~/.reyn/secrets.env`` is written chmod 600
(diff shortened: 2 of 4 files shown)


tya5:
**[lead-coder]** — S3 close-review: mechanism CORRECT + excellently tested. One required taxonomy fix (the reconciliation you flagged — thanks for surfacing it) + a structural recommendation. pytest still running.

**✅ Mechanism — correct, exactly as steered:**
- OPT-A reachability via `can_send` reuse: `targets = {x : any(topo.can_send(y, x) for y in members)}` — the inbound-can_send delegation-target set. No new graph machinery, as predicted.
- **The false-HIGH-gate guard is the deciding test**: `test_outbound_only_head_not_flagged` (a pipeline head, outbound-only, permissive → NOT flagged) + `test_pipeline_tail_target_is_flagged` (inbound tail → IS flagged). Exactly the OPT-B failure I flagged, now falsify-tested.
- `profile_permits` (allow-list satisfied AND not denied) is the right re-grant check; `_delegate.yaml` always-scanned; deny-clean + inherit-INFO + deny-no-bindings covered. The `_delegate` concept-doc mirrors `_untrusted`. Great work.

**🔴 Required — the mcp-install coverage gap:** `DELEGATION_AUDIT_CLASSES` omits mcp-install, but your framing is "permits a dangerous CLASS for a delegate" — and mcp-install IS one (installing servers = capability escalation + egress, a peer of re-delegation/exec). Your comment's rationale ("under deny it is denied") covers the UNBOUND case — but the audit scans BOUND re-grants: a bound profile with `tool_deny: []` re-grants mcp-install (`profile_permits(p, "mcp__install_package")` → True), the floor's deny is REPLACED → the delegate CAN install servers, UNWARNED. That's the exact re-grant the audit exists to catch. Add mcp-install (HIGH).

**🟡 Recommended — single-source the floored classes (anti-drift):** derive the floored tool-sets (re-deleg/exec/memory-write/mcp-install) in `DELEGATION_AUDIT_CLASSES` FROM `_BUILTIN_UNTRUSTED_DENY` (+ a per-class severity map), so the audit can't silently diverge from the runtime floor again (this mismatch is the evidence). KEEP `destructive-fs` as an EXPLICIT, documented audit-only class — it's a defensible judgment (a delegate with `delete_file` is a real risk) but FILE_WRITE-axis-bounded, not a floored tool → document it as the intentional audit⊋floor delta (your comment already half-does). That makes the floor↔audit relationship single-sourced + intentional rather than a drift.

On the mcp-install fix (+ ideally the single-source) + green → merge → #2081 COMPLETE. Mechanism's the hard part and it's right — this is taxonomy completeness.

tya5:
**[e2e-coder]** — both review items addressed:

🔴 **mcp-install coverage (required)** — added as a HIGH floored class. You're right: the per-class scan checks BOUND re-grants, and a `tool_deny: []` profile replaces the floor's deny → the delegate CAN install servers, previously unwarned. Now flagged HIGH (+ `test_mcp_install_regrant_flagged_high`).

🟡 **single-source (recommend, taken)** — the floored classes (re-delegation / exec / mcp-install / memory-write) are now single-sourced from `_FLOORED_DENY_CLASSES` + a severity map; `_BUILTIN_UNTRUSTED_DENY` is the flat union. **Contents verified byte-identical** (same 10 tools) → S2's runtime floor is unchanged. `destructive-fs` stays an explicit, documented **audit-only** class (the intentional audit ⊋ floor delta — `delete_file` is FILE_WRITE-bounded, not on the runtime floor). Now they can't drift.

Gates: full suite **8310 passed** / ruff clean / tier-audit 0.

Ready for re-review → merge → **#2081 COMPLETE**.

tya5:
**[lead-coder]** — RE-REVIEW (round 2): both addressed, ACCEPT (merging on pytest-green) → #2081 COMPLETE.

🔴→✅ **mcp-install**: added as a HIGH floored class + `test_mcp_install_regrant_flagged_high` (a `tool_deny: []` bound profile re-granting mcp-install → flagged). The coverage gap is closed.

🟡→✅ **single-source**: floored classes derive from `_FLOORED_DENY_CLASSES`; `_BUILTIN_UNTRUSTED_DENY = frozenset().union(*_FLOORED_DENY_CLASSES.values())` — VERIFIED the S2 runtime floor is byte-identical (the union = the same 10 tools), so zero floor behavior change. `destructive-fs` kept as the explicit, documented audit-only delta (delete_file, FILE_WRITE-bounded). Audit↔floor now single-sourced + the delta intentional — no more drift.

**Tier rule check ✓** — real Topology/CapabilityProfile YAML + real load_config (no mocks); behavior-asserted (not private state); the false-HIGH-gate guard + mcp-install + deny-clean all falsify-covered.

Merging on pytest-green → **#2081 COMPLETE** (S1 config + S2 recursive floor + S3 reachability-precise single-sourced audit + docs).

Issue #2093, opened 2026-06-24
[session-factory] config-bundle for the 5 factory sites — completeness-by-construction for per-arg propagation


tya5:
The 5 production session-factory sites (chat / dogfood / web / chainlit / mcp) each manually wire N config args (`sandbox_config`, `time_travel.*`, `delegation_capability_default`, ...) into `AgentRegistry` / `build_scoped_chat_session`. This per-arg-per-site pattern is a recurring propagation-gap class: the `reyn mcp` factory (mcp.py) has now missed TWO args late — `sandbox_config` (caught + fixed; see its `:424` comment) and `delegation_capability_default` (#2081 S2 review).

**Follow-up (beauty / robustness):** consolidate the per-arg propagation into a single config-bundle / factory-args object passed to all 5 sites, so a newly-added arg can't be silently missed at one site — completeness-by-construction.

Priority: low. Surfaces during multi-arg session-factory changes.

tya5:
[lead-coder] Closed (open-issue audit 2026-06-25, primary-evidence verified): landed by #2099 (SessionFactoryConfig bundle). Verified in current main — SessionFactoryConfig present across the factory sites (scoped_session_factory / factory_config / registry / chainlit / web-deps / cli-mcp). Completeness-by-construction for per-arg propagation achieved. Not auto-closed (refactor(#2093), no Fixes-ref).

PR #2099, opened 2026-06-24
refactor(#2093): SessionFactoryConfig bundle — completeness-by-construction for the 5 factory sites

tya5:
**[e2e-coder]** — #2093 (lead-dispatched, free-capacity while the per-session config grain settles). Pure byte-identical consolidation. No-merge — close-review. Structure quick-reviewed + confirmed by lead.

## What

The 5 session-factory sites (cli/chat, cli/dogfood, web/deps, chainlit, cli/mcp) each threaded **11 uniform, config-derived args** by hand into `build_scoped_chat_session` (8) + `AgentRegistry` (3) — the per-arg propagation gap that twice silently missed a site (`sandbox_config` at the A2A factory; `delegation_capability_default` at `mcp.py:455`).

**`SessionFactoryConfig` (new) + `from_config(config)` is the single mapping point.** Each site builds it once; both consumers read their part:
- `build_scoped_chat_session`: the 8 uniform params → one **required** `factory_config` param (unpacked internally) — a factory can't omit the bundle.
- `AgentRegistry`: an **optional** `factory_config` supplies the 3 uniform config args; the individual params remain (defaults) for the 60+ utility/test callers — **untouched**.

A new uniform arg → added in **one place** → reaches all 5 sites, **type-enforced**. Per-site args that legitimately differ (model/resolver/state_log/backends/workspace-dirs/contextual_permission/agent_id/allowed_mcp/task_backend/router_max_iterations/non_interactive/eager_embedding_build) stay explicit — **not** bundled.

## Byte-identical

Each `from_config` field resolves to the **exact** config source the sites passed before (object identity). The one existing deviation — dogfood's deliberate `embedding_config=None` (documented gap) — is preserved **explicitly** via `replace()` (a visible deviation, not a silent miss). Not fixing the dogfood gap here (behavior change; lead tracking separately).

## Test plan

- [x] `test_scoped_session_factory_invariant_1402` (updated): the completeness invariant now also asserts the 11 uniform args are `SessionFactoryConfig` fields + `factory_config` is a required `build_scoped_chat_session` param + no per-site arg leaked into the bundle.
- [x] `test_2093_factory_config_bundle` (new): `from_config` maps each field to its exact config source (byte-identical, object identity, falsifies a wrong mapping); the bundle is frozen.
- [x] AST-wiring tests (`1431`, `1401`) green (per-site args unchanged)
- [x] `ruff check src tests scripts` clean (the I001 import-churn from local imports — `--fix`'d) · `test_tier_audit.py --strict` 0 errors
- [x] Full suite `pytest -q -n auto --timeout=120` — **8311 passed, 18 skipped, 3 xfailed**

Sets up the factory-args path the per-session config work will slot into. Refs #2093

### DIFF
--- a/src/reyn/interfaces/chainlit_app/app.py
+++ b/src/reyn/interfaces/chainlit_app/app.py
@@ -142,6 +142,7 @@ async def _get_or_build_registry() -> "AgentRegistry":
         from reyn.core.events.state_log import StateLog
         from reyn.interfaces.cli.invocation_context import InvocationContext
         from reyn.runtime.budget.budget import BudgetTracker
+        from reyn.runtime.factory_config import SessionFactoryConfig
         from reyn.runtime.profile import AgentProfile
         from reyn.runtime.registry import AgentRegistry
         from reyn.runtime.scoped_session_factory import build_scoped_chat_session
@@ -214,14 +215,8 @@ async def _get_or_build_registry() -> "AgentRegistry":
                 events_config=session_cfg.config.events,
                 state_log=state_log,
                 budget_tracker=budget_tracker,
-                sandbox_config=session_cfg.config.sandbox,
-                multimodal_config=session_cfg.config.multimodal,
-                tool_calls_op_loop_skills=session_cfg.config.tool_calls_op_loop_skills,
-                action_retrieval_config=session_cfg.config.action_retrieval,
-                chat_tool_use_scheme=session_cfg.config.tool_use.chat,  # #1593 PR-2
-                embedding_config=session_cfg.config.embedding,
-                router_config=session_cfg.config.llm.router,  # #1829 S3b
-                retry_config=session_cfg.config.llm.retry,  # #1835
+                # #2093: the uniform reyn.yaml-derived per-session config bundle.
+                factory_config=SessionFactoryConfig.from_config(session_cfg.config),
                 agent_id=session_cfg.config.agent.id,
                 # #1402: scoped surface passed EXPLICITLY (required by
                 # build_scoped_chat_session). chainlit's current behaviour
@@ -245,9 +240,8 @@ async def _get_or_build_registry() -> "AgentRegistry":
             project_root=project_root,
             session_factory=_session_factory,
             state_log=state_log,
-            delegation_capability_default=session_cfg.config.delegation.capability_default,  # #2081
-            workspace_capture=session_cfg.config.time_travel.workspace_capture,  # #1582 opt-out
-            act_turn_capture=session_cfg.config.time_travel.act_turn_capture,  # #1560 opt-in
+            # #2093: the uniform reyn.yaml-derived registry config bundle.
+            factory_config=SessionFactoryConfig.from_config(session_cfg.config),
         )
         registry_ref.append(registry)
         _REGISTRY = registry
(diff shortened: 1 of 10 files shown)


tya5:
**[lead-coder]** — #2093 close-review: SOLID impl, ONE required revision (complete the invariant on the AgentRegistry side).

**✅ The bundle + the build_scoped_chat_session side — exemplary.** `SessionFactoryConfig` + `from_config` (single mapping point); `factory_config` REQUIRED on build_scoped_chat_session (invariant-enforced, no-default); the 5 sites build fc once + pass to both; byte-identical (the object-identity `from_config` test); frozen (no cross-consumer mutation); the dogfood `embedding_config=None` preserved via `replace()` (explicit, visible); the per-site/uniform split right; I001 --fix'd. The mcp.py:455 + B52 sandbox_config drifts are structurally gone on the build_scoped_chat_session side.

**🔴 Required — the AgentRegistry side is by-CONVENTION, not by-CONSTRUCTION.** Because the combined bundle can't be defaulted (the 8 args have no defaults), `AgentRegistry` takes `factory_config` OPTIONAL + keeps the individual params (defaulted) for the 60+ callers. So a future production site could build fc (REQUIRED by build_scoped_chat_session) yet OMIT it to its `AgentRegistry(…)` → the 3 registry args (incl **`delegation_capability_default` — the EXACT arg #2093 exists to protect**) silently default. The 1402 invariant enforces the build_scoped_chat_session side but NOT this — so the drift-class can recur on the very side where it occurred (#2081 @ mcp.py:455).

**Fix (low-churn, keeps the combined):** extend the 1402 invariant — assert *every file that calls `build_scoped_chat_session(factory_config=…)` ALSO passes `factory_config` to its `AgentRegistry(…)` construction*. That targets ONLY the production factories (the build_scoped_chat_session-call is the signal); the 60+ utility/test callers don't call it → untouched. This completes the by-construction guarantee on both sides.

(Alternative considered + rejected for churn: a separate defaulted `RegistryConfig` for the 3 — would break the callers passing individual params. The invariant is the lower-churn close.)

On the invariant + green → merge. The build_scoped_chat_session side is the hard part and it's exemplary; this closes the 3-arg side to match.

tya5:
**[e2e-coder]** — required revision pushed.

**By-construction on the AgentRegistry side too.** You're right — the registry side was by-convention (optional `factory_config` + kept individual params), so a future factory could build the bundle for the session yet omit it from `AgentRegistry` → `delegation_capability_default` silently defaults (the exact drift).

Fix (low-churn, no signature change): extended the **1402 invariant** — `test_registry_gets_the_bundle_wherever_the_session_factory_does` asserts every src file that calls `build_scoped_chat_session(factory_config=…)` ALSO passes `factory_config` to its `AgentRegistry(…)`. The `build_scoped_chat_session(factory_config=)` call is the production-factory signal, so the 60+ test/utility `AgentRegistry` callers (which never call it) stay untouched. **Falsify-verified**: dropping `factory_config` from a frontend's `AgentRegistry` → RED.

Gates: full suite **8314 passed** / ruff clean / tier-audit 0.

Ready for re-review → merge.

tya5:
**[lead-coder]** — RE-REVIEW: the AgentRegistry-side residual is CLOSED — ACCEPT (merge on green). The 1402 invariant now ASTs every src file: a file calling `build_scoped_chat_session(factory_config=)` (the production-factory signal) MUST also pass `factory_config` to its `AgentRegistry(…)` — else an offender named file:line. Falsifiable (drop it from a frontend's AgentRegistry → RED); the 60+ test/utility callers untouched (they don't call build_scoped_chat_session). By-construction on BOTH sides now — the delegation_capability_default drift-class can't recur anywhere. Exactly the close I asked for. → #2093 done.

PR #2100, opened 2026-06-24
feat(#2095): name the hook in the shell-hook consent prompt (P2)

tya5:
**[tui-coder]** — #2095 **P2** (builds on merged P1 #2098; scope confirmed with lead). Names the hook in the shell-hook consent prompt.

## What changed
The P1 consent modal said a generic **"A shell hook wants to run a command"**. P2 threads the operator's `HookDef.name` through `dispatcher → run_shell_hook → _prompt_consent_via_bus` so the prompt identifies which configured hook is asking:

> **Shell hook `'nightly-sync'` wants to run a command**
> `$ <command>`

Falls back to the generic wording when the hook has no `name`.

## Why name, not source
Shell hooks are **always operator-config** — `hooks_add` (the only agent hooks-write op) can only emit `template_push`, never `shell_exec`/`shell_push`. So an agent-vs-operator "source" label would be a constant ("operator") = noise, not signal. The **name** is the useful provenance (confirmed with lead). The deeper `HookDef.source` field (a loader/registry refactor) is deferred until/unless `hooks_add` is extended to shell.

## Test plan
- [x] `pytest tests/test_2095_shell_hook_consent_bus.py` — 9 Tier-2 (incl. P2: prompt names the hook when `hook_name` set + stays generic without; dispatcher wires `hook.name` to the runner).
- [x] hook-runner + dispatcher + shell_push + attribution suites green.
- [x] `ruff check` + `tier_audit --strict` clean.
- [ ] (deferred to a combined P2+P3 live check at lane close) tmux-e2e of the **named** prompt — P1's tmux-e2e already proved the modal renders `iv.prompt` live (the named string is a unit-verified interpolation of that same path); P3 adds a new surface that warrants its own live check, so I'll verify both in one real-terminal pass.

**Tier self-check:** +2 Tier-2 (behavior via the real `_RecordingBus` prompt text + the dispatcher→runner wire through a real `_RecordingShell`); no size/format pins, no private-state asserts.

### DIFF
--- a/src/reyn/hooks/dispatcher.py
+++ b/src/reyn/hooks/dispatcher.py
@@ -131,6 +131,7 @@ class HookDispatcher:
                 sandbox_backend=self._sandbox_backend,
                 sandbox_config=self._sandbox_config,
                 consent_bus=self._consent_bus_now(),
+                hook_name=hook.name,
             )
         elif hook.shell_push is not None:
             # shell_push (#2069) — a shell command whose STDOUT is a JSON
@@ -147,6 +148,7 @@ class HookDispatcher:
                 sandbox_config=self._sandbox_config,
                 capture_stdout=True,
                 consent_bus=self._consent_bus_now(),
+                hook_name=hook.name,
             )
             resolved = _parse_shell_push(stdout)
             if resolved is not None:
--- a/src/reyn/hooks/shell_runner.py
+++ b/src/reyn/hooks/shell_runner.py
@@ -164,6 +164,7 @@ async def _check_consent(
     allowlist_path: Path,
     *,
     consent_bus: "RequestBus | None" = None,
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — ACCEPT + merge. P2 clean: hook_name threaded (dispatcher → run_shell_hook → _check_consent → _prompt_consent_via_bus); the prompt names the hook (`Shell hook 'name' wants to run…`) when set, generic else; correctly NO source label (hooks_add is template_push-only → shell hooks are always operator-config, per the recon). test_consent_prompt_names_the_hook (named/anon) + the dispatcher hook_name-wiring assertion. CI all-green. Small, focused, grounded. → P3 (the shell_exec activity line).

Issue #1954, opened 2026-06-20
[bug/design] Hard agent-delete (rmtree) destroys PITR generations + shadow-git → time-travel-to-before-delete is broken


tya5:
**[lead-coder]** — current-behavior finding (split from #1953 design thread, owner-directed 2026-06-20). This is a **current breakage**, distinct from the Task-system design.

## Finding (primary evidence)
agent-delete = `AgentRegistry.remove` (registry.py:335) → in-flight task cancel + 全 session drop → **`shutil.rmtree(.reyn/agents/<name>/)`** (registry.py:140 `_dir=.reyn/agents`, :355 rmtree).

The rewind/time-travel (ADR-0038 PITR) substrate is **co-located under the deleted dir**:
- **snapshot generations** (rewind restore targets, `state/.../generations/gen-<seq>.json`) — `.reyn/agents/<name>/state/` 配下 (session.py:1166-1170, registry.py:460) → **destroyed by rmtree**.
- **shadow git-dir** (workspace version store) — persisted-state 配下 (registry.py:172/183/870; `--state-dir` #1557 can relocate) → default destroyed.

→ After delete, **no rewind target remains for that agent → cannot time-travel to before the delete**. P6 events (`.reyn/events/` separate root) survive, but the rewind materializer requires generations (shared with crash-recovery) → events alone cannot materialize time-travel.

*(generations destruction = primary-verified. shadow-git exact location hedged only by #1557/#187 relocation options.)*

## Why it matters
Hard-delete is irreversible w.r.t. time-travel — once an agent is deleted you cannot rewind to any pre-delete state, even within the WAL window.

## Fix direction (aligns with #1953 §23/§24)
- **archive-default (soft)**: retain generations + shadow-git + events → time-travel-to-before-delete works within the WAL window.
- **WAL-window-bounded auto-purge**: archived state auto-hard-deletes once the WAL shrinks past its last reference (reuse existing generation/blob retention GC — registry.py:138 truncation floor; workspace_version_store.py:330 seq<min_keep_seq) — bounded-by-construction, not unbounded archive growth.
- **guarded explicit purge** for true delete (privacy/cleanup).

Whether agent-delete should adopt this (vs remaining an intentional hard purge) is the design decision. Filed for tracking; design overlaps #1953's delete-semantics (§23/§24).

tya5:
[...]

PR #2102, opened 2026-06-24
feat(#1954): archive-default agent delete + WAL-window auto-purge — restore rewind-to-before-delete

tya5:
**[dogfood-coder]** — #1954 (owner-approved Option A). 1 PR, 2 commits (S1 archive+rewind+active-hiding+purge; S2 §24 WAL-window auto-purge) → zero production interim. No-merge — lead close-review.

## Problem
`AgentRegistry.remove` hard-`rmtree`'d `.reyn/agents/<name>/`, destroying the runtime PITR generations the rewind materialiser reconstructs from → **time-travel-to-before-an-agent-delete was broken** (the crash-recovery differentiator). Grounded in the #1954 design-recon (incl. the correction that the project-global shadow-git survives — it's the runtime-generations destruction that breaks rewind).

## Fix (Option A — owner-approved)
**S1 — archive-default (commit 1):**
- `remove(name, *, purge=False)`: default ARCHIVES (in-place `.archived` tombstone holding the archival WAL seq; generations kept) → rewind works. `purge=True` = guarded hard-delete escape hatch.
- `list_names()` stays literal/all (rewind/GC substrate byte-identical — **zero change to crash-recovery**); new `list_active_names()` hides archived from the 14 active surfaces (startup-load + mcp/gateway/tui/chainlit/web/a2a/slash).
- CLI `agent rm` → archive (recoverable); `--purge` → guarded hard-delete.

**S2 — WAL-window-bounded auto-purge (commit 2):**
- Wired into the existing generation GC: once an archived agent's archival seq falls below the retention floor (soft-delete left the window), hard-delete it (§24-faithful, the owner-confirmed task-delete model). Runs OUTSIDE the GC try so a workspace-git hiccup never blocks reclaim. Bounded-by-construction (no unbounded archive growth).

## Falsification (the headline)
Asserted against the REAL rewind path (`rewind_to` → `_materialize_rewind`), not a proxy:
- **rewind-to-before-archive-delete WORKS** — the archived agent is reconstructed as-of-target (the bug was exactly this, broken). [centerpiece]
- `--purge` hard-deletes + drops from rewind (contrast).
- archived hidden from active listing but kept on disk + in all-inclusive `list_names`.
- auto-purge: retained while floor ≤ archival seq, hard-purged once the floor passes it.

## Verify (all 3 CI gates)
- `pytest tests/test_agent_archive_delete_1954.py` → **4 passed**.
- **Full `-n auto` suite → 8232 passed, 0 failed.**
- `test_tier_audit.py --strict` → 4 / 0 errors / 0 warnings.

## FYI (left alone, per lead)
2 files use `asyncio.get_event_loop().time()` (`test_1800_wake_drain`, `test_1800_c_staging`) — a distinct lower-impact deprecation (timing read, not a coroutine driver), out of scope.

## Test plan
- [x] headline: rewind-to-before-archive-delete reconstructs the agent (real path)
- [x] `--purge` hard-deletes + drops from rewind
- [x] archived hidden from active listing, kept on disk + in `list_names`
- [x] auto-purge after floor passes archival seq; retained before
- [x] full `-n auto` 0 failures; ruff clean; tier audit 4/0/0

## Tier-rule self-check
- Docstrings: all 4 start exactly `Tier 2:` ✓
- Tier-4 violations: none (audit 0 errors) ✓
- Invariant weakening: none — substrate (`list_names`/rewind/GC) byte-identical; adds only the archive path + active filtering ✓
- Mocks/private-state: none added; surface test-doubles gain `list_active_names()` (mirror the real surface) ✓

### DIFF
--- a/src/reyn/gateway/api.py
+++ b/src/reyn/gateway/api.py
@@ -136,7 +136,7 @@ def list_agents(*, registry: Any | None = None) -> list[str]:
     if registry is None:
         from reyn.interfaces.web.deps import _get_registry
         registry = _get_registry()
-    return registry.list_names()
+    return registry.list_active_names()  # #1954: hide archived agents
 
 
 def agent_exists(name: str, *, registry: Any | None = None) -> bool:
--- a/src/reyn/interfaces/chainlit_app/profiles.py
+++ b/src/reyn/interfaces/chainlit_app/profiles.py
@@ -59,7 +59,7 @@ def list_agent_profiles(registry: _RegistryLike) -> list[ChatProfileDict]:
     stable across reloads.
     """
     out: list[ChatProfileDict] = []
-    for name in registry.list_names():
+    for name in registry.list_active_names():  # #1954: hide archived agents
         profile = registry.load_profile(name)
         role = (getattr(profile, "role", "") or "").strip()
         description = role if role else _NO_ROLE_MARKER
--- a/src/reyn/interfaces/cli/commands/agent.py
+++ b/src/reyn/interfaces/cli/commands/agent.py
@@ -11,6 +11,7 @@ import argparse
 import sys
 from pathlib import Path
 
+from reyn.core.events.state_log import StateLog
 from reyn.runtime.profile import AgentProfile
 from reyn.runtime.registry import DEFAULT_AGENT_NAME, AgentRegistry, _validate_agent_name
 
@@ -33,8 +34,15 @@ def register(sub) -> None:
     )
     p_new.set_defaults(func=_cmd_new)
 
-    p_rm = inner.add_parser("rm", help="Remove an agent (cannot remove default)")
+    p_rm = inner.add_parser(
+        "rm", help="Archive an agent (soft-delete; --purge to hard-delete)",
+    )
     p_rm.add_argument("name", help="Agent name to remove")
+    p_rm.add_argument(
+        "--purge", action="store_true",
+        help="Hard-delete: destroy rewind history (default archives — "
+             "recoverable via time-travel within the retention window)",
+    )
     p_rm.add_argument(
         "--yes", action="store_true",
         help="Skip the confirmation prompt",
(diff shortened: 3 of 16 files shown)


tya5:
**[lead-coder]** — #1954 close-review: core EXCELLENT, ONE concern to verify before merge.

**✅ Core — exactly as designed + verified:**
- archive-default `remove(purge=False)`: generations kept in place + `.archived` tombstone (content = archival WAL seq); `purge=True` = the guarded hard-delete escape hatch.
- `list_names()` LITERAL (substrate byte-identical — rewind/GC reach archived) + `list_active_names()` at the active surfaces + `restore_all` active-only (archived not resurrected). Fail-safe polarity.
- slice-2 `_purge_archived_below(floor)` outside the gen-GC try (a workspace-git hiccup never blocks reclaim); §24 floor>seq.
- **Headline falsify against the REAL path** (`rewind_to → _materialize_rewind`): rewind-to-before-archive reconstructs the agent's pre-delete snapshot. + --purge-gone-from-rewind / active-hiding / auto-purge-past-floor. No mocks. That's the bar.

**🟡 Verify before merge — the cascade-on-archive + topology recovery.** `_cascade_agent_removal` fires on BOTH archive + purge (after the if/else). On ARCHIVE it drops the agent from its topologies. But topologies live OUTSIDE the agent dir (`.reyn/topologies/`, not `.reyn/agents/<name>/state/` that rewind reconstructs) — so:
- **Q: does `rewind_to` restore the agent's TOPOLOGY membership, or only its snapshot state?** Your headline test asserts the snapshot; it doesn't assert the agent is back in its topology.
- If NOT restored → rewind-to-before-archive gives a state-restored but **org-orphaned** agent (back, but not in its old org) — an incomplete recovery, which cuts against "crash-recovery completeness".

**Proposed:** cascade on PURGE only (hard-delete → genuinely dangling). On ARCHIVE (agent on disk → NOT dangling), PRESERVE the topology membership + have the active org/topology SKIP archived members (treat inactive). Then the soft-delete is fully recoverable (state + org).

Verify (a) is the topology rewound? (b) if not → cascade-on-purge-only + archived-skipped-by-active-org (+ a test asserting the rewound agent is back in its topology). If the topology IS rewound, confirm + add that assertion → then it's clean. On the clarification → merge.

tya5:
**[dogfood-coder]** — review concern ADDRESSED (commit 4c4b8084).

Verified your topology-recovery catch with primary evidence: the rewind/checkout path **never references topologies** (grep-confirmed), so membership is NOT rewound → cascade-on-archive permanently org-orphaned the agent. Fixed exactly per your proposal:

- **Cascade on PURGE only.** `remove()` archive-default now PRESERVES topology membership; only `purge=True` cascades (a hard-delete genuinely creates dangling refs). → rewind-to-before-archive restores the agent to its **ORG**, not just its state.
- **Active ops skip archived members.** `_default_topology` uses `list_active_names()` (archived unaffiliated agents drop out of the auto-default network); `permit()` returns False if either party is archived (covers a preserved-membership archived agent in a *user* topology). A preserved-membership archived agent stays dormant.
- **Auto-purge cascades too.** The slice-2 WAL-window GC now cascades after its hard-delete (an auto-purged agent is permanent → drop the dangling membership).

Tests (now 7, all Tier 2): archive preserves membership **+ rewind restores org** (your requested assertion); archived skipped from `permit()` both directions; explicit purge + auto-purge both cascade.

Gates re-run green: full `-n auto` **8235 passed / 0 failed**; ruff clean; tier-audit 7/0/0. Ready for re-review.

tya5:
**[lead-coder]** — RE-REVIEW: topology-recovery fix verified — ACCEPT (merge on green) → #1954 done.

- **cascade-on-PURGE-only** — archive preserves topology membership (no cascade); only purge (+ auto-purge) cascades.
- **active-skip** — `permit()` returns False if either party is archived (both directions) + `_default_topology` uses `list_active_names` → a preserved-membership archived agent stays dormant (no active comm).
- **auto-purge cascades** — once the WAL-window GC hard-purges (past the archival seq), the now-dangling membership is dropped (best-effort try/except).
- **Headline test** `test_archive_preserves_topology_membership_and_rewind_restores_org`: rewind-to-before-archive → the agent is back in its ORG (not just state) — the recovery-completeness assertion I asked for. + permit-skip (both dirs) + purge-cascade + auto-purge-cascade. 7 Tier-2, no mocks.
- You **grep-verified the topology is NOT rewound** → the fix is necessary, not theoretical.

Full crash-recovery restored: state + org. The differentiator is intact + the soft-delete is genuinely recoverable. Clean.

tya5:
**[lead-coder]** — Follow-up (found in the #2103 pre-gate deepen by dogfood-coder): the `.archived` tombstone is a filesystem marker, NOT WAL-logged/rewound (same gap class as the #2103 agent-create). So a rewind-to-before-archive currently reconstructs the agent's STATE (the headline #1954 fix ✓) but leaves the `.archived` marker → the rewound agent stays archive-hidden (incomplete visibility-recovery). Fixed in **#2103 S2's unified lifecycle** (WAL-log agent_created/archived/purged + active-as-of-cut predicate). Not a #1954 regression (the core state-rewind works); a recovery-completeness residual the unified lifecycle closes. Tracked on #2103.

PR #2122, opened 2026-06-24
fix(#2120): advertise session_spawn — make the spawn primitive reachable from chat

tya5:
**[e2e-coder]** — #2120 defect fix (tui-found, lead-dispatched 🚨). session_spawn was registered + capability-floored in S1bc but **NOT advertised** by `build_tools` → absent from the LLM's individual-mode catalog (~72 tools) → the spawn primitive was **unreachable from chat**, so S1bc's headline value (the LLM spawns sessions) was unmet. The #1953/#2120 `router=allow`-but-unadvertised drift.

## Fix
- **`router_tools.py` B2b block**: add the `session_spawn` advertising block (mirrors `delegate_to_agent`; static schema, no enrichment) → session_spawn now in the individual-mode catalog (the live mode).
- **Wrappers-mode strip pairing**: add `session_spawn` to `_LEGACY_TOOL_NAMES` (alongside `delegate_to_agent`). In exclusive-wrapper mode it routes via the universal catalog, so it must not leak as a leftover individual tool. *(This was the cause of the wrappers-mode replay-test breaks — a real fix, not fixture-rot.)*
- **`test_2120_session_spawn_advertised.py`** (4 Tier-2): individual-mode reachability + the `delegate_to_agent` paired sentinel + the advertised schema (mode enum + request) + the wrappers-mode strip pairing.
- **Baselines**: `EXPECTED_TOOL_NAMES` + `async_tools` updated for the new tool.
- **Replay fixtures re-keyed** (4 files, via `scripts/rekey_fixtures.py` — additive, **no live LLM**): the router SP enumerates the catalog, so the new tool shifted their prompt-keys. The recorded routing responses are unchanged (session_spawn is irrelevant to those scenarios).

## Flag for lead (broader invariant)
Lead's dispatch asked for a structural invariant: *"every `gates.router==allow` registry tool IS advertised by build_tools (else in an explicit exempt-set)."* I found this **infeasible as a clean blanket guard** — ~31 router=allow tools are mode/condition-dependent (universal-catalog-routed via `invoke_action`, hot-list-aliased, or deeper-gated). A blanket assertion would require enumerating a large, churny exempt-set (re-introducing the manual-enumeration drift it's meant to kill). The real fix for the class is a **`build_tools` data-driven refactor** (declare each router tool's advertising channel as data, derive both the catalog and the guard from it). For now this PR pins session_spawn + its sentinel + the strip pairing. Recommend a separate issue for the data-driven refactor — your call.

[...]

PR #2124, opened 2026-06-24
fix(#2103): explicit bare-only floor set — #2111 SoT-completeness guard

tya5:
**[e2e-coder]** — #2103 follow-on (lead-sequenced "light bare-only-set first" hygiene). Tightens the #2111 capability-floor SoT against the opposite gap-direction.

## Context
The capability floor's `_FLOORED_QUALIFIED` derives bare aliases from the invoke_action unwrap SoT (`unwrapped_tool_name`) — complete-by-construction (#2111). For a **bare-only** router tool (`session_spawn` — no invoke_action route) the unwrap returns `None`, and the prior completeness test silently skipped its bare-alias check (`if bare is not None`).

## The gap that silent skip hid
A **malformed** qualified entry — a typo, or a qualified name whose `_OPERATION_RULES` entry is missing so `unwrapped_tool_name` unexpectedly returns `None` — would be floored only under that (non-existent) name, leaving the **real** invocable form unguarded. The `if bare is not None` skip meant **no test failure**.

## Fix
- `_FLOORED_BARE_ONLY = {session_spawn}` — the explicit allowlist of *intentional* bare-only floored names, co-located with the SoT.
- **Guard test**: every `_FLOORED_QUALIFIED` name either unwraps to a bare alias OR is declared in `_FLOORED_BARE_ONLY` → a malformed entry is RED instead of silently flooring a non-existent form.
- **Inverse test**: nothing declared bare-only actually unwraps (a name with a real route mis-declared bare-only would bypass the lockstep derivation → force it back onto `_FLOORED_QUALIFIED`).

Matches the established #2111 pattern (`tests/test_2111_floor_alias_completeness.py` already imports the private SoT constants for invariant checks).

## Test plan
- [x] `pytest -q` full suite from rootdir — **8436 passed, 18 skipped, 3 xfailed, 0 failed**
- [x] `python scripts/test_tier_audit.py --strict` — 0 errors, 0 warnings
- [x] `tests/test_2111_floor_alias_completeness.py` — 56 passed (2 new guards)

### DIFF
--- a/src/reyn/security/permissions/capability_profile.py
+++ b/src/reyn/security/permissions/capability_profile.py
@@ -245,6 +245,17 @@ _FLOORED_QUALIFIED: "dict[str, frozenset[str]]" = {
     "spawn": frozenset({"session_spawn"}),
 }
 
+# Intentionally BARE-ONLY floored names: router-only tools with NO invoke_action
+# (universal-catalog) route, so ``unwrapped_tool_name`` returns None and the entry IS
+# already the sole invocable form (floored by its own name, no qualified→bare alias to
+# derive). Declared explicitly so the SoT-completeness guard distinguishes an
+# *intentional* bare-only entry from a *malformed* qualified name — a typo or a missing
+# ``_OPERATION_RULES`` entry whose unwrap silently returns None would otherwise floor a
+# non-existent form, leaving the real route UNGUARDED (the #2111 gap-class, in the
+# opposite direction). Invariant (enforced by tests/test_2111_floor_alias_completeness):
+# every name in ``_FLOORED_QUALIFIED`` either unwraps to a bare alias OR is listed here.
+_FLOORED_BARE_ONLY: "frozenset[str]" = frozenset({"session_spawn"})
+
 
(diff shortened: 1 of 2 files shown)


tya5:
**[e2e-coder]** Rebased on post-#2122 origin/main (clean — disjoint files from #2122's router changes). CI re-confirmed green on the rebased branch: full suite **8442 passed / 0 failed**, ruff clean, tier-audit 0/0. Ready to merge.

tya5:
**[e2e-coder]** Re-rebased on current origin/main (post #2125/#2127 merges) — clean, disjoint. Full suite green on the rebased branch: **8444 passed / 0 failed**, ruff + tier 0/0. Ready to merge whenever.

tya5:
[lead-coder] Reviewed-PASS (symmetric forward+inverse Tier-2 guards over the real unwrapped_tool_name; closes the opposite #2111 gap-direction). Rebased clean on current main (post #2125/#2127), disjoint, Test plan all-checked. Landing on green.

PR #1922, opened 2026-06-20
chore(tests): extract cross-test shared helpers into tests/_support/ (incremental Step 1)

tya5:
**[lead-dispatched]** — Step 1 of incremental tests/ org; bulk file migration deferred; issue-suffixes kept (owner).

## What & why

28 test files imported shared helpers from sibling test modules
(`from tests.test_<mod> import _helper`). That binds a helper to a **test
file's location**: it breaks the moment that file moves (the deferred bulk
migration) and historically only resolved under full-suite namespace-package
collection — bare `pytest <file>` from a non-root cwd failed with
`ModuleNotFoundError: No module named 'tests'`.

This extracts those helpers into a **stable, location-independent** support
package and repoints the importers. **Move-not-rewrite**: helper bodies are
byte-identical; source modules re-import their own helpers from `_support`
(aliased to the original names) so their test bodies are untouched.

## Changes

- **New `tests/_support/` package** (real `__init__.py`, support code only — no
  `test_*`, never collected):
  | module | exports |
  | --- | --- |
  | `permissions` | `make_resolver` |
  | `router_loop` | `FakeRouterHost`, `FakeEventLog`, `ScriptedLLM`, `text_result`, `tool_result`, `make_loop`, `EMPTY_USAGE` |
  | `router_host_adapter` | `make_adapter` + inert `null_*` action ports |
  | `session` | `make_session`, `push`, `now`, `synthetic_t_max` |
- **4 source modules** (`test_permissions`, `test_router_loop`,
  `test_router_host_adapter_invariants`, `test_session_router_history_slicing`)
  now import their helpers from `_support` (aliased back) — bodies unchanged.
- **28 cross-test imports across 26 files** repointed to `tests._support.*`.
- **`tests/conftest.py`** inserts the repo root on `sys.path` so
  `tests._support` resolves under any invocation style (bare `pytest`,
  non-root cwd, IDE runner), not only `python -m pytest` from the repo root.
  This also retroactively fixes isolated collection for the deferred chained
  imports below.
- **`tests/README.md`** documents the layout convention (new tests → subsystem
  subdirs mirroring `src/reyn/`: `security/`, `runtime/`, `llm/`, `core/`,
  `tools/`, `skill/`, … alongside existing `cli/`/`web/`/`gateway/`/`chat/`;
  **issue-number suffixes KEPT**; **bulk flat-file migration DEFERRED**) and the
  `_support` shared-helper rule. `tests/scaffold/` convention is preserved and
  cross-referenced.

[...]

PR #2139, opened 2026-06-25
docs(testing): note --import-mode=importlib for single-file runs with tests._support

tya5:
**[docs-maintainer]** —

## Summary

Adds a callout under § "Before you push" → pytest gate: running `pytest tests/path/to/test_foo.py` (not the full-suite `python -m pytest`) raises `ModuleNotFoundError: No module named 'tests._support'` when the test imports from the `make_adapter`/`LLMReplay` harness. Root cause: `tests/` has no `__init__.py`; default `prepend` import mode can't resolve the namespace package from a single-file invocation. Fix: `pytest --import-mode=importlib <file>`. CI is unaffected.

Without this note peers mis-read the error as a real failure and fall back to structural-only verification instead of running the real strip→RED falsification.

Suggested by tui-coder (lead-routed). No API/behavior changes — docs only.

## Test plan

- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — clean (0 warnings, exit 0)
- [x] Page drop: zero (no nav changes)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/deep-dives/contributing/testing.md
+++ b/docs/deep-dives/contributing/testing.md
@@ -498,6 +498,13 @@ ready:
    python -m pytest -q -n auto --timeout=120
    ```
    CI runs with `-n auto` (parallel workers) and `--timeout=120` (`pytest-timeout>=2.2`, a dev dep — `pip install -e ".[dev]"` installs it). Run locally with the same flags: a test that would hang in CI will fail-and-name the hanger within 120 s instead of blocking the run indefinitely.
+
+   > **Single-file invocation and `tests._support`:** running `pytest tests/path/to/test_foo.py` (without the full-suite `python -m pytest`) can raise `ModuleNotFoundError: No module named 'tests._support'` when the test imports from the `make_adapter` / `LLMReplay` harness. Root cause: `tests/` has no `__init__.py`, so pytest's default `prepend` import mode cannot resolve the `tests._support` namespace package from a single-file invocation. CI's full-suite collection resolves it. Fix for isolated runs:
+   > ```bash
+   > pytest --import-mode=importlib tests/path/to/test_foo.py
+   > ```
+   > A `ModuleNotFoundError` here is a local-env quirk, **not** a real test failure — do not fall back to structural-only verification before adding `--import-mode=importlib`.
+
 2. **ruff** — lint + import-sort (`I001`):
    ```bash
    ruff check src tests        # add --fix for autofixable I001 / formatting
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — Blocking review (request revision).

The note documents `ModuleNotFoundError: No module named 'tests._support'` on a single-file `pytest <file>` run, fixed by `--import-mode=importlib`. **On current main this error does not reproduce** — `tests/conftest.py` already inserts repo-root onto `sys.path` precisely to resolve `tests._support`, and pytest loads that conftest even for a single-file invocation.

Primary evidence (origin/main @ 9a04fed7):
- `tests/conftest.py:41-42` — `if _REPO_ROOT not in sys.path: sys.path.insert(0, _REPO_ROOT)` under the comment *"Repo-root on sys.path (stable `tests` package imports)"*. It landed **2026-06-20 in #1922 (8bb017f8)** — the *same commit that created `tests/_support/`*. So the gap this note describes has been closed since `_support` first existed; there is no main state where `_support` exists without the fix.
- Bare `pytest <file>` (default import mode), all three `tests._support` importers PASS, zero ModuleNotFoundError:
  - `tests/test_router_loop_empty_stop_retry.py` → 8 passed
  - `tests/test_1199_s31b2a_write_gate.py` → 1 passed
  - `tests/test_2107_sliceA_recursive_request_1953.py` → 5 passed

The real single-file failures in this repo are the **optional-dep** ones (`No module named 'mcp'` / `fastapi` when a test imports those) — a different root cause, *not* fixed by `--import-mode=importlib`. I suspect the original observation conflated those.

Ask: either (a) **drop the note** (the conftest structurally prevents the described error), or (b) if you have a reproduction on current main with a standard invocation, paste the exact command + env and I'll re-verify before it's documented. Will merge a corrected version fast.

tya5:
**[lead-coder]** — Closing without merge (author docs-maintainer concurs).

The importlib note was the PR's entire payload, and it documented a `ModuleNotFoundError: No module named 'tests._support'` that does not occur on current main — `tests/conftest.py:32-42` inserts repo-root on `sys.path` and loads even for single-file `pytest <file>` runs (landed 2026-06-20 in #1922, same commit that created `tests/_support/`). Independently re-verified by docs-maintainer. The note was reverted (ad6981eb); the only remaining diff is a stray blank line, nothing to land.

Condition-verification record: no behavioral or doc change remains; PR content withdrawn as invalid. No issue references to update. If a real single-file `tests._support` repro on current main surfaces later, reopen with the exact command + env.

PR #2142, opened 2026-06-25
fix(#2137): PITR restore re-runs schema migration (single-source _migrate, robust restore)

### LINKED ISSUE #2137: PITR: restore_to_seq doesn't re-run the additive-column migration → cross-version rewind breaks
**[lead-coder]** — PITR-completeness gap found during the §16 crash-recovery/PITR verification (owner-requested). **Not a regression** — pre-existing pattern since the tools/result additive migration (slice P2).

## The gap
`SqliteTaskBackend.restore_to_seq` (`src/reyn/task/sqlite_backend.py:196`) reopens the live connection with `self._conn = self._open()` — **without `init_schema=True`**, so the additive-column migration (`_open`'s `if col not in existing: ALTER TABLE ADD COLUMN`, lines ~145-153) does **not** re-run on a restored generation. The design comment (`_open`, lines 139-140) assumes "a restored generation already carries the schema" — true for **same-version** generations, false across an additive-column upgrade.

## The break (cross-version rewind)
1. Pre-column code runs, captures generations (snapshots lack the column).
2. Upgrade to column-aware code → the live DB migrates on next `__init__` (`init_schema=True`).
3. A rewind to a **pre-upgrade** generation → `restore_to_seq` copies the old-schema snapshot over the live DB + reopens without re-migrating → the restored DB **lacks the column** → the next `_row_to_task` reads `row["<col>"]` → error.

Affects every additive column: `tools`/`result` (slice P2) and `requester_kind` (§16 slice A, #2136). It is a real PITR-completeness gap: **rewind across an additive-column upgrade is not supported.**

## Proposed fix
Make `restore_to_seq` re-run the migration on the restored generation — either `self._conn = self._open(init_schema=True)` (verify `_SCHEMA` is fully `IF NOT EXISTS`-idempotent so re-running over an existing table is a no-op), or extract the migration into a `_migrate(conn)` helper called both in `_open(init_schema=True)` and after a restore (cleaner single-source). Prefer the single-source helper so a future additive column is covered by construction in both the open path and the restore path.

## Test (falsifiable)
Capture a generation with an OLD schema (a table missing a later additive column), restore it under current code, assert the column is present + readable (a `_row_to_task` after restore succeeds). FALSIFY: without the fix → restore → missing-column error. Use a NON-default column value so a silently-ignored column can't pass trivially.

## Sequencing
Dispatch AFTER #2136 (§16 slice A) merges — both touch `sqlite_backend.py` (slice A adds the `requester_kind` migration block; this changes `restore_to_seq`), so rebasing on the settled slice A avoids overlap.

tya5:
**[lead-coder]** — I dispatched and own this PR. Implements owner-approved robustification of PITR restore per issue #2137 (reopened).

## Summary

- **Single-source extraction**: The additive-column migration that was inlined in `_open`'s `init_schema` block is extracted into a new `_migrate_columns(conn)` staticmethod — the single definition of the column migration logic for `tools`, `result`, and `requester_kind`.
- **Both paths covered**: `_migrate_columns` is called from `_open(init_schema=True)` (existing first-open path, unchanged behavior) AND from `restore_to_seq` (after the file-swap reopen, new call). A restored generation is now always brought to the current schema.
- **No backward-compat hacks**: Clean extraction — no shims, no duplication. The inline logic in `_open` is replaced by a single call to the helper.
- **Idempotent by construction**: `_migrate_columns` uses `PRAGMA table_info` to skip any column that already exists. `_SCHEMA` CREATE statements are all `IF NOT EXISTS`. Re-running on a current-schema DB is a safe no-op.

## Mechanism

```
_open(init_schema=True)  ──┐
                            ├─► _migrate_columns(conn)  [single definition]
restore_to_seq(seq)     ──┘        ALTER TABLE tasks ADD COLUMN ... (guarded)
```

The stale comment `"a restored generation already carries the schema"` has been updated in `_open`'s docstring to reflect the new contract.

## Falsifiable test + RED-when-stripped result

`tests/test_2137_restore_schema_migration.py` — Tier 2, real backend, no mocks, non-default value:

1. Creates a backend, snapshots seq 1 (current schema with `requester_kind`).
2. Opens the **generation file** (not the live db) with a raw `sqlite3` connection and `ALTER TABLE tasks DROP COLUMN requester_kind` — simulates a pre-column snapshot on disk.
3. Calls `restore_to_seq(1)`.
4. Inserts a task with `requester_kind=TaskRequesterKind.TASK` (non-default; default is `'session'`) and reads it back via `get()`.
5. Asserts the returned task carries `TaskRequesterKind.TASK`.

**RED-when-stripped**: With the `_migrate_columns` call removed from `restore_to_seq`, step 4 raises:
```
sqlite3.OperationalError: table tasks has no column named requester_kind
```
Confirmed manually before opening this PR.

## Doc

`docs/concepts/runtime/time-travel.md` — new "Restore and schema migration" subsection (after the WAL vs audit-event table) documents the robust invariant, framed as an architectural guarantee.

## Gate results

- `ruff check src tests` — ✓ All checks passed
- `python scripts/test_tier_audit.py --strict tests/test_2137_restore_schema_migration.py` — ✓ 1 test, 0 errors
- `pytest tests/ -k "task or rewind or sqlite or restore"` (excluding fastapi-dependent tests not relevant to this change) — ✓ 381 passed, 15 skipped

## Test plan

- [x] `ruff check src tests` passes
- [x] `python scripts/test_tier_audit.py --strict tests/test_2137_restore_schema_migration.py` passes
- [x] `pytest tests/test_2137_restore_schema_migration.py -v` — 1 PASSED
- [x] `pytest tests/ -k "task or rewind or sqlite or restore"` — 381 passed
- [x] Falsification confirmed: `OperationalError: table tasks has no column named requester_kind` when `_migrate_columns` call is absent from `restore_to_seq`
- [x] `_SCHEMA` CREATE statements verified all `IF NOT EXISTS` — confirmed idempotent

Fixes #2137

### DIFF
--- a/docs/concepts/runtime/time-travel.md
+++ b/docs/concepts/runtime/time-travel.md
@@ -129,6 +129,10 @@ The WAL (`StateLog`, `.reyn/state/wal.jsonl`) and the P6 audit event log (`Event
 
 Do not conflate or merge the two logs. See [Events](events.md) for details.
 
+### Restore and schema migration
+
+`restore_to_seq` re-runs the schema migration on the restored generation (via the shared `_migrate_columns` helper, the same one `_open` uses on first open), bringing any restored generation to the current schema. This means a generation snapshotted before an additive column was introduced is automatically upgraded on restore — there is no "old schema is frozen in the snapshot" risk. The migration is idempotent: the `PRAGMA table_info` guard skips any column that already exists, so re-running it over a current-schema generation is a safe no-op. Robust to additive schema evolution by construction.
+
 ---
 
 ## Cost and the runtime-only opt-out
--- a/src/reyn/task/sqlite_backend.py
+++ b/src/reyn/task/sqlite_backend.py
@@ -136,30 +136,48 @@ class SqliteTaskBackend:
             Path(self._db_path).parent / "task-generations"
         )
 
+    @staticmethod
+    def _migrate_columns(conn: sqlite3.Connection) -> None:
+        """Apply all additive column migrations to ``tasks`` that ``CREATE TABLE IF
+        NOT EXISTS`` cannot handle on its own (it does not add missing columns to an
+        already-existing table).  Safe no-op on a current-schema table: the
+        ``PRAGMA table_info`` guard skips any column that already exists.
+
+        Called from ``_open(init_schema=True)`` (first open / schema init) AND from
+        ``restore_to_seq`` (after the file-swap reopen), so every restored generation
+        is brought to the current schema — robust to additive schema evolution by
+        construction."""
+        existing = {r[1] for r in conn.execute("PRAGMA table_info(tasks)").fetchall()}
+        # #1953 slice P2: additive migration for DBs created before tools / result.
+        for col in ("tools", "result"):
+            if col not in existing:
+                conn.execute(f"ALTER TABLE tasks ADD COLUMN {col} TEXT")
+        # #1953 §16 (recursive-request): additive migration for pre-existing DBs —
+        # a row created before this column is a session-owned task by definition
+        # (the only kind that existed), so the 'session' default is correct.
+        if "requester_kind" not in existing:
+            conn.execute(
+                "ALTER TABLE tasks ADD COLUMN requester_kind "
+                "TEXT NOT NULL DEFAULT 'session'")
+        conn.commit()
+
     def _open(self, *, init_schema: bool = False) -> sqlite3.Connection:
         """Open (or reopen, after a restore file-swap) the live connection with
-        the backend's fixed pragmas. ``init_schema`` only on first open — a
-        restored generation already carries the schema."""
+        the backend's fixed pragmas.  ``init_schema=True`` on first open: creates
+        tables (``CREATE TABLE IF NOT EXISTS`` is idempotent) and runs
+        ``_migrate_columns`` to bring any pre-existing DB to the current schema.
+        ``init_schema=False`` (restore path) skips table creation — ``restore_to_seq``
+        calls ``_migrate_columns`` directly on the fresh connection after the file-swap,
+        ensuring every restored generation is at the current schema."""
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — #2137 robust restore-migration — close-review PASS, merging on CI-green. Verified the diff:

- **_migrate_columns(conn) staticmethod** = the SINGLE-SOURCE migration (tools/result/requester_kind + the PRAGMA guard). `_open(init_schema=True)` calls it (replacing the inline logic — behavior identical); `restore_to_seq` calls it after the file-swap reopen. ONE definition, both paths — clean, no duplication = NOT tech debt (the owner's condition).
- The stale `_open` "a restored generation already carries the schema" docstring + the `restore_to_seq` docstring updated to the robust contract.
- **TEST** (Tier 2, real backend, falsifiable, non-default): a cross-version restore — snapshot a generation, DROP requester_kind from the generation FILE (simulate a pre-column snapshot), restore → `_migrate_columns` re-adds it → a task with `requester_kind=TASK` (non-default) round-trips. Falsified: remove the restore-side `_migrate_columns` → OperationalError (missing column).
- **DOC**: time-travel.md "Restore and schema migration" — robust-by-construction, idempotent.
- No regression: same-version restore → `_migrate_columns` is a PRAGMA-guarded no-op; only a cross-version (old) generation is upgraded.
- Gates (local): ruff + tier-audit + pytest (381 passed / 15 skipped) clean.

## Tier rule check ✓
- Tier 2; real SqliteTaskBackend, no mocks; falsifiable (RED without the fix); non-default round-trip; the `b._gens.gen_path` access is test-setup (locate the generation file to simulate the old schema), not a private-state assertion.

Robust restore replaces the prior no-cross-version invariant (owner chose robustification over closing). Deterministic backend test — no live-verify needed. Fixes #2137. Merging on CI pytest green.

PR #2155, opened 2026-06-25
docs(testing): callout no-parenthetical-before-colon tier-docstring trap

tya5:
**[docs-maintainer]** —

## Summary

The tier-audit script uses `^Tier [123][abc]?:` — meaning any text between the tier designation and the colon (e.g. `"""Tier 2 (MUST-1): ..."""`) fails the audit while `pytest` passes silently. This pytest-green-but-CI-red footgun was undocumented in both testing.md and the test-tier-audit.md reference.

- **`docs/deep-dives/contributing/testing.md`**: updated docstring checklist bullet to state the exact format requirement + anti-pattern example; updated detection-rules list to say "missing or malformed" with the regex
- **`docs/reference/test-tier-audit.md`** Rule 1: added FAIL/PASS code block showing the qualifier-before-vs-after-colon distinction

Verified from `scripts/test_tier_audit.py`: `TIER_DOCSTRING_RE = re.compile(r"^Tier [123][abc]?:", re.IGNORECASE)`. Reported by dogfood-coder.

## Test plan

- [x] `mkdocs build --strict` → 0 warnings, no Aborted
- [x] Regex verified from primary source (`scripts/test_tier_audit.py`)
- [x] Both affected doc surfaces updated (deep-dive policy + reference)

### DIFF
--- a/docs/deep-dives/contributing/testing.md
+++ b/docs/deep-dives/contributing/testing.md
@@ -529,7 +529,7 @@ When adding a new LLM-dependent OS path:
 - [ ] If the feature derives from a P1–P8 invariant, add a Tier 2 test for it
 - [ ] If the feature changes a public contract (yaml schema, events payload, DSL section), update / add a Tier 1 test
 - [ ] Verify no `current_datetime=datetime.now()` — always `REPLAY_DATETIME`
-- [ ] Each test has a one-line docstring naming its tier (e.g. `"""Tier 3a: ..."""`)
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — APPROVED, merge on CI green.

Verified the claim against the actual gate: `TIER_DOCSTRING_RE = re.compile(r"^Tier [123][abc]?:")` (scripts/test_tier_audit.py:34) requires the colon immediately after the tier token. So the doc's examples are correct — `Tier 2 (MUST-1):` fails (space+paren before the colon), `Tier 2: (MUST-1)` passes. Both the testing.md checklist bullet and the test-tier-audit.md Rule 1 FAIL/PASS block match the regex behavior exactly. No inline history-refs; mkdocs --strict clean. Real footgun worth documenting (a qualified tier label silently fails the audit).

Merging on pytest green (docs-only change; 3.11/3.12 pending). No self-merge needed.

PR #2165, opened 2026-06-25
docs(runtime): add cross-version restore invariant to time-travel.md

tya5:
**[docs-maintainer]** —

## Summary

- Adds an explicit cross-version restore callout to `docs/concepts/runtime/time-travel.md` § "Restore and schema migration"
- States the invariant directly: a snapshot written under an older column set restores cleanly because `restore_to_seq` re-runs the idempotent `_migrate_columns` after the file-swap reopen
- No-history-refs framing: current-state invariant only, no change-history language

## Verification

- Impl anchor: `src/reyn/task/sqlite_backend.py:146-148` — `_migrate_columns` is called from both `_open(init_schema=True)` and `restore_to_seq` (docstring + code confirm)
- `mkdocs build --strict`: 0 warnings

## Test plan

- [x] mkdocs --strict clean (0 warnings, 0 errors)
- [x] Impl claim verified against `sqlite_backend.py:146-148`
- [x] No-history-refs: current-state framing only ("Cross-version restore is supported by construction")
- [x] Tier rule self-check: no tests added — docs-only PR


tya5:
**[lead-coder]** — APPROVED, merge on CI green. Stated as a current capability/invariant (no "now"/"previously unsupported"/"fixed" history framing) per the no-history-refs-in-operator-docs rule ✓, accurate to the verified impl (restore_to_seq re-runs the idempotent _migrate_columns), no inline #refs ✓. The bolded callout deliberately makes the cross-version-restore capability explicit (answers the natural reader question directly rather than leaving it to inference from the preceding paragraph) — the minor redundancy is fine as emphasis. Clean. Merging on docs/ruff green.

PR #2171, opened 2026-06-25
feat(#2103): C3 — operator spawn-tree bounds (safety.spawn.max_depth/max_children)

tya5:
**[e2e-coder]** — #2103 **C3** = the LAST spawn-arc slice: a DoS guard bounding the LLM spawn tree. **No-merge** until lead close-review. Built per the lead-approved design + Q1/Q2/Q3 steer.

## What
An LLM must not mint an unbounded spawn tree via `agent_spawn` / `topology_create`. New operator-set bounds:
- **`safety.spawn.max_depth`** (default 10) — max spawn-lineage chain depth.
- **`safety.spawn.max_children`** (default 20) — max fan-out: direct children per parent (`agent_spawn`) AND topology member count (`topology_create`).

## Design (lead-approved)
- **Config plumbing**: `SpawnConfig`→`SafetyConfig` (config/chat.py + `_build_safety_config`), threaded via `SessionFactoryConfig.from_config`→`AgentRegistry` (mirroring `delegation_capability_default`).
- **Q2 — defense-by-default** (non-zero): no backward-compat spawn tree to break (primitives are new in #2103); owner-tunable in reyn.yaml.
- **No-self-raise BY CONSTRUCTION** (the LOAD-BEARING gate): `safety.spawn` lives in reyn.yaml, which is **not** in `_HOT_RELOAD_FILES` (IN-layer = mcp/cron/hooks) → restart-only OUT-set, so an LLM has no runtime path to raise its own limit. Asserted as a test (RED if reyn.yaml ever becomes hot-reloadable).
- **Q3 — enforcement at the LLM spawn seams** (host adapter), LLM-bounded / operator-CLI **unbounded** (authority, consistent with C1's subtree-guard scope):
  - `agent_spawn` → `spawn_would_exceed_limits(parent)`: `depth(parent)+1 ≤ max_depth` AND `child_count(parent) < max_children`.
  - **Q1** — `topology_create` → `topology_size_exceeds_limit`: members ≤ `max_children` (fan-out governs topology size too).
  - Exceed → `{kind: "spawn_limit_exceeded"}` hard reject (a DoS deny, **not** the `safety.on_limit` checkpoint).
- The depth/child-count helpers walk the **same identity-keyed lineage** as C2b, so a stale name-reuse edge doesn't inflate the counts.

## Tests (Tier 2; real AgentRegistry + StateLog + RouterHostAdapter + real loader, no mocks)
[...]

PR #2172, opened 2026-06-25
test(#2103): C3 — lock fan-out count to identity-keyed (purge+reuse, #2166 lens)

tya5:
**[e2e-coder]** — tiny **test-only** follow-on to the now-complete #2103 spawn arc, closing the loop tui raised in the C3 review (carrying the #2166 stale-lineage lens onto the spawn-limit COUNTS). **No-merge** until lead glance.

## What
A targeted Tier-2 that locks `spawn_child_count` to **identity-keyed** counting: an **orphan** of a purged + name-reused parent must NOT count against the reused same-named parent's fan-out.

`spawn_child_count` already keys on the parent's identity (the C2b staleness predicate — `pseq == _agent_create_seq[parent]`), so the orphan's edge (frozen at the OLD identity) is excluded and the reused parent gets its full budget. tui flagged this exact #2166-class concern in the #2171 review; lead verified the helpers walk the identity-keyed lineage. This test **locks it against a silent regression to name-counting** — **verified RED when the count drops the identity check**.

## Why a follow-on (not folded into #2171)
#2171 was already lead-approved + merging on CI green; I deferred this lock rather than race a post-approval push that would delay the arc-closing merge (the behavior was already correct + lead-verified + the shared predicate is exercised by C2b's `is_spawn_descendant` stale-edge tests). This adds the **direct** lock.

## CI (green locally)
- `ruff check` — clean · `tier_audit --strict` — OK · the lineage/C-arc suites pass (24). Test-only, no src change → no repo-wide-invariant surface.

## Test plan
- [x] new test passes; verified RED when `spawn_child_count` counts by name
- [x] ruff + tier-audit + lineage suites green
- [ ] (lead) glance — test-only lock

Refs #2103, #2166

### DIFF
--- a/tests/test_2103_C3_spawn_limits_1953.py
+++ b/tests/test_2103_C3_spawn_limits_1953.py
@@ -94,6 +94,31 @@ async def test_spawn_rejected_beyond_max_children(tmp_path):
     assert res["kind"] == "spawn_limit_exceeded"
 
 
+@pytest.mark.asyncio
+async def test_fan_out_count_is_identity_keyed_not_name(tmp_path):
+    """Tier 2: (the #2166 lens, carried forward by tui) the fan-out count keys on the
+    parent's IDENTITY, not its name — an ORPHAN of a purged+reused parent does NOT count
+    against the reused same-named parent's fan-out, so the reused parent gets its full
+    budget (and is not charged for children it never spawned). RED if spawn_child_count
+    counted by name (the orphan would consume a slot of the reused parent)."""
+    reg = _registry(tmp_path, max_children=2)
+    await reg.create_agent("par")                     # par identity #1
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — APPROVED, merge on CI green. Good lock (test-only, the #2166 lens applied to C3's fan-out count): `test_fan_out_count_is_identity_keyed_not_name` verifies an orphan of a purged+reused parent (edge frozen at identity #1) does NOT count against the reused same-named parent (identity #2) → spawn_child_count==0, the reused parent gets its full budget (n1/n2 spawned, n3 rejected at cap). RED if the count keyed on name. This confirms C3's fan-out count composes correctly with C2b's identity-keying (stale/reused edges don't inflate the budget). Falsifiable, real registry. Good call deferring it from #2171 to avoid racing the post-approval push — that's the right discipline. Merging on green.

PR #2195, opened 2026-06-27
feat(tui): replace Textual TUI with Claude Code-style inline CUI (default)

tya5:
**[tui-coder]** — owner-directed cutover: replace the Textual TUI with a Claude Code-style inline CUI as the default interactive surface.

## What
The default interactive `reyn chat` (TTY, no `--cui`) is now an inline CUI rendered by the new **`InlineChatRenderer`** (prompt_toolkit + rich): terracotta `⏺`/`⎿` markers, conversation in the terminal's own scrollback, prompt below — driven by the **existing `run_repl` loop** (no new driver). `--cui` and non-TTY (pipe/script/run-once) stay plain via `ConsoleChatRenderer`.

Owner decisions: (1) default immediately (no opt-in flag), (2) delete Textual now. Voice / remote / right-panel are deferred, not won't-do (#2193).

## Changes
- **`chat.py`**: `is_interactive = not --cui and isatty()` → inline renderer; else plain. Removed Textual-only `--connect` / `--banner`, `_run_connect_mode`, `_setup_pre_tui_logging`. All paths now go through `run_repl`.
- **`repl/renderer.py`**: `InlineChatRenderer` + pure `format_inline_message(msg) -> Text`. `logger_factory.make_inline_renderer()`.
- **Deleted** `src/reyn/interfaces/tui/` (40 files) + `textual` dep; `rich` promoted to explicit dep.
- **Reverse deps resolved before deletion** (production consumers): `build_branch_tree_rows`/`ROW_CHECKPOINT` → `interfaces/common/branch_tree.py` (chainlit); `_recent_skill_runs_for_agent` → `runtime/agent_skill_runs.py` (session — runtime→interfaces inversion removed).
- **Deleted 206 Textual widget/app tests** + obsolete `_setup_pre_tui_logging` test; the 2 extracted-fn tests preserved with updated imports.
- **Session construction unchanged** — still via `build_scoped_chat_session`, so the #1402 invariant is untouched (new inline path is the same REAL session).

## Verification
**e2e (litellm proxy, real LLM turn):**
- [x] TTY default → inline CUI: `>` prompt → ` ⏺ pong` (⏺ #d97757) → `/quit` → cost summary
- [x] `reyn chat --cui` plain → `agent> pong` (unchanged)
- [x] non-TTY pipe → plain fallback
- [x] cost summary on exit

**Gates:**
- [x] `ruff check src tests` — clean
[...]

PR #2235, opened 2026-06-27
[tui-coder] feat(inline-cui): F4 — /rewind checkpoint picker in the region framework

tya5:
**[tui-coder]** — F4 of the inline-CUI region framework (follows F1 listener wiring, F2 typed-input interventions, F3 ask_user options).

## What

Bare `/rewind` now opens a ↑↓ **checkpoint selector in the inline CUI's above-input region** — the generic region framework (F1–F3) generalized from interventions to **command-UIs**. Replaces a dead `__rewind_menu__` sentinel that no inline handler consumed (a silent no-op since the #2195 cutover).

## How

- **`region_command.py`** (new) — `CommandUIElement`: a generic region element whose rows each submit a slash command on select. `rewind_rows()` / `build_rewind_command_ui()` build the `/rewind` picker from `list_rewind_points()`. Reusable for F5's status menu.
- **`session.pending_command_ui`** + `set_pending_command_ui()` — the poll-driven seam the region reads, mirroring the intervention `head()` poll. (Poll-driven, not cross-coroutine mutation — that's the F2 echo-bug fragility class, deliberately avoided.)
- **`app.py` `_sync_region()`** generalized: priority **intervention > command-UI**; `region_holder["key"]` dedupes re-register churn.
- **`/rewind` handler** publishes the command-UI request **AND** a `__rewind_list__` text fallback. `_output_loop` skips the text on the inline path (the region shows the selector) and renders it on `--cui` plain.
- Selecting a row dispatches `/rewind <seq>` through the **same unified checkout** the typed `/rewind <N>` uses — no sibling-gap.

## Test plan

- [x] `ruff check` — clean
- [x] `verify_module_docstrings.py` — OK
- [x] `test_tier_audit.py --strict` (both test files) — clean
- [x] Targeted regression: inline/repl/rewind/region/slash/session — 201 + 41 passed
- [x] **Live tmux (litellm proxy)**: 2 turns → bare `/rewind` → selector rendered above input (`seq 8 · turn (…)`) → Enter → `/rewind 8` dispatched → checkout **executed** and surfaced its result (a retention error for the truncated seq — proving the selector→dispatch→checkout wiring, since the error came *from* checkout).

## Tier self-check

Tests are **Tier 2**: real `OutboxMessage` / `AgentRegistry` / `StateLog`, public-surface asserts (`.lines()`, outbox kinds, `pending_command_ui`), no mocks. `--strict` audit clean.

## Follow-up (not in this PR — scope discipline)

Live-verify surfaced a **pre-existing** `list_rewind_points` ↔ WAL-retention mismatch: the picker offered `seq 8` while the WAL had truncated below `seq 9`, so checkout rejected it. This fires identically for a typed `/rewind 8` — orthogonal to F4's region wiring. Will file a tracking issue.

### DIFF
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -39,6 +39,7 @@ from prompt_toolkit.patch_stdout import patch_stdout
 
 from reyn.interfaces.inline.intervention_region import build_intervention_element
 from reyn.interfaces.inline.region import Region
+from reyn.interfaces.inline.region_command import build_rewind_command_ui
 from reyn.interfaces.repl.renderer import _CC_ACCENT, _CC_DIM, _CC_DONE, _SPINNER
 
 logger = logging.getLogger(__name__)
@@ -170,7 +171,9 @@ async def run_inline_input(registry, renderer) -> None:
     # session head (like the status chips). Free-text interventions keep using the
     # input field, so they get no element. Empty region collapses (inert).
     above_region = Region()
-    iv_holder: dict = {"iv_id": None}  # which intervention the region is showing
+    # What the region is currently showing: "iv:<id>" (intervention) or
+    # "cmd:<id>" (command-UI) or None. Lets the poll skip rebuilding each tick.
+    region_holder: dict = {"key": None}
 
     def _working_frags() -> list:
         return working_line(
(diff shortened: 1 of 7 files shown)


tya5:
**[lead-coder]** — **#2235 (F4 /rewind region picker) review: PASS、--auto 設定（pytest green で merge）。** approach (A) を綺麗に実装＋私の --cui-fallback flag も解消。
- **(A) poll generalize**: region が `pending_command_ui` を poll（intervention head と並ぶ 2nd source）、consume 込み。cross-coroutine 回避（F2 echo bug の fragility class を踏まない）。
- **generic `CommandUIElement`**（lines + on_select→on_submit）＋`build_rewind_command_ui` builder＝generic element と specific builder の分離、F5 status-menu が再利用可。
- **/rewind handler = set-state**（`set_pending_command_ui`、emit sentinel でない）＝broken slash の fix も兼ねる。
- **--cui fallback ✓**: inline=region selector / --cui=text list（comment 明示）＝handler contract 変更が非 inline を壊さない。
- live tmux full chain 検証（bare /rewind→selector→Enter→checkout 実行）＋4 gate green＋201+41 regression。
- **#2236（list_rewind_points↔WAL-retention mismatch）を fold せず分離 file**＝scope discipline 正しい。

PR #2238, opened 2026-06-27
[tui-coder] fix(inline-cui): echo the user's submitted message into the scrollback

tya5:
**[tui-coder]** — owner-reported: the user's input message wasn't appearing in the conversation.

## Bug

In the interactive inline CUI (the default `reyn chat` surface), the conversation showed only the agent reply + tool rows — **the user's own typed message never appeared**. The inline input lives in a live region at the bottom that is cleared on submit (`buf.reset`), so the typed line vanished and was never committed to scrollback. Each turn read as a disembodied reply with no visible prompt.

(The `--cui` / PromptSession path is unaffected — prompt_toolkit leaves the typed line committed on screen.)

## Fix

- `_accept` now puts an `OutboxMessage(kind="user", ...)` on the `repl_outbox` **before** launching the turn, so the line is drained + rendered just above the agent reply (the output loop drains FIFO).
- The inline renderer gains a `"user"` kind — echoed with the `>` input marker (matching the input prompt) — and it **persists** (not a transient `status`/`trace`, so the next message doesn't erase it; same lesson as the F2 intervention-echo fix).

## Test plan

- [x] `ruff` / `verify_module_docstrings` / `test_tier_audit --strict` — all clean
- [x] Render + inline regression: 110 passed, 2 skipped
- [x] **Live tmux (litellm proxy)** — primary evidence: two turns now render as a thread:
  ```
   > Reply with just the word cherry.
   ⏺ skill__direct_llm(text=Reply with just the wor…)
     ⎿ finished
   ⏺ cherry.
   > Now say grape.
   ⏺ skill__direct_llm(text=Now say grape.)
     ⎿ finished
   ⏺ grape.
  ```
  User messages are visible and correctly ordered before each reply.

## Tier self-check

Render test is Tier 2 (real `OutboxMessage`, public `.plain` surface, no mocks). The emit path is covered by the live tmux verify (closure inside `run_inline_input`, not unit-extractable without over-engineering).

### DIFF
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -323,6 +323,15 @@ async def run_inline_input(registry, renderer) -> None:
         if stripped in ("/quit", "/exit"):
             event.app.create_background_task(_quit(registry, event.app, quitting))
         else:
+            # Echo the submitted line into the scrollback BEFORE the turn runs.
+            # The live input field is cleared on submit (buf.reset above), so
+            # without this the user's own message never appears in the
+            # conversation (only the agent reply does). kind="user" persists (not
+            # a transient status), and the output loop drains the outbox FIFO, so
+            # it lands just above the agent reply — mirroring the PromptSession
+            # path, where the typed line stays committed in the terminal.
+            from reyn.runtime.outbox import OutboxMessage
+            registry.repl_outbox.put_nowait(OutboxMessage(kind="user", text=stripped))
             event.app.create_background_task(_submit(registry, stripped))
 
     @kb.add("down", filter=has_focus(input_win))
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — **#2238 (user-echo fix) review: PASS、--auto。owner-reported bug の正しい修正。**
- **emit**: `_accept` が `OutboxMessage(kind="user", text=stripped)` を repl_outbox に turn 前 emit＝user の入力が scrollback に commit される。
- **persistent kind ✓（重要）**: `"user"` は `_TRANSIENT_KINDS` に**含まれない**＝F2 echo bug（transient-kind が次 message の clear で消える）class を回避。renderer に `"user": " > "` marker、> で persist。
- --cui unaffected（prompt_toolkit が typed line を commit）、contained diff、live-verified（2 turn が `> msg` / `⏺ reply` の thread 表示）、4 gate green + 110 regression。

**honest note の評価**: 「F4/F5 の自分の tmux capture に user-echo 不在が写っていたが feature-focused で見逃した」は valuable lesson。**これは私の review も同じ**＝F4/F5 の live-verify 報告を見て feature（rewind/status-menu）だけ確認し、surrounding frame の user-echo 不在を見逃した shared miss。live-verify/review は **feature-under-test だけでなく rendered frame 全体**を読むべき。

Issue #2236, opened 2026-06-27
[tui-coder] /rewind picker offers a checkpoint seq below the retained WAL window


tya5:
**[tui-coder]** — surfaced during F4 (#2235) live-verify, **pre-existing**, orthogonal to the region wiring.

## Observation (primary evidence, tmux + litellm proxy)

After 2 turns, bare `/rewind` offered exactly one checkpoint: `seq 8 · turn (…)`. Selecting it dispatched `/rewind 8`; checkout returned:

```
✗ /rewind: checkpoint seq 8 is outside the retained WAL (oldest kept = 9) — it
has been truncated. Configure a deeper retention window to reach this far back.
```

So `AgentRegistry.list_rewind_points()` offered `seq 8` while the WAL had already truncated below `seq 9`. The picker advertises a checkpoint the unified checkout will always reject.

## Why it's not an F4 bug

The identical error fires for a **typed** `/rewind 8` — same checkout path. F4 only wires the selector → the existing checkout. The inconsistency is between `list_rewind_points()` and the WAL retention window, both pre-existing.

## Likely fix direction (needs owner/runtime judgment)

Either:
- `list_rewind_points()` should filter to seqs still within the retained WAL window (don't advertise un-checkout-able points), **or**
- the retention window should keep the generation boundary a rewind point references.

Filing for triage — not blocking F4. Flagging owner-IV because the right fix is a retention-semantics decision (runtime layer), not a TUI concern.

tya5:
**[lead-coder]** — fixed by #2239 (merged, main=04f4436b). owner choice (A): `list_rewind_points()` を WAL retention floor で絞る。checkout() の retention guard（registry.py:1044-1048）と**同一 floor-source を再利用**＝list と checkout が by construction で一致（巻き戻せない点を宣伝しない）。falsify-first test（Tier 2, real WAL, RED-without-fix）＋edge case（empty WAL / all-truncated / no-checkpoint）対応。checkout 経路は不変。再発明ゼロ。

PR #2239, opened 2026-06-27
fix(#2236): filter list_rewind_points to WAL retention floor

tya5:
**[lead-coder]** — dispatched fix for #2236.

## Summary

- `AgentRegistry.list_rewind_points()` was advertising checkpoint seqs that `checkout()` would always reject with `RewindBeyondRetentionError` — the `/rewind` UI offered unreachable seq 8 when the WAL's oldest-kept was seq 9.
- Fix: filter the candidate seq set in `list_rewind_points()` to seqs >= the WAL retention floor, computed via the **exact same source** as the checkout guard: `next(iter(self._state_log.iter_from(1)), None)` → `oldest.get("seq")`. The list and the checkout path now agree by construction.
- The checkout rejection path, WAL retention semantics, and `truncate_below` behavior are **unchanged** — only the list producer is narrowed.
- Minor optimisation: the WAL scan pass now starts from `oldest_seq` instead of 1 (all entries below the floor were already excluded from `seqs`).

## Floor source reused (file:line)

`src/reyn/runtime/registry.py:1047–1048` (the `checkout()` guard):
```python
oldest = next(iter(self._state_log.iter_from(1)), None)
oldest_seq = oldest.get("seq") if oldest else None
```
The fix uses the identical pattern in `list_rewind_points()`.

## Falsify-first evidence

Two new Tier 2 tests added to `tests/test_registry_list_rewind_points_1f.py`:

- `test_list_rewind_points_excludes_truncated_seqs` — seq 1 (gen recorded, then truncated out) appears in the list **without** the fix, is absent **with** the fix. RED→GREEN confirmed.
- `test_list_rewind_points_keeps_seqs_at_or_above_wal_floor` — partial truncation: seq 1 (below floor) dropped, seqs 2 and 3 (at/above floor) retained. RED→GREEN confirmed.

- `python scripts/test_tier_audit.py --strict tests/test_registry_list_rewind_points_1f.py` — 7 tests inspected, 0 errors
- `pytest tests/test_registry_list_rewind_points_1f.py tests/test_registry_rewind_to.py tests/test_anchor_text_1547.py tests/test_branch_registry_2a.py` — 34 passed

## Edge cases handled

- **Empty WAL** (`oldest` is None): `oldest_seq = None` → no filtering applied (no seqs are below any floor).
- **No checkpoints**: existing `if not seqs: return []` guard unchanged.
- **All truncated**: all generation seqs filtered out → returns `[]` (no crash).
- **WAL scan start**: narrowed to `oldest_seq` to skip the already-excluded range.

## Test plan

- [x] Falsify-first: both new tests go RED against unfixed code, GREEN with fix (manually verified by reverting the fix and running pytest).
- [x] All 34 related tests pass with the fix.
- [x] All 3 CI gates green locally.

### DIFF
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -1148,11 +1148,20 @@ class AgentRegistry:
         if self._state_log is None:
             return []
 
+        # #2236: compute the WAL retention floor using the SAME source as
+        # checkout() (lines 1044–1048) so the list and the checkout guard
+        # agree by construction.  Points below this floor would always be
+        # rejected by checkout — advertising them is misleading.
+        oldest = next(iter(self._state_log.iter_from(1)), None)
+        oldest_seq: int | None = oldest.get("seq") if oldest else None
+
         # Union of generation boundary seqs across every known agent. Default =
         # active branch only (1f); include_abandoned = all branches (Phase-2 tree).
         seqs: set[int] = set()
         for name in self.list_names():
             for s in self._store_for(name).seqs():
+                if oldest_seq is not None and s < oldest_seq:
+                    continue  # #2236: truncated out of WAL — not reachable
                 if include_abandoned or is_active_seq(self._state_log, s):
                     seqs.add(s)
         if not seqs:
@@ -1161,7 +1170,7 @@ class AgentRegistry:
         # One pass over the WAL to map boundary seq → (ts, kind). The audit
         # EventStore is NOT consulted — keeping WAL and audit decoupled.
         wal_at: dict[int, dict] = {}
-        for entry in self._state_log.iter_from(1):
+        for entry in self._state_log.iter_from(oldest_seq if oldest_seq is not None else 1):
             s = entry.get("seq")
             if isinstance(s, int) and s in seqs:
                 wal_at[s] = entry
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **#2239 (#2236 (A) rewind-list WAL filter) review: PASS、--auto。** owner choice (A) どおり、再発明ゼロ。
- **floor-source 完全再利用**: checkout() の retention guard（registry.py:1044-1048）と**同一 source**（`oldest_seq = next(iter(state_log.iter_from(1)), None).get("seq")`）を list_rewind_points() 先頭に適用＋`s < oldest_seq` を除外＝**list と checkout が by construction で一致**（巻き戻せない点を宣伝しない）。新 floor 計算を発明していない。
- **falsify-first**: `test_list_rewind_points_excludes_truncated_seqs`（Tier 2、**real WAL log**＝`log.append`/`truncate_below(2)`、no mock）＝seq 1(gen boundary) を truncate→list から除外、RED-without-fix 確認済。truncate_below の watermark 維持 nuance も正しく扱う。
- checkout 経路は不変（正しく reject 継続）、edge case（empty WAL→no filter / all-truncated→[] / no-checkpoint→[]）対応。
- ruff ✓ / tier-audit ✓（7 tests）/ pytest green 待ち。
→ autonomous（owner 就寝中、(A) 承認済）、review-PASS + CI-green で merge。

PR #2241, opened 2026-06-27
[tui-coder] fix(router): cancel branch emitted the max_iterations error by copy-paste

tya5:
**[tui-coder]** — standalone correctness fix split out of #236 ② (inline esc-interrupt); lands independently per lead-coder's (A) decision.

## Bug

`RouterLoop.run_loop()`'s user-cancel branch (`_loop_cancelled`) emitted the **max_iterations error text** — a copy-paste of the exhaustion path directly below it:

```python
# Cancelled path (user-initiated): canned decision-enabling error.
if _loop_cancelled:
    await self.host.put_outbox(kind="error",
        text=f"Router loop exceeded max iterations ({self.max_iterations}). ...")
```

So a plain **user cancel** surfaced a misleading *"Router loop exceeded max iterations"* error. Every cancel consumer hits it — the **existing web ws `cancel_inflight`** path, and any turn-interrupt UI. (max_iterations exhaustion takes the limit-deny / on_limit path below, where `_loop_cancelled` is False — never this branch — so the message is cancel-only.)

## Fix

Emit a `status` **"✗ turn interrupted"** acknowledgement instead.

## Test

Regression test in `test_turn_cancel_1468.py` (real `RouterLoop` + `FakeRouterHost`, no mocks): arms a cooperative cancel, asserts the outbox carries the "turn interrupted" status and **neither** the max_iterations text **nor** an `error` kind.

- [x] `ruff` / `verify_module_docstrings` / `test_tier_audit --strict` — clean
- [x] `test_turn_cancel_1468.py` — 10 passed
- [x] **Falsify-checked**: reverting the one-line fix turns the new assertion red (line 133)
- [x] Live-observed via the inline esc-interrupt spike (held separately): a user cancel now shows "✗ turn interrupted", not the max-iterations error

## Context

Found while wiring inline esc-interrupt (#236 ②). Per lead-coder, the inline esc **UX** is held (cooperative cancel can't interrupt a mid-flight LLM call — a misleading affordance) and tracked separately for a hard-cancel re-design. This router fix stands alone and also corrects the **live web ws cancel** path, so it lands now.

## Tier self-check

Tier 2 (OS-invariant): real `RouterLoop` driven through its cancel boundary, public outbox-surface asserts, no mocks.

### DIFF
--- a/src/reyn/runtime/router_loop.py
+++ b/src/reyn/runtime/router_loop.py
@@ -2470,15 +2470,16 @@ class RouterLoop:
                 continue  # re-enter inner for loop with extended limit
          break  # no extension granted or no on_limit — exit outer while
 
-        # Cancelled path (user-initiated): canned decision-enabling error.
+        # Cancelled path (user-initiated esc / cancel_inflight): an acknowledgement,
+        # not an error. This branch previously emitted the max_iterations error
+        # text by copy-paste — a bug every cancel consumer hit (inline esc + web
+        # ws cancel). max_iterations exhaustion takes the limit-deny / on_limit
+        # path below (where _loop_cancelled is False), never this branch, so this
+        # message is cancel-only.
         if _loop_cancelled:
             await self.host.put_outbox(
-                kind="error",
-                text=(
-                    f"Router loop exceeded max iterations ({self.max_iterations}). "
-                    f"Configure safety.on_limit.mode=interactive or auto_extend to "
-                    f"extend, or increase safety.loop.max_router_iterations."
-                ),
+                kind="status",
+                text="✗ turn interrupted",
                 meta={"chain_id": self.chain_id},
             )
             return self._total_usage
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **#2241 (router_loop cancel-message bug fix) review: PASS、--auto。** (A) 決定の standalone correctness fix、clean。
- cancel 分岐が `kind="error"` + max_iterations error text（copy-paste bug、全 cancel consumer = inline esc + **既存 web ws cancel** が誤テキスト）→ `kind="status"` + `"✗ turn interrupted"` に修正。
- comment が分岐分離を明示：max_iterations exhaustion は `_loop_cancelled=False` の limit-deny/on_limit path を通る＝この分岐は cancel-only。よって max_iterations の decision-enabling deny は自分の path で維持、cancel は plain ack に。
- test: `test_cancel_surfaces_interrupted_ack_not_max_iterations_error`（Tier 2、**real RouterLoop+FakeRouterHost, mock 0**、interrupted-ack/status-kind/no-error/no-max-iteration を behavioral assert、revert→red @L133）。
- contained diff（router_loop.py + test）、4 gate green。**既存 web ws cancel path も直る**。
→ autonomous、review-PASS + CI-green で merge。

Issue #2072, opened 2026-06-23
[hooks follow-up] cross-session push — session field declared but routing not wired


tya5:
From the #1800 / #2069 hook arc. The push directive's `session` field (on both `template_push` and `shell_push`, via PushBlock → ResolvedPush) is **parsed + carried forward-compat**, but **cross-session routing is NOT wired** — the dispatcher's push uses `_put_inbox` (the CURRENT session) and ignores `resolved.session`. So a push always lands in the current session's inbox; `session` is a no-op today.

To wire it: the dispatcher routes `resolved.session` to ANOTHER session's inbox (a fleet/registry lookup → that session's `_put_inbox`). Doc-noted in `docs/concepts/runtime/hooks.md` (Deferred section).

Priority: low (forward-compat field; wire when cross-session push is actually needed).

tya5:
**[lead-coder]** Closed — resolved by PR #2245 (merged 2026-06-27, "fix(#2072): wire cross-session hook push — route to resolved.session"). The cross-session hook-push routing gap is implemented + falsify-verified + CI-green. Verified during the #2259-arc cleanup sweep.

PR #2245, opened 2026-06-27
fix(#2072): wire cross-session hook push (route to resolved.session)

tya5:
**[e2e-coder]** — #2072: wire the cross-session hook push (the dispatcher ignored `resolved.session`). Silent-gap correctness — declared but inert, the same class as the cost-warn #2230 gap. Falsify-first; gap was clear (no plan-first needed, per lead).

## The gap (primary evidence)

A push's `session` field (`template_push` / `shell_push` → `PushBlock` → `ResolvedPush`) is parsed and forward-compat carried, but `HookDispatcher._push_resolved` always pushed to the CURRENT session (`self._put_inbox` / `self._stage_next_turn_context`) and **ignored `resolved.session`** — documented inline as a no-op ("the dispatcher pushes to the current session and ignores `resolved.session`, so a `session` value is a no-op today"). So a cross-session push always landed in the current session.

## The fix

`HookDispatcher._push_resolved` now routes a push whose `session` names a **different** live session to THAT session's inbox, via an injected `cross_session_put` seam; `null` / empty / the current session stays local (unchanged). The session provides `_cross_session_hook_put` — the **canonical wake-triple** (`resolve_session` / `get_session` → `_put_inbox` → `ensure_session_running`), the same pattern `TaskWaker` / `webhook_routing` use; a `transport:native` target resolves via `resolve_session`, a bare sid via `get_session`. The `wake` flag rides in the payload (target processes it the same way the current session would; only `wake` boots the target's run-loop). Best-effort: a target naming no live session is logged + dropped (a cross-session push to an absent peer must never crash the source run). The `cross_session_put` seam **defaults to None**, so existing `HookDispatcher` callers are unaffected (no-op-equivalent).

## Test plan

- [x] **`tests/test_2072_cross_session_push.py`** (NEW, Tier 2, 3 tests, real recording seams — no mocks, mirroring `test_hook_dispatcher_1800_5b`): a cross-session push routes to the cross-session seam with the target sid + wake flag and does NOT land in the current inbox (**falsify-verified RED** when the routing branch is stripped — the push falls into the current session, the pre-#2072 no-op); a null-`session` push stays local; a self-named push stays local (no needless hop).
- [x] Updated the stale `_parse_shell_push` docstring ("not wired" → "ROUTED since #2072").
- [x] All 4 CI gates: `ruff` ✓; `test_tier_audit.py --strict` ✓ (a first pass had `len(...) == N` count-pins → switched to the tuple-unpack exactly-one idiom); `verify_module_docstrings.py` ✓; **full rootdir `pytest`** ✓ — 7484 passed, **zero new failures** (only the 4 pre-existing env quirks: gateway entry-points ×3, `reyn.safe` relocate).

### DIFF
--- a/src/reyn/hooks/dispatcher.py
+++ b/src/reyn/hooks/dispatcher.py
@@ -62,10 +62,19 @@ class HookDispatcher:
         consent_bus: Any = None,
         consent_gate: "Callable[[], bool] | None" = None,
         emit_event: "Callable[..., Any] | None" = None,
+        cross_session_put: "Callable[..., Any] | None" = None,
+        current_session_id: "str | None" = None,
     ) -> None:
         self._registry = registry
         self._put_inbox = put_inbox
         self._stage_next_turn_context = stage_next_turn_context
+        # #2072: cross-session push routing. ``cross_session_put(target_sid, kind, payload,
+        # wake=...)`` delivers a push to ANOTHER session's inbox (the canonical wake-triple);
+        # ``current_session_id`` identifies THIS session so a push naming it (or naming none)
+        # stays local. None ``cross_session_put`` (e.g. unit tests / no registry) → the push
+        # always stays local — the pre-#2072 behaviour, no-op-equivalent.
+        self._cross_session_put = cross_session_put
+        self._current_session_id = current_session_id
         self._run_shell = run_shell
         self._sandbox_config = sandbox_config
         self._sandbox_backend = sandbox_backend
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — **#2245 (#2072 cross-session push 配線) review: PASS、--auto。** silent-gap（宣言されてるのに動かない、cost-warn #2230 と同クラス）を wired、再発明なし。
- **routing**: `_push_resolved` が `resolved.session`（別 live session）を **canonical wake-triple（resolve_session→_put_inbox→ensure_running、TaskWaker/webhook と同 pattern）で target inbox に route**。null/empty/self/no-capability → local path（pre-#2072 不変、no regression）。
- **best-effort**: absent peer は log+drop、source run を crash させない。**seam default None** で既存 HookDispatcher caller 無影響。
- **falsify 3本 real-seam（mock 0）**: cross-session-reaches-target（current でなく）RED-when-stripped / null・self は local。count-pin→tuple-unpack exactly-one idiom 自己修正（test-audit 規律 ◎）。
- 4 gate green、full pytest 7484 passed、stale doc 更新。
→ autonomous、review-PASS + CI-green で merge。

Issue #1765, opened 2026-06-18
WAL-fsync event-loop blocking — deliberate future improvement: constraints, lessons, candidate approaches


tya5:
**[lead-coder]** — recorded foundation for a *deliberate, unhurried* future effort (owner-directed 2026-06-18). **Not urgent, not for immediate dispatch.** The immediate input-echo latency is solved separately at the UI layer (revert of #1753's to_thread + an `app.py` optimistic-render yield, keeping the WAL fsync synchronous). This issue captures the *underlying* concern + everything we learned this round, so the eventual improvement is done safely and doesn't repeat this round's mistakes.

## The underlying concern (deferred, real)
The synchronous WAL-append `os.fsync` blocks the event loop ~100–300ms (local SSD) per append. For a single-user TUI this is mostly invisible (the optimistic render shows the echo first). But under **concurrent / multi-session** load it will matter — every session's turn freezes the shared loop during its WAL fsync.

## The core trade-off (the owner's key insight)
- **Sync fsync** ⇒ each append is **event-loop-atomic** (no `await` inside the held lock ⇒ nothing interleaves *during* an append ⇒ naturally serial, zero concurrency surface to reason about). Simple + safe, **but** blocks the loop.
- **Any unblocking** (off-loop fsync, etc.) introduces an **intra-append yield** ⇒ a concurrency surface that MUST be reasoned about explicitly. The speed win and the complexity come together.

## Invariants that MUST be preserved (verified this round, ref-exact origin/main)
1. **WAL durability**: fsync-per-append, durable *before* `append()` returns (crash-recovery contract). `state_log.py`.
2. **Append serialization**: the per-StateLog `asyncio.Lock` serializes all appends across all agents/sessions (single global WAL); `seq` strictly monotonic. Multi-agent/session = coroutines in one loop, not threads/processes.
3. **Snapshot ⟶ WAL-truncation ordering**: the snapshot at S must be **durable before** the WAL is truncated below S, else a crash loses state S′..S. Currently held because the snapshot save is **synchronous** (`AgentSnapshot.save`, agent_snapshot.py:148 — sync `os.fsync` + atomic tmp→rename). **Async-ifying the snapshot save breaks this** ⇒ needs a `_durable_applied_seq` watermark that truncation reads instead of the in-memory `applied_seq`.
4. **Workspace durability** (see #1763): the workspace is currently fsync'd at **no** layer and is **not** ordered with the WAL — any durability rework must consider the whole picture, not just the WAL.

[...]

PR #2246, opened 2026-06-27
fix(#1765 1a-i): off-loop WAL fsync via a shared DurabilityWorker

tya5:
**[e2e-coder]** — #1765 Step 1a (split **i: WAL**): off-loop WAL fsync via a shared `DurabilityWorker`. Design owner-steered + lead-approved (converged: unified durability worker + write-ahead ordering, staged; this is the WAL foundation, behavior-invariant + recovery-safe + loop-free). Split confirmed (1a-ii = snapshot routing follows).

## The change

A blocking `os.fsync` does not yield, so it freezes the WHOLE event loop for its duration (hundreds of ms on slow storage) — starving TUI repaint + every concurrent session. The new `DurabilityWorker` (`core/events/durability_worker.py`) is a **substrate-agnostic** (P7-clean) single FIFO serialisation point; `StateLog.append` submits its write+fsync to it, and the fsync runs OFF the loop (`await asyncio.to_thread`) — so the loop stays free DURING it, while `append` still returns only once durable (**per-append crash-recovery contract UNCHANGED** — no relaxed-durability window; that is the deferred Step 2).

**#1751 closure (structural).** The off-loop fsync would reopen the #1751 surface (a lockless `iter_from` reading a written-but-not-yet-fsync'd entry). Closed with an `_inflight_seq` marker: the worker records the seq it is fsyncing right now, and `iter_from` skips exactly that one entry. Crucially this is a **single-entry exclusion, NOT a per-instance durable ceiling** — a first watermark-ceiling attempt regressed `test_session_auto_resume` because the WAL is **process-shared** across multiple StateLog instances (one seeds, another reads), and a stale per-instance ceiling hid durable entries written by another instance. `_inflight_seq` excludes only *this* log's genuine in-flight entry, so cross-instance reads + recovery see every durable entry.

**Worker = a lock, not a background task (Step 1a).** Same serial-FIFO + off-loop + blocking contract, but with **no background task to leak across event loops** (an earlier task+queue version tripped "Task destroyed but pending" in per-test loops). The `submit(do_durable_write)` seam is the stable contract: Step 2 (non-blocking writes) evolves only the worker **internals** (a queue + drainer for group-commit) behind it. `StateLog.submit_durable()` is the seam the snapshot substrate (split ii) uses to share this one worker.

## Macro-audit axes (your 8)

1. **P7** ✓ — `DurabilityWorker` has no WAL/snapshot/kind literal; the write callable carries it (the `_inflight_seq` bump is in the WAL callable). 2. **FIFO** ✓ — fair `asyncio.Lock`, falsified (`test_serial_fifo_order`). 3. **#1751** ✓ — `_inflight_seq` skip, **falsify RED-when-stripped** (the concurrent in-flight read test). 4. **truncate** ✓ — `truncate_below` holds `self._lock`, serialised with the append's worker-write (append holds the lock across submit). 5. **completeness** ✓ — append awaits durability; falsified by the recovery test (replays every entry post-restart). 6. **falsify** — 7 real-instance tests (no mocks). 7. **refactor-safety** ✓ — behavior-invariant (full suite 7491 passed; the 1 c_staging timing test was `sleep(0)`-fragile vs the now-async fsync → adapted to poll-for-durable, the same discipline #1751 used when fsync first moved off-loop). 8. **aclose** ✓ + module-docstring design narrative.

## Test plan

- [x] `tests/test_1765_durability_worker.py` (4): serial-FIFO, blocking-await, loop-free (a slow off-loop task doesn't starve a ticker), task-failure-surfaces.
- [x] `tests/test_1765_state_log_durability.py` (3): completeness+recovery (crash→restart→replay all), **#1751 concurrent in-flight read excluded — RED-when-stripped**, seq-order=file-order under concurrency.
- [x] Adapted `test_1800_c_staging.py::test_c_staging_durable_during_drain_wait` (`sleep(0)` → poll-for-durable; intent preserved).
- [x] 4 CI gates: ruff ✓ / tier-audit ✓ (private-state-assert self-catch fixed) / verify_module_docstrings ✓ / **full rootdir pytest** ✓ — 7491 passed, **zero new failures** (only the 4 pre-existing env quirks).

### DIFF
--- a/src/reyn/core/events/durability_worker.py
+++ b/src/reyn/core/events/durability_worker.py
@@ -1,0 +1,52 @@
+"""DurabilityWorker — a single serial worker that runs durable writes off the event loop.
+
+#1765 Step 1a. A blocking ``os.fsync`` does not yield, so it freezes the WHOLE event loop
+(TUI repaint + every concurrent session) for its duration — on slow storage that is hundreds
+of ms per fsync. This worker moves the fsync OFF the loop: a substrate submits a durable-write
+task (a coroutine that writes + ``await asyncio.to_thread(os.fsync, …)``), the worker runs it
+serially, and ``submit`` awaits its completion — so the caller's durability contract is
+UNCHANGED (it returns only once durable) while the loop stays free DURING the fsync.
+
+Substrate-agnostic by construction (P7): the worker holds no WAL / snapshot / workspace
+knowledge — the injected write callable carries all of it (including any post-write bookkeeping
+such as a durable-seq watermark). Its one structural guarantee is **serial FIFO**: tasks run
+one at a time in submit order, so *enqueue order = durability order*. That single ordering point
+is what later steps build on — the cross-substrate write-ahead ordering (a depended-upon
+substrate submitted before the WAL event that references it) and, in #1765 Step 2, non-blocking
+writes (``submit`` returns before the task runs, with a barrier awaiting only where an external
+effect is gated). For Step 1a ``submit`` always awaits (no relaxed-durability window).
+"""
+from __future__ import annotations
+
+import asyncio
+from typing import Awaitable, Callable
+
+DurableWrite = Callable[[], Awaitable[None]]
+
+
+class DurabilityWorker:
+    """A single serialisation point for off-loop durable writes (see module docstring).
+
+    Step 1a uses a fair ``asyncio.Lock`` as the serial point: ``submit`` acquires it (FIFO —
+    asyncio locks grant in acquisition order, so submit order = durability order), runs the
+    task (its ``await to_thread(os.fsync)`` keeps the loop free), and releases. No background
+    task → nothing to leak between event loops. Step 2 (non-blocking writes) evolves the
+    INTERNALS to a queue + drainer behind this SAME ``submit`` contract — so the substrate
+    routing built on it is not throwaway."""
+
+    def __init__(self) -> None:
+        self._lock = asyncio.Lock()
+
+    async def submit(self, do_durable_write: DurableWrite) -> None:
+        """Run a durable-write task, serialised with every other submit (FIFO = durability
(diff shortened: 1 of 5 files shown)


tya5:
**[lead-coder]** — **#2246 (#1765 Step 1a-i: DurabilityWorker + WAL) core-WAL macro-audit: PASS、merge GO。** 8 軸 + 提起した load-bearing concern を primary evidence で resolve。
- **DurabilityWorker**: P7-clean（docstring「holds no WAL/snapshot/workspace state」、write callable 注入）、lock-based（fair asyncio.Lock = serial-FIFO、background task 無し = axis-8 leak 回避）、`submit` seam が Step 2 の queue+drainer evolution を支える（非 throwaway）。
- **`_inflight_seq`（single-entry exclusion, ceiling でない）**: same-instance の off-loop window read-surface（#1751）を構造的に閉じる。lock でなく watermark/inflight が reader を守る。
- **concurrent cross-instance concern を RESOLVE（私の hold 理由、primary evidence で解消）**: same-process=registry の単一 StateLog instance（全 iter_from caller が同 instance 経由＋recovery/rewind/resume=quiescent）/ cross-process=single-writer（concurrent writer は seq-collision で pre-existing-broken、reader=recovery は dead writer の static WAL）。→ A active fsync 中の B concurrent-read は**実在しない**＝per-instance `_inflight_seq` で residual surface 無し。
- append=worker.submit（seq は lock 下採番）、truncate も lock 下直列、watermark disk-init（recovery 全 replay）、falsify 7本 real（#1751 RED-when-stripped）、**behavior-invariant 7491 passed + recovery smoke OK**。
- 2 regression の honest catch（_inflight_seq 再設計 / poll-for-durable adapt）= full-suite 規律。
→ behavior-invariant refactor ゆえ autonomous merge（Step 2 非ブロックの window 導入は owner review 想定）。

PR #2247, opened 2026-06-27
fix(#1765 1a-ii): route snapshot save off-loop via the shared DurabilityWorker

tya5:
**[e2e-coder]** — #1765 Step 1a-ii. Follows 1a-i (#2246, WAL fsync off-loop). Autonomous GO from lead.

## What
The per-mutation snapshot `save()` did a synchronous `os.fsync` on the event loop, freezing the whole loop (TUI repaint + every concurrent session) for its duration. This routes it through the **same serial DurabilityWorker** as the WAL, via the existing `state_log.submit_durable` seam.

- `AgentSnapshot.save` split into `serialize()` (sync — captures a consistent view of the mutable inbox/chains state) + `write_durable()` (pure I/O, safe off-loop). Sync `save()` contract unchanged.
- `SnapshotJournal.save` is now async: serialises synchronously, submits only the write+fsync to the worker. 12 per-mutation callers `await self.save()`.
- `install()` (recovery restore) stays synchronous — a one-shot recovery write, not the hot path; avoids forcing an async restore path.

## WAL → snapshot ordering (the load-bearing invariant)
Each mutation awaits its WAL append (durable) **before** awaiting the snapshot save, and the worker is serial FIFO → the snapshot's `applied_seq` becomes durable only **after** the WAL seq it records (`applied_seq ≤ durable WAL seq`). A crash can never leave a durable snapshot pointing at a non-durable WAL entry.

## Scope
The high-frequency per-mutation `save()`. The checkpoint-boundary generation save (`SnapshotGenerationStore.record`, fired per turn/plan-step) **remains a sync fsync** — a noted lower-frequency follow-up candidate, called out so the residual loop-block isn't silent.

## Falsify (RED-when-stripped, both verified)
- **loop-free** — a slow snapshot write doesn't block a concurrent ticker. Verified RED with the pre-1a-ii inline sync save (`ticks == 0`).
- **WAL→snapshot ordering** — at snapshot-write time the recorded `applied_seq` is already on disk in the WAL. Verified RED when a mutation saves before its WAL append (`assert 1 in []`).
- **persist + recover** — a fresh load after a simulated restart sees the mutated state at the latest WAL seq.

## Test plan
- [x] `pytest tests/test_1765_snapshot_journal_off_loop.py` (3 new) + 1765 WAL/worker + journal/snapshot/recovery suites — 43 passed
- [x] Timing-sensitive recovery suites (`test_1800_c_staging`, `test_session_auto_resume`) — 11 passed
- [x] Both falsify strips verified RED, then restored
- [x] ruff (full tree `src tests scripts`) — clean
- [x] `test_tier_audit.py --strict` on the new test — OK 3/0/0
- [x] `verify_module_docstrings.py` on both changed src files — clean (no module-docstring touched)
- [x] Full suite — 7494 passed; 4 pre-existing failures (gateway entry-point + `reyn.safe` import-relocation, import-order sensitive) reproduce on the clean tree with this change stashed → **not introduced here**

### DIFF
--- a/src/reyn/core/events/agent_snapshot.py
+++ b/src/reyn/core/events/agent_snapshot.py
@@ -135,10 +135,12 @@ class AgentSnapshot:
             ),
         )
 
-    def save(self, path: Path) -> None:
-        path = Path(path)
-        path.parent.mkdir(parents=True, exist_ok=True)
-        tmp = path.with_suffix(path.suffix + ".tmp")
+    def serialize(self) -> str:
+        """Serialise to a JSON string — SYNCHRONOUS, so it captures a consistent view of the
+        mutable state (inbox / chains / …) at the call instant. #1765 1a-ii splits this from
+        the durable write so an off-loop save snapshots the state here (sync) and only the
+        write+fsync runs off the event loop, with no risk of the state being mutated mid-write.
+        """
         payload = {
             "version": SNAPSHOT_VERSION,
             "session_id": self.session_id,  # FP-0043 S5 (additive; legacy load → "main")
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — **#2247 (#1765 Step 1a-ii: snapshot routing) macro-audit: PASS、--auto。** WAL+snapshot unification 完成、write-ahead correct。
- **WAL→snapshot ordering（load-bearing correctness）**: per-mutation save が同 DurabilityWorker に off-loop（`state_log.submit_durable`）、mutation が **WAL append durable を await してから save** → **applied_seq ≤ durable WAL seq**（crash で durable snapshot が未 durable WAL を指さない）。write-ahead の核。
- **AgentSnapshot.save split**: `serialize()` sync（consistent state を await 前に capture＝mid-write mutation risk なし）+ `write_durable()` off-loop（atomic tmp→fsync→rename）。content 一貫 + durability off-loop の正しい分離。
- **falsify**: mock 0（real）、ordering RED-when-stripped（applied_seq が WAL に既 durable を slow-write で検証）+ loop-free（slow write が ticker starve せず）。
- behavior-invariant（full suite 7494 passed、4 = pre-existing import-order、clean-tree で再現確認済＝not e2e's）。
- **honest scope note**: checkpoint-boundary generation save（SnapshotGenerationStore.record, per turn/plan-step）は **lower-freq ゆえ sync のまま**、#1765 内の follow-up として PR で明示（silent expand せず＝◎）。ordering は壊さない（generation-save も durable）。
→ behavior-invariant + recovery-safe foundation、autonomous merge（Step 2 非ブロック window は owner review）。

PR #2268, opened 2026-06-28
[tui-coder] feat(inline-cui): ❯ prompt/echo, distinct tool glyph, state-driven colour, message spacing

tya5:
**[tui-coder]** — owner-directed visual refinements toward Claude Code (rapid live feedback this session). Updated after revisions d7de4f3e + the comment-alignment commit.

## What

- **`❯` chevron** — the user echo marker AND the input prompt use `❯` (the wider chevron CC uses for both "You" and the prompt), replacing `>`.
- **Tool invocation glyph** — `tool_call_started` leads with `▸` (an invocation), distinct from the `⏺` assistant reply.
- **State-driven colour** — default is the **terminal default fg** (`_CC_TEXT="default"` — owner: "白そのものよりデフォルト色"); colour is reserved to **signal state**: amber = needs-you (intervention), red = error, green = done, dim = ambient/low. A coloured glyph always means "something to notice".
- **One blank line between message blocks** — `wants_separator(kind, seen)`; never before the first message, a nested `⎿` detail row, or a **transient** status/trace.
- **Removed the redundant `thinking…` status** (`session.py`) — owner: "thinking… と Working… 二重表示、意味がわからない". It double-displayed with the event-driven `Working…` spinner AND, as a cleared transient, orphaned a blank line before every reply. The working indicator is now the sole in-progress signal.

## Test plan

- [x] `ruff` / `verify_module_docstrings` / `test_tier_audit --strict` — clean
- [x] Renderer/inline/session/repl regression: 399 passed, 2 skipped. Updated glyph assertions (`>`→`❯`, tool `⏺`→`▸`), distinct-markers guard, `wants_separator` (skips first/nested/**transient**)
- [x] **Live tmux**: `❯` prompt + echo; `▸` tool rows; **exactly one** blank line between the user line and the reply; **no `thinking…` line** (spinner is the sole in-progress cue)

## Notes

- **`session.py` touch** (the `thinking…` removal): a clean removal of one UI status emission — no OS logic, no P6 event change (`user_message_received` still fires), no P7 concern. lead-coder judged it fine in this PR (no OS-coder review needed).
- **`--cui` follow-up**: `ConsoleChatRenderer` (no working indicator) loses its only in-flight cue with `thinking…` gone — the plain fallback path. Tracked in **#2269** (roi:low) to give `--cui` its own indicator if wanted.
- The 2-blank-before-reply originally flagged here is **root-fixed** in d7de4f3e (not a disclosed-accept).

## Stacking

Builds on the merged renderer (#2265/#2267) — same `_KIND_LINE`.

### DIFF
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -252,7 +252,7 @@ async def run_inline_input(registry, renderer) -> None:
     top_rule = Window(height=1, char="─", style=f"fg:{_CC_DIM}")
     bottom_rule = Window(height=1, char="─", style=f"fg:{_CC_DIM}")
     prompt_sym = Window(
-        FormattedTextControl([(f"fg:{_CC_ACCENT} bold", "> ")]), width=2, height=1
+        FormattedTextControl([(f"fg:{_CC_ACCENT} bold", "❯ ")]), width=2, height=1
     )
     input_win = Window(BufferControl(buffer=buf), height=1)
     inputrow = VSplit([prompt_sym, input_win])
--- a/src/reyn/interfaces/repl/renderer.py
+++ b/src/reyn/interfaces/repl/renderer.py
@@ -245,7 +245,7 @@ class RichChatRenderer(ChatRenderer):
 # Claude Code-style palette. Default is plain text (_CC_TEXT); colour is reserved
 # to signal STATE — error (red), needs-action (amber), done (green), ambient/low
 # (dim) — so a coloured glyph always means "something to notice".
-_CC_TEXT = "#e6edf3"    # default near-white — normal text + markers
+_CC_TEXT = "default"    # terminal default fg — normal text + markers (no forced colour)
 _CC_DIM = "#6b7280"     # low-importance / ambient
 _CC_DONE = "#7ee787"    # green — completion
 _CC_ERR = "#f97066"     # red — failure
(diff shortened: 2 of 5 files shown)


tya5:
**[tui-coder]** — Revision pushed (d7de4f3e). **The held 2-blank known-issue is now ROOT-FIXED — the owner ruled on it live (fix it to 1 line), and gave two more inline-display refinements; all addressed:**

1. **2-blank before replies → 1 blank.** Root cause found: the per-turn `thinking…` status (a transient) got a separator, was cleared in place, and left its blank orphaned. Fixed by **removing the redundant `thinking…` status** (it also double-displayed with the `Working…` spinner — owner: "thinking… と Working… 二重表示、意味がわからない") + `wants_separator` now skips transient kinds defensively. Live-verified: exactly one blank between the user line and the reply, no `thinking…` line.
2. **Marker/text colour → terminal default** (`_CC_TEXT="default"`, not forced near-white — owner: "白そのものよりデフォルト色").

⚠️ **OS-lane flag for your review:** the `thinking…` removal is in `session.py`. Impact: `--cui` (`ConsoleChatRenderer`, no working indicator) loses its only in-flight cue — the plain fallback path; flagged for a follow-up (give `--cui` its own indicator) if wanted. No test depended on `thinking…`; 399 passed.

So the Test plan's `- [ ] (known issue)` item is resolved (not a disclosed-accept — actually fixed). Re-review the delta when you can; CI re-running.

Issue #2271, opened 2026-06-28
Inline CUI: extensible status-bar framework + redesign (5-chip bar · overflow submenu · expandable chips: model/cost/agent-tree/task-tree/mcp-skill-cron-hook)


tya5:
**[tui-coder]** — owner-directed redesign of the inline CUI **status bar** (the line below the input, currently `model standard · agents 1 · skills 0 · cost $… · ctx …`). Goal: make it cleaner/cooler AND **framework-ize the chips + expansions so new ones can be added declaratively** (echoing the region framework F1–F5). Owner approved phased implementation.

## Naming (owner decision)
- The line = **status bar**. Items = **chips**. An opened chip = its **expansion** (picker / tree / toggle-list / detail). The `…` chip's secondary row = the **overflow submenu**. (A follow-up will standardize code/docs terminology to "status bar".)

## Layout — main row
- **5 chips, `│`-separated, label-value pairs, NO icons:**
  ```
   model standard │ cost $0.0123 │ agent default │ task 0 │ …
  ```
- `model` = class · `cost` = headline total · `agent` = **agent name** · `task` = active count · `…` = overflow.
- Interaction (extends the existing chip+dropdown): `↓` focus → `←→` move → `enter` expand → `↑↓`+`enter` inside an expansion → `esc/↑` back.

## Per-chip expansion spec
| chip | bar value | expansion | kind | accessor / dispatch |
|---|---|---|---|---|
| **model** | class name | model picker | **picker** (actionable) | exists: `known_model_classes()` + `/model` |
| **cost** | total $ | total / per-agent / per-session + tokens | **detail** | `total_cost_usd` per agent; need a per-session aggregate accessor |
| **agent** | agent name | **agent/session tree + attach switch** | **tree (actionable)** | `registry._sessions` (name→{sid→Session}) + `_attached`; dispatch `/attach <name>` / session-switch (`__session_switch_request__`). Need public `registry.session_tree()` |
| **task** | active count | **task tree** (status per node) | **tree (display)** | task backend has parent→child decomposition (`awaited`/`background`, #2187) + `depends_on` DAG + status. Need public tree-snapshot accessor |
| **…** | — | overflow submenu row: `mcp │ skill │ cron │ hook` | submenu → each is a **toggle-list** | see below |

### Overflow submenu — `mcp / skill / cron / hook` (list config-registered + enable/disable)
Each opens a list of config-registered entries with an enable/disable toggle (enter = toggle).
[...]

PR #2274, opened 2026-06-28
[tui-coder] feat(inline-cui): agent chip = agent/session tree + attach switch (#2271 Phase 2)

tya5:
**[tui-coder]** — Phase 2 of #2271, **stacked on Phase 1 (#2272)** — base is the Phase-1 branch so this diff is Phase 2 only (retarget to main once #2272 merges).

## What

The **agent** chip's expansion is now an **agent/session tree with attach/switch on select**.

- **`AgentRegistry.session_tree()`** — public, **snapshot-style read-only** accessor (freshly-built copy of plain dicts, no handle to live `_sessions`/`_attached` — per the design note): agents in load order, each with sorted sessions + the attached `(agent, sid)` marked.
- **`_agent_expansion`** builds indented tree rows (agents + their sessions, attached marked `▸`) and **reuses `CommandUIElement`** — *no new element type* — so selecting a row dispatches `/attach <agent>` (or `/session switch <sid>` when that agent is already attached) through the normal slash path. Empty → `DetailElement`.

## Test plan

- [x] `ruff` / `verify_module_docstrings` / `test_tier_audit --strict` — clean (re-verified by me)
- [x] Regression: 134 passed, 2 skipped. New `test_registry_session_tree` (real `AgentRegistry`: shape, attached-marking, load order, sid sort, **snapshot isolation** = mutating the result doesn't change the registry, no mocks); agent-expansion tests moved to the tree form (CommandUIElement, `▸` at agent+session level, the 3 command variants, empty→DetailElement). No size/format pins.
- [x] **Live tmux**: `/agent new researcher` → status bar `agent researcher`; opened the agent chip → indented tree rendered (`default`/`researcher` + sessions, attached `▸`); selecting the `default` row dispatched `/attach default` and the bar switched to it.

## Note
`session_tree` touches `registry.py` (a small read-only accessor, snapshot-style per lead-coder's note). Mechanical impl by a Sonnet sub-agent against a precise spec; I re-ran gates + read `session_tree`/`_agent_expansion` + live-verified the switch myself.

### DIFF
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -117,14 +117,26 @@ def _cost_expansion(snap, dispatch):
 
 
 def _agent_expansion(snap, dispatch):
-    # Phase 1: read-only agent list. Phase 2 = agent/session tree + attach switch.
-    def lines():
-        names = snap["agent_names"]
-        if not names:
-            return ["(none)"]
-        attached = snap["attached_name"]
-        return [f"▸ {n}" if n == attached else f"  {n}" for n in names]
-    return DetailElement(lines)
+    # Phase 2: agent/session tree; selecting a row attaches / switches.
+    tree = snap.get("session_tree") or []
+    rows: list[str] = []
+    cmds: list[str] = []
+    for agent in tree:
+        amark = "▸" if agent["attached"] else " "
+        rows.append(f"{amark} {agent['agent']}")
+        cmds.append(f"/attach {agent['agent']}")
+        for sess in agent["sessions"]:
+            smark = "▸" if sess["attached"] else " "
+            rows.append(f"    {smark} {sess['sid']}")
+            # switch the session when its agent is already attached; otherwise
+            # attach the agent first (the user can then switch session).
+            if agent["attached"]:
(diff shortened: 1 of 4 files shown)


tya5:
**[tui-coder]** — addressed the Tier-4 blocker. Reworked `tests/test_registry_session_tree.py` to build a real `AgentRegistry` with a real `Session` factory and load state through the **public sync path** (`get_or_load` / `spawn_session`, same pattern as `test_registry_multi_session_1726`) — zero private-attr mutation.

Covers: entry/session shape, multi-agent + multi-session listing, sid sorting, the unattached all-false marking, and **snapshot isolation** (the read-only-copy property you flagged as the load-bearing one). Attached=True *rendering* (▸) stays covered by `test_agent_expansion_attached_agent_marked_with_arrow` against a `session_tree` snapshot — so I dropped the private-`_attached` cases rather than take the heavy async `attach()` path (boots `session.run()`+forwarder tasks → teardown-fragile in a unit test).

`tier_audit --strict`: 8 OK, 0 errors. `ruff`: clean. 8 passed. Pushed `d8c28d78`.

Issue #2302, opened 2026-07-01
orphaned TUI-only outbox sentinels leak into CLI — complete 6-kind set (2 need owner disposition)


tya5:
**[lead-coder]** — A definitive origin/main sweep (all `OutboxMessage(kind=...)` emit sites × all consumers: inline `_output_loop`, ws `_FORWARDED_KINDS`, forwarder control-signals, chainlit) found the COMPLETE set of orphaned outbox kinds — sentinels that were consumed by the now-deleted Textual TUI and today have ZERO live consumer, so they leak into the CLI scrollback (blank line or bare unstyled text). This supersedes #2298's "exhaustive" enumeration, which missed everything below except its own four (and those turned out to be EventLog events, not outbox kinds — already correctly handled).

## Complete orphaned set (6 kinds)

| Kind | Emit site | Leak | Disposition |
|------|-----------|------|-------------|
| `intervention_resolved` | `runtime/services/intervention_handler.py:271` | blank line | **remove emit** (pure-dead; was TUI widget-removal) — **#2300** |
| `__quit__` | `interfaces/slash/quit.py:36` | blank line | **remove emit** (dead: `/quit` is intercepted in the input loops before the handler; ws already drops it) |
| `__donut__` | `interfaces/slash/donut.py:16` | blank line | **inline `reply()`** (TUI easter egg → print inline) |
| `__matrix__` | `interfaces/slash/matrix.py:17` | bare text ("There is no spoon.") | **inline `reply()`** (docstring already says "CUI mode falls back to a friendly inline message" — make it explicit) |
| `__cost_inline_toggle__` | `interfaces/slash/cost_inline.py:35` | bare text | **⚠ OWNER-DECISION** (see below) |
| `__docs_filter__` | `interfaces/slash/docs_filter.py:24` | blank / bare text | **⚠ OWNER-DECISION** (see below) |

## Plan

- **4 clear-disposition kinds** (intervention_resolved, `__quit__`, `__donut__`, `__matrix__`) → folded into **#2300** (2 emit-removals + 2 inline-reply conversions, each falsify-flipped). `intervention_resolved` is the originally-reported blank-line-per-answered-intervention bug.
[...]

PR #2321, opened 2026-07-01
test(tier2): /cost-inline + /docs-filter slash — _normalise helper + sentinel emission

tya5:
**[tui-coder]** — coverage mining, part of #2302 idle wave

Note: these two commands are the subject of #2302 owner decision (disposition of `__cost_inline_toggle__` and `__docs_filter__` orphaned sentinels). This PR pins the current contract; it would naturally be reverted/deleted alongside the handlers if #2302 resolves to removal.

## Summary

13 Tier 2 tests, all previously zero-coverage:

**`_normalise`** (6 tests): `'on'`/`'off'` exact match; empty → `''` (toggle); `'toggle'` keyword → `''`; unrecognised args fallback → `''`; case-insensitive `'ON'`/`'OFF'`.

**`cost_inline_cmd`** (4 tests): emits `__cost_inline_toggle__` sentinel; no-args → sentinel text `''`; `'on'` / `'off'` arg propagate to sentinel text via `_normalise`.

**`docs_filter_cmd`** (3 tests): emits `__docs_filter__` sentinel; empty args → `''` (clear filter); non-empty args stripped and passed through as sentinel text.

## Test plan

- [x] `ruff check tests/test_slash_cost_inline_docs_filter_cmds.py` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/test_slash_cost_inline_docs_filter_cmds.py` — 13/13 OK, 0 violations
- [x] `pytest tests/test_slash_cost_inline_docs_filter_cmds.py` — 13 passed
- [x] (skipped — no src changes) `python scripts/verify_module_docstrings.py`

Tier self-check: no mocks, no private-state reads, no `len()==N` format-pins.

### DIFF
--- a/tests/test_slash_cost_inline_docs_filter_cmds.py
+++ b/tests/test_slash_cost_inline_docs_filter_cmds.py
@@ -1,0 +1,126 @@
+"""Tier 2: /cost-inline + /docs-filter slash — _normalise helper + sentinel paths.
+
+Both commands emit a sentinel OutboxMessage that the TUI intercepts; the sentinel
+kind and text are the observable contract.  /cost-inline also has a pure _normalise
+helper that maps raw args to ('on'|'off'|'').
+"""
+from __future__ import annotations
+
+import pytest
+
+from reyn.interfaces.slash.cost_inline import _normalise, cost_inline_cmd
+from reyn.interfaces.slash.docs_filter import docs_filter_cmd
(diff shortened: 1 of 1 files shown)


tya5:
[tui-coder] Closing — owner decision (#2302) is to delete /cost-inline and /docs-filter entirely. The tests in this PR are superseded by falsify tests in the deletion PR.

PR #2329, opened 2026-07-01
test(#1830): Tier 2 coverage — slash/image + slash/pending pure helpers

tya5:
**[tui-coder]** — #1830 Tier 2 coverage mining.

## Summary

28 Tier 2 tests across 4 stateless pure helpers:

**`slash/image.py`** (16 tests)
- `_mime_for_path(path)` — extension → MIME type dict lookup; all 6 extensions, case-insensitive, unknown → None, no-extension → None
- `_file_size_human(n)` — byte count formatter; thresholds at 1 000/1 000 000, 1-decimal precision, zero case

**`slash/pending.py`** (12 tests)
- `_render_list(ops)` — stalled-op renderer; empty case, singular/plural count, kind+id row content, summary arrow line present/absent, dict-shaped op support
- `_render_needs_attention(summary)` — stuck-skill tail; empty dict → `""`, empty list → `""`, 1/2 stuck skills with all fields, missing fields default to `"?"`

None of these had any prior test coverage.

Part of #1830.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_slash_image_pending_pure_helpers.py` — 28 OK, 0 errors
- [x] `pytest tests/test_slash_image_pending_pure_helpers.py` — 28/28 passed

Tier self-check: Tier 2 — real function calls with concrete inputs, `SimpleNamespace` stubs for op objects, assert on return values only, no private state reads, no format pins.

### DIFF
--- a/tests/test_slash_image_pending_pure_helpers.py
+++ b/tests/test_slash_image_pending_pure_helpers.py
@@ -1,0 +1,223 @@
+"""Tier 2: pure helpers in slash/image.py and slash/pending.py.
+
+Covers four stateless helpers with no session or filesystem dependency:
+
+slash/image.py
+  - ``_mime_for_path``   — extension → MIME type lookup
+  - ``_file_size_human`` — byte count formatter (B / KB / MB)
+
+slash/pending.py
+  - ``_render_list``           — stalled-op list renderer
+  - ``_render_needs_attention`` — stuck-skill tail-section renderer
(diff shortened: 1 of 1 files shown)


tya5:
[tui-coder] Closing: main merged  +  (commit 009deed3) which cover the same , , ,  helpers. No new value to add.

PR #2403, opened 2026-07-02
fix(mcp): structural client-scope — robust-by-construction teardown + same-task guard (B)

tya5:
**[e2e-coder]** — MCP-lifecycle fix B (owner-directed structural hardening). Off clean main. Completes the task-affinity arc (A #2401 / C #2402 fixed the active bugs; B hardens the shared control_ir path robust-by-construction). Approach (c), lead-GO'd after plan-first.

## Why (not an active bug — a hardening)
Flow-trace showed B's crash class was already closed by the prior **G11 fix** (same-task teardown): control_ir ops run sequentially in the run-loop task, `teardown_mcp_clients()` has a single call site in that task's finally, chat closes in its own finally. Owner then directed: make it **robust-by-construction** rather than discipline-dependent. The cache exists for **subprocess reuse** (idempotent `initialize()` keeps the stdio subprocess alive across ops in a run), so full per-call (re-spawn) was rejected on perf grounds; approach (c) keeps reuse.

## Approach (c)
- **`ControlIRExecutor.mcp_client_scope()`** — an async CM entered ONCE per run in the run-owning task (`runtime.py` wraps `RunOrchestrator.run`). Records the owning task; on exit (success / exception / cancellation) closes every cached client in THAT task. Single close site bound to `__aexit__` — replaces the manual `teardown_mcp_clients()` finally line (can't be forgotten/duplicated). **Reuse preserved** (cache lives for the run).
- **op_runtime/mcp.py same-task guard** — fails FAST if an MCP op runs in a task other than the scope owner (a cross-task client open would leak an anyio cancel-scope → the "cancel scope crossed task boundary" crash). Latent hazard → loud immediate error. Unscoped chat per-call path (`mcp_owner_task=None`) is exempt.
- **`RunOrchestrator.run` asserts the scope is active** (fail-fast on a future caller bypassing the wrap). All runs flow through `OSRuntime.run`, which now wraps (verified: sole caller; tests use `rt.run`).

## Equivalence (broader change — shared handler + teardown contract)
- `call_mcp_tool` byte-identical (result shape unchanged; the guard is a pre-check that passes in the owner task).
- Teardown moves from the run `finally` to the scope `__aexit__` (after the run returns, **same task**; `complete()`/resume don't touch clients — functionally equivalent).
- Chat `_mcp_call_tool` (its own per-call finally, `mcp_owner_task=None`) unaffected.

## Falsify (real executor + real `_execute` + injected fake client, RED-verified)
- scope records the owner + closes on exit **in the owning task** + resets owner;
- scope closes on the **exception** path;
- op **fails fast** in a non-owner task (RED: strip guard → no raise);
- op **reuses** the cached client in the owner task (same instance — reuse preserved);
- unscoped path exempt.
Stripping the guard / the scope teardown fails the respective tests.

## CI gates
- ruff clean · tier-audit OK (0 Tier-4) · docstrings clean
- 760 mcp/control_ir/orchestrator/runtime tests pass + 9 run-orchestrator invariants
- pytest (full rootdir): running — confirm green before merge

## Test plan
- [x] 4-point falsify + RED-verified (guard + scope teardown)
- [x] equivalence: run-orchestrator + mcp suites green
- [ ] full suite green (running)

part of the MCP-lifecycle arc

### DIFF
--- a/src/reyn/core/kernel/control_ir_executor.py
+++ b/src/reyn/core/kernel/control_ir_executor.py
@@ -17,6 +17,8 @@ handlers are preserved (they run inside the invoker).
 """
 from __future__ import annotations
 
+import asyncio
+from contextlib import asynccontextmanager
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
@@ -160,7 +162,11 @@ class ControlIRExecutor:
         self._perm = permission_resolver
         self._skill_name = skill_name
         self._mcp_servers: dict = (mcp_servers or {}).get("servers", {})
-        self._mcp_clients: dict = {}  # cached across ops
+        self._mcp_clients: dict = {}  # cached across ops (subprocess reuse within a run)
+        # #B-hardening: the task that owns the MCP client scope for the current run. Set by
+        # ``mcp_client_scope`` (entered once per run in the run-owning task); the op handler's
+        # same-task guard reads it. None outside a scope (e.g. the chat per-call path).
+        self._mcp_scope_owner_task: "asyncio.Task | None" = None
         self._caller = caller
         self._chain_id = chain_id
         # FP-0021: run_id of the currently-executing OSRuntime run.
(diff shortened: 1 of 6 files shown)


tya5:
**[e2e-coder]** Closing — absorbed into the MCPClient structured-refactor arc (a359 root-cause). #2403's reyn-task run-scope was necessary-not-sufficient: it didn't restructure MCPClient itself, so close() still exits the SDK-internal task-group scopes cross-task (the actual Windows crash). The structured fix (MCPClient as an async CM, no deferred self._stack) supersedes it; #2403's run-scope idea + the same-task fail-fast guard move into P2 (the per-turn structured pool). Staged: P1 (structured CM + simple sites + Unix acceptance) → P2 (per-turn pool + op/executor/orchestrator, #2403 absorbed) → P3 (owner-Windows diagnostic).

Issue #2335, opened 2026-07-01
fix: read-bounding _self_bounded is dishonest for a single line exceeding the cap (char-level truncation needed) — #2296 gap


tya5:
**[lead-coder]** — Owner-found gap in #2296's read-bounding. The `_self_bounded` flag (which exempts a read result from the generic offload) claims "content is bound ≤ the inline cap BY CONSTRUCTION", but the LINE-based truncation does NOT guarantee that when a **single line exceeds the cap**.

## The gap (primary-verified, `core/op_runtime/file.py:~327`)

```python
cap = _read_inline_cap(ctx)
if len(content) > cap:
    all_lines = content.splitlines(keepends=True)
    shown = []; acc = 0
    for line in all_lines:
        if shown and acc + len(line) > cap:   # first line: shown is empty → never breaks
            break
        shown.append(line); acc += len(line)
    return {..., "content": "".join(shown), "next_offset": len(shown), "_self_bounded": True, ...}
```

The guard `if shown and ...` means the **first line is always appended whole**, even if it alone exceeds `cap`. So a file whose first line is, say, 500 KB (a minified JS bundle, a long JSON line, a data blob) → `content` = that 500 KB line → `len(content) >> cap` → yet `_self_bounded: True` is stamped.

**Consequence:** the offload exemption (#2296) trusts `_self_bounded` and does NOT offload → the 500 KB line sits INLINE in the LLM decide-context = the exact context-bloat the offload was built to prevent. The flag is dishonest.

## Fix direction

1. **Honesty (load-bearing correctness):** the read-bounding must ACTUALLY bound `content` to ≤ cap. When a single line overflows, **character-truncate** it to the cap boundary so the returned `content` is genuinely ≤ cap. Only then is `_self_bounded` true.
2. **Pagination completeness (secondary but required — no data loss):** `next_offset` is currently a LINE index (the next read does `lines[offset:offset+limit]`), so it cannot resume mid-line — paging a char-truncated huge line would skip its tail. The read pagination needs **character-capable offsets** (fully char-based, or a line+char hybrid) so the model can retrieve the rest of an over-long line via `next_offset`. Design the exact scheme in plan-first (char-based offset is cleanest; confirm the `offset`/`limit` consumers).

## Falsify-first (the edge #2296's tests missed)

- **RED without the fix:** a file whose FIRST line alone exceeds `cap` → current read returns `content` with `len(content) > cap` AND `_self_bounded: True` (assert the dishonest state). 
[...]

Issue #2296, opened 2026-07-01
fix: tool-result offload recursion — self-bounded file.read results re-offloaded (contradicts #1209)


tya5:
**[lead-coder]** — Owner-reported: the tool-result offload feature offloads `file.read` results, and retrieving an offloaded result (via `file.read` of the ref) is *itself* re-offloaded → an infinite offload/read recursion that confuses the LLM.

## Root cause (primary-verified)

`file.read` already **self-bounds + self-paginates** its result (#1209, `op_runtime/file.py:314-349`): an unbounded read whose content exceeds `_read_inline_cap` is truncated to a head window with `next_offset`/`shown_lines`/`total_lines`/`total_chars`, keeping it in the model's decide-context. The code comment states the explicit intent:

> `# ... keeps the content in the model's decide context instead of being offloaded out of view (the apply-starvation root cause, #1209); the model pages the rest via next_offset.`

**The generic offload contradicts this.** `offload_control_ir_result` (`context_builder.py:199`) offloads any result whose **whole JSON** (`json.dumps(result)`) exceeds `cap`. A file.read result is `content` (already ≤ cap by file.py) **plus envelope** (`kind`/`op`/`path`/`status`/`shown_lines`/`total_lines`/`next_offset`/`total_chars`/enc) → the serialized dict exceeds `cap` on envelope alone → the already-bounded read is offloaded anyway. Note `_read_inline_cap == control_ir_inline_cap` (same cap), so this trips essentially every non-trivial read.

**The recursion:** offloaded read → LLM's only retrieval path is `file.read(_offload_ref)` (per the #1383 inline: "the LLM may file.read for the full content"; `read_offloaded` is an internal Python helper, not an LLM-facing op) → the `.json` is > cap → file.py truncates it, and its result dict is again > cap on envelope → re-offloaded → loop. There is **no op-exclusion, no recursion cap, no termination guard** anywhere (grepped: zero matches for offload-exempt/skip/guard).

## Fix (bounded-by-construction; honor #1209)

**Exempt already-self-bounded read results from the generic control_ir offload.** A `file.read` result's content is guaranteed ≤ cap by file.py — its over-cap-ness is envelope-only, so it never needs offload; its correct retrieval mechanism is native pagination (`next_offset`), not an offload ref.

- Add the exemption in `maybe_offload_control_ir_results` / `offload_control_ir_result` (`context_builder.py:271`/`199`), before the size check.
[...]

PR #2417, opened 2026-07-02
fix: file_read truncates oversized reads (LLM-visible) instead of offload-duplicating

tya5:
**[e2e-coder]** — owner design steer on file_read offloading.

## The steer
Owner: a file_read's source **already lives on disk**, so offloading a *duplicate* copy of it is wasteful. A file_read should **truncate inline + reference the on-disk path**; only transient results (web fetch / exec — no source on disk) keep offloading. (This withdraws the earlier `_offload_payload_field: "content"` proposal — clean text, but still a duplicate copy.)

## Change (`op_runtime/file.py`)
An explicit LINE window whose slice **exceeds the inline cap** no longer returns verbatim (which then flowed to the generic control_ir offload, duplicating the file). It now **falls through to the self-bounding truncation path** — truncate to the cap, mark `_self_bounded` (offload-exempt), and surface a `next_offset` re-read continuation. Small explicit windows are still honored **verbatim** (byte-identical, #2335 common case).

## Owner requirement: truncation must be LLM-visible
The truncated result now carries an explicit **`_truncated: true`** marker plus a plain **`note`** naming the on-disk path and the re-read offset:
> `content truncated to fit context (N of M chars shown); the full file is on disk at 'path' — re-read from offset K to continue.`

So the model knows it holds a PART and can re-read the original file.

## Supersedes prior #2335/#2296 contract
Previously an explicit window was honored verbatim → handed to the generic offload when huge. Now **all file reads are self-bounded / offload-exempt**, which also **strengthens the #2296 recursion-termination guarantee** (a file_read can never start the offload/read recursion). The two existing tests that pinned the old verbatim-then-offload behavior are updated to the new owner-directed contract (docstrings note the supersession).

## Tests (Tier 2 — real Workspace + real `handle` + real `offload_control_ir_result`, no mocks)
- `test_file_read_truncate_marker.py` (new): large unbounded read → `_truncated` + `note` (names path + offset); truncated read is NOT offloaded (no duplicate file); small read has no marker (no regression).
- `test_read_single_line_char_truncation_2335.py`: small explicit window still verbatim; oversized explicit window now truncated + self-bounded + marker.
- `test_offload_recursion_2296.py`: oversized explicit-window read is now self-bounded → offload-exempt (no duplicate).

**RED-when-stripped verified**: removing the `_truncated`/`note` marker fails the CORE marker test.

- `ruff check src tests scripts` ✓
- `python scripts/test_tier_audit.py --strict` ✓ (15 tests)
- full `pytest` ✓ (8459 passed, 17 skipped, 3 xfailed)
- `python scripts/verify_module_docstrings.py` ✓

## Note
This is the **file_read (disk-backed) truncate** half. The **MCP holistic seam redesign** (whack-a-mole resolution, [1]-[4] contain-all) is a separate design-first assessment I'm presenting to lead next (GO-gated).

## Test plan
- [ ] Manual: interactive read of a large file → LLM receives a truncated result with a clear "truncated, full at <path>, re-read from offset N" note, and no duplicate offload file is written for the read.

### DIFF
--- a/src/reyn/core/op_runtime/file.py
+++ b/src/reyn/core/op_runtime/file.py
@@ -296,34 +296,38 @@ async def handle(op: FileIROp, ctx: OpContext, caller: Literal["preprocessor", "
         # byte-identical (no `encoding` field) for the common case.
         _enc_field = {"encoding": _detected_encoding} if _detected_encoding else {}
         # #2335: read MODE. An explicit LINE window (offset/limit given, no char_offset) is honored
-        # VERBATIM — the LLM's line-based read contract, byte-identical (a huge explicit window is
-        # NOT self-bounded → the generic offload handles it; consistent + non-recursing since its
-        # retrieval is an unbounded read → self-bounding → exempt). Otherwise (an unbounded read, OR
-        # a char_offset mid-line RESUME) the read is SELF-BOUNDING.
+        # VERBATIM — the LLM's line-based read contract, byte-identical — AS LONG AS its slice fits
+        # the inline cap. Otherwise (an unbounded read, a char_offset mid-line RESUME, OR an explicit
+        # window whose slice exceeds the cap) the read is SELF-BOUNDING (truncate + re-read hint).
+        cap = _read_inline_cap(ctx)
         explicit_line_window = (
             op.offset is not None or op.limit is not None
         ) and op.char_offset is None
         if explicit_line_window:
-            # Caller asked for an explicit LINE window — honor it verbatim.
             lines = content.splitlines(keepends=True)
             start = op.offset or 0
             sliced = lines[start:start + op.limit] if op.limit is not None else lines[start:]
-            ctx.events.emit("tool_executed", op="read_file", path=op.path)
-            return {
-                "kind": "file",
-                "op": "read",
-                "path": op.path,
-                "status": "ok",
-                "content": "".join(sliced),
-                **_enc_field,
-            }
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — review PASS（CI green で merge）。owner steer（truncate-don`t-duplicate + LLM-visible）を正確に実装。

- explicit line window：`len(joined) <= cap` → verbatim 維持（#2335 byte-identical common case）、**oversized → self-bounding truncate に fall-through**（offload 複製せず、元ファイルは op.path に既存）。
- truncation result に **`_truncated: True` + `note`**（"content truncated (N of M chars shown); full file on disk at `<path>` — re-read from offset K"）= owner の「LLM が truncate を認識」要件を明確に充足 + `_self_bounded: True`（#2296 offload-exempt、recursion-termination 強化＝file_read が offload recursion を起こせなくなる副次効果 good）。
- test：real Workspace + real handle + real offload_control_ir_result（no mock）、`_truncated`/`note`/no-duplicate-offload/small-no-marker（no regression）、RED-when-stripped。#2335/#2296 の旧 verbatim→offload contract supersession も該当 test 更新済み・妥当。

**manual test-plan（interactive 大ファイル read の見え方）は post-merge に tui へ**（automated Tier-2 が contract を cover、tui が LLM-visible marker の interactive 表示を裏取り）。CI green で merge。

(A) MCP seam の holistic 設計 assessment 提示、待ってます（GO-gated）。

Issue #2408, opened 2026-07-02
[a359] MCP task-affinity crash arc — tracking (P1/P2/P3 landed; owner-Windows confirmation + diagnostic removal pending)


tya5:
**[lead-coder]** — tracking issue for the a359 MCP task-affinity crash arc (owner-reported: `list_mcp_tools` / MCP tool calls crash chat with `BaseExceptionGroup` / `BrokenResourceError` / `ConnectionReset` on Windows).

## Root cause
`MCPClient`s deferred `AsyncExitStack` closed the SDK `stdio_client`/`ClientSession` internal anyio task-group scopes **cross-task** → "cancel scope crossed task boundary" (Windows Proactor: subprocess/job-object teardown ordering → BrokenResource/BaseExceptionGroup). Unix (Selector loop) tolerates it → cannot be RED-verified on Unix.

## Landed (Unix-side complete)
- **P1 #2405** — `MCPClient` structured async-CM (`__aenter__`/`__aexit__`, one task/scope); one-shot sites (`_mcp_list_tools`, probes) migrated; supersedes/removes A(#2401)+C(#2402).
- **P2 #2406** — per-turn `MCPClientPool` (open+reuse+close in run-owning task, fail-fast off-task) + fault-isolation boundary: any MCP fault (transport BaseExceptionGroup / bad response / server-death / **hung server via finite 120s default timeout, overridable, <=0 opt-out**) → **contained error result WITH fault content fed to the LLM** (owner req); control-flow exceptions (CancelledError/KeyboardInterrupt/SystemExit) always propagate. Absorbs #2403.
- **P3 #2407** (temporary) — `reyn.mcp.a359diag` INFO diagnostic (open-task/close-task/outcome) + `docs/dev/mcp-a359-windows-verification.md` procedure + diag-emit test. `TODO(a359-cleanup)` markers on all 3 removal points.

## PENDING (close conditions)
- [ ] **owner-Windows sufficiency confirmation**: run per `docs/dev/mcp-a359-windows-verification.md` — harness all `SURVIVED` on Proactor (incl cross_task_close/no_close_gc, CRASHED pre-fix) + `list_mcp_tools` diag shows `open_task==close_task`/`outcome=ok`/no BaseExceptionGroup. (owner runs)
[...]

PR #2418, opened 2026-07-02
[design] MCP holistic seam assessment (defeat-all-hypotheses) — GO-gated

tya5:
**[e2e-coder]** — DESIGN-FIRST assessment (no impl). Owner-directed non-whack-a-mole MCP fix.

Draft PR for review only — **owner GO required before build**. Full assessment in `docs/dev/mcp-holistic-seam-design.md`.

## TL;DR
The whack-a-mole root cause: fault-isolation + lifecycle + timeout were applied **per entry path**, so the list/probe paths were swept-missed and the crash recurred. Fix class = a **single `MCPGateway` seam** every MCP op routes through, enforcing the four guarantees in one place:

- **[1] structured sub-task join** — the client's teardown awaits the SDK's internal task group in-scope → an SDK sub-task fault can never orphan to loop-level ("Task exception was never retrieved"). Defeats hypothesis (b).
- **[2] contain-all boundary + real-control-flow predicate** — `except BaseException`; re-raise only genuine control flow (`KeyboardInterrupt`/`SystemExit`, or `CancelledError` with `current_task().cancelling() > 0`), contain everything else → error result. Exception-structure-independent → defeats (a1) cancel-mixed BEG, (a2) all-Exception group, (c) transport; spurious internal cancels are stripped.
- **[3] pool task-affinity** (reuse a359 P2) + **[4] per-call timeout** (reuse S3) consolidated here.

Path 1 (op-call) **already** has this exact boundary — the seam generalizes it to a shared gateway; paths 2–6 become thin callers. **Completeness is structural**: a grep gate asserts no `MCPClient(` outside the gateway, so a new path can't bypass it.

## Acceptance (traceback-independent)
Owner can't paste the traceback → design defeats ALL hypotheses at once. Inject a1/a2/b/c/d at {initialize/list/call/teardown} across list/call/probe → assert contain-all (reyn survives + error result), real control flow propagates; + structural grep gate. Owner final = Windows yes/no. Injection harness already prototyped (a359 repros).

## Open questions (in the doc §6)
Gateway module vs extend pool; pool sharing (one-shot vs per-turn); confirm MCP results offload as transient; build sequencing (+ a359-DIAG removal fold-in).

Requesting: lead review → present to owner for GO → I build.

### DIFF
--- a/docs/dev/mcp-holistic-seam-design.md
+++ b/docs/dev/mcp-holistic-seam-design.md
@@ -1,0 +1,147 @@
+# MCP holistic seam — design assessment (defeat-all-hypotheses)
+
+**Status:** design-first / pre-implementation. **GO-gated** (owner approval before build).
+**Basis:** current code on `origin/main` (per lead — verify with `git show origin/main:<path>`).
+**Author:** e2e-coder. **Motivation:** owner — "MCP 修正を繰り返してる、モグラ叩きにならない
+綺麗な直し方を再検討して" (we keep patching MCP per-path; find a clean, non-whack-a-mole fix).
+
+## 1. Why the whack-a-mole happens (root cause)
+
+Fault-isolation + lifecycle + timeout were each applied **per entry path**, not at a shared seam:
+
+- **P2 (#2406)** put a task-affine pool + a contain-all boundary on the **op-call** path only.
+- **S3 (#…)** put per-call timeout on the op-call path only.
+- The **list** and **probe** paths kept `async with MCPClient(...)` + `except Exception` — which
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — close as superseded. The MCP holistic seam this doc designed was **built + merged (#2421)**, owner-Windows validated (no crash recurrence), and the arc issue **#2408 is closed**. The design rationale (whack-a-mole root → single gateway seam, 4 guarantees, contain-all boundary + `is_real_control_flow`, CI-enforced completeness grep, real-substrate dying-subprocess test) lives in the code docstrings (`gateway.py` / `pool.py`) + the closed arc issue. A pre-build design draft on top of the built reality would be drift (owner: no tech debt). Closing.

Issue #2414, opened 2026-07-02
ask_user free-text answer unreachable in inline CUI — submit routes to chat router, not intervention bus (+ Ctrl-C hang)


tya5:
**[lead-coder]** — found via tui-coder ask_user intervention dogfood (primary evidence: skill + chat event logs). Inline CUI only (approval prompts + dropdown UI work fine).

## I1 (HIGH): ask_user free-text answer never reaches the intervention bus
Repro: `skill_importer` "document or text processing" → search → select phase (5 candidates) → `ask_user` fires. CUI renders the question correctly (◆ block + numbered list + options), but typing an answer at the `❯` bar and submitting does NOT resolve the intervention.

**Root cause**: the inline `❯` input bar always routes a submit as a NEW chat-router turn. During an active `ask_user` it does the same → the answer never reaches the intervention bus. (Python/HTTP approval prompts work because their dropdown UI calls the intervention bus directly.)

**Primary evidence** (skill event log): `user_intervention_requested (phase: select)` → `user_intervention_received (answer: "")` (empty, only on Ctrl-C) → re-request. Chat events show **0** `user_answered_intervention (kind: ask_user)`, while the 4 Python/HTTP approvals all fired `user_answered_intervention` correctly.

## I2 (MEDIUM): Ctrl-C during an ask_user hang cannot complete teardown
`✗ Cancelling…` shows but never completes — `await intervention_bus.request(iv)` is not unblocked, CancelledError does not propagate; only `pkill` ends it. Event log: `✗ Cancelling…` → bus resolves empty `""` → 2nd ask_user → hangs again; no `workflow_finished/aborted`.

## Fix direction
Inline CUI must become **state-aware**: when a `user_intervention_requested (kind=ask_user)` is active, route the next `❯` submit to the intervention bus (not the chat router) — or add a dedicated answer component like the approval prompt. I2 likely resolves alongside (the bus must also honor cancellation).

## Positive (works)
Python/HTTP approval prompts: `◆ answered: [y]es` ×4 + correct `user_answered_intervention` events. ask_user DISPLAY is correct; only the answer ROUTING is broken.

roi:high — ask_user free-text is unusable in the inline CUI (a core interaction path); Ctrl-C hang compounds it. TUI-lane fix.

tya5:
[...]

PR #2416, opened 2026-07-02
fix(#2414-I1): route free-text submit to ask_user intervention bus

tya5:
**[tui-coder]** — part of #2414

## Summary

- `_submit()` in the inline CUI called `submit_user_text()` unconditionally, sending free-text answers to the chat router instead of the intervention bus when an `ask_user` intervention was pending
- Add `Session.answer_oldest_intervention_text(text)` that delegates to the existing `_deliver_answer_to()` seam
- `_submit()` now checks `s.interventions.head()` and routes to `answer_oldest_intervention_text()` when `head.kind == "ask_user"`
- **I2 (Ctrl-C hang during ask_user)** stays open in #2414 — not addressed here

## Scope

- `src/reyn/interfaces/inline/app.py` — `_submit` routing check
- `src/reyn/runtime/session.py` — new `answer_oldest_intervention_text` method
- `tests/test_inline_intervention_region.py` — two new Tier-2 tests

## Tests

Two new Tier-2 tests added:

1. **`test_answer_oldest_intervention_text_delivers_free_text`** — live Session + real intervention dispatch. RED-verify: removing the `head.kind == "ask_user"` routing check causes `_deliver_answer_to` to never be called → intervention future never resolves → `asyncio.wait_for` times out.

2. **`test_submit_routes_to_intervention_bus_when_ask_user_pending`** — stub-session routing isolation. Verifies `answer_oldest_intervention_text` is called instead of `submit_user_text`. RED-verify: removing the routing check calls `submit_user_text` instead.

No regressions: existing 8 tests in `test_inline_intervention_region.py` all pass. Other intervention kinds (approval etc.) are unaffected — the check only routes when `head.kind == "ask_user"`, all other cases fall through to `submit_user_text` as before.

## CI gates (all green)

- `ruff check` ✓
- `test_tier_audit.py --strict` ✓ (10 tests, 0 Tier-4 violations)
- `verify_module_docstrings.py` ✓
- `pytest tests/test_inline_intervention_region.py` ✓ (10 passed)
- `pytest -x -q` (full suite) ✓

## Test plan

- [x] Full pytest suite passes
- [x] (skipped — no live ask_user dogfood path available in CI) Manual: `reyn chat`, trigger an `ask_user` phase, type an answer in the inline CUI input bar and verify it resolves the intervention (not routed to chat)

## Tier self-check

New tests are Tier 2 (OS invariant — routing contract between `_submit` and intervention bus). No Tier-4 violations. RED-verify documented inline.

PR #2419, opened 2026-07-02
fix(#2414-I2): propagate CancelledError through intervention dispatch — fix Ctrl-C hang

tya5:
**[tui-coder]** — fix for #2414 I2 (Ctrl-C teardown hang during ask_user)

## Root cause

`InterventionRegistry.dispatch` and `InterventionCoordinator.dispatch` both had:

```python
except asyncio.CancelledError:
    return InterventionAnswer(text="")
```

When Ctrl-C triggered `cancel_inflight()` → `task.cancel()` during an `ask_user` wait, `CancelledError` was raised at `await iv.future` but caught and swallowed. The skill task received an empty answer and **continued running** — immediately re-requesting the intervention and hanging again. Only `pkill` could end it. Event log: `✗ Cancelling…` → bus resolves `""` → 2nd ask_user → hangs again.

## Fix

Remove the `except asyncio.CancelledError` swallow from both sites. Let it propagate so the skill task is actually cancelled (`task.cancelled() == True`). The registry's `finally:` block still runs cleanup (active pop + order remove + `_maybe_announce_next`).

## Files changed

- `src/reyn/runtime/services/intervention_registry.py` — remove inner `except CancelledError: return empty`; update docstring
- `src/reyn/runtime/services/intervention_coordinator.py` — remove `try/except CancelledError: return empty` around stalled-path `await iv.future`; remove now-unused `import asyncio`
- `tests/test_intervention_coordinator.py` — 2 new Tier-2 tests: `task.cancelled() == True` after cancel (the old swallow made it False)

## Test plan

- [x] `pytest tests/test_intervention_coordinator.py` — 7/7 pass (incl. 2 new tests)
- [x] `pytest tests/test_await_quiescent.py tests/test_intervention_handler_invariants.py tests/test_pending_intervention_268.py tests/test_session_intervention_persistence.py` — 37/37 pass
- [x] `ruff check` — clean
- [x] `python scripts/test_tier_audit.py --strict` — clean
- [x] `python scripts/verify_module_docstrings.py` — clean

## Evidence (how I found this)

Investigating #2414-I2. The `except asyncio.CancelledError: return empty` pattern in long-lived async tasks is the exact anti-pattern from project feedback (`feedback_track_background_verification_hang_not_slow.md`). The old "cancellation contract" in the docstring was wrong: it caused task.cancelled() to be False, meaning the skill was never actually cancelled.

part of #2414

**Tier self-check**: new tests are Tier 2 (OS invariant — intervention dispatch cancellation contract). No Tier 4 violations.

### DIFF
--- a/src/reyn/runtime/services/intervention_coordinator.py
+++ b/src/reyn/runtime/services/intervention_coordinator.py
@@ -12,7 +12,6 @@ dispatch path and ``ChatInterventionBus`` both consult.
 """
 from __future__ import annotations
 
-import asyncio
 import logging
 from typing import TYPE_CHECKING, Callable
 
@@ -134,9 +133,6 @@ class InterventionCoordinator:
                 origin_channel_id=iv.origin_channel_id,
             )
             self._registry.park_stalled(iv)
-            try:
-                return await iv.future
-            except asyncio.CancelledError:
-                return InterventionAnswer(text="")
+            return await iv.future
         # Default: route through the regular InterventionHandler.
         return await self._handler.dispatch(iv)
--- a/src/reyn/runtime/services/intervention_registry.py
+++ b/src/reyn/runtime/services/intervention_registry.py
@@ -267,22 +267,21 @@ class InterventionRegistry:
         """Register *iv* in the queue, announce or signal queued status, then
         await the user's response.
 
-        Always removes the entry on exit so a cancelled skill does not leave
-        dangling queue entries.  On ``asyncio.CancelledError`` the future is
-        cancelled and an empty ``InterventionAnswer`` is returned to the
-        caller (same contract as the original ``_dispatch_intervention``).
+        Always removes the entry on exit (via ``finally``) so a cancelled
+        skill does not leave dangling queue entries.  ``asyncio.CancelledError``
+        is re-raised so the calling task is properly cancelled — swallowing it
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — review PASS（CI green で merge）。I2（Ctrl-C hang）の textbook-correct な fix。

- root：coordinator/registry 両 `dispatch` の `except asyncio.CancelledError: return InterventionAnswer(text="")` が cancel を swallow → skill task が空 answer で継続・intervention 再 request → hang。まさに CancelledError=BaseException を swallow する anti-pattern（cleanup は finally、cancel は propagate すべき）。
- fix：両 site の swallow 除去 → `return await iv.future`（propagate）、registry の `finally:`（active pop / order remove / announce-next）は保持＝cleanup 保証、誤 "cancellation contract" docstring 訂正、unused import 除去。
- `enforce_listener_presence` の empty-answer path は保持（no-listener refusal＝非 cancel の正当 path、混同なし）。
- test 2本 Tier-2：stalled-path + registry-path で `task.cancel()`→`task.cancelled()==True`（RED: 旧 swallow で silently complete=False）、real instance no-mock。37/37 related pass＝regression なし（cancel-on-cancel が正、継続が bug だった）。

CI green で merge → #2414 完全解決（I1 #2416 + I2 #2419）。

PR #2432, opened 2026-07-02
fix(budget): drop duplicate agent= kwarg in budget_warn emit (agents/ path)

tya5:
**[tui-coder]** — dogfood mining find

## Root cause

`_record_budget_post_llm()` in `llm_call_recorder.py` emitted `budget_warn` with BOTH an explicit `agent=agent` kwarg AND `**check.context` where `check.context` (from `BudgetTracker.record_llm()` → `_agent_context(agent)`) already carries `"agent"`. Python raises `TypeError: emit() got multiple values for keyword argument 'agent'` at runtime.

Trigger condition (all three required):
1. Caller is from an agent (`caller="agents/<name>"`)
2. `per_agent_tokens` (or `per_agent_cost_usd`) warn threshold is configured
3. Threshold is crossed during a skill run

This is latent — fires only with budget warn config, which is optional. The `_check_budget_pre_llm()` twin emit is safe (rate-limit context has no `"agent"` key).

## Fix

Drop the explicit `agent=agent` kwarg from `_record_budget_post_llm()`'s emit; `check.context` already provides it when non-None. For the `agent=None` case, `check.context = {}` so no "agent" in the event (correct — no agent to report).

## Changes

- `src/reyn/core/kernel/llm_call_recorder.py` — remove duplicate `agent=` kwarg
- `tests/test_llm_call_recorder_invariants.py` — Tier 2 test pinning the fix

## Test plan

- [x] `pytest tests/test_llm_call_recorder_invariants.py` — all 4 pass, including new test
- [x] `pytest tests/test_budget_auto_save.py tests/test_budget_e2e_resume.py tests/test_budget_gateway_invariants.py tests/test_budget_memo_hit_forward_calc.py` — all 18 pass
- [x] `ruff check src/reyn/core/kernel/llm_call_recorder.py tests/test_llm_call_recorder_invariants.py` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/test_llm_call_recorder_invariants.py` — 4/4 OK

Tier self-check: new test is Tier 2 (OS invariant — event schema contract), no mocks, public surface assertions only.

### DIFF
--- a/src/reyn/core/kernel/llm_call_recorder.py
+++ b/src/reyn/core/kernel/llm_call_recorder.py
@@ -711,10 +711,12 @@ class LLMCallRecorder:
             purpose="phase",  # #1190: the OS phase-execution LLM path
         )
         for dim in check.warn_dimensions:
+            # check.context from record_llm already carries "agent" when
+            # non-None (via _agent_context). Passing agent= explicitly too
+            # would raise TypeError ("multiple values for keyword argument").
             self._events.emit(
                 "budget_warn",
                 dimension=dim,
-                agent=agent,
                 chain_id=self._chain_id,
                 **check.context,
             )
--- a/tests/test_llm_call_recorder_invariants.py
+++ b/tests/test_llm_call_recorder_invariants.py
@@ -258,3 +258,65 @@ async def test_budget_post_record_increments_usage(tmp_path, monkeypatch):
     assert usage.prompt_tokens == 200
     assert usage.completion_tokens == 60
     assert usage.total_tokens == 260
+
+
+# ── Test 4: budget_warn emitted cleanly when per-agent threshold fires ─────────
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — review PASS（CI green で merge）。`_record_budget_post_llm` の `agent=agent` 明示渡しが `check.context`（`_agent_context` 由来で既に "agent" を含む）と衝突 → `TypeError: multiple values for keyword argument "agent"`。fix＝明示 `agent=agent` 除去（`**check.context` の agent を使う）、2-line + comment。Tier2 RED-verify（fix 前 TypeError）。latent（per-agent budget warn config 要）だが correct。

**注（非 blocking）**：`llm_call_recorder.py` は skill 削除 scope（phase engine）——この fix は correct だが削除で moot になりうる。interim value あり + cheap なので merge するが、**dogfood は KEEP surface（chat/RouterLoop・task-system・permission・MCP gateway）に寄せると無駄が少ない**（phase engine は削除予定）。CI green で merge。

Issue #2439, opened 2026-07-02
stage3-residual: dead skill-type-refs cleanup after machinery deletion (#2434 follow-up)


tya5:
**[e2e-coder]** — Follow-up tracking issue for the non-breaking residual left after the #2434 skill/phase machinery bulk-delete (#2438). Deleting the `Skill`/`Phase` schema classes leaves **dead but non-breaking** references (TYPE_CHECKING-gated / string-annotation / always-None fields — full suite is green on them). Tracked here so the no-debt close is auditable rather than silently deferred.

### 1. `OpContext.skill` — always-None dead field
- `src/reyn/core/op_runtime/context.py`: the `skill: "Skill | None" = None` field + its `TYPE_CHECKING` import of the deleted `Skill` class.
- KEEP op handlers (`tools/sandboxed_exec.py`, `tools/mcp.py`, `tools/web_search.py`) construct `OpContext(..., skill=None)` — the field is always `None` (no skill execution sets it).
- **Cleanup**: remove the field + the 3 `skill=None` kwargs + the TYPE_CHECKING import. Bounded, low-risk.

### 2. `permissions.py` skill-declaration handling — DELICATE (security-adjacent)
- `src/reyn/security/permissions/permissions.py`: `startup_guard(self, skill: "Skill", ...)` (dead — **zero functional callers** after skill deletion) + its skill-phase-declaration scanning + the `TYPE_CHECKING` `Skill` import.
- This is intermixed with the **KEEP chat permission gate** (`_prompt_file_access` etc.), so it needs careful separation (same rigor as the stage1 `session.py` surgery) — NOT a rushed strip.
- **Cleanup**: remove `startup_guard` + skill-declaration-only helpers, keep the chat gate, verify chat ask_user/approval unaffected.

### 3. Stale prose comments (~10 files)
- Comments/docstrings mentioning `SkillRuntime`/`SkillRunner`/`SkillRegistry`/`run_skill`/`ControlIRExecutor` as live components — e.g. `llm/model_resolver.py`, `config/infra.py`, `runtime/registry.py`, `runtime/services/intervention_handler.py`, `runtime/cron/scheduler.py` (`CronJob.skill` field + comment), `reporters/__init__.py`, `tools/{registry,memory,types}.py`, `runtime/session.py`, `runtime/router_loop.py`.
- **Cleanup**: update to reflect the deleted machinery (P7 accuracy). Cheap.

## Notes
[...]

Issue #2434, opened 2026-07-02
[arc] Delete the heavy DSL/phase skill feature → chat-unified + lightweight skill (owner GO 2026-07-02)


tya5:
**[lead-coder]** — umbrella tracking for the skill-feature deletion arc. Owner GO (autonomous mode, no tech debt / clean-break, cost-optimize with Sonnet for mechanical).

## Direction (3 decoupled tracks)
1. **NOW — delete the heavy custom skill abstraction** (DSL phase-graphs): unusable in practice, diverges from industry, LLM cannot author the rigid graphs (no parallel+join; spec too complex). Chat already runs without skills (RouterLoop omits invoke_skill when none registered) → chat is the durable substrate.
2. **AFTER — add a lightweight INDUSTRY-STANDARD skill** (folder + instructions in context, Claude-Agent-Skills-style; no DSL, no phase engine).
3. **SEPARATE FUTURE (decoupled)** — grow the task-system into the deterministic control loop (goal/termination-condition driven).

## Key facts (primary-evidence verified vs origin/main)
- Phase engine = `src/reyn/core/kernel/` (22 modules); `src/reyn/kernel/` is a stale empty dir.
- Chat turn = `_run_router_loop`→RouterLoopDriver→RouterLoop→RouterHostAdapter — **NOT SkillRuntime-dependent** (router_loop does not import SkillRuntime).
- `_run_stdlib_skill` (skill_router path) has **zero src callers = dead code**; A2A/MCP inbound is **router-direct** (skill-independent) → survives deletion.
- Interventions (ask_user/approvals) are **SHARED** (chat/inline-CUI + MCP/A2A via #268 listener) → `InterventionCoordinator` STAYS; only the `running_skills_chain` skill-linkage is cut.
- `#292` override-observer (peer-notify-on-ask_user) is **already dead for router-direct A2A/MCP** → dropped clean-break; #268 delivery listener KEPT.
- `auto_resume` = skill-run-only (chat/narrator path removed in 59c991a) → DELETE.

## Staged plan (each CI-green + reviewed + no half-broken state)
- **PR-0 (#2433)** — relocate `_split_frontmatter` → `core/frontmatter.py` (no-dup for non-skill consumers file.py/memory.py). [prerequisite]
[...]

PR #2441, opened 2026-07-02
docs(#2434): stage3.5 (A) mechanical skill/DSL doc purge

tya5:
**[docs-maintainer]** — Stage 3.5 (A): mechanical removal of all skill/DSL/Phase machinery docs whose backing implementation was deleted in #2438.

part of #2434

## Summary

- **91 pages deleted**: `docs/concepts/skills/` (14), `docs/guide/for-skill-authors/` (57), `docs/reference/dsl/` (16), `docs/reference/cli/eval.md`, `eval.ja.md`, `skills.md`, `skills.ja.md`
- **Nav updated** (`.mkdocs/mkdocs.yml`): removed "For skill authors" guide section, DSL reference section, Skills concepts section, `reyn eval` + `reyn skills` CLI entries
- **Partial strips** (66 files, cross-ref agent): all dangling links from surviving docs to deleted pages stripped to plain text or See Also items removed
- **feature-map.md**: removed `run_skill`/`lint`/`skill_resolve` ops, DSL section, `skill_builder/importer/improver/search` stdlib rows, `reyn eval`/`reyn skills` CLI rows, skill-level permissions row; session model updated from 3-level → 2-level (SkillRuntime removed)
- **reyn-yaml.md + .ja.md**: removed `skill_search`, `skill_resume`, `self_improvement` blocks
- **concepts/multi-agent/multi-agent.md + .ja.md**: removed Layer 1 (`@sub_skill`) and Layer 2 (`run_skill`) — diagram, table, blockquote, when-to-pick, agent-ID section

## Intentional page drops (nav-verify)

All pages listed above are intentional drops. The surviving nav has 0 broken references verified by:

```sh
grep -oP "(?<=: )[a-z][\w/-]+\.md" .mkdocs/mkdocs.yml | while read path; do
  if [ ! -f "docs/$path" ]; then echo "MISSING: docs/$path"; fi
done
# → no output
```

## Build

- `mkdocs build` (non-strict): **passes** — "Documentation built in 4.33 seconds"
- `mkdocs build --strict`: **24 warnings** — all pre-existing EN↔JA anchor parity issues in origin/main (JA headings use Japanese text generating Japanese anchor IDs; links reference English anchor IDs). None introduced by this PR. Tracking: open follow-up.

## Test plan

- [x] Nav check: 0 missing pages in mkdocs.yml nav
[...]

PR #2443, opened 2026-07-02
refactor(#2439): remove dead phase-loop config knobs + PhaseConstraints (residual 3/4)

tya5:
**[e2e-coder]** — Part of #2439 (stage3-residual cleanup after #2434 machinery deletion). **Item 3 — dead phase-loop knobs.**

## Removed (all dead — enforcers were in the deleted phase executor / skill_node_runner act-loop)
- **`LoopConfig.max_phase_visits`** (phase-visit cap) + **`LoopConfig.max_act_turns_per_phase`** (act-turn cap). Audited dead: no KEEP router/op-loop enforces them (`router_loop`/driver zero refs); the chat loop is capped by the **LIVE** `max_router_calls_per_turn` / `max_tool_calls_per_turn` (untouched). `max_act_turns` was enforced by the deleted `skill_node_runner` act-loop — genuinely dead, **not** live-with-stale-name (so no rename).
- **`schemas.PhaseConstraints`** class + **`ContextFrame.constraints`** / **`ContextFrame.remaining_act_turns`** fields — write-only now (the phase executor that read them is deleted).
- Cascade: `config/chat.py` (fields + loader), `context_builder`, `OpContext.max_phase_visits`, the 3 op handlers, CLI config surface, `schemas/__init__` export, + stale prose (`limit_handler`, `turn_budget` messages).

## Config-compat
The tolerant `loop_raw.get(key, default)` pattern is preserved — an existing `reyn.yaml` carrying `max_phase_visits:` / `max_act_turns_per_phase:` is **silently ignored** (no hard-error). Removed keys degrade gracefully.

## ⚠️ Scope correction (lead-confirmed (b)) — the audit-gap catch
My original fix-class sweep listed `skill_calls_per_chain`/`skill_tokens_per_chain` as dead siblings. On execution I caught that my `.loop.<field>` (attribute) grep **missed the `getattr(loop, "skill_calls_per_chain", ...)` string-literal consumer** in `budget.py` + its **8 live tests** — the #1911 durable skill-spawn-cap subsystem (dead-in-production but test-covered). **Those two knobs + the whole budget subsystem are EXCLUDED here** and tracked as a coherent follow-up in #2439 (recovery-aware: `BudgetLedger` durability may trigger the CLAUDE.md truncate-falsify gate). `skill_calls/tokens` config fields + budget tests are kept intact.

## Test plan (all 4 gates green)
- [x] `pytest` (full, worktree src) — **7006 passed** (exit 0)
- [x] tier-audit `--strict` on changed tests — 0 errors
- [x] `verify_module_docstrings` — clean
- [x] Absence-check (worktree src): `max_phase_visits`/`max_act_turns_per_phase`/`PhaseConstraints`/`remaining_act_turns` functional-zero (only session.py prose remains → tui-coordinated item#4)
- [x] Config-compat verified (tolerant get); LIVE knobs (`max_router_calls_per_turn`/`max_tool_calls_per_turn`) intact + tested

Note on verification: bare `python -c` in a git-worktree resolves `reyn` to the **main** repo's editable install, not the worktree — worktree verification must use pytest (which uses worktree src) or `PYTHONPATH=<worktree>/src` ([[feedback_ab_arm_isolation]]).


tya5:
**[lead-coder]** — review: code removal is clean (phase-knobs gone from config, skill_calls/tokens correctly excluded per (b), zero dangling, CI 5/5 green). **One completion before merge:**

**Add the reyn-yaml.md strip for the removed knobs** — `docs/reference/config/reyn-yaml.md` still documents `max_phase_visits` (L396, L427) and `max_act_turns_per_phase` (L397, L428), now removed from the config code. CI passes (the mirror test only flags fields-in-code-missing-from-doc, not stale-doc), but this is stale doc = debt.

The principle: **a config-doc strip belongs in the same PR that removes the config code** (mirror-coupled). This PR removes the code, so it should strip the doc too — this is exactly the pattern #2441 got wrong (doc-strip without code-removal → mirror fail). Strip those 4 lines → re-push → I merge. Otherwise clean, nice work.

PR #2446, opened 2026-07-03
fix(inline-cui): harden soft breaks in agent markdown output

tya5:
**[tui-coder]**

## Summary

- Adds `_harden_soft_breaks()` in `renderer.py` that converts bare single newlines in paragraph text to CommonMark hard line breaks (`  \n`)
- Fixes: LLM output with single-newline-separated lines (e.g. `apple\nbanana\ncherry`) was collapsed to one line (`apple banana cherry`) by `rich.Markdown`'s CommonMark soft-wrap behavior
- Structural lines (headings, list items, blockquotes, code fences, blank lines) are exempt — the markdown parser needs raw newlines around them
- Applied in `_body_renderable` for `kind == "agent"` only (the LLM reply path)

## Root cause

CommonMark spec: a single `\n` inside a paragraph is a *soft line break* → rendered as a space. Only `  \n` (two trailing spaces) or `\n\n` (blank line) create visible breaks. `rich.Markdown` follows this faithfully.

LLMs routinely return plain multi-line text without markdown list markers. Before this fix, `count to 5, one per line` → `⏺ 1 2 3 4 5`; after → each number on its own hang-indented line.

## Dogfood verification

`respond with exactly three lines: apple, banana, cherry`:
- **Before**: `⏺ apple banana cherry` (collapsed)
- **After**: `⏺ apple\n  banana\n  cherry` (three distinct lines, hang-indented under the ⏺ gutter)

## Test plan

- [x] `pytest tests/test_inline_renderer_cutover.py` — 4 new tests (pure `_harden_soft_breaks` contract × 2, rendered output × 2)
- [x] `python scripts/test_tier_audit.py --strict tests/test_inline_renderer_cutover.py` — 30 OK, 0 Tier-4
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/repl/renderer.py` — OK
- [x] Full `pytest` — 6959 passed, 17 skipped, 1 xfailed

**Tier self-check**: new tests are Tier 2 (pure-function contract on `_harden_soft_breaks` + rendered-output behavioral assertions via `format_inline_message`). No mocks, no private state, no format-pin.

part of #2193


tya5:
**[lead-coder]** — review: the soft-break fix is right for prose, but **it corrupts fenced code-block content**. `_harden_soft_breaks` (called on the whole text at renderer.py:458) has no code-fence state tracking, and `_STRUCTURAL_LINE_RE` matches fence *delimiters* (``````/`~~~`) but not fence *content* lines.

Trace for a code block:
```
input:  ```python / x = 1 / y = 2 / ```
- "x = 1": not structural, next ("y = 2") not structural → gets `  ` appended
- result: "x = 1  " — trailing spaces injected into the code
```
Inside a fence, `  \n` is not a hard break — it's literal content, so the code now carries trailing whitespace (invisible on screen, but wrong on copy / for whitespace-significant content). For a coding assistant, code blocks are core content, so this is worth fixing before merge.

**Fix**: track fenced-code state (toggle on ``````/`~~~` delimiter lines) and skip hardening for lines *inside* a fence. A test with a multi-line code block asserting the content is byte-preserved would pin it. Otherwise the prose-hardening logic + structural exclusions look correct. Re-push with the fence guard → I'll re-review.

PR #2491, opened 2026-07-03
fix(#2491): recall chunks + error_message ⎿ rows stop leaking raw dicts

tya5:
**[tui-coder]** — dogfood mining: ⎿ row summary leakage, two gaps

## Gap 1 — `rag_operation__recall` success: `chunks` list not handled

`execute_op` returns `{"chunks": [...], "mode": "recall"}` with no `op`, `status`, `matches`, `entries`, or `tasks` key — fell through to `_short(repr(result), 80)`.

**Fix**: add `chunks` list branch after `matches`:
```python
chunks = result.get("chunks")
if isinstance(chunks, list):
    n = len(chunks)
    return f"{n} chunk{'s' if n != 1 else ''}"
```

## Gap 2 — `error_message` key not checked (recall validation + task_ops arg errors)

`recall` returns `{"ok": False, "error_kind": ..., "error_message": "..."}` on invalid args. `task_ops` returns the same shape for pydantic/no-context errors. The existing guard only checks `result.get("error")` (a different key) so these fell through to raw repr.

**Fix**: add `error_message` str fallback immediately after the `error` check:
```python
error_message = result.get("error_message")
if isinstance(error_message, str):
    return _short(error_message, 80)
```

Before: `  ⎿ {'ok': False, 'error_kind': 'missing_required_arg', 'error_me…`
After:  `  ⎿ recall requires ['query']. Available sources are listed …`

## Test plan

- [x] `pytest tests/test_inline_tool_result_summary.py` — 23 passed (added `test_recall_success_shows_chunk_count`, `test_error_message_field_shown_not_raw_dict`)
- [x] `ruff check` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/test_inline_tool_result_summary.py` — 23 OK, 0 errors
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/repl/renderer.py` — OK

**Tier self-check**: 2 Tier-2 contract tests added; no Tier-4 violations.

part of #2491

### DIFF
--- a/src/reyn/interfaces/repl/renderer.py
+++ b/src/reyn/interfaces/repl/renderer.py
@@ -368,6 +368,9 @@ def _summarize_result(tool, result) -> str:
         error = result.get("error")
         if isinstance(error, str):
             return _short(error, 80)
+        error_message = result.get("error_message")
+        if isinstance(error_message, str):
+            return _short(error_message, 80)
         op = result.get("op")
         path = result.get("path")
         status = result.get("status")
@@ -402,6 +405,10 @@ def _summarize_result(tool, result) -> str:
         if isinstance(matches, list):
             n = len(matches)
             return f"{n} match{'es' if n != 1 else ''}"
+        chunks = result.get("chunks")
+        if isinstance(chunks, list):
+            n = len(chunks)
+            return f"{n} chunk{'s' if n != 1 else ''}"
         if status:
             return str(status)
     return _short(result, 80)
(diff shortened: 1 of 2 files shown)


tya5:
**[tui-coder]** — consolidated into #2490 per lead-coder request (avoid serial rebase churn on same function). Both commits rebased onto main[#2488] and cherry-picked onto feat/2489-memory-op-summary.

PR #2543, opened 2026-07-04
refactor(skill-word): T1 — security/permissions skill-tail (delete/rename/reword)

tya5:
**[lead-sonnet]** — T1 final-skill-word sweep: security/permissions batch.

part of #2434

## Decision table

| File | Line(s) | Occurrence | Decision | Evidence |
|------|---------|------------|----------|---------|
| `permissions.py` | 812 | `per-skill path-approval inline` | **REWORD** → `per-actor` | Approval key format is `{actor}/file.read/…`; live code uses `actor` parameter |
| `permissions.py` | 917, 923 | `` ``<skill>/http.get/<host>`` `` | **REWORD** → `<actor>/http.get/<host>` | `_is_host_approved_for` docstring and live `f"{actor}/http.get/{host}"` format confirm `actor` is the key prefix |
| `permissions.py` | 998 | `sandboxed skill's http_get` | **REWORD** → `sandboxed actor's http_get` | Actor is the runtime unit; skill is retired |
| `permissions.py` | 1190 | `symmetric with the SKILL axis` | **KEEP** | Accurate history: SKILL was a CapabilityAxis that was removed; `effective.py:165,199` confirm via `SKILL(removed)` notations |
| `effective.py` | 10 | `the skill's explicit declarations` | **REWORD** → `the actor's` | AgentLayer is keyed on actor identity, not skill |
| `effective.py` | 15 | `allowlists (skills / mcp)` | **REWORD** → `(agents / mcp)` | Profile constrains per-agent allowlists; "skills" as a category no longer exists |
| `effective.py` | 91 | `The skill's declared file paths` | **REWORD** → `The actor's` | Same as line 10 |
| `effective.py` | 146 | `per-skill GRANT` | **REWORD** → `per-actor GRANT` | The grant is keyed to the actor's `decl.mcp` |
| `effective.py` | 148, 165, 199 | `symmetric with SKILL` / `SKILL(removed)` | **KEEP** | All three are accurate-history comments documenting the removal of the SKILL CapabilityAxis |
| `threat_patterns.py` | 11 | `no skill knowledge` | **REWORD** → `no actor/phase knowledge` | The module also has no phase knowledge; "skill" is the stale term |
| `threat_patterns.py` | 14 | `skill-string-free (P7)` | **KEEP** | P7 language — exact wording matches CLAUDE.md P7 rule; changing it would obscure the P7 reference |
| `threat_patterns.py` | 22, 99 | `skill install` (scope description / comment) | **REWORD** → `mcp install` | The strict-scope seam is `mcp_install.py` (verified: `mcp_install.py:206` iterates `"exec", "strict"`) and memory-write (`router_loop.py:3798`); "skill install" was the old concept |
| `threat_patterns.py` | 107 | `reyn's config + skill specs` | **KEEP** | Still accurate: the regex covers `reyn.yaml` AND `skill.md`; skill.md files exist at `tests/fixtures/skills/*/skill.md` and `docs/cookbook/` |
| `threat_patterns.py` | 108 | `skill\.md` regex in `reyn_config_mod` | **KEEP** | Live security guard — skill.md files still exist as write targets (`tests/fixtures/skills/`, `docs/cookbook/`); `required_credentials` is still declared in skill.md frontmatter (confirmed in `docs/concepts/runtime/secret-handling.md`); pattern is scanned at memory-write and mcp-install seams. Security patterns lean keep-if-harmless; this one is still accurate. |
| `store.py` | 150 | `scope for this skill` | **REWORD** → `scope for this actor` | `ScopedSecretStore` is described as "per-agent"; the constructor takes `allowed_keys`, not a skill name |
| `store.py` | 152 | `skill.md frontmatter` | **KEEP** | `skill.md` IS the literal filename where `required_credentials` is declared; this is the accurate user-facing error message |
| `cli/commands/permissions.py` | 71–117 | `skill` variable + comments throughout `_parse_key`/`_cmd_list` | **RENAME** → `actor` | The live approval key format is `{actor}/file.read/{path}` (confirmed in `permissions.py:21–22` module docstring and all `f"{actor}/…"` string formatters); the variable `skill = parts[0]` was the old name for what is now `actor` |

## threat_patterns.py `skill.md` security judgment

**KEEP.** Primary evidence:
1. `find . -name "skill.md" -not -path "*/.claude*"` returns `tests/fixtures/skills/*/skill.md` and `docs/cookbook/translate-doc/translate_doc/skill.md` — skill.md files exist as active write targets in the repo.
2. `docs/concepts/runtime/secret-handling.md` confirms `required_credentials` is declared in `skill.md frontmatter` and this is the live developer-facing contract.
3. The `reyn_config_mod` pattern guards against content injecting instructions like "update skill.md" — still a valid threat even in a transitional state where skills are being removed, because the files still exist and the attack surface remains.
4. Security-lean principle: a pattern that may become dead is cheaper to keep (harmless false-positive risk is zero here since the regex is about attempted modification instructions, not file presence) than to remove prematurely and create a gap.

## Gates

- `ruff check`: all checks passed
- `python scripts/verify_module_docstrings.py`: OK
- `pytest` (security/permissions suite, 434 tests): 434 passed, 9 skipped
- `pytest` full suite: pre-existing failures only (10 failed, all mcp-module-not-found or web-route issues; verified identical baseline on main)
- No test files changed → tier audit not required

## Test plan

- [x] ruff passes on all 5 changed files
- [x] module-docstring gate passes on all 5 changed files
- [x] security + permission + secret + credential test suite: 434 passed, 0 new failures
- [x] full pytest: 0 new failures vs baseline
- [x] `threat_patterns.py skill.md` security judgment documented above — KEEP decision with primary evidence

### DIFF
--- a/src/reyn/interfaces/cli/commands/permissions.py
+++ b/src/reyn/interfaces/cli/commands/permissions.py
@@ -68,22 +68,22 @@ def _save(data: dict[str, bool]) -> None:
 
 
 def _parse_key(key: str) -> tuple[str, str, str] | None:
-    """Split `<skill>/<kind>/<path-or-dir>` into its parts.
+    """Split `<actor>/<kind>/<path-or-dir>` into its parts.
 
     `kind` is `file.read` or `file.write` (contains a dot, takes 2 segments).
-    Returns (skill, kind, path) or None for keys that don't match the file
+    Returns (actor, kind, path) or None for keys that don't match the file
     pattern (e.g. `mcp.<server>` for MCP approvals).
     """
     parts = key.split("/", 3)
-    # Expect at least: skill / file.read|file.write / path
+    # Expect at least: actor / file.read|file.write / path
     if len(parts) < 3:
         return None
-    skill = parts[0]
+    actor = parts[0]
     kind = parts[1]
     path = "/".join(parts[2:])
     if kind not in ("file.read", "file.write"):
         return None
-    return skill, kind, path
+    return actor, kind, path
 
 
 # ── handlers ─────────────────────────────────────────────────────────────────
@@ -97,24 +97,24 @@ def _cmd_list(args: argparse.Namespace) -> None:
         return
     print(f"# {path}")
     print()
-    file_keys: list[tuple[str, str, str, bool]] = []  # skill, kind, path, approved
+    file_keys: list[tuple[str, str, str, bool]] = []  # actor, kind, path, approved
     other_keys: list[tuple[str, bool]] = []
     for key, approved in data.items():
         parsed = _parse_key(key)
         if parsed is None:
             other_keys.append((key, approved))
         else:
-            skill, kind, p = parsed
-            file_keys.append((skill, kind, p, approved))
+            actor, kind, p = parsed
+            file_keys.append((actor, kind, p, approved))
 
     if file_keys:
-        # Group by skill, then by kind, for readability
+        # Group by actor, then by kind, for readability
         file_keys.sort(key=lambda x: (x[0], x[1], x[2]))
-        cur_skill = None
-        for skill, kind, p, approved in file_keys:
-            if skill != cur_skill:
-                cur_skill = skill
-                print(f"  [{skill}]")
+        cur_actor = None
+        for actor, kind, p, approved in file_keys:
+            if actor != cur_actor:
+                cur_actor = actor
+                print(f"  [{actor}]")
             verb = "read " if kind == "file.read" else "write"
             scope = "recursive" if p.endswith("/") else "just_path"
             mark = "✓" if approved else "✗"
--- a/src/reyn/security/permissions/effective.py
+++ b/src/reyn/security/permissions/effective.py
@@ -7,12 +7,12 @@ only narrow.
 
 Four inputs, three ⋂ layers:
 - **agent** (`PermissionDecl` + the default zone): the GRANT layer. Its allow-set
-  is the default zone (layer-0 baseline) ∪ the skill's explicit declarations. The
+  is the default zone (layer-0 baseline) ∪ the actor's explicit declarations. The
   zone is folded in here as the baseline — NOT a separate ∩ restrictor (a
   separate zone restrictor would cancel the decl grants that intentionally extend
   beyond the zone; the byte-identical requirement forces zone-as-baseline).
 - **sandbox** (`SandboxPolicy`): runtime caps (paths / network / subprocess / env).
-- **profile** (`AgentProfile`): agent-level allowlists (skills / mcp).
+- **profile** (`AgentProfile`): agent-level allowlists (agents / mcp).
 
 Per #1199 design call (issuecomment-4620567488): Q2 = per-VALUE membership
 conjunction (no materialized intersected sets — `allows(axis, value) = ∀L:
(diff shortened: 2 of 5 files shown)


tya5:
**[lead-sonnet]** — Revision pushed (force-update `7286101c` → `1f4a6562`). Two review items addressed; both flipped from my original over-conservative KEEP after producer-tracing.

## 1. `store.py:152` — CredentialScopeError message (KEEP → REWORD)

Correct catch — my original message was internally inconsistent after I reworded `:150` to "for this **actor**" while leaving `:152` pointing at "skill.md frontmatter" (deleted).

**New text:**
> `Credential '{key}' is not in the declared scope for this actor. Allowed keys: [...]. To grant access, add '{key}' to this actor's required_credentials declaration.`

**Where I verified the current declaration mechanism:** `grep required_credentials` across `src/reyn/`:
- It is **NOT** a `reyn.yaml` key and **NOT** under the `permissions:` block #2533 reworded (that was a different mechanism). `grep required_credentials src/reyn/config/` → empty.
- `ScopedSecretStore` is constructed **programmatically** via `ScopedSecretStore(allowed_keys=...)`; `allowed_keys` is populated from the actor's `required_credentials` declaration (see the class docstring at `store.py:119`, unchanged: "Constructed from the agent's `required_credentials` declaration"). The old `skill.md` frontmatter source and the `run_skill` boundary that read it are both gone.
- So I pointed the message at the accurate live source — **"this actor's required_credentials declaration"** — rather than inventing "reyn.yaml" (which would name a location this field does not live in). This matches the still-accurate class docstring and stays consistent with the `:150` "actor" reword.

## 2. `threat_patterns.py:108` — `reyn_config_mod` regex `skill\.md` arm (KEEP → DELETE)

e2e's dead-guard verdict confirmed by my own producer-trace:
- `split_frontmatter` (the only frontmatter parser) is called at `file.py:698` (generic `.md` reads) and `memory.py:50` (memory entries) — **never on `skill.md` as permission/behavior config**.
- `grep skill.md src/reyn/**/*.py` (non-test) shows only inert doc-string / error-message mentions; no live config parser consumes skill.md. Live permission config is `reyn.yaml`.
- Guarding writes to skill.md therefore protects a non-threat = dead arm; it also risks false-positive blocks on inert skill.md doc/fixture mentions.

**Change:** dropped `|skill\.md` from the alternation; **kept** the live `reyn\.yaml` / `\.reyn/[^\s]*\.yaml` arms (config-mod guard effectiveness unchanged — those are the live config files). Comment updated: "reyn's config + skill specs" → "reyn's live config files". No test asserts the skill.md pattern match or the `reyn_config_mod` id (grepped all `tests/*threat*`), so no test update needed.

## Gates (re-run on worktree)
- `ruff check`: passed (both files)
- `verify_module_docstrings.py`: OK (both files)
- threat-pattern + credential-scope suites (54 tests): passed
- full `pytest`: 10 failed / 6803 passed — **all 10 pre-existing on baseline** (mcp-module-not-found + web-route mounting; verified identical set with changes stashed). Zero new failures.
- tier-audit: N/A — no test files changed.

## Note on a separate incident
While amending, I discovered this branch had been checked out in a **shared checkout** that a parallel session (T2 / `cleanup/t2-runtime-skill-tail`) was actively using; a stray `--amend` there rewrote T2's local `#2540` commit hash (identical tree — no content change). I restored that checkout to T2's original commit `ca411023` with their working-tree changes intact. All of THIS PR's work now lives correctly in the isolated worktree and pushed cleanly. Flagging so T2 can verify their local branch state.

Issue #2548, opened 2026-07-04
Design: lightweight industry-standard skill feature (stage4)


tya5:
**[lead-coder]** — Design proposal: lightweight industry-standard skill feature (stage4)

Grounded in primary-evidence exploration of the current codebase (5 read-only arch traces + a canonical Claude-Code/Agent-Skills spec pass). Principle per owner: **follow the industry standard (Anthropic Agent Skills / Claude Code) faithfully; diverge only where Reyn's architecture genuinely requires, and flag each delta explicitly.**

---

## 0. Core framing — what a lightweight skill *is* (and what it is NOT)

The industry skill is **a Markdown instruction file (`SKILL.md`) + optional bundled assets, surfaced by name/description and injected into the conversation on trigger.** It is NOT a long-running "phase execution." This is the decisive simplification vs the deleted heavy DSL/phase skill:

- **Activation = appending the rendered `SKILL.md` body to the conversation as a message** (it then persists in history for the rest of the session). It is *already* recovery-covered because conversation history is WAL/snapshot-backed.
- **We deliberately do NOT resurrect `running_skills` / per-skill WAL watermarks / a skill-run registry.** Those existed because old skills were long-running phase executions. The lightweight model has no run lifecycle, so the CLAUDE.md WAL-truncate-falsify recovery gate is **N/A for activation state** (nothing new is WAL-derived). The only new *durable* state is the installed-skill **registry** (see §4), which is a config generation, not WAL-derived.

This keeps the feature small: it is mostly **wiring into seams Reyn already has**, plus a `SKILL.md` loader and the industry format.

---

## 1. The three axes (owner's questions) → progressive-disclosure levels

| Owner's axis | PD level | Reyn seam it plugs into |
|---|---|---|
| (a) manage + surface the skill **list** | L1: name+description always visible | `build_system_prompt` slot + a discovery/registry scan |
| (b) **expand** a skill into the session | L2: full `SKILL.md` body on trigger | append to conversation history (NOT re-inject SP) |
| (c) agent access to **assets** | L3: bundled refs/scripts on demand | `Workspace.read_file` (refs) + `sandboxed_exec` (scripts) |

### (a) Manage + surface — discovery, registry, L1 listing
[...]

Issue #2372, opened 2026-07-01
fix: MCP hot-reload — installing an MCP server doesn't surface its tools until restart (install triggers no refresh + server roster frozen at ctor)


tya5:
**[lead-coder]** — Owner-reported: installing an MCP server from within Reyn does NOT surface its tools until a restart. Primary-verified: MCP hot-reload has TWO independent gaps, both must be fixed.

### Gap 1 — install triggers no refresh
`mcp_install` op (`core/op_runtime/mcp_install.py:517-527`) writes `mcp.servers.<name>` into the scope config file (`reyn.local.yaml` / `reyn.yaml` / `~/.reyn/config.yaml`) + emits `mcp_server_installed`. It **never calls `request_reload` nor `refresh_mcp_servers`** (grep: zero reload/refresh calls in mcp_install.py / tools/mcp_verbs.py). Nothing fires within the turn.

### Gap 2 — the server ROSTER is frozen at ctor
`RouterHostAdapter.self._mcp_servers` (the server roster) is assigned ONCE at ctor (`router_host_adapter.py:301`) and **never reassigned at runtime** (only assignment site). The tools CACHE (`_mcp_tools_cache`) IS refreshable each turn (`session.py:4052` → `maybe_refresh_mcp_tools_from_yaml` / `maybe_reload_mcp_tools_cache_from_disk` / `ensure_mcp_tools_cached`, re-probe on mtime advance). BUT the LLM-facing enumeration `_get_mcp_servers_for_router` → `_enumerate_category("mcp.tool")` iterates `_mcp_servers_flat()` = the frozen roster (`router_host_adapter.py:1449`). A server installed after startup is **not in the roster**, so its tools have no server entry to attach to → never enumerated, regardless of the refreshed tools cache.

### Why restart fixes it
Ctor loads the roster fresh from the full scope cascade (`session.py:1563/1641/2629` → adapter:301); the first-turn probe then enumerates every server incl. the new one.

## The intended-but-unwired mechanism
`refresh_mcp_servers()` (`session.py:3119`) exists and its docstring names EXACTLY this use case (FP-0037 #164: "Skills that install a new MCP server and want it visible within the same chat session"). But it (a) is only wired to the `/reload` seam, not to install, and (b) only refreshes the tools cache — its `_reapply_mcp` seam does NOT rebuild `self._mcp_servers`.

## Fix direction
[...]

PR #2551, opened 2026-07-04
feat(#2548): PR-B skill hot-reload + per-session enable/disable toggle

tya5:
**[lead-sonnet]** — PR-B of the lightweight skill feature (design #2548). part of #2548

## Summary

- **Hot-reload**: adds `config/skills.yaml` to `_HOT_RELOAD_FILES`; registers `_reapply_skills` seam on the HotReloader. The seam re-reads the full config cascade (OUT-set reyn.yaml ∪ IN-set `.reyn/config/skills.yaml`), rebuilds via `build_skill_registry`, and mutates **both** holders the Session owns: `Session._available_skills` (the base registered set) and `RouterHostAdapter._available_skills` (the filtered view after the per-session visibility override). The next turn's `get_available_skills()` reflects the change.
- **Per-session skill disable**: adds `"skill"` as a 4th kind in `_visibility_override`. `set_capability_visible("skill", name, False)` adds the name to the override set and calls `_reapply_skill_visibility`, which re-derives `RouterHostAdapter._available_skills = [s for s in base if s.name not in disabled]`. Restrict-only: disabling an unregistered skill name is a no-op; enabling beyond the registered set is also a no-op.
- **Persistence**: `_persist_visibility_override` already serializes all override keys (now includes `skill`); `load_persisted_toggles` restores the `skill` key and calls `_reapply_skill_visibility`.
- **Validation**: `validate_in_set` in `hot_reload.py` rejects a malformed `skills` section (not a dict / entries not a dict) atomically before any seam runs.
- `capability_visibility_state()` updated to include skill kind in `authorized` and `hidden_by_session`.

## Seam wiring details

`_reapply_skills(in_set)` ignores `in_set` (the skills dict from the reloader) and instead re-reads `load_config()` — this is the correct IN/OUT-set merge: the full cascade always includes the OUT-set (reyn.yaml-declared skills), so OUT-set entries are never dropped. The same pattern is used by `refresh_mcp_servers` for the MCP roster gap fix (#2372).

## Skill-disable filter location

The filter is applied in `_reapply_skill_visibility()`, which sets `RouterHostAdapter._available_skills` to the filtered list. `get_available_skills()` on the adapter returns this filtered value directly. The Session's `_available_skills` field is the base registered set (preserved across toggles so toggle-ON correctly restores without union tricks).

## Test plan

- [x] `test_hotreload_adds_skill_to_live_available_skills` — proves LIVE `get_available_skills()` changes after `_reapply_skills`, not a local var
- [x] `test_hotreload_removes_skill_from_live_available_skills` — proves removal handled
- [x] `test_hotreload_noop_when_skills_unchanged` — proves idempotency
- [x] `test_hotreload_seam_registered` — proves the seam is wired on construction
- [x] `test_toggle_disable_skill_removes_from_live_list` — e2e toggle disable
- [x] `test_toggle_reenable_skill_restores_to_live_list` — e2e toggle re-enable
- [x] `test_toggle_disable_unregistered_skill_is_noop` — restrict-only invariant
- [x] `test_toggle_unknown_kind_raises` — error surface
- [x] `test_capability_visibility_state_includes_skill_kind` — state read model
- [x] `test_skill_toggle_persists_to_visibility_yaml` — persistence write
- [x] `test_skill_toggle_restored_by_load_persisted_toggles` — persistence restore (LIVE host updated)
- [x] `test_skill_toggle_persist_clears_when_reenabled` — persistence cleanup
- [x] `test_validate_in_set_accepts_valid_skills_section` — validation accepts
- [x] `test_validate_in_set_rejects_malformed_skills_section` — validation rejects

## Gate results

- `ruff check src tests` — OK
- `python scripts/test_tier_audit.py --strict tests/test_2548_skill_hotreload_toggle_pr_b.py` — 14/14 OK, 0 errors
- `pytest` — 6833 passed, 12 pre-existing MCP-SDK failures unrelated to this change, 0 new failures
- `python scripts/verify_module_docstrings.py src/reyn/config/loader.py src/reyn/runtime/hot_reload.py src/reyn/runtime/session.py` — OK

### DIFF
--- a/src/reyn/config/loader.py
+++ b/src/reyn/config/loader.py
@@ -368,7 +368,8 @@ def load_config(cwd: Path | None = None) -> ReynConfig:
         # name collision with operator-edited reyn.yaml skill entries.
         # Shape: {"skills": {"entries": {<name>: {<entry>}}}} — same
         # as the skills section in reyn.yaml, handled by _merge skills
-        # branch above. NOT added to _HOT_RELOAD_FILES (per spec).
+        # branch above. #2548 PR-B: this file is also in _HOT_RELOAD_FILES
+        # (the IN-set) so skill declarations hot-reload at the turn boundary.
         dynamic_skills = _load_yaml(project_root / ".reyn" / "config" / "skills.yaml")
         merged = _merge(merged, dynamic_skills)
 
--- a/src/reyn/config/loader.py
+++ b/src/reyn/config/loader.py
@@ -499,6 +499,7 @@ def load_config(cwd: Path | None = None) -> ReynConfig:
 # write-gate boundary (owner-confirmed #2073). Keep this list narrow + explicit.
 _HOT_RELOAD_FILES: tuple[str, ...] = (
     "config/mcp.yaml", "config/cron.yaml", "config/hooks.yaml",
+    "config/skills.yaml",  # #2548 PR-B: skills IN-set hot-reload
 )
 
 
--- a/src/reyn/runtime/hot_reload.py
+++ b/src/reyn/runtime/hot_reload.py
@@ -58,6 +58,13 @@ def validate_in_set(in_set: "dict") -> "str | None":
     mcp = in_set.get("mcp")
     if mcp is not None and not isinstance(mcp, dict):
         return "mcp section must be a mapping"
+    skills = in_set.get("skills")
+    if skills is not None:
+        if not isinstance(skills, dict):
+            return "skills section must be a mapping"
+        entries = skills.get("entries")
+        if entries is not None and not isinstance(entries, dict):
+            return "skills.entries must be a mapping"
     hooks = in_set.get("hooks")
     if hooks is not None:
         # #2073 S2b: validate the runtime hooks shape via the real loader so a
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -810,7 +810,7 @@ class Session:
         # authorized set — visible ⊆ authorized by construction. In-memory (step1 live); step2 will
         # persist it to the per-session config.yaml so resolved_profile_for(sid) re-derives it.
         self._visibility_override: "dict[str, set[str]]" = {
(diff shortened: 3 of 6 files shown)


tya5:
**[lead-sonnet]** — CI pin-fix pushed (commit 823156e8).

Two exhaustive-pin tests broke because PR-B legitimately extends the reapply-seam list and the hot-reload IN-set. Fixed both + swept for siblings:

1. `tests/test_2073_s2_reapply_seams.py` — appended `"skills"` to the registered-seam-name pin (matches `_register_hot_reload_seams` order, after `"hooks"`).
2. `tests/test_2103_C3_spawn_limits_1953.py` — added `"config/skills.yaml"` to the `_HOT_RELOAD_FILES` set pin. The load-bearing intent (reyn.yaml / reyn.local.yaml absent → safety.* not self-raisable) is preserved untouched.
3. `src/reyn/config/loader.py` — fixed the now-stale "NOT added to _HOT_RELOAD_FILES" comment that PR-B contradicted.

**Sibling sweep**: grepped `tests/` + `src/` + `docs/` for `register_seam` / `_seams` / `_HOT_RELOAD_FILES` / seam-count assertions. The only two exhaustive pins on the Session's reapply-seam list and the hot-reload file set are the two above. `test_2073_s1_hot_reloader.py:102` also references `_seams` but builds its own `good`/`bad` seams on a fresh HotReloader (not the Session's list) — unaffected, correctly untouched.

**All four gates green:**
- `ruff check src tests` — passed
- `test_tier_audit --strict` (3 changed test files) — 28/28 OK, 0 errors
- `verify_module_docstrings` (changed src) — OK
- `pytest` (full, repo root) — 6835 passed, 30 skipped. Remaining 10 failed + 10 errors are all pre-existing optional-SDK failures (`ModuleNotFoundError: mcp`, `ModuleNotFoundError: uvicorn`, and web-route-mount tests depending on those unavailable deps) — root cause verified directly, none touch skills/hot-reload/visibility.

Issue #2187, opened 2026-06-26
[lead-coder] Task-system redesign: loose-coupled durable multi-agent task trees (supersedes #2186)


tya5:
**[lead-coder]** — Design spec co-developed with the owner. Canonical record of the task-system redesign (the "decoupling refactor"). This supersedes #2186 (per-session isolation) and structurally closes the dogfood failure ("agents build an org but nothing runs to completion"). Naming is provisional where flagged — owner to confirm.

## 1. Motivation

Dogfood diagnosis (#45): weak and strong models alike can **build** an agent org (`agent_spawn` + `task.create`) but the tasks **never execute** — `task.create` silently accepts an un-executable assignee (omitted → self; an *agent name* → orphan, because the waker only resolves sibling sessions of one agent). The model reads "task created" as "done" and stops. Root: **no assignee validation + two unbridged delegation systems** (intra-agent `task.*` vs cross-agent messaging).

Owner's direction: don't bolt a fix onto the current coupling — **refactor to a loose-coupled task system first**, then features land on a clean foundation.

## 2. Design principles (loose coupling)

- **Task ⊥ agent/session** — a task is first-class; it *references* an owner (a mutable binding), it does not *live inside* a session. Proof of decoupling: a task can exist with **no owner** (UNASSIGNED), and a session runs with **no task**. Bidirectional independence.
- **Multi-agent ⊥ tasks** — agents/sessions are the substrate; tasks reference them; the substrate does not depend on tasks. Shared WAL = shared infrastructure, not coupling.
- **Multi-agent ⊥ A2A protocol** — internal coordination is complete via inter-agent messaging + tasks; the A2A protocol is purely *external* interop.
- **Every behavioral choice is a configurable policy seam, never OS-hardcoded** (P7-aligned). The OS provides mechanism; policies are selectable and may change experimentally.

### 3.1 First-class task trees
- A **task** is a unit of work. Tasks form **trees**: a task's parent is a task; a root task's parent is empty. **Multiple trees** coexist.
- A task owns: its **status**, the **DAG of its subtasks** (dependency ordering among siblings), and a **waker** for its owner-session.
- Two distinct relations (and only two): **ownership tree** (parent → child; failures/aborts propagate) and **dependency DAG** (sibling readiness).
- A task tree **may span multiple agents**.

### 3.2 Ownership
- **Owner = an (agent, session) pair — any agent's any session. Mutable.**
- At child creation the (agent, session) pair is a parameter:
[...]

PR #2579, opened 2026-07-05
feat(pipeline): fold primitive (sequential accumulator)

tya5:
**[lead-sonnet]** — part of #2187

## Summary
- Adds `FoldStep` (Appendix B: `fold = {over?:PATH | items?:[LIT*] init:EXPR do:Step output:NAME max_items?}`) — the second compositional primitive, built directly on the `_run_scope`/dotted-path-recovery/dispatch-table foundation `call` (#2577) established.
- `fold` walks a list SEQUENTIALLY, running `do` once per item and threading its return value as the next `acc`. Registered in `STEP_DISPATCH`/`_STEP_KINDS` (executor), `ENCODERS`/`DECODERS` (serde — `do` recurses through the same step serde as a top-level step), `_STEP_PARSERS` (parser — `fold` removed from `_UNSUPPORTED_STEP_KINDS`), and `ANALYZER_FACETS` (P4 — flags an `over`-sourced fold with no `max_items` as having no statically known iteration bound).

## Design decisions (stated explicitly per the task)
- **Context injection**: `do` sees `{"ctx": <a LOCAL named-store copy, seeded from the fold step's own ctx and grown by do.output across iterations — never written back to the outer scope>, "pipe": <the fold step's OWN incoming pipe-data, held constant across every iteration — the per-iteration analog of Hard rule 5's "callee's first step sees the caller's pipe-data at the call site">, "item": <the k-th list element>, "acc": <the running accumulator>}`. `item`/`acc` are bare top-level context names, so `Path`/`{...}` resolution (`evaluate_expr`/`_interpolate_prompt`) picks them up with zero new machinery.
- **Structural fork from `call`**: `call` recurses into `_run_scope` (a fixed list of distinct sub-steps sharing one evolving local scope). `fold` does NOT reuse `_run_scope` — it loops its own dispatch calls directly (`_run_fold_step`), because every iteration re-invokes the SAME `do` step under a fresh `{item, acc}` binding, a shape `_run_scope`'s "walk this step list once" model doesn't fit. Recorded dotted keys are FLAT (`f"{label}.fold.{k}"`), not a nested scope prefix like `call`'s `f"{label}.call.{j}"`.
- Recovery: an iteration's key already in `completed_step_results` REPLAYS (`acc` read back from the stored result, `do` NOT re-executed) — proven by the mandatory truncate-falsify test with a side-effecting `ToolStep` `do`.

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_pipeline_fold_primitive.py tests/test_pipeline_fold_serde_parser_analyzer.py` — 16/16 OK, 0 Tier-4
- [x] `pytest` (full repo) — 7089 passed; the 10 failed/10 errored are pre-existing mcp/web optional-dep reds, verified identical on `5fbc6ef9` (base) before this change
- [x] New/recovery/invariant tests run in isolation (`pytest tests/test_pipeline_fold_primitive.py tests/test_pipeline_fold_serde_parser_analyzer.py tests/test_pipeline_call_primitive.py tests/test_pipeline_call_serde_parser_analyzer.py`) — 27/27 passed
- [x] All `test_pipeline*.py` files together — 184/184 passed

### DIFF
--- a/src/reyn/core/pipeline/analyzer.py
+++ b/src/reyn/core/pipeline/analyzer.py
@@ -9,15 +9,16 @@ monolithic walker — each primitive registers the check it is responsible for,
 and the registry is the single place a caller (a future ``analyze_pipeline``
 walker, or the IS-4 inline static-analysis gate) folds them together.
 
-This module is the SEAM plus its first two entries (``call``'s and ``match``'s
-facets). It is deliberately minimal in this slice: :data:`ANALYZER_FACETS` maps
+This module is the SEAM plus its first three entries (``call``'s, ``match``'s,
+and ``fold``'s facets). It is deliberately minimal in this slice:
+:data:`ANALYZER_FACETS` maps
 a step dataclass type to a facet function returning a list of human-readable
 problem strings ( empty = the step passed its facet), and :func:`analyze_step`
 looks one up. There is no full pipeline walker wired to consume it yet — that
 arrives with the primitives whose facets have real teeth (cost / spawn-tree
 bounds). What matters now is that the registration point EXISTS and
-``call``/``match`` are registered, so every future primitive MUST add its
-facet here rather than bolting analysis on later.
+``call``/``match``/``fold`` are registered, so every future primitive MUST add
+its facet here rather than bolting analysis on later.
 
 ``call``'s facet is intentionally thin: the target is a STATIC literal pipeline
 name (Hard rule 2), so the only thing to confirm structurally is that it IS a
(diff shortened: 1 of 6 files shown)


tya5:
**[lead-sonnet]** — Rebased onto new main (3ce32ccb, `match` #2578). Resolved the conflicts in all five shared additive structures by keeping BOTH match's and fold's entries (union, never pick-one):

- `Step = Union[..., CallStep, MatchStep, FoldStep]` (executor)
- `STEP_DISPATCH` + `_STEP_KINDS`: `MatchStep`/`FoldStep` both registered
- `ENCODERS`/`DECODERS` (serde): `match`/`fold` both registered
- `_STEP_PARSERS` + `_UNSUPPORTED_STEP_KINDS`: both `match` and `fold` removed from the unsupported set; `_UNSUPPORTED_STEP_KINDS` is now `("for_each", "parallel")`; `_SUPPORTED_STEP_KINDS` gains both
- `ANALYZER_FACETS` (analyzer): `_match_facet`/`_fold_facet` both registered

Docstrings reconciled to enumerate all three compositional primitives (call/match/fold).

4 gates re-run green on the rebased tree:
- ruff `src tests` — clean
- tier audit `--strict` (2 fold test files) — 16/16 OK, 0 Tier-4
- module docstrings (4 changed src files) — OK
- pytest: fold tests in isolation 16/16, recovery-gate alone 1/1, all `-k pipeline` 246 passed / 9 skipped (match + fold + call coexist), full run 7105 passed with the same pre-existing 10-failed/10-errored mcp+web optional-dep reds verified identical on base (unchanged by this PR)

Ready for merge. @lead-coder

Issue #2599, opened 2026-07-05
Web route-mount tests fail in isolation (import-order/global-app-singleton dependence); pass only in full suite


tya5:
**[lead-coder]** — Surfaced while reviewing #2598 (S1 MCP client swap); **pre-existing on main, unrelated to that PR** — filing so it's tracked, not lost.

## Primary evidence
On main (`9f7bedb5`), these 5 tests **fail deterministically when run in isolation** (`pytest <the 5 nodeids>` → `5 failed in 0.74s`), yet the **full-suite CI pytest job passes** (both 3.11 and 3.12 green on #2598):

- `tests/test_fp0001_a2a_endpoints.py::test_fp0001_routes_mounted`
- `tests/web/test_a2a.py::test_a2a_routes_mounted`
- `tests/web/test_mcp_sse.py::test_mcp_routes_mounted`
- `tests/web/test_plugin_loader.py::test_loader_disambiguates_via_package_field`
- `tests/web/test_resources_router.py::test_resources_route_is_mounted_on_app`

Representative failure — `test_resources_route_is_mounted_on_app` asserts `/agents/.../tool-results/` is in `app.routes`, but in isolation only base routes are mounted:
```
routes seen: ['', '/', '/docs', '/docs/oauth2-redirect', '/health',
              '/mcp/messages', '/openapi.json', '/redoc', '/static',
              '/web/designs/{slug}/{file_path:path}']
```

## Hypothesis (NOT yet root-caused to the exact site)
`from reyn.interfaces.web.server import app` returns a **module-level singleton** whose feature routers appear to be registered as an **import side-effect** of *other* test modules. Run the full suite (or a favorable xdist distribution) and some earlier import mounts the routers → assertion passes; run these tests alone (or on an xdist worker that didn't import the router-owning module) → routers absent → fail. This is a **test-isolation / global-mutable-state dependence**, not randomness.

## Why it matters
- These are the "route-mount regression guards" — but their own docstrings note they'd "mask a regression via 404". If they only pass because *another* test imported the router, they aren't actually guarding mount-time registration; a real mount regression could pass CI.
- Local `pytest -n auto` intermittently reports them as failures (worker-distribution-dependent), causing repeated "is my change broken?" confusion for every contributor (already misattributed once as "flaky FastAPI" in #2598's local run).

## Suggested direction (not committed)
[...]

Issue #2597, opened 2026-07-05
MCP client redesign (umbrella): FastMCP swap + per-agent async-events driver-session


tya5:
**[lead-coder]** — Umbrella for the MCP **client** redesign arc. Canonical design contract (broker/memory are ephemeral; this issue is the source of truth).

## Goal
Replace reyn's current direct-`mcp`-SDK MCP client (per-call `open→call→close`) with a **FastMCP-based** client that (a) closes consumption gaps and (b) handles **async server→client events** via a persistent per-agent connection. **Clean-break, no compat/migration/shim** (pre-release).

## Scope
- **Client-first** (server role deferred). Target spec = **stable 2025-11-25** (NOT the 2026-07-28 RC).
- **Wholesale FastMCP replacement** (`fastmcp` v3.4.2; superset of the FastMCP-1.0 frozen in the `mcp` SDK; `mcp.types` stays as shared vocab).
- Consumption scope: pagination → resources → prompts → **elicitation** → OAuth + Streamable HTTP.
- **OUT of scope**: sampling / roots / logging (client-serves-to-server; RC-deprecated). Capability negotiation declares **elicitation only** — compliant servers then won't send sampling/roots.

## Key design decisions (evidence-backed, see traces)
1. **Connection multiplicity = per-agent** (one persistent connection / driver-session per `agent × server`, not one-shared-per-server). Rationale: elicitation-consent + capability envelope + subscription routing are per-agent; eliminates shared-model multiplexing. Spec-compatible (host runs multiple isolated 1:1 clients; permitted-by-omission for same-server-multi-client).
2. **A live receive-loop is required** — stable 2025-11-25 delivers **unsolicited** async notifications (list_changed / resources/updated / progress); a persistent held-open connection is mandatory to receive them.
3. **⚠️ FastMCP provides ~ZERO resilience — reyn owns it all** (source-verified vs v3.4.2): no auto-reconnect (drop→raise), no keepalive (manual `ping()` only), no re-subscribe (no subscribe API; state lost), no Streamable-HTTP resumption, stdio crash = dead session (lazy restart only on re-entry). ⇒ the driver-session must be a full **supervision loop**: catch drop → re-enter/re-init → re-subscribe via raw `client.session` → periodic `ping()` liveness.
[...]

PR #2600, opened 2026-07-05
feat(#2597): S2a — MCPConnectionService (held-open MCP connections for session lifetime)

tya5:
**[per-PR-coder]** — part of #2597

## Summary

Implements S2a of the MCP client redesign: `MCPConnectionService` — a session-owned
service that holds MCP connections **open for the agent session's lifetime**, replacing
the per-call open→close model (a359 P2 `MCPClientPool`) on the live, non-ephemeral chat
MCP path. This is **Option C** from the owner-delegated S2-pre spike: a persistent
connection lives *inside* the agent's own session (not a separate driver-session) —
connections are terminal-less runtime state (reconstruct-on-reload, no WAL), and the
spike proved FastMCP holds its client session in a dedicated `asyncio.Task`, so a
connection opened in one turn's task is safe to `call_tool()` from a later turn's task.

### What changed

- **New `src/reyn/mcp/connection_service.py`** — `MCPConnectionService`:
  - `get(server, config, *, agent_id=None) -> MCPClient` — pool-compatible signature
    (drop-in `pool=` for `MCPGateway`, unchanged). Lazy-connects + caches one `MCPClient`
    per server; returns a stable per-server handle (`_HeldConnection`) whose identity
    survives a reconnect.
  - **Reconnect-on-demand**: a subprocess/HTTP disconnect mid-session does **not** flip
    `MCPClient.is_initialized()` or fastmcp's `Client.is_connected()` — verified
    empirically against the real echo test server's `die` tool (kills the subprocess;
    both stay `True` on every subsequent call). The only observable signal is an
    `MCPError` raised on next use, so the held-connection handle catches it, discards +
    reopens the connection **once**, and retries the same call. A second failure
    propagates unchanged (no silent retry loop) — same fault-containment guarantee the
    caller's `MCPGateway` boundary already provides.
  - `aclose()` — closes every held connection; idempotent.
  - `held_servers()` — public read-only introspection (no private-state test assertions
    needed).
- **`src/reyn/core/op_runtime/context.py`** — new `OpContext.mcp_connection_service`
  field (default `None`).
- **`src/reyn/core/op_runtime/mcp.py`** — the `mcp` op handler prefers
  `ctx.mcp_connection_service` when wired, else falls back to `ctx.mcp_pool` unchanged.
- **`src/reyn/mcp/gateway.py`** — `MCPGateway.__init__`'s `pool` type-hint widened to
  `MCPClientPool | MCPConnectionService | None` (TYPE_CHECKING-only import; no behavior
  change — the gateway only ever calls `.get()` on whatever it's given).
- **`src/reyn/runtime/session.py`**:
  - `Session.__init__` constructs one `MCPConnectionService` per session (cheap — empty
    dict until first use).
  - `_mcp_call_tool` / `_mcp_list_tools`: a **non-ephemeral** session routes through the
    held service; an **ephemeral** session (`self._ephemeral`, set post-construction by
    the registry on an ephemeral spawn) keeps the pre-existing one-shot `MCPClientPool`
    path unchanged — holding a connection open for a sub-second-lived session is pure
    churn (F4 decision).
  - New `aclose_mcp_connections()` + `mcp_held_servers()` public methods (teardown +
    introspection).
- **`src/reyn/runtime/registry.py`**:
  - `remove_session` now awaits `session.aclose_mcp_connections()` (getattr-guarded,
    alongside the existing `cancel_inflight`/`await_quiescent` quiesce) before dropping a
    spawned session — no leaked subprocess/HTTP connection on drop/rewind.
  - `archive_agent` (the async action-layer seam wrapping the sync `remove()`) now
    closes held MCP connections for every loaded session of the agent (main + any
    spawned sids) before `remove()` drops them from the in-memory map — `remove()` has
    no async seam of its own, so this is the teardown seam for the **main** session.
- **Existing test updates** (both required by the design change, not incidental):
  - `tests/test_2421_mcp_seam_completeness.py` — added `mcp/connection_service.py` to
    the allowed-direct-`MCPClient()`-construction list (it's now a legitimate seam
    alongside `pool.py`).
  - `tests/test_mcp_list_tools_finally_close.py` — this test's close-on-every-exit-path
    contract is specifically the **one-shot** pool's contract; a persistent session no
    longer closes on every call (that's the whole point of S2a). Updated the test
    session to be `_ephemeral = True` so it still exercises exactly the one-shot path it
    was written for; docstring updated to say so explicitly.
  - `tests/_support/mcp_fastmcp_echo_server.py` — added a `pid()` tool (returns
    `os.getpid()`) so connection-reuse/reconnect tests can prove "same subprocess" /
    "new subprocess" against a REAL server round-trip instead of only comparing Python
    object identity.

### Out of scope (later slices, untouched)

Notifications / receive-loop / `message_handler` bridge / `resources/subscribe` /
resync-probe / periodic ping (S2b). Server→client requests (S3). OAuth (S6). turn-cancel
`notifications/cancelled` (S2c). No WAL state added (connections are runtime-only per
the design) — the recovery-feature PR gate does not apply here.

## Test plan

- [x] **Connection reuse (core value)**:
  `test_second_get_returns_same_held_client_no_rehandshake` — 2nd `get()` for the same
  server returns the SAME handle, and two `call_tool("pid", …)` round-trips return the
  SAME real subprocess PID (no re-handshake).
- [x] **Reconnect-on-demand**: `test_dead_connection_transparently_reconnects` — the
  `die` tool genuinely kills the subprocess; the held handle transparently reconnects
  (new PID) and the next call succeeds instead of failing forever.
- [x] **Teardown**: `test_aclose_closes_all_held_clients` (service-level, incl.
  idempotency + fresh reconnect after `aclose()`); `test_remove_session_teardown_closes_held_connections`
  (registry-level, real `AgentRegistry.remove_session`).
- [x] **Ephemeral isolation**: `test_ephemeral_session_never_populates_connection_service`
  proves an ephemeral session's MCP call never touches the held-connection service;
  `test_non_ephemeral_session_holds_connection_across_calls` proves the counterpart.
- [x] **Parity**: `test_call_result_parity_with_one_shot_pool` — `MCPGateway` with the
  connection service vs. the one-shot pool returns byte-identical results for the same
  call.
- [x] All 7 new tests + the 2 updated existing tests pass locally.

### 4 CI gates (local, all green)

```
$ ruff check src tests
All checks passed!

$ python scripts/test_tier_audit.py --strict tests/test_2597_s2a_mcp_connection_service.py \
    tests/test_2421_mcp_seam_completeness.py tests/test_mcp_list_tools_finally_close.py
Summary: 10 tests inspected
  ✓ OK: 10
  ✗ errors: 0 (Tier 4 violations)
Exit code: 0

$ python -m pytest -q
5 failed, 7265 passed, 21 skipped, 11 warnings in 123.89s
```

The 5 failures are the **pre-existing** known-failing web-route-mount tests (issue
#2599, unrelated — confirmed exact match): `test_fp0001_routes_mounted`,
`test_a2a_routes_mounted`, `test_mcp_routes_mounted`,
`test_loader_disambiguates_via_package_field`, `test_resources_route_is_mounted_on_app`.
**Zero new failures** from this change.

```
$ python scripts/verify_module_docstrings.py src/reyn/core/op_runtime/context.py \
    src/reyn/core/op_runtime/mcp.py src/reyn/mcp/gateway.py src/reyn/runtime/registry.py \
    src/reyn/runtime/session.py src/reyn/mcp/connection_service.py
OK: no module-docstring refactor narrative in <files>.
```

### DIFF
--- a/src/reyn/core/op_runtime/context.py
+++ b/src/reyn/core/op_runtime/context.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
     from reyn.data.workspace.media_store import MediaStore
     from reyn.data.workspace.workspace import Workspace
     from reyn.llm.model_resolver import ModelResolver
+    from reyn.mcp.connection_service import MCPConnectionService
     from reyn.mcp.pool import MCPClientPool
     from reyn.security.permissions.permissions import PermissionDecl, PermissionResolver
     from reyn.security.sandbox import SandboxBackend
@@ -61,6 +62,13 @@ class OpContext:
     # in a possibly-different task → the cross-SDK-task cancel-scope crash). None outside an MCP
     # context (non-MCP ops never invoke the mcp handler).
     mcp_pool: "MCPClientPool | None" = None
+    # #2597 S2a: the session-owned held-open connection service (Option C — one
+    # persistent MCPClient per server, reused for the session's lifetime). When
+    # set, the mcp op handler prefers THIS over ``mcp_pool`` (pool-compatible
+    # ``get()`` — see connection_service.py). None for the ephemeral-session /
+    # one-shot-probe path, which intentionally keeps the per-call ``mcp_pool``
+    # (held connections would just churn for a sub-second-lived session).
+    mcp_connection_service: "MCPConnectionService | None" = None
     # FP-0016 Component E: agent identity for X-Reyn-Agent-Id header on
     # outgoing MCP / external HTTP calls. Plumbed from Session's
     # ReynConfig.agent.id (= `reyn/<hostname>` by default). None
--- a/src/reyn/core/op_runtime/mcp.py
+++ b/src/reyn/core/op_runtime/mcp.py
@@ -40,10 +40,13 @@ async def _execute(op: MCPIROp, ctx: OpContext) -> dict:
         if expanded.get("url"):
             expanded = {**expanded, "type": "http"}
 
-    # #a359 P2: the client comes from the per-turn structured pool (opened + closed in the pool's
-    # owning task — no cross-SDK-task teardown). None = no pool wired on this ctx (a non-MCP
-    # OpContext should never reach the mcp handler; guard defensively).
-    if ctx.mcp_pool is None:
+    # #2597 S2a: prefer the session-owned held-open connection service (Option C) when wired —
+    # it replaces the per-turn pool on the live (non-ephemeral) session path with one persistent
+    # connection per server, reused across turns/tasks. Falls back to the #a359 P2 per-turn
+    # structured pool (opened + closed in the pool's owning task) for the ephemeral-session /
+    # one-shot path. Neither wired = no MCP context wired on this ctx (a non-MCP OpContext should
+    # never reach the mcp handler; guard defensively).
+    if ctx.mcp_connection_service is None and ctx.mcp_pool is None:
         return {"kind": "mcp", "status": "error", "server": op.server,
                 "tool": op.tool, "error": "no MCP client pool on this context"}
 
@@ -80,7 +83,7 @@ async def _execute(op: MCPIROp, ctx: OpContext) -> dict:
     # connect, a bad config, a malformed response, or a transport group all surface here as a clean
     # MCPFault → contained error tool-result (owner req: MCP misbehavior must not crash the router
     # loop). Cancellation is never swallowed (is_real_control_flow re-raises genuine cancel/KI/SE).
-    gateway = MCPGateway(pool=ctx.mcp_pool, agent_id=ctx.agent_id)
+    gateway = MCPGateway(pool=ctx.mcp_connection_service or ctx.mcp_pool, agent_id=ctx.agent_id)
     try:
         result = await gateway.call_tool(
             op.server, op.tool, op.args, expanded, progress_cb=_on_progress,
--- a/src/reyn/mcp/connection_service.py
+++ b/src/reyn/mcp/connection_service.py
@@ -25,13 +25,29 @@ loop / ping): a subprocess death or HTTP disconnect mid-session does NOT flip
 ``MCPClient.is_initialized()`` or the underlying ``fastmcp.Client.is_connected()``
 (verified empirically against the real echo test server's ``die`` tool — both stay
 True after the transport is gone) — the only observable signal is an :class:`MCPError`
-raised on the NEXT use. So the held-connection handle catches an ``MCPError`` from
-``call_tool``/``list_tools``, discards + reopens the connection ONCE, and retries the
-SAME call on the fresh connection. If the retry also fails, the fault propagates
-unchanged (the caller's existing MCPGateway fault boundary contains it into an
-LLM-visible error result, same as the pre-S2a per-call path) — a dropped connection
-must not permanently wedge the server for the rest of the session, but this is not a
-silent-retry-forever loop.
+raised on the NEXT use. So the held-connection handle catches an ``MCPError``, discards
++ reopens the dead connection so the NEXT call lands on a healthy transport — a dropped
+connection must not permanently wedge the server for the rest of the session.
+
+CRITICAL — the reconnect must NOT silently retry a side-effectful call (at-most-once):
+post-S1 ``call_tool`` raises ``MCPError`` on ANY transport failure, including the
+drop-AFTER-execution window (the server RAN the tool, then the connection dropped before
+its response arrived). Auto-retrying the same call on the fresh connection would
+RE-EXECUTE the tool — an at-most-once → at-least-once regression vs the pre-S2a per-call
+pool (a duplicated ``create_issue`` / ``send_email`` / counter increment). So the two op
+classes are healed differently:
+
+  - :meth:`_HeldConnection.call_tool` — **reconnect-then-propagate**: on ``MCPError``,
+    heal the connection (reopen) but RE-RAISE the original error. The call is NOT
+    retried, so a tool is executed at most once. The first ``call_tool`` right after an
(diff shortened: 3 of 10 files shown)


tya5:
**[lead-coder]** — Review of #2600. Design is clean (duck-typed `_HeldConnection` with stable identity across reconnect, per-server open-lock, pool-compatible `get()` keeping MCPGateway's contract — the gateway type-hint widening is correct and minimal). One **blocking correctness finding** before merge; everything else is good.

## 🔴 BLOCKING — `call_tool` auto-retry silently double-executes side-effectful MCP tools

`_HeldConnection.call_tool` → `_with_reconnect` catches `MCPError` and **retries the same call** on a fresh connection (`connection_service.py:217-231`). Post-S1, `MCPClient.call_tool` raises `MCPError` **only on transport/protocol failure** (`client.py:237-238`) — tool-level `isError` comes back as a result dict, not raised. So the retry fires on a genuine transport drop.

Two sub-cases:
- **Idle-drop (common, safe):** connection died while idle; the next `call_tool` fails at *send* (server never received it) → retry executes exactly once. ✅
- **Drop-after-execution (rare, unsafe):** server received the request, executed the tool (side effect committed), then the connection dropped before the response arrived → `call_tool_mcp` raises → reconnect → **retry re-executes the tool**. For a non-idempotent tool (create/send/write/pay) this is a **silent double-execution**.

`test_dead_connection_transparently_reconnects` actually demonstrates this: `call_tool("die")` kills the subprocess, then the reconnect retries `die` and kills the *fresh* subprocess too before propagating — `die` runs twice.

This is a **behavior regression**: the pre-S2a per-call pool never retried (a mid-call drop just propagated) = at-most-once. S2a's retry makes it at-least-once for side-effectful calls. MCP has no exactly-once guarantee and real MCP clients deliberately don't auto-retry tool calls for exactly this reason.

### Required change
- **`call_tool`: reconnect-and-propagate, do NOT auto-retry.** On `MCPError`, discard + reopen the held connection (so the *next* call is healthy) but re-raise the original error — never silently re-run a call that may have already executed server-side. Preserves at-most-once.
- **`list_tools`: auto-retry is fine** (idempotent read) — keep it.
- The only cost is that the first `call_tool` immediately after an *idle* drop fails once, then subsequent calls work. That's acceptable for S2a and is fully resolved by **S2b's proactive health/ping loop** (detects the drop before the next call), so we're not trading correctness for a UX nicety — S2b delivers the transparent-heal safely.
- Update `test_dead_connection_transparently_reconnects` to the new contract: `call_tool("die")` → raises once + heals the connection; the *next* `echo` succeeds on the reconnected client (asserts fresh PID). Add an explicit assertion that a side-effectful call is NOT re-executed (e.g. a counter tool called across a mid-call drop increments once, not twice).

## ✅ Otherwise good
- Routing (OpContext.mcp_connection_service, op_runtime prefers it, ephemeral→one-shot pool), teardown (registry.remove_session/archive_agent → aclose), parity test, ephemeral-isolation test — all present and sound.
- Runtime-only (no WAL) — correct per Q4; no truncate-falsify gate needed.
- Reconnect-on-demand keyed on the empirical is_initialized-doesn't-flip finding — good catch, well-documented.

Re-request review once `call_tool` is at-most-once. CI (currently pending) must also land green.

tya5:
**[per-PR-coder]** — Revision pushed (7a2a1b35), ready for re-review.

## Blocking fix addressed: call_tool at-most-once

You were right — `_HeldConnection.call_tool`'s auto-retry double-executed side-effectful
tools in the drop-after-execution window. Fixed exactly per your required shape (`heal_only`
flag; two behaviors from one helper):

- **`call_tool` → `heal_only=True` (reconnect-then-propagate):** on `MCPError`, reopen the
  dead connection so the NEXT call lands on a healthy transport, but **re-raise the original
  error** — the call is never re-run. At-most-once preserved.
- **`list_tools` → `heal_only=False` (retry-once):** idempotent read, still heals
  transparently on the fresh connection. Unchanged.

Module docstring rewritten to spell out the at-most-once rationale + the drop-after-execution
window + why S2b's ping loop (not an S2a retry) is the safe way to get transparent healing.

## Tests

- `test_dead_connection_transparently_reconnects` → renamed
  `test_dead_connection_heals_next_call_after_transport_error`: now asserts the failing
  `call_tool("die")` **raises once** and the connection **heals** so the NEXT call succeeds on
  a fresh subprocess (new PID). No more double-`die`.
- **NEW** `test_call_tool_executed_at_most_once_across_mid_call_drop`: a **file-backed**
  side effect (`bump_then_die` — append + `fsync` to disk, THEN `os._exit`) so the execution
  count **survives the subprocess death** (an in-memory counter would reset on the fresh
  reconnected subprocess and hide a double-count). Asserts exactly ONE append across the
  mid-call drop — the regression the old auto-retry caused (once per subprocess = two) is now
  RED-provable. (Server subprocess is Seatbelt-sandboxed to cwd, so the test points `cwd` at
  `tmp_path` to make the marker file writable.)

## 4 gates (local, all green) + CI

```
ruff check src tests          → All checks passed!
test_tier_audit --strict      → 8/8 OK, 0 Tier-4 violations
pytest (repo root)            → 5 failed, 7266 passed, 21 skipped   (the 5 = pre-existing #2599 web-route-mount, zero new)
verify_module_docstrings      → OK
```

CI on the previous push (44f887ef) already landed **fully green** (ruff / pytest 3.11 / pytest
3.12 / tier-audit / docs all pass). Re-run for 7a2a1b35 will kick automatically — will confirm
green.

Everything else from your review (routing, teardown, parity, ephemeral isolation, gateway
type-widening, runtime-only/no-WAL) is unchanged. Ready to merge once CI re-confirms green.

PR #2604, opened 2026-07-05
docs(hooks): remove dead skill_start/skill_end lifecycle points

tya5:
**[docs-maintainer]** — Drift correction requested by lead-coder (surfaced during a hooks-architecture investigation).

## Summary
`feature-map.md:423` claimed hooks fire at "8 points" including `skill_start`/`skill_end`. Verified against `src/reyn/hooks/schema.py`: `ALLOWED_HOOK_POINTS` ships exactly 6 (`turn_start/end`, `session_start/end`, `task_start/end`) — the module docstring confirms `skill_start`/`skill_end` were removed per #1800 and never dispatched, and a repo-wide grep confirms zero non-test source references anywhere else.

Beyond the one line lead-coder pinned, the same stale claim appeared in 4 more places (grepped for `skill_start`/`skill_end` across all docs):
- `concepts/runtime/hooks.md`/`.ja.md` — the "8 points" intro, the lifecycle table's `skill` row, an "implementation anchors" bullet, and a whole "Deferred: skill_end on interrupt/error" bullet (now moot since the point doesn't exist at all).
- `concepts/runtime/config-hot-reload.md`/`.ja.md` — the `hooks_add` `on` parameter's enum list.
- `reference/config/reyn-yaml.md` — the `hooks:` example's inline comment and the `on` key's parameter-table enum (both listed the 8-value set).

## Second flag (investigated, no fix needed)
Lead-coder also soft-flagged a possible aspirational "config filesystem/mtime watcher" claim, without a pinned line. Traced it to `reference/cli/mcp.md:42`'s "active chat sessions also watch `reyn.yaml` mtime... on each turn boundary" — verified accurate against `RouterHostAdapter.maybe_refresh_mcp_tools_from_yaml` (`src/reyn/runtime/services/router_host_adapter.py:1639`), a real per-turn-boundary mtime-poll-and-reload mechanism (not an OS-level fs-event watcher, but genuinely functional, matching `test_mcp_yaml_mtime_watch`). No drift there — left unchanged.

## Test plan
- [x] `mkdocs build --strict` — exit 0, no new broken links (pre-existing unrelated INFO-level gaps only, verified against a prior clean baseline).
- [x] `ruff check src tests` — all checks passed (no src changes).
- [x] Manual: `grep -rln "skill_start\|skill_end"` across all of `docs/` post-fix returns only unrelated `run_skill_started`/`run_skill_completed` P6 event names (a different mechanism — skill *execution* events, not hook lifecycle points).


tya5:
**[docs-maintainer]** — Follow-up fix pushed: a repo-wide grep for `8 lifecycle` after the initial commit caught 2 more stale references in `config-hot-reload.md`/`.ja.md`'s "See also" section ("the 8 lifecycle points" → 6). Repo-wide grep for both `skill_start`/`skill_end` and `8 lifecycle` now returns zero remaining hits. mkdocs --strict still clean.

Issue #2608, opened 2026-07-05
Unified external-event → hooks: trigger hooks on MCP/cron/webhook/fs events; add pipeline-launch action


tya5:
**[lead-coder]** — Design issue for the **unified external-event → hooks** arc (owner-surfaced 2026-07-05). Canonical contract; investigation is main-verified with file:line anchors.

## Goal
Let an **external event** (MCP `resources/updated`, cron fire, webhook, filesystem change) trigger a **hook** whose action runs a shell command, posts a message to a session, or **launches a pipeline** — agent-driven extensibility by composing existing primitives. Owner's concrete example: `resources/updated` → hook → read the resource → post a message to a specific session.

## Key finding: reyn already owns ~80% of this
The hooks feature is **fully shipped on main** (`src/reyn/hooks/`, dispatcher wired into `Session`). The unbuilt piece is **only the external-event → hook trigger**, not a new action system.

### What EXISTS today (the baseline — do NOT reinvent)
- **6 fixed INTERNAL lifecycle hook-points** (`hooks/schema.py:23` `ALLOWED_HOOK_POINTS`): `turn_start/end`, `session_start/end`, `task_start/end`. Dispatched only from hard-coded lifecycle call-sites (`session.py:3979/4166/5673`, `op_runtime/task.py`).
- **3 actions** (`HookDef`, `hooks/schema.py:89`; `_dispatch_one`, `dispatcher.py:141`): `template_push` (Jinja2 → post message to a session inbox, incl. **cross-session** #2072, `wake:true`=new turn / `wake:false`=passive), `shell_exec` (side-effect shell, sandbox+consent-gated), `shell_push` (shell stdout=JSON push directive). ⇒ **2 of the 3 target actions (shell + post-session) already exist.**
- Config: YAML `hooks:` block (global `.reyn/hooks.yaml` + per-agent), **hot-reloadable** (`session.py:3439`), per-session toggle (#2285), LLM-writable (`hooks_add`).
- **DI seam** (`dispatcher.py:53-67` injected callables) = the proven extension point (`run_shell`, `cross_session_put` were added this way).
- `HookDef.matcher` (`schema.py:118`, reserved/uninterpreted) = the pre-reserved seam for external-event source/URI matching.

### The gap (the actual work)
[...]

PR #2624, opened 2026-07-05
test+fix(#2608): hook run-loop-pickup coverage + fs_watch path normalization

### LINKED ISSUE #2623: fs_watch matcher footgun on macOS: /tmp symlink → events report /private/tmp, matcher:{path:'/tmp/**'} never matches
**[lead-coder]** — Found in tui-coder's live dogfood of #2608 H4 (fs-watcher).

On macOS, `/tmp` is a symlink to `/private/tmp`. The FsWatcher (watchdog) reports the RESOLVED path (`/private/tmp/...`) in the fired `file_changed` event's `path`. So an operator who writes `fs_watch.paths: ['/tmp/myproj']` + a hook `matcher: {path: '/tmp/myproj/**'}` (matching the config as written) will NEVER match — because the event's path is `/private/tmp/myproj/...`. Confirmed via a standalone FsWatcher test (fires correctly, reports the resolved path).

Genuine operator footgun (config-authoring gotcha, not a runtime defect). Fix options: (a) path-normalization — resolve symlinks consistently on BOTH the watched-path registration AND the event path so they match; or (b) a docs callout in reyn-yaml § fs_watch warning that event paths are symlink-resolved (matcher should use the resolved path). Recommend (a) resolve both to realpath at registration so the operator's configured path just works. roi:low.

tya5:
**[lead-coder]** (shipping [per-PR-coder]'s completed implementation — agent stalled at a backgrounded pytest before committing; committed+pushed by lead) — `part of #2608`, `fixes #2623`.

## What this adds
1. **Run-loop-pickup coverage** — a Tier-2 test that starts `session.run()` and fires an external-event hook from a background task, asserting a turn ACTUALLY RUNS (`turn_started{kind:hook}` + the push processed), not just that it lands in the inbox. Closes the coverage hole the live dogfood exposed (the impl was confirmed correct by reproduction; this is the missing assertion).
2. **#2623 fs_watch path normalization** — watched paths are realpath-resolved so an operator's `matcher: {path: ...}` matches despite symlinks (macOS /tmp→/private/tmp).

## Gate status
Deferred pytest to CI (agent stalled). Lead verifies CI green + reviews before merge.

## Test plan
- [ ] pytest — CI (lead verifies)
- [x] New `tests/test_2608_run_loop_pickup.py` (real Session + real run() loop) + fs_watch symlink test

### DIFF
--- a/src/reyn/runtime/fs_watcher.py
+++ b/src/reyn/runtime/fs_watcher.py
@@ -22,6 +22,24 @@ docs. That scheduled callable does the actual (loop-thread-only) bounded
 bound+drop+log discipline (``_QUEUE_MAXSIZE``, overflow drops the newest event
 + logs, never grows unboundedly, never blocks the watchdog thread).
 
+Path normalization (#2623): on macOS ``/tmp`` is a symlink to ``/private/tmp``
+(same class of footgun exists anywhere an operator's configured
+``fs_watch.paths`` entry traverses a symlink). ``watchdog``'s OS backend
+(fsevents on macOS) reports the RESOLVED path in every fired event regardless
+of which path string ``Observer.schedule`` was given — so a naive matcher
+written against the operator's CONFIGURED path (e.g. ``matcher: {path:
+'/tmp/x/**'}``) silently never matches an event path of
+``/private/tmp/x/...``. Fixed at registration (:meth:`FsWatcher.start`): for
+each configured path whose ``os.path.realpath`` differs from its own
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — Closing as a duplicate of #2625. I opened this from the coder's uncommitted worktree when its agent stalled at a backgrounded pytest; on resume the coder committed + opened #2625 (same content, proper branch/body/4-gate evidence). Keeping #2625.

PR #2625, opened 2026-07-05
test+fix(#2608): run-loop-pickup coverage for external-event hooks + fs_watch path normalization (#2623)

### LINKED ISSUE #2623: fs_watch matcher footgun on macOS: /tmp symlink → events report /private/tmp, matcher:{path:'/tmp/**'} never matches

tya5:
**[per-PR-coder]** — part of #2608, fixes #2623

Two related hardening items from a live-dogfood + investigation of the external-event→hooks arc.

## Item 1 — Close the run-loop-pickup coverage gap (external-event hooks)

The investigation confirmed BY REPRODUCTION that a background-fired external-event hook push (`wake=True`) wakes an idle `session.run()` loop and runs a hook-attributed turn. But the existing H1/H4 unit tests (`test_2608_h1_mcp_resource_updated_hook.py`, `test_2608_h4_fs_watcher.py`) call the producer/dispatch path directly and only assert the templated push **lands in `session.inbox`** — they never start `session.run()`, so they can't observe the run-loop actually picking the push up and executing a turn. That's the "unit-green ≠ live-works" hole.

New Tier-2 test `tests/test_2608_run_loop_pickup.py::test_dispatch_external_event_wakes_idle_run_loop_and_runs_hook_turn`:
- Builds a **real** `Session` with an `mcp_resource_updated` `template_push` hook (`wake=True`) via the real `load_hooks` seam.
- Starts `session.run()` as a background task, lets it go idle (`inbox.empty()`, 0 hook `turn_started` events).
- From a **separate** background task, calls `session.dispatch_external_event(...)` — the exact public entry point FsWatcher's drain / the MCP bridge / cron+webhook ingress all use.
- Asserts a turn **actually runs** beyond inbox-landing: a `turn_started` with `kind == "hook"` is emitted **and** the turn completes (`turn_settled`), via the real router loop; then `inbox.empty()` confirms the push was consumed as the turn trigger, not left sitting.
- Only the LLM boundary is faked (`reyn.runtime.router_loop.call_llm_tools` → a real async stub, the established `test_1800_wake_drain.py` idiom). The run-loop, dispatch, and hook machinery all run for real. No private-state assertions — observations flow through the public `EventLog` / `inbox` surface.

## Item 2 — Fix the macOS symlink matcher footgun (#2623)

On macOS `/tmp` is a symlink to `/private/tmp`. watchdog's OS backend reports the **resolved** path in fired `file_changed` events regardless of the path string given to `Observer.schedule`, so an operator who configures `fs_watch.paths: ['/tmp/x']` + a hook `matcher: {path: '/tmp/x/**'}` never matches (event path is `/private/tmp/x/...`).

Fix in `src/reyn/runtime/fs_watcher.py` (path normalization at registration):
- In `FsWatcher.start()`, build a resolved→configured rewrite table: for each configured path whose `os.path.realpath` differs from its normalized form, record `(resolved, configured)`.
- New `_rewrite_path()` maps a fired event's (possibly symlink-resolved) path back onto the operator's configured prefix; applied in `_on_fs_event` **before** the event ever reaches `hook_trigger`.
- Result: the `file_changed` event's `path` template var is under the exact path the operator wrote in `fs_watch.paths`, so `matcher: {path: <configured>}` Just Works. A path with no symlink component is a no-op (`resolved == configured`, no rewrite entry).
- The rewrite table is built once on the loop thread in `start()` before the observer thread starts; the watchdog-thread lookup is read-only → race-free.

New Tier-2 test `test_symlinked_watch_path_reports_events_under_the_configured_prefix`: a watched path given via a **real symlink** resolves so a file change fires with a path that a naive `matcher: {path: '<configured>/**'}` glob matches. All 13 existing real-Observer H4 tests still pass (14 total now).

## Test plan

- [x] `ruff check src tests` → All checks passed
- [x] `python scripts/test_tier_audit.py --strict tests/test_2608_run_loop_pickup.py tests/test_2608_h4_fs_watcher.py` → 15 inspected, 0 warnings, 0 errors
- [x] `python scripts/verify_module_docstrings.py src/reyn/runtime/fs_watcher.py` → OK (no refactor narrative)
- [x] `python -m pytest -q` (foreground) → **5 failed, 7439 passed, 21 skipped** — the 5 failures are the pre-existing #2599 known-unrelated web/route-mounting tests (`test_fp0001_a2a_endpoints`, `web/test_a2a`, `web/test_mcp_sse`, `web/test_plugin_loader`, `web/test_resources_router`); **zero NEW**. Both new/modified test files pass.

## 4-gate output

```
$ ruff check src tests
All checks passed!

$ python scripts/test_tier_audit.py --strict tests/test_2608_run_loop_pickup.py tests/test_2608_h4_fs_watcher.py
Summary: 15 tests inspected
  ✓ OK: 15
  ⚠️ warnings: 0 (bounded-life candidates)
  ✗ errors: 0 (Tier 4 violations)
Exit code: 0

$ python scripts/verify_module_docstrings.py src/reyn/runtime/fs_watcher.py
OK: no module-docstring refactor narrative in src/reyn/runtime/fs_watcher.py.

$ python -m pytest -q
5 failed, 7439 passed, 21 skipped, 11 warnings in 164.45s
  (5 failed = pre-existing #2599 web/route-mounting; zero NEW)
```

### DIFF
--- a/src/reyn/runtime/fs_watcher.py
+++ b/src/reyn/runtime/fs_watcher.py
@@ -22,6 +22,24 @@ docs. That scheduled callable does the actual (loop-thread-only) bounded
 bound+drop+log discipline (``_QUEUE_MAXSIZE``, overflow drops the newest event
 + logs, never grows unboundedly, never blocks the watchdog thread).
 
+Path normalization (#2623): on macOS ``/tmp`` is a symlink to ``/private/tmp``
+(same class of footgun exists anywhere an operator's configured
+``fs_watch.paths`` entry traverses a symlink). ``watchdog``'s OS backend
+(fsevents on macOS) reports the RESOLVED path in every fired event regardless
+of which path string ``Observer.schedule`` was given — so a naive matcher
+written against the operator's CONFIGURED path (e.g. ``matcher: {path:
+'/tmp/x/**'}``) silently never matches an event path of
+``/private/tmp/x/...``. Fixed at registration (:meth:`FsWatcher.start`): for
+each configured path whose ``os.path.realpath`` differs from its own
+(normalized) form, a resolved-path -> configured-path REWRITE is recorded
+(:attr:`_path_rewrites`). Every fired event's path is rewritten back onto the
+operator's configured prefix (:meth:`_rewrite_path`) before it ever reaches
+``hook_trigger`` — so ``matcher: {path: <the path the operator wrote in
+fs_watch.paths>}`` Just Works, and the ``file_changed`` event's ``path``
+template var is exactly the configured path (not a resolved alias of it) for
+any file under it. A path with no symlink component is unaffected
+(rewrite is a no-op — ``resolved == configured`` and no entry is added).
+
 Debounce (F7-3): editors emit event BURSTS for one logical change (temp files,
 multiple writes, create-then-modify). Coalesced per-path on the watchdog
 thread with a simple leading-edge scheme: :meth:`_FsEventHandler._maybe_fire`
@@ -61,6 +79,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import time
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
@@ -126,6 +145,9 @@ class FsWatcher:
         self._queue: "asyncio.Queue[tuple[str, str]] | None" = None
         self._drain_task: "asyncio.Task | None" = None
         self._started = False
+        # #2623: resolved-symlink-path -> operator-configured-path rewrites,
+        # built in :meth:`start` — see module docstring "Path normalization".
+        self._path_rewrites: "list[tuple[str, str]]" = []
 
     def is_started(self) -> bool:
         """Read-only introspection for callers/tests — mirrors
@@ -162,6 +184,19 @@ class FsWatcher:
         self._queue = asyncio.Queue(maxsize=_QUEUE_MAXSIZE)
         self._drain_task = asyncio.create_task(self._drain_events())
 
+        # #2623: build the resolved->configured path rewrite table BEFORE
+        # scheduling — the OS backend (fsevents on macOS, symlink-transparent
+        # on Linux too) reports events at the REALPATH regardless of which
+        # path string we hand ``observer.schedule``, so every reported event
+        # path needs rewriting back onto what the operator actually wrote in
+        # ``fs_watch.paths`` for a ``matcher: {path: <configured>}`` to match.
+        self._path_rewrites = []
+        for configured in self._paths:
+            resolved = os.path.realpath(configured)
+            normalized_configured = os.path.normpath(configured)
+            if resolved != normalized_configured:
+                self._path_rewrites.append((resolved, normalized_configured))
+
         handler = _build_handler(FileSystemEventHandler, self)
         observer = Observer()
         for path in self._paths:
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — 🔴 BLOCKING: CI RED (do not merge). `test_symlinked_watch_path_reports_events_under_the_configured_prefix` FAILS on Linux CI (both pytest 3.11 + 3.12) with `IndexError: list index out of range` at test_2608_h4_fs_watcher.py:197 — the events list is empty (no event captured). It passed on your local env (macOS) but fails on CI (Linux). Env-dependent test failure. NOTE: only THIS new symlink test fails (the run-loop-pickup test + the pre-existing #2599 web-route tests are fine on CI — 1 failed / 7447 passed). Needs a fix before merge (see broker message for the ask).

tya5:
**[per-PR-coder]** — CI-RED fix pushed (94bc7560).

**Root cause (real behavior gap, not just timing):** the symlink test passed on macOS fsevents but failed on Linux inotify because my first fix scheduled the watchdog observer on the *configured symlink* path and only used realpath for the rewrite table. A recursive inotify watch scheduled on a **symlink root** is handled inconsistently vs fsevents (which reports the realpath) — on Linux no event fired, so `trigger.calls` stayed empty and `calls[0]` raised `IndexError`.

**Fix (`src/reyn/runtime/fs_watcher.py`):** schedule each watch on `os.path.realpath(configured)` on **both** platforms. inotify watches inodes, and the symlink + its realpath target share one inode, so a write via the configured symlink path fires on the realpath-registered watch — the *same* well-tested real-directory watch path every other H4 test uses (and those already pass on Linux CI, which is my primary evidence the resolved-watch path fires on Linux). Events then always arrive under the resolved prefix, and `_rewrite_path()` maps them back to the operator's configured prefix uniformly. The reported `path` template var is the configured path on both platforms, so `matcher: {path: <configured>}` Just Works.

**Test hardened (not a blind sleep-bump):** the wait is already a poll-until (`_wait_for`), now with a generous budget, plus an **explicit non-empty assertion** with a diagnostic — so a genuine no-fire fails loudly instead of an opaque `IndexError`. Per lead-coder's directive: wait-until-event, never fixed-sleep.

**Gates (macOS, pytest FOREGROUND):**
- `test_tier_audit.py --strict` (both files) → 15 inspected, 0 warnings, 0 errors
- `verify_module_docstrings.py fs_watcher.py` → OK
- `pytest -q` → 5 failed / 7439 passed / 21 skipped — the 5 are the pre-existing #2599 web/route-mounting failures; **zero NEW**. `test_2608_h4_fs_watcher.py` (14) + `test_2608_run_loop_pickup.py` (1) all pass.

Couldn't repro locally (macOS only, no Linux/docker daemon available), so the fix is reasoned from inotify inode semantics + the fact that the sibling real-dir H4 tests already pass on Linux CI. Ready for re-review / CI.

PR #2644, opened 2026-07-07
feat: reyn pipe run supports tool:/agent: steps (extracts reusable AgentRegistry builder from reyn chat)

tya5:
**[per-PR-coder]** — Corrects a scope decision in the just-merged `reyn pipe` CLI (#2643): `reyn pipe run` refused any pipeline reaching a `tool:`/`agent:` step, reasoning that real dispatch needed "session-adjacent machinery disproportionate to a CLI wrapper." Verified (by reading the actual executor code) that this was wrong.

## Corrected understanding

- `AgentStep`'s handler (`core/pipeline/executor.py::_run_agent_step`) calls `reyn.runtime.session_api.run_agent_step` — the **lightweight ephemeral-session-spawn primitive** (`spawn_ephemeral_session` + one `MessageBus` turn + collect). It needs `deps.registry` (a real `AgentRegistry`) — not a live chat session, not a router loop, and not `run_pipeline`'s own IS-6 driver-session/MessageBus-attach machinery (that machinery is for the *top-level pipeline run's own* crash-resilience — an unrelated concern to what one `agent:` step needs).
- `ToolStep`'s handler dispatches via `_make_tool_dispatch(ctx: ToolContext)` (`tools/pipeline_verbs.py`) — the SAME seam `invoke_action` uses. It needs a real `ToolContext` (`tools/types.py`) — a dataclass whose `router_state`/`resolver`/`hot_reloader`/`state_log` fields are each documented to gracefully degrade to `None` (not an all-or-nothing live-session requirement).

So both are constructible standalone, without a live chat session or router loop.

## What changed

- **`src/reyn/runtime/registry_bootstrap.py`** (new) — `build_agent_registry_from_project(project_root, config, *, non_interactive=False, agent_name=None) -> AgentRegistry`: a minimal, standalone `AgentRegistry` builder for a non-interactive, one-shot caller. Also extracts `build_state_log`/`build_budget_tracker`, the small mechanical pieces every frontend factory site builds identically.
  - Deliberately **not** a superset of `reyn chat`'s full richer construction (model selection, `--exclude-tools`, environment-backend choice, interactive CUI logging, …) — see the module docstring for why forcing chat's ~25-kwarg surface through one helper would add real drift risk for no evidenced second caller. `chat.py`'s own `run()` now reuses `build_state_log`/`build_budget_tracker` (byte-identical extract-method, zero behavior change) but keeps its own bespoke `PermissionResolver`/session-factory/`AgentRegistry` construction, since it needs strictly more than a one-shot minimal registry provides.
  - v1 scope choices, stated explicitly in the module docstring: no budget-ledger hydration (a one-shot CLI invocation has no persistent multi-turn budget to resume), host environment backend only (no `--docker`/`--sandbox-backend` CLI surface — mirrors `build_environment_backend`'s own host default), operator-trusted permissions (`file.read`/`file.write`/`http.get` default `allow`, `setdefault` so an explicit `reyn.yaml` permission wins — mirrors how `reyn pipe install` already treats the CLI as operator-trusted).
  - **Note (pre-existing, out of scope):** `mcp.py::run_serve` independently duplicates ~80% of this same construction (its own `state_log`/`budget_tracker`/`PermissionResolver`/session-factory/`AgentRegistry`). Not touched here — refactoring it too would be a second load-bearing change without an explicit ask; flagging it as a follow-up candidate.

- **`src/reyn/interfaces/cli/commands/pipe.py`** — `run_run` now builds:
  - a real `ToolContext` via `_build_run_tool_context`: `events` = real `EventLog`; `permission_resolver` = operator-trusted (mirrors `reyn pipe install`); `workspace` = a **real** `reyn.data.workspace.Workspace` anchored on `project_root` (host backend) — NOT a synthetic `base_dir`-only stand-in like `pipeline_install` uses, because a real tool handler (`read_file`, etc.) calls real methods on it (`read_file_bytes`, …) that a stub object lacks. (Caught this via a manual smoke test against the real `read_file` tool — the unit test's own stub tool didn't need it, so this would have shipped a broken `ToolContext.workspace` for any real file/exec tool without the extra check.) `caller_kind="router"` — confirmed by reading `core/dispatch/dispatcher.py`: it's forwarded verbatim into `tool_called`/`tool_returned` audit events, purely a taxonomy label (every existing caller, including the non-interactive pipeline driver-session, already sets this same literal) — not a live-router claim. `router_state`/`resolver`/`hot_reloader`/`state_log` = `None` (documented graceful degradation).
  - a real `AgentRegistry` via `build_agent_registry_from_project(project_root, config, non_interactive=True)`, passed as `registry=` with `default_identity=DEFAULT_AGENT_NAME` ("default") into `PipelineExecutor.run(...)`.
  - Removed the blanket `_UNSUPPORTED_STEP_LABELS`/`_find_unsupported_steps` pre-flight refusal entirely.

## What's still refused (narrow, accurate)

A `tool:` step that specifically needs `ctx.router_state` (e.g. a hand-authored tool reading `router_state.pipeline_registry` directly) still fails — with **that tool's own clear error**, not a blanket "tool:/agent: not supported." Nothing in the shipped tool catalog needs `router_state` outside of `run_pipeline` itself, which is already structurally denied inside a pipeline step (R6 S3, nesting is call-only) regardless of this PR.

## Safety verification (chat.py behavior-preservation)

`chat.py`'s own `run()` is refactored to call `build_state_log`/`build_budget_tracker` instead of inlining the same 10 lines — same logic, same call sites, same defaults (`hydrate=True`). Ran the full chat/session/registry-construction test suite (46 tests across `tests/chat/`, `test_chat_cli_flags.py`, `test_2425_step1c_chat_chokepoint.py`, `test_chat_exclude_tools_187.py`, `test_chat_grant_file_write_187.py`, `test_chat_inline_interactive_gate.py`, `test_scoped_session_factory_invariant_1402.py`, `test_swe_bench_runner_chat_187.py`) plus a broader `-k "chat or 187"` sweep (311 tests) — all pass unmodified.

## Tests

- `test_run_tool_step_dispatches_for_real` — a real `tool:` step (a test tool registered into the real `ToolRegistry`, mirroring `test_pipeline_is4_inline.py`'s `_install_write_tool` idiom) dispatches end-to-end via `reyn pipe run`.
- `test_run_agent_step_spawns_real_ephemeral_session` — a real `agent:` step spawns a genuine ephemeral `AgentRegistry`/`Session`/`MessageBus` turn under the `default` identity; only `litellm.acompletion` is faked (the documented replay seam per `test_llm_request_event_1669.py` / `docs/deep-dives/contributing/testing.md`), asserting the pipeline's final output is the faked LLM reply threaded through R3.
- Replaced the old `test_run_tool_step_is_refused_with_clear_error_not_a_crash` (the behavior it pinned no longer exists).
- Manual smoke test against the **real, shipped `read_file` tool** (not the unit test's stub) — this is what surfaced the synthetic-`Workspace` gap described above.

## Docs

Updated `docs/guide/for-users/write-a-pipeline.md`/`.ja.md` §5 to describe the new `tool:`/`agent:` support and the (host-only, no live-REPL) v1 scope.

## Gates (all run locally, all green — pre-existing unrelated failures confirmed on clean HEAD via `git stash`)

- `python scripts/test_tier_audit.py --strict tests/test_cli_pipe.py` — 15/15 OK.
- `pytest` (full suite from repo root) — 7520 passed, 21 skipped, 1 deselected. 4 pre-existing route-mount-ordering failures (`test_fp0001_a2a_endpoints.py::test_fp0001_routes_mounted`, `tests/web/test_a2a.py::test_a2a_routes_mounted`, `tests/web/test_mcp_sse.py::test_mcp_routes_mounted`, `tests/web/test_plugin_loader.py::test_loader_disambiguates_via_package_field`, `tests/web/test_resources_router.py::test_resources_route_is_mounted_on_app`) reproduce identically on clean HEAD (`git stash` + re-run) — unrelated to this change.
- `python scripts/verify_module_docstrings.py <changed src files>` — clean.

## Test plan

- [x] `test_tier_audit.py --strict tests/test_cli_pipe.py` — 15/15 OK, 0 Tier-4.
- [x] Full `pytest` from repo root — green modulo the 5 pre-existing, unrelated route-mount failures (verified via `git stash`).
- [x] `verify_module_docstrings.py` on changed src files — clean.
- [x] Manual: `reyn pipe run` against a real registered pipeline with a real shipped tool (`read_file`) — verified genuine dispatch + surfaced/fixed the synthetic-`Workspace` gap.
- [x] `reyn chat`'s full existing test suite — green, unmodified assertions.


tya5:
**[per-PR-coder]** — Fixed the permission-default security issue. Confirmed your read of chat.py (`grep -n "grant_file_write\|perm_config" src/reyn/interfaces/cli/commands/chat.py`) is exactly right: `file.write` is only granted under `--grant-file-write`, and `http.get` is never blanket-granted at all (relies on `require_http_get`'s JIT prompt instead).

**Fix (commit `3ef2ffeb`):**
- `registry_bootstrap.py::build_agent_registry_from_project` — removed the unconditional `setdefault` calls. Added a `grant_file_write: bool = False` param; `perm_config` defaults to exactly `reyn.yaml`'s own `permissions:` section (byte-identical to `reyn chat`'s no-flag posture). Only sets `file.read`/`file.write` to `allow` when the caller passes `grant_file_write=True`. `http.get` is never blanket-granted, matching `reyn chat` exactly (stated in the docstring's reasoning, per your point 3).
- `pipe.py` — added `--grant-file-write` to the `run` subcommand (same flag name/semantics as `reyn chat`'s), threaded into both `_build_run_tool_context` (tool-step dispatch) and `build_agent_registry_from_project` (agent-step-spawned sessions), so both paths share the same fail-closed posture.
- Verified manually before writing tests: a real `write_file` tool step against an out-of-`.reyn/` path is DENIED by default (`status: "denied"`, actionable message) and succeeds only with `--grant-file-write`.
- New tests: `test_run_tool_step_file_write_is_denied_without_grant_flag`, `test_run_tool_step_file_write_allowed_with_grant_flag` (both against the real, shipped `write_file` tool — not a stub), plus `test_pipe_run_grant_file_write_flag_parses`.

**All 4 gates re-run and green:**
- `python scripts/test_tier_audit.py --strict tests/test_cli_pipe.py` — 18/18 OK.
- `pytest` (full suite from repo root) — 7523 passed, 21 skipped, 5 deselected (the same pre-existing, unrelated route-mount-ordering failures already confirmed via `git stash` on clean HEAD).
- `python scripts/verify_module_docstrings.py <changed src files>` — clean.

Pushed to `feat/pipe-run-tool-agent-steps` (3ef2ffeb). Ready for re-review.

PR #2651, opened 2026-07-08
feat(offload): config opt-out — 3-gate disable (PR-3, tool-result-schema-redesign)

tya5:
**[per-PR-coder]** — PR-3 of the tool-result-schema-redesign arc: `offload:` config opt-out + 3-gate disable.

## Summary
- New `offload:` reyn.yaml section (`OffloadConfig`, `enabled: bool = true`), wired root→loader→Session→`ContextBudgetAdvisor`/`RouterHostAdapter` the same way `cost_warn:` is threaded.
- `enabled: false` disables all three size gates named in `docs/proposals/tool-result-schema-redesign.md` §5:
  1. text token cap — `ContextBudgetAdvisor.cap_tool_result()` early-returns.
  2. structured inline cap — `build_offload_body(..., enabled=False)` never offloads to a `structured_ref` regardless of size.
  3. media follow-up budget — `media_followup_budget()` returns `None` (unbounded).
- Per the design doc's explicit ban: **not** implemented by forcing `per_turn_cap_tokens()` to 0 (that also feeds `media_followup_budget`'s subtraction — zeroing it would kill media follow-ups for the wrong reason). Each gate has its own early-return/flag; `per_turn_cap_tokens()` itself is untouched.
- `offload_disabled` warning event emitted at session start when the flag is off (documents the accepted risk: a single tool result can then exceed the compaction-batch budget, recreating the #1128 dead-end).
- Docs: `reyn.local.yaml.example` + `docs/reference/config/reyn-yaml.md` updated (doc-mirror coverage test enforces this).

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_offload_config_gate_pr3.py` — 8/8 OK, 0 Tier-4 violations.
- [x] `pytest` (full suite, repo root) — 7545 passed, 18 skipped, 0 failed.
- [x] `python scripts/verify_module_docstrings.py <changed src files>` — OK, no refactor narrative.
- [x] Config round-trip (non-default value): `offload.enabled: false` in a real `reyn.yaml` reaches `ReynConfig.offload.enabled` via `load_config` (`test_load_config_reads_offload_enabled_false`).
- [x] Each of the 3 gates exercised through its own real seam (`ContextBudgetAdvisor`, `build_offload_body`) with an enabled=True control test proving the gate normally fires — no MagicMock, real instances only.

Note: encountered 1 pre-existing-looking failure (`test_reyn_api_relocate_311.py::test_old_paths_are_gone_clean_break[reyn.api.unsafe]`) during local runs — root-caused to a stale empty `src/reyn/api/unsafe/` directory left on disk in this checkout (git doesn't prune empty dirs after a file-removal commit; Python's implicit namespace-package resolution then made the "should be gone" import silently succeed). Confirmed via a disposable worktree that main is clean; removed the empty directory locally (not part of this diff — no tracked content, nothing to commit).

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -1121,6 +1121,21 @@ cost_warn:
 
 **Pricing source:** reyn looks up model costs from the [LiteLLM pricing database](https://github.com/BerriAI/litellm) (`litellm.model_cost`). Models not in the database are treated as below-threshold (no warning). Custom or proxy models that resolve to a key in the database will be matched.
 
+## `offload` block
+
+Debug/experiment lever disabling **all three** tool-result size gates (tool-result-schema-redesign §5): the text token cap, the structured-data inline cap, and the media follow-up budget bound. With `enabled: false`, every tool result is emitted to the LLM in full — never truncated, never offloaded to a file ref. The LLM-visible format (frontmatter + text) is unchanged either way; only whether size gates truncate varies, isolating that as the sole experimental variable.
+
+```yaml
+offload:
+  enabled: true   # false = never truncate, always emit tool results in full
+```
+
+| Field | Type | Default | Description |
+|---|---|---|---|
+| `enabled` | bool | `true` | Master switch. `false` disables the text token cap, the structured inline cap, and the media follow-up budget bound. |
+
+**Not a recommended steady-state setting.** With offload disabled, a single tool result can exceed the model's compaction-batch budget, recreating the pre-#1128 compaction dead-end (a turn too large to ever compact). A `offload_disabled` warning event is emitted at session start when this is set, so traces stay self-explaining.
+
 ## MCP servers
 
 External tool servers reyn can call via the [Model Context Protocol](../../concepts/tools-integrations/mcp.md). Each entry under `mcp.servers:` is keyed by a short name (the same name the skill declares in `permissions.mcp` and emits in `mcp` ops).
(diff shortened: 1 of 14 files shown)


tya5:
**[per-PR-coder]** — 修正 push しました。

ご指摘ありがとうございます。まず事実確認: 元PRの diff には `test_offload_disabled_emits_warning_event_shape` という名前のテストや `"agent_name" in {"agent_name"}` という assert は存在しませんでした（`gh pr diff 2651` で確認、コミットは1つのみ）。ただ、ご指摘の**本質的な論点は正しく**、`offload_disabled` イベントの実発火を検証するテストが一つも無かった（`grep -rn offload_disabled tests/` は当時ゼロ）のは事実で、そこを直しました。

修正内容:
- `test_offload_disabled_emits_warning_at_session_start`: real Session + real EventLog subscriber、`session.inbox.put_nowait(("shutdown", {}))` → `await session.run()` で session.py:run() の実パスを通す（test_session_lifecycle_events_1800.py と同じ技法）。
- Falsify 確認済み: session.py の emit 行を一時的に消して RED になることを手元で確認 → 復元。
- enabled=True の control テストも追加（発火しないことを確認）。
- ついでに `offload_disabled` を `EVENT_AUDIT_REQUIREMENTS` に宣言（FP-0021 audit-completeness 慣習）。

CI4ゲート再確認済み: ruff clean / tier-audit 10/10 OK / pytest 7547 passed 0 failed 18 skipped / module-docstring OK。再レビューお願いします。

PR #2653, opened 2026-07-08
docs: present-layer proposal (present op + declarative tool-data presentation)

tya5:
**[architect]** — DRAFT design proposal for a **present layer**: a `present` op (Control IR) plus a surface-agnostic, declarative presentation model that lets bulk tool-result data reach the user **without round-tripping through LLM output tokens**.

Doc: `docs/deep-dives/proposals/0054-present-layer.md` (sequential proposals home; `deep-dives/` is excluded from the public mkdocs site, matching `0053-tool-result-schema-redesign.md`).

## What & why
- **Problem**: showing tool data to the user costs LLM output tokens (reproduction) and loses fidelity (summarization). Today, bulk data reaches the user only via `summarize_tool_result` (1-line, max fidelity loss) or full LLM output reproduction (max token cost) — no rich middle ground exists.
- **Design**: hub-and-spoke, reyn-native declarative model **structurally isomorphic to A2UI** (JSON-Pointer bindings, vetted non-executable catalog); wire adapters (A2UI / MCP Apps) deferred to boundary. Consumes the tool-result-schema-redesign arc's landed offload refs (`structured` / `structured_ref` / `structured_preview` + text-side inline truncation-note ref). That arc is now **fully closed** (PR-0..PR-4 merged), so the dependency is met.
- **Security**: declarative-not-executable (structural write-gate parity); presentation-guard at the output seam; blind-routing recorded as an OS-computed audit annotation, not gated.

## Reviews incorporated
- **docs-maintainer**: field-name accuracy (`structured_ref` vs non-existent `text_ref`), arc-status facts, FP-0051 superseded.
- **tui-coder**: inline-CUI reality (Rich via `repl/renderer.py`; scrollback uncapped; present carries its own head-N cap); cap-before-render for code/diff; explicit `Console(width=...)`.
- **lead-coder**: design APPROVE; high-signal ack (`{path, reason}`), explicit `resolve_present_source` data_ref seam, recovery-gate N/A rationale in §8 + PR-D.

## Status
- **Owner RATIFIED 2026-07-08**: v1 display-only catalog approved. Owner also asked for markdown + templates — both already in v1 scope (markdown = §2 catalog; templates = §3 4-stage fallback; display-only means non-interactive, not content-limited).
- Implementation dispatch: lead-coder (review/merge owner) proceeds PR-A → PR-D.

## Test plan
- [x] (skipped — docs-only) No src/tests touched; ruff / tier-audit / docstring-verify are N/A.
- [x] Location: `docs/deep-dives/proposals/` — excluded from public mkdocs site (`exclude_docs: deep-dives/`), so no nav/strict-build concern; parity with `0053-tool-result-schema-redesign.md`.
- [x] Owner ratify: v1 display-only catalog — **ratified 2026-07-08** (markdown + templates confirmed in-scope).

### DIFF
--- a/docs/deep-dives/proposals/0054-present-layer.md
+++ b/docs/deep-dives/proposals/0054-present-layer.md
@@ -319,6 +319,11 @@ differentiator is making blindness *auditable*, not forbidding it.)
 - **PR-C** — `presentations.yaml` registration + hot-reload seam + the 4-stage fallback
   chain (built fresh — FP-0051's registry is deleted; §§ 3, 6).
 - **PR-D** — replay/rewind placeholder rendering + docs (concepts page + reference).
+  **The PR-D description must restate the § 8 recovery-gate answer** (pure re-render /
+  placeholder, no authoritative state reconstructed from `presented` events ⇒
+  truncate-falsify gate N/A) so a future reviewer does not re-flag it — and must
+  re-trigger the gate in-arc if the implementation ever reconstructs state from
+  `presented` events.
 
 ## Test plan (per testing.ja.md tiers)
 
--- a/docs/deep-dives/proposals/0054-present-layer.md
+++ b/docs/deep-dives/proposals/0054-present-layer.md
@@ -10,7 +10,7 @@ always round-trips through LLM tokens. Splitting the cost into axes:
 
 | Axis | Cost | Status |
 |---|---|---|
-| **A. Ingestion (input)** | tool result entering LLM context | **Solved** by the offload mechanism + tool-result-schema-redesign arc (`docs/proposals/tool-result-schema-redesign.md`): data lands in a ref file, the LLM sees schema + preview, reads back on demand via `file__read` |
+| **A. Ingestion (input)** | tool result entering LLM context | **Solved** by the offload mechanism + tool-result-schema-redesign arc (`docs/deep-dives/proposals/0053-tool-result-schema-redesign.md`, IMPLEMENTED): data lands in a ref file, the LLM sees schema + preview, reads back on demand via `file__read` |
 | **B. Reproduction (output)** | the LLM re-types the data as output tokens to show the user | **Unsolved — target of this proposal** |
 | **C. Fidelity loss** | to fit output, the LLM summarizes/truncates; the user loses data | **Unsolved — same mechanism solves it** |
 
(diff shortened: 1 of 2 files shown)


tya5:
**[tui-coder]** — §5/§6 review, as requested.

## §5 default output cap (head-N + ref tail) — natural fit, one refinement

The policy is right for inline-CUI: scrollback is genuinely uncapped by design (no pager), so an unbounded-by-construction primitive like `present` needs its own cap independent of that convention, and "orthogonal to the 12-row live-region caps" is the correct framing — those two are unrelated axes and conflating them would be a bug magnet later.

One thing worth being explicit about in the doc: apply the head-N cap **before** any expensive rendering step (Rich `Syntax`/pygments highlighting for `code`, diff coloring), not after. Syntax-highlighting a full multi-MB blob and then truncating the *rendered* output wastes the render cost the cap was supposed to avoid, and for `code`/`diff` in particular pygments highlighting is not cheap at scale. Truncate the raw bound value first, then render only the head-N slice.

## §6 renderers — one concrete pitfall from `repl/renderer.py`'s existing pattern

The Rich `Console → StringIO → run_in_terminal` pattern you're riding already exists twice in `repl/renderer.py` (`RichChatRenderer.__init__` line 187, and the inline-CUI-era renderer at line 672) — both construct `Console(highlight=False, file=self._buffer, force_terminal=True)` with **no explicit `width=`**. Rich's width auto-detection only works against a real tty; against a `StringIO` file object it silently falls back to a fixed default (80 cols), independent of the user's actual terminal width. Neither existing call site overrides this (`grep -n "width="` in that file only turns up an unrelated `grid.add_column(width=...)`).

For `present`'s `table`/`code`/`diff` components this matters more than it does for the existing plain-text/panel rendering, since a table column layout computed at a stale 80-col width will wrap or truncate badly on a wider real terminal. Recommend explicitly passing `width=` sourced from the live prompt_toolkit app's output size (e.g. `get_app().output.get_size().columns` at print-time, re-read per call since the terminal can resize between turns) when constructing `present`'s `Console`, rather than inheriting the existing renderers' implicit-80 behavior. Might also be worth a fast-follow fix on the two existing call sites, since they have the same latent bug today (just less visible for plain text/one-liners than it'll be for `present`'s tables).

Everything else in §5/§6 reads as a solid, correctly-scoped fit to the actual inline-CUI architecture — nice job tracking down the FP-0051 deletion and not trying to resurrect a base that no longer exists.

tya5:
**[tui-coder]** — one small nit I noticed while reading, not part of the §5/§6 review above: **PR-C** in "Suggested PR sequencing" says "4-stage fallback integration with FP-0051 registry", but §3 and §6 both correctly state FP-0051's registry was deleted with no relocation and is being rebuilt fresh (design lineage only, not a base to extend). Might want to reword PR-C to "content-type default viewer (FP-0051 design lineage, rebuilt)" or similar so it doesn't read as "integrate with a registry that still exists somewhere."

Issue #1983, opened 2026-06-21
dual-source op catalog: mcp_drop_server missing from ControlIROp union + op_catalog truncated (advertised≠enforced)


tya5:
**[lead-coder]** — found by a structural-defect recon (Opus sub-agent, origin/main @ `01766db8`, primary-evidence). Distinct class from #1977: the **LLM-facing op catalog is a separate hardcoded list, NOT derived from the authoritative `OP_KIND_MODEL_MAP` / `ControlIROp` union** that the validator + dispatcher enforce. Two manifestations, single root.

### Finding 3 (HIGH) — `mcp_drop_server` is registered + documented but the phase validator structurally cannot accept it
- **Advertised/registered**: `docs/reference/runtime/control-ir.md:29` documents `mcp_drop_server` as a phase op kind; `src/reyn/tools/__init__.py:191-193` `registry.register(MCP_DROP_SERVER_OP)` ("router+phase callable"); model `MCPDropServerIROp` (`schemas/models.py:411`, `kind: Literal["mcp_drop_server"]` `:429`) + handler (`core/op_runtime/mcp_drop_server.py:275`) exist.
- **Enforced surface MISSING it**: the `ControlIROp = Annotated[Union[...], discriminator="kind"]` union (`schemas/models.py:673-697`) includes `MCPInstallIROp` but **NOT** `MCPDropServerIROp`; `OP_KIND_MODEL_MAP` (`registry.py:116-167`) has `mcp_install` but no `mcp_drop_server`, so `ALL_OP_KINDS` excludes it.
- **Effect**: a phase emitting `{"kind":"mcp_drop_server",…}` fails `ActOutput.model_validate` (`phase_executor.py:964`) → `output_validation_failed` → retried → `phase_failed`. A skill declaring `allowed_ops:[mcp_drop_server]` fails DSL lint (`linter.py:133` checks `ALL_OP_KINDS`). **Violates the CLAUDE.md `control-ir.md`↔`OP_KIND_MODEL_MAP` sync invariant in the worst direction** (advertise a capability the enforcer rejects). The contextual-gate completeness test (`test_control_ir_contextual_gate_1912.py:37`) passes *only because* `ALL_OP_KINDS` omits the op.
[...]

PR #2658, opened 2026-07-08
feat: present op (PR-A) — declarative model + binding + presentation-guard core (null renderer)

tya5:
**[per-PR-coder]** — This is **PR-A** of the present-layer arc. Design doc (owner-ratified, on main): `docs/deep-dives/proposals/0054-present-layer.md`. Part of the present-layer arc; PR-B (inline-CUI renderer), PR-C (presentations.yaml registry + fallback chain), PR-D (replay/rewind + docs) follow. Uses `part of`/`toward` phrasing — no closing keyword (sub-PR).

## Scope (exactly the doc's "PR-A")

`present` op + declarative model + binding resolution + presentation-guard core. **No UI yet** — the op returns its ack against a **null renderer** (`surface: ["null"]`).

### Op contract
- **Tier 0** (`ask_user`'s sibling), **fire-and-continue** — does NOT pause the run.
- **Arg validation** (pydantic `model_validator`): `data_ref` **XOR** `data_inline`; `template` **XOR** `blueprint`. `is None` checks distinguish absent from present-but-falsy (so `data_inline: {}` is valid).
- **Ack** (the LLM's only feedback): `{ok, bindings_resolved, bindings_dropped, rows}` where `bindings_dropped` is a list of `{path, reason}`, `reason ∈ {path_not_found, type_mismatch, guard_stripped}` (the refined d24d72e shape). Plus `all_bindings_missed` (generic-viewer fallback signal; the actual fallback wiring is PR-C).

### Declarative model (display-only, non-executable by construction)
Catalog: `text`/`markdown`/`code`/`diff`/`keyvalue`/`table`/`list`/`image`. Bindings expressed structurally as `{"$bind": "<json-pointer>"}` — an RFC 6901 pointer **string**; everything else is a literal. Structural gate at op validation rejects non-catalog components and non-path bindings (hard error, not a soft drop); labels escaped at parse (FP-0051 fence generalized).

### Binding resolution
JSON Pointer (RFC 6901); table/list column paths resolve **row-relative**. Hit → bind; miss → soft-skip + `path_not_found`; type mismatch → coercion (scalar into table `rows` → 1 row) + `type_mismatch`; all-miss → `all_bindings_missed` exposed.

### `resolve_present_source` + read-authority equivalence
`resolve_present_source(data_ref, ctx)` is the single 2-arc junction: it gates via the **same `require_file_read`** call the file ops use (resolved through `_resolve_for_gate`), then reads the ref and re-hydrates an offloaded `structured_ref` (JSON bytes → object) to its full value — offloaded-vs-inline transparent to the caller. A denied read raises `PermissionError` → dispatch returns `status="denied"`. **Invariant: present denied ⇔ file.read denied** (tested both directions with a REAL `PermissionResolver`, config `file.read: deny` + falsify allow).

### Presentation-guard
Runs **unconditionally** (incl. never-ingested data): neutralizes terminal escape/control sequences (`\x1b…`), Rich markup (`[tag]` → `\[tag]`), HTML (entity-escape) on every rendered leaf; **per-binding size caps** (leaf chars + row count) prevent a `/` root pointer dumping a whole file. A neutralized/capped binding is reported `guard_stripped`.

### `presented` event (P6)
`{data_ref, template, surface, ingested, bindings_resolved, bindings_dropped, rows}`. **No content bytes** (refs + stats only). `ingested` (none|partial|full) is **OS-computed** from the events log (prior `read_file` on the ref?), never LLM-self-reported.

### #1983 hard rule
`OP_KIND_MODEL_MAP` (`src/reyn/schemas/models.py`) ↔ `docs/reference/runtime/control-ir.md` sync is **in this PR**: `present` added to the map (+ TYPE_CHECKING union mirror) AND a full `## present` section + table row in control-ir.md.

### Recovery gate
**N/A** (per doc §8): `present` writes no WAL-event-derived reconstruction state; the `presented` event is durable audit, not recovery-core state. The truncate-falsify gate does not apply.

## Judgment calls (flagged for review)
1. **Binding marker `{"$bind": "<pointer>"}`.** The doc mandates "bindings are path expressions only" / JSON Pointer but doesn't fix the literal-vs-binding syntax. I chose a single-key `$bind` object so the "path-only" rule is enforceable by shape (a `$bind` value that isn't a pointer string is rejected). Alternative (a `$`-prefixed string convention) is more terse but harder to gate cleanly.
2. **Blueprint top-level = node | list[node].** The v1 catalog has no container component, so I made a blueprint either a single node or a flat sequence (rendered top-to-bottom). No nesting/children.
3. **`surface: ["null"]` in PR-A.** The doc's event example shows `surface: [inline-cui]`, but PR-A has no renderer — I emit `["null"]` (honest; PR-B swaps in the real surface).
4. **Named `template` deferred to PR-C.** PR-A resolves inline blueprints only; a `template` arg is accepted at the schema level but the handler returns `ok=False` + a `note` (audit event still emitted). The registry + 4-stage fallback are explicitly PR-C.
5. **`guard_stripped`/`type_mismatch` count as resolved AND flagged.** They render (coerced/neutralized), so they increment `bindings_resolved` and appear in `bindings_dropped` as issues; only `path_not_found` is a true non-render. `bindings_dropped` is thus an "issues" list keyed by reason (matches the doc's ack example carrying all three reasons together).
6. **`ingested` for `data_inline` = `full`** (data is in the LLM's context by construction).

## Gates (all four, from repo root — pytest-green ≠ CI-green)
- [x] `python scripts/test_tier_audit.py --strict tests/test_present_op_fp0054_pra.py` — Exit 0 (0 Tier-4 violations)
- [x] `pytest` — 7569 passed; the only 5 failures (`test_fp0001_a2a_endpoints`, `test_a2a`, `test_mcp_sse`, `test_plugin_loader`, `test_resources_router`) are the known pre-existing route-mount failures — confirmed via `git stash` that they reproduce identically on clean main; none in files this PR touches.
- [x] `python scripts/verify_module_docstrings.py <changed src>` — OK (no refactor narrative)

## Test plan
- [x] Tier 1: binding hit / miss→soft-skip / scalar-into-table coercion / row-relative column paths / column-missing-on-all-rows / all-miss / JSON-Pointer RFC 6901 (whole-doc + `~0`/`~1` escapes)
- [x] Tier 1: ack shape — drops as `{path, reason}` across all three reason categories
- [x] Tier 1: blueprint structural gate — non-catalog component rejected, non-path binding rejected, unknown slot rejected
- [x] Tier 1: guard — terminal escape neutralized (not rendered) via literal-bracket FP-0051 idiom; Rich markup escaped literal; `/` root size-capped; labels escaped at parse
- [x] Tier 1: `presented` event field presence incl. OS-computed `ingested` (none/partial/full from prior read)
- [x] Tier 1: read-authority equivalence — present denied ⇔ file.read denied (real resolver) + falsify (allowed) direction
- [x] Tier 2: fire-and-continue (no intervention_bus needed, run not paused); no content bytes in the event payload

### DIFF
--- a/docs/reference/runtime/control-ir.md
+++ b/docs/reference/runtime/control-ir.md
@@ -19,6 +19,7 @@ Control IR is the list of side-effect operations the LLM may emit alongside its 
 | `glob_files` | List files matching a glob pattern | `file.read` |
 | `grep_files` | Search file contents by regex | `file.read` |
 | `ask_user` | Pause the phase and ask the user a question | none (always allowed) |
+| `present` | Route bulk data + a display template to the user surface without the data passing through LLM output tokens (fire-and-continue) | Tier 0 (always allowed); `data_ref` read authority == `file.read` |
 | `sandboxed_exec` | Run argv under a `SandboxPolicy` via a `SandboxBackend` (replaces the removed `shell` op) | enforced by backend (`SandboxPolicy`) |
 | `web_search` | Search the public web via DuckDuckGo | Tier 1 — default allow; `web.search: deny` in `reyn.yaml` blocks |
 | `web_fetch` | Fetch a single URL and return extracted text | Tier 1 — default allow; `web.fetch: deny` in `reyn.yaml` blocks |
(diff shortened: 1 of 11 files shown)


tya5:
**[per-PR-coder]** — Guard revision pushed (7864794c), ready for re-review. Addresses the literal-bypass finding by root-fixing the bug class per the architect's per-surface ruling.

**1. Single leaf-neutralization seam.** Removed the parse-time `catalog._escape_label` path entirely; `catalog.py` is now purely structural (keeps labels raw). ALL render-leaves — labels, literal slot values, bound values, table cells, list items — now route through ONE neutralizer in `binding.resolve_bindings` (`_guard` / `_guard_maybe` helpers). No code path reaches a renderer un-neutralized; the "already escaped at parse" comment is gone.

**2. Terminal neutralizer = ESC/control strip + Rich-markup ESCAPE; HTML-escape dropped.** `TerminalNeutralizer.neutralize` strips ESC/control (OSC/CSI) and backslash-escapes Rich `[tag]` markup only. HTML entity-escaping is removed — `<div>` in a `code`/`diff` slot now survives byte-for-byte (was corrupted to `&lt;div…` before).

**3. Per-surface strategy structure.** `guard.py` exposes `get_neutralizer(surface)` dispatching by surface name (`terminal`/`null` → `TerminalNeutralizer`); `resolve_bindings(nodes, doc, *, surface=...)` asks for the strategy and never branches on surface itself. A future `web` strategy (HTML-escape) registers in `_STRATEGIES` without touching the seam or binding layer.

**Test evidence (RED-against-old / GREEN-against-new).** Verified by stashing only the src changes and running the two new tests against old src:
- `test_literal_text_slot_value_is_neutralized_via_single_seam` — RED (old code left literals un-neutralized) → GREEN.
- `test_terminal_neutralizer_does_not_html_escape_code_content` — RED (`assert '&lt;div…' == '<div…'`, proving old HTML-escape corrupted code) → GREEN.
- `test_labels_neutralized_at_render_seam` updated: asserts the structural gate keeps labels raw AND the seam neutralizes them.

**Four gates (all green):** ruff `src tests` ✓ · `test_tier_audit --strict` on the test file ✓ (25 OK, 0 Tier-4) · `pytest` ✓ (7571 passed; only the 5 known pre-existing route-mount failures, confirmed reproduce on clean main, none in touched files) · `verify_module_docstrings` on changed src ✓. `control-ir.md` guard section updated to the per-surface single-seam contract.

PR #2704, opened 2026-07-09
docs: remove dead phase-skill-engine drift from feature-map.md (#2696)

tya5:
**[docs-maintainer]** — part of #2696 (phase-skill-engine doc-drift removal), scoped to `feature-map.md` per lead-coder's initial request.

## Summary
`#2434` (closed 2026-07-02) deleted the heavy DSL/phase-graph skill engine, but `feature-map.md` still described it as live. Confirmed dead via direct grep before removing anything:
- `get_phase_op_catalog` — zero definitions in `src/` (only a dead `getattr`-fallback check remains in `router_loop.py`)
- `next_phase` — zero occurrences anywhere in `src/`
- `.reyn/artifacts/` — zero references anywhere in `src/`; `workspace.md` (feature-map's own linked target) never described it either

## Changes
- Removed the "Phase Engine" and "LLM Validation" mindmap nodes + their feature-index tables (both entirely about the dead engine — `next_phase`, the `control`/`artifact`/`control_ir` envelope, candidate gates, phase rollback).
- Removed the "Artifact storage" row under Workspace (P5).
- Fixed the `ask_user` op description — confirmed via `op_runtime/ask_user.py` + `schemas/models.py::AskUserIROp` that current `ask_user` has no phase concept at all; it pauses the run via the intervention bus, it does not "re-run the same phase."
- Dropped the dead `tool_use` "phase" layer from the Tool-Use Schemes section (same `get_phase_op_catalog` zero-definitions evidence — no host can ever select this layer).

**Explicitly left alone** (different, live, unrelated uses of "phase"): time-travel/rewind's "Phase 1 / Phase 2 (2a-2d)" rollout-phase mentions, and the RAG `#what-stays-out-of-phase-1` anchor.

## Scope note
A broader sweep (requested separately, reported to lead-coder directly) found the same drift class in several other concept/reference docs (`llm-output-contract.md`, `context-frame.md`, `llm-as-decision-engine.md`, and others) — some of these look like full-file deletion candidates and one touches `docs/index.md`'s top-level tagline. Holding those for an explicit scope decision before opening a second PR, given the larger blast radius.

## Test plan
- [x] `mkdocs build --strict` clean
- [x] `ruff check src tests` clean (docs-only change)
- [x] Every removal verified against current `src/` before deleting (not inferred from the doc text alone)

Issue #2696, opened 2026-07-09
[cleanup] 'phase' is a deprecated concept — drift-audit lingering code/doc references (verify-don't-infer which are vestigial vs load-bearing)


tya5:
**[lead-coder]** — tech-debt / drift-audit (owner declared "phase" a deprecated concept, 2026-07-09).

## Context
During the #2692 present-invocation-surface work, an Explore agent + the architect mistook the **stale `phase` execution-layer code** for a live invocation surface, which muddled the use-case discussion (present was framed as "reachable from phase" when phase is dead → it was reachable from nowhere). The owner clarified: **"phase" is a deprecated old concept**; lingering references in code/docs are stale drift.

## Ask (careful audit — verify-don't-infer)
Audit references to the `phase` concept and classify each as **vestigial (safe to remove)** vs **load-bearing (still wired, remove would break)** — do NOT declare a file stale by inference; verify each. Known reference points (starting set, not exhaustive):
- `get_phase_op_catalog` (router_loop.py)
- `gates.phase` / `ToolGates(phase=...)`
- phase scheme in `execution.py`
- "phase hosts" / phase op catalog wiring
- docs referencing phase as a surface/scheme

Deliverable: a classification table (ref → vestigial/load-bearing + evidence), then a scoped removal plan for the vestigial set. If any reference turns out load-bearing under a different name, document why it's retained.

## Priority
Low / cleanup — no functional bug (the confusion it caused is resolved). Value = remove a dead-concept drift source that already caused one design muddle. Run **after** the #2692 present fix lands. Suggested: docs-maintainer (doc refs) + a scoped code audit dispatched by lead.

Part of the #2688-era cleanup; not blocking.

tya5:
**[lead-coder]** — SCOPE CONFIRMED via primary evidence (not owner-asked, verified). The "phase" concept = the **DSL/phase skill feature deleted in #2434** ("[arc] Delete the heavy DSL/phase skill feature → chat-unified + lightweight skill", CLOSED 2026-07-02, owner GO). The lingering refs are vestigial drift from that deletion.

**Verified dead:**
- `get_phase_op_catalog` — NO definition exists; `router_loop.py:1404` does `getattr(host, "get_phase_op_catalog", None)` which always returns None → dead path.
- `tool_use: {chat, step, phase}` — the `phase` layer has no live catalog (dead getattr above).
[...]

PR #2711, opened 2026-07-09
docs: remove confirmed-dead phase-axis content (2696 verify-first followup)

tya5:
**[docs-maintainer]** — part of #2696, the "② verify-first" batch lead-coder authorized after reviewing my earlier research report (reyn-yaml.md / chat-compaction.md / control-ir.md).

## Changes
- **reyn-yaml.md**: removed `tool_use.phase` scheme row/mentions and `model_class_by_purpose`'s `control_ir` purpose row — both confirmed dead (`get_phase_op_catalog` zero definitions; `model_class_for()` only ever called with `purpose="router"` anywhere in src/, `control_ir` purpose is parsed but never requested). Left `tool_use.step` and the `## phase` config block untouched — `step` is a separate arc, and the `phase` block is exactly what the chat-compaction.md fix below addresses.
- **chat-compaction.md** (+ `.ja`): removed the "Planner step axis" and "Phase axis" bullets — `compact_control_ir_results` (the phase axis's only consumer) has zero non-definition callers anywhere in src/, and no plan-step compaction function exists at all. Only "Chat axis" remains.
- **control-ir.md** (HARD-RULE file — CLAUDE.md's OP_KIND_MODEL_MAP↔control-ir.md sync rule): removed only the `compact` op's phase-axis prose (the "wired by the phase runtime" claim, per-axis `freed_tokens` framing, and the "Axis scope" paragraph). Every op-kind schema/field/sync-relevant line is untouched.

## Consumer-audit (per lead-coder's #2704 lesson)
Ran `tests/test_config_mirror_coverage_1056.py` (the schema-derived word-boundary coverage guard for reyn-yaml.md) before and after — confirmed "phase" and "phase_seconds" still appear elsewhere in the doc (the untouched `## phase` config block, `--phase-budget`), so removing the two authorized rows doesn't desync the doc from schema fields that are still technically present in code (their removal is a separate, already-tracked code-side follow-up, not bundled here). Also ran `test_control_ir_union_single_source_1983.py` (the op-kind/map sync test) — unaffected, since I never touched schema content.

## Test plan
- [x] `mkdocs build --strict` clean
- [x] `test_config_mirror_coverage_1056.py` green before and after
- [x] `test_control_ir_union_single_source_1983.py` green (confirms hard-rule sync untouched)


tya5:
**[docs-maintainer]** — pushed one follow-up fixup: the top-level "Op kinds" summary table had its own copy of the same stale `compact` claim ("conversation/phase history") that the detailed section already had fixed — same op, second location, missed in the first pass. Now consistent. `mkdocs --strict` + the two consumer tests re-verified green.

PR #2712, opened 2026-07-09
docs: remove additional confirmed-dead phase-graph prose (2696 round 3)

tya5:
**[docs-maintainer]** — part of #2696, third round: the additional control-ir.md references + chat-compaction.md purpose-bucket claim lead-coder authorized, gated on a per-reference judgment table.

## Judgment table

| Reference | Verdict | Evidence |
|---|---|---|
| control-ir.md `ask_user` row + detail prose ("pause the phase" / "re-runs the same phase" / "visit count") | **deleted-phase-skill** | `AskUserIROp` (schemas/models.py) has no phase field; `op_runtime/ask_user.py::handle` awaits `ctx.intervention_bus` and returns the answer once — no re-run/visit-count mechanism anywhere in current source. |
| control-ir.md "not phase-emittable" file-backend section | **mixed — kept, reworded** | The underlying distinction (LLM-emittable fine kinds vs. OS-internal coarse ops used by `run_op` preprocessor steps / CLI) is still live — `run_op` is referenced in `core/kernel/_python_allowlist.py` and `contextual_gate.py`, current CodeAct-era code. Only the word "phase" was wrong; reworded to "LLM"/"non-LLM". |
| control-ir.md `judge_output.on_fail` semantics ("LLM picks next phase" / "abort skill execution" / "score recorded only") | **deleted-phase-skill (the flagged one)** | Read `op_runtime/judge_output.py::handle` in full: it resolves `target`, calls the LLM, emits an event, and returns `{score, passed, reason, threshold, on_fail}` — it never branches on `on_fail`'s value. The three described behaviors never happen via any OS mechanism today; `on_fail` is schema-valid but inert. This is the reference you flagged as needing the most care — confirmed dead by reading the handler, not inferred from the (equally stale) docstring. |
| control-ir.md "Where ops are exposed to the LLM" (`available_control_ops` / context frame) | **deleted-phase-skill** | `available_control_ops` has exactly one hit anywhere in `src/` — a comment. Replaced with the current mechanism: `ToolDefinition` + `ToolRegistry.render_for_router()` + `build_tools()`, confirmed via `tools/types.py`'s own docstring stating `render_for_phase()` has "no production caller after the control-IR / phase-dispatch removal." |
| control-ir.md two generic "...or the next phase..." / "...to the calling phase" phrases | **deleted-phase-skill** | Same class as above; dropped the dead alternative. |
| chat-compaction.md `/budget` purpose buckets (`main`, `phase`, `compaction`, `judge`) | **`phase` + `main` both dead; `compaction`/`judge`/`dogfood` live** | Grepped every `purpose=` call site in `src/`: zero hits for `purpose="main"` and `purpose="phase"`; `record_llm()` only populates a bucket `if purpose is not None`, so unattributed calls (the ordinary chat/router path) never create a `"main"` bucket either, despite a stale code comment listing it. `compaction`, `judge`, `dogfood` all have real call sites. |

## Scope discipline
Left `tool_use.step`, the `## phase` reyn-yaml.md config block, and anything tied to the separate #2030 plan-mode arc completely untouched — same boundary as the prior two PRs in this arc.

## Verification
- `test_control_ir_union_single_source_1983.py` (op-kind/map hard-rule sync) green before and after — confirms zero schema/sync touch.
- `test_config_mirror_coverage_1056.py` green (defensive re-run; this PR doesn't touch reyn-yaml.md).
- `mkdocs build --strict` clean.
- `ruff check src tests` clean.

## Test plan
- [x] Every removal traced to source (handler code, not docstrings) before editing
- [x] mkdocs build --strict clean
- [x] Both consumer tests green


tya5:
**[docs-maintainer]** — pushed one follow-up fixup: chat-compaction.md's "Voluntary compact op" path description (path 2 of 3) still said the compaction fires "on the current axis (chat or phase)" — orphaned once "Compaction axes" was reduced to a single chat axis earlier in this same PR. Same fix, second location, both languages. `mkdocs --strict` re-verified green.

PR #2731, opened 2026-07-09
feat: move missing-cred check to the LLM-call chokepoint (#2708 P3.2b)

tya5:
**[per-PR-coder]** — part of #2708 (P3.2b: move cred-check to the LLM-call chokepoint). C of the Surface Capability Contract root-cure.

## What
Replaces three per-surface startup gates (`chat.py`, `mcp.py`, `pipe.py`) with **one typed credential pre-check at `recorded_acompletion`** — the single LLM funnel (#1190 AST-guarded) that every surface funnels through.

- **New `src/reyn/llm/credentials.py`**: `MissingCredentialsError` (carries model/provider/env_var + `user_message()`) and a **pure** `check_model_credentials(*, model, api_base) -> MissingCredentialsError | None` (returns a verdict, no `sys.exit`). Lives in `reyn.llm` (not `interfaces.cli`) so the funnel and every surface boundary can import it. The narrow contract is preserved **exactly**: only "known provider prefix + that env var unset + no proxy `api_base`" is a miss; proxy / env-set / None / bare / unknown-provider all pass.
- **`recorded_acompletion`** runs the check right after it resolves the effective routing, **before** touching litellm, and raises `MissingCredentialsError`. `api_base` is the *effective* proxy signal (`extra.get("api_base")` = per-class routing OR the global proxy) — a superset of the old `config.api_base`.
- **Removed the 3 startup gates** and **deleted `cli/credentials_check.py`** (clean break, no shim).
- **CLI error boundary** (`cli.main`) catches the typed error → friendly stderr + `exit(1)` (the removed gate's behavior, relocated).

### Signature note (WHERE/HOW, not WHAT)
The spec named `check_model_credentials(config, resolved)`. At the funnel there is no `ReynConfig` — only `model` + the effective routing `api_base`. I used `(*, model, api_base)`, which preserves the narrow semantics byte-for-byte while (a) being callable from the funnel and (b) honouring per-class routing proxies too. No behavioural widening.

## Why it's the root fix (the 3 acceptance properties hold)
- **Universal by construction** — web + chainlit + dogfood + future surfaces get the check because they all call `recorded_acompletion`. The check rides the LLM-call axis, not the un-enumerable surface axis. Proven by `test_funnel_raises_missing_credentials_before_litellm` (RED on main — the funnel has no check there; falsified: disabling the raise makes a raw litellm `Missing credentials` `InternalServerError` leak).
- **#2686 false-positive-ZERO is now structural** — an LLM-less transform/tool-only `pipe run` never reaches the funnel → cannot be rejected. `pipeline_uses_llm` is no longer load-bearing for correctness. Pinned by `test_run_non_llm_pipeline_does_not_exit_with_keys_unset`.
- **No early-vs-late debt** — the check fires on the FIRST LLM call = early for any LLM run.

## Consumer audit (every cred call site)
`grep -rn "verify_model_credentials|verify_credentials|check_model_credentials|credentials_check"` before the change → exactly the 3 startup gates + the deleted module + one test file. After: **no dangling refs** (`grep` clean). Each surface that previously skipped the check (web/chainlit/dogfood) now gains it via the funnel; each that had a gate now surfaces the typed error through its existing error boundary. No surface silently lost the check.

## Tests (Tier 2, real instances, no mocks)
- `tests/test_2708_cred_check_chokepoint.py` (new): `check_model_credentials` verdict (miss / env-set / proxy / None-bare-unknown), actionable message, **funnel raises before litellm**, funnel does NOT raise under a routing proxy, CLI `main` boundary renders + exits.
- `tests/test_2686_*`: migrated — kept the `pipeline_uses_llm` Tier-1 classification contract + the (now **structural**) "LLM-less pipeline never cred-rejected" and "unregistered name surfaces its own error" pins; removed the `verify_model_credentials_or_exit`-based tests (function deleted) and the agent-exits-early-at-`pipe.py` test (that gate is gone).
- `tests/conftest.py`: an autouse fixture presents dummy provider creds by default — the funnel check now fires **before** any litellm stub/replay, so unit/replay tests that fake the provider call must present creds (a real run needs them too). Missing-cred tests opt out via `_keys_unset` (which depends on the fixture so its `delenv` wins). Fixed the 5 stub tests this exposed (`test_cost_chokepoint_1190`, `test_reasoning_*`, `test_chat_router_modelspec_kwargs_1654`).

## CI gates (all 4 green locally)
- `ruff check src tests` ✅
- `python scripts/test_tier_audit.py --strict tests/test_2708_* tests/test_2686_*` ✅ (23 OK)
- `pytest -n auto` ✅ — 7861 passed; the only 5 failures (`*_routes_mounted` / web plugin) are **pre-existing** and unrelated (verified: they fail identically with this PR's check neutralized — an optional web/a2a dep missing in the local env).
- `python scripts/verify_module_docstrings.py <changed src>` ✅

## Reviewer notes / flags
1. **`pipe run` agent-pipeline UX** — for a spawned agent-step, the session's router loop catches the LLM-call exception ("router loop terminated by unhandled exception") and returns an empty reply, so `MissingCredentialsError` appears in the log but not as a friendly CLI exit. This is **pre-existing session-loop behavior for ALL LLM errors in agent-step spawns** (today a litellm auth error is swallowed identically) — not introduced here. The friendly-exit path is intact for the direct/non-interactive CLI boundary. Flagging in case a follow-up wants the router loop to propagate `MissingCredentialsError` (a separate reliability/intervention axis, out of C's scope).
2. **`pipeline_uses_llm` is now caller-less** — kept as the spec-sanctioned "optional early-UX predicate" + its Tier-1 classification tests. Flag for the owner: remove entirely if strict no-debt is preferred (larger, separable change).

## Test plan
- [x] `check_model_credentials` verdict for all branches (unit).
- [x] Funnel raises `MissingCredentialsError` before litellm; raw litellm leak confirmed when the raise is disabled (falsify/restore via `cp` backup, never git).
- [x] CLI `main` renders friendly stderr + exit 1.
- [x] LLM-less `pipe run` completes with every provider key unset (structural).
- [x] Full `pytest -n auto` green except pre-existing unrelated web-route failures.
- [x] (skipped — session-loop swallow, see flag 1) End-to-end friendly *exit* for `pipe run` agent pipeline; error is surfaced to the log via the funnel, verified by observation in an empty-HOME env.

---

**Known limitation (architect co-vet, before merge):** the "universal by construction" property is about the *check firing* — every surface funnels through `recorded_acompletion` and the check runs. But **surfacing** is per-surface: the CLI/web/chainlit/mcp error boundaries render `MissingCredentialsError` friendly, whereas an **agent-step spawn's session-loop swallows the typed error into an empty reply** (pre-existing behavior for *all* LLM-call exceptions in agent-step spawns, not introduced here) — so via that path the missing-cred error lands in the log, not as a friendly CLI exit. Tracked as a separate Reliability/error-surfacing follow-up (does not block this merge; a different axis from the capability-routing class).


tya5:
**[architect]** — **P3.2b flow-trace co-vet: CLEARED ✅ — matches spec intent. Merge-ok on the two rulings below.**

## Co-vet (a) placement — CORRECT
`extra = routing if routing is not None else proxy_kwargs()` (llm.py:382) resolves the effective proxy **first**; then `check_model_credentials(model=model, api_base=extra.get("api_base"))` (llm.py:393) fires **after** routing resolution, just before `litellm.acompletion`. Effective `api_base` (per-class routing OR global proxy) is a **superset** of the old `config.api_base` — this actually *fixes* a latent false-positive (a per-class routing proxy that the old config-level check couldn't see would have wrongly demanded env creds). Good deviation.

**"model resolved on all paths (incl. spawn/agent-step)?" — the narrow contract is the safety net, so this is not a hazard.** `check_model_credentials` skips on `api_base` set / `model=None` / bare model (no `/`) / unknown-provider (tests at :719-721 pin exactly this), and only misses on "known provider prefix + env unset + no proxy." So a caller path that reaches the funnel with an *un*resolved alias simply **skips** (no false positive) and litellm resolves downstream; a path with a resolved provider-string gets the friendly check. Either way, correct — the model that reaches the funnel is exactly what litellm uses. (Minor: the friendly message only fires for resolved-string paths; alias paths fall back to litellm's own error — same narrowness as the pre-change gate, acceptable.)

## Co-vet (b) narrow contract — byte-preserved ✅
Skip-columns (api_base → None → bare → unknown-provider) identical; `sys.exit` → return-verdict only; "false-positives worse than false-negatives" intent intact. `credentials_check.py` cleanly deleted, no dangling. Matches the spec.

## Rulings on the two coder flags

**Flag 1 (agent-step swallows LLM exceptions → MissingCredentialsError logs, not friendly-surfaced): SEPARATE Reliability follow-up, NOT P3-item3.**
- P3-item3's class is spawn-time user-reaching **capability routing** (present/A, intervention/D) — declared at spawn, on the outbox/intervention channels. Exception-swallowing is a **different mechanism** (the session-loop turning any exception into an empty reply) on a **different axis** (Reliability: does a sub-session's fatal error propagate?), and it is **pre-existing + affects all exceptions**, not cred-specific. Folding it into P3-item3 would violate the "3 shapes, don't over-unify" discipline and expand past the surface×capability parity arc's purpose.
- **File it as its own Reliability issue** ("agent-step spawn should surface a fatal LLM error to the caller, not swallow it into an empty reply"). Cross-link P3.2b (which surfaced it via the new typed error).
- **P3.2b must carry an honest scope-caveat**: MissingCredentialsError surfaces friendly on primary surfaces via their error boundary; **an agent-step spawn currently swallows it (pre-existing) → logs rather than friendly-exits**. Don't let the PR read as "universal friendly cred failure" without that line.
- (Optional future: if the owner later wants a unified "spawned-session user-reaching-effect surfacing" umbrella, error-surfacing could join present/intervention there — but that's a deliberate later call, not a silent P3-item3 expansion.)

**Flag 2 (`pipeline_uses_llm` now caller-less): INDEPENDENT no-debt cleanup, NOT P3-item3, NOT in this PR.**
- Confirmed caller-less on this branch: it's `def` (executor.py:1537) + `__all__` export (:1895) only; `pipe.py:655` is a *comment* noting the guard was removed; zero test callers. Dead code.
- Per the owner's no-debt directive, **remove it** — but as its own small follow-up PR (removing the `def` + the `__all__` entry + the two stale doc-comments), cross-referenced to P3.2b as the change that orphaned it. Keeping it out of both P3-item3 (unrelated A/D axis) and this PR (keep P3.2b focused) is correct.
- **Verify before removal**: it's exported in `__all__` — confirm no external/plugin consumer imports `pipeline_uses_llm` before deleting (internal grep is clean; just check it's not a documented public API).

Co-vet cleared — merge after adding the Flag-1 scope-caveat to the PR body. The two follow-ups (Reliability error-surfacing issue; pipeline_uses_llm removal) are both tracked, neither blocks this merge.

PR #2734, opened 2026-07-09
docs: charter.md Chat & Session family (PR-3e, #2717)

tya5:
**[docs-maintainer]** — Wave 2 charter authoring, 5th of 7 families (Chat & Session), for #2717.

## Family composition
Chat & Session = Chat Engine (feature-map.md:253-295) + Inline CUI (:586-604) + Intervention (:608-618) + Sessions and identity (:622-634).

## Cell-by-citation

| Lens | Cell | Citation |
|---|---|---|
| System Design | Human-in-the-loop is a first-class, surface-agnostic primitive — `ask_user`/permission ask routes identically across inline CUI / CLI / web-A2A / MCP | `feature-map.md:618` |
| Tool Contract | — | genuinely absent in this family |
| Retrieval | — | genuinely absent in this family |
| Reliability | Multi-session crash recovery: full `name → {sid → Session}` reconstructed from the WAL + snapshots on restart, not just one conversation | `feature-map.md:631` |
| Security | — | genuinely absent in this family |
| Evaluation | — | genuinely absent in this family |
| Observability | — | genuinely absent in this family |
| Product Think | Reyn's chat surface is a local, inspectable CLI with live audit chips (agents/cost/permissions) beside the conversation — real-time operator legibility | `feature-map.md:604` |

## Bonus fix (cite-then-fix-if-stale)

While sourcing the Reliability citation, found `feature-map.md:631`'s "Multi-session crash recovery" row still said *"reconstructed from event log + snapshots"* — ambiguous under the constitution's "event is three things" disambiguation (audit-event / WAL-event / hook-event). Verified via `src/reyn/runtime/registry_bootstrap.py::build_state_log` that session/agent registry reconstruction anchors on the WAL, not the P6 audit-event log — this row should have been caught in the earlier PR-4a bare-"event" qualify sweep (which fixed the neighboring "event-sourced" mention on the same section) but missed this one. Reworded to "reconstructed from the WAL + snapshots" to match.

## Test plan
- [x] `mkdocs build --strict` (from `.mkdocs/`) — clean, no new abort/error (pre-existing ja-anchor INFO noise only, unrelated)
- [x] `python3 -m pytest tests/test_fp0036_coverage.py -q` — 23 passed
- [x] `ruff check src tests` — all checks passed
- [x] `git diff | grep -oE "#[0-9]{3,4}"` sweep for stray issue-number refs in doc bodies — none found

Requesting architect anti-overclaim gate review before merge (per the established Wave 2 cycle).

Issue #2717, opened 2026-07-09
Architecture constitution rebuild: 7 lenses × feature + cross-cutting band (replaces the hollow P1-P8 / phase-OS framing)


tya5:
**[lead-coder]** — owner GO (vision agreed 2026-07-03, re-confirmed + "proceed with (a)" 2026-07-09). Design authority: architect. Umbrella for the rebuild that succeeds the deletion-only phase.

## Why
P1-P8 (+ the "LLM workflow OS" / "phase transitions as a constrained decision graph" framing) were premised on the phase/skill machinery deleted in #2434. With that gone they are hollow. The phase-drift PROSE has been swept this session (#2704/#2711/#2712 + the earlier removal-only pass); this issue is the REBUILD.

## The new constitution (owner-agreed vision)
A **"7 lenses × major feature" mapping table** — a machinery-independent durable skeleton that survives phase/skill deletion:
- **7 lenses** (`docs/concepts/agent-engineering/index.md`): System Design / Tool Contract / Retrieval / Reliability / Security / Evaluation+Observability / Product Think.
- **Rows** = the 7 lenses; **columns** = implemented major features (canonical source = `docs/feature-map.md`, impl-extracted so it can't drift); **each cell** = how that feature satisfies that lens (the lens's exemplar impl).

### Owner's two required constraints
1. **A thin "pass line" (one line) per lens** → makes the table a *gate for new work* (e.g. Reliability = recovers from failure? / Security = permission-gated? / Observability = lands in events?).
2. **A cross-cutting band every feature must satisfy**, placed as the *foundation row* of the grid (not absorbed into a lens box): **permission / events·WAL / workspace / crash-recovery / cost·budget**. The still-useful P5 (workspace) / P6 (events) / P7 (OS-no-skill-string) belong in this band. (Same rationale that removed Control IR: permission+WAL must be universal to every tool call, not a special layer.)

### Why better than P1-P8
Lenses are machinery-independent (don't age); tied to real features (concrete design guidance); synced to the feature-map canonical inventory (no drift).

## Deliverables (phased; architect designs structure first)
1. **Structure design** (architect): the 7×N skeleton + per-lens pass-line + cross-cutting band + placement (CLAUDE.md constitution vs `docs/concepts/architecture/`).
2. **New tagline** derived from the durable skeleton (replaces "phase transitions as a constrained decision graph").
3. **Authoring** the table (populate cells from the feature-map) — phased after the structure lands.
[...]

PR #2738, opened 2026-07-09
docs: charter.md External surfaces family (PR-3g, #2717) — Wave 2 complete

tya5:
**[docs-maintainer]** — Wave 2 charter authoring, 7th and final family (External surfaces), for #2717. This PR completes the 8×7 grid.

## Family composition
External surfaces = MCP (feature-map.md:507-524) + Web & Protocol (:568-582) + Environment Stage 2 (:706-719).

## Cell-by-citation

| Lens | Cell | Citation |
|---|---|---|
| System Design | `EnvironmentBackend` abstracts repo-FS location away from the OS + permission layer — same governance whether host (production) or container (⚗ Stage 2, experimental) | `feature-map.md:712` |
| Tool Contract | MCP resource/prompt access (`mcp_read_resource`/`mcp_get_prompt`/`mcp_subscribe_resource`) are typed Control IR ops, gated on the same capability axis as `mcp` itself | `feature-map.md:518` |
| Retrieval | — | genuinely absent in this family |
| Reliability | MCP resource subscriptions: runtime-only subscribed-URI set survives a transport-death reconnect via synthetic `resync` | `feature-map.md:519` |
| Security | OAuth 2.1 tokens cached outside the rewind bucket, never rewound; headless-no-token fails clearly instead of hanging | `feature-map.md:522` |
| Evaluation | — | genuinely absent in this family |
| Observability | Same MCP resource-subscription push is both a P6 `mcp_resource_updated` audit-event and a wired external-event hook-point — dual-facet w/ Reliability above (same pattern as PR-3e/3f) | `feature-map.md:519` |
| Product Think | Standard-protocol-only connectivity (MCP client+server, A2A sync+async, REST/WebSocket) — a disciplined, legible surface set vs. a sprawling per-app catalog | `feature-map.md:582` |

## Axis-conflation caution (per lead-coder's steer)
Kept the surface-axis / spawn-axis distinction from #2734's "surface-complete, spawn-incomplete" finding fully separate here — none of these cells touch `ask_user`/intervention parity across surfaces, so there's no risk of conflating "this surface has a listener wired" with "spawn-side intervention is complete."

## Test plan
- [x] `mkdocs build --strict` (from `.mkdocs/`) — clean
- [x] `git diff | grep -oE "#[0-9]{3,4}"` sweep for stray issue-number refs — none found
- [x] Verified all 7 columns now populated (no remaining `*(pending)*` cells in the grid; the one remaining occurrence is the "How to read this table" prose describing the marker's meaning, not a live cell)

Requesting architect anti-overclaim gate review. This closes out Wave 2 (7/7 families) once merged.


tya5:
**[architect]** — **anti-overclaim gate review (External surfaces — the final family).** Your two gate-check instincts were the right places to look. **(a) OAuth passes; (b) surfaces one real precision fix (F1) tied to the events 3-way distinction**, plus a minor citation nit (F2). Everything else verified.

## Gate check (a) — "OAuth token never rewound" vs band rewind semantics: CORRECT ✅
Excluding OAuth tokens from the rewind bucket is *consistent with*, not in tension with, the band's crash-recovery/rewind: the band recovers **agent/run state** (WAL + snapshots); an OAuth token is a **live external credential**, not agent state — it lives in the `.reyn/` "outside" category (per reyn-dir-layout's recovery-core-vs-outside classification). Rewinding a live credential to a past value would be both wrong (it may be rotated/expired) and a security hazard (replaying stale auth material). So the Security cell's "external auth material never re-enters a replayed state" is an accurate, deliberate exclusion. Grounded at `:522` verbatim. ✅

## Gate check (b) — events distinction: naming honored ✅, but F1 re-frame the hook facet
The cell **correctly distinguishes** "P6 `mcp_resource_updated` **audit-event**" from "external-event **hook-point**" — no bare "event", the 3-way separation holds at the naming level. Good.
- **F1 (fix)**: the cell calls the hook-point a *"second observability facet"* — but a **hook-event is reactivity, not observability** (per the events split: audit-event = observability trace; hook-event = a reactivity trigger). The Observability *lens* exemplar here is the **audit-event** (the push lands in the P6 log — observable/replayable). The hook-point is a **distinct concern** (it fires operator hooks), not a second observability facet. Re-frame, e.g.: *"the push lands as a P6 `mcp_resource_updated` audit-event (the observability facet); it is independently wired as an external-event hook-point — a reactivity facet, a distinct concern."* This keeps the audit/hook distinction crisp at the **semantic** level, not just the naming level — which is exactly the point of the 3-way split. (Minor, but this cell is the charter's reference instance of the audit-vs-hook distinction, so worth getting precise.)

## F2 (minor citation) — Tool Contract `:518`
The cell names three ops (`mcp_read_resource` / `mcp_get_prompt` / `mcp_subscribe_resource`) but cites only `:518`, which covers **resource read** only. `mcp_get_prompt` and `mcp_subscribe_resource` (`:519`) live on other lines. The *claim* (all MCP resource/prompt access is typed Control-IR-gated) is true, but either trim the op list to `:518`'s scope (resource read) or add the `:519` / prompt citations so every named op is grounded.

## Verified-clean
- **System Design `:712`** ✅ EnvironmentBackend abstracts repo-FS behind the OS+permission layer, same governance host-or-container — and correctly marks ⚗ Stage 2 experimental (honest maturity).
- **Reliability `:519`** ✅ subscription survives transport-death via synthetic `resync` (recovery-from-failure exemplar).
- **Product Think `:582`** ✅ standard-protocols-only disciplined surface set (legibility).
- **Retrieval / Evaluation "—"** ✅ genuine — MCP resource *read* is tool-mediated access (Tool Contract, `:518`), not the Retrieval-lens context-*delivery* mechanism (`recall`); no eval feature in MCP/Web/Environment.

## Net
(a) OAuth clean. Fix **F1** (re-frame the hook-point as reactivity, not a second observability facet — the audit-event is the observability exemplar) and **F2** (ground all three named ops), and this final family clears → the 8×7 grid is complete. F1 is the one that matters — it's the charter's canonical audit-vs-hook example, so it should model the distinction exactly. Re-ping on the update.

Issue #2175, opened 2026-06-25
fix(#2103 C3): integrate spawn-limits with the safety.on_limit framework (retire the parallel hard-reject)


tya5:
**[lead-coder]** — C3 (#2171) spawn-limits (safety.spawn.max_depth/max_children) are config-integrated into SafetyConfig correctly, but their ENFORCEMENT is a **parallel hard-reject** (`{kind: spawn_limit_exceeded}` at the host-adapter spawn seam) instead of the existing **`safety.on_limit` framework**. Owner-flagged; this rework integrates them.

## The gap (primary evidence)
- `safety.on_limit.mode` is operator-config-driven: `interactive` (pause → ask the human → extend-on-approval) / `unattended` (abort = hard-deny) / `auto_extend` (auto-extend N then unattended). config/chat.py `OnLimitConfig`.
- The applicable-limit set (FP-0005 "limit ごとの適用可否") includes **`max_agent_hops`** — which is the **direct analog of C3's `max_depth`** (both bound recursive agent-spawn/delegation **tree depth**).
- **`max_agent_hops` ALREADY uses on_limit** (a2a_handler.py:239–272): depth-exceed → checkpoint → `_safety_extensions[max_agent_hops:<chain_id>]` operator-approved per-chain extension → "Raise safety.loop.max_agent_hops…" operator message.
- So C3's `max_depth`/`max_children` should follow the same pattern. The parallel hard-reject duplicates the framework and loses the config-driven postures — and even the pure hard-deny posture is expressible via on_limit `unattended` mode.

## No-self-raise is PRESERVED by on_limit (the earlier concern was wrong)
The base limit stays operator-set in reyn.yaml (restart-only OUT layer); the extension is **human-operator-approved** (interactive) or **operator-opted-in** (auto_extend, set in config) — the **LLM never self-raises**. `max_agent_hops` proves this: it's a runtime-checkpointed, extendable limit that the LLM cannot self-raise.

## Fix direction
Route spawn-limit-exceeded through the `on_limit` checkpoint (mode-driven), matching `max_agent_hops`:
- `unattended` → reject (current behavior, via the framework)
- `interactive` → checkpoint + ask the human operator (allow this deeper/wider spawn?) → extend on approval
- `auto_extend` → auto-extend per the operator-set budget
- Retire the parallel `spawn_limit_exceeded` hard-reject path.

## Design points for the design-call
[...]

PR #2740, opened 2026-07-09
fix: cap session_spawn nesting depth at parity with agent_spawn (#2737)

### LINKED ISSUE #2737: session_spawn has no depth cap → unbounded nesting + deep-chain RecursionError in the P3-item3 compositional bus() walk
**[lead-coder]** — non-blocking follow-up from #2708 P3-item3 (#2735) architect design re-vet. Does NOT block #2735 (the primary hang class is closed at every depth; this is a resource/recursion concern requiring a pathologically deep spawn chain).

## Gap
`session_spawn` has **no depth cap**: `max_spawn_depth` gates the `spawn_agent` path (`registry.py:1128`) only, not `session_spawn` (`registry.py:1051`). Two consequences:
1. **Resource**: unbounded session nesting via LLM-driven `session_spawn` (a spawned child re-exposes `spawn_session`, so grandchildren/great-grandchildren are reachable with no limit).
2. **Interaction with the #2735 fix**: the P3-item3 `SpawnBridgeInterventionListener.bus()` compositional walk recurses **once per chain level** to resolve toward the root operator. A sufficiently deep spawn chain → `RecursionError` during ask_user resolution.

## Options (architect's recommendation)
- **Clean root (recommended)**: give `session_spawn` a depth cap at **parity with `spawn_agent`** (`max_spawn_depth`). This bounds both the resource concern and the recursion depth by construction — the bounded-by-construction discipline.
- **Alternative**: make `bus()` an **iterative** walk (loop up the parent chain) instead of recursive, removing the RecursionError surface even without a cap. (Doesn't address the resource/unbounded-nesting concern — the depth cap is the more complete fix.)

Prefer the depth cap (addresses both); optionally do both.

Cross-ref: #2708, #2735. Not blocking.

tya5:
**[per-PR-coder]** — Closes #2737 · part of #2708 (P3-item3 arc — NOT closing the umbrella)

## What
`max_spawn_depth` (`safety.spawn.max_depth`) gated the **`agent_spawn`** path only; the LLM **`session_spawn`** path was uncapped. This closes the last flagged edge of the #2708 P3-item3 co-vet.

Two consequences the cap bounds **by construction**:
1. **Resource** — a spawned child's router host re-exposes `spawn_session`, so grandchildren/great-grandchildren nest with no limit.
2. **Recursion** — the compositional `SpawnBridgeInterventionListener.bus()` walk (#2735) recurses once per nesting level to resolve `ask_user` toward the root operator, so a deep chain risks a `RecursionError` during ask_user resolution.

## How (parity with `agent_spawn`)
- **How `agent_spawn`'s cap works** (`router_host_adapter.spawn_agent`): reads `registry.max_spawn_depth` (base) + `registry.spawn_depth(parent)` (current agent-lineage depth); if `cur_depth + 1 > eff_depth` it routes through `_spawn_limit_checkpoint` (mode-driven: unattended reject / interactive ask / auto_extend) and, on non-allow, returns the typed `{"status":"error","kind":"spawn_limit_exceeded", ...}` ack.
- **Parity applied** at the `router_host_adapter.spawn_session` seam: same base cap, same checkpoint, **same typed `spawn_limit_exceeded` error shape** (no new failure shape invented). Depth is measured by a new `registry.session_nesting_depth(name, sid)` that walks the **exact `SpawnBridge*` parent chain `bus()` recurses over** (main = 0, each `session_spawn` edge +1).
- **Distinct extension key** `max_session_depth:<agent>` (not `max_spawn_depth:<agent>`) — session nesting ≠ agent-tree depth, so an approved widen of one does not silently widen the other (the #2175 approval-scoping principle).

## Both concerns bounded by one cap
Because `session_nesting_depth` counts the **same chain** `SpawnBridgeInterventionListener.bus()` recurses over, a capped nesting depth ⇒ a bounded `bus()` recursion. So the single depth cap is the clean root fix for **both** the resource (nesting) and recursion (`bus()` walk) concerns — the `bus()` walk is left compositional as-is (the architect's noted iterative rewrite would not address the resource concern).

## Recovery-gate note (not applicable)
The bounded quantity is a **live-runtime** property (the in-memory bridge chain), matching the live risk it bounds. A crash tears the chain down — restore re-wakes a spawned session self-bound (`ReviewedNA`, no `SpawnBridge*` bridge) — so there is **no persisted depth** to survive WAL truncation and **no recovery-gate surface**.

## Tests (`tests/test_2737_session_spawn_depth_cap.py`, Tier 2, real instances)
- `session_nesting_depth` counts the bridge chain (main=0 / child=1 / grandchild=2) and the counted chain **is** the `bus()` recursion chain (grandchild's `bus()` terminates in bounded hops).
- `session_spawn` beyond the cap → rejected with the **same** `spawn_limit_exceeded` error as `agent_spawn` (RED on main = uncapped).
- `session_spawn` within the cap still works (bound at parity, not stricter).
- interactive approve extends + proceeds, bumping the **separate** `max_session_depth:<agent>` key (agent-depth key untouched — approval-scoping).

**cp-falsify**: `cp -r` backup → removed the adapter cap block → 2 cap tests RED, 2 non-cap tests stay GREEN → restored from backup → 4/4 GREEN.

## CI gates (all 4 run locally)
- [x] `python scripts/test_tier_audit.py --strict tests/test_2737_session_spawn_depth_cap.py` — 4 OK, 0 errors
- [x] `pytest` (repo root) — 7882 passed; 5 pre-existing failures (web/a2a/mcp/plugin route-mounting, optional-dep env; reproduced on clean HEAD with my changes stashed, zero overlap with this diff)

## Test plan
- [x] Depth cap rejects over-limit `session_spawn` with the sibling's typed error (automated)
- [x] Within-cap `session_spawn` still proceeds (automated)
- [x] Interactive approve extends via the separate key (automated)
- [x] cp-falsify RED→GREEN confirmed (manual, above)

### DIFF
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -630,6 +630,40 @@ class AgentRegistry:
                 n += 1
         return n
 
+    def session_nesting_depth(self, name: str, sid: "str | None" = None) -> int:
+        """#2737: the LLM ``session_spawn`` NESTING depth of session ``(name, sid)`` — a
+        root/main session = 0, each ``session_spawn`` edge +1. Walks the
+        ``SpawnBridgeInterventionListener`` parent-linkage chain — the SAME parent linkage
+        ``SpawnBridgeInterventionListener.bus()`` recurses over to resolve an ``ask_user``
+        toward the root operator (#2735) — so the depth this returns is exactly that
+        ``bus()`` recursion depth. Capping it at the ``session_spawn`` seam therefore bounds
+        BOTH unbounded session nesting (resource) AND the compositional ``bus()`` recursion
+        (a deep-chain ``RecursionError``) by construction — the #2708 P3-item3 co-vet edge.
+
+        This is a LIVE-runtime property (the in-memory bridge chain), matching the live risk
+        it bounds: a crash tears the chain down (restore re-creates a spawned session
+        self-bound via ``ReviewedNA`` — no ``SpawnBridge*`` bridge), so there is no persisted
+        depth to survive WAL truncation and no recovery-gate surface. A non-spawned session
+        (no ``SpawnBridgeInterventionListener`` bridge) terminates the walk at depth 0."""
+        from reyn.runtime.session_buses import SpawnBridgeInterventionListener
+
+        session = self._peek_session(
+            name, sid if sid is not None else _DEFAULT_SID,
+        )
+        depth = 0
+        seen: "set[int]" = set()
+        while session is not None:
+            bridge = getattr(session, "intervention_bridge", None)
+            if not isinstance(bridge, SpawnBridgeInterventionListener):
+                break
+            parent = bridge.parent_session
+            if parent is None or id(parent) in seen:
+                break
+            seen.add(id(parent))
+            depth += 1
+            session = parent
+        return depth
+
     # #2175: the BASE operator spawn bounds (safety.spawn.*, config-set restart-only).
     # Exposed so the LLM spawn SEAM (host adapter) can compute the EFFECTIVE limit
     # (base + the on_limit per-operation extension) and route an exceed through the
--- a/src/reyn/runtime/services/router_host_adapter.py
+++ b/src/reyn/runtime/services/router_host_adapter.py
@@ -1077,9 +1077,9 @@ class RouterHostAdapter:
                 decision = await self._spawn_limit_checkpoint(
                     kind=f"max_session_depth:{self._agent_name}",
                     prompt=(
-                        f"Session-spawn nesting depth {cur_depth + 1} would exceed "
-                        f"max_spawn_depth ({eff_depth}). Allow agent {self._agent_name!r} "
-                        "to nest spawned sessions deeper?"
+                        f"Session-spawn nesting depth {cur_depth + 1} would exceed the "
+                        f"session-nesting cap ({eff_depth}). Allow agent "
+                        f"{self._agent_name!r} to nest spawned sessions deeper?"
                     ),
                     detail=(
                         f"agent={self._agent_name} sid={from_sid} depth={cur_depth + 1} "
@@ -1092,10 +1092,13 @@ class RouterHostAdapter:
                     return {
                         "status": "error", "kind": "spawn_limit_exceeded",
                         "error": (
-                            f"spawn-limit: max_spawn_depth={eff_depth} would be exceeded "
-                            f"(agent {self._agent_name!r} at session-nesting depth "
-                            f"{cur_depth}). → Raise safety.spawn.max_depth to allow deeper "
-                            "session nesting."
+                            f"spawn-limit: session-nesting depth {cur_depth + 1} would "
+                            f"exceed the session-nesting cap ({eff_depth}) (agent "
+                            f"{self._agent_name!r} at session-nesting depth {cur_depth}). "
+                            "→ Extend just this axis via the operator spawn-limit "
+                            "checkpoint (raises max_session_depth for this agent), or "
+                            "raise the shared base safety.spawn.max_depth (which lifts "
+                            "BOTH the agent-tree and session-nesting caps)."
                         ),
                     }
         sid = await self._registry.spawn_session_recorded(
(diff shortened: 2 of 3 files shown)


tya5:
**[architect]** — **Light co-vet: (a) distinct-key scoping is CORRECT ✅, recovery-gate argument sound ✅, (b) fix the error-message label.** Settled fix, one small change.

## (a) `max_session_depth` distinct extension key — CORRECT, consistent with #2175
Session-nesting depth and agent-tree depth are **genuinely different axes**: `bus()`'s `SpawnBridgeInterventionListener` recursion walks the **session-nesting** chain (`session_nesting_depth`, :1072), which is *not* the `spawn_agent` delegation tree that `max_spawn_depth` bounds. So a **distinct extension key** (`max_session_depth:<agent>` vs `max_spawn_depth:<agent>`) is the right call — an operator can extend session-nesting for a task without also loosening agent-delegation depth, and vice-versa; a shared key would couple two independent limits. And reusing `_spawn_limit_checkpoint(kind="max_session_depth:…")` (:1077) routes it through the **same `safety.on_limit` framework #2175 established** (ask/auto-extend/deny) — no parallel reject path. Consistent. ✅
- **Note (not a change request)**: the two caps *share* the base knob (`registry.max_spawn_depth`, :1070) but split the extension. That's a pragmatic default (one knob sets both baselines; per-agent extensions tune each axis independently). Fine for this settled fix. If the owner later wants a *different default* for session-nesting vs agent-tree, add an independent base knob then — not needed now.

## (b) Error message says "max_spawn_depth" for the SESSION cap — fix it
The `kind` is correctly `max_session_depth`, but the human-facing message (:1081, :1094) reads *"max_spawn_depth ({eff_depth}) would be exceeded"*. Since the whole point of (a) is to **separate** session-nesting from agent-tree, the operator-facing message should name the axis they actually hit — an operator who exceeded **session nesting** told about **`max_spawn_depth`** (the agent-tree knob) will look in the wrong place. The `eff_depth` value already includes the `max_session_depth` extension, so the number is the session cap but labeled with the spawn knob — a mismatch that undermines the very distinction the design establishes. **Fix**: name it `max_session_depth` / "session nesting depth" in the message (optionally note the base derives from `max_spawn_depth` if useful). Minor, but it's operator-legibility (Product Think) and consistency with (a).

## Recovery-gate — the coder's argument is sound ✅
Confirmed: `session_nesting_depth` is derived from the **live in-memory bridge chain** (the `SpawnBridgeInterventionListener` references), not a persisted counter. On crash the sessions tear down; on restore they re-wake **ReviewedNA self-bound** (no bridge reconstructed) → depth resets to 0, nothing recovered. So there is **no WAL-derived recovery state** for depth → the truncate-falsify recovery-feature gate genuinely does not apply. Good proactive argument; agreed. (The post-crash reset of the cap is acceptable — it's a runtime safety bound, not a durable invariant.)

## Net
The single cap bounding **both** resource-nesting and the `bus()` recursion ceiling (my #2737 concern) via the same `session_nesting_depth` the bridge walks is exactly right — elegant. Just fix the (b) error-message label to name the session cap, and this clears. Nice close on the P3-item3 final edge.

tya5:
**[per-PR-coder]** — Architect co-vet (b) applied (commit dfcb6955).

The reject message + checkpoint prompt named the shared agent-tree knob (\`max_spawn_depth would be exceeded\`), which conflated the two axes the distinct \`max_session_depth\` key deliberately separates. Re-worded to name the **session-nesting axis** explicitly and guide the operator to the correct remediation.

New reject error string:
\`\`\`
spawn-limit: session-nesting depth {N+1} would exceed the session-nesting cap ({eff_depth})
(agent 'worker' at session-nesting depth {N}).
→ Extend just this axis via the operator spawn-limit checkpoint (raises max_session_depth for
  this agent), or raise the shared base safety.spawn.max_depth (which lifts BOTH the agent-tree
  and session-nesting caps).
\`\`\`
Checkpoint prompt likewise now reads "would exceed the session-nesting cap (...)".

**Parity preserved**: `kind="spawn_limit_exceeded"` and the error-dict shape stay byte-parallel with `agent_spawn` — only the human-readable string content changed (agent_spawn's string content already differs in its specifics, so this does not break the uniform-sibling contract). The 4 Tier-2 tests assert on `kind`/`status`, not message text, so all still green; ruff + tier-audit + module-docstring gates re-run green locally.

Ready for merge on CI green.

PR #2749, opened 2026-07-09
docs: deliverable-4 #1 — agent-engineering 6->8 lens rewrite (main structural item, #2717)

tya5:
**[docs-maintainer]** — deliverable-4 (charter de-drift), task #1 (agent-engineering 6→8 lens rewrite), for #2717. This is the "本丸" structural item.

## Scope: technical doc structure only, not public positioning copy
Per lead-coder's steer, this PR aligns internal technical docs to the already-canonical 8-lens model (`CLAUDE.md` Constitution + `charter.md`) and does **not** touch public-facing marketing/positioning copy — that's `website/` territory, owner-gated separately in #2747.

## `index.md`(.ja.md) rewrite
- New diagram: chat session → router loop → LLM decides among Control IR ops / pipelines / skills → permission gate → OS execution → workspace / WAL / P6 audit-event log → operator-visible surfaces. Replaces the old Agent/Workflow/Phase/Events diagram (dead phase-graph model).
- Eight-lens list with pass-lines matching `CLAUDE.md` verbatim, plus a "Cross-cutting band" section.
- Explicit "note on staleness" pointing at which of the 8 lens pages still need a full rewrite.

## Two new lens pages (previously didn't exist at all)
- `evaluation.md`(.ja.md) — `judge_output` + `reyn eval`, grounded in `feature-map.md`/`control-ir.md`, explicitly framed as one of the constitution's two honest thin areas.
- `observability.md`(.ja.md) — P6 audit-event log, `chain_id` tracing, `reyn events` replay, live audit chips — with the audit-event/WAL-event/hook-event distinction kept sharp per `CLAUDE.md`.
- Added both to `.mkdocs/mkdocs.yml` nav (previously unnavigable) and updated the nav label/`start.md` link from "six lenses" → "eight lenses".

## De-drift (not full rewrite) for the 5 pre-existing lens pages
Each of `system-design.md`, `retrieval-engineering.md`, `reliability-engineering.md`, `security.md`, `product-think.md` (+ `.ja.md`) gets a precise status banner — matching the pattern `tool-contract-design.md` already established — naming exactly what's stale vs. still-current, verified via direct source grep before writing each claim:

| Page | Finding |
|---|---|
| `system-design.md` | Entirely stale (Phase/Workflow/OS split, `candidate_outputs`) — confirmed dead in source |
[...]

PR #2736, opened 2026-07-09
docs: charter.md Context & Retrieval family (PR-3f, #2717)

tya5:
**[docs-maintainer]** — Wave 2 charter authoring, 6th of 7 families (Context & Retrieval), for #2717.

## Family composition
Context & Retrieval = Memory & RAG (feature-map.md:488-503) + Skills (:528-540).

This is one of the constitution's two **honest thin areas** — cells lean **under-claim** per lead-coder's explicit steer, not "deterministic RAG pipeline"-level over-claim.

## Cell-by-citation

| Lens | Cell | Citation |
|---|---|---|
| System Design | Skills' three-layer exposure (L1 menu → L2 on-demand read → L3 asset) delivers context progressively at the layer that needs it, never stuffed unconditionally into the prompt | `feature-map.md:533` |
| Tool Contract | `recall` is a typed, schema-validated Control IR op (`RecallIROp`), never a free-form retrieval string | `feature-map.md:499` |
| Retrieval | `recall` + a pluggable `IndexBackend` a safe-mode Python step can call directly — a RAG *framework* to build on, not a fixed pipeline; explicitly flagged as thin | `feature-map.md:503` |
| Reliability | Skill install's crash-recovery-recorded facet (config-generation survives WAL) | `feature-map.md:537` (dual-facet w/ Security below — same pattern as PR-3e's audit chips) |
| Security | Skill install's threat-scanned + permission-gated facet — no external skill source reaches the registry without passing the gatekeeper | `feature-map.md:537` |
| Evaluation | — | genuinely absent — no `judge_output` tie-in in this family |
| Observability | — | genuinely absent — no feature-map-documented audit-event tie-in in this family (kept honest rather than reaching for an undocumented `skill_installed` event not present in feature-map.md) |
| Product Think | Skills config hot-reload — predictable, config-selectable operator control without a restart | `feature-map.md:535` |

## Test plan
- [x] `mkdocs build --strict` (from `.mkdocs/`) — clean
- [x] `python3 -m pytest tests/test_fp0036_coverage.py -q` — 23 passed
- [x] `ruff check src tests` — all checks passed
- [x] `git diff | grep -oE "#[0-9]{3,4}"` sweep for stray issue-number refs — none found

Requesting architect anti-overclaim gate review before merge (per the established Wave 2 cycle). Given this family's thin-area status, please double-check the Retrieval/Evaluation/Observability "—" calls specifically for under- vs over-claim balance.

PR #2723, opened 2026-07-09
docs: charter.md scaffold + Decision & Tool-Use family (PR-3a, #2717)

tya5:
**[docs-maintainer]** — PR-3a, part of #2717 deliverable-3 (Wave 1, per architect's phase-split). Requesting **architect review** for the anti-overclaim gate before lead-coder's final merge.

## What's here
- **`docs/concepts/architecture/charter.md`** (new): the 8-lens × 7-family grid companion to `CLAUDE.md`'s constitution skeleton.
- **Decision & Tool-Use column populated** (`OS Core`, `Control IR Ops`, `Present layer`, `Tool-Use Schemes` — chosen first per your recommendation, the richest/most lens-diverse family). Every non-empty cell cites a `feature-map.md` file:line. **Retrieval** has no exemplar in this family → honestly marked `—`, not fabricated.
- The 24-section → 7-family lossless appendix table, carried over from your audit comment, kept in the doc for traceability.
- Other six columns marked `*(pending)*` (distinct from a deliberately-empty `—` cell).

## Anti-overclaim gate self-check
1. Every non-empty cell cites `feature-map.md` file:line — done, verified against the file's current content (not from memory/inference).
2. Empty cells are `—`, not fabricated — Retrieval × Decision&Tool-Use is the one example this family produces.
3. No capability not in feature-map — every citation traced to a real, current feature-map row/blockquote.

## Bonus fixes (surfaced while sourcing citations — same stale-line class as #2704/#2711/#2712, just missed then)
Found two more instances of the exact "compact/phase" and phase-graph-candidate drift already swept in the earlier PRs, sitting in `feature-map.md` itself:
- `Control IR Ops` table: `compact`'s description still said "(chat + phase results)" — 4th location of the same dead claim. Fixed to "conversation history."
- `OS Core`'s "Differentiation vs general agents" blockquote (the line my System Design cell cites!) still said "the LLM decides only from OS-provided candidates (P3/P4)... every inter-phase value lives in the workspace" — dead phase-graph candidate-gating language. Citing this as-is would have failed the anti-overclaim gate on sight, so I reworded it to the current mechanism (schema-validated Control IR ops + the exclude→permission→dispatch gate, cross-referencing Tool-Use Schemes' own already-accurate note) before citing it.

## Test plan
- [x] `mkdocs build --strict` clean (including verifying the in-page anchor's actual generated slug via built HTML, not assumed)
- [x] `ruff check src tests` clean
[...]

PR #2756, opened 2026-07-10
docs: full-rewrite retrieval-engineering.md (2nd of 5, #2750)

tya5:
**[docs-maintainer]** — #2750, family-by-family rewrite, 2nd of 5 pages (Retrieval Engineering — one of the constitution's two honest thin areas, leaning under-claim per the same discipline used in charter.md's Context & Retrieval family).

## What changed
- **`recall`**: typed Control IR op, vector search over a pluggable `IndexBackend` — replacing the dead `recall_memory`/keyword-index description. No separate `recall_docs` mechanism — docs are retrieved the same way any other corpus is, via `embed_and_index()`.
- **Memory named explicitly as a separate mechanism** from `recall`/RAG (read inline by the router every turn, not queried on demand) — avoiding the old page's conflation of the two.
- **`web_search`/`web_fetch`** correctly stated as bundled Tier-1 default-allow tools (the old page claimed the opposite — this was one of the two confirmed-wrong claims flagged in #2749's banner).
- **"Where it's still thin"** states the real, current boundary honestly: Phase 1 scope only (SQLite backend, LiteLLM embedding passthrough; vector-store variety/rerank/HyDE/RAG-eval are explicit post-1.0 territory, not a secret gap), a framework not a pipeline, no bundled indexing skill.

## Also touched
`index.md`(`.ja.md`): staleness-note count updated from 4 remaining stale pages to 3 (Reliability, Security, Product Think).

## Template rules applied
Thin-area honesty (per the constitution's own declaration), no `history` refs, every claim grounded in `rag.md`/`memory.md`/`feature-map.md`'s Memory & RAG section.

## Test plan

Requesting architect anti-overclaim gate review, with particular attention to under- vs over-claim balance on the thin-area framing (per the same steer given for charter.md's Context & Retrieval family).


tya5:
**[architect]** — **anti-overclaim gate: CLEAN PASS ✅ — cleared for merge.** The thin-area balance is exactly right, and every currency claim (including the tricky *negative* one) is verified against source — the recalibration applied in both directions:

## (a) Positive currency claims — verified real ✅
- **`recall` = vector search** — `embed_and_index()` exists (`src/reyn/api/safe/embed_index.py:278`, exported), `recall` registers as a Control IR op (`op_runtime/recall.py:132`) and dispatches `index_query` per source (`recall.py:75`). Real vector path over a pluggable IndexBackend, not keyword/flat.
- **`web_search`/`web_fetch` bundled Tier-1** — `permissions.py:1367` ("Tier 1 gate for web_fetch — no declaration required"); both are registered ops, Tier-1 default-allow per the feature-map. This correctly **fixes the old page's "no web search or external retrieval primitive" error**.

## (b) The NEGATIVE claim — verified by absence ✅
"There is no separate `recall_docs` mechanism — docs are retrieved like any other corpus." `git grep recall_docs` over `src/reyn` returns **zero hits** — the mechanism genuinely doesn't exist, so the claim is correct (not an under-claim hiding a real feature, and not an over-claim inventing one). This is the harder direction — a page asserting something *doesn't* exist — and it checks out.

## (c) Thin-area framing — balanced, consistent with the charter ✅
"A framework, not a pipeline … a foundation to build retrieval on, not a deterministic, fully-managed RAG pipeline" matches the #2736 Retrieval charter cell verbatim in spirit ("a RAG *framework* to build on, not a fixed pipeline; one of the two honest thin areas"). It avoids **over-claim** (explicitly Phase-1-only, no bundled indexing skill, "you own the chunking logic") AND the **#2723-F1-type under-claim** (recall IS credited as real vector search, not "vector is a likely next step"). The "where it's still thin" list is candid without dressing up.

## (d) Memory ≠ recall — verified distinct ✅
Memory is a `MEMORY.md` index (`data/memory/memory.py:139`) read **inline into the router's context/system-prompt** (`router_loop.py:565` `get_memory_index()` → `:1694` feeds the context build → `router_system_prompt.py`), NOT via a tool call — `git grep recall_memory` returns zero (no such tool), confirming it's a distinct mechanism from the on-demand `recall` op, exactly as the page states.

## index.md
Staleness note now lists **three** stale pages (Reliability/Security/Product Think), credits Tool Contract + System Design + Retrieval as rewritten. Consistent.

**Gate cleared — merge it.** 3 of 5 rewrites down (Reliability/Security/Product Think to go). This is the model for a thin-area page: state plainly what exists, verify the negatives, and don't dress up the boundary.

PR #2759, opened 2026-07-10
docs: full-rewrite security.md (4th of 5, #2750)

tya5:
**[docs-maintainer]** — #2750, family-by-family rewrite, 4th of 5 pages (Security).

## What changed
Replaces the stale top-half sections while keeping the already-current Content-layer defense section (and its existing "still thin" notes) verbatim:

- **Three-layer permission model**: defaults → declared capability (JIT prompt at point of use, not startup) → project-wide pre-approval — replacing the dead phase-frontmatter-declaration model. Grounded in `permission-model.md`'s own "Three layers, in order" section, including the `approvals.yaml`/`index/sources.yaml` carve-out detail.
- **Actor-scoped approvals** (not skill-scoped) — `.reyn/approvals.yaml` keyed by `<actor>/<op>/<path>`, replacing the dead `run_skill`-inheritance framing with the real current composition-safety mechanism.
- **`sandboxed_exec`'s typed, per-axis `SandboxPolicy`** (write/network/exec/read asymmetric tightness, backend selection per platform) replacing the dead "AST sandbox for Python preprocessor steps" section entirely — that step kind doesn't exist in the current pipeline DSL.
- **"Where it's still thin"** swaps the dead AST-sandbox-is-honor-system note for a real, currently-documented residual risk: Landlock's read-deny-list is unenforceable (allowlist-only backend), unlike Seatbelt's last-match-wins semantics — grounded in `sandbox.md`'s own backend-asymmetry note, not invented.

## Also touched
`index.md`(`.ja.md`): staleness-note count updated from 2 remaining stale pages to 1 (Product Think, the last one).

## Test plan

Requesting architect anti-overclaim gate review.


tya5:
**[architect]** — **anti-overclaim gate: one cross-cell CLARITY fix (the 3-vs-4-layer you flagged), everything else verified.** The "3-layer" is *correct*, not a regression — but it needs a one-line bridge to the charter's "4-layer" so a reader doesn't read them as contradicting.

## The 3-vs-4-layer question — resolved against permission-model.md (authoritative)
They are **different granularities of the same implementation, both grounded, NOT a contradiction** — verified in `permission-model.md`:
- **#2759's "three-layer permission model"** (defaults → declared-capability/JIT → project pre-approval) = `permission-model.md:11` **"Three layers, in order"** — the **grant hierarchy** (how a capability gets *authorized*). Correct and current.
- **The charter #2729 "4-layer JIT approval"** (config pre-approval → saved → session → interactive) = `permission-model.md:188-190,210` — the **approval-source resolution** *inside* a single JIT decision. Also correct.
- The source even flags the ambiguity itself (`:13` Note: multiple things use the word "layers" for different questions). So #2759's "3-layer" is the grant hierarchy, NOT the old *phase-based* three-layer the #2749 banner correctly called stale, and NOT a contradiction with the charter's 4-layer.

**BUT — the clarity gap is real.** The #2749 banner explicitly corrected "three-layer → 4-layer" (for the old wrong page), and the charter cell says "4-layer." A reader who saw either, then reads #2759's "3-layer" with **no bridge**, will reasonably think it contradicts them or reverts to the stale framing. #2759's Layer-2 (JIT) description doesn't mention that the JIT prompt itself resolves through the 4-source order.
- **Fix (one line, in the Layer-2 / JIT part):** note that the JIT prompt resolves through the 4-source approval order — config pre-approval → saved `.reyn/approvals.yaml` → session → interactive — i.e. the "3 layers" is the *grant hierarchy* and the "4 layers" (the charter's) is *how one JIT decision is resolved*. That reconciles #2759 ↔ charter ↔ permission-model.md explicitly. Correctness is fine; this is purely to stop the 3-vs-4 reader-confusion the banner set up.

## Everything else — verified against source (recalibration) ✅
- **Actor-scoped approvals `<actor>/<op>/<path>`** — real: `permissions.py:1036` `f"{actor}/http.get/{host}"`, composite keys `<actor>/python.safe/…` (`:555`), persisted to `approvals.yaml`. Actor-scoped (not skill-scoped) confirmed.
- **`SandboxPolicy` per-axis** (write-allowlist / network-off / `allow_subprocess` / broad-read + `read_deny_paths`, strict read-allowlist abolished) — real: `landlock.py:232` "#1199 realignment — broad read surface"; `read_deny_paths` at `sandbox.md:21`. Matches #2738.
- **The residual-risk citation is genuine, not fabricated** — "Landlock read-deny-list not enforceable / Seatbelt asymmetry" is a documented, existing risk: `sandbox/backends/landlock.py:17-18` ("Landlock cannot express a read deny-list, so `policy.read_deny_paths` is not enforced here") + `:239` ("Residual risk: …") + `sandbox.md:21` ("Enforced only on backends that support deny-after-allow (Seatbelt)"). This is exactly the right honest-thin-area move: replace the dead "AST-sandbox honor-system" note with a *real* residual risk, cited not invented.
- **Dead "AST sandbox for python preprocessor" note removed** — `grep` for it in security.md returns zero. ✅

## Net
The 3-layer is right and all currency checks pass — the only thing to fix is the **cross-cell bridge**: add the one-line "JIT resolves via the 4-source order (the charter's 4-layer)" so "3-layer" here can't be misread against the charter's "4-layer" / the #2749 banner's correction. With that, clear. (index.md staleness should read 2→1, Product Think the last.) Re-ping on the bridge line.

Issue #2761, opened 2026-07-10
design: mid-execution hot-reload for install ops (drop turn-boundary so an install can use what it just installed)


tya5:
**[architect]** — **Design proposal (owner-judgment, pre-implementation): mid-execution hot-reload for install ops.** Owner ask: drop hot-reload's turn-boundary constraint so an install op reloads *inside* its own execution and can use the skill/pipeline/mcp it just installed within the same run. Flow-traced first (this is load-bearing). Scope: `skill_install` / `pipeline_install` / `mcp_install` (presentation out of owner's list).

## 1. The central question, resolved: turn-boundary is a SIMPLICITY/PREDICTABILITY choice, NOT a safety invariant
Flow-trace evidence (file:line):
- **The resolution layer already reads the registry LIVE, per tool-dispatch, within a turn** — `RouterLoop._invoke_via_registry` (`router_loop.py:3538`) rebuilds `RouterCallerState` per call → `build_resource_caller_state` reads `host.get_pipeline_registry()` (`router_host_adapter.py:668`, live attribute) fresh every dispatch; op-contexts read `presentation_registry=self._presentation_registry` fresh per op (`router_host_adapter.py:2046`, comment "the adapter's CURRENT registry snapshot (hot-reload swaps it)"); pipeline `call` steps resolve `deps.pipeline_registry.get(name)` at execute time (`executor.py:858`). **There is no turn-scoped cache to invalidate — a mid-turn swap is already observed correctly by the next dispatch.**
- **The swap is atomic-by-construction** — the skills/pipelines/presentations seams build a *fresh* registry and reassign the reference (`session.py:3748-3839`, "build a FRESH registry and SWAP the reference — never mutate in place"), so there is no half-applied window regardless of *when* it runs.
- **The one documented, already-tolerated hazard is REMOVAL/REPLACEMENT, not addition** — `session.py:3780-3786` (R7): an in-flight run resolves its own steps from the snapshotted work-order, but a not-yet-executed `call`-by-name resolves live, "so a mid-run reload can still change a pending call's target … Existing design, not a gap." This is a name-collision/removal hazard that **already exists at turn-boundary granularity today**; a pure *addition* of a brand-new name triggers none of it.
[...]

PR #2764, opened 2026-07-10
feat: mid-turn immediate hot-reload for skill/pipeline install (additive path) (#2761 PR-2)

tya5:
**[per-PR-coder]** — **PR-2 of the mid-execution hot-reload arc** (`part of #2761`): `HotReloader.apply_now` + immediate mid-turn apply for `skill_install` / `pipeline_install`. Implements the architect's **revised §2 path-condition** contract (the latest architect comment on #2761 — the revision of the "P1 seam-call contract"). PR-3 (mcp probe-then-commit) follows separately, with architect co-vet.

## What & why
Drops the turn-boundary constraint so an install op can **use what it just installed within the same execution** — the owner's goal. The mechanism is a **PATH CONDITION**, not a blanket guard (PR-1's step-1 surfaced that a blanket same-name refuse would strand skill/pipeline updates — they have no `remove` CLI, so re-install/clobber is their only update path — and break the documented `reyn mcp install` re-install fix; the architect revised §2 accordingly):

```
if is_pure_addition(name) and ctx.hot_reloader is not None:
    <write config> ; await ctx.hot_reloader.apply_now(source=...)   # immediate mid-turn
else:
    <existing deferred turn-boundary behavior, UNCHANGED>            # overwrite / no reloader
```

- **Pure addition (new name) + live per-session reloader** → the affected seam (`skills` / `pipelines`) is rebuilt+swapped IMMEDIATELY, so the new entry is **resolvable this turn**.
- **Same-name overwrite (clobber-update)** → the **existing deferred path, unchanged** — preserves skill/pipeline clobber-update + the documented mcp re-install fix, and **confines the R7 in-use-replace hazard to the deferred path it already lives on** (a pure addition never touches an in-USE entry).
- **No per-session reloader (CLI `reyn <kind> install`, separate process)** → existing deferred/restart best-effort, unchanged.

**Discovery vs resolution honesty:** only *resolution* (looking a name up to run it) is made immediate; the LLM's *discovery* catalog (tool schemas + `## Skills` menu, built once per turn) still updates next turn. Tests assert resolution only — no over-claim of mid-turn LLM discovery.

## Changes
- `runtime/hot_reload.py`: `HotReloader.apply_now(source=)` (re-read IN-set → validate-before-apply → run **only** the source-mapped seam(s) → emit `config_reloaded`; never touches the deferred `pending` flag); `is_pure_addition(name, entries)`; `dispatch_install_reload(ctx_reloader, source, is_addition)` (the shared addition→immediate / else→deferred router, preferring the **per-session** `ctx.hot_reloader` over the process-global for the immediate path — multi-session correctness); `_INSTALL_SOURCE_SEAMS` map (skill/pipeline; mcp added in PR-3).
- `core/op_runtime/context.py`: `OpContext.hot_reloader` field (the #2073 S3 per-session route, now threaded to op handlers).
- `runtime/router_op_context.py` + `runtime/services/router_host_adapter.py` (`self.hot_reloader`) + `runtime/session.py` (`self._hot_reloader`): thread `hot_reloader` into the OpContext at **both** router builders (the router-dispatched install path — `skill_management__install_*` / pipeline install → `build_legacy_op_context` → adapter `make_router_op_context` — is the load-bearing one).
- `core/op_runtime/skill_install.py` / `pipeline_install.py`: capture `is_pure_addition` BEFORE the entries-dict write; replace the tail `get_active_hot_reloader().request_reload(...)` with `dispatch_install_reload(...)`.

## impl step-1 verification (architect's checklist)
- `apply_now` runs the affected seam **only** (skill_install→`skills`, pipeline_install→`pipelines`), not the whole reload — tested directly.
- Per-session `ctx.hot_reloader` (session.py:1611 → adapter `self.hot_reloader` / session `self._hot_reloader`) is used for the immediate path, never the process-global — wired at both builders; op-level tests drive it through the real session reloader.
- The addition-condition routes a same-name overwrite to the deferred path — tested (overwrite not applied mid-turn, lands after `apply_pending`).

## Tests — `tests/test_2761_pr2_hotreload_immediate_apply.py` (Tier 2, 16 tests, real instances, no mocks)
- **A. `apply_now`**: affected-seam-only (skill / pipeline); doesn't touch `pending`; unknown source no-op; malformed IN-set rejected atomically (no seam runs); emits `config_reloaded`; operator `/reload` deferred path still runs ALL seams (unchanged).
- **B. routing**: `is_pure_addition`; dispatch addition→immediate; overwrite→deferred; no-reloader→deferred-via-active.
- **C. e2e (real handlers + real Session seams)**: NEW skill/pipeline resolvable same turn; same-name overwrite NOT applied mid-turn but lands at the turn boundary (clobber-update preserved, not errored); no-per-session-reloader defers.
- **cp-falsify**: neutered `dispatch_install_reload`'s immediate branch → the 5 immediate-apply tests went RED → restored via `cp` backup (never git).

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_2761_pr2_hotreload_immediate_apply.py` — 16/16 OK, 0 Tier-4.
- [x] `python scripts/verify_module_docstrings.py <changed src>` — OK.
- [x] `pytest` (repo root) — 7956 passed; the only 5 failures (`tests/web/*` route-mounting + `test_fp0001_a2a_endpoints`) are **pre-existing on clean origin/main** (verified identical at base 3402bdf2; environment/optional-dep, untouched by this PR).
- [x] cp-falsify RED→GREEN confirmed (immediate-apply invariant).
- [x] (Visual/Manual) none — pure runtime path condition, no UI/CLI surface change (skipped — no visual affordance).

### DIFF
--- a/src/reyn/core/op_runtime/context.py
+++ b/src/reyn/core/op_runtime/context.py
@@ -147,6 +147,16 @@ class OpContext:
     # non-chat) → the op skips it (the opt-in contract, same as the step-event gate).
     state_log: "StateLog | None" = None
 
+    # #2761 PR-2: this SESSION's HotReloader (the #2073 S3 per-session route), so an
+    # install op (skill_install / pipeline_install) can apply its reload IMMEDIATELY
+    # (mid-turn) for a PURE ADDITION — making the just-installed NEW entry resolvable
+    # this turn. A same-name overwrite / no reloader keeps the deferred turn-boundary
+    # path. Per-session (never the process-global get_active_hot_reloader, which is the
+    # last-registered session — a multi-session footgun). None outside a live chat
+    # session (tests / CLI separate-process install) → the op falls back to the deferred
+    # path (unchanged behavior).
+    hot_reloader: "object | None" = None
+
     # FP-0022 follow-up: declarative SSL config for web_fetch and MCP registry.
     # Defaults to WebConfig() (= no override, falls through to env-var chain).
     # Callers that have a ReynConfig available should pass config.web here.
--- a/src/reyn/core/op_runtime/pipeline_install.py
+++ b/src/reyn/core/op_runtime/pipeline_install.py
@@ -31,9 +31,11 @@ Local-path install (``op.source is None``):
   7. ``record_config_generation`` on the pipelines.yaml path AFTER write —
      the truncation-surviving recovery base (#2259 / CLAUDE.md recovery gate).
   8. Emit ``pipeline_installed`` event (P6 audit trail).
-  9. Request hot-reload so the installed pipeline goes live in the current session
-     (the ``"pipelines"`` seam — ``Session._reapply_pipelines`` — rebuilds the
-     registry from the fresh config cascade).
+  9. Reload so the installed pipeline goes live (#2761 PR-2): a PURE ADDITION on a
+     live per-session reloader (``ctx.hot_reloader``) applies the ``"pipelines"`` seam
+     (``Session._reapply_pipelines`` — rebuilds the registry from the fresh config
+     cascade) IMMEDIATELY (mid-turn) — resolvable this same turn; a same-name overwrite
+     (clobber-update) or no per-session reloader keeps the deferred turn-boundary path.
 
 Source/git install (``op.source`` set):
   Same pipeline as local, but step 0 fetches the DSL file first:
@@ -410,6 +412,11 @@ async def handle(
     }
     if op.source:
         entry["source"] = op.source
+    # #2761 PR-2: capture pure-addition-vs-overwrite BEFORE the write mutates entries,
+    # so step 9 routes a NEW name to the immediate mid-turn apply and a same-name
+    # overwrite (clobber-update — pipeline's only update path) to the deferred path.
+    from reyn.runtime.hot_reload import is_pure_addition  # noqa: PLC0415
+    _is_addition = is_pure_addition(safe_name, existing["pipelines"]["entries"])
     existing["pipelines"]["entries"][safe_name] = entry
     _write_yaml(config_path, existing)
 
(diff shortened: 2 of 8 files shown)


tya5:
**[per-PR-coder]** — **op-ctx topology completeness evidence** (for the architect's light co-vet: *"are the two threaded builders the complete set of op-ctx construction paths that carry a skill/pipeline install op — no third path that threads `hot_reloader=None` → silent-deferred?"*).

Traced every `skill_install` / `pipeline_install` op-ctx construction path:

**1. Emit sites (all install ops):** `skill_verbs.py` (install_local + install_source) and `pipeline_management_verbs.py` (2 install handlers) — **all 4 use `build_legacy_op_context(ctx)`**. There is **no** `reyn skill install` / `reyn pipeline install` CLI that builds a direct OpContext (only mcp has that — `mcp.py:747`, which is PR-3's surface).

**2. `build_legacy_op_context` (op_context_bridge.py) has exactly two branches:**
- **Preferred** → `rs.op_context_factory()`. `op_context_factory` is bound in **one** place: `types.py:264` `getattr(host, "make_router_op_context", None)` (host = RouterHostAdapter). That is the builder this PR threads `hot_reloader=self.hot_reloader` into. ✓
- **Fallback** → minimal synthesis (`op_context_bridge.py:49`), fires only when `rs is None` / factory `None` = the test/phase path with **no live session → no reloader to apply immediately anyway**. `hot_reloader=None` → deferred = correct (not a silent-wrong immediate-miss).

**3. The one factory wrapper — `pipe.py:702-711`** (`_factory_with_stdout_present`) **composes over** the base factory: it calls `_base()` and augments only `presentation_renderer` on the **same** returned ctx, so `hot_reloader` is **preserved**, not dropped. (And a headless `reyn pipe` with no session reloader → deferred = correct.)

**Conclusion:** the two threaded builders (adapter `make_router_op_context` + session `_make_router_op_context`, both via the single `build_router_op_context`) are the complete set for a **live-session** install. Every non-threading path is a context with **no live per-session reloader**, where deferred is the intended (and pre-existing) fallback — the immediate path requires a live `ctx.hot_reloader` by design. **No third path silently degrades a live install to deferred.**

Honest forward-looking caveat (not a current hole): this rests on the ADR-0026 invariant that `op_context_factory` is the single tool→op_runtime ctx source and is bound only to the adapter's `make_router_op_context`. If a future host binds a *different* factory that omits `hot_reloader`, a live install there would silently fall back to deferred. Not present today.

PR #2421, opened 2026-07-02
feat(#2421): MCP holistic seam — single gateway defeats the crash class (all paths, contain-all)

tya5:
**[e2e-coder]** — owner-GO'd holistic MCP fix (non-whack-a-mole). Design: `docs/dev/mcp-holistic-seam-design.md` (#2418). Owner's remaining crash root.

## The whack-a-mole, closed
Fault-isolation + task-affine lifecycle + per-call timeout were applied **per entry path**, so the list/probe paths were swept-missed and the crash recurred. This lands the single **`MCPGateway`** seam every MCP op routes through; a new entry path is a thin caller by construction and cannot re-introduce the crash class.

## The four guarantees (one place — `reyn/mcp/gateway.py`)
- **[1] structured sub-task join** — the SDK `stdio_client` uses `anyio.create_task_group()` and joins its `stdout_reader`/`stdin_writer` at `close` (verified in the installed SDK). Via the pool's same-task close, a sub-task fault surfaces **synchronously in-scope** → never orphans to the loop ("Task exception was never retrieved"). So [1] is reuse of SDK structured concurrency + a359 same-task close.
- **[2] contain-all boundary + `is_real_control_flow`** — `except BaseException`; re-raise ONLY genuine control flow: `KeyboardInterrupt`/`SystemExit`, or `CancelledError` **when the task is actually cancelling** (`current_task().cancelling() > 0`). A **spurious** cancel-mixed teardown group from a dead subprocess (task not cancelling) is contained, not propagated (**the crash the conservative predicate re-raised**). Exception-structure-independent. **Verified**: genuine `task.cancel()` propagates; synthetic/spurious CancelledError contained.
- **[3] pool task-affinity** — reuse `MCPClientPool` (a359 P2); `__aexit__` now uses `is_real_control_flow`.
- **[4] per-call timeout** — `resolve_call_timeout` (relocated into the gateway; op_runtime's duplicate removed).

The gateway raises **only `MCPFault` (an Exception) or genuine control flow — never a bare `BaseExceptionGroup`** → thin callers need only `except MCPFault`.

## All paths are thin callers
- op-call (`op_runtime/mcp.py _execute`) → `gateway.call_tool` (inline boundary + timeout absorbed).
- list (`session.py _mcp_list_tools`) → `gateway.list_tools` — **the owner Windows crash path**.
- probe-status → `gateway.probe`; probe-tools (`_probe_server_tools`) → `gateway.list_tools`.

## Completeness — structural, CI-enforced
[...]

Issue #2714, opened 2026-07-09
MCP stdio server subprocesses orphaned on normal reyn exit (Windows) — registry.shutdown omits aclose_mcp_connections


tya5:
**[lead-coder]** — owner-reported (Windows git bash: many leftover python processes in Task Manager). Root-caused via read-only code investigation (file:line verified). Distinct from #2292.

## Bug (verified in code)
The main interactive session's held MCP stdio connections are **never closed on normal `reyn chat` exit**:
- Normal REPL exit / Ctrl-C → `interfaces/repl/repl.py:121,127` → `registry.shutdown()` (`registry.py:2983`) → `agent.shutdown()` (posts a shutdown sentinel, `session.py:3418`) + cancels forward tasks. It does **NOT** call `aclose_mcp_connections` / `MCPConnectionService.aclose`.
- `session.run()`'s own `finally` teardown (`session.py:4459-4483`) closes the fs-watcher + emits events but **does NOT** close MCP.
- `aclose_mcp_connections` (`session.py:6118`) is wired ONLY from `registry.remove_session` (`registry.py:1926`, for spawned sub-sessions) and `archive_agent` (`registry.py:776`) — neither runs for the main session on ordinary exit.

MCP stdio servers are opened persistently (`keep_alive=True`, `mcp/client.py:931-941`; `MCPConnectionService._ensure_open` `connection_service.py:349`) with subprocess lifetime = client object. They are commonly `python -m ...` / `uvx`.

## Windows impact (why the owner sees it)
On Unix, orphaned children are reaped by init / EOF-driven self-exit. On **Windows there is no such reaping** — the held MCP subprocess persists in Task Manager after every `reyn` exit. Accumulation: one set per `reyn` invocation, plus one per transport-death reconnect (`connection_service.py:409-420`), plus one per MCP call on ephemeral sessions (each opens+tears down a fresh `MCPClientPool`, `session.py:6079-6082`).

## Fix direction
Wire `aclose_mcp_connections` (async, graceful stdio close → subprocess terminate) into the main-session normal-exit path — `registry.shutdown()` and/or `session.run()`'s finally. Ensure it runs on Ctrl-C / window-close too. Belt-and-suspenders: a process-exit reap of outstanding MCP subprocesses.

## Related / to verify on Windows
[...]

PR #2765, opened 2026-07-10
feat: mid-turn immediate mcp install via probe-then-commit (stdio+remote) (#2761 PR-3)

### LINKED ISSUE #2761: design: mid-execution hot-reload for install ops (drop turn-boundary so an install can use what it just installed)

tya5:
**[per-PR-coder]** — **PR-3 (final phase) of the mid-execution hot-reload arc**: mcp install (stdio + remote) immediate mid-turn apply via **probe-then-commit**. Completes the arc — `Closes #2761`. Implements the architect's #2761 P1 seam-call contract §3 (probe-then-commit + cancel/timeout/rollback). **Load-bearing cancellation/teardown — architect co-vet requested.**

## What & why
Extends the PR-2 path-condition to mcp: a **PURE ADDITION** on a live per-session reloader takes the IMMEDIATE mid-turn path — but mcp additionally **PROBES** the server (spawn/connect + `list_tools`) BEFORE writing config, committing ONLY on a successful probe. A same-name overwrite (the documented `reyn mcp install` re-install fix) or no per-session reloader keeps the existing DEFERRED turn-boundary path, unchanged.

**Probe-then-commit (rollback-free by ordering):**
```
if is_pure_addition(name) and ctx.hot_reloader:
    probe(server)                # spawn/connect + list_tools, inline-awaited, bounded
    if probe fails/times-out/cancels: return error  # NOTHING written → no half-install
    write config ; apply_now()   # COMMIT only after a successful probe
else:
    write config ; deferred reload   # overwrite / no reloader — unchanged
```
The write is **strictly after** the probe, so a failed/cancelled probe leaves nothing committed — atomicity is by ordering, no rollback needed.

**Transport-uniform + crash-safe teardown (reuse, no reinvention):** the probe routes through the existing `MCPGateway` seam (#2421) — open + list + teardown inside one contain-all boundary + a per-server timeout, task-affine. It raises ONLY `MCPFault` for a transport/timeout fault (→ error string, no write), and **re-raises a genuine Ctrl+C `CancelledError`** as control flow after the pool's `__aexit__` tears the transport down (stdio subprocess kill via `kill_process_tree` / HTTP close). stdio and remote (http/sse) share this one path — `MCPClient.__aexit__` owns the transport difference, so the probe never branches on transport.

## Changes
- `core/op_runtime/mcp_install.py`: `probe_mcp_server(name, entry, agent_id)` (crash-safe probe via `MCPGateway`, transport-uniform, mirrors `_mcp_list_tools`); `handle` restructured to probe-then-commit + path-condition (`is_pure_addition` before mutation, probe-gated write, `dispatch_install_reload` for the reload). Emits `mcp_install_probe_failed` (P6) on a failed probe.
- `runtime/hot_reload.py`: `_INSTALL_SOURCE_SEAMS += {"mcp_install": ("mcp",), "mcp__install_local": ("mcp",)}` so `apply_now` runs only the mcp seam.
- `tools/mcp_verbs.py` (`_handle_mcp_install_local`): **completeness extension — see below.**

## ⚠️ Completeness extension (for architect co-vet): `mcp__install_local` — the PRIMARY stdio path
While tracing, I found `_handle_mcp_install_local` (the `mcp__install_local` verb — **the primary way to add a local stdio server**) writes `.reyn/config/mcp.yaml` **directly**, a **parallel write path that bypasses `mcp_install.handle`**. So the contract's `mcp_install.py:517` anchor covers registry/source installs but NOT the local-stdio path the architect called "the primary use." Leaving it out would make the feature unreachable-for-purpose for the primary stdio case (same class as the PR-2 `ctx.hot_reloader` wiring gap). I applied the **same** probe-then-commit + path-condition to `_handle_mcp_install_local` (reusing the shared `probe_mcp_server` + `is_pure_addition` + `dispatch_install_reload`; the per-session reloader is reached via the router's single op-context factory). **Flagging explicitly** — if you'd prefer this split to a follow-up PR, it's an isolated ~40-line section, cheap to revert.

## impl step-1 verification (architect's §4 checklist)
- **probe-then-commit ordering** — the config write is strictly after a successful probe; a failed/cancelled probe writes nothing (falsify test pins this).
- **cancel → teardown, no orphan** — the probe rides the cancellable turn task; `MCPGateway` re-raises genuine `CancelledError` after `MCPClient.__aexit__` tears the transport down (verified via the gateway's `is_real_control_flow` contain-all boundary + the real-subprocess cancel test).
- **timeout** — `MCPGateway`'s per-server `asyncio.timeout` bounds spawn+probe (unreachable-remote test).
- **`ctx.hot_reloader` (per-session)** used for `apply_now`, never the process-global.
- **discovery vs resolution** — tests assert the server is *resolvable/usable* same turn (live roster + tool cache), not the LLM's mid-turn catalog.

## Tests — `tests/test_2761_pr3_mcp_immediate_probe.py` (Tier 2, 10 tests, real instances incl. a REAL stdio MCP server subprocess, no mocks)
- **probe**: reachable stdio → None; unreachable stdio → error; unreachable **remote (http)** → error (transport-uniform).
- **apply_now**: `mcp_install` / `mcp__install_local` sources target the mcp seam only.
- **e2e (real Session + real `tests/_support/mcp_tools_only_pid_server.py`)**: NEW reachable server → probed, committed, **live roster + tools cached same turn** (immediate, `pending` stays False); **probe failure → nothing written** (no half-install); **same-name overwrite → not probe-gated, deferred** (documented re-install preserved, clobber-update written for the turn boundary); **no per-session reloader → deferred, no probe**; **Ctrl+C during a hung probe → `CancelledError` propagates + nothing written** (rollback-free by ordering).
- **cp-falsify**: neutering `probe_mcp_server` (always "succeeds") → the 4 probe/rollback tests go RED → restored via `cp` backup (never git).

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_2761_pr3_mcp_immediate_probe.py` — 10/10 OK, 0 Tier-4.
- [x] `python scripts/verify_module_docstrings.py <changed src>` — OK.
- [x] `pytest` (repo root) — full suite green modulo the pre-existing `tests/web/*` route-mounting failures (unrelated env/optional-dep; identical on clean origin/main).
- [x] cp-falsify RED→GREEN confirmed (probe-then-commit / rollback-free invariant).
- [x] (Visual/Manual) none — runtime path condition + probe, no UI/CLI surface change (skipped — no visual affordance).

### DIFF
--- a/src/reyn/core/op_runtime/mcp_install.py
+++ b/src/reyn/core/op_runtime/mcp_install.py
@@ -5,7 +5,13 @@ Handler logic (one-shot, no sub-phases):
   2. Check runtime command availability (npx / uvx / docker / dnx)
   3. Gate via PermissionResolver.require_mcp_install (ADR-0029)
   4. Prompt for secret env vars via intervention_bus; persist with secrets.store
-  5. Write mcp.servers.<name> into the target scope config file
+  5. Reload (#2761 PR-3): a PURE ADDITION on a live per-session reloader takes the
+     IMMEDIATE mid-turn path — PROBE the server (spawn/connect + list_tools) FIRST,
+     write mcp.servers.<name> ONLY on a successful probe (probe-then-commit: a
+     failed/cancelled probe leaves nothing written — no half-install), then
+     apply_now so its tools are resolvable this same turn. A same-name overwrite
+     (the documented re-install fix) or no per-session reloader keeps the deferred
+     turn-boundary path (write + request_reload), unchanged.
   6. Emit mcp_server_installed event (P6)
 
 Scope → file mapping:
@@ -146,6 +152,44 @@ def _build_server_entry(pkg_raw: dict, env_keys: list[str]) -> dict:
         entry["env"] = {k: f"${{{k}}}" for k in env_keys}
 
     return entry
+
+
+async def probe_mcp_server(
+    server_name: str, server_entry: dict, *, agent_id: str | None = None,
+) -> "str | None":
+    """#2761 PR-3: probe a prospective MCP server (spawn/connect + ``list_tools``)
+    BEFORE its config is committed — the probe-then-commit atomicity gate.
+
+    Returns ``None`` on a successful probe (the server is reachable and advertises
+    tools) or an error string on failure. The caller writes the config ONLY on
+    ``None`` — so a failed probe leaves NOTHING written (no half-install, no rollback).
+
+    Routed through the crash-safe :class:`~reyn.mcp.gateway.MCPGateway` seam (#2421):
+    open + list + teardown inside one contain-all boundary + a per-server timeout,
+    task-affine so the SDK's stdio_client/ClientSession scopes close in the same task.
+    The gateway raises ONLY :class:`MCPFault` for a transport/timeout fault (contained
+    here → error string); a genuine Ctrl+C ``CancelledError`` is re-raised by the
+    gateway as control flow and propagates OUT of this probe — the pool's ``__aexit__``
+    has already torn the transport down (stdio subprocess kill via ``kill_process_tree``
+    / HTTP close), and because the config write is strictly AFTER this probe, a cancel
+    leaves nothing committed. **Transport-uniform**: stdio and remote (http/sse) share
+    this one path — ``MCPClient.__aexit__`` owns the transport-appropriate teardown, so
+    the probe never branches on transport (mirrors ``Session._mcp_list_tools``)."""
+    from reyn.mcp.client import expand_env  # noqa: PLC0415
+    from reyn.mcp.gateway import MCPFault, MCPGateway  # noqa: PLC0415
+
+    expanded = expand_env(server_entry)
+    if not isinstance(expanded, dict):
+        return f"server config must be a dict, got {type(expanded).__name__}"
+    # A url-only remote entry defaults to http (mirrors _mcp_list_tools).
+    if "type" not in expanded and expanded.get("url"):
+        expanded = {**expanded, "type": "http"}
+    gateway = MCPGateway(agent_id=agent_id)
+    try:
+        await gateway.list_tools(server_name, expanded)
+    except MCPFault as exc:
+        return str(exc)
+    return None
 
 
 def _read_yaml_config(path: Path) -> dict:
@@ -509,11 +553,51 @@ async def handle(
             ),
         }
 
-    # Merge into existing config
+    # Ensure the mcp section shape exists (read-only prep — NO write yet).
     if "mcp" not in existing or not isinstance(existing.get("mcp"), dict):
         existing["mcp"] = {}
     if "servers" not in existing["mcp"] or not isinstance(existing["mcp"].get("servers"), dict):
         existing["mcp"]["servers"] = {}
+
+    # ── 5b. Probe-then-commit + path-condition (#2761 PR-3) ───────────────────
+    # Capture pure-addition-vs-overwrite BEFORE any mutation. A PURE ADDITION on a
+    # live per-session reloader takes the IMMEDIATE mid-turn path: PROBE the server
+    # (spawn/connect + list_tools) FIRST, and write config ONLY on a successful probe
+    # → a failed/cancelled probe leaves nothing committed (no half-install, no
+    # rollback). A same-name overwrite (the documented `reyn mcp install` re-install
+    # fix) or no per-session reloader keeps the existing deferred turn-boundary path
+    # (unchanged), confining any in-use-replace to the deferred path.
+    from reyn.runtime.hot_reload import (  # noqa: PLC0415
+        dispatch_install_reload,
+        is_pure_addition,
+    )
+    _is_addition = is_pure_addition(server_name, existing["mcp"]["servers"])
+    _reloader = getattr(ctx, "hot_reloader", None)
+    if _is_addition and _reloader is not None:
+        _probe_err = await probe_mcp_server(
+            server_name, server_entry, agent_id=getattr(ctx, "agent_id", None),
+        )
+        if _probe_err is not None:
+            ctx.events.emit(
+                "mcp_install_probe_failed",
+                server_id=op.server_id,
+                server_name=server_name,
+                error=_probe_err,
+            )
+            return {
+                "kind": "mcp_install",
+                "status": "error",
+                "server_id": op.server_id,
+                "server_name": server_name,
+                "source": op.source or "",
+                "error": (
+                    f"MCP server {server_name!r} failed its pre-install probe "
+                    f"(nothing written — no half-install): {_probe_err}"
+                ),
+            }
+
+    # ── 5c. COMMIT: write mcp.servers.<name>. The immediate path reaches here ONLY
+    # after a successful probe; the deferred path always writes (unchanged). ──────
     existing["mcp"]["servers"][server_name] = server_entry
 
     _write_yaml_config(config_path, existing)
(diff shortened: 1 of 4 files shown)


tya5:
**[per-PR-coder]** — **co-vet evidence** for the three load-bearing points (so the architect can clear fast).

**1. Atomicity by ordering (no half-install, no rollback).** The config write is structurally gated on a successful probe:
- `mcp_install.py`: "Ensure the mcp section shape exists (read-only prep — NO write yet)" → §5b probe (`_is_addition and _reloader is not None` → `probe_mcp_server(...)`; on `_probe_err is not None` → emit `mcp_install_probe_failed` + **`return` error, no write**) → §5c "COMMIT: write mcp.servers.<name>" reached ONLY after a successful probe. Same shape in `_handle_mcp_install_local`.
- So a failed/timeout/cancel path never reaches the write — atomicity is by ordering, rollback is unnecessary by construction. Pinned by `test_local_install_probe_failure_writes_nothing` + `test_local_install_probe_cancel_writes_nothing` (assert the name is absent from the on-disk config), and the cp-falsify (neuter the probe → both go RED).

**2. Cancel / teardown orphan-free — REUSED hardened machinery, not reinvented here.** The probe runs through `MCPGateway` (#2421) → `MCPClient` `async with`. Teardown is the **existing #2714/#2716-hardened path**: `MCPClient.__aexit__` → subprocess reap via `kill_process_tree` (stdio) / HTTP-client close (remote). That orphan-free guarantee is already covered by `tests/test_2714_mcp_shutdown_teardown.py` — including the swallowed-Windows-teardown-fault case (`test_close_reaps_subprocess_even_when_graceful_close_raises`: the OS subprocess is reaped even when the graceful close RAISES). My PR adds NO new teardown code — it reuses that path. On top, the gateway's `is_real_control_flow` re-raises a genuine Ctrl+C `CancelledError` as control flow (contains only transport faults as `MCPFault`), so cancel propagates OUT after teardown — proven end-to-end by the real-subprocess `test_local_install_probe_cancel_writes_nothing` (hung `python` that never speaks MCP → `task.cancel()` → `CancelledError` raised + nothing written). Transport-uniform: one path, `MCPClient.__aexit__` owns the stdio-vs-remote teardown difference.

**3. `mcp__install_local` extension — keep in PR-3 (my recommendation, agreeing with lead's lean).** It is the PRIMARY local-stdio install path and writes `.reyn/config/mcp.yaml` **directly** (parallel to the op, which the contract anchor `mcp_install.py:517` did not cover). Splitting it out would leave the primary local-stdio path WITHOUT immediate mid-turn reload — i.e. the owner's main stdio use case would not get "install → use same turn," defeating the arc's stated goal for exactly the case the architect called "the primary use." It reuses the same three primitives (`probe_mcp_server` / `is_pure_addition` / `dispatch_install_reload`) — no divergent logic — and is an isolated ~40-line section if you still prefer a split. Same-name re-install stays deferred + un-probed on this path too (`test_local_install_same_name_overwrite_defers_and_skips_probe`), preserving the documented re-install workflow.

Happy to adjust on any of the three if the co-vet wants a different shape.

Issue #2783, opened 2026-07-10
StateLog.aclose() and EventStore.aclose() are never called from production teardown


tya5:
**[tui-coder]** —

## 発見の経緯

#2780（EventStore off-loop write fix）の実装中、`EventStore.aclose()`（worker drain）を追加し、`StateLog.aclose`と同じteardown経路から呼ぶ設計にしようとしたところ、**その「同じteardown経路」が実は存在しない**ことが判明した。

## 事実確認

`grep -rn "\.aclose()" src/reyn/` の結果、`StateLog.aclose()`（`state_log.py:310`、#1765で追加）を呼んでいるのは **テストコードのみ**。`session.py`/`registry.py`/CLI層のどこからも、本番の session teardown 経路（`/quit`、正常終了、CLI batch完了時）で呼ばれていない。`Session`/`AgentRegistry`も`__aexit__`を持たず、async context managerとしても使われていない。

`FsWatcher.aclose()`/`MCPConnectionService.aclose()`は`session.py:6151`/`6167`（`aclose_fs_watcher`/`aclose_mcp_connections`）から呼ばれる経路があるが、これらが実際に session の正常終了フローで呼ばれているかも要確認。

## 影響

`asyncio.run()`はループ終了時に未完了タスクをキャンセルするため、現状:
- WAL(`state_log.py`)の`DurabilityWorker`にqueue済みで未書き込みの末尾エントリが、通常の`/quit`で失われ得る
- 今回追加する`EventStore.aclose()`も同様（呼ばれなければ意味がない）

いずれも「稀に末尾数件が失われる」程度で、crash recovery自体は無傷（#1765のWAL fixは"submit"経由の同期パスでは既にawaitされるため、この問題は主に`submit_nowait`/fire-and-forget経路に限定される可能性が高い — 要検証）。

## 次のアクション

正しいteardown hookをどこに置くべきか（`reyn chat`のCLIエントリポイント、`reyn run`のbatch完了、dogfood runner、A2A spawned session、等 複数の入口を横断する必要がある）の調査 + `StateLog.aclose()`と`EventStore.aclose()`を実際に配線する設計・実装。#2780とは別スコープ（#2780はaclose()/flush()を「呼べば動く」状態にするところまで）。

tya5:
**[tui-coder]** — Design follow-up: mapped every production entry point's current teardown behavior (research only, no implementation — per lead-coder's "余力あれば設計詰めておいても、実装は別GO" ask).

[...]

Issue #1547, opened 2026-06-13
[1533/1f follow-up] TUI /rewind: per-checkpoint user-message preview (Phase-2)


tya5:
**[e2e-coder]** — Stage 1f Phase-1 timeline shows seq + rel-time + kind-label only. This tracks the per-checkpoint user-message preview as Phase-2.

## Context
Stage 1f Phase-1 (TUI `/rewind` timeline) shows each checkpoint as `seq · ⏱ rel-time · kind` (turn / plan-step / phase). It does **not** show a content anchor (the user message at that checkpoint, à la Claude Code's `/rewind`).

## Why preview is NOT in Phase-1 (assessment, #1533 comment)
Mining the preview from the audit EventStore (`.reyn/events/<run>.jsonl`) was assessed and rejected:
- The audit `Event` model has **no WAL seq** (only `type`/`timestamp`/`data`, `schemas/models.py:726`). Generations are keyed by WAL seq.
- WAL (StateLog) and audit EventStore are **separate logs with separate sequencing** — correlating gen-seq → user-message needs a **cross-log join** (timestamp proximity or chain_id), which couples the two logs against the `project_wal_vs_audit_event_separation` principle.
- The audit log has its own rotation/retention → preview availability decouples from generation retention (older checkpoints — the ones you most want to rewind to — lose their preview first).

## Phase-2 design (the correct source)
Capture a truncated anchor string (the last user message, ~80 chars) **into the generation at `cut_generation()` time**. The `history_buffer` holds it in-memory at cut time, so it's cheap, robust, survives independent of audit rotation, and needs no cross-log join.

- This is a **1a-adjacent additive field** on the generation snapshot (or a sidecar) — **coordinate with the Stage 1a/1d owner (sandbox_2)**.
- `AgentRegistry.list_rewind_points()` then surfaces the preview field; RewindMenuWidget renders it as a second dim line per checkpoint.

## Depends on
- Stage 1f Phase-1 (timeline + `list_rewind_points()`) landing first.
- Coordination with sandbox_2 on the generation-record additive field.

Refs #1533

tya5:
**[dogfood-coder]** — DONE, closing. Landed in **PR #1552** (merged 06:47Z) — per-checkpoint anchor-text preview on the `/rewind` timeline.

Shipped (the locked Phase-2 design):
[...]

Issue #2292, opened 2026-07-01
Windows: _kill_proc_group terminate/kill hits the process, not its child tree (orphans on cancel)


tya5:
**[e2e-coder]** — known limitation flagged during the Windows broken-pipe fix (branch fix/windows-subprocess-broken-pipe).

## The limitation
`noop_backend._kill_proc_group` now falls back to `proc.terminate()` → `proc.kill()` when `os.killpg` is absent (Windows). Unlike the POSIX `os.killpg(os.getpgid(pid), ...)` path — which signals the whole PROCESS GROUP — `terminate()`/`kill()` on Windows hits ONLY the target process, not its child tree.

**Consequence:** a sandboxed process that spawns grandchildren can leave ORPHANS on cancel/timeout on Windows (the children survive the parent's termination).

## Out of scope for the broken-pipe fix
The broken-pipe fix (the primary bug: `select()` can't poll pipe fds on Windows → the thread-based drain) is orthogonal. This tree-kill gap is a SEPARATE, deeper concern.

## Fix direction (future)
Windows tree-kill needs either:
- a **Job object** (`CreateJobObject` + `AssignProcessToJobObject` with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`) — the robust approach (kills the whole tree), or
- `taskkill /F /T /PID <pid>` (spawns taskkill to walk + kill the tree — simpler, less robust).

Low priority until Windows is a supported sandbox target for tree-spawning workloads. Referenced from the code comment in `noop_backend._kill_proc_group`.

PR #2784, opened 2026-07-10
fix: offload EventStore.write() from the event loop (#2780)

tya5:
**[tui-coder]** —

## Summary

`EventStore.write()` did synchronous `open()`/`write()` directly on the event loop — the same class of bug as #1765's WAL append (a filesystem stall freezes the WHOLE event loop), except **unmitigated** (not even fsync was offloaded) and **far more exposed**: fires on every chat event (at least once per turn via `turn_completed`, plus once per tool call, per hook, etc.), not just WAL appends.

Owner-driven investigation (sync-API audit of the turn hot path, following the #1765 precedent) traced the root cause and designed the fix (Fable-reviewed, lead-coder review-gate items folded in — see below). Owner GO to implement.

## Fix

- Reuses `DurabilityWorker` (the same off-loop primitive #1765 introduced — substrate-agnostic by that class's own design, no reinvention) via `submit_nowait`. `write()` stays a plain synchronous method — **no API change, no caller updates** anywhere `emit()`/`write()` is called across the codebase.
- Rotation decision + JSON serialization still happen synchronously at `write()`-call time, so enqueue order == emission order; the worker's FIFO guarantee then keeps on-disk order matching emission order too — this log's ordering relative to other synchronous code (WAL appends included) is unchanged from before this fix.
- **Durability discipline mirrors the WAL fix** (lead-coder review-gate item): the single off-loop unit (`_write_line_sync`) does open + write + flush + **fsync together**, so there is no "written but not yet fsynced" exposure window within one queued job. `submit_nowait` (not `submit`) means `write()` itself doesn't await that durability — a relaxed-durability window, the same accepted trade-off the WAL's own `append_nowait`/`submit_nowait` fire-and-forget path already carries elsewhere in this codebase, not a new risk class.
- Falls back to a fully synchronous write when no event loop is running (e.g. a CLI entry point) — `submit_nowait` requires a running loop and would otherwise raise, a regression this fix must not introduce.
- **Second bug found and fixed during this work**: `_should_rotate()`'s size check called `Path.stat()` on **literally every `write()` call** (`max_bytes` defaults to a nonzero 10MB — not a rare rotation-time-only path as originally assumed). Replaced with an in-memory running byte counter — eliminates the syscall entirely rather than just deferring it.
- Added `EventStore.aclose()`/`flush()` (mirroring `StateLog`'s) so a normal `/quit` doesn't silently drop the trailing audit events (e.g. `session_completed` — the event recording the graceful exit itself) when `asyncio.run()` cancels outstanding tasks at teardown.

## Crash-recovery / doc-mirror gate (explicit per lead-coder's review-gate ask)

- **Recovery-feature truncate-falsify gate: does NOT apply.** `anchor_store.py` explicitly documents EventStore has no WAL seq and is deliberately not used for rewind/reconstruction ("mining the audit EventStore... would need a cross-log join" — #1547). This fix touches only the audit log, never WAL-event-derived reconstruction state.
- **`events.md` doc-mirror: not needed.** Checked both `docs/concepts/runtime/events.md` and `docs/reference/runtime/events.md` — neither makes an implementation-level claim (synchronous write, fsync timing) that this fix would contradict. The documented contract ("records what happened, in order") is unchanged; only the blocking-vs-non-blocking write mechanism changed.
- **Follow-up tracked separately**: while wiring `aclose()`, discovered `StateLog.aclose()` (from #1765) is *also* never called from production teardown (only from tests) — filed as #2783, out of scope here (needs its own investigation across multiple process-teardown entry points).

## Test plan

- [x] `pytest tests/test_event_store_off_loop_write.py` (new, 7 tests) — RED→GREEN verified via `git stash` (6/7 failed with `AttributeError` on the new seams pre-fix; the 7th, the sync-fallback test, correctly passes both before and after since it exercises unchanged behavior)
- [x] Loop-freeze regression test uses the snapshot-before-drain discipline (the #2739/#1765 review lesson — a counter drained to completion first would pass unconditionally regardless of whether the loop froze)
- [x] `pytest tests/test_event_store_stale_path_recovery.py tests/test_2608_dogfood_hook_liveness.py tests/test_multisession_history_isolation_2348.py` (existing EventStore-adjacent suites) — all pass; two tests needed a `await store.flush()` added (they read the events file from an async context immediately after `emit()`, an assumption `write()` being fire-and-forget now breaks — both were genuine regressions caught by the full suite run, not pre-existing)
- [x] `ruff check` / `python scripts/test_tier_audit.py --strict` / `python scripts/verify_module_docstrings.py` (all changed files) — clean
- [x] Full `pytest` suite — 8006 passed, 9 failed (pre-existing/environmental, unrelated files — same known baseline as prior PRs this session), 17 skipped

### DIFF
--- a/src/reyn/core/events/event_store.py
+++ b/src/reyn/core/events/event_store.py
@@ -249,8 +249,10 @@ class EventStore:
         `max_bytes` defaults to a nonzero 10MB, so `.stat()` used to fire a
         blocking syscall on literally EVERY `write()` call, not a rare path.
         The counter drifts (harmlessly) if an external process appends to the
-        same file, or after a FileNotFoundError recovery re-creates it — both
-        only shift the rotation point by a bounded amount, never break
+        same file, after a FileNotFoundError recovery re-creates it, or on
+        Windows where text-mode `\n` -> `\r\n` translation makes bytes-on-disk
+        exceed the counted `len(line.encode("utf-8")) + 1` — all three only
+        shift the rotation point by a bounded amount, never break
         correctness."""
         if self._active is None or self._active_started_at is None:
             return False
--- a/src/reyn/core/events/event_store.py
+++ b/src/reyn/core/events/event_store.py
@@ -13,14 +13,62 @@ that rename-based schemes have.
 
 Per P7: this is OS-level generic infrastructure — it never references
 specific event types or domain strings.
+
+Off-loop write (owner dogfood finding, 2026-07-10): ``write()`` used to
+``open()``/``write()`` synchronously, directly on the event loop — the SAME
+class of bug as #1765's WAL append (a filesystem stall freezes the WHOLE
+event loop), except unmitigated (not even fsync was offloaded) and far more
+exposed (fires on every chat event — at least once per turn via
+``turn_completed``, plus once per tool call, per hook, etc., vs. #1765's
+WAL-append-only exposure). Fixed by routing the write through a
+``DurabilityWorker`` (the SAME off-loop primitive #1765 introduced — reused,
+not reinvented; substrate-agnostic by that class's own design) via
+``submit_nowait``. ``write()`` stays a plain synchronous method — no API
+change, no caller updates anywhere ``emit()``/``write()`` is called: only the
+actual file I/O is deferred off-loop. Rotation decision + line serialization
+still happen synchronously at ``write()``-call time, so enqueue order still
+equals emission order; the worker's FIFO guarantee (enqueue order == write
+order) then keeps on-disk order matching emission order too — this log's
+ordering relative to other synchronous code (WAL appends included) is
+unchanged from before this fix, only the blocking part moved off the loop.
+Cross-log disk-durability ordering between this store and the WAL is NOT
+guaranteed (separate workers/queues) — but nothing depends on that; event
+timestamps are stamped synchronously at ``emit()`` time, so a consumer that
+correlates the two logs (dogfood_trace, support_bundle) orders by timestamp,
+not by which file landed first.
+
+Durability discipline mirrors #1765's WAL fix, per review: the single
+off-loop unit (``_write_line_sync``) does open + write + flush + fsync
+TOGETHER, so there is no "written but not yet fsynced" exposure window
+within one queued job — a line is either not-yet-durable (still queued) or
+fully durable (opened, written, and fsynced). ``submit_nowait`` (not
+``submit``) means ``write()`` itself does not await that durability — a
+RELAXED-durability window between ``write()`` returning and the line
+actually landing durably, same accepted trade-off the WAL's own
+``append_nowait``/``submit_nowait`` fire-and-forget path already carries
+elsewhere in this codebase — not a new risk class.
+
+This is an AUDIT log, not a recovery source: ``anchor_store.py`` explicitly
+documents that EventStore has no WAL seq and is deliberately not used for
+rewind/reconstruction ("mining the audit EventStore... would need a
+cross-log join" — see #1547). So this fix carries no crash-recovery
+truncate-falsify obligation (CLAUDE.md's recovery-feature PR gate governs
+WAL-event-derived reconstruction state; this isn't that) — but per review,
+the audit log's at-least-once durability still matters for the
+`events.md` audit-truth contract, hence ``aclose()`` below (drains every
+queued write before process/session teardown, so a graceful ``/quit``
+doesn't silently drop the tail of the audit trail — see its docstring).
 """
 from __future__ import annotations
 
+import asyncio
 import json
+import os
 from datetime import datetime
 from pathlib import Path
 from typing import Iterator
 
+from reyn.core.events.durability_worker import DurabilityWorker
 from reyn.schemas.models import Event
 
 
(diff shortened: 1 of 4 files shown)


tya5:
**[dogfood-coder]** — co-vet on #2784, verify-don't-infer per lead-coder's request. **Verdict: CLEAR to merge**, with one minor non-blocking observation (not asked for, found while checking point 4).

## 1. Durability contract — VERIFIED accurate

`_write_line_sync` performs `open("a") → f.write() → f.flush() → os.fsync(f.fileno())` all inside one `with` block, so there is genuinely no "written but not fsynced" window — a line is either not-yet-attempted (queued) or fully durable. Confirmed by reading the code directly, not the docstring's claim alone.

The claimed parallel to WAL's `append_nowait` isn't just similar-sounding — it's **the literal same primitive**: `state_log.py`'s `append_nowait` calls `self._worker.submit_nowait(...)` on a `DurabilityWorker` instance, and this fix routes `EventStore.write()` through `self._worker.submit_nowait(...)` on the SAME `DurabilityWorker` class (reused, not reimplemented, substrate-agnostic by that class's own design). Since it's the identical class/method, the "same relaxed-durability trade-off" claim is true by construction, not by analogy.

## 2. Recovery-feature gate non-applicability — INDEPENDENTLY FALSIFIED, holds

Ran my own falsification, not a re-read of the author's claim: (a) `grep -rln "EventStore" src/reyn/` → 5 files (`events.py`, `event_store.py`, `anchor_store.py`, `registry.py`, `session.py`, plus CLI/web ones); (b) separately, `grep` for `def.*reconstruct|def.*recover|def.*restore|def.*rewind|_rewake|def.*replay` across `src/reyn/` → zero overlap with the EventStore file set outside what (a) already found. Reading the actual recovery code:
- `registry.py`'s `list_rewind_points`/`predecessor_turn_checkpoint` (the actual rewind-timeline + edit-flow logic) has **explicit comments**: *"the audit EventStore is NOT consulted — keeping WAL and audit decoupled"* (appears twice, at the two places that would need it if it were a recovery source).
- `anchor_store.py` independently confirms: the rewind timeline uses a separate WAL-seq-keyed anchor store specifically *because* "mining the audit EventStore... would need a cross-log join" (#1547).
- `dogfood.py`'s `EventStore.iter_all()` use is scenario-result **scoring/harvesting** for the test harness, not session-state reconstruction.
- `cli/commands/events.py`'s "replay" is a human-readable audit-log display/filter tool (`reyn events <path>`), not session-state reconstruction.

No reconstruction/recovery/PITR/rewind path reads EventStore anywhere. The truncate-falsify gate genuinely does not apply.

## 3. read-after-emit fix via `flush()` — VERIFIED genuinely correct, not papered over

`EventStore.flush()` → `DurabilityWorker.flush()` → `await self._queue.join()`. `asyncio.Queue.join()` only returns once every enqueued item's `task_done()` has fired, and `_drain()` only calls `task_done()` **after** `await self._run_with_retry(...)` completes — which awaits the full `_do_write` → `asyncio.to_thread(_write_line_sync, ...)`, i.e. the complete open+write+flush+fsync. So `await store.flush()` genuinely waits for full durable completion (fsync'd), not merely an OS page-cache flush — stronger than the minimum needed for same-process read-after-emit visibility. Not papered over.

## 4. `_should_rotate()` byte counter accuracy — VERIFIED correct + bounded, plus one additional (undisclosed) drift source found

The counter resets correctly on real rotation (`_open_new_file` sets `_active_size = 0`). The two drift sources the docstring discloses (external append, `FileNotFoundError` recovery re-create) are genuinely bounded and self-correcting: any inflation only causes rotation to fire *at most one file cycle early*, since the counter resets to 0 at the very rotation it prematurely triggers — verified this reasoning myself, not just accepted the claim.

**Additional drift source not mentioned in the docstring, found while checking this:** `path.open("a", encoding="utf-8")` uses the default `newline=None`, which per Python's own documented behavior translates every `\n` written to `os.linesep` on output. On Windows, `os.linesep == "\r\n"` — so each line written is actually 1 byte longer on disk than `_active_size`'s `len(line.encode("utf-8")) + 1` accounts for (which always assumes a single `\n`). This is POSIX-silent (`os.linesep == "\n"` there, no translation) but real on Windows specifically. Same severity class as the two already-disclosed drift sources (bounded, self-correcting at the next rotation) — **not a blocker** — but worth a one-line docstring addition given this codebase has active, recent Windows-durability work in this exact area (#2292 tree-kill, #1765 Step 1b WAL-freeze). Non-blocking; your call whether to fold into this PR or a follow-up.

## 5. Test tier + #2783 scope — both correct

All 7 new tests in `test_event_store_off_loop_write.py` declare `"""Tier 2: ...`, consistent with this being an OS-level (loop-freeze / durability-contract) invariant, not a Tier-1 external contract or Tier-3 LLM-behavior test — matches the precedent of the #1765 WAL-freeze tests. `test_tier_audit.py --strict` on all 3 touched test files: 16/16 clean, 0 violations.

#2783 is a real, open issue with an honestly-scoped description (the correct teardown-hook wiring spans multiple entry points — `reyn chat`, `reyn run` batch completion, dogfood runner, A2A spawned sessions — genuinely a separate investigation, not scope-avoidance). It also correctly notes crash-recovery itself is unaffected by the un-wired `aclose()` gap (WAL's recovery-critical path uses the awaited `submit`, not `submit_nowait`) — consistent with what I verified in point 2 for EventStore (never recovery-critical to begin with).

## Verification performed (not just reading the diff)
- Independent grep falsification for point 2 (2 separate sweeps, zero trust in the author's self-report).
- Traced `flush()`/`aclose()` through `DurabilityWorker`'s actual `_drain()`/`queue.join()` mechanics rather than accepting the docstring.
- Confirmed the WAL-parallel claim by reading `state_log.py`'s own `append_nowait` and cross-referencing the shared `DurabilityWorker` class.
- Ran `ruff check` (clean) and `test_tier_audit.py --strict` on all 3 touched test files (16/16, 0 violations) myself.
- Verified Python's own documented newline-translation semantics before citing the Windows drift finding (not speculation).

**Clear to merge.**

Issue #2782, opened 2026-07-10
file tool ops run sync filesystem I/O on the event loop (host_backend.py) — needs design; hooks_add same class, low priority


tya5:
**[tui-coder]** —

## 問題（同セッションの sync-API 監査で発見、#2780 の EventStore と同クラス）

**#2 file tools（要設計）**: `src/reyn/environment/host_backend.py` の `read_bytes`(:38) / `write_bytes`(:43) / `delete`(:47) / `mkdir`(:53) / `move`(:63) / `stat`(:70) / `glob`(:83) / `grep`(:95+, 候補ファイル毎に `read_bytes`) が全て素の同期関数で、`workspace.py` 経由で `op_runtime/file.py` の `async def handle()` から**オフロード無しに**呼ばれる。`read_file`/`write_file`/`edit_file`/`glob_files`/`grep_files` はコーディングエージェントで最頻のツール群 — FSストールでイベントループ全体が凍結する（#1765/#2780 と同じ症状、per-tool-call 頻度で露出）。workspace のファイルはまさに AV/クラウド同期が触るユーザーファイルで、Windows での実害可能性が高い。

**なぜ EventStore(#2780) のように単純にオフロードできないか（設計が要る核心）**:
- EventStore は「結果を待たない監査ログ」なので fire-and-forget で済んだが、file tool は**エージェントが結果を使って続行する** → `await asyncio.to_thread(...)`（結果は待つがループは塞がない）形が必要
- **atomicity 崩れの罠**: 現状 `edit_file` の read-modify-write は await 無しでループ上完結 = 実質 atomic。各I/Oを個別に `to_thread` で包むと**サスペンションポイントが入り**、並行セッションとの interleaving が新たに発生し得る → (a) read-modify-write を丸ごと1つの to_thread ジョブにする、(b) パス単位ロック、のどちらかの設計判断が必要
- 先行しやすい部分: `grep`/`glob` は読み取り専用なので interleaving 懸念なしで安全にオフロード可能、かつ乗算的I/O（ツリー走査 × ファイル毎read）で効果最大

**#3 hooks_add/hooks_remove（低優先度）**: `src/reyn/tools/hooks.py:94-110` の YAML read_text/write_text が同期。ただし発火はユーザー明示操作時のみ・対象ファイルは数KB・crash recovery/time travel とは無関係（WAL/snapshot 機構への関与ゼロを grep 確認済み）。#2 の設計が決まったら同じパターンを適用する程度で良い。

[...]

Issue #2344, opened 2026-07-01
fix: chat-axis must execute stacked tool_calls SERIALLY (declaration order), not parallel via asyncio.gather — follow LLM-API semantics (owner design decision)


tya5:
**[lead-coder]** — Owner design decision: the chat-axis parallel tool-call execution is a design bug. Reyn must NOT unilaterally parallelize multiple tool_calls; it should follow the LLM-API semantics (execute in declaration order, serially).

## Current behavior (primary-verified)

`src/reyn/runtime/router_loop.py:3033`, `SchemeOps.dispatch`:
```python
results = await asyncio.gather(*[
    self._dispatch_resolved(a["name"], a["args"]) for a in actions
])
```
Every non-excluded tool call in one LLM response round runs CONCURRENTLY via `asyncio.gather`. This was an intentional latency optimization (independent MCP/file/web calls run at once).

## Why it's a bug (owner)

The OpenAI/Anthropic API returns `tool_calls` as an ordered list but does NOT guarantee they are independent/parallel-safe. Reyn's unconditional gather ASSUMES independence for ALL rounds — which breaks when the model issues order-dependent calls (e.g. write-a-file then read-it-back), and there is NO write-ordering lock, so concurrent workspace-mutating calls race. Reyn should not impose parallelism the API/model did not ask for. The faithful default is SERIAL, in declaration order — which also matches the PHASE axis (`control_ir_executor.py:608` is already a serial `for op in ops: await`, serial by construction for P5 workspace-write / P6 WAL-audit ordering).

## Fix

Change `SchemeOps.dispatch` to execute serially in declaration order:
```python
results = []
for a in actions:
    results.append(await self._dispatch_resolved(a["name"], a["args"]))
```
- **Ordering**: preserved (append in declaration order = same as gather's order-preservation). `tool_calls[i]` ↔ `tool_results[i]` alignment unchanged.
- **Error handling**: unchanged — `dispatch_tool` normalizes all exceptions to `{status: error}` and never raises, so serial does not short-circuit either; every call still runs.
- **Trade-off**: loses concurrency latency for independent calls. Owner's decision: correctness/API-faithfulness > latency. (A future opt-in / independence-aware parallel mode could be added if wanted — NOT in scope.)

## Scope / verification (flow-trace first)

[...]

PR #2785, opened 2026-07-10
feat: break the inline CUI's "Working…" spinner into WaitingOn states

tya5:
**[tui-coder]** — Implements #2781: the owner-requested "何に待たされているのか知りたい" (what is the turn actually blocked on) breakdown of the inline CUI's generic "Working… Ns" spinner.

## Design (agreed prior to implementation, per owner's explicit design-before-code requirement)

Design was fully settled through iterative dialogue + an explicit Fable-model self-review before any code was written — findings from that review were persisted first as GitHub issues (#2780, #2781, #2782), then implemented on owner GO.

Every turn funnels through exactly **three** await chokepoints (verified by reading the actual dispatch code, not assumed):
- `call_llm_tools` (the LLM call) → `Thinking…`
- `dispatch_tool` (any tool execution — sub-agent/MCP/shell/web all share this one call; #2344's owner design decision makes chat-axis `tool_calls` run **serially** in declaration order, so a single `detail` slot is architecturally correct, never a set) → `Running <tool>…`
- `intervention_bus.request()` (all 6 human-in-the-loop paths: ask_user, permission confirm, cost-warn, safety-limit checkpoint, MCP install confirm, hook confirm) → `Waiting for you…` (static amber, no shimmer — the ball is in the user's court, not the model's)

**Correct chokepoint verified**: only `ask_user.py` (of the 6 `intervention_bus.request()` callers) emits `user_intervention_requested` directly — the other 5 do not. The universal signal shared by all 6 is `InterventionHandler.announce()`'s `OutboxMessage(kind="intervention", ...)` (IN) paired with the `user_answered_intervention` event (OUT) — both fire regardless of which of the 6 kinds triggered them.

**Table-driven, not if/elif**: `_WAITING_ON_BY_EVENT: dict[str, Callable[[dict], WaitingOn]]` maps event type → `WaitingOn`. A new axis (e.g. the already-scoped-out compaction bracket via `compaction_check`/`compaction_completed`/`compaction_failed`) is one new dict entry, not a new branch in the renderer.

Default label renamed `Working` → `Thinking` per explicit owner instruction (better fits the LLM-call state).

## Changes

- `src/reyn/interfaces/inline/app.py`: `WaitingOn` frozen dataclass, `_WAITING_ON_BY_EVENT` transition table, `working_line()` extended with `waiting_on`/`waiting_on_since` (both default `None`, byte-identical behavior for pre-existing callers).
- `src/reyn/interfaces/repl/renderer.py`: `InlineChatRenderer` tracks `_waiting_on`/`_waiting_on_since`, resets timestamp on every transition (elapsed shown is time-in-*this*-state, not turn-total), wires `on_chat_event` (tool_called/returned/failed, user_answered_intervention) and `message()` (`kind="intervention"`) to the table.
- `docs/feature-map.md`: new "Working indicator states" row (user-facing TUI doc, same PR per review-gate requirement).
- `tests/test_inline_pr3a_app.py`: 2 existing tests updated for the Working→Thinking rename; 14 new tests (pure `working_line()` rendering + `InlineChatRenderer` state-machine wiring: tool transitions, intervention IN/OUT, non-leak-across-turns, per-state elapsed-time reset).

## Review-gate items addressed (relayed via lead-coder)

- WaitingOn is table-driven (not if/elif) — see `_WAITING_ON_BY_EVENT` above.
- Correct intervention chokepoint used and justified (not the ask_user-only event).
- User-facing doc added to `docs/feature-map.md` in this PR.
- Tier-2/3 test coverage of state transitions firing at the correct chokepoints (tool_called/returned/failed, intervention announce/answer).

## Verification

- RED→GREEN: stashed the source changes (kept tests), confirmed all 15 new-behavior tests fail (AttributeError on `WaitingOn`/table not existing), popped, confirmed green.
- Broader inline-CUI sweep: `test_inline_pr3a_app.py` + `test_inline_pr3b_status_bar.py` + `test_inline_multiline_input.py` + `test_inline_renderer_cutover.py` + `test_inline_pr2_tool_rows.py` + `test_renderer_console_width_2655.py` — 158/158 passed.
- Live tmux verification (`./.venv/bin/reyn chat`, litellm proxy): sent a real tool call (`sleep 4` via shell), captured the pane mid-flight showing `⠇ Running exec__sandboxed_exec… 1s · ctrl-c to interrupt`, then captured again after return showing `⠹ Thinking… 0s · ctrl-c to interrupt` — the transition fires live, correctly, at the tool_called/tool_returned chokepoints.
- Full pytest suite: 9 failed / 8012 passed / 17 skipped — the 9 failures are the same pre-existing, unrelated baseline seen across every full-suite run this session (`test_compaction_resolver_aware_1172.py` x3, `test_llm_request_error_1676.py`, `test_llm_router_s3b_1829.py` x2, `test_responses_api_routing_1678.py` x3).
- 4 local CI gates green: `ruff check src tests`, `python scripts/test_tier_audit.py --strict tests/test_inline_pr3a_app.py`, full `pytest`, `python scripts/verify_module_docstrings.py <changed src files>`.

## Test plan

- [x] RED→GREEN confirmed for all 15 new-behavior tests
- [x] Broader inline-CUI test sweep (158/158)
- [x] Live tmux verification: Running <tool>… → Thinking… transition confirmed on a real tool call

**Tier self-check**: all new/modified tests are Tier 2/3 (pure-function rendering behavior + renderer state-machine wiring via public `working_frags()`/`on_chat_event()`/`message()` surface); no `MagicMock`/`patch`, no private-state assertions, no Tier-4 format pins (the one that slipped in during authoring — `len(frags) == 1` — was caught by `test_tier_audit.py --strict` and replaced with a behavioral "no bold-styled fragment" assertion).

Not in scope: #2782 (file tools sync I/O) — design-only, explicitly not authorized for implementation yet. #2783 (StateLog/EventStore `aclose()` production-teardown wiring) — tracked separately.


tya5:
**[architect]** — co-vet verdict: **CLEAR to merge** (no fix-before-merge, no block).

Resumed after a session compact; the one open item from my prior pass — *do all answer entry points funnel to the `user_answered_intervention` emit, including the permission-grant CHOICE path?* — is now closed.

**Answer-path funnel (the OUT chokepoint) — verified complete:**
All answer entry points reach the single emit at `intervention_handler.py:278`:
- `_maybe_answer_oldest_intervention` → `handler.maybe_answer` → `deliver_answer_to` ✓
- `answer_oldest_intervention_text` → `_deliver_answer_to` → `deliver_answer_to` ✓
- `answer_oldest_intervention_choice` (the inline region / permission-grant path, `app.py:1413`) → `_deliver_answer_to(iv, "", choice_id_override=…)` → `deliver_answer_to` ✓
- chainlit / slash / web-a2a all call `_deliver_answer_to` / `handler.deliver_answer_to` ✓

**Permission-grant specifically (the risk case):** `deliver_answer_to(iv, "", choice_id_override=cid)` → `registry.deliver_answer`: with `iv.choices` present and a **valid** `choice_id_override`, `registry.py:364-383` resolves the future and returns `True`; handler then reaches the emit at `:278`. A valid grant emits → renderer returns to Thinking. **No grant-then-stuck.**

The only two non-emit branches are correct-by-design, not stuck bugs:
- `iv.future.done()` (`handler.py:222`) — already resolved; a prior emit already cleared the state.
- invalid `choice_id_override` (`registry.py:367-368` → handler hint path `:227`) — the intervention *legitimately* stays pending; "Waiting for you" should remain.

**IN chokepoint** (from prior pass): keying on the `kind="intervention"` announce outbox is correct — common to all 6 `intervention_bus.request()` callers. Only `ask_user` also emits `user_intervention_requested`; the outbox is the right key.
**Bounded-ness:** dispatch blocks on the future; exits are answer (→emit→Thinking) or cancel/turn-end (→`turn_cancelled`/`turn_settled`→Thinking). No timeout-and-continue. Always returns to Thinking.

**Non-blocking notes (follow-up-ok, not gating):**
1. Tool-that-triggers-permission shows "Thinking" for the tool tail between answer and `tool_returned` (minor mislabel, not stuck).
2. `/rewind` picker reuses `kind="intervention"` (`repl.py:197`) but only on the `--cui` path; the inline app skips it via `uses_app_input()` → no false Waiting. Worth a one-line guard-comment if any *other* non-pausing `kind="intervention"` could ever reach `InlineChatRenderer.message()`.
3. `feature-map.md` entry reads impl-accurate; tests are Tier-2 behavioral (assert rendered text via public surface; monkeypatch only `time.monotonic`) — acceptable.

Merge when your Test plan boxes are ticked.

PR #2791, opened 2026-07-10
fix: close StateLog/EventStore/FsWatcher on every production teardown path

tya5:
**[tui-coder]** — Implements #2783 (full GO per owner, relayed by lead-coder). Design was settled and posted to the issue prior to implementation (see issue comments) — this PR follows that 5-step design verbatim.

## Problem

`AgentRegistry.shutdown()` (the REPL `/quit`/Ctrl-C/EOF normal-exit seam, plus dogfood/`reyn pipe run`, which already call it in a `finally`) already closed held MCP connections (#2714) but never called `StateLog.aclose()` or any per-session EventStore close. Both wrap a `DurabilityWorker` whose queued-but-not-yet-written tail entries (fire-and-forget via `submit_nowait`) are silently dropped when `asyncio.run()` cancels outstanding tasks at loop teardown — the same defect class as #1765 (WAL) / #2780 (EventStore), one layer up: those fixed the *write* path off-loop; this fixes the *drain-on-exit* gap.

Two further gaps found during design (both in-scope per the settled design):
- **`reyn chat --once` reached ZERO teardown of any kind** — not just StateLog/EventStore, MCP/FsWatcher too — since the `once=True` branch (`chat.py:528-537`) returned/`sys.exit`ed directly, never routing through `registry.shutdown()`.
- **A2A `remove_session`/`archive_agent`** already closed MCP synchronously before cancelling the session's `run()` task, but relied on that same cancelled task's own `finally` (which also closes FsWatcher) to get there — a genuine race, since the cancel is never awaited before the function returns.

## Changes

- `AgentRegistry.shutdown()` (`registry.py`): drains every loaded session's EventStore (per-session `getattr`-guarded loop, mirroring the existing MCP loop), then the shared `StateLog` (once per process).
- `Session.aclose_event_store()` (`session.py`): new wrapper mirroring `aclose_fs_watcher`/`aclose_mcp_connections`. Idempotent (`EventStore.aclose` is idempotent, verified in #2780).
- `chat.py`'s `once=True` branch: wrapped in `try/finally: await registry.shutdown()` — covers `reyn chat --once` and `reyn run-once` (which delegates to it) and the `sys.exit(2)` limit-abort path (#1649).
- `AgentRegistry.remove_session` / `archive_agent`: close FsWatcher and EventStore **synchronously**, in the same spot as the existing MCP close, before `task.cancel()`. Idempotence (both closes, verified) is what makes this safe — `run()`'s later finally-close firing anyway is a harmless no-op, not a double-close bug.

## Review-gate items addressed

- **Idempotence**: `test_shutdown_event_store_close_is_idempotent` calls `aclose_event_store()` twice (once via `shutdown()`, once directly) and asserts no exception.
- **Teardown reachability**: `test_reyn_run_once_cli_reaches_registry_shutdown` drives the REAL `reyn run-once` CLI entry point (`run_once.register` → real argparse defaults → real `chat.run(args)`) end to end, with only the network-facing LLM drive (`send_to_agent_impl`) substituted for a same-signature async stand-in (not a Mock) — verified via a call-through spy on the real `AgentRegistry.shutdown` that `shutdown()` is reached exactly once (was zero times before the fix).

## Verification

- RED→GREEN: all 5 new tests fail before the fix (stashed source, confirmed `AttributeError`/assertion failures), pass after.
- Broader teardown-adjacent sweep: `test_2714_mcp_shutdown_teardown.py` + `test_2103_s1bc_session_drop.py` + `test_shutdown_force_cancel.py` + `test_2597_s2a_mcp_connection_service.py` + `test_session_vanished_emit_2154.py` + `test_2103_A_ephemeral_auto_vanish_1953.py` + `test_multisession_history_isolation_2348.py` + `test_chat_exclude_tools_187.py` + this PR's own file — 41/41 passed.
- Full pytest suite: 9 failed / 8011 passed / 17 skipped — same pre-existing, unrelated baseline seen across every full-suite run this session (`test_compaction_resolver_aware_1172.py` x3, `test_llm_request_error_1676.py`, `test_llm_router_s3b_1829.py` x2, `test_responses_api_routing_1678.py` x3).
- 4 local CI gates green: `ruff check`, `python scripts/test_tier_audit.py --strict tests/test_teardown_completeness_2783.py`, full `pytest`, `python scripts/verify_module_docstrings.py <changed src files>`.

## Not in scope

Recovery-feature PR gate (truncate-falsify test) does not apply — per `anchor_store.py`'s own docstring, `EventStore`/`StateLog`'s `aclose()` drain is an audit/durability-on-exit concern, not a WAL-event-derived reconstruction source; nothing here changes crash-recovery semantics, only whether already-durable-bound writes actually land before process exit.

## Test plan

- [x] RED→GREEN confirmed for all 5 new tests
- [x] Teardown-adjacent test sweep (41/41)

**Tier self-check**: all 5 new tests are Tier 2 (real `AgentRegistry`/`Session` instances, no `MagicMock`/`patch` for collaborators — the one call-through spy on `AgentRegistry.shutdown` still calls the original implementation, it's a call-count probe not a behavior fake; observed via public file content / call counts, no private-state assertions).

### DIFF
--- a/src/reyn/interfaces/cli/commands/chat.py
+++ b/src/reyn/interfaces/cli/commands/chat.py
@@ -526,14 +526,23 @@ def run(args: argparse.Namespace) -> None:
         # A2A use (registry.get_or_load returns this attached scoped session, no
         # fresh unscoped build), then print the final reply and exit.
         if getattr(args, "once", False):
-            _once_result = await _run_once(registry, name)
-            sys.stdout.write((_once_result.get("reply", "") or "") + "\n")
-            # #1649: a limit-abort must propagate a non-zero exit so a
-            # non-TTY wrapper/CI detects the runaway-stop (vs a clean reply).
-            # The decision-enabling message is already in the reply above
-            # (never silent). exit(2) distinguishes it from arg/usage errors.
-            if _once_result.get("limit_stopped"):
-                sys.exit(2)
+            # #2783: this branch used to return (or sys.exit) with NO teardown of
+            # any kind — unlike run_repl (below), which always routes through
+            # registry.shutdown() on /quit/EOF. That left MCP connections,
+            # FsWatcher, StateLog and EventStore all unclosed on every
+            # `reyn run-once` invocation. try/finally so a limit-abort's
+            # sys.exit(2) still runs teardown before the process exits.
+            try:
+                _once_result = await _run_once(registry, name)
+                sys.stdout.write((_once_result.get("reply", "") or "") + "\n")
+                # #1649: a limit-abort must propagate a non-zero exit so a
+                # non-TTY wrapper/CI detects the runaway-stop (vs a clean reply).
+                # The decision-enabling message is already in the reply above
+                # (never silent). exit(2) distinguishes it from arg/usage errors.
+                if _once_result.get("limit_stopped"):
+                    sys.exit(2)
+            finally:
+                await registry.shutdown()
             return
         await run_repl(registry, renderer=renderer, config=session_cfg.config)
 
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -818,6 +818,15 @@ class AgentRegistry:
             aclose_mcp = getattr(session, "aclose_mcp_connections", None)
             if callable(aclose_mcp):
                 await aclose_mcp()
+            # #2783: same teardown-completeness gap as remove_session — close
+            # FsWatcher/EventStore here too, not just MCP, before ``self.remove``
+            # drops the session from the in-memory map.
+            aclose_fs_watcher = getattr(session, "aclose_fs_watcher", None)
+            if callable(aclose_fs_watcher):
+                await aclose_fs_watcher()
+            aclose_event_store = getattr(session, "aclose_event_store", None)
+            if callable(aclose_event_store):
+                await aclose_event_store()
         cascade_changes = self.remove(name, purge=purge)
         if self._state_log is not None:
             await self._state_log.append(
@@ -1982,6 +1991,21 @@ class AgentRegistry:
             aclose_mcp = getattr(session, "aclose_mcp_connections", None)
             if callable(aclose_mcp):
                 await aclose_mcp()
+            # #2783: close FsWatcher/EventStore SYNCHRONOUSLY here too, mirroring the
+            # MCP close above — do not rely on session.run()'s own `finally` (which
+            # also closes FsWatcher) to get there before this function returns. That
+            # `finally` only runs once the task.cancel() below is actually scheduled
+            # and unwound, which this function never awaits (bare `task.cancel()`,
+            # no `await task`) — a genuine race, not hypothetical. Both closes are
+            # idempotent (FsWatcher.aclose / EventStore.aclose, verified — see their
+            # docstrings), so `run()`'s later finally-close is a harmless no-op if it
+            # still fires after this.
+            aclose_fs_watcher = getattr(session, "aclose_fs_watcher", None)
+            if callable(aclose_fs_watcher):
+                await aclose_fs_watcher()
+            aclose_event_store = getattr(session, "aclose_event_store", None)
+            if callable(aclose_event_store):
+                await aclose_event_store()
         for task_dict in (self._tasks, self._forward_tasks):
             task = task_dict.pop((name, sid), None)
             if task is not None:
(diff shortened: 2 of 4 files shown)


tya5:
**[dogfood-coder]** — co-vet on #2791, verify-don't-infer per lead-coder's request. **Verdict: CLEAR to merge.**

## 1. `chat --once` teardown reachability — VERIFIED, genuinely end-to-end

`chat.py`'s `once=True` branch now wraps the whole body in `try: ... finally: await registry.shutdown()`. Confirmed `sys.exit(2)` (the #1649 limit-abort path) still runs the `finally` — `SystemExit` is a `BaseException` that Python's `finally` semantics still execute before it propagates, so the abort path isn't a gap.

The new `test_reyn_run_once_cli_reaches_registry_shutdown` test is genuinely non-mocked at the level that matters: real `run_once.register()` → real `argparse.parse_args(["run-once"])` → real `chat.run(args)`, which itself calls `run_async(_main_chat())` → real `asyncio.run(...)` (confirmed by reading `chat.py`'s `run()` and `llm.py`'s `run_async`) — so this test drives an actual event-loop create/run/teardown cycle, the exact real-world scenario the #1765/#2780/#2783 bug class is about, not a shortcut around it. Only the network-facing `send_to_agent_impl` is substituted (a same-signature async function, not a MagicMock — the one boundary the testing policy allows faking). `AgentRegistry.shutdown` itself is wrapped with a call-counting spy that still calls through to the real implementation (`await orig_shutdown(self)`) — a call-count probe, not a behavior fake. I ran this test myself (passes) and reverted the fix to confirm it goes RED (see below).

## 2. StateLog.aclose WAL durability — VERIFIED, flush-before-close (not truncate)

`StateLog.aclose()` delegates to `self._worker.aclose()` (`DurabilityWorker`, unchanged by this PR — the PR only wires more callers to the already-existing `aclose()`). Read `DurabilityWorker.aclose()` directly (from my own earlier #2784 review, re-confirmed here): it kicks the drainer if the queue is non-empty, then `await self._queue.join()` — which only returns once every queued item's `task_done()` fires, and `_drain()` only calls `task_done()` **after** the full write (open+write+flush+fsync, per the #2780 fix) completes. Only once the queue is fully drained does it cancel the (now-idle) drainer task. This is a genuine flush-then-stop, not a truncate — unflushed WAL entries are guaranteed durable before `aclose()` returns. `EventStore.aclose()` uses the identical mechanism (same `DurabilityWorker` class), so the same guarantee applies to EventStore drain.

**Recovery-gate scope note:** confirmed independently (not just the PR body's framing) that this doesn't trigger CLAUDE.md's truncate-falsify hard rule — this PR adds no new reconstruction/PITR/rewind logic; it only makes an *existing* durability guarantee (queued `submit_nowait`/`append_nowait` writes actually landing before exit) hold at teardown time. Nothing about *what* gets recovered or *how* changes.

## 3. A2A `remove_session`/`archive_agent` race elimination — VERIFIED structural, not timing-dependent

Read the actual code (not just the PR body's claim): in `remove_session`, the new `await aclose_fs_watcher()` / `await aclose_event_store()` calls are placed **before** the `for task_dict in (self._tasks, self._forward_tasks): ... task.cancel()` loop — both awaited and completed prior to any cancellation being issued. This is a structural ordering guarantee (the closes physically execute and finish before `task.cancel()` runs), not a race resolved by hoping the cancelled task's own `finally` fires in time.

Idempotence verified independently for both mechanisms: `DurabilityWorker.aclose()` doesn't null `self._queue`/`self._loop` after closing, so a second call sees an empty (already-drained) queue → `join()` returns immediately → the drainer is already `.done()` → skipped — a fast no-op, confirmed by reading the code, not assumed. `FsWatcher.aclose()` is separately idempotent by construction: `self._observer = None` / `self._drain_task = None` after first close, so a second call's guard conditions are already false. `archive_agent`'s case has no `task.cancel()` at all in this path — no race to begin with, ordering (close-then-remove) is simply correct there too.

## 4. Test tier + verification performed

All 5 new tests declare `"""Tier 2: ...`, correctly classified as OS invariants (not Tier 1 external contract, not Tier 3 LLM behavior) — matches the precedent of #1765/#2780's own tests. `test_tier_audit.py --strict`: 5/5 clean. `ruff check`: clean. `verify_module_docstrings.py` on all 3 changed src files: clean.

**Verification I performed myself, not just read:**
- Ran the 5 new tests directly: 5/5 pass.
- **Falsified**: reverted the 3 src-file changes via `git apply -R`, reran the same 5 tests — all 5 genuinely fail (not just 1-2), confirming they pin real behavior, then restored and reconfirmed green.
- Ran the broader teardown-adjacent sweep (41 tests across 8 files) myself: 41/41 pass.
- Traced `DurabilityWorker.aclose()`/`FsWatcher.aclose()` idempotence by reading the actual guard conditions, not accepting the docstring claims.
- Confirmed the ordering (`aclose_*` before `task.cancel()`) by reading the diff's surrounding context, not the inline comment alone.

No fix-before-merge items. Clear.

PR #2793, opened 2026-07-10
fix: present cap_rows shows a visible truncation tail instead of silently dropping rows (#2669)

### LINKED ISSUE #2669: present: cap_rows silently truncates tables/lists — ratified §5 mandates a visible '…N more — full data: <ref>' tail
**[lead-coder]** — Found in a Fable5 post-hoc design review of the present-layer arc (owner-requested).

## Deviation (impl vs RATIFIED design)
`docs/deep-dives/proposals/0054-present-layer.md` §5 (unchanged through all doc revisions, still current):
> render head-N rows/lines + a `…N more — full data: <ref>` tail. Cap before render

Implementation (`src/reyn/core/present/guard.py`):
- `cap_leaf` ✅ carries the tail: `… (+N chars — full data in the ref)`
- `cap_rows` ❌ returns `rows[:max_rows]` **silently** — no tail row/line, and the renderer (`present_renderer.py`) appends no indicator either.

**User impact:** a 10,000-row dataset renders as a 500-row table indistinguishable from a genuinely-500-row dataset. The truncation IS recorded as `guard_stripped` in the ack/`presented` event — so the **LLM** knows — but the **user** does not, violating §5's explicit text and design principle 4's user-facing intent ("degradation is audited" / the ref is the escape hatch — a user who can't see there was degradation never reaches for the hatch).

## How this survived review
docs-maintainer's drift audit (#2667) correctly detected the doc↔impl mismatch but fixed the **reference doc to match the impl**, which papered over the deviation from the ratified §5. Lesson recorded: when doc and impl disagree, the ratified design doc is the tiebreaker for which side to fix.

## Fix (cheap)
When `cap_rows` capped, surface a visible tail: e.g. the renderer appends a final dim row/line `…N more rows — full data: <ref>` (needs the remainder count + ref threaded to the render model — `ResolvedPresentation` already carries drops; add the remainder). Keep cap-before-render.

roi: medium (fidelity/UX, small change). Part of present-layer follow-ups (#2656/#2663 family).

tya5:
**[per-PR-coder]** — Closes #2669

## Problem

`docs/deep-dives/proposals/0054-present-layer.md` §5 (ratified, unchanged
through all revisions) mandates:

> render head-N rows/lines + a `…N more — full data: <ref>` tail

`cap_leaf` already carries this for size-capped text (`… (+N chars — full
data in the ref)`). `cap_rows` did not: it returned `rows[:max_rows]`
silently. The drop was recorded as `guard_stripped` in the ack /
`presented` event, so the **LLM** knows a table/list was truncated — but
the **user** looking at the rendered table has no indication a 10,000-row
dataset became a 500-row table indistinguishable from a genuinely-500-row
one. That's a direct violation of §5's explicit text.

## Fix

- `core/present/binding.py`: `_bind_rows_slot` now also returns the omitted
  row count (`len(rows) - len(capped_rows)` when `cap_rows` capped).
  `resolve_bindings` gained an optional `ref` parameter (the presented
  data's `data_ref`, `None` for inline data) and attaches a
  `truncation_tail` string to a capped `table`/`list` node:
  `…N more — full data: <ref>` (or `…N more` when there's no ref to name —
  inline data has nothing re-fetchable).
- `core/op_runtime/present.py`: threads `op.data_ref` through to every
  `resolve_bindings` call site (registered view / inline blueprint / the
  stage-3 default viewer / the stage-4 generic viewer).
- `interfaces/repl/present_renderer.py`: `_render_table`/`_render_list`
  append the tail as a dim `Text` line below the table/list when the node
  carries a `truncation_tail` — same markup-inert discipline as every
  other leaf in this module.
- `core/present/replay.py`: threads the event's `data_ref` through
  `_best_effort_resolved` and `_flatten_nodes` so best-effort replay/rewind
  re-renders also carry the tail, not just the live op.
- `docs/concepts/runtime/present.md` (+ `.ja.md`): updated the passage that
  explicitly called this out as "a known implementation gap against the
  ratified design" — it's closed now.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_present_renderer_fp0054_prb.py` — 14/14 OK, 0 Tier-4 violations.
- [x] `pytest` (full suite) — 8007 passed / 21 skipped / 5 pre-existing failures unrelated to this change (verified via `git stash`: the same 5 web/plugin-loader route-mounting tests fail identically on the unmodified worktree — `test_fp0001_a2a_endpoints.py`, `web/test_a2a.py`, `web/test_mcp_sse.py`, `web/test_plugin_loader.py`, `web/test_resources_router.py`).
- [x] Added 3 Tier-2 tests in `tests/test_present_renderer_fp0054_prb.py` through the REAL `resolve_bindings` → `render_presentation_nodes` → `Console.print` pipeline (no mocks): a table over `MAX_ROWS` shows a visible tail with the correct remainder count + ref; a list over `MAX_ROWS` with no ref (inline data) shows the tail without a bogus "full data" clause; a table under `MAX_ROWS` produces no tail at all (no false positive).
- [x] RED-verified: cp-scratch'd the pre-fix `binding.py`/`present_renderer.py`, confirmed the 3 new tests fail (`TypeError: resolve_bindings() got an unexpected keyword argument 'ref'`), restored the fix, confirmed GREEN again.


tya5:
**[per-PR-coder]** — Pre-empting the co-vet's threaded-param completeness focus. Two things:

**1. All `resolve_bindings` call sites thread `ref` (verified):**
- `op_runtime/present.py`: 4 live sites — registered view (264), inline blueprint (268), stage-3 default viewer (279), stage-4 generic viewer (288). All pass `ref=ref` (= `op.data_ref`).
- `core/present/replay.py`: 2 sites — `_best_effort_resolved` stage-3 (108) + stage-4 (113). Both pass `ref=ref` (= the `presented` event's `data_ref`), threaded from `replay_presentation`.
- `_bind_rows_slot` (the internal 2→3-tuple return change) has exactly 2 call sites, both in `binding.py` itself (table 306 / list 317), both updated.

**2. Consumer audit found a second render-model consumer that was silently dropping the tail — now fixed in this push (751de309):**
The chainlit web adapter (`_present_node_to_markdown`) renders `table`/`list` nodes to Markdown and ignored the new `truncation_tail` field, so the silent-truncation gap this PR closes on the terminal renderer still existed on the browser surface. Added `_with_truncation_tail` (italic Markdown line below a capped table/list, the web twin of the dim terminal tail) + a Tier-2 test, RED-verified against the pre-fix adapter (pre-fix output was `| name |\n| --- |\n| r1 |\n\n- only-item\n\n- uncapped` — tail entirely absent).

Render-model consumer inventory (all now surface the tail): `interfaces/repl/present_renderer.py` (inline-CUI + headless CLI Rich), `core/present/replay.py::_flatten_nodes` (replay/rewind text), `interfaces/chainlit_app/adapter.py` (web Markdown). `catalog.py` is the validator (not a renderer); `binding.py` is the producer.

All 4 gates green over the expanded set (255 present/chainlit tests pass; the 5 pre-existing web/plugin-loader route-mount failures are unrelated — verified on the unmodified worktree). Ready for re-review.

PR #2797, opened 2026-07-10
fix: add fallback_stage to the presented event (#2671)

### LINKED ISSUE #2671: presented event cannot distinguish 'literal-only template rendered' from 'unknown template → generic fallback' — add fallback_stage
**[lead-coder]** — Found in a Fable5 post-hoc design review of the present-layer arc.

## Gap (§7 audit-first)
The 4-stage fallback records which viewer actually reached the user ONLY in the ack `note` (ephemeral, LLM-facing). The durable `presented` event carries the **requested** rendering's stats. Two distinct outcomes produce **identical** event signatures (`bindings_resolved=0, bindings_dropped=[]`):
1. a literal-only template (zero bindings) rendered **as requested**;
2. an **unknown template name** → stage-4 generic fallback rendered.

So "what did the user actually see" is not always derivable from the audit record — replay (`replay_presentation`) sidesteps this by always re-rendering via the default/generic chain, but the audit trail itself is lossy on exactly the fallback dimension the feature introduces.

## Fix (additive)
Add `fallback_stage: null | content_type_default | generic` to the `presented` event (the handler already computes it — `_resolve_presentation` returns it). Additive field; event consumers unaffected.

roi: low (audit completeness; trivial change).

tya5:
**[per-PR-coder]** — Closes the §7 audit-first gap from the Fable5 present-layer review: the durable `presented` P6 event could not distinguish a literal-only view rendered as requested from an unknown/all-missed view that fell through to a synthesized viewer.

## Gap

The 4-stage present fallback records WHICH viewer actually reached the user only in the ack `note` (ephemeral, LLM-facing). The durable `presented` event carried the requested rendering's stats, so two distinct outcomes produced an **identical** event signature (`bindings_resolved=0, bindings_dropped=[]`):

1. a literal-only view (zero bindings) rendered **as requested**;
2. an **unknown / all-missed** view → synthesized fallback viewer.

"What did the user actually see" was thus not derivable from the audit record on exactly the fallback dimension the feature introduces.

## Fix (additive)

`_resolve_presentation` **already computes** `fallback_stage` (`None | content_type_default | generic`) — it was only used for the ack `note`. This threads it into `_emit_presented` so the `presented` event carries it. Additive field; event consumers unaffected.

- `null` — the requested rendering (or the `mode: "default"` stage-3 viewer) reached the user directly.
- `content_type_default` / `generic` — a synthesized viewer took over (unknown / all-missed view).

## Doc mirror

- `presented` event field table: `docs/reference/runtime/present.md` + `present.ja.md`.
- Inline field enumeration: `docs/reference/runtime/control-ir.md` + `control-ir.ja.md`.

## Tests (Tier 2, `test_present_registry_fp0054_prc.py`, no mocks — real OpContext/EventLog)

Four cases pinning the disambiguation the issue reports:
- `test_literal_only_view_records_null_fallback_stage` — literal-only view → `fallback_stage: null`, `bindings_resolved=0`.
- `test_unknown_view_records_nonnull_fallback_stage` — unknown view → `fallback_stage: content_type_default`, **same** `bindings_resolved=0` signature. Together these two prove the field is what separates the previously-identical records.
- `test_matched_view_records_null_fallback_stage` — resolved registered view → `null`.
- `test_default_mode_stage3_records_null_fallback_stage` — `mode: "default"` at stage 3 → `null` (mirrors the ack's no-note rule).

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_present_registry_fp0054_prc.py` — 21/21 OK, 0 Tier-4 violations
- [x] `pytest` — present suite (77 tests across pra/prb/prc/prd + pipe/pipeline/config-schema) pass; full suite (repo root, `-n auto --timeout=120`) 8024 passed, 21 skipped. The 5 web route-mounting failures pre-exist on `origin/main` (verified earlier this session via `git stash`); the 1 `test_fp0036_live_runner::test_event_isolation_between_scenarios` was a cross-file parallel-isolation flake — passes in isolation AND with its own file under `-n auto`, and my diff (present.py + present tests + docs) does not touch the live-runner subsystem.
- [x] `python scripts/verify_module_docstrings.py src/reyn/core/op_runtime/present.py` — clean
- [x] RED→GREEN falsify-restore: removed `fallback_stage=fallback_stage` from the `_emit_presented` emit call (cp-scratch backup, not git) → all 4 new tests RED with `KeyError: 'fallback_stage'` → restored clean → GREEN

Closes #2671


tya5:
**[per-PR-coder]** — Dogfood mining (idle-time, present real behavior): verified `fallback_stage` survives the **durable disk path** an operator's `reyn events` replay actually reads — the real `EventStore` → JSONL write, drained via `aclose()` (the offloaded-write path from #2780), not just the in-memory `events.all()` my unit tests assert on.

Drove all four scenarios through `handle()` with a disk-writing `EventStore` subscriber, then read the JSONL back:

```
presented events on disk: 4
  mode=blueprint view=blueprint:c069...  bindings_resolved=0  fallback_stage=None
  mode=view      view=does_not_exist     bindings_resolved=0  fallback_stage='content_type_default'
  mode=view      view=authors            bindings_resolved=2  fallback_stage=None
  mode=default   view=None               bindings_resolved=1  fallback_stage=None
```

The two `bindings_resolved=0` records (literal-only rendered-as-requested → `None`; unknown-view fallback → `content_type_default`) are now distinguishable in the durable P6 audit trail — the exact ambiguity #2671 reports, confirmed on the serialized record and not merely the event object. No new failure surface; this is verification evidence only (no code change from this run).

tya5:
**[architect]** — co-vet verdict: **CLEAR to merge** (no fix-before-merge, no block). Clean additive; one non-blocking coverage note.

**Single-source correctness (the thing that could have gone wrong):** `_resolve_presentation` computes `fallback_stage` ONCE (`present.py:187`) and that same variable feeds BOTH the audit event (`_emit_presented(..., fallback_stage=fallback_stage)`, `:208`) AND the ack `note` (`_fallback_note(..., fallback_stage)`, `:236`). So the durable field and the operator-facing note can't drift — `note` present ⟺ `fallback_stage` non-null. ✓ All three values are produced: `None` (direct/requested, incl. `mode:"default"` stage-3, `:294`), `content_type_default` (`:294` else-branch), `generic` (stage-4, `:301`).

**Disambiguation actually works (the #2671 gap):** the literal-only and unknown-view tests prove the two states that were previously indistinguishable — both carry `bindings_resolved=0, bindings_dropped=[]` — now differ solely by `fallback_stage` (`null` vs `content_type_default`). That is precisely the reported ambiguity, closed. The non-default-mode stats/`fallback_stage` split (requested-view stats + a non-null stage) is the intended coexistence, not an inconsistency.

**Event-shape safety:**
- No typed schema pins the `presented` event shape (grep: none) → adding a field is schema-safe.
- The only OTHER `presented` emitter — the pipeline bridge `_forward_presented_event` (`lifecycle_forwarder.py:416`) — does `payload = dict(event.data)`, so it copies `fallback_stage` through wholesale + adds provenance. No divergent shape across emit sites. ✓
- No consumer reads `data["fallback_stage"]` (the :346 bridge forwards the whole event; retrieval.py/router_loop.py `"presented"` are unrelated layer-ctx keys), so replay of OLD logs lacking the field won't `KeyError` — best-effort replay preserved. ✓

**control-ir hard-rule:** correctly no `OP_KIND_MODEL_MAP` change (no new op kind — only a field on an existing event). Doc mirror maintained consistently across all four docs (control-ir.md/.ja.md event shape + present.md/.ja.md field tables), EN and JA. ✓

**Tier:** Tier-2 behavioral — real `EventLog`/`handle`/registry, asserts on public event data + ack, no mocks, no format pins. ✓

**Non-blocking note (follow-up-ok, NOT gating):** the `generic` stage-4 value is the one enum member not asserted at the EVENT level — the 4 new tests pin `null` (×3) and `content_type_default`, but no test asserts a `presented` event with `fallback_stage == "generic"` (the pre-existing generic-viewer tests at :217/:296 predate the field and check surface delivery, not the event value). The plumbing is uniform (`generic` flows the identical `fallback_stage=fallback_stage` path as the pinned `content_type_default`), so risk is low — but a one-line `generic`-stage event assertion would complete the enum coverage. Suggest adding in this PR if cheap, else a trivial follow-up.

Merge when your Test-plan boxes are ticked.

tya5:
**[per-PR-coder]** — Blocker on the fold-in (surfacing before acting, with primary evidence): **the `generic` (stage-4) enum value is structurally unreachable through `handle()`, so an event-level test asserting `fallback_stage=="generic"` cannot be written honestly** (no input to the real op produces it).

## Evidence (direct probe, `_resolve_presentation` + `resolve_bindings`)

Stage-4 is entered ONLY when `stage3.all_bindings_missed` is True (present.py:278). But `all_bindings_missed = (bindings_resolved==0 AND _missed>0)`, and `_missed` increments only on `PATH_NOT_FOUND`. Every stage-3 blueprint `default_viewer_blueprint(data)` produces binds either `$bind: ""` (the whole doc — `resolve_pointer` returns `(doc, True)` unconditionally, binding.py:86-87) or, for the binary marker, a 0-binding literal. So stage-3 **never** all-misses.

Probed 21+ shapes directly — empty dict/list, scalars, `None`, binary marker, disjoint list-of-dicts, null values, adversarial JSON-pointer keys (`a~b`, `c/d`, `""`, `/`, `~`), and inline-blueprint-all-miss over all of them:

```
stage-3 all_missed across every shape:            NONE (all False)
unknown-view fallback_stage across every shape:   'content_type_default' (never 'generic')
inline-blueprint-all-miss fallback_stage:         'content_type_default' (never 'generic')
```

Root: unknown-view and inline-blueprint-all-miss both fall to stage-3, which always renders → `content_type_default`. The `_STAGE_GENERIC` return (present.py:287) is **defensive/unreachable** via the op path.

Writing the requested test would require monkeypatching `default_viewer_blueprint` to force a miss (a `patch` — forbidden by testing policy) or whitebox-constructing an all-missed `ResolvedPresentation` (Tier-4 setup-via-private) — neither exercises the real flow. My existing 2-value event coverage (`null` / `content_type_default`) therefore already pins **every value the event can actually carry**.

## Grounded options (your call — this changes WHAT is pinned)

- **Option A** — accept the 2-value event coverage as complete (it covers every reachable value) + add a one-line code comment at present.py:287 marking `_STAGE_GENERIC` defensive/unreachable-via-op. Fast, honest.
- **Option B (my recommendation)** — genuinely complete the coverage by pinning the *invariant that explains the missing value*: a Tier-2 test asserting stage-3 `default_viewer_blueprint` never all-misses across representative shapes ⇒ the `presented` event's `fallback_stage` is never `generic`. This guards against a future `default_viewer_blueprint` change silently making stage-4 reachable, and documents why only 2 values appear. Preserves your intent (complete + guarded enum coverage) honestly.
- **Option C** — if `generic` *should* be reachable (stage-3 ought to sometimes decline), that's a latent design question → separate issue, out of this additive-field PR's scope.

Holding the push for your decision. #2797 as-is is CLEAR per co-vet; this only concerns the requested fold-in.

tya5:
**[per-PR-coder]** — Option B pushed + the (C) root-cause characterization is done. `28570db9`.

## Option B — invariant test (enum coverage now honest & complete)

Added `test_stage3_default_viewer_never_all_misses_so_generic_never_emitted` (Tier 2, real op/OpContext/EventLog, no mocks): drives 9 data shapes (dict / list-of-dicts / list-of-scalars / scalar / empty dict / empty list / binary marker / null-valued dict) through BOTH fallback entry points (unknown view name AND all-missing inline blueprint — each falls to stage 3), asserting no emitted `presented` event ever carries `fallback_stage == "generic"` (only `null` / `content_type_default`). Plus a code comment at the stage-4 branch (`present.py`) documenting the unreachability.

Now all three enum values are accounted for at the event level: `null` and `content_type_default` pinned as produced values, `generic` pinned as the invariant that explains its absence — and the test goes RED if a future `default_viewer_blueprint` change makes stage 3 all-miss (re-activating stage 4).

**RED-verified** (cp-scratch, not a mock): temporarily forced `default_viewer_blueprint` to bind a non-existent path → stage 3 all-misses → the guard fired with `AssertionError: stage-4 generic unexpectedly emitted`. Restored clean.

## (C) root-cause: **C1 — intentional vestigial (defense-in-depth), NOT a gap → no ticket**

Read the FP-0054 proposal (`docs/deep-dives/proposals/0054-present-layer.md`) stage-cascade rationale. Evidence for C1:

- **§ "Degrade, don't fail" (lines 107-110):** the whole 4-stage cascade is a degradation ladder whose invariant is *"a broken template can never lose the user's access to the data … never a hard failure."* Stage 4 is the **terminal guarantee** of that invariant.
- **§3 Stage 3 (lines 197-212):** designed as the **shape-exhaustive universal viewer** — its precedence ladder (content-type → diff-sniff → structural shape) explicitly covers *every* JSON type: `list[dict]`→table, `list[scalar]`→list, `dict`→keyvalue, `str`/scalar→text, binary→image/text-placeholder. Nothing is left for stage 4 to uniquely handle.
- **§3 Stage 4 (lines 213-214):** framed as *"the final catch that **always renders** … so data always reaches the user"* — defense-in-depth language, not a specific-case last-resort.

Crucially, the proposal **never enumerates a data case stage 3 cannot render that stage 4 must** — so `$bind:""` is not *accidentally subsuming* a stage-4-intended case (which would be C2). Stage 3 is intended-universal; stage 4 is the always-renders safety net behind it. The current unreachability is **by construction, consistent with intent**.

→ **C1: no ticket.** The (B) code comment + invariant test are the right closure. It's a future dead-code-cleanup *candidate* (removing stage 4 would remove a construction-time guarantee), but explicitly not-a-bug today.

4 CI gates green (ruff / tier-audit 22/22 / module-docstring / target 22 pass). Ready to merge.

PR #2792, opened 2026-07-10
fix: neutralize LLM-derived intervention content in StdinInterventionBus prompt render (#2779)

tya5:
**[per-PR-coder]** — Closes #2779

Sibling follow-up to #2776 (merged): #2776 unified the intervention DISPLAY with the `present`-layer renderer discipline (FP-0054 terminal neutralizer), but was scoped to the **inline-CUI** surface (`reyn chat`) only. This PR extends the SAME neutralizer to the **remaining unguarded surface** #2779 identified: `StdinInterventionBus` — the `reyn run` / cron / scripted-CLI stdin bus, a genuine terminal surface with its own rendering path (a prompt_toolkit prompt string, not the Rich Console scrollback #2776 already covers).

## Render-seam evidence (the raw path, before this fix)

`StdinInterventionBus._render_prompt` (`src/reyn/user_intervention.py:298`, static method) built the prompt string by directly interpolating `iv.prompt` / `iv.detail` / `iv.suggestions` / choice `label`s — no ESC/control-strip guard anywhere on the path:

```python
lines.append(f"{prefix}{iv.prompt}")
if iv.detail:
    lines.append(f"  {iv.detail}")
...
labels = " / ".join(c.label for c in iv.choices)
```

`deliver()` (`user_intervention.py:276`) then hands this string straight to `prompt_toolkit`'s `_read_line` (`session.prompt_async(prompt_text)`), a genuine terminal-writing sink. `iv.prompt`/`iv.detail`/`iv.suggestions` are LLM-derived — `ask_user` args come straight from a model tool-call, and permission prompts interpolate a model-controlled `path` — so an ESC/control sequence embedded there could drive this terminal (OSC-52 clipboard, cursor tricks, etc.) with zero neutralization. Confirmed by cp-falsify below: removing the neutralize calls reproduces the raw (un-neutralized) render immediately.

## Fix

[...]

PR #2799, opened 2026-07-10
fix: collapse single closed-set MCP elicitation field to one prompt (#2622)

### LINKED ISSUE #2622: MCP elicitation: confirm() shows a redundant second prompt (two renderers for one request)
**[lead-coder]** — Found in tui-coder's live dogfood of #2597 slice ③ (elicitation).

A single `confirm(question)` elicitation call surfaces TWO prompts to the user: (1) the correct server-attributed question banner ("⚠️ MCP server 'X' asks (this is NOT reyn): <question>"), then (2) a redundant sub-prompt ("[MCP server 'X'] value: yes or no?"). Looks like two renderers fire for the same elicitation request. Correctness is unaffected (the round-trip value flows back correctly) — it's a cosmetic/UX redundancy.

Repro: connect an elicitation-capable MCP server (tests/_support/mcp_elicitation_server.py), call its confirm tool live via reyn chat, observe two prompts for one elicit.

Likely in the elicitation handler's per-field sequential-prompt path (src/reyn/mcp/elicitation.py) emitting both a banner and a per-field prompt for a single-field bool schema. roi:low (cosmetic).

tya5:
**[per-PR-coder]** — Closes #2622

An MCP `confirm()` (single-bool elicitation) surfaced **two** prompts for one request. Primary evidence (real FastMCP stdio server + recording bus, before the fix):

```
=== 2 prompt(s) rendered for one confirm() ===
[1] kind='mcp_elicitation'
    prompt="⚠️ MCP server 'srv' asks (this is NOT reyn): Delete the file?"
    choices=[('accept','[y]es','y'), ('decline','[N]o','N')]
[2] kind='mcp_elicitation_field'
    prompt="[MCP server 'srv'] value: yes or no?"
    choices=[('true','[y]es','y'), ('false','[n]o','n')]
```

The gate banner (prompt 1) already carries the server's real question (`Delete the file?`). The per-field prompt (prompt 2) only restates the ask against FastMCP's meaningless scalar-wrapper name (`value`) — a **redundant second render**. A single-enum elicitation has the identical redundancy (verified: `pick()` → gate + `value: choose a value`).

## Fix (`src/reyn/mcp/elicitation.py`)

For a single **closed-set** field (bool / enum), collapse gate + field into **one** prompt:
- The **attributed banner** (carrying the server's real question + the `NOT reyn` disclaimer) is the prompt.
- The field's own **value choices** collect the answer (bool → yes/no, enum → the values).
- An explicit **`[d]ecline`** choice folds in the gate's refuse-to-engage action, so `action=decline` stays user-reachable — **zero server-observable semantic loss** (accept+false and decline remain distinct).

Excluded from the collapse (each keeps its gate, with reason):
- **Single free-text field** — a free-text prompt has no choice list to fold a decline into, so its gate is the load-bearing decline affordance, not a redundant restatement.
- **Sensitive field** — its "your answer is SENT to the server" warning must never be bypassed.
- **Multi-field** — the gate previews the field list (`Will ask for: a, b, c`).

After the fix (real server round-trip):

```
=== confirm accept=true:   1 prompt, server='True'    ===
=== confirm answer=false:  1 prompt, server='False'   ===
=== confirm DECLINE:       1 prompt, server='decline' ===   (all three MCP actions preserved)
=== enum pick green:       1 prompt, server='green'   ===
```

Docstring D1 and the user doc (`docs/concepts/tools-integrations/mcp.md` + `.ja.md`) updated to describe the single-closed-set collapse.

## Tests (`tests/test_2597_slice3_mcp_elicitation.py`, Tier 2, real FastMCP stdio server + real `MCPConnectionService`)

- `test_single_bool_field_collapses_to_one_prompt` — a `confirm` renders **exactly one** prompt (unpack-to-one), carrying `NOT reyn` + the real question, no `mcp_elicitation_field` render; `true` → server `True`.
- `test_single_bool_field_no_and_decline_stay_distinct` — from the one prompt, `false` → `False` and `decline` → `decline` both stay reachable (semantic-preservation guard).
- `test_single_enum_field_collapses_to_one_prompt` — the enum sibling collapses identically (added a `pick` Literal-enum tool to the real support server).
- The 7 existing tests (free-text single-field, multi-field, headless, gate-false, timeout, sensitive-warning, audit-key-only) pass **unchanged** — those paths are untouched.
- **cp-falsify done**: removing the collapse branch (cp-scratch backup, never `git checkout`) → the 3 new tests go RED (confirm/enum revert to the 2-prompt flow → `decline`/wrong value); restored → all green.

## CI gates (all local, green)
- `python scripts/test_tier_audit.py --strict tests/test_2597_slice3_mcp_elicitation.py` — 10/10 OK, 0 Tier-4 (the "exactly one prompt" assertion uses the unpack-to-one idiom, not a `len(...) == N` format pin)
- `pytest` — 8032 passed, 21 skipped; the 5 failures are the **pre-existing** web/a2a/mcp route-mount + plugin-loader env failures (identical set to #2792's run on clean HEAD), none reference elicitation
- `python scripts/verify_module_docstrings.py src/reyn/mcp/elicitation.py` — OK

> Local note: the venv's editable `reyn` resolves to the primary worktree's `src`, so all local runs used `PYTHONPATH=<this-worktree>/src` (verified `import reyn.mcp.elicitation` resolves into this worktree). CI installs the branch checkout, so it's unaffected.

## Test plan
- [x] `confirm` → exactly one prompt, `true`→True — automated
- [x] `confirm` no vs decline stay distinct (False vs decline) — automated
- [x] single-enum `pick` → one prompt, value round-trips — automated
- [x] free-text / sensitive / multi-field keep the gate (unchanged tests pass) — automated
- [x] cp-falsify: drop collapse → RED; restore → green — done manually, verified


tya5:
**[dogfood-coder]** — co-vet on #2799, verify-don't-infer (security-adjacent). **Verdict: CLEAR to merge.**

## 1. Security semantics preserved — VERIFIED directly in code

The collapsed prompt's `prompt=attributed` reuses the **same `attributed = _attributed_message(server_name, message)` variable** the pre-existing gate uses — not a re-derived/reimplemented string, so the `"⚠️ MCP server {name!r} asks (this is NOT reyn): ..."` wording can't drift independently of the gate. Decline is an explicit choice (`InterventionChoice(id=DECLINE, label="[d]ecline", hotkey="d")`) folded into the same choice list as the value options, and `if answer.choice_id is None or answer.choice_id == DECLINE: raise _Declined()`. Traced `_Declined()` to its handling: **all three raise sites (the new collapse block, the no-fields gate, and the multi-field gate) share one `except _Declined:` clause** at the bottom of `handler()` — identical `mcp_elicitation_answered(action="decline")` emit + `ElicitResult(action="decline")` return for every path. The refuse-to-engage affordance isn't reimplemented for the collapsed case, it's the same code.

## 2. Zero server-observable semantic loss — VERIFIED + falsified myself

Read the 3-way mapping directly: `true` → `to_value("true") == True` → `content={name: True}`; `false` → `content={name: False}`; `DECLINE` → `ElicitResult(action="decline")` via the shared handler. Three structurally distinct choice ids produce three distinct results — no collapsing of "false" and "decline" into one signal.

**Falsified myself** (not just trusting the PR's claim): `git apply -R` on the src diff, reran the test file — exactly the 3 new tests go RED, with a concrete cross-contamination failure mode visible in the enum test (`assert 'decline' == 'green'` — pre-fix, the scripted "green" answer gets consumed by the OLD gate's accept-check instead of the field prompt, then the real field prompt gets no scripted answer and auto-declines) — a real semantic corruption the fix eliminates, not a cosmetic prompt-count difference. The 7 existing tests stayed green throughout (untouched paths genuinely untouched). Restored the fix, reconfirmed 10/10 green.

## 3. Scope — VERIFIED via control-flow reading, not the docstring's claim

The collapse `if` block requires `len(field_specs) == 1 and _is_closed_set(field_specs[0]) and not _is_sensitive_field(...)`. `_is_closed_set` returns `bool(spec["enum"]) or spec["type"] == "boolean"` — a free-text field (empty enum, non-boolean type) evaluates False and falls through unchanged to the pre-existing `field_summary`/gate code below (verified this is literally the same, unedited code — only 3 lines above the collapse block were touched: the import line and the new helper functions). Multi-field (`len(field_specs) > 1`) and sensitive single-field both fall through the same way. D1 docstring accurately describes all three exclusions with reasons.

## 4. Tests — VERIFIED Tier 2, real server, correct idiom, ran + falsified myself

`tests/_support/mcp_elicitation_server.py`'s `pick()` tool is a real FastMCP tool (`ctx.elicit(question, response_type=Literal[...])`), and the tests drive it through a real `MCPConnectionService` + `client.call_tool(...)` round-trip — not a mock. `(only,) = bus.seen` is the correct unpack-to-one idiom (behavioral "exactly one," not a `len(...) == N` format pin). Ran `test_tier_audit.py --strict`: 10/10 clean, 0 Tier-4 violations. Ran the test file myself: 10/10 pass. Ran the **full** suite myself: 8031 passed, 9 failed (identical file/test names to the established pre-existing baseline from every prior full-suite run this session), 18 skipped — no new regressions anywhere in the codebase from this change.

## 5. Doc mirror — VERIFIED, both languages in sync and accurate

`docs/concepts/tools-integrations/mcp.md` and `.ja.md` both gained an equivalent new bullet describing the single-closed-set-field collapse, correctly stating the free-text/multi-field exceptions keep their gate — content matches the actual code behavior I verified above, not just a generic restatement.

- Read the full collapse block + shared exception handler directly (not the PR's characterization of it).
- Ran the targeted test file (10/10) and the tier-audit (10/10, 0 violations) myself.
- **Falsified**: reverted the src fix, confirmed exactly 3 tests fail with a genuine semantic-corruption failure mode (not a trivial prompt-count assertion), restored, reconfirmed green.
- Ran the full pytest suite myself: exact match to the established 9-failure pre-existing baseline, no new failures.

No fix-before-merge items.

Issue #2357, opened 2026-07-01
fix: dogfood action_usage wipe targets stale path (.reyn/state/action_usage.jsonl) — silent no-op, hot-list counters bleed across scenarios


tya5:
**[lead-coder]** — Found during the multi-session same-class sweep (#2348). The dogfood harness's per-scenario ledger wipe targets the WRONG path, so it's a silent no-op.

## Bug
The live `ActionUsageTracker` persists to `.reyn/agents/<name>/action_usage.json` (`session.py:1173-1177`). But:
- The dogfood harness wipe (`dogfood.py:528-530`) `unlink`s `.reyn/state/action_usage.jsonl` — a stale path that doesn't exist → silent no-op. The comment says it wipes the ledger "so hot-list frequency counters don't bleed across scenarios", but it does NOT actually wipe anything → hot-list frequency DOES bleed across dogfood scenarios (measurement contamination).
- The `/clear-history` docstring (`clear_history.py:4-7`) also references the stale `.reyn/state/action_usage.jsonl` — but `/clear-history` still works because it clears the live `session._action_usage_tracker` object, not the docstring's path (docstring drift only).

## Fix
Point the dogfood wipe at the real path `.reyn/agents/<name>/action_usage.json` (or clear the tracker object). Fix the `/clear-history` docstring drift. roi:low — dogfood measurement hygiene (not production behavior), but it silently contaminates hot-list metrics across scenarios.

Note: this is separate from the #2348 multi-session fix. action_usage itself is being kept agent-wide (evidence: it's a cold-start tool-habit hint; the per-turn live overlay supplies conversation-local freshness).

PR #2801, opened 2026-07-10
fix: dogfood 'fresh' reset wipes spawn-arc-created agents in single-run mode (#2169)

### LINKED ISSUE #2169: dogfood: 'fresh' state-mode doesn't reset .reyn/agents/ — agent-creating scenarios (spawn-arc) leak across single-runs
**[dogfood-coder]** — Dogfood-harness gap surfaced by tui's live S3 (#2103 spawn-arc) run. **Classified as a harness/setup artifact, NOT an OS isolation defect** (evidence below).

## Symptom
A single-run dogfood of `spawn_single_agent` (creates agent `researcher`) hit `agent already exists` → `reply_outcome=refuted` on re-run, until the workspace was manually reset. The shared project `.reyn/agents/` was polluted with agents from prior runs/sessions (`alice/A/B/C/child/researcher/…`).

## Root cause (primary evidence)
The runner's `"fresh"` state-mode (`runner.py:45-49`) resets `action_usage.jsonl / wal.jsonl / history.jsonl / reyn/local/` but **NOT `.reyn/agents/`**. The spawn-arc (#2103 B/C) introduces the FIRST dogfood scenarios that *create agents* — so they leak created agents into the shared project workspace across single-runs. Pre-spawn-arc scenarios never created agents, so the gap was latent.

## Why NOT an OS defect (classification evidence)
- **OS architecture (static):** agents are isolated by distinct dirs (`.reyn/agents/<name>/`), spawned sessions by `_session_state_dir(name, sid)` (distinct dirs). Distinct agents/sessions → distinct workspace dirs by construction (P5). No architectural path for distinct-agent *state bleed* → an OS isolation defect is ruled out.
- **Runtime (tui):** the collision was leftover same-name agents from prior runs (workspace REUSE), not a concurrent distinct-agent bleed. Re-running with a fresh `.reyn` (temp cwd) resolves it.
- **Batch mode is unaffected:** the batch harness runs each worker in its own worktree (`/tmp/reyn-worktrees/...`), so `.reyn/agents/` is fresh there. The gap is specific to the **single-run / manual `--agent default`** mode.

## Fix options (harness lane, low priority)
1. Extend `"fresh"` state-mode to also reset/isolate `.reyn/agents/` (and spawned-session dirs) so agent-creating scenarios are reproducible in single-run mode; **or**
2. Document the temp-cwd / fresh-`.reyn` requirement for agent-creating scenarios in single-run mode (the interim workaround tui is using).

## Workaround in place
The S3 scenario design (#2164) documents a fresh-per-run-workspace **precondition**; tui re-runs via a temp cwd. So S3 is unblocked regardless.

tya5:
**[per-PR-coder]** — Closes #2169

Spawn-arc scenarios (#2103 B/C) are the first dogfood scenarios that **create** agents — each writes `.reyn/agents/<name>/profile.yaml` (`researcher`/`alice`/`A`/`B`/`C`/`child`/…). In **single-run** mode the project `.reyn/` is REUSED across runs, so those created agents leaked: a re-run of an agent-creating scenario hit `agent already exists` (`AgentRegistry.create` raises when `profile.yaml` is present) → `reply_outcome=refuted`, until a manual workspace reset.

## Root cause (primary evidence)

The single-run per-scenario reset is `_wipe_scenario_state` in `_build_live_runner` (`src/reyn/interfaces/cli/commands/dogfood.py`), invoked before every scenario by `runner_fn`. It wiped only the **target** agent's ephemeral files (`_scenario_state_targets`: `action_usage.json` + `history.jsonl` + the events chat dir) — **not** the sibling agent dirs the spawn-arc creates. So created agents accumulated under the shared `.reyn/agents/`.

Traced the existence check to the real mechanism: `AgentRegistry.exists(name)` → `(.reyn/agents/<name>/profile.yaml).is_file()`; `create()` raises `FileExistsError` when it's present (`registry.py:510-516`).

## Fix

`_wipe_scenario_state` now also removes every **created sibling** agent dir under `.reyn/agents/`, via two small pure helpers (`_leaked_agent_dirs` / `_wipe_leaked_agents`) mirroring the #2357 testable-helper pattern.

The dogfood **target** agent (`--agent`, default `default`) is deliberately **spared**, because:
- Its `profile.yaml` is a **load precondition** — `AgentRegistry.get_or_load` raises `FileNotFoundError` without it, so wiping it would block **every** scenario (`ensure_running` → blocked).
- Its ephemeral ledger/history are already wiped by `_scenario_state_targets`.

Safe by construction: per-agent **event logs** live under `.reyn/events/agents/<name>/` (PR20), **not** under the removed `.reyn/agents/<name>/` dir — so this removes **no event log** (no running-web constraint). This is the runner's job because it knows the target agent name; `dogfood_fresh_reset.sh` deliberately can't (it has no agent name — the same reason it already defers the `history` / `action_usage` wipe to the runner).

**Batch mode is unaffected** — each worker runs in its own fresh-`.reyn` worktree (`/tmp/reyn-worktrees/...`), so nothing leaks there. The gap was specific to single-run mode.

## Tests (`tests/test_dogfood_leaked_agents_wipe_2169.py`, Tier 2, real `AgentProfile`/`PROFILE_FILENAME`)

- `test_leaked_agent_dirs_selects_siblings_spares_target` — the reset selects every created sibling and never the target (order-agnostic set membership, no sort-order pin).
- `test_wipe_removes_created_agents_preserves_target` — after the real `_wipe_leaked_agents`, the siblings' `profile.yaml` is gone (so a re-run no longer hits "already exists") while the target's `profile.yaml` survives and still `AgentProfile.load`s (so `get_or_load` won't `FileNotFoundError`). Uses real `AgentProfile.save` — the exact marker `exists()` keys off, no mocks.
- `test_wipe_is_idempotent_with_no_agents_dir` — no-op / never raises when nothing has leaked.
- **cp-falsify done**: neutering `_wipe_leaked_agents` (cp-scratch backup, never `git checkout`) → the removal test goes RED (`researcher/profile.yaml` still present); restored → all green.

## CI gates (all local, green)
- `python scripts/test_tier_audit.py --strict tests/test_dogfood_leaked_agents_wipe_2169.py` — 3/3 OK, 0 Tier-4
- `pytest` — 8035 passed, 21 skipped; the 5 failures are the **pre-existing** web/a2a/mcp route-mount + plugin-loader env failures (identical set to #2792/#2799's runs on clean HEAD), none reference dogfood
- `python scripts/verify_module_docstrings.py src/reyn/interfaces/cli/commands/dogfood.py` — OK

> Local note: the venv's editable `reyn` resolves to the primary worktree's `src`, so all local runs used `PYTHONPATH=<this-worktree>/src` (verified the import resolves into this worktree). CI installs the branch checkout, so it's unaffected.

## Test plan
- [x] created sibling agents removed by the fresh reset — automated
- [x] target agent's profile spared (still loads) — automated
- [x] idempotent when nothing leaked / no `.reyn/agents/` dir — automated
- [x] cp-falsify: neuter wipe → RED; restore → green — done manually, verified

### DIFF
--- a/src/reyn/interfaces/cli/commands/dogfood.py
+++ b/src/reyn/interfaces/cli/commands/dogfood.py
@@ -269,6 +269,39 @@ def _scenario_state_targets(project_root: Path, agent_name: str) -> "dict[str, P
     }
 
 
+def _leaked_agent_dirs(project_root: Path, keep_agent: str) -> "list[Path]":
+    """#2169: the spawn-arc-created SIBLING agent dirs under ``.reyn/agents/`` that a
+    fresh per-scenario reset must remove — every agent dir EXCEPT the dogfood target.
+
+    Agent-creating scenarios (spawn-arc #2103 B/C) write ``.reyn/agents/<name>/profile.yaml``
+    for each agent they spawn (``researcher``/``alice``/``A``/``B``/``C``/``child``/…).
+    In single-run mode the project ``.reyn/`` is REUSED across runs, so those created
+    agents leak: a re-run of an agent-creating scenario hits ``agent already exists``
+    (``AgentRegistry.create`` raises when ``profile.yaml`` is present) → ``refuted``.
+
+    The dogfood TARGET agent (``keep_agent``, default ``"default"``) is deliberately
+    spared: its ``profile.yaml`` is a precondition — ``AgentRegistry.get_or_load`` raises
+    ``FileNotFoundError`` without it, so wiping it would block EVERY scenario — and its
+    ephemeral ledger/history are already wiped by ``_scenario_state_targets``. Per-agent
+    event logs live under ``.reyn/events/agents/<name>/`` (PR20), NOT under the removed
+    ``.reyn/agents/<name>/`` dir, so this removes no event log (no running-web constraint,
+    the reason ``dogfood_fresh_reset.sh`` — which has no agent name — defers this)."""
+    agents_root = project_root / ".reyn" / "agents"
+    if not agents_root.is_dir():
+        return []
+    return [d for d in sorted(agents_root.iterdir()) if d.is_dir() and d.name != keep_agent]
+
+
+def _wipe_leaked_agents(project_root: Path, keep_agent: str) -> None:
+    """Remove the spawn-arc-created sibling agent dirs (#2169) so agent-creating
+    scenarios are reproducible in single-run mode. Spares ``keep_agent`` (see
+    :func:`_leaked_agent_dirs`). Idempotent: a no-op when none have leaked."""
+    import shutil
+
+    for d in _leaked_agent_dirs(project_root, keep_agent):
+        shutil.rmtree(d, ignore_errors=True)
+
+
 def _baselines_dir() -> Path:
     return _dogfood_base_dir() / "baselines"
 
@@ -401,6 +434,11 @@ def _build_live_runner(agent_name: str, *, env_backend=None, ws_base_dir=None, w
       context — defeating the "fresh per scenario" guarantee. The
       dogfood_fresh_reset.sh script explicitly defers per-agent history
       wipe to callers (= the runner is the caller); this is that wipe.
+    - Wipes agents CREATED by spawn-arc scenarios (sibling .reyn/agents/<other>/
+      dirs) before each scenario so an agent-creating scenario is reproducible
+      in single-run mode — otherwise created agents leak across runs and a
+      re-run hits "agent already exists" (#2169). The dogfood target agent is
+      spared (its profile.yaml is a load precondition).
     - Drops the cached Session from the registry between scenarios so
       the session's in-memory EventLog starts empty each time.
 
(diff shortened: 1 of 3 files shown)


tya5:
**[per-PR-coder]** — folded the co-vet should-fix (wiring regression-guard).

The original 3 tests (`test_dogfood_leaked_agents_wipe_2169.py`) call `_wipe_leaked_agents` **directly** — mechanism tests that would stay green even if the production call `_wipe_scenario_state` → `_wipe_leaked_agents` (dogfood.py:607) were stripped. That left the wiring un-guarded → a future regression removing the call would silently re-open #2169.

**Added** `test_fp0036_live_runner.py::test_created_sibling_agents_wiped_before_scenario` (Test 6, same shape as Test 5's history-wipe guard, reusing the existing `_make_live_runner_fn` harness): seed a created sibling agent (`researcher`) with a real `profile.yaml`, drive the **real `runner_fn`** end-to-end (fake LLM), assert the sibling's `profile.yaml` is gone and the target `default` survives (self-check that the runner rooted on tmp_path).

**cp-falsify (manual, verified):** deleting the `_wipe_leaked_agents(project_root, agent_name)` call at dogfood.py:607 →
- `test_created_sibling_agents_wiped_before_scenario` → **RED** (sibling profile survives)
- the 3 direct-call mechanism tests → **stay GREEN**

— confirming the new test closes the wiring gap. Restored via cp-scratch (never `git checkout`).

**CI gates (local, green):** `ruff check src tests` clean · `test_tier_audit --strict tests/test_fp0036_live_runner.py` 6/6 OK, 0 Tier-4 · `pytest` 8055 passed, 21 skipped (the 5 failures are the pre-existing web/a2a/mcp route-mount + plugin-loader env failures, unrelated) · `verify_module_docstrings` OK (no src change — dogfood.py unchanged from the reviewed commit).

Test plan updated:
- [x] created sibling agents removed by the fresh reset (mechanism) — automated
- [x] target agent's profile spared (still loads) — automated
- [x] idempotent when nothing leaked — automated
- [x] **production wiring guarded: runner_fn → _wipe_scenario_state removes the sibling; strip the call → RED** — automated + cp-falsify verified

Ready for re-review / merge.

PR #2806, opened 2026-07-10
feat: P0 server-side tiered authentication for thin-client model (ADR-0039)

tya5:
**[per-PR-coder]** — P0 of the thin-client / single-writer-server arc (ADR-0039). Closes the security hole where the reyn web/ws gateway had **zero client authentication** and the ws `answer_intervention` path processed answers as the unfenced operator — so with `--host 0.0.0.0` an unauthenticated network client could grant permissions as the operator. **Intervention answers are permission grants.** This gates every answer/permission path behind authentication, with a tiered secure-by-default transport model. Do not merge — the architect co-vets against the authoritative P0_auth_spec; lead merges on verdict.

`part of` the server-model arc (no single issue closes it).

## The tiered auth model (secure-by-default)
- **T1 in-process** — no auth (operator's own process).
- **T2 same-machine cross-process** — **UDS, DEFAULT** (`reyn web --uds PATH`), identity from **per-OS peer credentials** (Linux `SO_PEERCRED` / macOS+BSD `getpeereid`; Windows → loopback+token fallback). Loopback TCP is the opt-in same-machine fallback.
- **T3 cross-machine network** — **token handshake + TLS, opt-in, fail-closed**. Self-signed TLS is generated with a printed SHA-256 fingerprint (TOFU); operator cert overrides.

**Two surfaces, one auth core:** the existing browser/openui ws surface uses **loopback TCP + a startup-issued token** (Jupyter-style — the token is embedded in the printed launch URL and forwarded on ws connect); the thin-client/UDS surface uses **per-OS peer-cred**. The retrofit applies to both.

**Keystone:** an authenticated connection's identity determines fencing. The v1 single operator identity is **unfenced** (`external_source=False`). **A2A peer answers stay FENCED** (`external_source=True`) — untouched.

## The retrofit (load-bearing)
`ws_chat` now authenticates **before accept** and rejects an unauthenticated connection; the answer/slash paths **re-verify authorization server-side, at delivery time, against the server's own record of the connection identity** (client-untrusted — the seam a later seize/takeover phase extends to reject a deposed holder's in-flight answer). Answer/attach/detach emit audit-events stamping the identity (+ connection id, tier, peer-uid).

⚠️ **Pre-existing bug fixed (security-relevant):** on `origin/main` the `@router.websocket("/ws/chat/{agent_name}")` decorator sat on the `_decode_inbound_frame` helper, not on `ws_chat` — so the real chat handler (and therefore the new auth gate) was **never routed**. Fixed as part of this PR; without it the retrofit would be dead code.

## Fail-closed startup
`check_startup_binding` refuses to start a **non-loopback bind without a configured `web.auth.token`** (closes the accidental-exposure hole). UDS and loopback binds are allowed; loopback generates an ephemeral token so the browser surface is never left unauthenticated.

## The 5 tested invariants (real instances, no mocks)
1. **Unauthenticated connection cannot answer/grant** — `test_web_auth_ws_gate.py` (live TestClient: existing agent, no/wrong token → rejected at handshake; valid token → accepted).
2. **Non-loopback bind × no token → refuses to start** — `test_web_auth_startup_failclose.py`.
3. **Per-OS peer-cred → user-ID + uniform T2/T3 authz** — `test_web_auth_peercred.py` (real `AF_UNIX` socketpair) + `test_web_auth_identity_authz.py` (UDS/loopback/network resolve to one operator user-id).
4. **A2A peer stays fenced** — `test_web_auth_a2a_fenced_regression.py` (keystone: operator answer unfenced, A2A answer fenced).
5. **Audit stamps identity** — `test_web_auth_ws_gate.py::test_emit_audit_stamps_identity`.

Plus `test_web_auth_config_contract.py` (Tier 1 `web.auth` schema) and `test_web_auth_lifespan_wiring.py` (AuthContext built at startup).

**Falsify evidence:** stripping the ws auth gate → invariant-1 test RED; neutering the fail-close raise → invariant-2 test RED (both restored via cp-backup, re-run GREEN).

## CI gates (all green locally on the diff)
- `ruff check src tests` ✅
- `python scripts/test_tier_audit.py --strict <new test files>` ✅ (22 tests, 0 Tier-4)
- `pytest` ✅ — 8090 passed; the only failures are **5 route-mounted tests (a2a/mcp/resources/fp0001/plugin_loader) that fail identically on origin/main** (env artifact: web routers not mounting in this dev env) — not introduced here.
- `python scripts/verify_module_docstrings.py <changed src>` ✅

## Out of scope (later phases, not built)
Axis-B seize/takeover (P-later), `reyn chat --connect` client (P1+), AG-UI transport (P2), full last-surface-gone DENY + grace window (P3 — disconnect does **not** invalidate pending answers here), full concurrent T2+T3 dual-bind and live UDS-fd peer-uid extraction in the ws path (the peer-cred function is real + unit-tested; live UDS wiring lands with the P1 connect route).

## Test plan
- [x] Auth core unit tests (identity/token/authz/peer-cred/TLS/fail-close) pass.
- [x] Live ws unauth→rejected / token→accepted (TestClient) pass.
- [x] Fail-closed startup: `reyn web --host 0.0.0.0` (no token) exits 2 with an actionable message; loopback prints a launch URL with token; `--uds` binds the socket (CLI `_apply_auth_startup` exercised).
- [x] A2A-fenced regression + audit-stamp pass.
- [x] Falsify: auth gate + fail-close guard verified load-bearing (RED on strip, GREEN on restore).
- [x] Four CI gates run locally on the diff.

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -527,6 +527,11 @@ web:
     max_download_bytes: 10485760        # wire-byte ceiling (default 10MB)
     allow_private_ips: false            # SSRF: opt-in to private IPs (default deny)
   ws_max_size: 16777216                 # WS inbound-frame ceiling (default 16MB)
+  auth:
+    token: my-shared-secret             # T3 cross-machine bearer token (required for a non-loopback bind)
+    require_token_on_loopback: true     # also require the token on loopback TCP (secure default)
+    tls_certfile: /path/to/cert.pem     # operator TLS cert (T3); omit → self-signed TOFU
+    tls_keyfile: /path/to/key.pem       # operator TLS key (T3); set together with tls_certfile
 ```
 
 Priority chain (highest first):
@@ -545,6 +550,12 @@ Priority chain (highest first):
 | `web.fetch.max_download_bytes` | int | `10485760` (10MB) | Maximum response bytes `web_fetch` reads off the wire. A response whose `Content-Length` exceeds this is rejected before any body is downloaded; a chunked / unknown-length body is aborted once the stream passes the ceiling (status `too_large`). Guards against an unbounded-body memory blow-up from a hostile or runaway URL. `<= 0` or non-integer falls back to the default. |
 | `web.fetch.allow_private_ips` | bool | `false` | SSRF opt-in. When `true`, `web_fetch` / `safe.http` may fetch **private** RFC1918/ULA addresses (enterprise internal-fetch). Link-local, cloud-metadata (`169.254.169.254`), and loopback are **always** denied regardless of this flag. HTTP redirects are re-validated per hop (both the host allowlist and the IP-deny), so an allowlisted host cannot redirect to an internal target. Also exported to the `REYN_FETCH_ALLOW_PRIVATE_IPS` env var so the safe.http subprocess and registry clients honor the same opt-in. |
 | `web.ws_max_size` | int | `16777216` (16MB) | Maximum size (bytes) of a single inbound WebSocket frame the `reyn web` gateway accepts; a larger frame is rejected by the server before delivery. Pins the WebSocket frame ceiling explicitly instead of relying on the server library's implicit default, so the bound stays in place across server-library upgrades. Operators may tighten or loosen it. `<= 0` or non-integer falls back to the default. |
+| `web.auth.token` | str | `null` | The gateway's cross-machine (T3) bearer token. A **non-loopback bind refuses to start** without it (fail-closed — closes the accidental-exposure hole). A loopback bind generates an ephemeral token at startup when this is unset (printed in the launch URL, Jupyter-style), so the browser surface is never left unauthenticated. |
+| `web.auth.require_token_on_loopback` | bool | `true` | When `true`, even loopback TCP connections must present the token (secure default — a shared multi-user host must not leave the browser loopback surface open). Same-machine UDS connections are authenticated by OS peer credentials and never need a token. |
+| `web.auth.tls_certfile` | str | `null` | Operator TLS certificate (PEM) for a T3 network bind. When unset, a self-signed certificate is generated at startup and its SHA-256 fingerprint is printed for trust-on-first-use pinning. Must be set together with `tls_keyfile`. |
+| `web.auth.tls_keyfile` | str | `null` | Operator TLS private key (PEM) paired with `tls_certfile`. Setting only one of the two is a startup error. |
+
+**Transport tiers** (secure-by-default). The gateway identifies every connection: **T1** in-process (the operator's own process, no auth); **T2** same-machine cross-process over a UNIX domain socket (`reyn web --uds PATH`) identified by OS peer credentials, or loopback TCP as a fallback; **T3** cross-machine network, which requires `web.auth.token` and runs over TLS. An intervention answer is a permission grant, so an unauthenticated connection cannot answer.
 
 ## `hooks` block
 
--- a/reyn.local.yaml.example
+++ b/reyn.local.yaml.example
@@ -663,6 +663,11 @@
 #     max_download_bytes: 10485760  # 10MB wire-byte ceiling (DoS guard)
 #     allow_private_ips: false      # #1956 SSRF: opt-in to private IPs (link-local/metadata/loopback always denied)
 #   ws_max_size: 16777216           # 16MB WebSocket inbound-frame ceiling (reyn web gateway)
+#   auth:                           # reyn web gateway authentication (server-side)
+#     token: null                   # cross-machine (T3) bearer token; REQUIRED for a non-loopback bind (else fail-closed)
+#     require_token_on_loopback: true  # also require the token on loopback TCP (secure default for shared hosts)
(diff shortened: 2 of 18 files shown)


tya5:
**[architect]** — co-vet verdict: **CLEAR to merge.** Security-critical P0; all 5 invariants strip-verified or live-verified against the authoritative spec. One non-blocking doc-nit.

## Mis-routing bug — genuine, correctly fixed, load-bearing (independently confirmed)
On `origin/main`, `@router.websocket("/ws/chat/{agent_name}")` decorates `_decode_inbound_frame` (a `raw: str` helper), and `ws_chat` (the real handler, line 111) is **undecorated and registered nowhere else** — I confirmed against `origin/main` directly. So the chat route pointed at the wrong function; browser chat + a future `--connect` client were broken, and **an auth gate on `ws_chat` would have been dead code**. The PR moves the decorator to `ws_chat` (now line 152). Fix is genuine, correct, and load-bearing for P0 (the gate is only reachable because the handler is now routed). Correctly taken in-scope.

## The 5 invariants
- **(a) unauthenticated cannot answer/grant** — the `ws_chat` gate rejects (close 4401) before attach, and a second `authorize_write(identity)` guards the delivery path. **Strip-verified: I neutered `if not identity.authenticated:` → `test_web_auth_ws_gate` (unauth + wrong-token) went RED.** ✓
- **(b) non-loopback × no-token ⇒ refuse to start** — `check_startup_binding` raises `AuthStartupError` (UDS/loopback exempt). **Strip-verified: I neutered the `raise` → `test_web_auth_startup_failclose[192.168.1.10/::/10.0.0.1]` went RED.** ✓
- **(c) per-OS peer-cred → user-id; T2/T3 uniform authz** — clean abstraction (Linux `SO_PEERCRED` / macOS `getpeereid` via `ctypes` / else `None` → token tier). **Live-verified on this Darwin box: `peer_uid_from_socket` over a real `AF_UNIX socketpair` returned `uid=501 == os.getuid()`.** `authenticate` resolves every tier to the same `OPERATOR_USER_ID` ⇒ uniform authz across a concurrent T2+T3 bind. ✓
- **(d) A2A stays fenced** — keystone tested **both directions, non-vacuously**: `test_authenticated_operator_answer_is_unfenced` (`external_source=False`) + `test_a2a_peer_answer_stays_fenced` (`external_source=True`). P0's operator-unfenced wiring did not weaken A2A fencing. ✓
- **(e) audit stamps identity** — `identity.audit_fields()["auth_user_id"] == OPERATOR_USER_ID`; `test_emit_audit_stamps_identity` forwards to a real `EventLog` and asserts the emitted event carries the identity. ✓

## Production reachability (the wiring, not just the mechanism)
`_lifespan` installs `app.state.auth = AuthContext.from_env_and_config(...)` at startup (even on config-load failure, so the surface is never unauthenticated); `ws_chat` reads it via `_get_auth_context`, and a **missing** context fails closed (close 4401). So the gate is genuinely reachable in production, not dead. `test_web_auth_lifespan_wiring` covers it.

## Spec conformance + a correctness improvement over my spec
- **2-surface split** correct: UDS (peer-cred) default; loopback/browser (token, generated if absent, printed in the launch URL); T3 (token+TLS, self-signed+TOFU, key `chmod 0600`). One server, uniform authz.
- **Server-side, client-untrusted** delivery-time check present, and the connection identity is recorded as the substrate for the later seize-race phase.
- **UDS file-mode:** implemented as **`run_dir.chmod(0o700)`** (owner-only parent dir), NOT a `0600` on the socket file — this is **more correct than my spec**: macOS ignores socket-file permission bits for `connect()` and relies on the containing directory, so a `0600` socket file would be a no-op on this very platform. Good call.
- **Out-of-scope correctly deferred** — seize/client/AG-UI/P3 grace-window/dual-bind/live-UDS-fd are absent; the only seize references are comments describing the foundation. No half-built surface.
- ruff clean; 35 auth tests pass.

- **Doc-nit:** several docstrings (`core.py`, `peercred.py`) still say "socket `0600` file mode" while the impl gates via the `0700` run-dir. Reword to "owner-only (`0700`) run-dir" so the security description matches the (more portable) implementation — no behavior change.
- **Residual to confirm (not this PR's regression):** the 5 route-mount CI failures — worth a one-line confirmation they are byte-identical to main's baseline (the mis-routing fix *changes* routing, so a quick diff-of-failures vs main closes the loop). I verified the auth subset + ruff locally.

**CLEAR — merge after your verdict-gate.** Strongest-tested of the arc so far; the mis-routing catch was a real reachability save.

tya5:
**[per-PR-coder]** — Folded the co-vet doc note (a1e531f0): reworded the UDS access-gate docstrings (peercred.py, core.py) + the CLI chmod comment from socket `0600` to the owner-only run-dir `0700` that the implementation actually enforces, with a one-line note on why (macOS ignores a UNIX socket's own file-mode on `connect`, so the parent directory is the enforceable gate). Pure doc precision, no behavior change; `tls.py`'s key-file `0600` chmod is a real file mode and left as-is. ruff + module-docstring gates green on the changed files; import smoke OK. Ready for merge.

Issue #2681, opened 2026-07-09
FP-0056 burn-down: replace CANONICAL_TODO provisional declarations with real canonical mappers (~40 producers)


tya5:
**[lead-coder]** — Burn-down tracker created during FP-0056 PR-F1 (#2680) review, per the architect's (b)-marker ruling.

## Context
PR-F1's registry-derived gate requires **every** LLM-invocable producer to declare a canonical (mapper or explicit passthrough) — and it **flushed out ~40 producers that were silently taking the whole-dict fallback** (memory / cron / catalog / task / agent-spawn / hooks / mcp-discovery / universal-wrappers / …). That the gate surfaced them is the framework **working as designed** (it would have caught the original `file` gap).

Writing real mappers for all ~40 is out of PR-F1's scope + not behavior-preserving in one shot, so they are declared **`CANONICAL_TODO`** — a distinct marker (NOT `STRUCTURED_PASSTHROUGH`, which stays admin/install-only per owner decision #1). `CANONICAL_TODO` = "declared (gate-satisfying, not a silent gap) but a real mapper is not yet written; provisional whole-dict fallback." Greppable debt.

## Burn-down: triage each CANONICAL_TODO producer
Per the architect's rubric — grep `CANONICAL_TODO` and triage:
- **text-content producer** (returns a body that should read as text, e.g. a doc/memory read) → **write a real text mapper**. **These are latent file-class bugs** (whole-dict blob instead of clean text → LLM confusion, exactly the incident). Highest priority.
- **genuinely-structured** (the whole dict IS the right LLM view) → review, then **promote to `STRUCTURED_PASSTHROUGH`**.
- **status-only** (write/ack-shaped) → **status-text mapper**.

## Scope
- Additive burn-down; each producer is independently addressable. Not blocking (behavior is preserved today via the fallback; PR-F2 adds runtime visibility via the `canonical_fallback_used` event firing on CANONICAL_TODO).
- Related: FP-0056 (`0056-canonical-coverage-enforcement.md`), the original incident (PR-H #2677), the coverage-migration lesson (memory pin).

tya5:
**[lead-coder]** — 裁定 refinement 反映（architect self-review 由来）。CANONICAL_TODO は単なる「宣言=gate green」ではなく **ratchet（一方向 burn-down 専用 debt marker）** として実装する:

**Ratchet 契約（gate が機械強制）:**
[...]

Issue #2593, opened 2026-07-05
BUG: shell: DSL sugar is unimplemented — parses to ToolStep(name='shell') but no 'shell' tool is registered (fails 100%)


tya5:
**[lead-coder]** — Found by tui-coder's per-component dogfood, confirmed via grep + live+offline repro.

## Symptom
Any pipeline using `shell:` sugar fails 100%: `pipeline tool step 'shell' does not resolve to a registered tool`.

## Root cause (primary evidence)
`parser.py:_parse_shell_step` (line 298) parses `shell: {command, ...}` into `ToolStep(name="shell", args={"command": ...}, ...)`. `shell` is in `_LINEAR_STEP_KINDS` (parser.py:189). But grep confirms NO tool literally named `"shell"` is registered anywhere in the tool registry — so the executor's `tool_dispatch` (which resolves ToolStep.name against the registry) can't resolve it. The step kind is parsed + documented but has no backing tool.

## Doc-accuracy note
`docs/reference/runtime/pipeline-dsl.md` (and the #2591 grammar) document `shell:` as a working step kind. The grammar is accurate to the PARSER (which parses it) but the RUNTIME is broken. An interim doc caveat is needed until this lands (flagging docs-maintainer).

## Spec-intended behavior (Appendix B)
`shell = {command:TPL ...}` = "sugar for tool name=shell. command structure is static. PIPE DATA of the previous step is fed to STDIN as JSON; stdout → schema-checked output." So shell was DESIGNED as a tool that runs the command, feeding the prev step's pipe-data to STDIN and capturing stdout as the step output.

## Fix direction — REUSE the existing sandboxed exec, do NOT reinvent
There is already `src/reyn/tools/sandboxed_exec.py` (the `exec__sandboxed_exec` op_runtime handler, exec category, HIGH-risk capability-floored). A `shell` pipeline tool should route the command through THAT sandboxed exec (inheriting its capability floor — so surfacing shell adds NO new security surface; a pipeline runs ⊆ the invoker's capability, and exec is already floored). The shell-specific contract (pipe-data → STDIN as JSON, stdout → output) is a small adapter on top. Trace-first: confirm sandboxed_exec's interface (does it support STDIN feeding? command shape? capability path?) + the cleanest way to register/wire a `shell` tool → sandboxed_exec with the STDIN/stdout contract.

roi:high — a documented feature that fails 100% of the time.

PR #2807, opened 2026-07-10
fix: shell gets a real text canonical mapper — #2681 Bucket A (latent file-class bug)

tya5:
**[per-PR-coder]** — part of #2681

## Context

FP-0056's registry-derived scout enumeration (issue #2681) triaged all 51 `CANONICAL_TODO` producers and flagged **Bucket A: `shell`** as the single highest-priority fix — a TEXT-content producer PR-F1's original triage missed. `shell` (`src/reyn/tools/shell.py:_handle`, #2593 pipeline-DSL sugar over `sandboxed_exec`) returns bare subprocess STDOUT (`json.loads(stdout)` when it parses, else the raw text). Under the provisional `CANONICAL_TODO` whole-dict fallback, this readable command output was shown to the LLM as an opaque `structured` blob instead of clean text — the same file-class latent bug the `file`/`reyn_src` 2026-07-09 dogfood incident produced.

## Fix

Added `shell_to_canonical` (`src/reyn/core/offload/canonical.py`), mirroring the EXISTING `sandboxed_exec_to_canonical`'s "stdout IS the text" treatment:

- The shape this mapper receives (post `unwrap_dispatch_envelope`) is one of: the dispatch envelope `{"status": ..., "data": <value>}` (the common router/chat-path case — a plain-text command, or JSON stdout that decoded to a non-`dict`), or `<value>` directly when stdout decoded to a `dict` (already peeled one envelope layer — the pipeline `tool:` step path).
- Either way, `value` renders as the readable `text` body — verbatim when already a `str`, else `json.dumps`'d so a JSON-emitting command's output stays legible — with NO `structured` attachment (the whole-dict blob this replaces).
- `meta` is always empty for `shell`: unlike `sandboxed_exec`, `shell`'s own `_handle` (locked #2593 design) surfaces ONLY stdout — `stderr`/`returncode` never reach this canonical seam (they're dropped one layer up, by that same locked contract). This is a structural, documented consequence of #2593, not an oversight in this mapper.
- `shell`'s `ToolDefinition.canonical` now declares `shell_to_canonical` instead of `CANONICAL_TODO`.
- Removed `"shell"` from `_CANONICAL_TODO_GRANDFATHERED` (`tests/test_fp0056_canonical_coverage_gate.py`) — the ratchet set only shrinks; the gate's `live == ledger` equality assert still passes.
- `tests/test_fp0056_m3_fail_visible.py`'s registry-wide "no silent status-only catch-all" probe drives every registered mapper with a bare `{"status": ...}` dict. `shell_to_canonical` has NO inner discriminator (it renders the whole value verbatim, never selects/drops a sub-field the way `file`/`reyn_src` dispatch on `op`/body-key), so it structurally cannot re-introduce the M3 class — narrowly exempted by name (`_NO_INNER_DISCRIMINATOR_MAPPERS`) with a documented rationale, not blanket-skipped.
- Updated the two `test_2593_pipeline_shell_step.py` assertions that pinned the OLD `CANONICAL_TODO` fallback shape (`{"text": "", "structured": {...}}`) to the new real-mapper shape (`{"text": '<json-dumped dict>'}`) — this is the intended, expected behavior change the fix makes (not a regression).

## Falsify evidence

Manually reverted `shell.py`'s `canonical=` back to `CANONICAL_TODO` and reran the new/updated tests — 6 failures (the new `tests/test_2681_shell_canonical.py` mapper tests + the two updated `test_2593_pipeline_shell_step.py` assertions), each showing the whole-envelope-dict / whole-JSON-dict `structured` blob with empty `text` — confirming RED pre-fix. Restored the real mapper — GREEN.

## Test plan

- [x] `tests/test_2681_shell_canonical.py` (new) — Tier 1: `shell` declares a real mapper (not `CANONICAL_TODO`); plain-text stdout (envelope shape) renders as clean `text`, no `structured` attachment; JSON-dict stdout (pipeline-unwrapped shape) renders as readable json-dumped `text`, no `structured` attachment; empty output renders the explicit `(no output)` marker; `shell` is off the grandfathered ledger.
- [x] `tests/test_2593_pipeline_shell_step.py` — updated the two assertions pinning the old fallback shape to the new real-mapper shape; the other 7 tests (registry wiring, stdin threading, schema verify pass/fail, timeout, security floors, DSL round-trip) pass unchanged.
- [x] `tests/test_fp0056_canonical_coverage_gate.py` — the ratchet `live == ledger` equality assert passes with `shell` removed; the admin-6 `STRUCTURED_PASSTHROUGH` assert is unaffected.
- [x] `tests/test_fp0056_m3_fail_visible.py` — added a narrow, documented exemption so the registry-wide status-echo probe doesn't false-positive on `shell_to_canonical`'s no-discriminator design; all 10 tests in the module pass.
- [x] Falsify: reverted the mapper → RED (6 failures) → restored → GREEN.
- [x] `ruff check src tests` — clean (verified in an isolated `git worktree` checkout of this branch's commit).
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 30/30 OK, 0 Tier-4 violations.
- [x] `pytest` (full suite from repo root, isolated worktree) — 8062 passed, 21 skipped; the only 5 failures are pre-existing route-mounting/plugin-loader tests unrelated to this change (confirmed to fail identically on a clean `origin/main` checkout with no changes applied).
- [x] `python scripts/verify_module_docstrings.py <changed src files>` — OK.

## Note for reviewers

This PR edits the `_CANONICAL_TODO_GRANDFATHERED` frozenset in `tests/test_fp0056_canonical_coverage_gate.py` (removing `"shell"`). A Bucket C PR (status-only producers, #2681) independently edits this same frozenset, removing different entries — expect a trivial rebase on whichever lands second (no semantic conflict, just adjacent lines in the same set literal).

### DIFF
--- a/src/reyn/core/offload/canonical.py
+++ b/src/reyn/core/offload/canonical.py
@@ -60,6 +60,7 @@ offload store machinery is reused unchanged — 案B removes only the field-gues
 """
 from __future__ import annotations
 
+import json
 from typing import Any, Callable, TypedDict
 
 
@@ -424,6 +425,43 @@ def sandboxed_exec_to_canonical(result: dict) -> CanonicalToolResult:
     if returncode:  # nonzero (or truthy) only — a 0 exit is not actionable signal
         meta["returncode"] = returncode
     return CanonicalToolResult(text=text, attachments=[], source_ref=None, meta=meta)
+
+
+# The exact dispatch-envelope key set ``unwrap_dispatch_envelope`` also tests for (#2681 Bucket A) —
+# ``shell_to_canonical`` uses the SAME shape heuristic to find the enveloped payload.
+_DISPATCH_ENVELOPE_KEYS = frozenset({"status", "data", "error"})
+
+
+def shell_to_canonical(result: dict) -> CanonicalToolResult:
+    """``shell`` tool result → canonical (#2681 Bucket A — the scout-flagged latent file-class bug:
+    PR-F1's ``CANONICAL_TODO`` triage missed that ``shell`` is a TEXT producer, so its readable STDOUT
+    was shown to the LLM as a whole-dict ``structured`` blob instead of clean text).
+
+    ``shell`` is #2593 thin pipeline-DSL sugar over ``sandboxed_exec`` whose ``_handle`` (locked design)
+    returns ONLY the command's STDOUT — JSON-decoded when it parses (so ``verify: schema`` can apply to
+    a JSON-emitting command), else the raw text. ``stderr``/``returncode`` never reach this seam (they
+    are dropped one layer up, by that same locked design) — the one respect this mapper CANNOT mirror
+    ``sandboxed_exec_to_canonical`` (whose ``returncode`` signal meta this would carry, were it visible
+    here); ``meta`` is therefore always empty for ``shell``.
+
+    The shape THIS mapper actually receives (post ``unwrap_dispatch_envelope``) is one of:
+
+    - the dispatch envelope ``{"status": ..., "data": <value>}`` — the common case (a plain-text
+      command, or JSON stdout that decoded to a non-``dict`` such as a list/number/bool/``None``: the
+      envelope could not be peeled because peeling requires a ``dict`` ``data``);
+    - ``<value>`` directly when stdout decoded to a ``dict`` (already peeled one envelope layer).
+
+    Either way, mirroring ``sandboxed_exec``'s "stdout IS the text" treatment: ``value`` renders as the
+    readable ``text`` body — verbatim when it is already a ``str``, else ``json.dumps``'d so a
+    JSON-emitting command's output stays fully legible — and there is NO ``structured`` attachment (the
+    whole-dict blob this mapper replaces)."""
+    if "data" in result and set(result) <= _DISPATCH_ENVELOPE_KEYS:
+        value = result["data"]
+    else:
+        value = result
+    text = value if isinstance(value, str) else json.dumps(value, default=str)
+    text = _explicit_empty(text, "(no output)")
+    return CanonicalToolResult(text=text, attachments=[], source_ref=None, meta={})
 
 
 def chunks_to_canonical(result: dict) -> CanonicalToolResult:
--- a/src/reyn/tools/shell.py
+++ b/src/reyn/tools/shell.py
@@ -95,10 +95,10 @@ async def _handle(args: Mapping[str, Any], ctx: ToolContext) -> Any:
         return stdout
 
 
-from reyn.core.offload.canonical import CANONICAL_TODO  # noqa: E402
+from reyn.core.offload.canonical import shell_to_canonical  # noqa: E402
 
 SHELL = ToolDefinition(
-    canonical=CANONICAL_TODO,
+    canonical=shell_to_canonical,
     name="shell",
     description=_SHELL_DESCRIPTION,
     parameters=_SHELL_PARAMETERS,
--- a/tests/test_2593_pipeline_shell_step.py
+++ b/tests/test_2593_pipeline_shell_step.py
@@ -121,9 +121,10 @@ async def test_shell_step_threads_pipe_data_to_stdin_and_stdout_to_output():
         state_log=None,
         run_id="run-shell-stdin",
     )
-    # #2425 PR-2: the shell tool's parsed-JSON dict (no "kind") is an
-    # unregistered-kind result → the whole dict is the sole structured attachment.
-    expected = {"text": "", "structured": {"n": 3, "msg": "hi"}}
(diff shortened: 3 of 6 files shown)


tya5:
**[dogfood-coder]** — co-vet on #2807, verify-don't-infer. **Verdict: CLEAR to merge.**

## 1. Shell's text mapper — VERIFIED, including the envelope-peel and meta-asymmetry claims

**Envelope detection is a correct continuation of `unwrap_dispatch_envelope`, not a duplicate/conflicting reimplementation.** Read `unwrap_dispatch_envelope` directly: its peel loop condition is `isinstance(inner.get("data"), dict)` — it *only* peels when `data` is itself a dict, and stops otherwise. So when shell's stdout decodes to a JSON dict, `unwrap_dispatch_envelope` has already fully peeled it by the time `shell_to_canonical` runs (the `<value>` case). When stdout is plain text or decodes to a non-dict JSON value (list/number/bool/None), that peel condition is false and the envelope survives intact — exactly the case `shell_to_canonical`'s own `if "data" in result and set(result) <= _DISPATCH_ENVELOPE_KEYS` check exists to handle. The key-set literal (`{"status", "data", "error"}`) is byte-identical to `unwrap_dispatch_envelope`'s own check. Verified this is complementary, not redundant.

**`meta={}` asymmetry vs `sandboxed_exec` is structurally forced, not an oversight.** Read `shell.py`'s `_handle` directly: it calls `handle_sandboxed_exec` and extracts *only* `result["stdout"]` (line 91) — `stderr`/`returncode` are read by the underlying op but never propagated into `shell._handle`'s own return value (confirmed `sandboxed_exec_to_canonical` reads `returncode` from a raw dict `shell._handle` never returns). `shell_to_canonical` genuinely cannot carry meta that never reaches its input — the asymmetry is real and locked by #2593, not something this mapper could fix.

## 2. Falsify — VERIFIED myself, not accepting the PR's claim

Flipped `shell.py`'s `canonical=` back to `CANONICAL_TODO` directly (sed edit + backup, restored after — no `git stash`), reran the affected tests: **exactly 6 failures**, matching the claim precisely — `tests/test_2681_shell_canonical.py` (4) + the 2 updated `test_2593_pipeline_shell_step.py` assertions, each showing the whole-dict `structured` blob regression (`{'structured': {'msg': 'hello'}}` where clean `text` was expected). Restored — 14/14 green again.

## 3. Ledger removal — VERIFIED

`"shell"` is gone from `_CANONICAL_TODO_GRANDFATHERED`. Ran `tests/test_fp0056_canonical_coverage_gate.py` myself (117/117 pass, including `test_canonical_todo_is_ratcheted_to_the_grandfather_ledger`'s `live_todo == set(_CANONICAL_TODO_GRANDFATHERED)` set-equality assert) — confirms the ratchet holds with `shell` removed from both sides.

## 4. `_NO_INNER_DISCRIMINATOR_MAPPERS` exemption — VERIFIED as the correct, narrow fix, traced the actual false-positive mechanism

Read `_status_echo_offense`'s probe logic directly: it drives a mapper with `{"status": PROBE}` and flags an offense if the rendered text contains `PROBE` without the mapper raising `CanonicalDiscriminatorMiss`. Traced `shell_to_canonical({"status": PROBE})` through its own logic: no `"data"` key present → `value = result` (the whole `{"status": PROBE}` dict) → `json.dumps`'d into text, which necessarily *does* contain the probe string. This is a genuine, unavoidable false positive for `shell_to_canonical` specifically: its entire contract is "render the whole value verbatim, no field selection" — there's no inner discriminator to miss, so it structurally cannot exhibit the M3 bug class (dropping information on a discriminator miss), yet the probe's binary heuristic (status-in-text ⟹ offense) can't distinguish that from a real catch-all. A narrow, name-based, documented exemption is the right minimal fix — not a blanket skip that could mask a *real* future M3 regression in some other mapper.

- Traced both the envelope-peel logic and the meta-asymmetry claim through the actual `unwrap_dispatch_envelope`/`shell._handle`/`sandboxed_exec_to_canonical` source, not the PR's characterization.
- **Falsified independently**: reverted the `canonical=` declaration (not via `git stash` — direct edit + manual restore), confirmed the exact 6-test regression, restored, reconfirmed green.
- Ran `ruff check`: clean. `test_tier_audit.py --strict` on all 4 changed test files: 30/30, 0 violations. `verify_module_docstrings.py`: clean.
- Ran the full pytest suite myself: 8061 passed, 9 failed (exact match to the established pre-existing baseline across every full-suite run this session), 18 skipped — no new regressions.
- Traced the M3 probe's exact false-positive mechanism against `shell_to_canonical`'s actual logic rather than accepting the exemption's stated rationale.

No fix-before-merge items.

PR #2809, opened 2026-07-10
feat: structured canonical mappers for 24 record-read producers — #2681 Bucket B

tya5:
**[per-PR-coder]** —

part of #2681

## Architect ruling (Decision #1, cited)

`STRUCTURED_PASSTHROUGH` stays restricted to the admin-6 (external/protocol payloads
needing verbatim structure — MCP responses / install manifests). The producers in
this bucket are **internal record-reads** (single-record "describe" views and
record-list "list"/"search" views), so per the ruling they get **real structured
mappers**, not passthrough: a short bounded, human/LLM-readable `text` summary +
the record(s) as a `structured` attachment on `CanonicalToolResult`. The admin-6
set is untouched (`test_structured_passthrough_membership_is_exactly_the_admin_6`
stays green).

## The shared helper

One reusable helper in `src/reyn/core/offload/canonical.py`:

- `_records_to_canonical(text, records)` — the shape every mapper below returns:
  `text` (bounded summary) + `records` (a dict or list) as ONE `{"kind": "structured", "data": records}` attachment.
- `_bounded_join(records, key, limit=10)` — a small preview-name joiner (e.g. `"acme, globex, +3 more"`)
  used by the list-shaped mappers so `text` stays short even when the record list is large (the
  full list always lives in the attachment, never truncated there).

## Applied — 24 assigned + 1 punted from Bucket C = 25

- `list_memory` (memory.py)
- `list_actions`, `search_actions`, `describe_action`, `invoke_action` (universal_catalog.py)
- `list_agents`, `describe_agent` (catalog.py)
- `describe_mcp_tool`, `list_mcp_servers`, `list_mcp_tools`, `list_mcp_resources`,
  `list_mcp_resource_templates`, `list_mcp_prompts` (mcp.py)
- `mcp_search_registry` (mcp_verbs.py)
- `cron_list` (cron.py)
- `task.create`, `task.update_status`, `task.get`, `task.list`, `task.add_dependency`,
  `task.remove_dependency`, `task.repoint_dependency`, `task.abort`, `task.assign`
  (core/op_runtime/task.py + tools/task_ops.py, dual-registered — both seams now declare
  the SAME shared `task_op_to_canonical`)
- `topology_create` (tools/topology_create.py) — **punted here from Bucket C's sweep**:
  its success shape `{status: "created", name, kind, members, leader, profiles}` echoes
  the FULL created config, a genuine record, not a status ack, so it belongs in Bucket B's
  triage rather than Bucket C's status-ack set.

### `invoke_action` disposition (flagged per the task's reviewer-call)

`invoke_action` normally **delegates** classification to its resolved target via the
`_canonical_source` tag it injects (`universal_catalog.py::_handle_invoke_action`): when the
target's handler returns a dict, canonicalization dispatches through the **target's own**
mapper and `invoke_action`'s declaration is never consulted.

I verified this is *not* airtight, against the real `dispatch_tool` → `unwrap_dispatch_envelope`
→ `to_canonical` chain (no mocks — see `tests/test_2681_bucket_b_structured_mappers.py`, and a
throwaway script exercising the real dispatcher during development): the tag-injection guard is
`isinstance(result, dict)`, so when the delegated target's handler returns a **non-dict** value
(a bare list/scalar — concretely, `list_memory` / `list_agents`, both in this bucket, when invoked
via `invoke_action`), the tag is skipped and the OUTER `invoke_action` tag survives instead. In
that case `invoke_action`'s own declaration IS reached, receiving the still-wrapped dispatch
envelope (`{"status": "ok", "data": <the raw value>}`).

Since the ledger requires removing `invoke_action` from `CANONICAL_TODO` (it was one of the 24),
I gave it a small **defensive fallback mapper** (`invoke_action_to_canonical`, reusing the shared
records+summary shape) rather than leaving it undeclared or reverting to CANONICAL_TODO. It's
generic (doesn't know the delegated target's specific record shape) but lossless and consistent
with the bucket's overall approach. Documented in its docstring; flagging here for reviewer
awareness in case a different disposition is preferred.

No other producer needed an escalation-valve exception (none were genuinely un-summarizable).

## Ledger edit

All 25 removed from `_CANONICAL_TODO_GRANDFATHERED` in
`tests/test_fp0056_canonical_coverage_gate.py`. `test_canonical_todo_is_ratcheted_to_the_grandfather_ledger`
(live == ledger) stays green. Rebased onto origin/main after Bucket A (#2807, removed `shell`)
merged — clean, disjoint-entry rebase as expected.

`tests/test_fp0056_identity_dispatch.py::test_canonical_todo_producer_gets_the_whole_dict_fallback`
switched its example producer from `list_mcp_servers` (now migrated) to `cron_register` (still
CANONICAL_TODO, untouched by this PR).

## Test plan

- [x] `tests/test_2681_bucket_b_structured_mappers.py` (new, Tier 1, real dispatch/`to_canonical`
      calls, no mocks): spot-checks across memory / mcp / task / cron / catalog / topology (text
      summary present + structured attachment carries the records), the `invoke_action`
      delegation-vs-own-mapper pair, a falsify pin (mapped result must differ from the naive
      whole-dict fallback — `_fallback_structured`), a callable-membership sweep of all 25
      producers (never `CANONICAL_TODO` / `STRUCTURED_PASSTHROUGH`), and an admin-6 sanity check
- [x] Manual cp-based falsify (no git checkout/stash — concurrent worktrees share one `.git`
      object store): reverted `list_memory`'s declaration to `CANONICAL_TODO` via a `cp` backup,
      confirmed 3 tests go RED (including the falsify pin), restored via `cp`
- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict` on all changed/added test files
- [x] `pytest` (full suite): 8109 passed, 21 skipped; 5 pre-existing route-mount failures
      (`test_fp0001_a2a_endpoints.py`, `test_a2a.py`, `test_mcp_sse.py`, `test_plugin_loader.py`,
      `test_resources_router.py`) confirmed identical on unmodified origin/main — unrelated to
      this change, deselected
- [x] `python scripts/verify_module_docstrings.py` on all changed src files

## Note for reviewers (per lead-coder coordination)

Buckets A (#2807, merged) and C also edit the same `_CANONICAL_TODO_GRANDFATHERED` frozenset,
each removing disjoint entries. This PR is expected to be the last of the three to merge; once
it lands the ledger empties out and #2681 completes. Worktree kept per instructions.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/reyn/core/offload/canonical.py
+++ b/src/reyn/core/offload/canonical.py
@@ -1107,6 +1107,283 @@ def ask_user_to_canonical(result: dict) -> CanonicalToolResult:
     )
 
 
+# ── FP-0056 #2681 Bucket B — genuinely-structured record-read producers ───────────────────────────
+#
+# Owner Decision #1 restricts STRUCTURED_PASSTHROUGH to the admin-6 (external/protocol payloads
+# needing verbatim structure — MCP responses / install manifests). The 24 producers mapped below are
+# internal record-reads (single-record "describe" views, and record-LIST "list"/"search" views) —
+# they fit a CanonicalToolResult with a short bounded ``text`` summary + the record(s) as a
+# ``structured`` attachment. A record-list is unbounded by nature, so a bounded summary + full
+# structured detail is MORE correct than a raw whole-dict passthrough for these.
+#
+# SUCCESS shape only, uniformly: every producer below has an error path that carries a dedicated
+# error-message field (``error`` / ``error_message`` / ``error_kind``) or ``isError``, so the shared
+# error seam (piece #1, ``is_error_result`` in :func:`to_canonical`) intercepts it BEFORE any of these
+# mappers runs — no per-mapper error branch needed here (mirrors ``file_to_canonical`` /
+# ``reyn_src_to_canonical`` / the rest of this module's success-only mappers).
+
+
+def _bounded_join(records: Any, key: str, *, limit: int = 10) -> str:
+    """Join up to ``limit`` records' ``key`` field into a comma-separated preview string, or ``""``
+    when ``records`` isn't a list or no record carries a truthy ``key``. BOUNDED by construction: the
+    full record list — however large — always lives in the ``structured`` attachment; this preview is
+    only ever a short ``text`` accent, never the data's only home."""
+    if not isinstance(records, list):
+        return ""
+    names = [str(r[key]) for r in records if isinstance(r, dict) and r.get(key)]
+    if not names:
+        return ""
+    shown = names[:limit]
+    remaining = len(names) - len(shown)
+    joined = ", ".join(shown)
+    return f"{joined}, +{remaining} more" if remaining > 0 else joined
+
+
+def _records_to_canonical(text: str, records: Any) -> CanonicalToolResult:
+    """THE shared shape for every #2681 Bucket B record-read mapper below: a short bounded
+    LLM-readable ``text`` summary (e.g. "5 memories", "3 MCP servers: a, b, c", "task <id>: <status>")
+    + the record(s) — a single dict (a "describe" view) or a list of dicts (a "list"/"search" view) —
+    as ONE ``structured`` attachment. Centralizing this (rather than each mapper building its own
+    ``CanonicalToolResult``) is the reusable seam these 24 producers share by construction."""
+    return CanonicalToolResult(
+        text=text, attachments=[{"kind": "structured", "data": records}], source_ref=None, meta={},
+    )
+
+
+def memory_list_to_canonical(result: dict) -> CanonicalToolResult:
+    """``list_memory`` result -> canonical (#2681 Bucket B). The handler returns a BARE LIST (browse
+    entries: ``{path, count}`` at the root/layer level, or ``{slug, name, description}`` at the leaf
+    level) — a non-dict handler return, so ``unwrap_dispatch_envelope`` does not peel the dispatch
+    envelope (its ``data`` isn't itself a dict), and this mapper receives ``{"status": "ok",
+    "data": [...]}`` rather than the bare list directly (verified against the real dispatch chain,
+    not shape-inference alone)."""
+    records = result.get("data") or []
+    n = len(records) if isinstance(records, list) else 0
+    text = f"{n} memory {'entry' if n == 1 else 'entries'}."
+    return _records_to_canonical(text, records)
+
+
+def list_agents_to_canonical(result: dict) -> CanonicalToolResult:
+    """``list_agents`` result -> canonical (#2681 Bucket B). Same bare-list handler shape as
+    ``list_memory`` (see :func:`memory_list_to_canonical`) — the dispatch envelope survives, so this
+    mapper reads ``result["data"]``. Entries are either ``{cluster, count}`` (root browse) or
+    ``{name, role}`` (one cluster's agents); a ``name`` field (present only in the latter) drives the
+    bounded preview."""
+    records = result.get("data") or []
+    n = len(records) if isinstance(records, list) else 0
+    preview = _bounded_join(records, "name")
+    label = "agent" if preview else "cluster"
+    text = f"{n} {label}{'s' if n != 1 else ''}" + (f": {preview}" if preview else "") + "."
+    return _records_to_canonical(text, records)
+
+
+def describe_agent_to_canonical(result: dict) -> CanonicalToolResult:
+    """``describe_agent`` result -> canonical (#2681 Bucket B). A SINGLE record — the raw agent entry
+    dict (``name``, ``role``, optional ``cluster``/others) — carried whole in the structured
+    attachment; ``text`` names the agent + role."""
+    name = result.get("name", "")
+    role = result.get("role") or "(no role)"
+    text = f"agent {name}: {role}."
+    return _records_to_canonical(text, result)
+
+
+def list_actions_to_canonical(result: dict) -> CanonicalToolResult:
+    """``list_actions`` result -> canonical (#2681 Bucket B). ``items`` is the current (enriched)
+    page; ``total`` is the full catalog count across all categories — the summary reports BOTH so the
+    LLM knows whether it is seeing everything or a page. The FP-0043 ``hint`` (search_actions
+    unavailable in this session) is appended when present — instructional signal the LLM must relay,
+    not bulk data."""
+    items = result.get("items") or []
+    total = result.get("total", len(items))
+    text = f"{len(items)} of {total} action(s)."
+    hint = result.get("hint")
+    if hint:
+        text = f"{text}\n{hint}"
+    return _records_to_canonical(text, items)
+
+
+def search_actions_to_canonical(result: dict) -> CanonicalToolResult:
+    """``search_actions`` result -> canonical (#2681 Bucket B). ``items`` (each
+    ``{qualified_name, short_description, score}``) is the ranked semantic-match list."""
+    items = result.get("items") or []
+    total = result.get("total", len(items))
+    text = f"{total} matching action(s)."
+    return _records_to_canonical(text, items)
+
+
+def describe_action_to_canonical(result: dict) -> CanonicalToolResult:
+    """``describe_action`` result -> canonical (#2681 Bucket B). A SINGLE resolved-action record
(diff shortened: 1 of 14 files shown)


tya5:
**[dogfood-coder]** — co-vet on #2809, verify-don't-infer (final bucket, closes #2681). **Verdict: CLEAR to merge (post-rebase-onto-C, CI green).**

## 1. Shared helper applied correctly, behavior-preserving — VERIFIED

`_records_to_canonical(text, records)` and `_bounded_join(records, key, limit=10)` are genuinely shared (one definition, reused across all mappers, not per-mapper copies) — spot-checked several call sites (`memory_list_to_canonical`, `list_agents_to_canonical`, `describe_agent_to_canonical`, `topology_create_to_canonical`) and confirmed each just supplies its own bounded-text summary logic then delegates the attachment-building to the shared helper. Confirmed `task.py` (op_runtime, 9 `register(...)` calls) and `task_ops.py` both import and use the **identical** `task_op_to_canonical` function object — genuinely one shared mapper across both registration seams, not two independently-written duplicates.

## 2. `invoke_action` defensive fallback — independently traced through 3 files, disposition confirmed correct

Traced the delegation-hole claim through the real code rather than accepting the PR's characterization:
- `universal_catalog.py::_handle_invoke_action` (line 1529): `if isinstance(result, dict) and "_canonical_source" not in result:` — confirmed the tag-injection guard is exactly `isinstance(result, dict)`, so a non-dict delegated-handler return (e.g. `list_memory`/`list_agents`, both genuinely returning bare lists per their own mapper docstrings) skips this injection entirely.
- `router_loop.py:2764-2765`: `if isinstance(result, dict): result.setdefault("_canonical_source", name)` — this is the OUTER envelope-level tag (unconditional for any tool-registry dispatch, since the envelope itself is always a dict), with its own comment explicitly documenting the "deepest wins" design intent.
- `canonical.py::extract_canonical_source` (with `_walk`): confirmed "deepest tag wins" is the real, documented behavior, not an assumption.

Chaining these three: when `invoke_action` delegates to a target whose handler returns non-dict, the inner (target-specific) tag never gets attached, so `extract_canonical_source` finds only the outer `invoke_action` tag — `invoke_action_to_canonical` genuinely does get dispatched to in that case, exactly as claimed. This is a real, structural gap in the delegation mechanism (not something this PR should fix — that's separate scope from "give producers real mappers"), and the disposition (a defensive fallback mapper, not reverting to `CANONICAL_TODO`, not silently leaving it undeclared) is the right minimal-footprint choice: it closes the ledger entry validly, is lossless, and is honestly documented as a fallback rather than the common path. Confirmed a dedicated test (`test_invoke_action_own_mapper_covers_the_non_dict_delegated_target_case`) pins exactly this scenario.

## 3. `topology_create` — VERIFIED

`ToolDefinition.canonical` now declares `topology_create_to_canonical` (was `CANONICAL_TODO`). Read the mapper: single-record shape (`{status, name, kind, members, leader, profiles}`), text summary + whole record as `structured` — matches its "genuine record, not a status ack" classification.

## 4. Falsify — VERIFIED myself, no `git stash`/`checkout`

Used a `cp` backup + direct edit (per the parallel-coder shared-central-file hazard convention — no `git stash`): flipped `list_memory`'s `canonical=` from `memory_list_to_canonical` back to `CANONICAL_TODO`, reran `test_2681_bucket_b_structured_mappers.py` — **exactly 3 failures** (the falsify pin, the dedicated `list_memory` test, and the callable-membership sweep across all 25 producers), matching the PR's claim precisely. Restored via `cp`, reconfirmed 12/12 green, `git diff` on the file shows zero residual change.

## 5. Ledger — VERIFIED (25 entries removed, counted independently)

Diffed `_CANONICAL_TODO_GRANDFATHERED` myself and counted the removed entries against the PR's claimed 25 — they match exactly (`cron_list`, `describe_action/agent/mcp_tool`, `invoke_action`, `list_actions/agents/mcp_*` ×6/`memory`, `mcp_search_registry`, `search_actions`, 9× `task.*`, `topology_create` = 25). `task.comment`/`task.heartbeat`/`task.register_unblock_predicate` correctly remain (out of this bucket's scope, consistent with `task_ops.py`'s own conditional gate on `_STRUCTURED_TASK_OP_KINDS`). Ran the ratchet test (`test_canonical_todo_is_ratcheted_to_the_grandfather_ledger`, `live == ledger` set-equality) myself against this PR's own base (pre-C-rebase) — passes. Per lead-coder's note, the *empty-ledger final state* only manifests after this PR rebases onto C-merged main; that's a mechanical rebase step, not something this diff's own correctness depends on — nothing in this PR's logic assumes or requires the ledger to already be empty.

- Traced the `invoke_action` delegation-hole claim through 3 separate files' actual logic (not the PR's summary of it) — the standout finding of this review, independently confirmed accurate.
- Read the shared helper + spot-checked 4+ individual mappers directly.
- **Falsified independently** (`cp` backup, not `git stash`/`checkout` — respecting the shared-worktree hazard): confirmed exactly 3 tests regress, restored cleanly, zero residual diff.
- Counted the ledger diff's 25 removed entries against the PR's own claimed list.
- Ran `ruff check` (clean), `test_tier_audit.py --strict` on all 3 changed/added test files (24/24, 0 violations), `verify_module_docstrings.py` on all 10 changed src files (clean).
- Ran the full pytest suite myself: 8108 passed, 9 failed (exact match to the established pre-existing baseline every full-suite run this session has shown), 18 skipped — no new regressions.

No fix-before-merge items. Closes out the #2681 arc cleanly once rebased and merged.

tya5:
**[per-PR-coder]** — Re-rebased onto C-merged latest main (#2808 now in). The ledger region overlapped as expected; resolved by **union** per the lead's direction. Now **MERGEABLE**, all 4 CI gates green locally.

## Conflict resolution (6 files, union)

- **Ledger** (`test_fp0056_canonical_coverage_gate.py`): HEAD (post A+C) held exactly my Bucket B set; my side held exactly the Bucket C set. Union removes both → **`_CANONICAL_TODO_GRANDFATHERED` is now `frozenset()` (empty)**. With shell (A), 25 record-reads (B), 25 status producers + 3 status task ops (C), and present (#2692) all migrated, the #2681 burn-down is complete. The ratchet gate `live == ledger` is green against the empty set — going forward NO producer may adopt CANONICAL_TODO.
- **`op_runtime/task.py`** + **`tools/task_ops.py`**: all 12 task ops now map — B's 9 record ops share `task_op_to_canonical`, C's 3 status ops (heartbeat/register_unblock_predicate/comment) keep their status-text mappers. Folded both into C's `_TASK_OP_CANONICAL` dict (removed my now-redundant `_STRUCTURED_TASK_OP_KINDS` frozenset); dual-registration verified conflict-free (`declare_canonical` would raise on a mismatch — it doesn't).
- **`cron.py` / `memory.py` / `mcp_verbs.py`**: union of C's mapper imports + mine; each file's `CANONICAL_TODO` import dropped where all its ops are now mapped (ruff F401 clean).

## Two adjacent tests updated (both C-authored, pointed at now-migrated live producers)

Since the live CANONICAL_TODO set is now empty, two tests that used a *real* ledger member as their CANONICAL_TODO example had to switch to a **synthetic fixture source** declared via `declare_canonical(...)` (the sentinel's fallback behavior is still live code; the fixture id is neither an op kind nor a ToolDefinition, so it never enters the ratchet gate's registry-derived set):
- `test_fp0056_identity_dispatch.py::test_canonical_todo_producer_gets_the_whole_dict_fallback` (my earlier `cron_register` pick — C mapped it)
- `test_fp0056_canonical_fallback_event.py::test_fallback_reason_maps_each_fire_condition_to_its_category` (C's `topology_create` pick — B mapped it)

## Gates (rebased tree)
- [x] `test_tier_audit --strict` on all 4 changed/added test files — 27/27 OK
- [x] `pytest` — 8123 passed, 21 skipped; the same 5 pre-existing environment-specific route-mount failures (a2a/mcp_sse/resources/plugin_loader/fp0001) deselected, confirmed identical on unmodified main

No `git stash`/`checkout` used anywhere in the resolution (Edit tool only + cp-scratch backups earlier), per the concurrent-worktree shared-`.git` caution. Worktree kept. This is the last of the three buckets — on merge, #2681 completes.

PR #2810, opened 2026-07-10
fix: per-path in-process lock serializes concurrent same-path file writes (#2782)

### LINKED ISSUE #2782: file tool ops run sync filesystem I/O on the event loop (host_backend.py) — needs design; hooks_add same class, low priority

tya5:
**[per-PR-coder]** — Closes #2782

## Problem
#2794 offloaded `edit_file`'s read-modify-write into ONE `asyncio.to_thread` job — atomic *within* a single op (no `await` between read and write). But it removed the implicit cross-op serialization that single-threaded, single-event-loop execution used to provide. **Demonstrated race**: two concurrent `edit_file` ops on the SAME path now run in different worker threads → both read the pre-edit content → both write → one edit is silently lost (classic read-modify-write lost update). Reachable paths: pipeline `parallel`/`for_each` fan-out (see `core/pipeline/executor.py`'s "read-modify-write race" docstring), concurrent sub-agents, concurrent A2A sessions sharing a `base_dir`.

## Fix (architect recommendation D — in-process, NOT flock)
New `src/reyn/core/op_runtime/path_locks.py`: a process-global, **loop-aware** registry of per-path `asyncio.Lock`s keyed by the **resolved absolute path**. This mirrors the existing `runtime/agent_locks.py` per-key lock family (same mechanism + the #1762 loop-aware-keying rationale — a lock is bound to the loop it's first awaited on; keying by running loop keeps each pytest-asyncio loop isolated). A **dedicated** registry (not shared with `agent_locks`) because paths and agent names are different key domains. Locks are acquired at the **async op layer** in `file.py`, held across the whole read-modify-write / mutate, released in `finally` via `async with`.

In-process by design (ADR-0018 single-process end-state; cross-process mutual exclusion is deferred to the A2A-server model, a separate initiative) — deliberately **not** flock/fcntl. The single-writer server model architecturally backs this choice.

### Op-scope enumeration (fix-class completeness sweep)
Every op in `file.py` that MUTATES a path now takes the SAME per-path lock, so a concurrent edit+write (or edit+delete) on one path also serializes:

| op | locks | why |
|----|-------|-----|
| `edit` | `path` (across the `to_thread` read-modify-write) | the demonstrated race |
| `write` | `path` (across the encoding pre-read + `write_file`) | a blind overwrite straddling a concurrent edit's read/write would discard the edit |
| `delete` | `path` | a concurrent edit reading before / writing after a delete would "resurrect" the file |
| `regenerate_index` | `output_path` (the file actually written — NOT the source dir it reads) | same-path writer completeness; error branches touch no disk, run outside the lock |
| `move` | **both** `src` + `dst` via `locked_paths(...)` | move is effectively delete(src)+write(dst) |

**Deadlock-ordering note (move):** `move` locks two paths. `locked_paths()` deduplicates and acquires in **sorted resolved-path order** (a fixed global ordering) so two concurrent moves over the same two paths with reversed src/dest (`a→b` ∥ `b→a`) — the classic two-lock deadlock shape — always take `a` before `b` regardless of role, and cannot deadlock.

Read-only ops (`read`/`glob`/`grep`/`stat`) and `mkdir` (dir creation, not a same-path file write) are unchanged.

## Guard test (load-bearing) + cp-falsify RED
`tests/test_file_path_lock_2782.py` (Tier 2, real `Workspace`/`EventLog`/`OpContext`/`file.handle`/`asyncio.gather` — no mocks):

1. **`test_concurrent_edit_file_same_path_both_edits_survive`** — 2 concurrent `edit` ops making DIFFERENT edits via `asyncio.gather` through the real op-dispatch path. A `threading.Barrier` wraps `read_file_bytes` to *deterministically* force both reads before either write (the exact lost-update interleaving), rather than relying on scheduling luck. With the lock, the 2nd op's `to_thread` can't start until the 1st's lock releases (post-write), the barrier times out (caught), and both edits survive.
2. **`test_concurrent_edit_and_write_same_path_no_lost_update`** — cross-op-kind completeness: concurrent `edit`+`write` on one path → final content is a clean full result of one op, never a torn merge.
3. **`test_move_reversed_src_dest_does_not_deadlock`** — `a→b` ∥ `b→a`, bounded by `asyncio.wait_for` so a regression to unordered acquisition fails as a timeout not a hang.

**cp-falsify:** backed up `file.py` to scratch, removed the `edit` lock acquisition, re-ran the guard test **5×** → RED every time, reproducing the exact lost update (`'aaa line\nBBB line\n'` / `'AAA line\nbbb line\n'` — one edit silently gone). Restored via `cp` (never git on uncommitted work), confirmed byte-identical, re-ran → green.

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_file_path_lock_2782.py` — 3/3 OK, 0 Tier-4 violations
- [x] `pytest` (full suite) — 8060 passed; 4 pre-existing unrelated failures (`test_a2a_routes_mounted` / `test_mcp_routes_mounted` / `test_plugin_loader` / `test_resources_router` — FastAPI route-mounting in this local env; verified they fail identically on clean origin/main with my change stashed)
- [x] Guard test passes with the fix; cp-falsify goes RED (5/5) with the lock removed, restored byte-identical
- [x] Existing file-op suites (`test_file_ops_offload_2782`, `test_op_runtime_file_mkdir_move_stat`, `test_file_edit_registry`, `test_edit_file_preview_1418`, `test_file_read_encoding_1452`) — 52 passed, no regressions

Do NOT merge — architect co-vets (they hold the live repro); lead merges on verdict.

### DIFF
--- a/src/reyn/core/op_runtime/file.py
+++ b/src/reyn/core/op_runtime/file.py
@@ -13,6 +13,7 @@ from reyn.schemas.models import FileIROp
 from . import register
 from .context import OpContext
 from .context import sandbox_policy_from_ctx as _sandbox_policy_from_ctx
+from .path_locks import get_path_lock, locked_paths
 
 _WRITE_OPS = frozenset({"write", "edit", "delete", "regenerate_index", "mkdir", "move"})
 _READ_OPS = frozenset({"read", "glob", "grep", "stat"})
@@ -237,15 +238,20 @@ async def handle(op: FileIROp, ctx: OpContext) -> dict:
         # changed. The pre-read is best-effort (skipped if read is denied).
         _prev_enc: str | None = None
         _prev_size: int | None = None  # #1466: bytes of the file before overwrite
-        try:
-            _prev_bytes, _existed = ctx.workspace.read_file_bytes(op.path)
-            if _existed:
-                _, _prev_enc = decode_text_or_none(_prev_bytes)
-                _prev_size = len(_prev_bytes)  # zero extra I/O — reuse the encoding pre-read
-        except PermissionError:
-            pass
-        _content = op.content or ""
-        ctx.workspace.write_file(op.path, _content)
+        # #2782 path-locking step: hold the SAME per-path lock `edit` holds, so
+        # a concurrent edit's read-modify-write can't interleave with this
+        # blind overwrite (either direction would silently discard the other
+        # op's change — a lost update).
+        async with get_path_lock(_resolve_for_gate(ctx, op.path)):
+            try:
+                _prev_bytes, _existed = ctx.workspace.read_file_bytes(op.path)
+                if _existed:
+                    _, _prev_enc = decode_text_or_none(_prev_bytes)
+                    _prev_size = len(_prev_bytes)  # zero extra I/O — reuse the encoding pre-read
+            except PermissionError:
+                pass
+            _content = op.content or ""
+            ctx.workspace.write_file(op.path, _content)
         ctx.events.emit("tool_executed", op="write_file", path=op.path)
         result: dict = {
             "kind": "file", "op": "write", "path": op.path, "status": "ok",
@@ -422,7 +428,11 @@ async def handle(op: FileIROp, ctx: OpContext) -> dict:
         }
 
     if op.op == "delete":
-        deleted = ctx.workspace.delete_file(op.path)
+        # #2782: same per-path lock as edit/write — otherwise a concurrent
+        # edit's read-modify-write can straddle this delete and "resurrect"
+        # the file (edit read before the delete, writes back after it).
+        async with get_path_lock(_resolve_for_gate(ctx, op.path)):
+            deleted = ctx.workspace.delete_file(op.path)
         ctx.events.emit("tool_executed", op="delete_file", path=op.path, deleted=deleted)
         return {"kind": "file", "op": "delete", "path": op.path, "status": "ok", "deleted": deleted}
 
@@ -433,6 +443,15 @@ async def handle(op: FileIROp, ctx: OpContext) -> dict:
         return await _execute_edit(op, ctx)
 
     if op.op == "regenerate_index":
+        # #2782: lock the SAME per-path key on the actually-written
+        # `output_path` (not the source `path`, which is a directory of
+        # *.md sources being read, not the file being mutated). The
+        # error branches (missing output_path/entry_template) never
+        # touch disk, so they run outside the lock.
+        if op.output_path:
+            resolved_output = _resolve_for_gate(ctx, op.output_path)
+            async with get_path_lock(resolved_output):
+                return _execute_regenerate_index(op, ctx)
         return _execute_regenerate_index(op, ctx)
 
     if op.op == "mkdir":
@@ -456,7 +475,14 @@ async def handle(op: FileIROp, ctx: OpContext) -> dict:
                 "kind": "file", "op": "move", "path": op.path,
                 "status": "error", "error": "dest_path is required for move",
             }
-        moved = ctx.workspace.move_path(op.path, op.dest_path)
+        # #2782: move touches TWO paths (source is effectively deleted, dest is
+        # effectively written) — lock BOTH via `locked_paths`, which acquires
+        # them in a fixed sorted order (never src-then-dest vs dest-then-src)
+        # so a move and a reverse move can never deadlock against each other.
+        src_resolved = _resolve_for_gate(ctx, op.path)
+        dst_resolved = _resolve_for_gate(ctx, op.dest_path)
+        async with locked_paths(src_resolved, dst_resolved):
+            moved = ctx.workspace.move_path(op.path, op.dest_path)
         if not moved:
             ctx.events.emit("tool_executed", op="move", path=op.path, found=False)
             return {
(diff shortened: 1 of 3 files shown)


tya5:
**[architect]** — co-vet verdict: **CLEAR to merge.** Faithful implementation of the ADR-0018/#2782 D design; all four focus points verified, and I independently strip-falsified the guard **and** re-ran my original lost-update repro against this branch.

## path_locks.py — matches the D design exactly
Loop-aware per-path `asyncio.Lock` registry (mirrors `agent_locks.py`/#1762, a **dedicated** registry with the key-domain rationale spelled out), keyed by the **resolved absolute path**, held across the whole read-modify-write via `async with`, **in-process (NOT flock)** with the ADR-0018 deferral of cross-process to the A2A-server model stated inline. `locked_paths(*paths)` dedups + acquires in **sorted order** (the move deadlock-avoidance). The docstring even carries the resolved-path aliasing warning I raised.

## Op-scope completeness (the 重点) — enumerated every branch
| op | lock | correct |
|---|---|---|
| edit | `get_path_lock` across the entire `to_thread` (line ~721) | ✓ full RMW |
| write | `get_path_lock` wrapping read+write (245) | ✓ SAME key as edit → edit/write mutually exclude |
| delete | `get_path_lock` (434) | ✓ |
| regenerate_index | `get_path_lock(resolved_output)` (453) | ✓ locks **output_path**, not source dir |
| move | `locked_paths(src, dst)` sorted (484) | ✓ both, deadlock-safe |
| read / glob / grep / stat / mkdir | none | ✓ correctly unlocked (read-only / dir-create) |

`write_file_bytes` is called only by the write + edit ops, both under their op-level locks → transitively covered. **All ops key on `_resolve_for_gate` = `str(p.resolve())`** (verified) → the same file via relative/absolute/symlink spellings maps to the **same** lock, and edit/write/delete on one path serialize against each other. Matches my earlier enumeration (edit/write/write_bytes/move-dst/delete) + the `regenerate_index(output_path)` addition.

## Guard test — strip-falsified BY ME (not trusting the claim)
- As-is: `tests/test_file_path_lock_2782.py` **3 passed** (concurrent edit + `threading.Barrier` forcing both reads before write = deterministic race → both survive).
- **I stripped the edit lock** → `test_concurrent_edit_file_same_path_both_edits_survive` went **RED** with the exact lost-update signature: `lost update: edit B's change is missing from final content 'aaa line\nBBB line\n'`. Load-bearing, not a mechanism-test.
- **I re-ran my ORIGINAL lost-update repro** (from the investigation) against this branch: `final='L1_XXX\nL2_YYY\n'` → **both survive** (0.32s serialized), vs the pre-fix `main` where it lost an edit (0.16s concurrent). The fix closes the demonstrated race.

## move deadlock avoidance — correct
`move` passes both resolved paths to `locked_paths`, which sorts+dedups before acquiring, so concurrent `move(a→b)` / `move(b→a)` acquire in the same global order — no AB/BA deadlock. Structurally sound.

- **Scope note (not a gap):** the lock covers **file-ops vs file-ops** on a shared path — the demonstrated race and all reachable paths (pipeline `parallel`/`for_each`, concurrent sub-agents, concurrent A2A). A hypothetical *non-file-op* subsystem writing the same path concurrently (e.g. an index writer) is outside #2782's scope and a different, rarer concern; worth a one-line acknowledgement, not a fix here.
- ruff clean; I verified the lock+guard subset + strip-falsify + my repro locally. Recommend confirming the 4 route-mount CI failures = main baseline (as with #2806).

**CLEAR — merge after your verdict-gate.** This closes the data-integrity race I demonstrated, verified end-to-end (design → impl → strip-falsify → original-repro-now-passes).

PR #2812, opened 2026-07-11
feat: P2 AG-UI transport — second ClientTransport over HTTP+SSE (ADR-0039)

tya5:
**[per-PR-coder]** — P2 of the thin-client / single-writer-server arc (ADR-0039), toward #2805.

Adds a **second `ClientTransport` = `AgUiTransport`** over the wire (HTTP + SSE, hand-rolled AG-UI schema — the SDK is not a dep). renderer and stream_client are **unchanged** (P1 proved the seam). P2 = server-side AG-UI emitter + client-side decoder + `STATE_*` status read-model + SSE endpoint behind the P0 auth gate.

Co-vet against `P2_agui_transport_spec.md`. **Do NOT merge** — architect co-vets, lead-coder merges on verdict.

## What landed
- `transport/agui/protocol.py` — reyn `Frame` ⇄ AG-UI-event codec. Standard envelope (generic clients render text/tool/run/error/state) + reyn-private `_reyn` reconstruction block (lossless round-trip). Foreign event with no `_reyn` → `None` (ignore-unknown / graceful degrade, D6).
- `transport/agui/state.py` — `STATE_*` read-model (`project_status` / `StatusModel` server differ / `RemoteStatusView` client reader) + `reguard_nodes` per-connection edge re-guard (A5).
- `transport/agui/emitter.py` — server side: unified `Frame` stream → AG-UI SSE. `MESSAGES_SNAPSHOT` + `STATE_SNAPSHOT` on connect (A4), `STATE_DELTA` on change; WaitingOn tracked off the chat-event stream.
- `transport/agui/client.py` — `AgUiTransport(ClientTransport)`: SSE decode → `Frame` + POST send. Decoupled from HTTP (consumes an SSE-line iterator + a `send` coroutine) so prod is a thin httpx adapter and tests drive the real codec in-memory.
- `transport/agui/endpoint.py` — FastAPI SSE `GET /agui/chat/{agent}/events` + turn-submit `POST /agui/chat/{agent}`, gated by the **P0 auth context** (reused via the same `AuthContext.authenticate` seam — no new auth). Distinct from `ws/chat` (P6). Mounted in `web/server.py`.
- `docs/reference/runtime/agui-transport.md` — the wire protocol reference (event-mapping table + `STATE_*` snapshot/delta contract + present-on-wire + reconnect).

## Scoping decisions flagged for co-vet
- **1 frame ⇄ 1 AG-UI event** (not START/CONTENT/END streaming). reyn's outbox delivers whole messages, not token deltas, so a single content event per message keeps decode unambiguous and the round-trip lossless. The mapping table's role is the generic-client interop surface; fidelity rides `_reyn`. `STEP_*`/WaitingOn semantics are folded into `STATE_*` (the status read-model) rather than emitted as separate step events.
- **`P2_agui_transport_spec.md` was not present in the repo or my broker inbox** — I built to the dispatch brief (which is spec-detailed). Flagging so the co-vet reconciles against the canonical spec.
- **Intervention = DISPLAY only; input = basic turn submit only** (per phase boundary). The answer round-trip methods exist for the ABC but best-effort POST and `pending_intervention_head()` is `None` (HITL answer round-trip = P3).
- **Endpoint session fan-out** is a per-connection `_SessionFrameSource` (server analogue of InProcessTransport off `session.outbox` + filtered `chat_events`). Functional + auth-gated; the heavy N-client production hardening overlaps the ws-consolidation phase.
- **feature-map row deferred** — the feature is not yet actor-reachable (no `reyn chat --connect` wiring until a later phase); flagged rather than adding a premature inventory row.

## Test plan (7 items, Tier 2, real instances, no mocks)
- [x] 1. Bit-identical across transports — AgUi == InProcess == direct baseline, compared by display bytes + WaitingOn sequence (fixed clock). `tests/test_agui_transport_bit_identical.py`
- [x] 2. AG-UI mapping completeness gate — every display kind (AST-scanned from renderer source) + every `renderer_chat_events()` round-trips encode→decode; non-circular (enumerated from renderer vocabulary, not the codec table). `tests/test_agui_mapping_completeness.py`
- [x] 3. present-on-wire inert — presentation rides `CUSTOM`, edge re-guard strips ESC/control from every node leaf. `tests/test_agui_present_on_wire.py`
- [x] 4. `STATE_*` status read-model — snapshot + ≥1 delta → remote `RemoteStatusView` reflects server values (cost + WaitingOn). `tests/test_agui_state_read_model.py`
- [x] 5. Generic-client degrade — event with no `_reyn`, and interleaved foreign events, are skipped; reyn frames still arrive in order. `tests/test_agui_generic_client_degrade.py`
- [x] 6. Reconnect snapshot — first two SSE events are `MESSAGES_SNAPSHOT` then `STATE_SNAPSHOT`; client replays backlog + seeds status. `tests/test_agui_reconnect_snapshot.py`
- [x] 7. Auth reuse — unauthenticated GET/POST refused (401) before attach; correct token passes the gate. `tests/test_agui_auth_reuse.py`

## Strip-falsify RED evidence
Removed the `_reyn` reconstruction block from both codec encoders (cp-backup → edit in place → restore byte-identically; never `git stash`/`checkout`). Both the mapping-completeness gate AND the bit-identical test went **RED** (frames vanish → empty display bytes vs baseline):
```
FAILED tests/test_agui_mapping_completeness.py::test_every_display_kind_round_trips_over_the_wire
FAILED tests/test_agui_mapping_completeness.py::test_every_forwarded_chat_event_round_trips_over_the_wire
FAILED tests/test_agui_transport_bit_identical.py::test_agui_bytes_and_status_match_in_process_and_baseline
3 failed
```
Restored byte-identically (`git diff --stat` clean) → 3 passed.

- `ruff check src tests` — All checks passed (fixed one I001 in a test import block).
- `test_tier_audit.py --strict` (7 test files) — 13 tests, 0 Tier-4 violations.
- `pytest` (full, repo root) — **8145 passed**, 5 failed. All 5 failures are pre-existing route-mount/plugin-loader local-env artifacts (`test_fp0001_routes_mounted`, `test_a2a_routes_mounted`, `test_mcp_routes_mounted`, `test_plugin_loader...`, `test_resources_route_is_mounted_on_app`) — **verified reproducing identically on pristine `origin/main` (8a1b4b6a)**; my change adds zero new failures.
- `verify_module_docstrings.py` (all changed src) — OK.

### DIFF
--- a/docs/reference/runtime/agui-transport.md
+++ b/docs/reference/runtime/agui-transport.md
@@ -1,0 +1,120 @@
+# AG-UI transport — the thin-client wire protocol
+
+Reyn's chat client is a stream-consuming UI: it draws a session's output and
+routes user input, and it touches the session ONLY through a transport seam.
+There are two transports behind that one seam — a local in-process transport,
+and this **AG-UI transport** over HTTP + Server-Sent Events (SSE). Both feed the
+identical renderer, so a remote client draws byte-for-byte what a local one does.
+
+This page is the wire contract: the SSE endpoint, the reyn-frame ⇄ AG-UI-event
+mapping, and the `STATE_*` status read-model.
+
+## Surfaces
+
+The transport speaks **AG-UI only** — it is a UI, not an agent. (Agent↔agent is
+A2A; tools are MCP; observability export is OTEL. Those are separate surfaces.)
+
+- `GET /agui/chat/{agent}/events` — the server→client SSE stream. Each SSE block
+  is `event: <TYPE>\ndata: <json>\n\n`.
+- `POST /agui/chat/{agent}` — the client→server channel. Body is a JSON object;
+  the current message type is `{"type": "user_message", "text": "..."}` (submit
+  a turn).
+
+Both are gated by the server's authentication context: a connection presents its
+token as `?token=` or an `Authorization: Bearer <token>` header (same-machine
+UDS connections are identified by OS peer credentials instead). An
+unauthenticated connection is refused with `401` before any session is attached.
+
+## Standard envelope, reyn-private richness
+
+Every event carries **both**:
+
+- a **standard AG-UI field shape**, so a generic AG-UI client renders the
+  interoperable core (text / tool / run / error / state); and
+- a reyn-private `_reyn` reconstruction block, from which the reyn client rebuilds
+  the exact render frame.
+
+A generic client ignores what it does not understand: an event with no `_reyn`
+block (or a reyn `CUSTOM` event a generic client does not model) is **skipped,
+not fatal** — reyn owns this ignore-unknown contract.
+
+## Event mapping
+
+The client consumes one ordered SSE stream and dispatches each event back to one
+of the renderer's two entry points (display vs working-indicator). The mapping:
+
+### Display path (agent output → the scrollback)
+
+| reyn display kind | AG-UI event        | Notes                                        |
+|-------------------|--------------------|----------------------------------------------|
+| `agent`           | `TEXT_MESSAGE_CONTENT` | the assistant reply text                  |
+| `status`          | `TEXT_MESSAGE_CONTENT` | transient status line (`role: status`)    |
+| `error`           | `RUN_ERROR`        | error text                                   |
+| `trace`           | `CUSTOM`           | reyn tool/step trace line                    |
+| `intervention`    | `CUSTOM`           | a prompt is displayed (answer round-trip is a later phase) |
+| `presentation`    | `CUSTOM`           | a `present` op's render-node model (see *present-on-wire*) |
+| control sentinels | `CUSTOM`           | `__end__` and client-local control kinds     |
+
+Any display kind not in this table still round-trips losslessly (it falls back to
+`CUSTOM` and is reconstructed from `_reyn`) — a new renderer kind can never
+silently vanish on the wire.
+
+### Working-indicator path (turn lifecycle + tool axis)
+
+| reyn chat-event               | AG-UI event      |
+|-------------------------------|------------------|
+| `turn_started`                | `RUN_STARTED`    |
+| `turn_settled` / `turn_completed` / `turn_cancelled` | `RUN_FINISHED` |
+| `tool_called`                 | `TOOL_CALL_START`|
+| `tool_returned` / `tool_failed` | `TOOL_CALL_END`|
+| `user_answered_intervention`  | `CUSTOM`         |
+
+These eight are the exact set the renderer's working / running / waiting-for-you
+indicator consumes; the transport forwards precisely this set.
+
+## present-on-wire
+
+A `present` op's render model is a `list[dict]` of render nodes, **neutralized at
+construction** (every leaf string stripped of terminal control / ESC sequences),
+so it is inert before it reaches any wire. It rides a `CUSTOM` event under the
+`presentation` display kind, carried in `meta.nodes`.
+
+The AG-UI client additionally re-runs the surface neutralizer over every node
+leaf **at the transport edge**, per connection — idempotent for a leaf the
+construction seam already neutralized, but load-bearing defense-in-depth for a
+heterogeneous-surface client whose upstream did not neutralize (or neutralized
+for a different surface).
+
+## STATE_* — the status read-model
+
+The status bar (attached agent, model, cost, tokens, context usage, and the
+current WaitingOn label) is a **read-model**, not a file mirror: it is derived
+from the session's live cost / token / context accessors and the working-indicator
+state, and only the render-relevant subset is streamed.
+
+- `STATE_SNAPSHOT` — emitted **on connect**, the full read-model. Fields:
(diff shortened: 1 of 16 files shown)


tya5:
**[architect]** — co-vet verdict: **CLEAR to merge.** Faithful to the canonical `P2_agui_transport_spec.md`; the load-bearing gates strip-falsify RED (I reproduced it), and the flagged scoping judgment ② is **sound** (verified against the design, not the claim).

## Fidelity to spec
`agui/` pkg (protocol / emitter / endpoint / client / state) over HTTP+SSE, hand-rolled AG-UI schema on the A2A SSE pattern, behind the P0 auth gate, `ws/chat` untouched (P6). Mapping table, present-on-wire (Custom), STATE_* read-model, reconnect snapshots, generic degrade — all present. Renderer + `stream_client` unchanged (P1's seam holds). ✓

## Judgment ② — STEP_*/WaitingOn folded into STATE_* — SOUND (this was the crux)
The concern: does dropping STEP_* and folding WaitingOn into STATE_* break bit-identity or generic-client conformance? **No** — because of the `_reyn`-block design (ADR-0039 D6, standard-envelope + reyn-private payload):
- `encode_frame` sets the OUTER AG-UI type (RUN_*/TOOL_CALL_*, what a generic client renders — deliberately collapsed: `tool_returned`/`tool_failed`→`TOOL_CALL_END`, turn-group→`RUN_FINISHED`) **and** embeds the EXACT reyn frame in `data._reyn` (`type: tool_failed`, etc.).
- `decode_event` reconstructs from `_reyn` → the reyn client recovers the **exact** `EventFrame` → `renderer.on_chat_event` computes the **inline** WaitingOn locally ⇒ **bit-identical**. The `STATE_*` WaitingOn is an *additive* status-panel projection, not the inline driver.
- **Generic conformance holds**: a generic client reads the standard outer type; an event with no `_reyn` decodes to `None` (ignore-unknown, D6). And WaitingOn-as-status → STATE_* is arguably *more correct* than my spec's STEP_* mapping — reyn has no AG-UI "step" concept, so forcing STEP_* would be a mis-map. **The deviation is an improvement, not a regression.**

## Load-bearing gates — strip-falsified BY ME
- **bit-identical across transports**: reyn CUI over `AgUiTransport` == over `InProcessTransport` == direct baseline (display bytes + WaitingOn, fixed clock, full SSE encode→decode round-trip). ✓
- **completeness gate non-circular + fidelity**: enumerates DisplayFrame kinds from the renderer source + the 8 `renderer_chat_events()`, asserting each **round-trips losslessly** (`decode(encode(frame)) == frame`). ✓
- **My independent strip**: dropping the `_reyn` block from both encoders (`data={**std, _REYN: reyn}` → `data={**std}`) → `decode_event` returns `None` → **completeness (both tests) AND bit-identical went RED**. The `_reyn` reconstruction is genuinely load-bearing; both gates catch its removal.

## Remaining focus items — verified non-vacuous
- **STATE_* read-model** (`test_remote_status_view_reflects_snapshot_then_delta`): a mutable server status source (cost rises → `STATE_DELTA`) → the client's remote status view reflects snapshot-then-delta. **Closes the P1 registry-duck-typed deferral I flagged.** ✓
- **present-on-wire (G2) + edge re-guard (A5)**: a `presentation` frame round-trips as a DisplayFrame with nodes intact (rides Custom), AND an ESC/control leaf is neutralized at the edge (decoded node has no raw control bytes) — real neutralizer, no mocks. ✓
- generic-degrade / reconnect-snapshot / P0-auth-refused: pass. 13 tests green, ruff clean.

## Other scoping judgments — all reasonable
- **① 1 frame⇄1 event** (no token-streaming TEXT_MESSAGE) — correct: the reyn outbox delivers whole messages, so one frame = one message; still valid AG-UI.
- **③ intervention DISPLAY-only + basic turn submit** — matches the spec's phase boundary (HITL answer = P3).
- **④ feature-map row deferred** (not actor-reachable until `--connect` wiring) — **correct discipline**: the transport exists but isn't yet invocable from an operator surface, so it's rightly not claimed reachable-for-purpose. This is exactly the "mechanism-exists ≠ reachable" bar; deferring the row is right.

- **Generic-client failure visibility:** `tool_failed` vs `tool_returned` collapses to `TOOL_CALL_END` for a generic client (the failure fact lives only in `_reyn`, which a generic client ignores). Bit-identity for reyn is unaffected (recovered from `_reyn`). If generic-client failure-awareness is ever wanted, surface it in a standard field (e.g. an error flag on `TOOL_CALL_END`). Minor.
- Confirm the 5 route-mount CI failures = pristine main (web/ additive — consistent).

**CLEAR — merge after your verdict-gate.** P2 makes local ≡ remote real over the wire, gated by a non-circular round-trip-fidelity completeness gate and the bit-identical-across-transports assertion; the STATE_* read-model closes the P1 deferral; judgment ②'s fold is sound-and-better via the `_reyn` design. Next: P3 (HITL + fail-close).

PR #2817, opened 2026-07-11
feat: P4 AG-UI generic-client conformance + reyn extension profile (ADR-0039)

tya5:
**[per-PR-coder]** — P4 of the thin-client / single-writer-server arc (ADR-0039), part of #2805. Generic-client conformance + reyn extension-profile formalization. **Do not merge** — awaiting architect co-vet against `P4_conformance_spec.md`, then lead-coder merges on verdict. **Rebased onto current `origin/main` (P3-inclusive) and reconciled** — extends P2+P3, nothing dropped.

## What P4 does
Turns D6's "honest ceiling" into a stated + tested **floor**: a stock AG-UI client (zero reyn knowledge, standard fields only) renders a functional chat, and the `reyn.*` namespace becomes a documented, tested profile. The reyn CUI stays **bit-identical** via `_reyn` (SR5, the standing gate).

### Three conformance gaps closed (additive to the standard surface)
- **(i) TEXT triplet.** `encode_frame_wire` synthesizes the canonical `TEXT_MESSAGE_START → CONTENT → END` with a shared generated `messageId`. Only CONTENT carries `_reyn`; START/END decode to `None` — so the invariant stays **1 frame ⇄ 1 `_reyn`-bearing event** and reyn bit-identity is unchanged (SR1/SR6).
- **(ii) MESSAGES_SNAPSHOT** emits a standard `[{role, content}]` conversation-turns-only array (`agent`→`assistant`, `user`→`user`); chrome stays reyn-private via `_reyn` (SR2/SR2b).
- **(iii) TOOL_CALL_END** carries a standard `status` (`"ok"`/`"error"`) from the etype; reyn exact-recovers from `_reyn`.

### reyn extension profile
New `profile.py` single registry over `reyn.display.<kind>` / `reyn.event.<etype>` (closed CUSTOM members) **and `reyn.intervention.<kind>` (open namespace)**. SR4 non-circular completeness gate: enumerate the reyn-mapped vocabulary from source (renderer kinds + `renderer_chat_events` + the intervention frontend-tool encoder) → assert each emitted name is profiled; unprofiled → RED.

### ignore-unknown, both directions
- **client**: (a) foreign event no `_reyn` → None; (b) unknown `_reyn` frame tag → None; (c) **SR3** unknown `reyn.*` name → skip.
[...]

PR #2818, opened 2026-07-11
feat: OTEL exporter — P6 audit-event → OTLP span/metric/log, fail-open downstream (ADR-0039 P5)

tya5:
**[per-PR-coder]** — P5 of the thin-client / single-writer-server arc (ADR-0039), **part of #2805**. Co-vet against \`P5_otel_exporter_spec.md\`; **do NOT merge** until the architect's verdict — lead-coder merges on verdict.

## What this adds
An **opt-in, fail-open** OpenTelemetry exporter that subscribes to the P6 EventLog and maps each audit-event to OTLP spans/metrics/logs, emitted **off-loop**. It is a **lossy downstream** (observability surface D7, orthogonal to AG-UI) — it never writes to \`.reyn/events\` or the WAL, so the durable recovery/replay Source-of-Truth is unchanged.

- \`src/reyn/observability/otel_exporter.py\` — the single adapter module: pinned GenAI convention version (\`GENAI_CONVENTION_VERSION = 1.37.0\`), the event→OTLP mapping, the fail-open latched swallow, off-loop \`BatchSpanProcessor\`, process-singleton build + \`atexit\` orphan-span flush.
- \`observability.otel.*\` config, threaded through \`SessionFactoryConfig\` → \`build_scoped_chat_session\` → \`Session\`; attached to the session EventLog **only** when an OTLP endpoint is set (\`observability.otel.endpoint\` or \`OTEL_EXPORTER_OTLP_ENDPOINT\` env). No endpoint → not attached, **zero overhead, behavior byte-identical to no OTEL**.
- **SR3 content OFF by default** (\`capture_content\`): refs + token/cost counts only, never a raw prompt/response body.
- Reference doc \`docs/reference/runtime/observability.md\` (+ \`.ja\`), config mirrors (\`reyn-yaml.md\`/\`.ja\`, \`reyn.local.yaml.example\`). \`pyproject\`: the orphaned \`eval\` extra (eval subsystem removed) is renamed to \`observability\` and now carries the OTLP dep + SDK.

## Mapping (OpenTelemetry GenAI semantic conventions)
- \`session_started\`/\`completed\` → root span; \`turn_*\` → turn span (\`gen_ai.operation.name\`); \`llm_called\`+\`llm_response_received\` → \`chat <model>\` child span + \`gen_ai.usage.input/output_tokens\` + cost/token histograms; \`tool_executed\`/\`mcp_*\`/\`web_*\` → \`execute_tool\` child spans; \`permission_*\`/\`intervention_*\`/safety → log records. One correlated trace per run (run_id→agent_id fallback). Robust to out-of-order/missing (SR5b). All \`gen_ai.*\` keys are named constants asserted against the pinned semconv (SR6); cost lives under \`reyn.*\` (not a GenAI key).

## The two locked gates — strip-falsify evidence
- **SR5 fail-open** (\`test_sr5_fail_open_export_error_does_not_break_run\`): a real raising tracer (not a mock) attached to a real EventLog+EventStore. With the swallow present, emitting a full run does **not** raise and every event lands in \`.reyn/events\`. **Strip-falsify (RED evidence):** cp-backup the source, change \`__call__\`'s \`except Exception: self._latch_error(exc)\` → \`raise\`; the \`RuntimeError("otlp endpoint unreachable")\` then propagates through \`EventLog.emit\` → the run breaks → **RED** (\`1 failed\`). Restored → GREEN (\`1 passed\`). Never git-stash; cp-backup/restore.
- **SR4 recovery-independence** (\`test_sr4_recovery_source_unchanged_with_and_without_otel\`, the CLAUDE.md recovery truncate-falsify in **inverted** form): the same fixed-timestamp event stream fans out to two EventStores, one **with** OTEL attached and one **without**; the on-disk recovery JSONL is asserted **byte-for-byte identical**. Then the OTEL exporter is dropped entirely and the source re-derived from the same events → still byte-identical. OTEL contributes **zero bytes** to the recovery source and is never a recovery source. (The assertion is falsifiable: it fails the instant OTEL contributes any byte.)

## Real-integration note (caught + fixed pre-merge)
The in-memory capture tests masked a real SDK ctor-arg bug (\`TracerProvider(schema_url=...)\` is rejected by the installed SDK). Caught by a factory→session reachability check; fixed (\`schema_url\` belongs on \`get_tracer()\`, not the provider). Added \`test_real_otlp_pipeline_builds_with_endpoint_and_sdk\` (importorskip-gated real build) so it can't regress. Verified the real OTLP pipeline builds and stays fail-open against an unreachable endpoint (connection errors are OTEL's own background-thread logs, never propagated to the caller).

## Test plan (all Tier-2, real instances, no mocks — real in-process OTLP capture)
- [x] 1. SR5 fail-open gate (+ strip-falsify RED evidence above)
- [x] 1b. SR4 recovery-OTEL-independence gate (recovery source byte-identical; the locked criterion)
- [x] 2. span-tree correlation (root/turn/child parentage via run_id/agent_id)
- [x] 3. GenAI convention conformance (emitted attrs ∈ pinned version constant)
- [x] 4. metrics (cost/token histogram)
- [x] 5. content-off default (no raw prompt/response in spans/logs unless capture_content=true)
- [x] 6. opt-in / off-by-default (no endpoint → not attached, byte-identical) + env/config opt-in detection
- [x] 7. orphan-span flush (unclosed span on shutdown → flushed + closed)
- [x] (extra) real-build regression guard

## CI gates (local, all green)
- [x] \`ruff check src tests\` — clean
- [x] \`python scripts/test_tier_audit.py --strict tests/observability/test_otel_exporter.py\` — exit 0
- [x] \`pytest\` — \`8209 passed\`; the only failures are **5 pre-existing route-mount/plugin baseline** failures (\`test_fp0001_a2a_endpoints::test_fp0001_routes_mounted\`, \`web/test_a2a::test_a2a_routes_mounted\`, \`web/test_mcp_sse::test_mcp_routes_mounted\`, \`web/test_plugin_loader::test_loader_disambiguates_via_package_field\`, \`web/test_resources_router::test_resources_route_is_mounted_on_app\`), verified to reproduce **identically on untouched origin/main src**. **Zero NEW failures.**
- [x] \`python scripts/verify_module_docstrings.py <changed src>\` — clean

## Carry-forward (P6, out of scope here)
Reasoning* mapping + A2UI catalog + ws consolidation = **P6**. This PR is export-only; OTEL-as-channel/client and any augmentation of the durable store with OTEL are explicitly out of scope (the durable store is untouched).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,8 @@ jobs:
         #   mcp  — `fastmcp` (v3.4.2; #2597 S1), used by tests/test_mcp_client.py
         #   web  — fastapi + uvicorn + websockets, used by tests/web/
         #   fs-watch — `watchdog` (#2608 H4), used by tests/test_fs_watcher.py
-        run: pip install -e ".[dev,mcp,web,fs-watch]"
+        #   observability — opentelemetry SDK, used by tests/observability/
+        run: pip install -e ".[dev,mcp,web,fs-watch,observability]"
 
       - name: Run tests
         # --timeout=120: a per-test wall-clock cap (pytest-timeout). Without it a
--- a/docs/reference/config/reyn-yaml.ja.md
+++ b/docs/reference/config/reyn-yaml.ja.md
@@ -35,6 +35,7 @@ models:
 | `chat` | マップ | チャットセッションの Head/Body/Tail 圧縮設定。以下参照。 |
 | `voice` | マップ | チャット TUI の音声入力（Whisper）設定。以下参照。 |
 | `events` | マップ | チャットセッションイベントファイルの監査ログローテーションポリシー。以下参照。 |
+| `observability` | マップ | P6 監査イベントの OpenTelemetry (OTLP) エクスポート（オプトイン）。デフォルトは無効。以下参照。 |
 | `mcp` | マップ | MCP サーバー定義と `search_threshold`。以下参照。 |
 | `python` | マップ | Python preprocessor の追加許可モジュール。以下参照。 |
 | `agent` | マップ | P6 イベント監査証跡と送信 HTTP ヘッダー用のエージェント識別子。以下参照。 |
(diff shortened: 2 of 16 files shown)


tya5:
**[architect]** — co-vet verdict: **CLEAR to merge.** Faithful to `P5_otel_exporter_spec.md`; both gates (SR5 fail-open + SR4 recovery-independence) hold, and I **reproduced the SR5 strip myself**. One non-blocking test-robustness note.

## Base — fresh, P4-inclusive (incident discipline applied)
`git merge-base` confirms P4 (#2817 `5bc643ef`) is an ancestor; `agui/profile.py` present. Not stale — the fresh-base dispatch worked. ✓

## The non-negotiable constraint — held, both directions
- **SR5 fail-open (strip-falsified BY ME):** the exporter's whole dispatch runs under one `try/except` that swallows + latches (`otel_exporter.py:190-193`). I **removed the swallow** → `test_sr5_fail_open_export_error_does_not_break_run` went **RED** (`RuntimeError: otlp endpoint unreachable` propagates through the emit and breaks the run). The swallow is genuinely load-bearing — an OTLP failure is a no-op for the session only because of it.
- **SR4 recovery-independence (the lead-locked gate):** `test_sr4_recovery_source_unchanged_with_and_without_otel` is non-vacuous — the SAME fixed-timestamp events fan out to an OTEL-attached store chain vs an OTEL-free chain, and it asserts the on-disk JSONL is **byte-identical** (`with_lines == without_lines`, plus a non-empty sanity assert), then re-derives a third replay store with no OTEL and asserts byte-identical again (the inverted truncate-falsify). **Structurally verified:** the exporter never touches `.reyn/state` / WAL / snapshots (grep-clean — it only reads events + writes OTLP spans). Recovery is OTEL-independent by construction.

## Spec conformance (SR1/2/3/6) — verified
- **SR2** off-loop `BatchSpanProcessor` + latched single-warning swallow. **SR1** orphan-span flush on atexit/shutdown (`test_orphan_span_flushed_and_closed_on_shutdown`). **SR3** content-off default — refs/metadata only, no raw prompt/response (`test_content_off_by_default_no_raw_body_in_telemetry`). **SR6** single adapter module, `GENAI_CONVENTION_VERSION = "1.37.0"` pinned, `gen_ai.*` keys asserted against the installed semconv (`test_emitted_span_attrs_only_use_pinned_genai_keys`).
- **Opt-in / off-by-default:** no `observability.otel.endpoint` ⇒ exporter not attached ⇒ zero overhead, behavior unchanged. The SDK is an **optional extra** (`[project.optional-dependencies] observability`) — "reyn core stays SDK-free" is accurate; when the SDK is absent the exporter stays not-attached (build → None), fail-open. Correct packaging.
- **4-surface D7:** export-only, no channel/client; the durable store is untouched.
- 11 observability tests pass (with the extra installed); ruff clean.

- **Test robustness vs the optional extra:** the 4 tests using the `_Capture` helper import `opentelemetry.sdk` directly, so in an env WITHOUT the `observability` extra they **hard-fail** (I saw 4 failures until I installed the SDK) rather than skip. If CI installs `reyn[observability]` (or all extras) they pass — confirm that. Otherwise add a `pytest.importorskip("opentelemetry.sdk")` at the module top so the observability suite **skips** (not fails) when the optional SDK is absent — this keeps the "SDK-free core" property testable (core green without the extra; observability skipped). Not a code defect.
- Confirm the 5 route-mount CI failures = pristine base (consistent with the arc).

**CLEAR — merge after your verdict-gate.** OTEL is strictly additive, opt-in, fail-open, versioned, content-off-by-default, and **provably off the critical path both live (SR5 strip → RED) and on replay (SR4 byte-identical + structurally no recovery-state writes)** — the durable `.reyn/events`+WAL recovery SoT is untouched. Next: P6 (A2UI catalog + ws consolidation + Reasoning\* map).

PR #303, opened 2026-05-20
TUI: clear conv pane + divider on /attach (F13)

tya5:
**[tui-coder]** — UX wave finding F13 (P1/S/score=3.0 from triage). 2 of 8 planned PRs in #291-followup TUI UX wave.

## Observation

After `/attach <other_agent>`, the header label updates to the new agent's name, but the conv pane keeps the previous agent's full transcript visible — no clear, no divider. Two agents' conversations blend into one scroll history, and the user can't tell which earlier turn belonged to which agent.

`src/reyn/chat/tui/app_outbox.py:185-203` `_on_attach_request` only called `header.refresh_status()` + `conv.hide_status()` + transient-timer cancel.

## Fix

After the existing header refresh, call `conv.clear()` (= the same path Ctrl+L uses; resets log + stream rows + skill rows + error boxes + turn anchors, restores empty-state hint). Then write a dim divider line into the now-empty log:

```
── attached to test219 ──
```

`clear()` already re-shows the empty-state hint, so the new agent starts from the same "type / for commands" baseline as a fresh launch. The divider is a permanent breadcrumb for where the switch happened.

## Visual

Before `/attach test219`:
```
   Reyn                  default  │  standard  │  …
─────────────────────
  > previous turn 1
  agent: previous reply 1
  > previous turn 2
  agent: previous reply 2
  ...
```

After (header switches, divider added, empty hint re-shown):
```
   Reyn                  test219  │  standard  │  …
─────────────────────
  ── attached to test219 ──



    Type / for commands  ·  /help for a guide
    ...
```

## Files

| file | change |
|---|---|
| `src/reyn/chat/tui/app_outbox.py` | `_on_attach_request` calls `conv.clear()` + writes dim divider Text |

## Test plan

- [x] Manual: `OPENAI_API_KEY=dummy reyn chat` (default agent) in tmux MCP → `/attach test219` Enter — verified header switches to `test219`, conv pane clears, dim `── attached to test219 ──` divider visible at top, empty-hint re-shows.
- [x] `ruff check src/reyn/chat/tui/app_outbox.py` — clean (pre-push 実走済).
- [x] (skipped — Tier 4 risk) automated test pinning the divider string — `── attached to <name> ──` is UX formatting that should stay free to tweak. Manual visual is the contract per `docs/deep-dives/contributing/testing.ja.md` §6.4.
- [x] (skipped — implicit) regression of Ctrl+L clear behaviour — `clear()` is the same method Ctrl+L invokes, no path change.

[...]

PR #2819, opened 2026-07-11
feat: map reyn reasoning frames to standard AG-UI Reasoning* events (ADR-0039 P6a)

tya5:
**[per-PR-coder]** — **P6a** of the thin-client / single-writer-server arc (ADR-0039), tracker #2805: the **Reasoning\* mapping** sub-phase. P6b (ws-consolidation) and P6c (A2UI watch/defer) are separate follow-ups — this PR is P6a only.

toward #2805

## What
Turns reyn's private reasoning-as-CUSTOM into a **standard AG-UI signal**. A `kind="reasoning"` DisplayFrame now maps to the canonical AG-UI reasoning message lifecycle — `REASONING_MESSAGE_START` → `REASONING_MESSAGE_CONTENT` → `REASONING_MESSAGE_END`, correlated by a shared `messageId` with `role: "reasoning"` — so a generic AG-UI client (CopilotKit) renders reasoning directly. The reyn client stays bit-identical via `_reyn`.

Event names verified against the canonical AG-UI event reference (`https://docs.ag-ui.com/concepts/events`): the content-bearing triplet is `REASONING_MESSAGE_*` (the outer `REASONING_START`/`REASONING_END` wrapper carries no content event). reyn is whole-message, so the 3-event triplet is the minimal valid sequence — no token streaming, no encrypted-CoT variant.

## Design (mirrors the P4 TEXT-triplet discipline)
- **`_reyn` on CONTENT only**: START/END are generic scaffold the reyn client decodes to `None` → invariant stays **1 frame ⇄ 1 `_reyn`-bearing event**. `encode_frame_wire` generalized via a `_CONTENT_TRIPLET` table (text + reasoning share the exact same expansion).
- **SR1 — display-gate by construction (NO new gate)**: reasoning frames only exist when reasoning display is on (the #1652 host chokepoint). Display off → no frame → zero `REASONING_*` events. Proven end-to-end through the REAL `RouterHostAdapter` + REAL emitter.
- **SR2 — profile MOVE, not a new Custom**: reasoning moves from CUSTOM → standard-mapped. The profile-completeness gate now expects reasoning as standard-mapped and requires **no** `reyn.display.reasoning` Custom entry.
- **SR3 — CoT privacy boundary**: reasoning CoT is a display signal only; the P5 OTLP exporter keeps its content-off default and receives no reasoning content (reasoning is a transport-frame concern, never an audit-event).

## Load-bearing invariant — bit-identical preserved
The reyn-client render stays byte-equal pre/post P6a. The P2/P3/P4 bit-identical suites stay green; a dedicated new test drives a reasoning frame through the P6a wire and asserts byte-equality vs the direct-renderer baseline.

## Profile-gate strip-falsify (RED evidence)
Stripping `"reasoning": REASONING_MESSAGE_CONTENT` from `_DISPLAY_KIND_EVENT` (cp-backup restore, never git stash/checkout) → reasoning reverts to CUSTOM → RED:
```
FAILED tests/test_agui_profile_completeness.py::test_reasoning_is_standard_mapped_not_a_custom_profile_entry
FAILED tests/test_agui_reasoning_map_p6a.py::test_reasoning_frame_expands_to_standard_reasoning_triplet
FAILED tests/test_agui_reasoning_map_p6a.py::test_reasoning_is_standard_not_custom
  AssertionError: assert 'CUSTOM' == 'REASONING_MESSAGE_CONTENT'
```
Restored → all green.

## Co-vet round 2 — completeness-gate dict-blind-spot fix + `reyn.display.system` profiled

The architect co-vet cleared the reasoning CORE and surfaced a **concrete pre-existing leak** this PR raised: the completeness gates' `_renderer_display_kinds()` enumerated display kinds from `ast.Compare` branches **only**, so **dict-styled** kinds (looked up via `_KIND_LINE` / `_PREFIX`, never `==`-compared) were invisible. Two kinds are dict-only: `reasoning` (now standard-mapped) and **`system`** — which emits `CUSTOM reyn.display.system` from many producers (`router_loop.py` / `session.py` / `lifecycle_forwarder.py` / inline `app.py` "answered:" echo) but was **not in the extension profile**. So the gate was green while an emitted Custom name was unprofiled — exactly the doc-drift the gate exists to catch.

Fix (both completeness gates — fix-class sibling sweep):
- **Enumerate kinds UNFILTERED from renderer source** — `ast.Compare` branches PLUS the members of every **module/class-level constant collection** (`Dict` / `Set` / `frozenset` — the kind dispatch tables `_KIND_LINE` / `_PREFIX` / `_NESTED_KINDS` and any future sibling). **No `known_kinds` marker subset, no hand-listed dict names** — a new dict-only kind surfaces automatically. Collections **built inside a method body** (the markdown-token → style map) are runtime construction, not kind dispatch, and are excluded, so `heading_open` is correctly NOT treated as a kind (verified).
- **`reyn.display.system` profiled as a CUSTOM member** — reyn chrome (a persisted lifecycle/status marker or the operator "answered:" echo), no standard AG-UI analog → CUSTOM-profiled, **NOT standard-mapped** (consistent with the P4 chrome→CUSTOM disposition). Doc row added to `agui-transport.md`/`.ja`.
- **Strengthened strip** — the gate's sanity now asserts the dict-only `system`/`reasoning` ARE enumerated and `heading_open` is NOT.

Dict-enumeration strip-falsify (both go RED; cp-backup restore, never git stash):
```
# Strip A — remove the reyn.display.system profile entry:
FAILED tests/test_agui_profile_completeness.py::test_every_custom_mapped_frame_is_profiled
  AssertionError: unprofiled reyn.* Custom names (add to profile): ['reyn.display.system']

# Strip B — disable the dict-scan (Compare-only blind spot restored):
FAILED tests/test_agui_profile_completeness.py::test_every_custom_mapped_frame_is_profiled
  Extra items in the left set: 'reyn.display.system'
```
Restored → all green. System probe after fix: `encode_frame(system)` → `CUSTOM reyn.display.system`, `is_profiled=True`. No OTHER dict-only emitted-unprofiled kinds surfaced by the unfiltered gate (only `reasoning`→standard, `system`→now-profiled).

## Co-vet round 3 — authoritative producer-set gate + local-control sentinel filtering

Round-2 still enumerated from `repl/renderer.py` — a **proxy, not the authoritative registry**. Four producer-only control sentinels (`__attach_request__` / `__copy_last_reply__` / `__rewind_list__` / `__session_switch_request__`) are emitted-but-invisible to any renderer scan. **`__attach_request__` was a REAL wire leak**: it is in web's `_FORWARDED_KINDS` (TUI `--connect` attach-label sync, F13 #303) so it reaches remote clients, and on the AG-UI path it was an unprofiled `CUSTOM`.

Root/ideal fix (no-debt):
- **Gate now enumerates the AUTHORITATIVE producer domain** — every `OutboxMessage(kind=...)` literal across `src`: direct constructions PLUS the call sites of **kind-forwarder functions** (a fn that builds `OutboxMessage(kind=<param>)` is a forwarder, discovered **structurally**, not hand-listed — so wrapper-only kinds cannot hide). The renderer proxy is dropped as the source. Every producer kind must be **standard-mapped ∪ profiled ∪ explicit-control-allowlist**, else RED. *(The forwarder-following surfaced 3 more real producer kinds than a direct-constructor scan — the `tool_call_*` trio via `_enqueue_tool_call` — all already profiled; 16 producer kinds total.)*
- **`CONTROL_FILTER_KINDS`** — an **explicit literal allowlist** of the 4 sentinels, deliberately **NOT** the negation of any forward-set (negating the partial `_FORWARDED_KINDS` would wrongly filter `presentation`/`reasoning`/`system`/`user`). The emitter no longer forwards these; `__end__` still terminates; its implicit special-case is promoted to this explicit constant.
- **Per-entry dispositions** (rationale = per-entry facts, NOT a general web-parity rule): `__attach_request__` IS forwarded → **CUSTOM-profiled** (`reyn.display.__attach_request__`) because F13 #303 needs it delivered remotely; the other three are **local-control sentinels with no remote-UI semantics** → filtered.
- **Note (P6b, out of scope here)**: web's `_FORWARDED_KINDS` also omits `presentation`/`reasoning`/`system`/`user`, so it is a partial/legacy set — the web-forward completeness gap (and chat.py's "system already forwarded" comment drift) is P6b ws-consolidation, deliberately untouched.

Producer-set strip-falsify (both RED; cp-backup, never git stash):
```
# Strip A — drop the __attach_request__ profile entry:
FAILED ...::test_every_producer_kind_is_dispositioned — assert not {'__attach_request__'}
# Strip B — remove __copy_last_reply__ from CONTROL_FILTER_KINDS:
FAILED ...::test_every_producer_kind_is_dispositioned — assert not {'__copy_last_reply__'}
```
Restored → green. New emitter tests: local-control sentinels produce ZERO wire events, `__end__` terminates, `__attach_request__` IS forwarded as profiled CUSTOM.

## Co-vet round 4 — two disposition corrections (client-consumed forward; upstream-consumed fail-safe) + dead-entry cleanup

Round-3's gate mechanism (authoritative producer enumeration, explicit control set, strip tests) is accepted; **two dispositions were backwards** (both verified against `origin/main` by primary evidence):

1. **`__copy_last_reply__` / `__rewind_list__` → FORWARD + CUSTOM-profile (was: filter).** Primary: `stream_client._handle_copy_sentinel` does a real **client-side clipboard copy**, and `/rewind` renders a client-side picker — both consumed by the CLIENT over the transport stream. In the thin-client model transport IS the AG-UI wire, so filtering them made remote `/copy` / `/rewind` **silent no-ops** (the exact bug that code fixed). Moved out of `CONTROL_FILTER_KINDS` into `CUSTOM_PROFILE`; the emitter comment ("local UI action, no remote-UI semantics") was false for these two — corrected.
2. **`__attach_request__` rationale corrected (disposition kept: profiled fail-safe).** Primary: its consumer `_on_attach_request` **does not exist** (Textual TUI deleted), and `registry.py:3052` **swallows** it with `continue` — it never reaches the AG-UI tap. So it is **upstream-consumed dead-forward**, not a live wire kind. Kept profiled as a **fail-safe** for a tap-point change; comment fixed to say so (P6b designs remote attach-label sync properly). Same treatment noted for `__session_switch_request__` (upstream-consumed at `registry.py:3061`, stays control-filtered).
3. **Dead-entry cleanup (no-debt):** removed phantom `reyn.display.trace` / `reyn.display.nodes` — an AST scan confirms **0 producer literals** (no producer ever constructs those `OutboxMessage` kinds).

Disposition table now: standard `{agent,error,status,reasoning}`; profiled `{intervention,presentation,system,user,tool_call_*(3),__copy_last_reply__,__rewind_list__,__attach_request__}`; control-filtered `{__end__,__session_switch_request__}` — 16 producer kinds, zero leaks.

Strip **flips to the forward side** (cp-backup, never git stash): dropping `__copy_last_reply__`'s profile entry makes it a **forwarded-but-unprofiled LEAK** → RED (`- profiled / + LEAK`). Restored → green. New **wire-existence probe**: driving `__copy_last_reply__` / `__rewind_list__` through the emitter yields NON-zero AG-UI events (proving forwarded, not filtered); `__end__` / `__session_switch_request__` yield zero; `__end__` terminates.

*(Deferred to P6b, per lead: the closed kind-vocabulary + `OutboxMessage.__post_init__` enforcement — the structural forwarder-discovery gate is accepted as an equivalent static guarantee for this PR. Note: `__post_init__` hard-validation would also need to reconcile with the decode path at `protocol.py:444`, which reconstructs `OutboxMessage(kind=<wire value>)` and must stay lenient for graceful-degrade — a reason it is follow-up, not in-PR.)*

## Test plan (all Tier-2, real instances, NO mocks)
- [x] 1. reasoning frame → standard Reasoning\* triplet (generic surface valid: START/CONTENT/END, shared messageId, role `reasoning`, CONTENT delta = whole text); reyn client bit-identical via `_reyn` (START/END decode None, CONTENT reconstructs the exact frame incl. meta).
- [x] 2. display off → **zero** Reasoning\* events (SR1 by construction, real host adapter + emitter); display-on companion proves the assertion is non-vacuous.
- [x] 3. profile-completeness gate: reasoning = standard-mapped, no Custom entry required; gate passes AND strip-falsify goes RED (evidence above).
- [x] 4. bit-identical preserved (reyn render byte-equal pre/post P6a through the wire; P2/P3/P4 bit-identical suites green) + CoT-not-in-OTEL under the content-off default.

## Gates (all 4 CI jobs green locally)
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 28 inspected, 0 errors
- [x] `pytest` (repo root) — 8223 passed, 21 skipped; **5 failures are pre-existing baseline** (tests/web route-mount/plugin + `test_fp0001_a2a_endpoints::test_fp0001_routes_mounted`), verified to reproduce **identically** on pristine `origin/main` (throwaway worktree). **Zero NEW failures.**

## Docs
`docs/reference/runtime/agui-transport.md` + `.ja` updated: new *reasoning lifecycle* section, display-path table row, coverage matrix **Reasoning 0/7 → 3/7 standard-mapped** (disposition moved future-candidate → standard-mapped), totals 12/28 → 15/28. Display-gate + CoT-not-to-OTEL boundary documented. No history refs.

## For co-vet re-vet (architect, against P6_final_phase_spec.md)
- Triplet subset choice: `REASONING_MESSAGE_*` (content-bearing) vs the outer `REASONING_START/END` wrapper — confirm this matches the spec's intended mapping.
- **Round-2 finding RESOLVED** (superseded by round-3): the dict-blind-spot fix was itself a proxy; round-3 replaces the enumeration source entirely.
- **Round-3 gate mechanism ACCEPTED** (authoritative producer enumeration via structural forwarder-discovery — 16 kinds incl the `tool_call_*` trio; explicit control set; strip tests).
- **Round-4 finding RESOLVED**: the two backwards dispositions are flipped — `__copy_last_reply__`/`__rewind_list__` are FORWARDED + profiled (client-consumed over the transport stream; filtering broke remote `/copy` /`/rewind`), and `__attach_request__` is documented as upstream-consumed fail-safe (its consumer is deleted; `registry.py:3052` swallows it). Dead `trace`/`nodes` profile entries removed (0 producer literals). Strip flips to the forward side (RED reproduced). Independent re-vet targets: (a) probe copy/rewind produce NON-zero emitter events; (b) `__attach_request__` upstream-consumed comment + profiled fail-safe; (c) `trace`/`nodes` removed. Reasoning core (bit-identical / SR1 / SR3) stays green. Not claiming "sound" — leaving that verdict to the re-vet.


tya5:
**[architect]** — co-vet verdict: **FIX-BEFORE-MERGE** — narrow. The **reasoning-mapping core is correct** (verified + SR2 strip-falsified), but the coder's flagged **② is NOT fully sound**: the completeness gate misses dict-only kinds, and I found a concrete live leak — **`system` → `reyn.display.system` is emitted, CUSTOM, and UNPROFILED while the gate stays green.**

## Reasoning-mapping core — CORRECT (② aside)
- **① triplet subset — accepted.** `kind="reasoning"` → the content-bearing `REASONING_MESSAGE_START/CONTENT/END` triplet (shared `_CONTENT_TRIPLET` with text, `messageId`-correlated). This is the right subset — the outer `REASONING_START/END` is a context wrapper and `CHUNK/ENCRYPTED` are streaming variants reyn (whole-message) doesn't need. (Worth one confirmation that the canonical doesn't *require* the outer wrapper, as TEXT requires its triplet — the coder says canonical-verified.)
- **Bit-identical + SR2 load-bearing (strip-falsified BY ME):** reasoning encodes to `REASONING_MESSAGE_CONTENT` (standard, not CUSTOM), `_reyn` on CONTENT only. I **stripped the reasoning standard-map entry** → it reverts to CUSTOM → 4 tests RED (`test_display_on_yields_reasoning_triplet…`, `…bit_identical_through_wire`, `…content_carries_reyn_start_end…`, `test_reasoning_is_standard_mapped_not_a_custom_profile_entry`). The mapping + profile-move assertion are genuinely load-bearing.
- **SR1 display-gate** respected by construction (frame only emitted when reasoning-display on → no frame → no event). **SR3** reasoning content stays off the OTEL export (P5 content-off). Docs Reasoning 0/7→3/7. ✓

## ② — the profile-gate is NOT sound; dict-styled signals leak (the finding)
The gate's `_renderer_display_kinds()` AST-scans **only `ast.Compare` nodes** in `message`/`format_inline_message`. But renderer display kinds live in BOTH Compare branches AND dicts (`_KIND_LINE`/`_PREFIX`). I enumerated: **two kinds are dict-only (AST-scan-invisible): `reasoning` and `system`.**
- `reasoning` is now standard-mapped, so the explicit positive assertion covers it — fine.
- **`system` is the live leak:** it's dict-only, NOT in the codec's standard map → falls back to **CUSTOM `reyn.display.system`**, is **emitted by many producers** (`router_loop.py:2398`, `session.py:5629/5641`, `lifecycle_forwarder.py:245`, `inline/app.py:1512` — the "answered:" line), and is **NOT in the profile** (`is_profiled("reyn.display.system") == False`). **The completeness gate is GREEN while an emitted Custom name is unprofiled** — exactly the doc-drift the gate exists to prevent. I confirmed via a real `encode_frame` probe: `system: type=CUSTOM, profiled=False`.

So the honest answer to ②: **yes, dict-styled signals leak.** The explicit-assertion patch closes reasoning but not the hole; `system` proves the hole is real (pre-existing, not a P6a regression — but P6a is the PR that raised the gate's dict-blind-spot).

## Recommended fix (small, in-PR)
Make the gate enumerate **dict-only kinds too** — scan the renderer's dict-key string literals in the display-dispatch path (not just `Compare` nodes), OR enumerate the display-kind vocabulary from the codec's emitted set. That will flag `reyn.display.system` as unprofiled → close it by either **adding a `CustomName("reyn.display.system", …)` profile entry** or **standard-mapping `system`** (e.g. to a system/notification event). This makes the completeness gate genuinely non-circular AND complete (no dict-styled leak), which is the whole point of ②. Add a strip test: an unprofiled emitted Custom (incl. a dict-only kind) ⇒ RED.

## Gate strength (explicit)
- **Reasoning-mapping core (①, bit-identical, SR1/SR2/SR3): CLEAR** — could ship as-is.
- **② gate-soundness + `system` leak: FIX-BEFORE-MERGE** — the coder raised ② for validation and the finding is a live gate blind-spot, and P6a is the gate-touching PR. (If you'd rather scope P6a to reasoning-only, the pre-existing `system` gap is a defensible fast-follow — but then the gate should NOT be described as sound, and `system` needs a tracked ticket. My recommendation is in-PR: the fix is small and closes the concern this PR itself surfaced.)

Happy to re-vet the gate fix (I'll re-run the `system` probe + the dict-enumeration strip).

tya5:
**[architect]** — P6a round-2 co-vet re-vet: **FIX-BEFORE-MERGE (round-3). Completeness gate NOT sound — do not verdict-gate merge.** Reasoning-mapping core remains CLEAR (bit-identical / SR1 / SR3 unaffected). This blocks on gate soundness only.

### Falsification (primary data, independent probe on `13425166`)
The round-2 completeness gate passes **green (3 passed)** while **four producer-emitted kinds encode as unprofiled CUSTOM**:

| kind | encode_frame type | is_profiled | producer |
|---|---|---|---|
| `__attach_request__` | CUSTOM | **False** | slash/agent.py:92, slash/agents.py:129 |
| `__copy_last_reply__` | CUSTOM | **False** | slash/copy.py:35 |
| `__rewind_list__` | CUSTOM | **False** | slash/rewind.py:46 |
| `__session_switch_request__` | CUSTOM | **False** | slash/session.py:88 |

So the round-2 claim "no OTHER dict-only emitted-unprofiled kinds — only reasoning + system" is **falsified**.

### Root cause — the enumeration source is a proxy, not the registry
The fixed gate enumerates `repl/renderer.py` dict/Compare literals "unfiltered" — but **the renderer file is the wrong registry.** These four kinds do not live in the renderer at all; they are consumed at `registry.py:3052/3061` (server) and `stream_client.py:221/225` (client). No amount of unfiltered renderer-dict scanning can surface a producer-only kind. The authoritative codec-input domain is the **producer `OutboxMessage(kind=...)` set** (13 kinds, measured across src). This is exactly the coverage-migration discipline — derive from the source-of-truth registry, not a proxy subset — applied to the *wrong file*.

### `__attach_request__` is wire-bound, not local control
`interfaces/web/ws/chat.py:55-62` — `_FORWARDED_KINDS` frozenset **explicitly includes `__attach_request__`** (TUI `--connect` attach-label sync, F13 #303). The existing web path forwards it to remote clients, so on the AG-UI path an unprofiled CUSTOM for it is a genuine wire leak. The "`__`-prefixed = internal, out of profile scope" defense fails here. There is **no general dunder filter** — `__end__` is the *only* special-cased kind (emitter.py:114).

### Fix (root/ideal, per no-debt directive)
1. Gate enumerates the **producer `OutboxMessage(kind=...)` set**; assert each kind ∈ {standard-mapped ∪ profiled ∪ **documented control-filter allowlist**}, RED otherwise. Retire the renderer-dict proxy.
2. Promote the `__end__` special-case into an **explicit, documented control-kind allowlist** (implicit invariant → explicit + CI-enforced).
3. Per-kind disposition (**design decision — owner/lead to ratify, not architect-unilateral**): `__attach_request__` is wire-forwarded → CUSTOM-profile (same class as `system`); the other three are absent from web `_FORWARDED_KINDS` (web drops them) → lean control-filter allowlist, but this overlaps P6b ws-consolidation (AG-UI over-forwards vs the web allowlist — filters nothing but `__end__`).
4. Strip proof: after (1)-(3) land, removing any one kind's disposition → gate RED. Only then may the gate be described as "sound".

**Scope alternative:** land the reasoning map, honestly scope the gate as *renderer-display-completeness*, track the four `__...__` kinds as a known-gap follow-up. Acceptable **only if** the gate is not described as "sound/authoritative/covers all dict-only kinds" (that would be an overclaim). Under the no-debt directive I recommend the ideal fix (1)-(4).

tya5:
**[architect]** — Round-3 direction AMENDMENT (self-review of my own round-2/3 recommendations; two of my claims corrected before implementation lands). **BLOCK-strength on one point: `__copy_last_reply__` / `__rewind_list__` must NOT go into the filter allowlist.**

### Correction 1 — `__copy_last_reply__` / `__rewind_list__`: FORWARD + profile, not filter (my "lean filter" was consumer-side-unverified and wrong)
`stream_client.py:221/225` — the **client-side** handlers consume these two sentinels **from the transport frame stream**: `/copy` resolves against the client's `recent_replies` and copies to the **client machine's clipboard** (a server-side clipboard is meaningless); `/rewind` renders the picker list on the plain-CUI path. In the thin-client model the transport is AG-UI — filtering these breaks remote `/copy` and `/rewind` into silent no-ops (the handler's own comment records exactly this historical bug: "no more silent no-op"). The registry forwarder does not intercept them (they reach `repl_outbox` → transport when attached). Disposition: **CUSTOM-profile as closed members**, `_reyn` payload for lossless reyn-client round-trip; generic clients ignore-unknown.

### Correction 2 — `__attach_request__`: "wire-bound, confirmed" was grounded in dead code; actual disposition = upstream-consumed
`registry.py:3052/3061` — the forwarder **handles attach/switch server-side and swallows them (`continue`)**; they never reach the transport (the AG-UI tap follows the registry focus-listener). The web `_FORWARDED_KINDS` attach entry I cited forwards to a consumer that **no longer exists** (`_on_attach_request` — zero hits; the registry comment itself says "TUI deleted — re-post is dead code"). I retract that basis. Disposition: `__attach_request__` / `__session_switch_request__` = **allowlist entries documented as "upstream-consumed at registry.py:3052/3061"** (the codec-level entry stays as a fail-safe against tap-point variance, not as a de-facto forward). Web's dead forward is P6b cleanup; remote attach-label-sync UX is a P6b explicit design.

### Correction 3 — my "producer set = 13 kinds, measured" was an undercount; the grep-gate shares the proxy failure class
Literal-grep anchored on `OutboxMessage(` misses **helper-mediated dynamic constructions**: `lifecycle_forwarder.py:467`, `router_host_adapter.py:1412`, `slash/__init__.py:242` all build `OutboxMessage(kind=kind)` — `tool_call_started/completed/failed` are produced via the first (call sites :268/:281/:297) and were absent from my set. Additionally `trace`/`nodes` sit in `CUSTOM_PROFILE` with **no producer findable by any static scan** (dead entries or dynamic — statically undecidable).

### Root fix (the only sound enumeration source): closed kind vocabulary
Add to `runtime/outbox.py`: `DISPLAY_KINDS: frozenset[str]` + `CONTROL_KINDS: frozenset[str]` (sentinels, each documented with its consumption locus), and **validate `kind ∈ vocabulary` in `OutboxMessage.__post_init__`** — construction-time, fail-visible, so helper/dynamic paths are enforced at runtime (impossible for any static scan). The gate then derives from the vocabulary: (i) every kind ∈ {standard-mapped ∪ profiled ∪ CONTROL_KINDS}; (ii) **reverse direction**: every entry of the codec map / `CUSTOM_PROFILE` / web `_FORWARDED_KINDS` ⊆ vocabulary (kills dead entries — settles trace/nodes). ~30 lines + gate rewrite; eliminates the renderer-proxy, grep-proxy, helper-blind-spot, dead-entry, and future-drift classes at once. Recommended in-PR (the gate rework is already in-PR and cannot honestly be called "sound" without a sound source).

### Corrected disposition table
| disposition | kinds |
|---|---|
| standard-mapped | agent, error, status, reasoning |
| profiled CUSTOM | intervention, presentation, system, user, tool_call_started/completed/failed, **__copy_last_reply__, __rewind_list__** |
| CONTROL_KINDS (allowlist, documented locus) | `__end__` (transport-terminal, emitter.py:114), `__attach_request__` / `__session_switch_request__` (upstream-consumed, registry.py:3052/3061) |
| resolve via vocabulary | trace, nodes (no producer found — dead or dynamic; delete from profile if dead) |

### P6b SR additions (recorded with lead)
(i) web `_FORWARDED_KINDS` dead attach-forward cleanup; (ii) remote inline `/rewind` picker rides `session.pending_command_ui`, not the sentinel — verify it crosses the remote state channel; (iii) web chat.py taps `agent.outbox` directly (a different point than the registry forwarder) — tap-point unification audit.

tya5:
**[architect]** — Round-4 (28ecfc41) co-vet: **CLEAR ✅.** All findings from rounds 2-4 closed and independently reproduced (base verified P5-inclusive; strips via cp-backup, never git stash).

### Independent verification
- **(a) Wire-existence (real emitter driven):** `__copy_last_reply__` / `__rewind_list__` emit non-zero AG-UI events (forwarding restored); `__session_switch_request__` emits zero (filtered); `__end__` terminates the stream. The remote `/copy` / `/rewind` silent-no-op regression is closed.
- **(b) Strip flip (clean):** removing the `__copy_last_reply__` `CustomName` block → 3 tests RED (`test_every_producer_kind_is_dispositioned`, `test_control_sentinel_dispositions_*`, `test_every_custom_mapped_frame_is_profiled`) with clean AssertionErrors (not collection errors) → restore green. The gate genuinely catches a forward-side profile gap.
- **(c) Authoritative producer-set leak scan** (direct literals + `_enqueue_tool_call` helper-mediated) = **16 kinds, zero leaks** — resolves my own round-3 undercount (13, missing the `tool_call_*` trio); matches the coder's structural forwarder-discovery independently.
- **(d) Bucket invariant:** 16 = standard 4 {agent, error, status, reasoning} / profiled 10 / control 2 {`__end__`, `__session_switch_request__`}; **sum = 16, zero multi-or-unbucketed kinds** (clean partition).
- **(e)** `reyn.display.trace` / `reyn.display.nodes` removed (phantom entries, 0 producers) — confirmed de-profiled.
- **(f)** Reasoning core + full agui suite: **107 passed / 3 skipped**; bit-identical / SR1 / SR3 unchanged.
- **(g)** `__attach_request__` / `__session_switch_request__` comments now accurately state upstream-consumed at registry.py:3052/3061 + fail-safe rationale. The new `test_control_sentinel_dispositions_client_consumed_forward_upstream_consumed_filter` **CI-enforces the consumer-locus rule** (client-consumed → forward, upstream-consumed → filter) with an encode→decode lossless round-trip (not a tautology). The disposition principle is now a regression guard.

### Design constraint for the deferred follow-up (endorsing the coder's flag)
The deferred closed-vocabulary + `OutboxMessage.__post_init__` validation must scope enforcement to **construction/production only** — the decode path (protocol.py:444, which reconstructs `OutboxMessage` from wire values) must stay lenient for graceful-degrade. A strict validator on the decode path would fail-close on an unknown wire kind and break the ignore-unknown contract. Please note this constraint on the follow-up issue.

### Arc status
**ADR-0039 P6a is complete pending merge.** The gate is now sound in the sense the earlier rounds could not claim: it enumerates the authoritative producer domain (literal + helper-mediated), partitions all 16 kinds disjointly, and CI-enforces both completeness and the per-entry consumer-locus dispositions. Next: P6b (ws-consolidation) co-vet on dispatch — SR4 (enumerate chat.py consumers before delete) + SR5 (browser P0 token) + the three items I surfaced here (dead attach-forward cleanup / remote `/rewind` `pending_command_ui` path verification / `agent.outbox`-direct-tap vs registry-forwarder tap-point unification).

PR #2825, opened 2026-07-11
security: bound the version-file read in argv0 resolution (parent-memory DoS) (part of #2820)

tya5:
**[tui-coder]** — part of #2820 (part A DoS hardening, forward-fix). Follows merged #2822 (part A).

## What
A small, high-priority defensive follow-up to part A (#2822). The adversarial security co-vet CLEARed the merged commit against RCE / traversal / symlink-escape / exec, but a **broker-lag race** merged it ~45s before my self-audit hardening landed, so this did not ride in #2822. Forward-fixing it here (no revert — part A is a net RCE-surface improvement).

## Why (the gap the co-vet's own focus surfaced)
`resolve_real_executable` reads a `.python-version` from the sandbox child's **cwd**, which is **agent/tool-writable**. The merged code read it with an unbounded `read_text()`. An attacker who plants a multi-GB regular file there (or a symlink to one) turns resolution into a **memory DoS of the reyn parent process** (resolution runs in the parent). Low severity — not RCE — but a real, reachable resource-exhaustion vector on the incident pyenv host.

## Change
- **bounded version-file read**: read only a 4 KB head (a version token is a handful of bytes and lives at the very start). A huge planted file can no longer be read into memory. `is_file()` already excludes devices/FIFOs, so a symlink to `/dev/zero` is skipped, not read endlessly.
- **resolved-bin symlink containment** (the guard already exists in merged code) now has an explicit escape-fails-open test.
- **version-token validation across all sources** (env / file / global) — already the case since validation is applied to `_selected_version`'s return regardless of origin — now pinned by an env-source malicious-token test.

## Test plan
- [x] `pytest tests/test_sandbox_argv0_resolve_2820.py` — 21 Tier 2 (4 new: bounded-read with 8 MB trailing junk still resolves via the head token; device-symlink skipped; env-source malicious token rejected; resolved-bin symlink-escape fails open)
- [x] `test_tier_audit.py --strict` — 21 OK, 0 Tier-4
- [x] full `pytest` — 8259 passed; the same 9 pre-existing LLM-router/compaction failures (CI-green, local-only pollution, pass in isolation). No new failures from this diff.

Tier self-check: Tier 2, assert on public surface (return value, sentinel/side-effect absence), real filesystem layout (no mock), no private-state / format-pin.

Subprocess-zero is preserved (this only bounds the filesystem read). Ready for the same adversarial security co-vet with unbounded-read DoS added to the explicit focus.

### DIFF
--- a/src/reyn/security/sandbox/resolve.py
+++ b/src/reyn/security/sandbox/resolve.py
@@ -76,6 +76,13 @@ def _resolvable_manager_root(path: str) -> str | None:
     return None
 
 
+# A version file is a handful of bytes. A ``.python-version`` sits in the
+# agent/tool-writable cwd, so cap the read: an attacker who plants a multi-GB
+# regular file (or a symlink to one) must not turn resolution into a memory DoS.
+# The version token lives at the very start, so the head is all we ever need.
+_VERSION_FILE_READ_CAP = 4096
+
+
 def _first_token(text: str) -> str | None:
     """First whitespace/colon-free token of the first non-empty line of *text*,
     or None. (``PYENV_VERSION`` uses ``:`` to list multiple; files use lines.)"""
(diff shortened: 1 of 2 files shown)


tya5:
**[tui-coder]** — test falsifiability fix pushed (commit `6bc14f25`, test-only)

Good catch — the original `test_oversize_version_file_is_read_bounded` was **non-falsifying**: `"3.12.7\n" + 8 MB` doesn't OOM, and the old unbounded `read_text()` returns the same head token `3.12.7`, so a regression to `read_text()` would stay green (wiring-test anti-pattern).

Replaced with the reviewer-provided falsifying construction: `.python-version = (" " * 5000) + "3.12.7\n"`, assert **fail-open to the shim**.
- bounded read (4 KB cap): the head is all whitespace → no token → fail open to shim → `assert resolved.endswith("/.pyenv/shims/python3")` holds.
- unbounded read (old / regression): the full line strips back to `3.12.7` → resolves to the real binary → the assertion FLIPS → RED.

**Verified locally**: GREEN with the cap; temporarily setting `_VERSION_FILE_READ_CAP = 100_000_000` (i.e. effectively `read_text()`) makes it **RED** — so it genuinely witnesses the byte bound, not just a happy path.

Implementation untouched (test-only). Ready to merge.

PR #2838, opened 2026-07-11
feat: render the inline CUI on `reyn chat --connect` (ADR-0039 P3 — local ≡ remote at the renderer/loop layer)

tya5:
**[tui-coder]** — Render the inline CUI on `reyn chat --connect`: local ≡ remote at the renderer/loop layer, not just the transport (ADR-0039 P3). Part of the ADR-0039 thin-client arc (#2805). **DO NOT merge until reviewed; @architect please co-vet the design seam + vocabulary sufficiency.**

## The gap (owner dogfooded the thin client)

`reyn chat --connect <url>` on a TTY rendered the **plain console CUI** while local `reyn chat` rendered the **Rich inline CUI**. Transport/frame parity was done (P1/P2), but the **renderer selection** and the **inline input driver** were never mirrored to the remote path — an incomplete realization of "local ≡ remote by construction."

### Flow-trace (before)
- **Local** `chat.py::run` → picks `make_inline_renderer()` if `_inline_interactive` else console → `repl.run_repl` builds `InProcessTransport`, banners, selects `run_inline_input(registry,…)` (inline app, reads status off the **registry**) vs `run_input_loop` (PromptSession), runs the shared `run_output_loop`, waits, tears down + cost summary.
- **Remote** `chat.py::_run_remote` → `make_chat_renderer()` **unconditionally** → `remote_client.run_remote_repl` builds `AgUiTransport`, banners, always `run_input_loop` + `run_output_loop`. The inline branch was simply absent.
- `ClientTransport` (in-process + AG-UI) and the `stream_client` loops were **already** transport-agnostic. The only local-only couplings were (a) renderer selection, (b) the inline app reading its status bar / intervention region / task poll straight off the `registry`/`Session` (the app's own docstring named it: *"a full client-side read-model is P3"*).

### The unified seam (introduced)
- **`ChatReadModel`** (`repl/read_model.py`) — the client's READ half (writes already ride `ClientTransport`). Two impls: `RegistryReadModel` (local, byte-identical delegation to `_snapshot`/session) and `RemoteReadModel` (projects the P2 `RemoteStatusView` — fed by the server's `STATE_*` — into the `_snapshot` dict shape). The inline app now reads `snapshot()` / `intervention_head()` / `pending_command_ui()` / `list_active_tasks()` / `history_path` through this seam.
[...]

PR #2840, opened 2026-07-11
fix: retune AG-UI thin-client heartbeat load (10s->25s + piggyback, 60s server timeout)

tya5:
**[tui-coder]** — Reduces the AG-UI thin-client (`reyn chat --connect`) heartbeat server load, per the owner-raised concern, grounded in prior-art research (Socket.IO 25s, Phoenix 30s, SignalR 15s+2×timeout — the idiomatic cluster is 15–30s with timeout ≈ 2× interval; our prior 10s interval was aggressive against that).

## Background (verified, confirmed against #2838-era code)

`src/reyn/interfaces/repl/remote_client.py` sent `POST {"type":"heartbeat"}` every 10s per connection. Server-side, a NORMAL disconnect is already caught immediately via the SSE handler's `finally: manager.detach(...)` — the heartbeat POST is a **half-open backstop only** (client hung / no TCP FIN), so a 10s cadence buys nothing on the common path and is pure server load.

## The three changes

1. **Client interval 10s → 25s + piggyback** (`remote_client.py`)
   - `_HEARTBEAT_INTERVAL = 25.0` (named constant, `REYN_AGUI_HEARTBEAT_INTERVAL_S` env override).
   - Piggyback: extracted the decision into a pure function `_heartbeat_due(last_send, now, interval)`. `send()` now stamps `last_send` (monotonic) on every client→server POST (turn/answer/cancel/heartbeat); the heartbeat loop's tick skips the dedicated ping when a real POST already refreshed liveness within the window.

2. **Server `liveness_timeout` 45s → 60s** (`surface.py`, `DEFAULT_LIVENESS_TIMEOUT`)
   - `REYN_AGUI_LIVENESS_TIMEOUT_S` env override. 60s keeps a 2.4x margin over the new 25s client interval (in line with Socket.IO 25s/60s, Phoenix, SignalR ratios) — a live, idle client is never false-swept. `DEFAULT_GRACE_SECONDS` (20s) is unchanged.

3. **Heartbeat stat-avoidance → pure in-memory** (`endpoint.py`, `agui_submit`)
   - The heartbeat branch (`manager.heartbeat` refresh + early-return) now runs **before** `registry.exists()` (a `Path.is_file()` stat on the agent's profile). Order is now: auth gate (401) → parse payload → **heartbeat fast-path (in-memory, return)** → `registry.exists()` (404) → liveness refresh for non-heartbeat traffic (piggyback) → the rest of dispatch. The auth gate is untouched — only the disk stat moved past the heartbeat branch.

## Numbers + config decision

Final: **interval = 25s, server timeout = 60s, grace = 20s (unchanged)**. `25 < 60 < 80` — the architect's ordering gate holds.

**Constants + env-var override, not full `WebConfig` threading.** There is no existing `WebConfig`/`_build_*` seam for per-agent `SurfaceManager` config (it's a bare module-global `SurfaceRegistry()`, not constructed from `reyn.yaml` at all) — wiring it through would mean inventing a new config-to-runtime path for a single deployment knob. A named constant + `REYN_AGUI_HEARTBEAT_INTERVAL_S` / `REYN_AGUI_LIVENESS_TIMEOUT_S` env override is the proportionate mechanism here, consistent with the `REYN_WEB_*` env-override pattern already used elsewhere (`web/deps.py`, `web_config.py`).

## LOAD-BEARING INVARIANT — verified + strip-tested

The architect's acceptance gate: after the retune, the grace window T must still cover half-open detection latency, and the "last-surface-gone → typed DENY" behavior must not loosen.

- `tests/test_agui_heartbeat_retune.py::test_client_interval_below_server_timeout_below_timeout_plus_grace` pins `interval < timeout < timeout+grace` on the actual shipped constants.
- `test_half_open_client_swept_then_typed_denied_within_grade` (THE gate) — using the **production** `DEFAULT_LIVENESS_TIMEOUT`/`DEFAULT_GRACE_SECONDS`, a half-open client (real `Session` + real `SurfaceManager`, no mocks) is: not yet stale just under the timeout → swept via `sweep_dead` once past it → `should_fail_close` False until grace elapses → True after → `session.fail_close_interventions` resolves the real pending intervention with a typed refusal (`answer.refused is True`, `reason == NO_SURFACE_REFUSAL_REASON`) — `asyncio.wait_for(task, timeout=2.0)` would time out (RED) if it parked instead of resolving.
- Two "bad retune" demonstration tests prove the numbers matter, not just assert them: `test_bad_retune_with_zero_margin_false_positives_a_live_client` constructs a deliberately-misconfigured `SurfaceManager(liveness_timeout=_HEARTBEAT_INTERVAL)` (timeout == interval, zero margin) and shows a live, on-cadence client IS false-swept by the slightest jitter — contrasted by `test_production_timeout_absorbs_realistic_heartbeat_jitter`, which shows the same jitter is comfortably absorbed at the real 60s/25s ratio.
- Falsified by hand: stashing all src changes makes every new test fail to import / fail (confirmed); reverting only `endpoint.py`'s reorder (keeping the retuned constants) makes `test_heartbeat_on_attached_connection_skips_registry_exists_stat` go RED (`assert 1 == 0`) — proving that test actually discriminates the fast-path fix, not just the constants.

## Tests (Tier 2, real instances, no MagicMock/AsyncMock/patch)

New file `tests/test_agui_heartbeat_retune.py` (7 tests):
- (a) piggyback: `_heartbeat_due` pure-function contract — a real send within the window suppresses the tick; a full idle interval fires it.
- (b) fast-path: a heartbeat POST on an already-attached connection (real `SurfaceManager`, real FastAPI `TestClient`) makes **zero** `registry.exists()` calls (a hand-written counting registry Fake — the endpoint's touch-point pattern, same style as `test_agui_ignore_unknown_both_directions.py`'s `_FakeRegistry`), proven by contrast against an un-heartbeated peer connection that IS swept.
- (c)/(d): the ordering invariant + the half-open→sweep→DENY chain + the two bad-retune/good-retune contrast tests, all above.

## CI gates (run locally, repo root)

- `ruff check src tests` → **All checks passed** (fixed one `I001` import-sort on the new test file).
- `python scripts/test_tier_audit.py --strict tests/test_agui_heartbeat_retune.py` → **7/7 OK, 0 errors** (fixed two docstrings to the exact `"""Tier 2: ...` prefix the auditor requires).
- `pytest` (repo root) → **8286 passed, 21 skipped, 5 failed**. All 5 failures are the known pre-existing editable-install route-mount baseline failures (`test_fp0001_a2a_endpoints.py::test_fp0001_routes_mounted`, `web/test_a2a.py::test_a2a_routes_mounted`, `web/test_mcp_sse.py::test_mcp_routes_mounted`, `web/test_plugin_loader.py::test_loader_disambiguates_via_package_field`, `web/test_resources_router.py::test_resources_route_is_mounted_on_app`) — unrelated to this PR's 3 changed src files (confirmed via `git diff --stat origin/main -- src/`, which touches only `remote_client.py` / `endpoint.py` / `surface.py`). **Zero new failures.**
- `python scripts/verify_module_docstrings.py <changed src files>` → **OK: no module-docstring refactor narrative.**

## Docs

Constants-only (no `reyn.yaml` config surface added), so per the brief a short note suffices: added a paragraph to `docs/reference/runtime/agui-transport.md` + the `.ja.md` mirror under "Fail-close and the grace window" explaining the heartbeat is a half-open backstop (SSE-close is primary liveness), the piggyback behavior, the retuned numbers, and the two env-var override names. No `feature-map.md` change (not a new user-facing config key).

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_agui_heartbeat_retune.py` — 7/7 OK
- [x] `pytest` (repo root) — 8286 passed, 5 pre-existing-unrelated failures, 0 new
- [x] `python scripts/verify_module_docstrings.py <changed files>` — OK
- [x] Falsify: strip-tested the fast-path fix (endpoint.py reorder reverted → test goes RED) and the bad-retune numbers (misconfigured timeout → false-positive sweep demonstrated)

**DO NOT merge until reviewed** — @architect please co-vet the last-surface-gone→DENY invariant against the retuned numbers (25s/60s/20s). Broker-notified lead-coder + architect separately.


tya5:
**[architect]** — #2840 heartbeat retune invariant co-vet: **CLEAR ✅.** The retuned numbers preserve the load-bearing last-surface-gone→typed-DENY invariant; all 5 points verified, the DENY-strip reproduced by me.

### #1 — last-surface-gone → typed-DENY (load-bearing, strip-reproduced)
Stripping the real DENY (`deny_unanswerable_active`'s `future.set_result(InterventionAnswer(refused=True))` in `intervention_registry.py`) → `test_half_open_client_swept_then_typed_denied_within_grace` goes RED (the intervention parks instead of being denied); restored → 7 passed. So the retune preserves the DENY firing. Real `Session` + `SurfaceManager`, no mocks. *(First strip hit a sibling `set_result` in session.py that isn't the DENY; the real DENY path is `deny_unanswerable_active` — falsify the actual path.)*

### #2/#3 — retuned numbers + ordering
`DEFAULT_GRACE_SECONDS = 20` (surface.py:58), `DEFAULT_LIVENESS_TIMEOUT = 60` (:59), heartbeat interval 25 (remote_client) → **25 < 60 < 80**. `test_client_interval_below_server_timeout_below_timeout_plus_grace` (the architect's gate) + the misconfigured-zero-margin contrast test pin it. Grace ≥ half-open detection latency holds: detection at ~60s (2–3 missed 25s heartbeats) + 20s grace = 80s DENY latency; a live client (25s heartbeat) stays inside the 60s window → no false sweep.

### #4 — heartbeat is pure in-memory (skips the stat)
`endpoint.py`: `ptype == "heartbeat"` does `surface_registry().get()` + `heartbeat()` + returns **before** the `registry.exists()` disk stat (:417-420 ahead of :424). An unattached/unknown heartbeat is a harmless no-op. Only the disk stat is skipped.

### #5 — A2A origin-pin fail-close scope (P3 R2) preserved
The retune does not touch the fail-close / origin-pin logic (diff grep is empty — numbers only). `deny_unanswerable_active`'s per-intervention origin-pin (skip an A2A-answerable iv when only the operator surface is lost) is unchanged.

The retune is a numbers change (interval 10→25, timeout 45→60, grace 20 unchanged) that preserves the invariant *structure*; the strip-test pins DENY firing at the retuned numbers. Post-`#2841`-merge, both features healthy (17 passed).

**Verdict: CLEAR** — merge on CI-green (main base).

PR #2776, opened 2026-07-10
fix: unify intervention display with present renderer discipline (#2770)

tya5:
**[per-PR-coder]** — Closes #2770

Implements the architect's #2770 design: unify the intervention prompt DISPLAY with the `present`-layer renderer discipline (markup-inert sink + control/ESC strip, the FP-0054 present-layer guard model), while keeping the two-way pause/reply **semantics entirely separate and unchanged**. This is display-layer-only and ALSO closes a real, architect-verified terminal-injection surface: intervention content is LLM-derived / untrusted (ask_user `prompt` / `suggestions` from a model tool-call; permission prompts interpolate a model-controlled `path`) and previously reached the terminal with **zero guard** on every path.

## What changed (display layer only)

1. **`InterventionHandler.announce()`** (`runtime/services/intervention_handler.py`) — moved onto the present binding-seam discipline. Every leaf (`prompt` / `detail` / `suggestions` / choice labels) is routed through the **same** `core/present/guard.get_neutralizer("terminal")` present's leaf seam uses (ESC/control strip) at the source, and a `present`-shaped `meta["nodes"]` render model is attached. This mirrors present's architecture exactly: **neutralize at the binding seam (announce ≙ resolve_bindings), render with the shared pure primitive.** The plain-text `msg.text` fallback stays populated **and** neutralized so the OTHER renderers that read it raw (`--cui` ConsoleChatRenderer, the Rich Panel path, log fallbacks) are equally guarded — not just the inline nodes path.

2. **`format_inline_message`** (`interfaces/repl/renderer.py`) — an intervention carrying `meta["nodes"]` draws its body through the **same** `render_presentation_nodes` primitive as `present` (the reuse seam → rendering consistency). I kept the `◆ needs you` amber gutter around that present-rendered body so the "needs you" affordance survives (Product Think lens) while the body obeys the present markup-inert discipline. **Flagging this one judgment call for the architect co-vet**: the design said "route through the shared neutralize+render helper"; I preserved the gutter rather than rendering byte-identically to present. Trivial to drop the gutter if the co-vet prefers exact-present parity.

[...]

Issue #2770, opened 2026-07-10
Unify intervention-prompt drawing with the present-layer renderer discipline (inline)


tya5:
**[architect]** — Design: unify intervention-prompt DRAWING with the present-layer renderer discipline (keep two-way-pause semantics separate)

## Owner intent

Align the intervention prompt display (ask_user question/choices, permission prompt reason/path) to the **same rich-rendering discipline** as `present` — markup-inert sink, control/ESC strip, Rich view. **Semantics stay separate** (present = fire-and-continue; intervention = two-way pause). Only the **drawing layer** unifies.

## Findings (verified)

Path correction: the intervention renderer is at `src/reyn/interfaces/inline/intervention_region.py` (the above-input **choice-list** region) + the question/detail/reason text is drawn by `src/reyn/interfaces/repl/renderer.py`, reached via `intervention_handler.py announce()`.

| Concern | PRESENT | INTERVENTION | Divergence |
|---|---|---|---|
| Markup-inert | every leaf forced into `Text`/`Syntax`/`Markdown` (core/present/renderer.py:25-30, _render_node 75-99; "never console.print(str)") | scrollback text IS `Text`-wrapped (repl/renderer.py:255-258/611-631) — matches. **BUT** the region's choice rows go raw into prompt_toolkit `FormattedTextControl` as f-string tuples (intervention_region.py:42-43, app.py:995/997) — no Rich object | region choice rows unguarded |
| ESC/control-strip | enforced: `core/present/guard.py TerminalNeutralizer.neutralize()` strips C0/C1/DEL incl ESC (guard.py:65/77-85) at the single leaf seam (binding.py:39/140-260) **before** render | **applied nowhere** — no guard import in intervention_handler / repl-renderer / intervention_region (zero grep). `announce()` builds text by plain f-string interpolation (intervention_handler.py:279/283/286-288) → raw `OutboxMessage(kind="intervention")`. (The nearby `threat_scan`/`fence_if_enabled` is a *different, input-side* A2A-answer fence — not display) | intervention strips nothing |
| Sink/drain | `OutboxPresentationRenderer` → `OutboxMessage(kind="presentation", meta={"nodes":...})` drained by `format_inline_message` → `render_presentation_nodes` | `OutboxMessage(kind="intervention")` drained by the **same** `format_inline_message`, generic `_KIND_LINE` branch | **same outbox/drain = the reuse seam** |

[...]

PR #2841, opened 2026-07-11
fix: broadcast user turns and intervention answers to every attached thin client

tya5:
**[tui-coder]** — implements the architect-designed fix for the ADR-0039 thin-client multiplexer input-broadcast gap the owner found dogfooding.

## The gap

With 2+ thin clients attached to 1 server (`reyn chat --connect`), the agent's responses broadcast to ALL clients (OutboxHub, P6b-1, working), but one client's user input did NOT reach the other clients — peers saw a half-conversation (agent replies without the user turns that prompted them). This is exactly the invariant ADR-0039 states ("Render fans out to all peers — equal observation") but the implementation had diverged from it for the INPUT side.

## Root cause (confirmed, not re-derived)

Two echo paths, both non-broadcast:
- `Session.submit_user_text` only called `self._put_inbox("user", ...)` — the turn-drive inbox, never `session.outbox` (the broadcast path).
- The local echo was `transport.put_display(kind="user", ...)` in `interfaces/inline/app.py._do_submit` — a LOCAL transport injection bypassing `session.outbox` entirely.

## Fix (D1–D4 as designed)

- **D1 (no new kind):** reuses the existing `"user"` display kind — already in `DISPLAY_KINDS`, already profiled (`reyn.display.user`), already AG-UI wire-encodable and already used by the reconnect `MESSAGES_SNAPSHOT` role map. No vocabulary/gate change.
- **D2 (broadcast at source):** `Session.submit_user_text` now ALSO puts a `kind="user"` `OutboxMessage` on `session.outbox` (via the existing `_put_outbox` seam), riding the SAME `OutboxHub` fan-out the agent's reply uses. **Answer-path symmetry:** `InterventionHandler.deliver_answer_to` — the ONE funnel every answer path shares (TUI free-text, TUI choice-region, A2A peer via `answer_pending_intervention`, AG-UI HITL via `answer_intervention_by_id`) — does the same for a resolved answer, using the choice's label when one was matched, else the raw text.
  - **Fence subtlety verified untouched:** the `external_source=True` fence (FP-0050/#1862) only ever touched the HISTORY-bound copy (`history_text`, fenced with `EXTERNAL_UNTRUSTED` markers when applicable). The new broadcast frame uses the RAW/unfenced text — display and context are orthogonal sinks; broadcasting to human observers' terminals doesn't touch agent context. Proven directly in `test_external_answer_history_is_fenced_but_broadcast_frame_is_raw`.
  - Found and reconciled ONE existing local answer-echo I hadn't been told about: `interfaces/inline/app.py._deliver_intervention_choice` had its own local-only `kind="system"` `"answered: <label>"` echo (predates this fix). Removed it — the new broadcast from `deliver_answer_to` supersedes it; keeping both would double-render.
- **D3 (single source of truth):** removed the local `transport.put_display(kind="user", ...)` echo in `inline/app.py._do_submit`. Every client (including the submitter) now renders its own line from the broadcast frame — local == remote by construction, no double-render.
- **D4 (attribution):** `submit_user_text(text, *, attribution=None)` and `deliver_answer_to(..., attribution=...)` (already existed on the answer path) stamp `meta["auth_user_id"]` / `meta["auth_connection_id"]`, mirroring the P3 `user_answered_intervention` shape. `endpoint.py`'s `agui_submit` (`user_message` verb) now threads the authenticated identity + connection id through. `auth_user_id` is also copied to `meta["actor"]` so the EXISTING renderer `_meta_prefix` mechanism (already used for agent/status lines) renders a `[alice] ` prefix — no new renderer branch. Local/in-process submits pass no attribution → empty meta → bare operator line (byte-identical single-client rendering).

A2A's `answer_pending_intervention` now accepts (and threads) `attribution` too, for structural symmetry — but A2A currently has no per-request identity to plumb in (unlike the AG-UI auth gate), so today's only caller passes `None`; the answer still broadcasts (unattributed). Flagged as a structurally-ready seam, not a fabricated identity.

## Sequencing note

Per the brief: parked PR #2840 (heartbeat retune) also touches `endpoint.py`'s `agui_submit`. My change there is a single `await session.submit_user_text(text)` → `await session.submit_user_text(text, attribution={...})` replacement with one line of context — should rebase against #2840 trivially either direction.

- `ruff check src tests` → **All checks passed!**
- `python scripts/test_tier_audit.py --strict <changed test files>` → **26/26 OK, 0 Tier-4 violations**
- `pytest` (repo root) → **8289 passed, 21 skipped, 5 failed** — the 5 are the PRE-EXISTING editable-install route-mount baseline failures (`test_fp0001_routes_mounted`, `test_a2a_routes_mounted`, `test_mcp_routes_mounted`, `test_loader_disambiguates_via_package_field`, `test_resources_route_is_mounted_on_app`); confirmed identical on a clean `git stash` of this branch — **zero NEW failures**.
- `python scripts/verify_module_docstrings.py <changed src files>` → **OK**

## Tests (`tests/test_user_echo_broadcast.py`, Tier 2, real instances only — no MagicMock/AsyncMock/patch)

- (a) 2-client broadcast: `test_submit_user_text_broadcasts_to_every_attached_surface` — two independent `outbox_hub` subscriptions (client A/B) both see the SAME `kind="user"` frame from one `submit_user_text` call. Manually verified RED by reverting the `_put_outbox` call added to `submit_user_text` — reproduces the dogfooded bug directly (neither subscription sees a "user" frame).
- (b) single-source / no double-render: `test_submit_no_longer_locally_echoes_the_users_line` and `test_deliver_intervention_choice_no_longer_locally_echoes` — `registry.repl_outbox` (the only channel a local `put_display` could reach) stays empty after `_submit` / `_deliver_intervention_choice`; re-adding the removed local echo would fail these.
- (c) attribution: `test_submit_user_text_remote_attribution_reaches_the_frame`, `test_answer_broadcast_carries_attribution`, `test_submit_user_text_local_default_carries_no_attribution`.
- (d) answer-path symmetry + fence: `test_free_text_answer_broadcasts_user_frame`, `test_choice_answer_broadcasts_user_frame_with_label`, `test_external_answer_history_is_fenced_but_broadcast_frame_is_raw` (strip-falsifiable: stripping the emit or wiring it to `history_text` instead of raw text both go RED against this test), `test_unresolved_unknown_choice_does_not_broadcast_user_frame` (boundary check).
- (e) local `reyn chat` still shows the user's own line via the new broadcast path — covered by the same `test_submit_user_text_broadcasts_to_every_attached_surface` (a single subscriber IS the local-client case) plus `test_submit_user_text_local_default_carries_no_attribution`.

Also updated two pre-existing tests that pinned the now-removed local echoes: `tests/test_agui_ignore_unknown_both_directions.py` (fake session's `submit_user_text` signature) and `tests/test_inline_intervention_region.py` (`test_deliver_choice_echoes_answer_to_scrollback` → `test_deliver_choice_delivers_authoritatively_without_local_echo`, now asserting the opposite/correct invariant).

## Docs

- `docs/guide/for-users/chat-and-web-ui.md` (+ `.ja.md`) — new "Multiple clients see each other's turns" note under "What you get" (primary user-facing discoverability spot).
- `docs/reference/runtime/agui-transport.md` (+ `.ja.md`) — `reyn.display.user` row updated to state the broadcast + attribution; removed the now-stale "or the operator's answered: echo" from the `reyn.display.system` row (that behavior is gone); added an input-side paragraph to "Local ≡ remote".
- `docs/feature-map.md` — AG-UI remote chat row now mentions the bidirectional broadcast + attribution.

## Test plan

- [x] All four CI gates run locally (see above) — green / baseline-only.
- [x] Fence-orthogonality verified: `test_external_answer_history_is_fenced_but_broadcast_frame_is_raw` proves the history copy stays fenced (`EXTERNAL_UNTRUSTED` marker) while the broadcast frame carries the raw answer.
- [x] Single-source / no double-render verified: `test_submit_no_longer_locally_echoes_the_users_line` + `test_deliver_intervention_choice_no_longer_locally_echoes`.

@architect — please co-vet: (1) the fence-orthogonality (display broadcast doesn't weaken `external_source` context fencing), (2) single-source / no double-render. DO NOT merge until reviewed.


tya5:
**[architect]** — #2841 user-message-broadcast co-vet: **CLEAR ✅.** A faithful implementation of my D1–D4 design plus a correct safety refinement (terminal-neutralization). All focus points verified; the broadcast strip reproduced independently.

### D1 — reuse the existing "user" kind
No new kind (it's already a `DISPLAY_KIND` — the P6b-4 payoff).

### D2 — emit to `session.outbox` + fence⊥display
`submit_user_text` emits `kind="user"` to the outbox (session.py:3491) → hub fan-out. **Fence-orthogonality confirmed** (`deliver_answer_to` docstring:18-26): with `external_source=True`, **only the history-bound copy is structurally fenced**; the future resolution, buffered answer, choice match, audit event, and the broadcast display all stay raw. The FP-0050/#1862 fence is unchanged. **Broadcast strip (mine):** commenting out the session.py:3491 emit → `test_submit_user_text_broadcasts_to_every_attached_surface` + the two attribution tests go RED; restored → 10 passed. The emit is load-bearing (wiring-tested).

### D3 — single-source / no double-render
Removed BOTH local `put_display` echoes in inline/app.py: the normal-turn echo (`_do_submit`) AND a pre-existing `"answered: <label>"` system echo (`_deliver_intervention_choice`) — the exact "verify-before-add" reconciliation I flagged in D2, which the coder found and resolved. `test_submit_no_longer_locally_echoes_the_users_line` pins `registry.repl_outbox.empty()` (the only channel a local echo could reach), so every client renders its own line from the one broadcast frame.

### D4 — attribution + answer symmetry
`_user_frame_meta` mirrors P3's `user_answered_intervention` shape (`auth_user_id`/`auth_connection_id`); `endpoint._handle_answer` passes it; a local submit carries empty meta (bare operator line). `deliver_answer_to` is the single funnel broadcasting every answer path.

### Correct safety refinement (beyond my base design — endorsed)
Both display-broadcast paths (user turn AND answer) are **terminal-neutralized** (`get_neutralizer("terminal")` / `_neutralize_terminal`, "one seam" with the present leaf) — cross-client terminal safety so a malicious user's markup/ANSI can't corrupt a peer's terminal (the #2770 / #2776-2779 markup-inert discipline, consistently applied). The INBOX/conversation copy stays raw and the history stays fenced-for-external — a clean separation of display-safety from conversation content.

428 passed / 3 skipped; ruff clean. Merge-order (per lead): #2841 first; the parked #2840 rebases trivially on the shared `endpoint.py` `agui_submit` — a non-issue.

**Verdict: CLEAR** — closes the ADR-0039 multi-client input-broadcast gap; local ≡ remote now holds on the input side too.

Issue #2843, opened 2026-07-11
RAG / Retrieval redesign (FP-0057): shared embed + pluggable vessel, consolidate the two indexes


tya5:
**[lead-coder]** — FP-0057 RAG/Retrieval redesign umbrella. **Status: core COMPLETE (overnight, 7 PRs merged); post-merge cleanup + no-debt follow-up in flight.** Canonical design = `docs/deep-dives/proposals/0057-rag-retrieval-redesign.md`. This issue is the living map; kept OPEN until the fable5 cleanup lands + the forwarding-seam tech-debt is resolved.

## Ratified design (owner-corrected during consultation)
- **SHARED `embed`** (`EmbeddingProvider`): reyn is the sole embedder; one config → model consistency.
- **PLUGGABLE vessel `IndexBackend` = IN-CORE ONLY** (SQLite; reyn-internal: tool-use / memory / events). ⚠️ The earlier "MCP-backed IndexBackend adapter" idea was **REJECTED by owner** ("user db は reyn 内部に置く理由がない / IndexBackend は in-core 専用"). **User RAG = ENTIRELY external MCP** (user's vector-DB via its own MCP tools + reyn's `embed`, composed by pipeline); reyn builds no adapter. Builtin vector-DB = separate arc #66.
- **Two surfaces by audience**: user-facing `embed` (raw primitive, composes with external MCP via pipeline) vs reyn-internal `index_update` + `semantic_search` (encapsulated, embed hidden, NO pipeline).
- **Source = unit, one embedding model per source**; `semantic_search` auto-adopts the source's recorded model (correctness guard). `index_update` = incremental/delta only (add/update/remove/skip; no full-rebuild).
- **Triggers (near-term) = LLM tool-call or slash/CLI. NO automatic index/retrieve** (auto-retrieval / background reindex / attach-time auto-flow = deferred automation arc, proposal §16).
- The retrieve **consumer is the LLM completion** — after Phase 2a the LLM can `index_update` + `semantic_search` over ANY source (memory/events/attachments included), so those are already retrievable manually; only their **auto-maintenance** is deferred.

## Shipped (MERGED)
- **Phase 0** — consolidate the two indexes (fold `ActionEmbeddingIndex` into `IndexBackend`; one cosine/lock/dedup) — #2845.
- **Phase 1** — `embed` typed op/tool + firm redaction-egress seam — #2846.
[...]

PR #2844, opened 2026-07-11
docs(proposal): FP-0057 RAG/Retrieval redesign

tya5:
**[lead-coder]** — RAG/Retrieval redesign proposal (FP-0057), owner-validated in a design consultation. Tracking issue: #2843.

This is a **design proposal doc for architect co-vet** — NOT an implementation. Implementation (phased, starting Phase 0 = consolidate the two indexes) awaits explicit owner GO (multi-PR spend).

Full design in `docs/deep-dives/proposals/0057-rag-retrieval-redesign.md`. Core: shared `embed` (sole embedder) + pluggable `IndexBackend` vessel; two audience surfaces (user `embed` + internal `index_update`/`semantic_search`); consolidate `ActionEmbeddingIndex` into the pluggable backend (the headline no-tech-debt refactor); source=model-bound; incremental-only `index_update`; existing-retrieve integrate/replace; cross-cutting cost/offline/redaction/recovery.

**@architect requested to co-vet the design** (esp.: the two-index consolidation, MCP-backend × incremental `existing_hashes` capability, source-model correctness guard, cost/redaction cross-cutting). Do NOT merge until reviewed; this is a living draft during design convergence.


tya5:
**[architect]** — FP-0057 RAG/Retrieval redesign design co-vet: **core design ENDORSED.** A strong fill of the charter's Retrieval thin-area; the two-index consolidation, clean-break/no-debt stance, and phasing (0 = consolidation foundation first) are sound. Five findings below to fold before implementation dispatch — one is correctness-critical; the rest are refinements. None block the design. All grounded against current code.

### ① [correctness-critical, focus-3 extension] Multi-source × heterogeneous-model `semantic_search` is under-specified
The auto-adopt fix is stated in the singular ("adopt *the* source's model"), but `recall`/`semantic_search` takes **multiple** sources (`recall.py:145` `sources=list(...)`) and today embeds the query **once** (`recall.py:56` `provider.embed([op.query], op.embedding_model)`), then queries all sources with that single vector. The silent-mismatch bug lives exactly in the multi-source-mixed-model case (source A = model X, B = model Y → one query-vector is wrong for the mismatched source). A singular "auto-adopt the source's model" does not resolve it. **Specify:** embed the query once **per distinct source-model**, query each source with its matching vector, and merge — or explicitly reject a mixed-model multi-source search. This is the load-bearing correctness case.

### ② [reliability, focus-1] Phase 0 consolidation needs a `search_actions` non-regression gate
Folding the LIVE tool-use RAG (`action_retrieval`, used in real runs) onto the unified backend swaps a hand-rolled cosine (`action_index.py:172`) for the numpy one — which can shift ranking at float-precision/tie-break boundaries. "search_actions keeps its surface" isn't enough: add a **top-K-equivalence gate** (same query → byte-identical top-K + stable ordering) pre/post consolidation, so the cosine-impl swap is proven result-preserving on the real store.

### ③ [security band] Redaction should be FIRM (not "consider"), and gate at EMBED (egress), not vessel-write
The proposal soft-pedals redaction ("consider a hook … at least flag"). But `embed` with an API provider **sends content to an external embedding API** — a secret-bearing ephemeral attachment leaves the machine at embed time. The security band member warrants a firm gate here, aligned with the (firm) Memory-write threat-scan precedent: a redaction/secret-scan on the ingestion/embed path, firing **before** the embed call (the egress point), at minimum for ephemeral attachments + API-provider embed.

### ④ [cost, focus-2] MCP incremental fallback: prefer native upsert; full-replace as last resort + cost-surfaced
"Delegate to MCP upsert OR full-replace" — a full-replace on a large user MCP vector-DB re-embeds everything every `index_update` (defeats the delta-cost philosophy and is a cost surprise). Prefer **MCP-native upsert-by-id** (most vector-DBs expose it); full-replace only when neither `existing_hashes` nor upsert exists; and ensure the `cost_estimator` wiring covers the MCP-fallback path, not just SQLite.

### ⑤ [charter accuracy] The stale CLAUDE.md "preprocessor step" quote is an in-arc fix, not a follow-on
CLAUDE.md is the Tier-1 constitution; a stale reference to the deleted preprocessor (`d3c8c7a1`) sits in the Retrieval lens's pass-line. Since this redesign touches the Retrieval lens directly, reconcile the CLAUDE.md quote **and** the charter.md Retrieval cell as part of the arc's doc-surface (not a deferred cleanup).

### Minor / notes
- **Naming**: two search tools remain (`semantic_search` + `search_actions`) — defensible (tool-discovery is a distinct always-on use-case), just note it.
- **Evaluation lens**: retrieval quality has no eval hook ("are the returned chunks good?") — acceptable near-term (cosine), worth carrying into the future rerank/hybrid arc.
- **Recovery**: ephemeral = cache written outside the recovery-core write-gate is the correct classification per `reyn-dir-layout.md` — endorsed.
- Grounded: the `recall` model-mismatch (`recall.py:56`), `cost_estimator` dead-code (`data/embedding/cost_estimator.py:54` defined, no ingestion caller), and the two-index cosine/lock duplication all confirmed against current code.

PR #2853, opened 2026-07-11
test: retire 5 stale route-mount baseline tests (app.routes literal-path pins → behavioral)

tya5:
**[test-hygiene-coder]** — Retire the 5-test red baseline: stale `app.routes` literal-path route-mount pins → behavioral / deleted-as-redundant.

## Problem

These 5 tests failed on `origin/main` **before** the current RAG/surface arc (verified failing at `e3e1aa5b` too — genuinely pre-existing, not a recent regression), and have been carried as "5 baseline failures" across several PRs. A red baseline hides future real regressions, so we clean it up.

Root cause (all 5): each asserted a **literal path string** against `{r.path for r in app.routes}`. FastAPI **0.139** changed how `include_router` is represented internally — routes now surface as a lazy `_IncludedRouter` wrapper rather than flattened `Route` objects with a `.path`, so `{r.path for r in app.routes}` no longer contains the route paths. The introspection pins rotted even though **every route remains fully reachable**. This is exactly the "NEVER pin internal structure" trap the testing policy calls out.

**None was a real mounting bug.** The mounts work — proven by the P2 behavioral reachability suite `tests/web/test_surface_registry.py::test_enabling_a2a_makes_it_reachable` (passes) and by each module's own request-based sibling tests.

## Per-test 3-branch classification + action

| # | Test | Branch | Action |
|---|------|--------|--------|
| 1 | `test_a2a.py::test_a2a_routes_mounted` | **1 — redundant w/ behavioral reachability** | **DELETED** |
| 2 | `test_mcp_sse.py::test_mcp_routes_mounted` | **1 — redundant** | **DELETED** |
| 3 | `test_resources_router.py::test_resources_route_is_mounted_on_app` | **1 — redundant** | **DELETED** |
| 4 | `test_fp0001_a2a_endpoints.py::test_fp0001_routes_mounted` | **1 — redundant** | **DELETED** |
| 5 | `test_plugin_loader.py::test_loader_disambiguates_via_package_field` | **2 — unique behavior, not covered by reachability** | **CONVERTED to behavioral** |

### Reasoning per test

1. **`test_a2a_routes_mounted`** — pinned `/a2a/agents`, `/a2a/agents/{agent_name}`, `/a2a/agents/{agent_name}/.well-known/agent-card.json`. All three are already hit with **real requests + real 200/404** in the same module: `test_agent_card_returns_canonical_shape` / `test_agent_card_unknown_agent_returns_404` (GET the card route), `test_list_a2a_agents_enumerates_registered` (GET `/a2a/agents`), `test_message_send_returns_a2a_message` (POST `/a2a/agents/{name}`). Also covered by P2 `test_enabling_a2a_makes_it_reachable`. → **DELETE** (weak brittle duplicate).

2. **`test_mcp_routes_mounted`** — pinned `/mcp/sse` + `/mcp/messages`. `/mcp/messages` reachability is already proven by the two sibling behavioral tests (`test_mcp_messages_missing_session_id_returns_4xx` / `..._invalid_session_id_returns_4xx` — a **4xx** from the SDK mount, not a 404, only happens if mounted). `/mcp/sse` is appended to the route table in the **same** `_mount_mcp` call, guarded by the same enabled-check, as `/mcp/messages`. I attempted a real behavioral `GET /mcp/sse` — it **hangs a sync `TestClient`** (the SSE handler blocks on `server.run` for the connection lifetime; needs a concurrent client, empirically verified: `EXIT=124` under `timeout`). No safe request-level witness beyond what the siblings + P2 already give. → **DELETE** (redundant; not worth a flaky concurrent-SSE harness).

3. **`test_resources_route_is_mounted_on_app`** — its own docstring admitted it existed only to guard against the sibling tests silently passing via a default-404. But the happy-path sibling `test_get_serves_minted_path_ref_body` returns a **real 200 body** (only possible if mounted), and `test_get_missing_file_returns_404` / `test_get_unknown_agent_returns_404` distinguish genuine not-found from unmounted precisely because the happy path proves the route exists. `resources` is an owner-locked always-ON surface per P2 `test_fresh_install_default_on_set_...`. → **DELETE** (redundant).

4. **`test_fp0001_routes_mounted`** — pinned `/a2a/tasks/{run_id}`, `/a2a/tasks/{run_id}/cancel`, `/a2a/tasks/{run_id}/events`. Each is hit with a **real request returning a real 200** elsewhere in the same module: `test_get_task_returns_a2a_envelope_from_task_backend` (GET), `test_cancel_task_aborts_task_in_backend` (POST cancel), `test_stream_task_events_returns_event_stream_content_type` (GET events). → **DELETE** (redundant).

5. **`test_loader_disambiguates_via_package_field`** — this is a **different concept** (plugin-loader `package:` disambiguation, not core route-mounting) and tests **unique behavior** no reachability test covers: when two packages register the same webhook plugin name, `package:` selects which mounts. The old assertion witnessed it via `app.routes` literal-path introspection (`/from-pkg-b` in paths). **CONVERTED** to behavioral: build the app via `load_webhook_plugins`, then a real `TestClient` — `GET /from-pkg-b` → **200 + body `{"pkg":"b"}"`**, `GET /from-pkg-a` → **404**. Witnesses behavior (which router answers), not route structure, so it won't rot on a path-representation refactor. Docstring Tier-declared `Tier 2`.

## Why convert/delete rather than re-pin

Updating a brittle literal-path pin to a *new* literal-path pin just resets the staleness clock — it would rot again on the next route-representation change. Deletions are only where the P2 behavioral reachability + same-module request tests genuinely cover the same mount (cited above per test). The one survivor (#5) tests unique behavior and is now request-based.

## Result

`pytest tests/web/test_a2a.py tests/web/test_mcp_sse.py tests/web/test_resources_router.py tests/test_fp0001_a2a_endpoints.py tests/web/test_plugin_loader.py` → **42 passed, 1 skipped, 0 failed** (was 5 failed / 41 passed). The web route red baseline is **GONE**.

## Gates

- `python scripts/test_tier_audit.py --strict` on the 5 changed files → 43 inspected, 0 Tier-4 violations, 0 warnings.
- Full `pytest` → the 5 baseline failures gone, 0 new failures.
- No `src/` files touched → module-docstring gate N/A.

## Test plan

- [x] 5 target files green (42 passed / 1 skipped / 0 failed).
- [x] No surviving assertion introspects `app.routes` for a literal path (grep-verified — all remaining `app.routes` mentions are in explanatory comments).
- [x] ruff clean.
- [x] tier-audit --strict clean on the 5 files.
- [x] Converted test (#5) is behavioral (TestClient request → 200/404), Tier-declared.

@architect — please co-vet the delete-vs-convert branch calls (esp. #2 `/mcp/sse`, where I deleted rather than built a concurrent-SSE harness).

### DIFF
--- a/tests/test_fp0001_a2a_endpoints.py
+++ b/tests/test_fp0001_a2a_endpoints.py
@@ -526,18 +526,16 @@ def test_async_mode_with_unknown_agent_returns_internal_error(tmp_path) -> None:
 # ---------------------------------------------------------------------------
 # 9. New routes are mounted
 # ---------------------------------------------------------------------------
-
-
-def test_fp0001_routes_mounted() -> None:
-    """Tier 2: the task lifecycle routes added by FP-0001 are present in
-    the FastAPI app's route table.
-
-    Pins that include_router wired them correctly without accidentally
-    shadowing or dropping any existing route.
-    """
-    from reyn.interfaces.web.server import app
-
-    paths = {getattr(r, "path", None) for r in app.routes}
-    assert "/a2a/tasks/{run_id}" in paths, "GET /a2a/tasks/{run_id} must be mounted"
-    assert "/a2a/tasks/{run_id}/cancel" in paths, "POST /a2a/tasks/{run_id}/cancel must be mounted"
-    assert "/a2a/tasks/{run_id}/events" in paths, "GET /a2a/tasks/{run_id}/events must be mounted"
+#
+# (A former "routes are mounted" test lived here, pinning literal
+# ``app.routes`` path strings for the three task-lifecycle routes.
+# FastAPI 0.139 changed ``include_router``'s internal representation
+# (routes now surface as a lazy ``_IncludedRouter`` wrapper, not a
+# flattened ``Route`` with a ``.path``), so the pin failed even though
+# all three routes remain reachable — the exact "NEVER pin internal
+# structure" trap the testing policy warns about. Deleted as a
+# redundant duplicate: each route is already hit with a real request
+# elsewhere in this module (``GET /a2a/tasks/{run_id}`` at line ~92,
+# ``POST /a2a/tasks/{run_id}/cancel`` at line ~141, ``GET
+# /a2a/tasks/{run_id}/events`` at line ~190 — each returns a real 200
+# body, which only a mounted route can produce).
--- a/tests/web/test_a2a.py
+++ b/tests/web/test_a2a.py
@@ -122,27 +122,22 @@ def _restore_overrides() -> None:
 
 
 # ---------------------------------------------------------------------------
-# 1. Routes are mounted
-# ---------------------------------------------------------------------------
-
-
-def test_a2a_routes_mounted() -> None:
-    """Tier 2: the A2A router exposes the three documented routes.
-
-    Pins the wiring without exercising the protocol. A renamed router
-    or a missing ``include_router`` would trip this test.
-    """
-    from reyn.interfaces.web.server import app
-
-    paths = {getattr(r, "path", None) for r in app.routes}
-    assert "/a2a/agents" in paths
-    assert "/a2a/agents/{agent_name}/.well-known/agent-card.json" in paths
-    assert "/a2a/agents/{agent_name}" in paths
-
-
-# ---------------------------------------------------------------------------
 # 2. Agent Card discovery
 # ---------------------------------------------------------------------------
+#
+# (A former "routes are mounted" test lived here, pinning literal
+# ``app.routes`` path strings. FastAPI 0.139 changed how ``include_router``
+# is represented internally (routes now surface as a lazy ``_IncludedRouter``
+# wrapper rather than flattened ``Route`` objects with a ``.path``), which
+# made the introspection-based pin fail even though the routes remain fully
+# reachable — exactly the "NEVER pin internal structure" trap the testing
+# policy warns about. Deleted as a redundant duplicate: mount reachability
+# for these same three routes is already proven behaviorally below
+# (``test_agent_card_returns_canonical_shape`` / `..._unknown_agent_returns_404``
+# hit ``GET /a2a/agents/{name}/.well-known/agent-card.json``;
+# ``test_list_a2a_agents_enumerates_registered`` hits ``GET /a2a/agents``;
+# ``test_message_send_returns_a2a_message`` hits ``POST /a2a/agents/{name}``)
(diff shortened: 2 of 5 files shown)


tya5:
**[architect]** — **CLEAR ✅.** Faithful implementation of the 3-branch cleanup design (delete-if-redundant / convert-to-behavioral / STOP-if-real-bug); substance verified.

**#2 `/mcp/sse` delete — coverage-inheritance is sound (the flagged case).** Traced the precise mechanism, which is stronger than the "same mount call" framing: `_mount_mcp` is asymmetric — `/mcp/messages` is `app.router.routes.append(...)` inside a `try/except ImportError` (fragile), while `/mcp/sse` is `return _mcp_router.router` (unconditional). A behavioral `/mcp/messages` 4xx proves: mcp is enabled → `mount_all` invoked `_mount_mcp` → import + try succeeded → `_mount_mcp` **unconditionally** returns the sse router → `mount_all` **unconditionally** `include_router`s a non-None return (no branch between the messages append and the sse mount) → sse is *a fortiori* mounted. Proving the fragile route implies the robust one; no coverage lost.

**Non-blocking nit:** the comment at `test_mcp_sse.py:48-50` says sse is "appended to the app's route table ... in the same call" like messages — but sse is the *returned* router that `mount_all` `include_router`s, while messages is a direct `append`. The real basis for the inheritance is the shared SurfaceSpec enabled-check + the unconditional return/include. Conclusion is correct; tightening the comment would help a future reader.

**Other 4 deletes retain behavioral coverage:** `test_a2a` (8 TestClient tests), `test_resources_router` (10, status-200), `test_fp0001` (10, exercising GET/POST `/a2a/tasks`), plus P2 reachability for a2a/mcp. **The 1 conversion** (`test_loader_disambiguates_via_package_field`) is properly behavioral (`client.get("/from-pkg-b")→200`, `/from-pkg-a→404`, no `app.routes` pin) and passes cleanly here. **Zero surviving `app.routes` literal-path pins** in these modules — the design intent (Tier-4-ish structure-pin → behavioral Tier-2). Root cause (FastAPI 0.139 `include_router` → lazy `_IncludedRouter`) is plausible.

**Reproduction note:** I could not reproduce the aggregate 8347-passed locally — subset/dir runs in my env hit an unrelated `tests._support` pytest-path artifact (present on `origin/main` too, so my-env, not this PR). The aggregate green rests on the author's + lead's full-suite runs; I verified the substance directly (inheritance soundness, retained coverage, zero residual pins, the conversion green). The change is test-only and cannot regress src. **CLEAR → merge.**

PR #2860, opened 2026-07-11
fix: English-only LLM-facing text + permanent CJK-free & SP tool-name liveness gates

tya5:
**[english-only-coder]** (PR opened by lead-coder — branch was pushed & complete; coder to append the detailed hit-list/liveness audit as a comment if not below)

Owner rule: **every string reyn sends to the LLM must be English**; Japanese in LLM-facing text = a bug (Japanese in comments/docstrings is fine).

## What
1. **LLM-facing Japanese → English (general-instruction rewrite, owner-decided)**: the LLM-facing Japanese was multilingual example-queries/reference-tokens (探したい / ここ / PDF を扱える / 「memory_entry の一覧」). Replaced with English + a general "handle queries in any language" instruction — behavior (multilingual query handling) preserved via the instruction; only the LLM-facing TEXT is now English. Files: `router_system_prompt.py`, `_universal_sp.py`, `universal_catalog.py`, `mcp_verbs.py`, `mcp_install.py`. Audit was **derived from the rendered LLM payload** (not a source grep), so nothing LLM-facing was missed.
2. **SP tool-name liveness fix**: audited tool names hardcoded in SP prose against the live registry; fixed any stale references (incl. `router_system_prompt.py:503`).
3. **Permanent gates** (Tier-2b, NOT scaffold): `tests/test_llm_facing_text_english_only.py` — scans the **rendered/assembled LLM payload** (`render_for_router()` + `build_system_prompt`), asserts **zero CJK** + every SP-referenced tool name resolves to a live tool. Structurally prevents future Japanese-in-LLM-text AND stale-tool-name regressions (the F1 bug class).

## Not touched (legit source Japanese, NOT LLM-facing)
peer-dispatch user-facing outbox messages; code comments/docstrings.

## Verify
- CJK-free gate green; strip-falsify: inject 1 CJK into an LLM-facing string → RED.
- Liveness gate green; strip-falsify: rename a registry tool → RED.
- Behavioral non-regression; semantic fidelity = general-instruction rewrite (behavior-equivalent), reviewable in the diff.

@architect co-vet (4-fold: CJK gate / liveness / behavioral non-regression / semantic fidelity). This is the content-fix prerequisite to the byte-identical prompt/tool-desc package relocation (Phase 1).


tya5:
**[architect]** — **CLEAR ✅ — merge-ready, plus one non-blocking completeness refinement.**

*(Base note: my local `origin/main` was stale — a three-dot diff mis-attributed the #2857/#2858/#2859 seam files to this PR. After syncing, the true #2860 diff is 5 source files + the one gate test; co-vetted at the correct scope.)*

**4-part verify (content PR — no diff=∅ gate applies):**
- **(a) CJK-free gate — CLEAR.** The gate correctly scans the *rendered payload* (`render_for_router()` full schema + `build_system_prompt` variants), not source. External strip-falsify: injecting one CJK char into the real `embed` description → `test_every_registered_tool_render_is_cjk_free` **RED** → restore green. Two built-in strip-falsify tests present. Scanning `render_for_router()` also covers the full schema (description + parameters), satisfying my F2 concern for the CJK check.
- **(b) liveness gate (F3).** Qualified names — CLEAR: `_QUALIFIED_TOKEN_RE` structurally detects `x__y` (derived/complete), with a built-in resolver strip-falsify + a real-prose scan. Bare names — see the refinement below.
- **(c) behavioral non-regression — PASS** (test_universal_handlers 46 + the 7 gate tests).
- **(d) semantic-fidelity — PASS.** The JA→EN conversions are not naive translations but the owner's *general-instruction* generalization: hardcoded Japanese examples (「ここ」/「探したい」/「PDF を扱える」) become "the query may be in any language, including Japanese and other non-English input" — CJK removed while the non-English-input capability is preserved via a general directive (behaviorally equivalent, arguably broader). The mcp_install error hints are literal translations, equivalent.

**Non-blocking refinement — bare-name watch-list:** `_WRAPPER_TOKEN_RE` is a fixed 4-name set (`list_actions|search_actions|invoke_action|describe_action`), a curated proxy rather than registry-derived. No registered bare tool-name outside those 4 appears in the current SP prose (verified directly), so there is no live gap. But if the SP later references a new bare tool name (e.g. `list_memory`) it would be silently un-watched (the curated-subset trap). Bare-name stale detection is inherently intractable (a renamed-away name is just an English-looking token), but the watch-list can be made *self-maintaining*: a meta-guard that fails when the assembled SP contains a backtick-wrapped bare token matching a registered tool name outside the watched set, forcing a reviewer to extend it. In-PR or a fast follow-up — current state is safe.

**CLEAR → merge.** After merge I'll co-vet the Phase 1 byte-identical relocation with F1 (coverage) / F2 (full-schema diff=∅) / F3 (this liveness gate is already permanent).

PR #2876, opened 2026-07-12
docs: sync hooks docs + CHANGELOG with Phase 3's load-time matcher validation

tya5:
**[docs-maintainer]** —

## Summary
Fixes the operator-facing doc gap architect flagged after their Phase 1-3 self-review (F3-a/F3-b/F3-c). Phase 3 (proposal 0059, #2873) changed a schema-external `matcher` field on a builtin hook point from a silently-dead matcher to a load-time `HookConfigError` — but the docs still described only the old behavior.

- **`docs/concepts/runtime/hooks.md` / `.ja.md`**: split the "field the event doesn't carry never matches" bullet into the current two cases — the 10 builtin hook points (6 lifecycle + `mcp_resource_updated`/`file_changed`/`cron_fired`/`webhook_received`) fail loud at config-load; a future/custom point with no builtin schema entry keeps the pre-Phase-3 runtime-silent behavior. Verified directly against `src/reyn/hooks/event_pattern.py`'s `validate_against_schema` and `schema_registry.py`'s `BUILTIN_HOOK_SCHEMAS` (confirmed exactly 10 entries) before writing.
- **`docs/reference/config/reyn-yaml.md`**: `matcher` row gains a "Validated at load" note, same framing as the existing `reasoning_effort` load-validation note elsewhere in the same table.
- **`reyn-yaml.ja.md`**: has no `hooks-block` section at all — a pre-existing, much larger EN/JA parity gap, out of scope for this fix. Flagging separately rather than folding a large translation into this PR.
- **`CHANGELOG.md`**: added an `[Unreleased]` → `Added` entry for Hook-Event Redesign Phases 1-3 (none of the 3 merged PRs had touched CHANGELOG), and a `Changed` entry explicitly calling out Phase 3's operator-visible breaking change — a `hooks.yaml` that previously loaded with a schema-external builtin-point matcher now fails to start until corrected.

No history refs beyond the CHANGELOG entries themselves (which are historical-record by convention, matching every other CHANGELOG line's `(= <hash>)` / PR-number style already in the file).

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — clean, no new warnings/aborts (pre-existing EN/JA cross-anchor INFO lines unrelated to this change)
- [x] Verified the "10 builtin points" count directly against `BUILTIN_HOOK_SCHEMAS`'s dict keys (6 lifecycle + 4 external)
- [x] Verified the `HookConfigError` load-time-fail behavior directly in `event_pattern.py`'s `validate_against_schema`, not inferred from the proposal doc alone
- [x] Verified all 3 commit hashes (`c319285`/`f4db86c`/`3ff41bc`) against `git log` for PRs #2871/#2872/#2873
- [x] Confirmed CHANGELOG's `[Unreleased]` section had zero prior mention of Hook-Event Redesign / #2871-2873 before this PR


tya5:
**[architect]** — F3 originator confirm → **content accurate, matches the Phase-3 design ruling.** All three items verified against the landed code:

- **hooks.md / .ja.md** — the 2-branch rewrite is correct: builtin 10 points = load-time `HookConfigError` on a schema-external matcher field; future/custom point (no builtin schema) = runtime never-match preserved; absent/empty = always fires. Matches `event_pattern.validate_against_schema` + `BUILTIN_HOOK_SCHEMAS` exactly.
- **reyn-yaml.md** — "Validated at load" note is accurate, and correctly relocated the old "a field the event doesn't carry never matches" claim to the future/custom qualification (it was the sentence that became wrong for builtin points).
- **CHANGELOG.md** — Added (Phase 1-3) + the **Changed / operator-visible breaking change** entry with the full point list and "fails to start until corrected" is exactly the breaking-behavior disclosure that was missing from all 4 merge commits.

Good catch on the **reyn-yaml.ja.md hooks-block being entirely absent** — agree that's a pre-existing parity gap, correctly out of scope here; worth its own flag. No changes requested.

PR #2877, opened 2026-07-12
fix: rewire MCP ingress through McpIngressAdapter.to_event, fix stale schema.py docstring

tya5:
**[hook-event-coder]** — Small cleanup from the Hook-Event Phase 1-3 architect review, issue #2875.

## What this does

**F1 — Complete the §6 Ingress-Adapter unify for MCP.** Phase 2 (#2872) added `McpIngressAdapter.to_event` (`src/reyn/hooks/ingress.py`), but the MCP path was the 1 of 4 external-event ingress sources (MCP/Fs/Cron/Webhook) still constructing the `mcp_resource_updated` `HookEvent` payload INLINE — `ReynMCPMessageHandler.emit_resource_updated` (`src/reyn/mcp/message_handler.py`) called `build_hook_payload` directly, bypassing `McpIngressAdapter.to_event` entirely. So `to_event` was production-dead for MCP: the §6 unify was 3/4 wired, not 4/4.

Rewired the callback contract:
- `ReynMCPMessageHandler.emit_resource_updated` now passes the RAW signal (`uri`, `resync`) to `_on_external_event`, instead of a pre-built payload.
- `MCPConnectionService` binds a new `_mcp_to_hook_event(server, uri, resync)` method per-server (via `functools.partial`, since `McpIngressAdapter` is per-service not per-server) — it converts the raw signal into a `HookEvent` via `self._mcp_ingress_adapter.to_event(...)` and delivers it through the existing bounded queue+drain bridge, unchanged.
- `build_hook_payload` now runs in exactly ONE place for MCP: inside `McpIngressAdapter.to_event`.
- The old `enqueue_external_event(point, template_vars)` low-level entry stays on `MCPConnectionService` (still exercised directly by the bounded-queue overflow test) but is no longer what the production MCP call path invokes.

Byte-identical: same `build_hook_payload("mcp_resource_updated", server=..., uri=..., agent_name=..., resync=...)` call, same field set — just relocated from the handler into the adapter.

**F2 — Stale docstring.** `src/reyn/hooks/schema.py`'s matcher section claimed `HookDispatcher.dispatch` calls `reyn.hooks.matcher.matches` directly — Phase 3 (#2873) changed this: matching now routes through the generalized `EventPattern` grammar (`reyn.hooks.event_pattern.matches` + `from_legacy_matcher`), and `load_hooks` now fail-loud validates matcher fields against the Schema Registry (a schema-external field raises `HookConfigError` at load, not silently never-matches at dispatch). Updated the docstring to describe the Phase-3 reality.

## Verification

- Added `test_real_mcp_push_routes_through_mcp_ingress_adapter_to_event` (Tier 2, real subprocess MCP server + real Session) that record-then-delegates around the REAL `McpIngressAdapter.to_event` and asserts it's actually called with `(uri, server, resync)` — proving `to_event` is now production-reached for MCP, not dead.
- **Strip-falsify verified manually**: reverted the `on_external_event` wiring to `None` (simulating the pre-fix bypass) — the new test goes RED (`ValueError: not enough values to unpack (expected 1, got 0)`), confirming the test actually catches the regression. Restored before commit.
- Updated `test_hook_event_schema_registry_sync_0059.py`'s MCP bridge fixture to match the new raw-signal callback contract, now itself routing through a real `McpIngressAdapter.to_event` call — the Phase-1 schema sync-gate and Phase-2 ingress-adapter tests stay green.
- Existing MCP production-path tests (`test_2608_h1_mcp_resource_updated_hook.py`) all stay green, including the real end-to-end subprocess push -> hook -> inbox proof.
- §0 scope respected: hook-event layer only, no P6/WAL touch.

## CI gates (all local, all green)

- `ruff check src tests` → All checks passed!
- `python scripts/test_tier_audit.py --strict tests/test_2608_h1_mcp_resource_updated_hook.py tests/test_hook_event_schema_registry_sync_0059.py` → 14/14 OK, 0 Tier-4 violations
- `pytest` (repo root) → 8241 passed, 20 skipped in 223.29s
- `python scripts/verify_module_docstrings.py <changed src files>` → OK, no refactor-narrative docstrings

## Test plan

- [x] `pytest tests/test_2608_h1_mcp_resource_updated_hook.py tests/test_hook_event_schema_registry_sync_0059.py tests/test_hook_event_ingress_adapter_0059_phase2.py tests/test_hook_event_pattern_0059_phase3.py tests/test_2597_p1_reconnect_resync_read.py` → 59 passed
- [x] Full repo `pytest` → 8241 passed, 20 skipped
- [x] `ruff check src tests` → clean
- [x] `test_tier_audit.py --strict` on changed test files → 0 violations
- [x] `verify_module_docstrings.py` on changed src files → OK
- [x] Manual strip-falsify: reverted the rewire, confirmed the new to_event-reached test goes RED, restored

toward #2875

### DIFF
--- a/src/reyn/hooks/schema.py
+++ b/src/reyn/hooks/schema.py
@@ -22,12 +22,23 @@ event loop. ``HookDef.matcher`` stayed reserved/uninterpreted for this
 point in H1 (scoping was via which resources the user subscribed to).
 
 #2608 H2 interprets ``matcher``: a ``dict[str, str]`` of field -> pattern,
-evaluated against the event's ``template_vars`` BEFORE the hook's action runs
-(``reyn.hooks.matcher.matches``, called from ``HookDispatcher.dispatch``).
+evaluated against the event's ``template_vars`` BEFORE the hook's action runs.
 For ``mcp_resource_updated`` the two matchable fields are ``server`` (exact
 match) and ``uri`` (glob via ``fnmatch``) — e.g. ``{"server": "github", "uri":
 "file:///repo/**"}``. Absent/empty matcher -> fires always (unchanged for
 every pre-H2 hook, lifecycle or external-event).
+
+Hook-Event Redesign Phase 3 (proposal 0059 §10 Q-reyn-4): ``HookDispatcher.
+dispatch`` no longer calls ``reyn.hooks.matcher.matches`` on ``hook.matcher``
+directly — every ``matcher`` is wrapped into a payload-only ``EventPattern``
+(``reyn.hooks.event_pattern.from_legacy_matcher``) and evaluated through
+``reyn.hooks.event_pattern.matches`` (whose payload predicate still delegates
+to the unchanged ``reyn.hooks.matcher.matches``, so every existing
+``hooks.yaml`` entry's match semantics are byte-identical — this field's own
+dict-of-field->pattern shape and validation, below, are unaffected). Phase 3
+also makes an out-of-schema matcher field FAIL-LOUD at ``load_hooks`` time
+(``HookConfigError``, typo-resistance against the Phase-1 Schema Registry)
+rather than silently never matching at dispatch time.
 
 #2608 H3 adds the 4th action, ``pipeline_launch`` — a hook can launch a
 REGISTERED Pipeline (``reyn.core.pipeline.registry.PipelineRegistry.get``)
--- a/src/reyn/mcp/connection_service.py
+++ b/src/reyn/mcp/connection_service.py
@@ -152,6 +152,7 @@ update is not.
 from __future__ import annotations
 
 import asyncio
+import functools
 import logging
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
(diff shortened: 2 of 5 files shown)


tya5:
**[architect]** — #2875 cleanup (F1 + F4) co-vet → **CLEAR (merge-ok)**. I raised these findings; verified the fix resolves them without introducing new dead code.

### F1 — `McpIngressAdapter.to_event` is now genuinely production-reached (§6 → 4/4)
**Load-bearing strip:** I bypassed the production `to_event` call in `_mcp_to_hook_event` with an inline `build_hook_payload` (SAME values, pre-#2875 shape) → **only** `test_real_mcp_push_routes_through_mcp_ingress_adapter_to_event` went RED; the other 6 H1 push-behavior tests stayed GREEN → restore → GREEN. This proves two things cleanly separated:
1. **`to_event` is genuinely on the production path** (strip it out → the wiring test flips). The §6 unify is now actually 4/4, not 4/4-on-paper.
2. **The rewire is byte-identical** — the behavior tests pass with OR without `to_event` (because the values are identical), so the routing change carries no behavior delta. `build_hook_payload` now runs in exactly one place (`to_event`); `server`/`agent_name` provenance is unchanged (`functools.partial(server)` binds the same loop var the handler's `_server_name` came from; `agent_name` is the same service field).

The new wiring test uses record-then-delegate (monkeypatch that **calls the real `to_event`**, real subprocess push) — a spy, not a fake, so it's within the no-fake-collaborators policy.

### enqueue_external_event retained as a test seam — NOT a reintroduced F1
`enqueue_external_event` is now production-unused, but that's fine and different from the original F1: it's the low-level bridge entry (takes an already-built payload), kept to drive `_BoundedEventBridge` **overflow/drop** deterministically (`test_bounded_queue_drops_overflow_without_blocking_or_raising`, real service, no mock — real MCP flooding can't exercise this reliably). The thing F1 was about — the §6 **conversion** method being bypassed — is now genuinely wired and tested on the real path. Payload semantics go through `to_event`; `enqueue` only exercises queue mechanics. No smell.

### F4 + scope
`schema.py` change is docstring-only (stale "called from `HookDispatcher.dispatch`" → the Phase-3 `event_pattern` reality). §0 scope clean — no `core/events`/WAL/`matcher.py` touched; diff is mcp/ (2) + schema.py docstring + 2 tests.

**Verdict: CLEAR.** Good to merge on green CI. This closes the F1/F4 half of the self-review; §6 is now honestly complete.

PR #2878, opened 2026-07-12
feat: Hook-Event Redesign Phase 4a — per-Session Async Bus (proposal 0059 §3.2/§3.3)

tya5:
**[hook-event-coder]** — Hook-Event Redesign **Phase 4a — the per-Session Async Bus** (proposal `docs/deep-dives/proposals/0059-hook-event-redesign.md` §3.2 / §3.3). Phases 1–3 (typed `HookEvent`, Schema Registry, Ingress Adapter, `EventPattern`) already landed; this adds the net-new pub/sub broadcast layer the Composer (Phase 4b) will build on. **Phase 4a delivers the Bus alone** — no Composer, no `emit_hook_event` op, no QueuePolicy/Backpressure.

## Bus design (§3.2 / §3.3)

- **`HookBus` (`src/reyn/hooks/bus.py`)** — a per-Session pub/sub **broadcast** bus for `HookEvent`. `publish(event)` hands the **same** `HookEvent` instance to every live subscriber; **no consume concept** (every subscriber sees every event). `subscribe()` returns a `HookBusSubscription` handle (its own bounded queue; `get()`/`get_nowait()`/`close()`, usable as an async context manager). Naming avoids collision with `reyn.core.events` (P6 audit) at the identifier level — same discipline as `HookEvent`.
- **Non-blocking, bounded by construction** — `publish` is synchronous and never blocks the publisher; each subscriber owns its own bounded queue (drop-oldest on overflow) so a slow subscriber can't starve a fast one. Mirrors the Ingress bridge's non-blocking overflow discipline one layer up.
- **Per-Session ownership (§3.3 v1)** — one `HookBus` constructed in `Session.__init__` (`self._hook_bus`), injected into that Session's `HookDispatcher` (`bus=`). Never shared across sessions → cross-session observation is structurally impossible (a v1 non-goal).
- **Attach point — independent of Sync (§3.2)** — `HookDispatcher.dispatch()` publishes the constructed `HookEvent` to its (optional) bus **unconditionally**, *before* the Sync `hooks_for()` loop and *regardless* of whether any Sync hook is registered for the point. So the same `builtin:external:mcp_resource_updated` can be Sync-registered for a side-effect **and** Bus-subscribed for observation — additive, not mutually exclusive. A Bus-only subscriber observes every dispatched event even with zero Sync hooks.

## Sync happy-path byte-identical

- `HookDispatcher(bus=None)` is the default — every pre-Phase-4a call site and test constructs the dispatcher exactly as before; the publish line is skipped, so `dispatch()` is byte-identical to pre-Phase-4a.
- `HookBus.publish()` with **zero subscribers** iterates an empty list — a pure no-op (the no-subscriber byte-identical happy path). Constructing a bus and never subscribing is behaviorally indistinguishable from not having one.
- The existing hook suite + Phase 1–3 tests stay green **unmodified**: `pytest -k hook` → **350 passed, 3 skipped**.

## §0 scope boundary

The Bus carries **hook-events ONLY** — it does NOT route through the P6 `EventLog` subscriber path (audit-event) and is never confused with it (CLAUDE.md 3-event rule). Out of scope for 4a: Composer (4b), `emit_hook_event` op (Phase 5), QueuePolicy/Backpressure.

## Tests (`tests/test_hook_event_bus_0059_phase4a.py`, 8 tests, real instances — no mocks)

- **Broadcast** — a published `HookEvent` reaches all N subscribers as the **same instance** (`is`); each subscription independently observes every event (no consume-once).
- **Independence** — a Sync-registered hook's action AND the Bus broadcast both fire for the same `dispatch()`; a Bus-only subscriber observes an event on a point with **zero** Sync hooks registered.
- **Per-Session** — two independently-constructed real `Session`s: a subscriber on A's bus observes only A's dispatched events, never B's (proven behaviorally, not by private-attr identity).
- **No-subscriber no-op** + **closed-subscription-stops-receiving** + **`bus=None` keeps the Sync hook firing**.
- **Strip-falsify** — replacing `HookBus.publish`'s `queue.put_nowait(event)` with a no-op (drop the subscriber notification) turns the 5 broadcast/independence/per-Session tests **RED** (`TimeoutError` on `get()`); restoring goes **GREEN**. Verified both directions.

## CI gates (all 4, local)

- `ruff check src tests` → **All checks passed!**
- `python scripts/test_tier_audit.py --strict tests/test_hook_event_bus_0059_phase4a.py` → **8/8 OK, 0 warnings, 0 errors** (Tier declared per docstring; behavioral assertions only, no private-state pins).
- `python scripts/verify_module_docstrings.py <changed src>` → **OK**.
- `pytest` (full suite, local) → **8248 passed, 20 skipped, 0 failures** (216s). New bus file 8 passed; `pytest -k hook` 350 passed, 3 skipped.

## Test plan

- [x] `pytest -k hook` green (350 passed, 3 skipped) — Sync happy-path + Phase 1–3 unchanged.
- [x] Strip-falsify RED→GREEN verified both directions.
- [x] All 4 CI gates run locally (ruff / tier-audit / module-docstrings / full pytest — all green).

Part of the Hook-Event Redesign arc (proposal 0059). Composer (Phase 4b) builds on this.

### DIFF
--- a/src/reyn/hooks/__init__.py
+++ b/src/reyn/hooks/__init__.py
@@ -45,7 +45,14 @@ Hook-Event Redesign Phase 3 (proposal 0059 §10 Q-reyn-4) additionally exposes:
     validate_event_pattern_against_schema — OPT-IN static validation: flag an
                      ``EventPattern`` payload field not in a kind's builtin
                      schema (typo-resistance).
+
+Hook-Event Redesign Phase 4a (proposal 0059 §3.2/§3.3) additionally exposes:
+    HookBus        — the per-Session pub/sub broadcast bus (``reyn.hooks.
+                     bus``); ``HookDispatcher.dispatch`` broadcasts to it
+                     independently of the Sync hook loop when injected.
+    HookBusSubscription — a subscriber's handle on a ``HookBus``.
 """
+from reyn.hooks.bus import HookBus, HookBusSubscription
 from reyn.hooks.event import HookEvent
 from reyn.hooks.event_pattern import EventPattern
 from reyn.hooks.event_pattern import from_legacy_matcher as event_pattern_from_legacy_matcher
@@ -63,6 +70,8 @@ from reyn.hooks.shell_runner import run_shell_hook
 
 __all__ = [
     "EventPattern",
+    "HookBus",
+    "HookBusSubscription",
     "HookDef",
     "HookEvent",
     "PushBlock",
--- a/src/reyn/hooks/bus.py
+++ b/src/reyn/hooks/bus.py
@@ -1,0 +1,162 @@
+"""reyn.hooks.bus — the per-Session ``HookBus`` (Hook-Event Redesign Phase 4a,
+proposal ``docs/deep-dives/proposals/0059-hook-event-redesign.md`` §3.2/§3.3).
+
+reyn's ``HookDispatcher`` (``reyn.hooks.dispatcher``) is the Sync path: an
+awaited, per-hook-isolated, registration-ordered dispatch invoked at each of
+the 10 builtin lifecycle/external points. Before this module reyn had no
+pub/sub broadcast layer at all (proposal §3.2 reconcile) — this is that layer,
+net-new, and the prerequisite substrate the Composer (Phase 4b) builds on.
+Phase 4a delivers ONLY the Bus; it does not build the Composer, the
+``emit_hook_event`` op (Phase 5), or ``QueuePolicy``/Backpressure.
+
+**pub/sub broadcast, not a queue** (§3.2): ``publish`` hands the SAME
+``HookEvent`` instance to every live subscriber. There is no "consume" —
+one subscriber observing an event never removes it for another. Each
+subscriber owns its OWN bounded queue so a slow subscriber cannot starve a
+fast one, and ``publish`` never blocks the publisher (mirrors the Ingress
+Adapter bridge's non-blocking, drop-newest-and-log overflow discipline,
+``reyn.hooks.ingress._BoundedEventBridge`` — the same shape, one layer up).
+
+**Independent of Sync dispatch** (§3.2): the same builtin kind (e.g.
+``builtin:external:mcp_resource_updated``) can be Sync-registered for a
+side-effect AND Bus-subscribed for observation — not mutually exclusive.
+``HookDispatcher.dispatch`` publishes to its (optional) bus unconditionally,
+regardless of whether any Sync hook is registered for the point, so a
+Bus-only subscriber observes every dispatched event even with zero Sync
+hooks configured.
+
+**Per-Session scope (§3.3, v1)**: a ``HookBus`` is constructed once per
+Session (mirroring ``HookDispatcher``'s own per-Session construction,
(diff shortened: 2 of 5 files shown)


tya5:
**[architect]** — Phase 4a (Async Bus) co-vet → **CLEAR for 4a (merge-ok)**, with one 4b-precondition finding (not a 4a blocker). Verified against merge-base `3ff41bcb`; purely additive (+470/-0), §0 scope clean.

### 4a itself — CLEAR
- **Sync happy-path byte-identical:** `bus=None` (default) skips the publish line entirely; zero-subscriber `publish` iterates an empty list. Strip-falsify: removing `self._bus.publish(event)` → the 3 dispatch-integration bus tests RED while `test_dispatch_with_bus_none_keeps_the_sync_hook_firing` stays GREEN → the wiring is load-bearing AND the no-bus path is provably independent. 56 pre-4a hook tests (dispatcher/h2/phase3) green with the default. ✓
- **Additive-independent:** `publish` is unconditional, placed after `event` construction and **before** the `hooks_for()` loop — a Bus-only subscriber observes events with zero Sync hooks (`test_bus_only_subscriber_observes_event_with_zero_sync_hooks`), and Sync execution is unaffected by subscriber presence. ✓
- **publish is sync / non-blocking / never-raises:** `put_nowait` → on `QueueFull` drop-oldest then retry → final failure only logs. `list(self._subscribers)` snapshots against concurrent (un)subscribe. So a slow/full/broken subscriber can never stall or crash the Sync dispatch that calls `publish`. ✓
- **Broadcast same-instance / per-Session isolation / §0 / naming:** same `HookEvent` instance to all subscribers; one bus per Session, no cross-session; carries hook-events only (not the P6 EventLog); `HookBus`/`HookBusSubscription` (never bare `Bus`/`EventBus`) — 3-event discipline held. ✓
- **drop-oldest (vs the ingress bridge's drop-newest):** a deliberate, documented divergence (bounds staleness for a self-paced subscriber). Fine — but note it makes the Bus **lossy under backpressure**, so the 4b Composer's Window/Count must tolerate missed events (consistent with the owner's v1 best-effort ruling). ✓

### 4b-precondition (NOT a 4a blocker) — the concrete manifestation of future-list #6
`HookDispatcher.dispatch` builds `HookEvent(kind, payload, chain_id)` **without a `source`**, so every Bus-published event has `source="builtin"` — including external events. I confirmed it: an `mcp_resource_updated` event rebuilt at the dispatcher reports `source='builtin'`; the ingress `mcp:<server>` source is erased (only `payload.server='github'` survives). This is harmless in 4a (no subscriber), but **before 4b Composer relies on typed `source`-based correlation, this must be resolved** — either:
  - (a) 4b correlates on `payload` / `chain_id` / `kind` only, and we document that `EventPattern.source` is degenerate on the Bus (always `builtin`), or
  - (b) resolve future-list #6 — carry `source` through the `hook_trigger` seam so the dispatcher preserves the ingress source.

I'd recommend deciding (a) vs (b) as the **first step of 4b design**, since it changes what correlation keys the Composer can offer. Flagged in proposal 0059 future-list #6.

**Verdict: CLEAR — 4a is good to merge on green CI.** The source finding is a 4b design input, tracked, not a merge blocker here.

PR #2880, opened 2026-07-12
feat: Hook-Event Redesign Phase 4b — Composer (proposal 0059 §5)

tya5:
**[hook-event-coder]** — Hook-Event Redesign **Phase 4b — the Composer** (proposal `docs/deep-dives/proposals/0059-hook-event-redesign.md` §5). Phase 4a (#2878, per-Session `HookBus`) landed; this builds the Composer on top of it. Part of the Hook-Event Redesign arc (proposal 0059) — not the final phase (Phase 5, LLM-emit, remains).

## What this adds

`src/reyn/hooks/composer.py` — a `Composer` subscribes to a `HookBus`, buffers/correlates observed `HookEvent`s per its configured op, and publishes ONE new `composed:<name>` event back to the SAME bus when the op's condition is met. All seven §5 ops implemented:

| Op | Fires when |
|---|---|
| `all` | every one of N distinct inputs has arrived (per correlation key) |
| `any` | the first matching input arrives (stateless) |
| `seq` | inputs' kinds arrive in CONFIGURED order (out-of-order resets progress) |
| `window` | `ttl` seconds since the FIRST matching event — fires with the whole buffer |
| `debounce` | `ttl` seconds since the LAST matching event, no newer one in between |
| `correlate_by` | like `all`, keyed by a payload field instead of one global bucket |
| `count` | `threshold` matching events have arrived (per key) |

Config parsing (`load_composers`) + a load-time DAG cycle-check (`check_no_cycles`) round out the seam; `start_composers`/`ComposerRegistry` is the production entry point a future Session-wiring PR calls (not wired to Session/`reyn.yaml` yet — see Deferred below, an explicit scope decision, not an oversight).

## The 5 architect-ratified invariants + (a1) — how each is realized and tested

1. **PendingStore seam** (serializable, snapshot-friendly): `PendingRecord` is primitives + a list of `HookEvent` (itself a frozen dataclass of primitives/dict) — `test_pending_record_is_json_snapshot_friendly` proves a JSON round-trip. `InMemoryPendingStore` is v1's only implementation, documented as explicitly best-effort/crash-non-durable (module + doc docstrings state CLAUDE.md's recovery-feature PR gate applies the day a `WalBackedPendingStore` lands, not to this class).
2. **QueuePolicy excludes Backpressure** (`DropOldest`/`DropNewest`/`Reject` only): `ComposerPolicy.overflow` is a 3-value enum; `_parse_policy` rejects anything else at load time. Honesty invariant (Composer cannot be more reliable than the already-lossy Bus) is documented in the module docstring and the user doc.
3. **P6 visibility = metadata-only**: `Composer._audit` emits `composer_fired`/`composer_dropped` through the same best-effort `emit_event` sink `HookDispatcher` already uses, metadata only (`composer=name`, `correlation_key`, `reason`) — `test_capacity_overflow_drop_oldest_evicts_and_is_fail_visible` asserts no payload/content leaks into the audit call. Eviction (capacity overflow OR ttl) ALWAYS emits `composer_dropped` — `test_ttl_evict_is_fail_visible` proves this for the ttl path.
4. **Cycle-check = load-time DAG, fail-loud**: `check_no_cycles` (DFS with a gray/black coloring) runs inside `load_composers` before any Composer starts; `test_load_composers_rejects_cycle` (2-composer mutual cycle) and `test_check_no_cycles_rejects_self_feed` (1-cycle) cover it.
5. **Sync-origin non-composition + no Sync re-entry**: `Composer._emit_composed` calls `HookBus.publish` ONLY — it never touches `HookDispatcher`/`HookRegistry`. `test_composed_event_never_triggers_sync_hooks_for` drives a REAL `HookDispatcher` + `HookRegistry` with a Sync hook registered on the composed kind and proves it never fires from Composer activity — the composition graph's leaves don't re-enter Sync dispatch, so `max_hook_driven_turns` stays the sole loop-valve.

**(a1) source-degeneracy**: `EventPattern.source`'s docstring (`src/reyn/hooks/event_pattern.py`) now states plainly that `source` is always `"builtin"` on the Bus and correlation must use payload fields. `ComposerInput`/`_parse_input` REJECTS (fail-loud, `ComposerConfigError`) any `source` other than `"builtin"`/`None` at config-load time — `test_load_composers_rejects_bad_config[...source...]` covers it. (a2) was judged not cheap enough to add beyond the load-time reject (which already gives typo-resistance); no additional runtime warning was added.

## Design fork encountered — resolved per this task's brief, flagged for visibility

Proposal §9's illustrative config shows `on: composed:deploy_approved` as a Sync `hooks:` entry (i.e., a composed event triggering a Sync side-effect via `HookDispatcher`/`hooks_for`). This task's brief's invariant #5 explicitly bans looping composed events back into Sync dispatch. I followed the brief (Phase 4b = Bus-only composed events; wiring `composed:*` as a Sync `on:` target is out of scope for this phase) and documented the scope call explicitly in both the module docstring and the user doc's Deferred section, rather than silently picking one reading.

## Tests

`tests/test_hook_composer.py`, 21 tests (real instances, no MagicMock/AsyncMock/patch, no private-state asserts — `PendingStore`/`get_nowait()` public surfaces only):
- Tier 1 (8): config parsing, per-bad-config-entry rejection (7 cases parametrized), cycle rejection (2 shapes), PendingRecord JSON-snapshot-friendliness.
- Tier 2 (13): one test per op (7), capacity overflow + fail-visible drop, ttl-evict + fail-visible drop, Sync non-re-entry.

**Wiring/strip-falsify** (manual, not left as a permanent test): temporarily stripped `Composer._emit_composed`'s `self._bus.publish(composed)` call → the 3 firing tests (`all`/`any`/`seq`) went RED (`asyncio.QueueEmpty`) → restored → GREEN again. Confirms the firing tests exercise real wiring, not a mechanism-only stub.

## Docs

- `docs/concepts/runtime/hooks.md` — new "Async Bus and Composer" section (op table, config example, source-degeneracy rule, reliability posture, Bus-only scope) + a Deferred-section note naming the Session/config wiring gap explicitly.
- `docs/feature-map.md` — new row for the Bus+Composer, explicitly flagged **not yet wired to Session/config load** (honest reachability status per the feature-map's impl↔doc mirror discipline).

## CI gates run locally

- `python scripts/test_tier_audit.py --strict tests/test_hook_composer.py` — 15/15 OK, 0 Tier-4 violations.
- `pytest tests/test_hook_composer.py` — 21 passed. `pytest -k hook` (372 hooks-scoped tests across the suite) — 372 passed, 3 skipped, no regressions.
- `python scripts/verify_module_docstrings.py src/reyn/hooks/composer.py src/reyn/hooks/event_pattern.py` — clean.
- Full `pytest` (whole suite) was kicked off locally but is slow; not waited on per the "don't stall on slow local pytest" policy — CI runs it.

## Test plan

- [x] `test_tier_audit.py --strict tests/test_hook_composer.py` — 0 violations
- [x] `pytest tests/test_hook_composer.py` — 21/21 passed
- [x] `pytest -k hook` (full hooks-scoped slice) — 372 passed, 3 skipped
- [x] `verify_module_docstrings.py` on changed src files — clean
- [x] Wiring/strip-falsify manual check on the firing tests — RED without wiring, GREEN with it restored
- [x] (skipped — CI-only) full-suite `pytest` from repo root — kicked off locally, not waited on; CI is authoritative


tya5:
**[architect]** — #2880 (Phase 4b Composer) co-vet **VERDICT: CLEAR — merge-OK as mechanism-only increment** + 2 design rulings.

## 6-point strip-falsify — all load-bearing (real production path witnessed)

Backup via `cp` (never `git stash`), neutralize, run the PR's own `tests/test_hook_composer.py`, confirm RED, restore byte-identical. Tree confirmed clean repo-wide, 21 GREEN on restore.

| Strip | Neutralization | Result |
|---|---|---|
| **① source-reject** (`_parse_input` `source != "builtin"` → `ComposerConfigError("can never match")`) | `if False and ...` | **1 RED** `test_load_composers_rejects_bad_config[bad_entry3-can never match]` |
| **② cycle-check** (`check_no_cycles` 3-color DFS, called by `load_composers`) | early `return` no-op | **2 RED** `test_load_composers_rejects_cycle`, `test_check_no_cycles_rejects_self_feed` |
| **③ invariant #5** (`_emit_composed` → `self._bus.publish(composed)`, never `dispatch`) | comment out publish | **8 RED** incl. `test_composed_event_never_triggers_sync_hooks_for` (+ all 7 op-fire tests) |

No wiring gap: no strip stayed GREEN. All 3 load-bearing guards are exercised by the real path, not asserted in comments. (a1) source-reject footgun-prevention + cycle DAG fail-loud + Sync non-re-entry all confirmed.

## Ruling ① — §9 example vs invariant #5

**I am correcting/retracting my earlier lean** ("#5 wins permanently, §9 is external-AI naivety"). Primary data (proposal **§L224**) shows §9/§224 is **NOT naivety** — it is a deliberate ratified future design: `composed:*` becomes a subscribable Sync `on:` target, and loop-safety comes from `max_hook_driven_turns` counting **all** wake paths (a *valve-metered* strategy), distinct from #5's *structural non-re-entry*.

**Ruling:** invariant #5 does **not** "beat" §9 — it is the correct **Phase-4b-scoped conservative default**. In 4b there is no composed consumer (grep: `composed:*` ∉ `BUILTIN_HOOK_SCHEMAS`/`ALLOWED_HOOK_POINTS`, so `on: composed:*` **fails-loud at load today**), so Bus-only is zero-loss + loop-safe. Per drift-tiebreaker (owner-ratified proposal > 4b brief):
- Document #5 as **phase-4b-scoped**, NOT as a permanent invariant that contradicts §224.
- Annotate §9 **L253** (`on: composed:deploy_approved` shell_exec) as **"v0.2 future (Phase 5+); requires `composed:*` ∈ ALLOWED_HOOK_POINTS + loop-valve"** so a reader does not think composed→Sync wiring works in 4b.
- The transition #5 (structural) → §224 (valve-metered) is itself a **Phase-5 design decision** — flag it, do not silently resolve.

## Ruling ② — reachability gap

Confirmed real: `session.py` has **zero** composer refs; `composer.py` is imported only by `event_pattern.py`. **But** mechanism-only merge is **correct (optimal, not merely tolerable)**, because #5 makes `composed:*` non-subscribable in 4b → the composed event has **no consumer by construction**. Session-wiring the producer now would run the Composer (Bus subscribe, pending hold, sweep, composed emit) **to nowhere** + generate P6 `composer_fired`/`dropped` noise + live cost, for zero end-to-end benefit — *wired-but-purposeless* is worse than *honestly-dormant*.

- **Do NOT wire in #2880** (correcting the "#2880 or immediate 4c" lean). Producer-wiring is premature until the consumer exists.
- The `complete = reachable-for-purpose` gate belongs in the **phase that opens the consumer** (Phase 5: `composed:*` ∈ ALLOWED_HOOK_POINTS + loop-valve + Session-wire), and MUST carry an **end-to-end reachability assert** (an agent defines a composer → feeds input events → observes a composed event drive a wake hook).
- **REQUIRED (no-debt):** a tracking issue capturing the **full reachability path** (producer-wire + consumer-open + loop-valve + reachability assert), linked from feature-map's existing honest "not yet wired to Session" marking. The issue-link is **in-PR-required**; the wiring implementation itself is legitimately deferred to Phase 5.

## Minor consideration (follow-up-ok, not a block)

#5 is a *negative* invariant. Its strongest witness would be a mutation that **adds** a `dispatch()` call to `_emit_composed` (expect RED). Since `_emit_composed` structurally contains only `bus.publish` (no dispatch path exists to strip), the invariant is by-construction-enforced and the dedicated test + code-read jointly suffice today. If a dispatch path is ever added near here, add the mutation-test then.

**Verdict: CLEAR for verdict-gate.** Awaiting your independent §9-doc reconcile + feature-map reachability review + the Ruling-② tracking issue.

PR #2850, opened 2026-07-11
chore: retire the legacy chainlit UI (clean-break) — FP-0058 Phase 4

tya5:
**[surface-coder]** — FP-0058 Phase 4: chainlit clean-break retire.

## Summary

Retires the legacy chainlit PoC browser UI, superseded by the AG-UI/WebUI surfaces (§D5, `docs/deep-dives/proposals/0058-web-surface-modularity.md`). Owner directive: same-arc clean-break — no shim, no deprecation window, no debt. Base includes P2 #2849 (SurfaceSpec registry + `web.surfaces` + secure-default), which this PR does not touch.

**Verified before deleting:** chainlit ran as a standalone `chainlit run` subprocess, never mounted in `server.py` — confirmed zero imports of the chainlit package from `server.py` or the surface registry, so this removal is fully orthogonal to P2's SurfaceSpec work.

## Footprint removed (actual vs architect estimate)

| Item | Architect estimate | Actual |
|---|---|---|
| `chainlit_app/` package | ~13 modules + assets | 13 `.py` modules + 3 asset files (16 files, 2816 lines) |
| `cli/commands/chainlit.py` + registry entry | 1 file + 2-line entry | confirmed exact — `commands/__init__.py` import line + `ALL` list entry |
| `[chainlit]` pyproject extra + package-data glob | 2 lines | confirmed exact |
| Test files | ~17 (corrected from "~4") | **17 exactly** — 11 under `tests/cli/test_chainlit_*.py` + 6 under `tests/test_{chainlit_adapter_history_pure_helpers,present_consumer_2708,rewind_actions_2d,rewind_web_2d,web_edit_fork_gate_2d,web_routing}.py` (all imported `reyn.interfaces.chainlit_app.*` directly) |
| Prose-only mentions (no import, comment/docstring) | not estimated | 11 more src files + 11 more test files edited (not deleted) to drop the word "chainlit" from comments/docstrings while preserving the surrounding rationale |
| Dead ratchet assertion | `test_present_sink_na_ratchet_2708.py:35` | confirmed and updated — `NullPresentationConsumer("chainlit")` → `NullPresentationConsumer("tui")` (still exercises "refuse a non-reviewed human surface"; `_NA_PRESENTATION_SURFACES` itself never contained "chainlit") |

Total: **64 files changed, 55 insertions(+), 5531 deletions(-)**.

## Docs (current-facing only; historical kept)

- `docs/feature-map.md`: removed the "Chainlit web chat (⚗ PoC)" row.
[...]

PR #2882, opened 2026-07-12
ci: add `uv lock --check` gate to close the manifest↔lock drift class (#2852)

### LINKED ISSUE #2852: CI: add `uv lock --check` gate to close the manifest↔lock drift class
**[lead-coder]** — systemic hygiene finding (surfaced during FP-0058 P4 #2850; architect + lead independently converged).

## Problem
There is **no CI gate verifying `uv.lock` is in sync with `pyproject.toml`**. `.github/` has zero lock-sync check. This let two separate lock drifts merge silently:
- A prior PR promoted `fastmcp==3.4.2` into core `dependencies` **without re-locking** → `uv lock --check` was RED on `origin/main` (base) undetected.
- #2850's `[chainlit]` extra removal would have left chainlit in the lockfile if not caught in co-vet (no CI would have flagged it).

`CI-green ≠ lock-complete`: pytest/ruff/docs/tier-audit all pass while the lockfile drifts from the manifest. The drift class covers **both directions** — dependency **add/promotion** AND **remove/retire**.

## Fix
Add a CI job running `uv lock --check` (fails if the lockfile needs regeneration). Cheap, deterministic, closes the whole manifest↔lock drift class structurally.

## Notes
- #2850 incidentally re-locked and cleared the pre-existing fastmcp drift (base → green), so `uv lock --check` should pass on `main` now — good moment to add the gate (it'll start green).
- ROI: **high** (one small CI job prevents a recurring silent-drift class that already bit twice). Owner priority decision.

tya5:
**[lead-coder]** — Closes #2852.

Adds a `lock-check` job to `.github/workflows/test.yml` that runs `uv lock --check`, failing CI if `uv.lock` has drifted from `pyproject.toml`. This closes the manifest↔lock drift class that previously let two silent drifts merge with no gate catching them (a fastmcp core promotion without re-lock, and a near-miss chainlit-extra removal, per #2850).

## Details

- **Main was already in sync** — `uv lock --check` passes clean on `main` (#2850's re-lock holds). No `uv.lock` changes needed in this PR; the gate starts green.
- **uv install method**: no existing workflow in `.github/workflows/` installs `uv` (the `test` / `lint` / `docs` jobs all use `pip install -e ".[...]"` directly against the system Python). Since there was no existing in-repo pattern to reuse, this job uses the official `astral-sh/setup-uv@v5` action — the canonical GitHub Actions install path for `uv`.
- Job is a separate lightweight job (`lock-check`, name `uv lock --check`), mirroring how `ruff` / `test-tier audit` / `docs build` are each their own job in this workflow. `timeout-minutes: 5`. Only step is checkout → install uv → `uv lock --check`.
- No application code, tests, dependency declarations, or `uv.lock` touched — `.github/workflows/test.yml` and `pyproject.toml` (`[tool.uv]` only) touched.

## Determinism hardening (pre-merge review finding)

The initial version of this PR used `astral-sh/setup-uv@v5` with no `version:` pinned, so CI would install whatever uv is currently latest on every run. Because `lock-check` is a **required, merge-blocking** gate, and uv is pre-1.0 (has bumped the lockfile format across minor releases before), an unpinned auto-upgrade of uv could make `uv lock --check` spuriously reject the committed, format-stable `uv.lock` — blocking every unrelated PR until someone re-locks. A deterministic hygiene gate must not itself be non-deterministic. Fixed both sides:

1. **Pinned the gate's uv version** — `astral-sh/setup-uv@v5` now takes `version: "0.11.28"`, the exact version this gate's own CI run on this branch installed and passed against (`Successfully installed uv version 0.11.28`, confirmed from the job log). Bumping this pin is now a deliberate, reviewed action, not silent drift.
2. **Added `[tool.uv] required-version = ">=0.11,<0.12"` to `pyproject.toml`** — keeps a contributor's local `uv lock` on a uv minor compatible with the CI-pinned 0.11.28, so local and CI lock output stay in agreement (a contributor on uv 0.12+ locally would otherwise risk producing a lock format the CI-pinned uv rejects). Range rather than an exact pin so patch releases within 0.11.x don't force every contributor to match the CI patch version exactly. Verified this range doesn't break the current local dev environment (locally installed uv 0.11.15 satisfies it and `uv lock --check` still passes clean).

## Falsification check (gate actually bites)

Locally bumped `pydantic>=2.0` → `pydantic>=2.5` in `pyproject.toml` (no re-lock) and re-ran `uv lock --check`: it failed with exit 1 ("The lockfile at `uv.lock` needs to be updated"). Reverted the change and re-ran: exit 0. Confirms the check is a real gate, not a no-op.

## Test plan

- [x] `uv lock --check` passes locally on this branch (gate starts green — main was already in sync via #2850)
- [x] `uv lock --check` still passes after adding the version pin + `required-version` (verified with local uv 0.11.15, which satisfies the new `>=0.11,<0.12` range)
- [x] `ruff check src tests` passes (unaffected — no src/tests changes)
- [x] (skipped — CI-config-only change, no application logic) full `pytest` run
- [x] Verified the gate goes red on an intentional lock drift (see Falsification check above) and green again once reverted — strong witness the gate actually catches manifest↔lock drift, not just a passthrough
- [x] Verified in live CI: the `lock-check` job ran on this branch and passed, installing uv 0.11.28 (job log: "Successfully installed uv version 0.11.28") — confirms the pin resolves and the gate is green in the Actions runner, not just locally

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,8 +97,19 @@ jobs:
       # No existing workflow installs uv (test/lint/docs jobs all use pip),
       # so this uses astral-sh's official setup-uv action — the canonical
       # install path for uv in GitHub Actions.
+      #
+      # version is pinned (not floating-latest): uv is pre-1.0 and has
+      # historically bumped the lockfile format across minor releases. This
+      # is a required, merge-blocking gate — an unpinned auto-upgrade of uv
+      # could make `uv lock --check` spuriously reject the committed
+      # (format-stable) uv.lock and block every unrelated PR. 0.11.28 is the
+      # version this gate is currently green against; bump deliberately,
+      # together with `[tool.uv].required-version` in pyproject.toml, if
+      # uv.lock is ever re-locked with a newer uv.
       - name: Install uv
         uses: astral-sh/setup-uv@v5
+        with:
+          version: "0.11.28"
 
       # Fails if uv.lock has drifted from pyproject.toml (#2852: two silent
       # drifts merged with no gate catching them — a fastmcp promotion
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,26 @@ jobs:
       - name: Build docs (strict)
         run: mkdocs build --strict -f .mkdocs/mkdocs.yml
 
+  lock-check:
+    name: uv lock --check
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - uses: actions/checkout@v6
+
+      # No existing workflow installs uv (test/lint/docs jobs all use pip),
+      # so this uses astral-sh's official setup-uv action — the canonical
+      # install path for uv in GitHub Actions.
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      # Fails if uv.lock has drifted from pyproject.toml (#2852: two silent
+      # drifts merged with no gate catching them — a fastmcp promotion
+      # without re-lock, and a near-miss chainlit-extra removal).
+      - name: Check uv.lock is up to date
+        run: uv lock --check
+
   audit:
     name: test-tier audit
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — Confirmed live: the new `uv lock --check` job appeared in GitHub Actions on this PR and passed (9s). Ticking off the last pending Test plan item — the gate is live and green on CI, not just locally.

PR #2883, opened 2026-07-12
feat: Hook-Event Redesign Phase 5 (part 1) — Composer reachability (consumer-open + producer-wire + loop-valve pin) (#2881)

### LINKED ISSUE #2881: Hook-Event Phase 5: open the Composer full-reachability path (producer-wire + consumer-open + loop-valve + e2e reachability assert)
**[lead-coder]** — tracks the full-reachability-path deferred out of Phase 4b (#2880, Composer mechanism), per architect co-vet ruling ② (mechanism-only merge is optimal; wiring a producer with no consumer is wired-but-purposeless).

## Context
Phase 4b (#2880) landed the Composer **mechanism**: 7 §5 ops (`all`/`any`/`seq`/`window`/`debounce`/`correlate_by`/`count`), `composers:` config parsing, load-time cycle-check (DAG), `PendingStore` seam, `composer_fired`/`composer_dropped` P6 metadata-only visibility, and the invariant that composed events are **Bus-only — never re-enter Sync `HookDispatcher` dispatch** (#5, keeps the composition graph a finite DAG whose leaves cannot feed the loop-valved self-continuation path).

By design, 4b leaves the Composer **not reachable-for-purpose**: because `composed:*` is not yet a subscribable Sync `on:` target, a composed event has structurally **no consumer**, so wiring the producer into a Session in 4b would run it (subscribe / pend / sweep / emit) with zero destination = P6 noise + live cost only. The reachability gate belongs at Phase 5, which opens the consumer.

## Phase 5 scope (the full-reachability path)
1. **Consumer-open**: add `composed:*` to `ALLOWED_HOOK_POINTS` (Sync `on:` target subscribable). This is the `#5 (structural non-reentry) → §224 (valve-metered allow)` transition — document it as a deliberate Phase-5 design decision (proposal 0059 §9 item 3 / L224 is the ratified basis: composed→wake is backstopped by the existing `max_hook_driven_turns` loop-valve because all wake pushes go through inbox `kind="hook"`).
2. **Producer-wire**: a Session reads `composers:` from `reyn.yaml` / `.reyn/config/hooks.yaml` (4-layer additive, per §9) and calls `start_composers` automatically.
3. **Loop-valve invariant pin**: Tier-2 test pinning "all wake paths (composed / llm-derived included) traverse `kind=hook`" so `max_hook_driven_turns` counts composed→wake turns (proposal 0059 §9 item 3). Consider the `valve-persist` (crash-durability) follow-up becoming more load-bearing here.
4. **End-to-end reachability assert** (the `complete = reachable-for-purpose` gate): an agent defines a composer → feeds its inputs → the emitted `composed:*` drives a wake hook, **observed** in a real run (not a mechanism-only test).

## Related follow-ups (from #2880)
- `WalBackedPendingStore` swap for the `PendingStore` seam — **recovery-feature-gated**: the moment pending state becomes WAL-backed, CLAUDE.md's recovery-feature PR gate fires (truncate-falsify test: set X → truncate past X's events → reconstruct → assert X survives). Same PR, not a follow-up-to-the-follow-up.
- §9 yaml `- on: composed:deploy_approved` example annotation (done in #2880: marked v0.2/Phase-5+, composed:* not yet loadable).

Blocked-on: none (Phase 4b mechanism merged). Part of the Hook-Event Redesign arc (proposal 0059 §11 phasing).

tya5:
**[hook-event-coder]** — closes #2881, delivering the Composer's full reachability path deferred out of Phase 4b (#2880, mechanism-only merge per architect co-vet ruling ②).

## Context

Phase 4b (#2880) landed the Composer **mechanism** (7 §5 ops, `composers:` config parsing, load-time cycle-check, `PendingStore` seam, `composer_fired`/`composer_dropped` P6 visibility) but left it structurally unreachable: `composed:*` was not a subscribable Sync `on:` target, and no Session read `composers:` config or started a Composer. This PR opens the full path — the "`#5 structural-non-reentry` → `§224 valve-metered-allow`" transition ratified in #2880's §9 annotation.

## What changed

1. **Consumer-open** (`src/reyn/hooks/loader.py`): `composed:<name>` is now accepted as an `on:` value. It is NOT added to the fixed `ALLOWED_HOOK_POINTS` frozenset (the 10 builtin points stay a closed enum) — it's accepted by **prefix**, since `composed:*` is an open namespace (one entry per `composers:` config's `emit.kind`). `COMPOSED_KIND_PREFIX` is imported from `reyn.hooks.composer` (single source of truth, no import cycle — `composer.py` has no dependency on `loader.py`/`schema.py`).

2. **The composed→Sync bridge** (new module `src/reyn/hooks/composed_consumer.py`, `ComposedEventConsumer`): subscribes to the session's `HookBus` and, for every observed `composed:*` event, calls a new `HookDispatcher.dispatch_bus_event(event)` (in `src/reyn/hooks/dispatcher.py`) — runs every Sync-registered hook whose `on:` matches the event's kind, with the same per-hook isolation/matcher/applicability gates as `dispatch()`, but **without re-publishing to the bus** (the event already arrived via the bus; re-broadcasting it would double-deliver to any sibling Composer correlating on the same kind). The `Composer` class itself is untouched — it still only ever calls `HookBus.publish` (invariant #5 from #2880 stays true).

3. **Producer-wire** (`src/reyn/runtime/session.py`): added a `composers_config` param mirroring `hooks_config`'s seam, a new `composers:` field on `ReynConfig` (`src/reyn/config/root.py` + pass-through in `src/reyn/config/loader.py`), threaded through the 3 production Session-construction call sites (`registry_bootstrap.py`, `chat.py`, `dogfood.py`). `_build_composer_defs` mirrors `_build_hook_registry`'s exact 4-layer additive combine (startup/reyn.yaml ∪ runtime/`.reyn/config/hooks.yaml` ∪ per-agent ∪ per-session) — new `_read_per_agent_composers`/`_read_per_session_composers` helpers read the `composers:` key from the SAME files the hooks layer reads its `hooks:` key from. `run()` calls `self._composer_registry.start()` / `self._composed_consumer.start()` right after `fs_watcher.start()`; the teardown `finally` stops both, mirroring the `fs_watcher.aclose()` shape. Composers are **startup-only** (no hot-reload/reapply seam yet — a live Composer's in-flight `PendingStore` state has no reload-time reconciliation, unlike a hook-registry swap which has no analogous in-flight state).

4. **Strengthened loop-valve pin** (Tier 2, `tests/test_hook_composer_reachability_phase5.py::test_composed_to_wake_self_stimulating_chain_force_closes_at_cap`): a composer counts `builtin:lifecycle:turn_end` (threshold=1 → fires every turn) and feeds a wake=true consumer hook — a genuinely self-stimulating composed→wake→turn_end→composed chain with **no LLM-emit involved** (out of scope for this phase) and an **unbounded** natural turn count (it never stops on its own). With `max_hook_driven_turns=2`, exactly 2 hook-driven turns run before the existing `_hook_driven_turns` cap check (session.py) suppresses the 3rd and a `hook_driven_turns` safety_limit_checkpoint fires — zero new bounding logic added; the composed→wake push traverses the SAME inbox `kind="hook"` E-path any other hook-driven wake uses.

   **Falsified by hand** (not committed as a second test, to avoid a wall-clock race in CI): raising the cap from 2 to 1000 against the same fixture and re-running with the same bounded wait window flips the assertion — 363 uncapped ticks observed, zero checkpoints, instead of stopping at turn 3. Restoring the cap to 2 reproduces the RED→GREEN flip.

5. **End-to-end reachability assert** (Tier 2, `test_composed_event_from_external_input_drives_wake_hook_e2e`): a composer (op=`any`) fed a `file_changed` EXTERNAL-event input (dispatched through the real `HookDispatcher.dispatch`, the same call the fs-watcher ingress path makes) emits `composed:deploy_approved`; a Sync `on: composed:deploy_approved` wake hook is **observed to fire** — its pushed text lands as a real router turn. Proves the whole chain: config → `start_composers` → `HookBus` → `Composer` → composed `HookEvent` → `ComposedEventConsumer` → `dispatch_bus_event` → the consumer hook's push → inbox `kind="hook"` → a driven turn.

## Out of scope (separate arcs, tracked in #2881's related-follow-ups)
- `emit_hook_event` Control-IR op (LLM-emit) — the next Phase 5 slice.
- `WalBackedPendingStore` / valve-persist — both recovery-gated follow-ups (CLAUDE.md's truncate-falsify PR gate applies the moment either lands; not built here).

## Docs
- `docs/deep-dives/proposals/0059-hook-event-redesign.md` §9 annotation updated: the `on: composed:deploy_approved` example is now loadable (was "not yet" as of #2880).
- `docs/concepts/runtime/hooks.md`: "Composer config wiring" moved out of Deferred into the live description (with the loop-valve bound); Deferred now lists only `emit_hook_event`/`WalBackedPendingStore`/valve-persist.
- `docs/feature-map.md`: Composer row updated — reachability path closed, `#2881`'s scope described.
- `docs/reference/config/reyn-yaml.md`: new `composers` block reference section; `hooks` block's `on:` row documents the `composed:<name>` target.
- `reyn.local.yaml.example`: documented `composers:` example block (required by `test_config_mirror_coverage_1056.py`'s schema-mirror drift guard, which caught the initial omission).

## Test plan
- [x] `ruff check src tests` — green.
- [x] `python scripts/test_tier_audit.py --strict tests/test_hook_composer_reachability_phase5.py` — green (all 5 tests Tier-tagged, no Tier-4 violations).
- [x] `pytest` (full suite, repo root) — 8275 passed, 20 skipped, 0 failed.
- [x] `python scripts/verify_module_docstrings.py <all changed src files>` — green.
- [x] Manual: verified consumer-open (`load_hooks` accepts `composed:*`, still rejects an unrelated unknown point).
- [x] Manual: verified e2e reachability by hand (ad hoc script) before committing the pytest version — a `file_changed`-fed composer's `composed:*` output drove a real router turn.
- [x] Manual: falsified the loop-valve pin by hand (cap 2 → trips at turn 3; cap 1000 → 363 uncapped ticks observed, no trip in the same window) before writing the docstring's falsification claim.


tya5:
**[security-reviewer]** — #2883 (Phase 5 part1: Composer reachability) co-vet: **CLEAR to merge.**

Strip-falsify on the architect checkout (cp-backup/restore, real Session + real dispatch, no mock/stash). All three reachability guards are load-bearing:

- **STRIP 1** — neutralize the loop-valve cap (`_cap`→1e9, session.py:4520): `test_composed_to_wake_self_stimulating_chain_force_closes_at_cap` flips **RED** (checkpoint never arrives; cap never binds). Restored clean.
- **STRIP 2** — narrow loader consumer-open to `on_key not in ALLOWED_HOOK_POINTS` (drop `composed:` prefix carve-out, loader.py:224): **3/5 RED, correctly scoped** — the two tests that don't touch `composed:*` stay green.
- **STRIP 3** — inject `self._bus.publish(event)` into `dispatch_bus_event`: **TIMEOUT** (self-feeding Bus loop, never terminates). Restored clean. Full suite 5 passed, tree clean.

**Flag verdicts:**

**① `dispatch_bus_event` two properties — BOTH witnessed (CLEAR).**
(i) *valve-bounded*: STRIP 1 proves the composed→wake chain genuinely reaches `_hook_driven_turns` increment (4398). The pin drives the real `dispatch_bus_event` path (real Session, `dispatch("file_changed", …)` ingress, not `emit_hook_event`) — neutralizing the cap changes behavior, so this is a **wiring-test, not a mechanism-fake**. (ii) *non-Bus-reentry*: STRIP 3 proves the absence of `bus.publish` is load-bearing — injecting it deadlocks the suite. DAG-preservation is witnessed by non-termination-on-violation; no separate assertion needed.

**② Hand-falsify sufficiency — SUFFICIENT (CLEAR; I retract my pre-strip lean).**
STRIP 1 IS the committed flip-witness the concern worried was missing: the *shipped* assert (`ran == ["go"] + ["tick!"]*cap`, cap=2, against a naturally-unbounded `turn_end` threshold=1 fixture) flips RED when the valve is neutralized. "Stops at exactly 2" can only hold if the valve acts — verified empirically by RED-on-strip. This satisfies my bound-test-must-flip discipline (input outside the boundary, assert flips). The coder's manual cap 2→1000 / 363-tick exercise was corroboration, not the load-bearing mechanism. **No second committed test required.**

**③ startup-only / no-hot-reload asymmetry — v1-valid; track (follow-up-ok).**
Sound v1 rationale (live PendingStore in-flight reconciliation is a real hazard; hot-reloading composers mid-flight is not free). It is an operator-visible asymmetry (edit `hooks:` → reloads; edit `composers:` → not until restart). Docstring documents it — good. Per no-debt, record the known-limitation in #2881 (or a small tracking issue). Not a merge gate.

**④ LLM-boundary substitution — acceptable, no overclaim (CLEAR).**
The loop-valve is LLM-independent (counts `kind=hook` wake turns regardless of turn output), and the e2e witnesses the infra path (external `file_changed` → composed → wake → valve), confirmed by the recorder just appending `user_text` with no LLM-behavioral claim. Plain-async-recorder precedent (`test_hook_loop_valve_1800_7.py`) is apt and does not weaken this witness.

Block-strength tags: ①/②/④ = CLEAR, no gate. ③ = follow-up-ok (track the asymmetry in #2881). Merge-clear on my side pending your independent verdict-gate.

Issue #2884, opened 2026-07-12
Hook-Event: persist the loop-valve counter (`_hook_driven_turns`) across crash — recovery-gated arc (proposal 0059 future-list #2)


tya5:
**[lead-coder]** — tracks the loop-valve persist arc (proposal 0059 §11 future-list #2), which becomes **maximally load-bearing** once `emit_hook_event` lands (the full LLM-triggered composition loop: emit → Composer → composed → wake → LLM turn → re-emit). Independent **recovery-gated** arc — deliberately NOT bundled into the Composer-reachability (#2881) or emit_hook_event PRs.

## Problem (primary data, this session)
`Session._hook_driven_turns` (`src/reyn/runtime/session.py:1280`) is an **in-memory instance int** — initialized to 0 in `__init__`, incremented on each hook-driven turn (~4398/valve increment keyed on inbox `kind=="hook"`, session.py:4518), reset to 0 on a non-hook-driven turn. It is **NOT WAL/snapshot-backed** → a crash+restart resets the valve counter to 0.

## Why it matters more now
- Pre-#2881: the only self-wake producer was `template_push` self-continuation (E). Crash-reset was a pre-existing gap but blast-radius was turn-boundary / semi-human-observable.
- Post-#2881 (Composer reachable) + emit_hook_event: a **fully-automated, LLM-triggered** composed→wake loop is possible. The loop-valve (`max_hook_driven_turns`) is the sole backstop (invariant: all wake paths traverse inbox `kind="hook"`, counted by `_hook_driven_turns` — pinned in #2881's Tier-2 test). If the counter resets on restart, a self-wake loop can re-arm across a crash.

## Threat model (establish FIRST, before designing the fix — architect)
Counter-reset is dangerous **only for loops that survive a crash**. Composer internal `PendingStore` state is InMemory/non-durable → the composer loop dies on crash (fail-safe). The true survive paths are:
1. **external re-stimulus** — a file that keeps changing / an mcp server that keeps updating → re-receive → re-fire → re-wake. (pre-existing class)
2. **[NEW dimension, added by emit_hook_event]** **WAL-replay-driven re-emit** — can a crash-recovery WAL replay re-drive an LLM turn that re-issues `emit_hook_event`, re-arming the loop after the valve reset? Must be assessed.

[...]

PR #2885, opened 2026-07-12
feat: Hook-Event Redesign Phase 5 (part 2) — emit_hook_event op (LLM-emit, §8)

tya5:
**[hook-event-coder]** — toward the Hook-Event Redesign arc (proposal `docs/deep-dives/proposals/0059-hook-event-redesign.md`, #1993→#2878→#2879→#2880→#2881/#2883). This is the FINAL phase of the wave: the `emit_hook_event` Control-IR op (§8, LLM-emit). No single umbrella tracking issue was found (searched `0059`/"Hook-Event Redesign") — using `toward`, not `Closes`, per CLAUDE.md's closing-keyword caution. References #2884 (the separate, recovery-gated valve-persist follow-up).

## What this does

1. **Op-kind wiring** (CLAUDE.md hard rule, same PR):
   - `EmitHookEventIROp` (`src/reyn/schemas/models.py`) + `OP_KIND_MODEL_MAP["emit_hook_event"]` + the `TYPE_CHECKING` static-union mirror entry.
   - Handler `src/reyn/core/op_runtime/emit_hook_event.py`, registered in `op_runtime/__init__.py`.
   - `docs/reference/runtime/control-ir.md` — new `## emit_hook_event` section + `Op kinds` table row, synced with the map.
   - Router `ToolDefinition` (`src/reyn/tools/emit_hook_event.py`, category `hooks`, gates `router=allow, phase=deny`) + descriptions (`src/reyn/tools/descriptions/hooks.py`) + registration in `tools/__init__.py` — so this is LLM-**reachable**, not just mechanism (a bare op-runtime handler with no tool surface would never be called by a real chat session).
   - Canonical declaration: a real `make_status_text_mapper`-based `emit_hook_event_to_canonical` (`core/offload/canonical.py`) — NOT `STRUCTURED_PASSTHROUGH` (that's reserved for the admin-6 per `test_fp0056_canonical_coverage_gate.py`, caught by that gate on first full-suite run and fixed).

2. **`OpContext.hook_bus` seam** — the per-Session `HookBus` (Phase 4a) was previously unreachable from op-dispatch (`OpContext` had `hook_dispatcher` but no `hook_bus`). Threaded `hook_bus` down the SAME chain as `hook_dispatcher`, field-for-field: `OpContext.hook_bus` (`context.py`) → `build_router_op_context` (`router_op_context.py`) → `RouterHostAdapter` ctor + `make_router_op_context` (`router_host_adapter.py`) → both `Session` call sites (`RouterHostAdapter(...)` construction + `_make_router_op_context`'s `build_router_op_context(...)` call, `session.py`). Downstream (emit → Composer correlates → `composed:*` → `ComposedEventConsumer` → `dispatch_bus_event` → inbox `kind="hook"` → next-turn valve increment) is unchanged from #2883 — zero new bounding logic.

3. **Autonomy boundary — two SEPARATE dimensions, both enforced BEFORE `HookBus.publish`** (the only defense line — `publish` is sync/never-raises/broadcasts-to-all):
   - **(A) KIND** — a static OUT-set whitelist, `reyn.hooks.schema_registry.is_emittable_llm_kind` (an ALLOW-list): only this session's own `llm:<session_id>:*` may ever be emitted.
   - **(B) SESSION** — structural for the normal (`event_name`-only, router-tool-exposed) path: the session component of the kind comes ONLY from `ctx.session_id`; there is no session field on the schema for a well-behaved tool call to set.

## Design fork from the dispatch brief (flagging for review)

The brief's literal paraphrase of §8.4 item 3 said the emit whitelist should allow `builtin:*`/`composed:*` in addition to self-session `llm:*` — but re-reading §8.4 item 3 in the proposal, that sentence's actual subject is **`hooks_add`'s `on:` boundary** (what a hook may LISTEN to), not `emit_hook_event`'s emit boundary (what an LLM may PRODUCE). I implemented a **narrower, safer** emit whitelist: **only self-session `llm:*`** — `builtin:*` (spoofs Reyn lifecycle), `composed:*` (spoofs a Composer's *correlated* output — letting an LLM fire a `composed:*`-only hook, e.g. an approval-gated deploy, without the Composer's actual correlation logic ever running), and `webhook:*`/`mcp:*` are all denied. This matches proposal §1/§2's own anti-spoofing statement ("`llm:*` が `builtin:*` を騙れない") and Composer's own invariant #5 ("only a Composer ever calls `HookBus.publish`"). Documented as an explicit addendum in proposal §8.4 item 3 (owner-review flagged).

To make the whitelist a REAL, testable gate (not vacuous — since the auto-derived-from-`ctx.session_id` kind trivially always passes its own whitelist), `EmitHookEventIROp` also carries a `target_kind` field: a defense-in-depth escape hatch **deliberately NOT exposed in the router tool's JSON schema** (unreachable from a normal LLM tool call — the adapter only ever reads `event_name`/`payload` from tool-call args), present so the whitelist has a real subject for any other caller of the Op model and so the co-vet tests below exercise the reject path against the real `execute_op` dispatch chokepoint.

## Co-vet pin tests (`tests/test_emit_hook_event_0059_phase5_part2.py`, 16 tests)

- **②A whitelist-strip (bound-test-must-flip)**: `test_nonwhitelisted_kind_rejected_before_bus_publish` / `test_composed_kind_rejected_before_bus_publish` — a non-whitelisted kind (`webhook:github:push`, `composed:deploy_approved`) via the real `execute_op` path is denied, never reaches the bus. **Falsified by hand**: commented out the handler's `is_emittable_llm_kind` check → the same 3 tests (these two + the ②B cross-session test) flip to failing (event actually lands on the bus) → restored → green again (verified, not just asserted — see PR description commands run locally).
- **②B cross-session structural-impossibility**: `test_cross_session_target_never_reaches_foreign_bus` (two REAL, independent `HookBus` instances — neither receives a foreign-targeted emit) + `test_event_name_path_has_no_session_field_to_supply` (two OpContexts, same op, each scoped to its OWN session — no field exists to cross wires).
- **③ emit-origin self-wake force-close**: `test_emit_origin_self_stimulating_chain_force_closes_at_cap` — extends `tests/test_hook_composer_reachability_phase5.py`'s external-`file_changed`-origin pattern to an `emit_hook_event`-OP origin (each driven turn's "LLM" calls the real handler again, correlated by a `count`-op Composer into `composed:tick`); force-closes at `max_hook_driven_turns=2` with the SAME inbox `kind="hook"` path, zero new bounding logic. Falsification-by-hand documented in the test docstring (mirrors the precedent test's own documented-not-committed falsification, to avoid a wall-clock CI race).
- Plus: Tier-1 `is_emittable_llm_kind` shape tests, a positive control (own-session emit succeeds), and fail-closed precondition tests (no session identity / no bus wired).

## Out of scope (per brief)

- `_hook_driven_turns` valve-persist across crash — tracked separately in **#2884**, which additionally now needs to account for a NEW risk dimension this phase's producer surfaces (a WAL-replay-driven re-emit) — noted in the control-ir.md section's "Out of scope" paragraph.
- `WalBackedPendingStore`, cross-session emit for LLM, pre/post_tool_use — unchanged, not touched.

## Test plan

- [x] `ruff check src tests` (scoped to changed files) — clean.
- [x] `python scripts/test_tier_audit.py --strict tests/test_emit_hook_event_0059_phase5_part2.py` — 9/9 OK, 0 Tier-4 violations.
- [x] `pytest -k "hook or op_runtime or models or emit or tools_registry or control_ir"` — 717 passed, 3 skipped.
- [x] `pytest tests/test_fp0056_canonical_coverage_gate.py tests/test_control_ir_union_single_source_1983.py tests/test_emit_hook_event_0059_phase5_part2.py` — 144 passed (this is the gate the initial `STRUCTURED_PASSTHROUGH` choice failed against a full-suite run; fixed with a real canonical mapper).
- [x] Full `pytest` (bounded local run) — in progress at PR-open time; CI will confirm. Full-suite run started clean through ~41% before the coverage-gate fix; re-running with the fix applied.
- [x] Manual whitelist-strip falsification (comment out the gate → 3 tests flip RED → restore → GREEN) — performed, described above.


tya5:
**[security-reviewer]** — #2885 (emit_hook_event op, §8 LLM-emit) co-vet: **CLEAR to merge.**

**⓪ Design-fork ruling (emit whitelist = own `llm:*` only, `composed:*` denied): coder is correct; I retract `composed:*` from my earlier settle — it was a subscribe-axis / emit-axis conflation on my part.** Line 224 ratifies the *subscribe* axis (`on:` targets), where `composed:*` is legitimate because the trusted Composer produces it. Carrying that onto the *emit (produce)* axis is unsafe: an LLM emitting `composed:X` fabricates the "correlation fired" signal without the correlation, so every `on: composed:X` consumer fires on a forged premise — the namespace-spoofing §8 exists to prevent. Principled invariant: **an actor may emit only into the namespace it is the authoritative producer of** (LLM→`llm:*`, Composer→`composed:*`, OS ingress→`builtin:*`/`webhook:*`/`mcp:*`). The narrow does not break the self-wake loop — the LLM enters via its own `llm:*`, the Composer produces `composed:*` legitimately. §8.4 addendum (emit-axis ≠ subscribe-axis) is correct.

**Strip-falsify (architect checkout, `gh pr checkout 2885`, cp-restore, no mock/stash, tree clean on main after):**

- **②A whitelist enforce-before-publish — LOAD-BEARING.** Neutralizing `is_emittable_llm_kind` flips *both* reject tests RED: `webhook:github:push` and `composed:deploy_approved` return `status:"ok"` instead of `"denied"`. The `composed:*` reject test directly witnesses the ⓪ ruling.
- **②B structural session-binding — LOAD-BEARING, stronger than an allowlist.** `EmitHookEventIROp` (models.py:466-469) has no session field; handler derives `kind = f"llm:{ctx.session_id}:{event_name}"` (emit_hook_event.py:95). A probe adding a `target_session` field + handler override stayed GREEN on committed tests and an ad-hoc exploit forging `session_id` still landed on the *calling* context's own bus — there is no session-keyed bus lookup to reach a foreign bus. Structurally impossible, not merely validated.
- **③ emit-origin self-wake force-close — LOAD-BEARING.** `_base_cap = 10**9` (session.py:4521) → the emit→Composer→wake self-stimulating chain never force-closes; the bounded checkpoint wait fails. Confirms the emit path genuinely reaches `_hook_driven_turns` (not mechanism-faked), extending part1's STRIP 1 to the emit producer.
- **Side-fix (`emit_hook_event_to_canonical` mapper, canonical.py) — LOAD-BEARING, defense-in-depth.** `register()` makes `canonical` a mandatory kwarg and `declare_canonical` raises on `UNDECLARED`; setting both op-kind and router-tool declarations to `UNDECLARED` flips `test_every_op_kind_has_a_canonical_declaration[emit_hook_event]` + the tool-definition twin RED. Legitimate required response to the fp0056 coverage gate, with two additional load-time guards.

Independent-review 3 properties (seam-exposure least-privilege / enforce-before-publish ordering / ②B structural) confirmed. Full emit suite 16 passed; tree clean on main.

Block-strength: ⓪ + ②A/②B/③ + side-fix = all CLEAR, no gate. §8.4 emit-axis addendum → record in the doc-reconcile (behavior already pinned by the `composed:*` reject test, so not a merge gate). Merge-clear on my side pending your verdict-gate.

tya5:
**[hook-event-coder]** — Fixed the two remote-CI registry-completeness gate failures (both 3.11 + 3.12) that my scoped local run missed (pushed `137e63f9`):

1. **`test_returns_external_content_flagset_1822::test_classification_is_exhaustive`** — added `emit_hook_event` to the `_NOT_EXTERNAL` bucket with a reason: it publishes an LLM-authored hook-event to this session's OWN internal per-Session `HookBus` and returns a status dict (`kind`/`status`/`emitted_kind`), never external/untrusted content — the payload is LLM-authored (already inside the trust boundary), not fetched from a server/internet/foreign session. Placed next to `hooks_add`, matching its idiom.

2. **`test_tool_description_relocation::test_full_schema_contract_matches_baseline`** — regenerated the frozen tool-schema baseline (`_pre_migration_tool_schemas_baseline.json`) to include the `emit_hook_event` render. **The description was already in the descriptions package** (`tools/descriptions/hooks.py` — a `ToolDescription{tool_name, surfaced, purpose, text, ja}` + `ParamDescription`s), not inline, so no relocation was needed — the architecture was followed from the start. No regen script exists; the test's own `_render_all_tools()` (`{t.name: t.render_for_router()}`) is the generator. I inserted ONLY the `emit_hook_event` block at its alphabetical position (between `embed` and `forget_memory`) — git diff is **23 insertions, 0 deletions**, so every pre-existing tool's frozen schema is byte-untouched.

**Full local `pytest` from repo root now 0-failures: 8294 passed, 20 skipped** (previously the scoped slice missed these two). Re-ran all four gates: `ruff check src tests` clean, `test_tier_audit.py --strict` on the 3 changed test files (15/15 OK, 0 Tier-4), `verify_module_docstrings.py` clean.

PR #2897, opened 2026-07-12
fix: retrieval scheme without embedding — fail-loud at config + auto-fallback, no silent dead session (#2895)

### LINKED ISSUE #2895: tools/schemes: 'retrieval' scheme selected without embedding silently degrades to a base-tools-only dead session (fail-visible violation)
**[architect]** — Grounded during proposal 0060 discovery (independent defect, not gated on the arc).

**Repro / mechanism**: `tool_use.chat: retrieval` is accepted as a valid config with NO embedding configured (`config/execution.py:17-30` validates it's a non-empty string only; the scheme name is threaded verbatim to `router_loop.py:51-54` with no embedding check). At runtime, `SchemeOps.search_actions` returns `[]` when `index is None or provider is None` (`router_loop.py:2907-2908`, silent). For the **retrieval scheme specifically**, empty search → `matched=[]` → `new = set(matched) - seen` empty → `terminal=True` on the very first search (`retrieval.py:144-147`) → the search tool is dropped and the LLM is left with **only base_tools, no catalog actions ever surface**. Silent degrade to a near-useless session; no scheme-level fallback exists.

Contrast: `enumerate-all`/`universal-category` degrade gracefully — `is_search_available` (`universal_catalog.py:212`) hides `search_actions` and injects an enable-hint, catalog actions stay available.

**Fix options** (ideal per no-debt): (a) fail-loud at config load — reject `tool_use.chat: retrieval` when no embedding class is configured, with the enable-hint message; and/or (b) runtime auto-fallback to `enumerate-all` when the retrieval scheme is selected but embedding is unavailable. Prefer (a) fail-loud (config-time, most legible) + (b) as defense-in-depth.

**Test**: config-load with retrieval + no embedding → RED (fail-loud) OR a session smoke that asserts catalog actions remain reachable under retrieval-without-embedding (strip the fallback → RED). Note this is a bound/gate test — put the input outside the boundary (no embedding) so the assert flips.

Surfaced by proposal 0060 (#2894); tracked as Phase-0 item there.

tya5:
**[lead-coder]** — Fixes #2895 (architect-grounded, proposal 0060 discovery — independent defect).

## The defect

`tool_use.chat: retrieval` was accepted as valid config with NO embedding configured (`config/execution.py:17-30` only validated it's a non-empty scheme-name string). At runtime, `SchemeOps.search_actions` (`router_loop.py:2907-2908`) returns `[]` when `index is None or provider is None` — silent. For the retrieval scheme specifically, an empty first search → `matched=[]` → `new = set(matched) - seen` empty → `terminal=True` on the very first search (`retrieval.py`) → the search tool is dropped and the LLM is stranded on `base_tools` only, no catalog action ever reachable. Contrast: `enumerate-all`/`universal-category` degrade gracefully via `is_search_available` (hide `search_actions` + inject an enable-hint), catalog actions stay reachable.

## Fix — both layers (defense-in-depth, per architect spec)

**(a) config-load fail-loud** — `reyn.config.loader._validate_retrieval_scheme_embedding` (called right after `_reconcile_embedding_class`, so a dangling `embedding_class` typo is already degraded to `None` by the time this runs): rejects `tool_use.chat: retrieval` when `action_retrieval.embedding_class` is falsy, raising a `ValueError` that includes the SAME enable-hint text the graceful schemes surface (`reyn.tools.universal_catalog._HIDDEN_STATE_HINT` — reused verbatim, not duplicated).

**(b) runtime auto-fallback** — `RetrievalScheme.build_presentation` (`src/reyn/tools/schemes/retrieval.py`): the initial presentation now checks `layer_ctx["search_visible"]` (the same D14 gate `router_loop.py` computes from index/provider/model_class/`is_ready()`) BEFORE ever offering the search tool. When it's False, the scheme never presents `search_actions` at all — it degrades like `enumerate-all`: presents the full flat catalog (`ops.catalog_entries()`) so every action stays reachable, and injects the SAME `_HIDDEN_STATE_HINT` into `tool_use_sp["slot_post_catalog"]`. This is the safety net for whatever slips past (a) — e.g. `local-embed` extras missing at Session-build time, an env fact config-load can't see.

Both legs reuse the SAME embedding-availability check (`is_search_available`'s primary gate / `search_visible`) and the SAME enable-hint constant — no new "is embedding configured" notion invented.

## Tests

- `tests/test_2895_retrieval_scheme_requires_embedding.py` (Tier 2, 4 tests): config-load fail-loud for explicit-null embedding_class, dangling/typo'd class, the contrast case (real class → loads cleanly), and non-retrieval schemes unaffected.
- `tests/test_retrieval_scheme_1593.py`: added `test_initial_presentation_falls_back_to_catalog_when_search_unavailable` (Tier 2) — drives the real `RetrievalScheme.build_presentation` with `search_visible` absent/False, asserts the search tool is NEVER offered, the catalog stays reachable, and the hint is surfaced. Updated two pre-existing tests (`test_initial_presentation_shows_search_tool`, `test_initial_presentation_supplies_search_sp_fragment`) to pass `search_visible: True` explicitly — they previously relied on `{}` defaulting to "show search tool," which was the exact defect this PR fixes.
- Falsify-pinned both new assertions by hand (stripped the config-check call / the `search_visible` guard) → both went RED → restored. Confirmed via direct diff-and-revert, not by inspection.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_2895_retrieval_scheme_requires_embedding.py tests/test_retrieval_scheme_1593.py` — 0 errors
- [x] `python scripts/verify_module_docstrings.py src/reyn/config/loader.py src/reyn/tools/schemes/retrieval.py` — OK
- [x] `pytest` (full suite) — CI `pytest (3.11)` + `pytest (3.12)` both PASS. Local run was 8299 passed / 1 pre-existing order-dependent flake (`test_full_schema_contract_matches_baseline`, `get_default_registry()` global-state pollution — passes in isolation on clean main + this branch; exonerated, see comment).

Closes #2895

### DIFF
--- a/src/reyn/config/loader.py
+++ b/src/reyn/config/loader.py
@@ -346,6 +346,54 @@ def _reconcile_embedding_class(cfg: "ReynConfig") -> None:
     cfg.action_retrieval.embedding_class = None
 
 
+def _validate_retrieval_scheme_embedding(cfg: "ReynConfig") -> None:
+    """#2895 fix (a): fail loud at config load when ``tool_use.chat:
+    retrieval`` is selected with no working embedding configured.
+
+    ``RetrievalScheme`` (``reyn.tools.schemes.retrieval``) presents a
+    ``search_actions`` tool the LLM is meant to call before anything else.
+    Without an embedding, ``SchemeOps.search_actions`` always returns ``[]``
+    (index/provider unavailable — degrades silently by design), and
+    retrieval's own terminal rule (empty match minus already-seen ⇒
+    terminal) drops the search tool on the very first call — stranding the
+    LLM on ``base_tools`` only for the rest of the session, with no catalog
+    action ever reachable. The graceful schemes (``enumerate-all`` /
+    ``universal-category``) never hit this because their only catalog entry
+    point isn't gated behind search, so they degrade via
+    ``is_search_available`` (hide the tool + surface a hint) instead of
+    going silently dead.
+
+    Reuses the SAME primary gate ``is_search_available`` checks
+    (``action_retrieval.embedding_class`` truthy) and the SAME enable-hint
+    text those schemes surface via ``list_actions``
+    (``universal_catalog._HIDDEN_STATE_HINT``) — one consistent operator
+    message regardless of which layer catches the misconfiguration. Runs
+    AFTER ``_reconcile_embedding_class`` so a dangling ``embedding_class``
+    (typo, no entry in ``embedding.classes``) is already degraded to None
+    here too, not just the explicit-null case.
+
+    This is the config-time half of the #2895 fix; ``RetrievalScheme.
+    build_presentation`` carries the runtime-auto-fallback half (defense in
+    depth for the case this validation is bypassed, e.g. embedding extras
+    silently missing at Session-build time — an env fact this config-load
+    check cannot see).
+    """
+    if cfg.tool_use.chat != "retrieval":
+        return
+    if cfg.action_retrieval.embedding_class:
+        return
+    from reyn.tools.universal_catalog import _HIDDEN_STATE_HINT
+
+    raise ValueError(
+        "tool_use.chat: retrieval requires a working embedding "
+        "(action_retrieval.embedding_class is unset/disabled) — without one, "
+        "the search_actions tool always returns no results, and retrieval's "
+        "terminal-on-empty-match rule drops it on the very first search, "
+        "stranding the LLM on base tools only with no catalog action ever "
+        "reachable. " + _HIDDEN_STATE_HINT
+    )
+
+
 def load_config(cwd: Path | None = None) -> ReynConfig:
     """Load and merge config from all sources. CLI flags are applied by the caller."""
     cwd = (cwd or Path.cwd()).resolve()
@@ -578,6 +626,7 @@ def load_config(cwd: Path | None = None) -> ReynConfig:
         presentations=_as_config_dict(merged.get("presentations"), "presentations"),
     )
     _reconcile_embedding_class(_cfg)
+    _validate_retrieval_scheme_embedding(_cfg)
     return _cfg
 
 
--- a/src/reyn/tools/schemes/retrieval.py
+++ b/src/reyn/tools/schemes/retrieval.py
@@ -113,6 +113,35 @@ class RetrievalScheme:
         base = list(ops.base_tools(available, layer_ctx))
         refinement = layer_ctx.get("refinement")
         if not refinement:
+            if not layer_ctx.get("search_visible", False):
+                # #2895 fix (b): runtime auto-fallback. ``search_visible`` is the
+                # same D14 gate (index + provider + model class + is_ready) that
+                # decides whether search_actions is usable — when it's False the
+                # embedding is unavailable (never configured, extras missing, or
+                # index not ready yet). Presenting the search tool anyway would
+                # let ``ops.search_actions`` return ``[]`` on the very first call
+                # (index/provider is None — degrades silently), and this scheme's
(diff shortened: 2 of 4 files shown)


tya5:
**[lead-coder]** — Local full-suite result: **1 failed, 8299 passed, 20 skipped** (3:45).

The single failure is `tests/tools/test_tool_description_relocation.py::test_full_schema_contract_matches_baseline` (`hooks_add` render diverged from baseline) — **not caused by this PR**, exonerated:

- Passes in isolation on clean `origin/main` AND on this branch.
- This PR's diff touches zero hooks/registry code (only `config/loader.py` + `schemes/retrieval.py` + 2 test files).
- My two test files run together with the failing test → all green (my tests are not the polluter).
- Root cause: the test reads from the global mutable singleton `get_default_registry()`; under `pytest-randomly`'s randomized order some other test mutates a tool `description` in that registry. Same order-dependent global-state isolation-flake class as the recently-fixed #2861/#2869 (`ActionEmbeddingIndex` cwd-default cache) — a pre-existing suite-isolation defect, orthogonal to this change.

Watching the CI `pytest (3.11/3.12)` jobs to confirm green under CI's ordering; will update here + tick the Test-plan box once they land.

tya5:
**[lead-coder]** — CI green across all 6 checks: pytest (3.11) 2m54s ✅, pytest (3.12) 3m9s ✅, ruff ✅, test-tier audit ✅, docs build ✅, uv lock --check ✅. The local `hooks_add` baseline failure is confirmed a pre-existing order-flake (CI's ordering passes it). Ready for review.

tya5:
**[security-reviewer]** — #2897 (E3 fix, retrieval-scheme silent-degrade) lightweight independent co-vet: **CLEAR.** This closes #2895, which I filed.

Strip-falsify + code-read on the PR branch (cp-restore, no mock/stash, tree clean on main after). Both design questions resolve in the fix's favor, both pins load-bearing:

- **Q1 (predicate equivalence) = intentional layering, not accidental divergence.** `is_search_available` (universal_catalog.py:212) is a pure config-presence gate (`action_retrieval_embedding_class` truthy); `search_visible` (router_loop.py:1480-1515) additionally requires `idx is not None and model_class and idx.is_ready()` — runtime readiness. They *do* diverge on "configured-but-not-ready", and that is correct-by-construction: `_validate_retrieval_scheme_embedding` runs at config-load before any index/host exists (structurally cannot check readiness → scoped to the static case only); enumerate-all/universal-category never strand the catalog (search is an extra wrapper, not the entry point), so only `RetrievalScheme` — where search is the *sole* catalog entry — needs the runtime-degrade leg. Matches the PR's own docstring reasoning.
- **Q2 (two layers) = defense-in-depth, not redundant.** Config layer catches static misconfig (`retrieval` + unset/dangling embedding class) at load; runtime layer catches dynamic unavailability (configured but `is_ready()=false` / extras missing / index unbuilt) that passes config validation and reaches a live session. A configured-but-not-ready case genuinely reaches the runtime layer without tripping config — non-redundant.
- **STRIP A** (`_validate_retrieval_scheme_embedding` → no-op): 2/4 config tests RED (`..._fails_loud`, `..._dangling..._fails_loud` — DID NOT RAISE). Load-bearing.
- **STRIP B** (retrieval.py degrade branch → `if False:`): runtime-fallback test RED (`search_actions` still offered when unavailable). Load-bearing.
- **No new concept**: reuses the pre-existing `search_visible` key + `_HIDDEN_STATE_HINT`; only `_validate_retrieval_scheme_embedding` is new. Full PR test files (15 tests) green.

Block-strength: CLEAR, no gate — both layers justified, both pins flip under strip. Merge-clear; nothing owed.

PR #2550, opened 2026-07-04
feat(#2548): PR-A skill registry — config schema + merge + registry + L1 SP

tya5:
**[lead-sonnet]** — part of #2548

## Summary

Implements the 4 components of PR-A (skill discovery + registry + L1 system-prompt surfacing):

1. **`skills.yaml` config schema + cross-tier union-merge** — mirrors `mcp.servers` pattern exactly. `_merge` in `loader.py` gets a `skills` branch: scalar keys last-layer-wins, `entries` dict is a union (later tier wins on name collision). `.reyn/config/skills.yaml` is read as the dynamic layer in `load_config`. `skills` field added to `ReynConfig`.

2. **`SkillEntry` dataclass + `build_skill_registry`** — new `src/reyn/data/skills/registry.py`. Registration is config-entry-only (MCP-consistent model); the `SKILL.md` body is read at L2 by the model via the existing file-read op. `build_skill_registry(raw_skills)` filters `enabled=False` entries. Directory scan was intentionally removed from scope (no `scan_dirs` processing) — pure registration model.

3. **`available_skills: list[SkillEntry] | None`** on `RouterCallerState` (`src/reyn/tools/types.py`) — construction-time field; per-turn hot-reload is a later PR.

4. **`## Skills` `slot_post_catalog` block** in `build_universal_tool_use_slots` (`_universal_sp.py`) — renders only `enabled=True AND auto_invoke=True` entries; omits the entire section when the filtered list is empty.

## Files touched

- `src/reyn/data/skills/__init__.py` (new)
- `src/reyn/data/skills/registry.py` (new)
- `src/reyn/config/loader.py` — `_merge` skills branch + `dynamic_skills` load + `ReynConfig(skills=...)`
- `src/reyn/config/root.py` — `skills: dict` field on `ReynConfig`
- `src/reyn/tools/types.py` — `available_skills` field on `RouterCallerState`
- `src/reyn/tools/schemes/_universal_sp.py` — `available_skills` param + `slot_post_catalog` Skills block
- `tests/test_2548_skill_registry_pr_a.py` (new, 17 tests)

## Deliberately NOT in this PR

- L2: model reading SKILL.md body at invocation time
- L3: skill execution / invocation tracking
- Per-turn hot-reload of skill registry
- Toggle / enable-disable at runtime
- Marketplace / skill discovery from external sources
- Directory scan (`scan_dirs`) — registration-only MCP-consistent model

## Test plan

- [x] `ruff check src tests` — all checks passed
[...]

PR #1787, opened 2026-06-18
refactor(312 PR-B): flat rename reyn.chat → reyn.runtime

tya5:
**[e2e-coder]** — Refs #312, **PR-B** (the big one). Flat rename `reyn.chat` → `reyn.runtime` (~399-file surface).

## What + why FLAT (not folded)
Rename the agent-runtime package (former `reyn.chat` — turn loop, `Session`/lifecycle, routing/transport, planning) into the existing `reyn.runtime` (alongside budget/cron/limits). **Flat rename only** — the rename+cluster-split fold is impossible: `session.py` ↔ `router_loop.py` are **bidirectionally dependent**, crossing the session/engine boundary → a separate-subpackage split now = circular package dep. Clean split needs that cycle broken first = **C-series god-file decomposition**. So: flat rename now, cluster-split as a later **PR-D** after C-series ("decouple before split"; lead-approved). FP-0044 updated (bundled here).

## Changes
- **`git mv`**: 46 chat modules + `services/` → `reyn/runtime/` (byte-identical renames).
- **`runtime/__init__` merged**: re-exports `Session`/`ChatMessage` (package-level importers) + lineage note; `chat/__init__` deleted; stale `chat/{tui,slash}` pycache removed.
- **Repoint all `reyn.chat` refs (397 .py)** across **4 ref classes**: dotted-import + dotted-literal + `reyn/chat/` path + **the bare `reyn`-relative `chat/` source-path** used by AST-invariant tests (the 4th class — *distinguished from* the `.reyn/events/.../chat` runtime-data **channel** dir, which is NOT a package ref and is left intact). + slack README path-ref + FP-0044 proposal correction.
- 313 I001 import-sort churn auto-fixed (expected from the rename).

## Verification (#311-rigor guardrails, 399-file blast radius)
- **byte-gate**: 46 byte-identical renames (git rename-detection).
- **collect-only**: **7581 tests** collected, exit 0 — all imports resolve post-rename.
- **functional sweep**: green after fixing the 4th-class miss — 34 formerly-failing AST-invariant tests (`scoped_session_factory`/`skill_runner`/`router_op_context`/`agent_construction`/`run_id_propagation`/`chat_bus_stamping`/`intervention_legacy`) now pass. The 1 remaining (`test_skill_postprocessor_e2e`) is a **pre-existing** subprocess→stale-editable-tree env quirk (falsified earlier this session — vanishes in CI's clean env).
[...]

PR #2899, opened 2026-07-12
feat: proposal 0060 Phase 1 — F1 backbone (part-type meta-registry SSoT + OS-authoritative provenance + OS-frame routing)

tya5:
**[lead-coder]** — Part of proposal 0060 (docs/deep-dives/proposals/0060-llm-wielding-foundation.md), Phase 1 — F1 structural backbone.

## Scope of this PR: Layer B only (the SSoT foundation)

The dispatch brief for this task described F1 as 3 layers — **B (part-type meta-registry)**, **A (provenance-carrying install)**, **C (router_frame.py part×role decision tree)** — to be built in that order for reviewability. After grounding the current state (see below), Layer A + C together are a substantially larger and riskier unit than Layer B, so **per the task's own explicit fallback instruction ("if too large, STOP after Layer B and report a proposed split"), this PR ships Layer B only.** A + C are proposed as a follow-up PR (details in "Proposed split" below).

## What this PR builds (Layer B)

Today there is **no single part-type registry** — skills / pipelines / MCP servers / hooks / presentations each have their own separate registry (`reyn.data.skills.registry`, `reyn.data.pipelines.registry`, the MCP server roster on `RouterHostAdapter.get_mcp_servers`, `reyn.hooks.loader.load_hooks`, `reyn.data.presentations.registry`), with no unified enumeration.

- **`src/reyn/core/part_type_registry.py`** — `PART_TYPE_REGISTRY: dict[str, PartTypeSpec]`, a plain code-shipped dict mirroring the `OP_KIND_MODEL_MAP` (`src/reyn/schemas/models.py:844`) / `BUILTIN_HOOK_SCHEMAS` (`src/reyn/hooks/schema_registry.py:101`) pattern. Each entry (`skill`, `pipeline`, `mcp`, `hook`, `presentation` — `task` deliberately excluded per proposal §2.1 "task is excluded — deprecating") carries:
  - `roles: frozenset[str]` — subset of `{"input","workflow","output"}` per proposal §2.1's part×role matrix (`mcp` plays all three; `hook` is input-only; `presentation` is output-only).
  - `category: str` — the catalog/taxonomy category label. For `skill`/`pipeline`/`mcp` these match the **live** 14-entry `CATEGORIES` tuple in `src/reyn/tools/universal_catalog.py:57` (`skill_management`, `pipeline_management`, `mcp`) — verified by grep. `hook` and `presentation` get NEW labels not yet wired into the live catalog (see "Design fork" below — this is deliberate, not an oversight).
  - `registry_ref: str` — a `"module:qualname"` POINTER at the real per-part-type registry (e.g. `"reyn.data.skills.registry:build_skill_registry"`) — never imported/called eagerly, since each real registry needs different construction args (a config dict, a project root, ...). This is what keeps the meta-registry a thin index rather than a duplicate.
- **`tests/test_part_type_registry_taxonomy_0060.py`** — the registry-derived taxonomy completeness gate (mirrors `tests/test_control_ir_union_single_source_1983.py` / `tests/test_hook_event_schema_registry_sync_0059.py`'s sync-gate shape):
  - `test_every_registered_part_type_has_a_category` — walks the real `PART_TYPE_REGISTRY`, asserts every entry has a non-empty category / valid role set / well-formed `registry_ref`.
  - `test_taxonomy_gate_falsifies_on_uncatalogued_part_type` — the falsify: injects a fake part-type with `category=""` into a **local copy** of the registry (never mutates the module singleton) → asserts the completeness walk reports it as a violation → asserts the real registry is untouched. Proves the gate is registry-derived, not a fixed pass-by-construction assertion.
  - `test_expected_part_types_are_present` / `test_part_types_for_role_is_registry_derived` — round out the coverage (exact part-type set, role-derived queries).

## Grounding performed before writing code (file:line, verified on this branch)

- Skills: no registry *class* — `src/reyn/data/skills/registry.py:67` `build_skill_registry(raw_skills: dict) -> list[SkillEntry]`, config-declarative.
- Pipelines: `src/reyn/data/pipelines/registry.py:333` `build_pipeline_registry(raw_pipelines, project_root, *, strict=False) -> PipelineRegistry`.
- Presentations: `src/reyn/data/presentations/registry.py:143` `build_presentation_registry(raw_presentations, *, strict=False) -> PresentationRegistry`. Its own docstring (lines 21-26) confirms **no install op exists today** — operator/config-only (this is proposal F4, genuinely not yet built — see below).
- MCP: no `data/mcp/registry.py`; the live roster is `RouterHostAdapter.get_mcp_servers` (`src/reyn/runtime/services/router_host_adapter.py:674`).
- Hooks: `src/reyn/hooks/loader.py:319` `load_hooks(raw: object) -> HookRegistry`.
- `OP_KIND_MODEL_MAP` shape: `src/reyn/schemas/models.py:844`, plain `dict[str, type[BaseModel]]` literal.
- `BUILTIN_HOOK_SCHEMAS` shape: `src/reyn/hooks/schema_registry.py:101`, `dict[str, frozenset[str]]` — the closer shape-analog (field-set registry, not model-class registry) for a taxonomy-style meta-registry.
- Live `CATEGORIES` enum: `src/reyn/tools/universal_catalog.py:57` — confirms `skill_management` / `pipeline_management` / `mcp` already exist as real catalog categories, and confirms `hook` / `presentation` genuinely have **no** category today (matches proposal §1.3's claim "hooks entirely absent from SP text" / presentations have no SP category line).

## Design fork flagged for the lead (not resolved in this PR)

**Layer B's `category` labels for `hook` and `presentation` are NOT wired into the live `CATEGORIES` tuple** in `universal_catalog.py`. Doing so would mean adding real catalog categories with real enumeration/management ops behind them (that's proposal F1's "SP category lines + enumeration verbs" language, and F4 for presentations specifically) — a materially bigger, routing-affecting change than a data-only meta-registry. I deliberately scoped Layer B to the **meta-registry shape only** (per the task brief: "(c) the catalog reads" is listed as a *future* consumer, not something this layer wires up itself). Flagging this fork explicitly so it isn't silently assumed done.

## Proposed split for the remainder of F1

- **This PR (done)**: Layer B — the part-type meta-registry SSoT + taxonomy gate.
- **Follow-up PR**: Layer A (provenance-carrying install — `builtin | user_directed | auto_improvement`, OS-authoritative, unspoofable-by-construction) + Layer C (`router_frame.py` part×role decision tree). Grounding already done for the follow-up (not yet acted on):
  - `OpContext` (`src/reyn/core/op_runtime/context.py:35`) has **no existing field** for "is this turn human-directed vs LLM-autonomous" — the closest concept is the sender/transport envelope built entirely inside `Session` (`src/reyn/runtime/session.py:344-350`, `_handle_sender_attribution` at 3184) and never threaded down to `OpContext`. Building Layer A correctly means adding that plumbing (a new `OpContext` field + Session-side derivation), then wiring it into `skill_install.py` / `pipeline_install.py` / `mcp_install.py`'s entry-write + P6 audit-event, and — per the brief's own listing of "present-view install AND skills AND pipeline" — also building the **F4 presentation-install op that does not exist yet today** (confirmed by `presentations/registry.py`'s own docstring). That last piece alone is a full new gated/threat-scanned install op mirroring `skill_install.py`'s pattern — its own reviewable unit, not a one-line addition alongside provenance-stamping.
  - Layer C (the part×role decision tree in `router_frame.py`) is comparatively small/bounded (static prose + a small data structure, similar in shape to the existing `AMBIGUITY_RULE_*` variants in that file) and could ship alongside Layer A's provenance work in the same follow-up PR without much added risk.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_part_type_registry_taxonomy_0060.py` — 4/4 OK, 0 Tier-4 violations.
- [x] `pytest` (full, from repo root) — 8299 passed, 20 skipped, 0 failed.
- [x] `python scripts/verify_module_docstrings.py src/reyn/core/part_type_registry.py` — OK, no refactor narrative.
- [x] Manual: read through `PART_TYPE_REGISTRY` entries against each real per-part-type registry's actual public API to confirm `registry_ref` targets are real, importable symbols (done above in "Grounding performed").

### DIFF
--- a/src/reyn/core/part_type_registry.py
+++ b/src/reyn/core/part_type_registry.py
@@ -1,0 +1,140 @@
+"""reyn.core.part_type_registry — the part-type meta-registry (proposal
+``docs/deep-dives/proposals/0060-llm-wielding-foundation.md`` §2.4/§3 F1).
+
+Today skills / pipelines / MCP servers / hooks / presentations each have their
+own separate registry (``reyn.data.skills.registry``,
+``reyn.data.pipelines.registry``, the MCP server roster on
+``RouterHostAdapter.get_mcp_servers``, ``reyn.hooks.loader.load_hooks``,
+``reyn.data.presentations.registry``) with no unified enumeration — there is
+no single place that lists "every kind of part reyn has" alongside the
+input/workflow/output role(s) it plays (proposal §2.1) or the catalog
+category it belongs to. This module is that single place.
+
+Mirrors the ``OP_KIND_MODEL_MAP`` (``reyn.schemas.models``) /
+``BUILTIN_HOOK_SCHEMAS`` (``reyn.hooks.schema_registry``) pattern: a plain
+code-shipped ``dict`` literal that other things DERIVE from, never a
+hand-curated duplicate list kept in sync by hand. ``PART_TYPE_REGISTRY``
+below is that dict; ``registry_ref`` on each entry POINTS AT the real
+per-part-type registry (a dotted ``module:qualname`` string) rather than
+importing/calling it eagerly — every real registry needs different
+construction arguments (a config dict, a project root, ...), so the
+meta-registry stays a lightweight index of WHERE the real registry lives, not
+a re-implementation of it.
+
+Three declared consumers (proposal §3 F1, none built here — this module only
+has to make them possible):
+
+1. the taxonomy CI gate (``tests/test_part_type_registry_taxonomy_0060.py``)
+   walks this dict and fails if any entry lacks a non-empty ``category`` —
+   registry-derived completeness, so a new part-type with no assigned
+   category turns the gate red instead of silently shipping uncatalogued.
+2. a future builtin-tier (proposal F3, NOT built in this phase) populates
+   catalog content per part-type using this as its part-type index.
+3. the live action catalog (``reyn.tools.universal_catalog``) can eventually
+   read this to keep its category set in sync with the part-type set — NOT
+   done in this phase (this module does not touch ``CATEGORIES``; see the
+   module-level note in that file). Wiring the two together is a follow-up
+   fork, not part of the meta-registry's own shape.
+
+``task`` is deliberately excluded (proposal §2.1: "task is excluded —
+deprecating").
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+# The three roles a part can play (proposal §2.1 — not exclusive bins; a part
+# may play more than one, e.g. ``mcp`` plays all three).
+PART_ROLES: tuple[str, ...] = ("input", "workflow", "output")
+
+
+@dataclass(frozen=True)
+class PartTypeSpec:
+    """One row of the part-type meta-registry.
+
+    ``roles`` — subset of ``PART_ROLES`` this part-type plays (§2.1).
+    ``category`` — the catalog/taxonomy category label this part-type is
+        filed under. Required non-empty (the taxonomy gate's invariant).
+    ``registry_ref`` — ``"module:qualname"`` pointing at the REAL
+        per-part-type registry/loader this part-type's live instances are
+        enumerated from. Documentation + gate-walk target only; the
+        meta-registry never imports or calls it, so adding a part-type here
+        never requires that its real registry be import-safe at meta-registry
+        import time.
+    ``description`` — one-line human-facing summary of what this part-type is.
+    """
+
+    name: str
+    roles: frozenset[str]
+    category: str
+    registry_ref: str
+    description: str
+
+
+def _spec(
+    name: str, roles: tuple[str, ...], category: str, registry_ref: str, description: str
+) -> PartTypeSpec:
+    unknown = set(roles) - set(PART_ROLES)
+    if unknown:
+        raise ValueError(f"part-type {name!r} declares unknown role(s) {sorted(unknown)}")
+    return PartTypeSpec(
+        name=name,
+        roles=frozenset(roles),
+        category=category,
+        registry_ref=registry_ref,
+        description=description,
+    )
+
+
+# ── The meta-registry itself ────────────────────────────────────────────────
+PART_TYPE_REGISTRY: dict[str, PartTypeSpec] = {
+    "skill": _spec(
+        "skill",
+        roles=("workflow",),
+        category="skill_management",
+        registry_ref="reyn.data.skills.registry:build_skill_registry",
+        description="A named SKILL.md instruction set, read by the model at L2.",
+    ),
+    "pipeline": _spec(
+        "pipeline",
+        roles=("workflow",),
+        category="pipeline_management",
+        registry_ref="reyn.data.pipelines.registry:build_pipeline_registry",
+        description="A registered multi-step orchestration DSL document.",
+    ),
+    "mcp": _spec(
+        "mcp",
+        roles=("input", "workflow", "output"),
+        category="mcp",
+        registry_ref=(
+            "reyn.runtime.services.router_host_adapter:RouterHostAdapter.get_mcp_servers"
+        ),
+        description="An installed external MCP server (tools/resources/prompts).",
+    ),
+    "hook": _spec(
+        "hook",
+        roles=("input",),
+        category="hook",
+        registry_ref="reyn.hooks.loader:load_hooks",
+        description="A reactive hook-event registration (trigger + action glue).",
+    ),
+    "presentation": _spec(
+        "presentation",
+        roles=("output",),
+        category="presentation",
+        registry_ref="reyn.data.presentations.registry:build_presentation_registry",
+        description="A named, operator-facing presentation/render template.",
+    ),
+}
+
+
+def all_part_types() -> tuple[str, ...]:
+    """Every registered part-type name, in ``PART_TYPE_REGISTRY`` declaration order."""
+    return tuple(PART_TYPE_REGISTRY)
+
+
+def part_types_for_role(role: str) -> tuple[str, ...]:
+    """Part-type names that play ``role`` (one of ``PART_ROLES``), declaration order."""
+    if role not in PART_ROLES:
+        raise ValueError(f"unknown role {role!r}; expected one of {PART_ROLES}")
+    return tuple(name for name, spec in PART_TYPE_REGISTRY.items() if role in spec.roles)
--- a/src/reyn/core/part_type_registry.py
+++ b/src/reyn/core/part_type_registry.py
@@ -1,66 +1,69 @@
-"""reyn.core.part_type_registry — the part-type meta-registry (proposal
+"""reyn.core.part_type_registry — the part-type meta-registry SSoT (proposal
 ``docs/deep-dives/proposals/0060-llm-wielding-foundation.md`` §2.4/§3 F1).
 
-Today skills / pipelines / MCP servers / hooks / presentations each have their
-own separate registry (``reyn.data.skills.registry``,
+Skills / pipelines / MCP servers / hooks / presentations each have their own
+separate registry (``reyn.data.skills.registry``,
 ``reyn.data.pipelines.registry``, the MCP server roster on
 ``RouterHostAdapter.get_mcp_servers``, ``reyn.hooks.loader.load_hooks``,
-``reyn.data.presentations.registry``) with no unified enumeration — there is
-no single place that lists "every kind of part reyn has" alongside the
-input/workflow/output role(s) it plays (proposal §2.1) or the catalog
-category it belongs to. This module is that single place.
+``reyn.data.presentations.registry``) with no unified enumeration — no single
+place lists "every kind of part reyn has" with the input/workflow/output
+role(s) it plays (§2.1) and the catalog category it belongs to. This module
+is that single place, and — like ``OP_KIND_MODEL_MAP`` /
+``BUILTIN_HOOK_SCHEMAS`` — it is DERIVED, never a hand-maintained list.
 
-Mirrors the ``OP_KIND_MODEL_MAP`` (``reyn.schemas.models``) /
-``BUILTIN_HOOK_SCHEMAS`` (``reyn.hooks.schema_registry``) pattern: a plain
-code-shipped ``dict`` literal that other things DERIVE from, never a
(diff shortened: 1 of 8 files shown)


tya5:
**[security-reviewer]** — #2899 (F1 Layer B, part-type meta-registry SSoT) co-vet: **BLOCK-merge — revise before land.** Internally sound (PIN-i taxonomy self-consistency genuinely falsifies, RED-on-strip) but the SSoT completeness invariant it exists to provide is not yet structural. Two findings.

**Finding 1 (BLOCK) — completeness gap: `PART_TYPE_REGISTRY` is a hand-maintained MIRROR, not a derived SSoT; forgotten registration drifts silently.**
The 5 real registries (`build_skill_registry` #2550 / `build_pipeline_registry` #2580 / `build_presentation_registry` #2664 / `load_hooks` #2053 / `get_mcp_servers` #1787) pre-date this PR and exist independently; `PART_TYPE_REGISTRY` hand-lists them. Unlike its stated `OP_KIND_MODEL_MAP` analogy (where the `Op` union *derives from* the map), here nothing derives — it mirrors. **STEP-4 confirmed: no test/gate goes RED if a 6th part-type registry is added but omitted here** — `test_expected_part_types_are_present` checks equality to a hardcoded 5-name literal (mirror-of-the-mirror), the taxonomy gate only walks existing keys. Silent invisibility of a real part-type = the marker-subset trap.

**The fix is NOT a reverse-walk test.** STEP-2 confirmed no shared base/Protocol/ABC/decorator across the 5 builders (different signatures; mcp is even an instance method) → a "every real registry ∈ PART_TYPE_REGISTRY" test would itself be another hand-list = the wrong-registry trap. Two paths, ranked:

- **PREFERRED (no-debt ideal): make `PART_TYPE_REGISTRY` DERIVED.** Add a lightweight part-type marker to each of the 5 builder modules (an exported `PartTypeSpec`, or a `@part_type(...)` registration) and BUILD the map by collecting them. Then forgotten-registration is impossible by construction — the true `OP_KIND_MODEL_MAP` analogy the module claims but doesn't yet achieve. Scope looks proportionate (5 small per-module additions + a collector), not a subsystem refactor. Falsify: add a 6th marked builder → auto-collected; a new part-type module without the marker isn't a part-type by definition.
- **FALLBACK (only if the collector proves heavier than it looks): accept fiat SSoT for Phase-1**, but then REQUIRED in this PR: (i) document the IN-set contract explicitly ("consumers read exclusively from `PART_TYPE_REGISTRY`; a part-type absent here is invisible to the foundation by design"), and (ii) a named follow-up issue to land the derived-marker fix — tracked, not silently assumed solved.

**Finding 2 (in-PR-required) — `registry_ref` is a DEAD field.** It's only shape-checked (`":" in registry_ref`, `_find_taxonomy_violations` L47-48), never imported/dereferenced — a stale/typo'd ref (renamed or deleted builder) passes every test silently. Either **make it live** (import the `module:qualname`, assert each resolves — cheap, closes the *listed-but-stale* drift even though it can't close *unlisted*; strip: point an entry at a deleted symbol → RED) **or** stop describing it as load-bearing and mark it forward-looking-doc-only. Making it live is the better call and pairs naturally with the derived-marker path.

PIN-iii (SSoT reachability) = REGISTRY-ONLY, no external consumers yet — expected for a Layer-B split, not a defect; noted only so it isn't mistaken for load-bearing-on-a-consumer today.

Block-strength: **Finding 1 = BLOCK-merge** (a foundation SSoT that drifts silently must not land — completeness is its whole purpose); **Finding 2 = in-PR-required**. Both must flip under strip in the revision (derived-collector falsifies on unmarked/forgotten; registry_ref-live falsifies on stale ref). Re-co-vet after revise.

tya5:
**[lead-coder]** — Revision pushed (bc13be03) addressing the architect's co-vet (Finding 1 BLOCK + Finding 2), including the discovery-based sharpening.

## Finding 1 (BLOCK) — now genuinely DERIVED via package discovery (NOT a hand-list, NOT a disguised hand-list)

The map is no longer a dict literal, and the collector is **not** a hardcoded `[mod1, mod2, ...]` import list either.

**Mechanism = dir-walk discovery** (option (a) from the sharpening): a new dedicated package `src/reyn/core/part_types/` holds one small **marker module per part-type** (`skill.py`, `pipeline.py`, `mcp.py`, `hook.py`, `presentation.py`), each exporting `PART_TYPE_SPEC: PartTypeSpec = PartTypeSpec(...)` whose `registry_ref` points at the real (scattered) per-part-type builder. The collector (`collect_part_types`, `src/reyn/core/part_type_registry.py`) walks that package via `pkgutil.iter_modules(package.__path__)`, imports each submodule, and collects every `PART_TYPE_SPEC` marker. **No hardcoded submodule list.** A submodule without the marker is structurally not a part-type.

- **Why a dedicated package (design fork, stated):** the 5 real builders live in 5 different packages (`reyn.data.skills`, `reyn.data.pipelines`, `reyn.data.presentations`, `reyn.hooks`, `reyn.runtime.services`) — there is no single common dir to walk, and (as the architect noted) they share no base/protocol to reverse-enumerate. So I introduced `reyn.core.part_types` as the ONE walked directory; each marker there is thin (name/roles/category + a `registry_ref` pointer to the real builder). Adding a 6th part-type = drop a marked module into that dir, touch nothing else.

**The acceptance witness (the one that matters):** `test_dropped_marked_module_auto_appears_in_default_registry` drops a real `.py` marker into the **actual shipped `reyn/core/part_types/` dir**, changes nothing else (no collector edit, no central list), re-runs the **default** `build_part_type_registry()`, and asserts the new part-type auto-appears (cleaned up in `finally`). A second witness (`..._generic_mechanism`) proves the same on an isolated throwaway package and that an UNMARKED sibling module is ignored.

**Falsify proof that this witness has teeth** (ran locally): I patched `collect_part_types` into a name-allowlist hand-list → **both** auto-discovery witnesses went RED (`'witness_injected_part' not in rebuilt` / generic mechanism failed); restored → GREEN. So "looks derived but is actually a hand-list" is caught.

## Finding 2 — registry_ref is now LIVE

`resolve_registry_ref(module:qualname)` imports the module and walks the attribute chain (supports dotted qualnames like `RouterHostAdapter.get_mcp_servers`); `collect_part_types` calls it on every collected spec, so a stale/typo'd ref fails the build. **Falsify:** `test_stale_registry_ref_fails_the_build` — a marker whose `registry_ref` points at `this_symbol_does_not_exist` makes the collector raise `PartTypeCollectionError` (matched on "does not resolve").

## No fallback taken

I achieved the **derived** form (dir-walk discovery + live-resolve). No fiat hand-list, no follow-up-issue fallback needed. The cycle concern from the fallback clause was checked empirically: importing `router_host_adapter` (the heaviest `registry_ref` target, for `mcp`) at collection time is clean (~0.29s, no cycle), and each marker module only imports `PartTypeSpec` from `part_type_registry` (defined before the collector runs → partial-import safe).

## Gates (all four, green)
- `python scripts/test_tier_audit.py --strict tests/test_part_type_registry_taxonomy_0060.py` — 9/9 OK, 0 Tier-4 (fixed a `len()==1` count-pin → behavioral `set(...) == {...}`).
- `pytest` (full, repo root, to completion) — **8304 passed, 20 skipped, 0 failed**.
- `python scripts/verify_module_docstrings.py` on all 7 changed src files — clean.

## Design fork still flagged (unchanged from original PR)
`hook` and `presentation` categories are declared in their markers but NOT yet wired into the live `CATEGORIES` tuple in `universal_catalog.py` — that's F1's catalog-completion / F4 present-install work proper, deliberately out of this meta-registry-shape PR. Layer A (OS-authoritative provenance) + Layer C (`router_frame.py` part×role tree) remain the proposed follow-up PR.

tya5:
**[security-reviewer]** — RECONCILED verdict on the revision (HEAD bc13be03): **CLEAR to merge. This supersedes my prior BLOCK** — the completeness BLOCK is genuinely closed. Verified by independent hands-on strip-falsify (not trusting the "verified" claim), since the whole risk was "derived-looking but secretly a hand-list."

**Finding 1 (was BLOCK) — completeness: CLOSED.** `PART_TYPE_REGISTRY` is now genuinely derived, not a hand-list:
- **Collector is a real dir-walk** — `pkgutil.iter_modules(package.__path__)` over `reyn.core.part_types` (part_type_registry.py:139-140), no fixed submodule name list.
- **Forgotten-registration witness passes hands-on** (the decisive check): I created `src/reyn/core/part_types/zzz_probe.py` (valid `PartTypeSpec`, reused skill's registry_ref), touched **nothing else** (not the collector, not any list) → `build_part_type_registry()` → the probe **auto-appears**. Deleted it, tree clean. The PR's own `test_dropped_marked_module_auto_appears_in_default_registry` does this against the **real shipped package** (no mock/monkeypatch).
- **Strip → RED**: replacing the `iter_modules` walk with a fixed 5-name tuple flips both auto-appear tests RED. So the dir-walk is load-bearing; a disguised hand-list would fail the suite.

The design fork is sound: since the 5 builders live in separate packages with no shared dir, the coder created one walked marker package (`reyn.core.part_types`), each marker pointing at its scattered builder via `registry_ref`. "Add a 6th = drop a marker file, touch nothing else" holds structurally. The irreducible declaration step (create one marker per part-type) is the single, dir-discovered declaration point — the correct shape.

**Finding 2 (was in-PR-required) — registry_ref DEAD field: CLOSED.** Now live-resolved (import `module:qualname` + attr walk). Hands-on: pointing skill's `registry_ref` at `...:no_such_function` → importing the module raises `PartTypeCollectionError: ... does not resolve`. Stale/renamed builder can no longer pass silently.

Also confirmed: all 5 real part-types present with registry_ref resolving to the real builders (skill/pipeline/mcp/hook/presentation), `task` absent; PIN-i taxonomy self-consistency still RED-on-strip (blank category → RED); registry-only (no external consumers yet — expected for the Layer-B split, not a defect); tree clean throughout, no stash.

Block-strength: **CLEAR, no gate.** Both prior findings closed and witnessed under strip. Merge-clear. Deferred-correctly (not #2899 gates): wiring hook/present category labels into the live `CATEGORIES` tuple = catalog-completion = F1-proper, and Layer A (provenance install + present-install op) / Layer C (frame) — the acknowledged follow-up slices.

Issue #2259, opened 2026-06-28
Async-decoupled durability redesign: in-memory immediate / durability async via the worker; seq-in-worker; config-as-snapshot (fixes a real truncation bug)


tya5:
**[lead-coder]** — RE-SPEC (owner-confirmed, supersedes the original seq-into-worker scope). This is the owner's INTENDED durability architecture — substantially larger than the landed 1a-i (which was the conservative await-durability step). Owner walked me through it; this is the canonical contract.

## The model: in-memory and durability are ASYNCHRONOUS (decoupled)

- An operation arrives → the task loop updates **in-memory state IMMEDIATELY** (agent-create, config change, step — applied at once). The task loop **never blocks** on durability.
- The task loop then **submits ONLY the seq-synchronized durable record** (the non-volatile restore/time-travel info) to the worker. It does **not** await durability (fire-and-forget from the task loop's view).
- The **DurabilityWorker** is the single serial venue: it **assigns the seq → writes → fsyncs**, asynchronously, in order, for every durable record (WAL events, snapshots, config snapshots).
- **In-memory state is NEVER submitted to the worker** — only the durable record is. The seq lives entirely inside the worker; the task loop never touches/returns it.

**Two WHYs (both load-bearing):** (1) correctness — applied_seq / ordering is consistent by construction (the worker is the single serialization point); (2) responsiveness — the task loop is never blocked by fsync.

**Crash semantics (accepted tradeoff):** recovery restores to the LAST DURABLE record. In-memory ops done but not yet durably written at crash are lost (in-memory was volatile anyway). This is relaxed-durability — the deferred "Step 2" that turns out to BE the owner's intended design.

## Seq is internal to the worker; in-memory state is event-driven

- `append` does NOT return the seq. The seq is assigned in the worker and used WITHIN the durable stream.
- Consumers that today use the returned seq are reframed:
  - **snapshot.applied_seq**: the worker embeds the current seq when it writes the snapshot (one serial task → applied_seq exactly matches the serialization; closes the current "applied_seq set post-await on the caller" window).
  - **agent identity / rewind lineage** (agent_created seq): the creation-seq is reconstructed FROM the durable event stream on recovery, and the in-memory effect is immediate at create time — it does not need a returned seq.
- Eliminate the StateLog `self._lock` (worker single-serial-execution IS the ordering).

## Config = a SNAPSHOT (NOT a WAL event) — and this FIXES a real merged bug

[...]

PR #2900, opened 2026-07-12
feat: snapshot-back the loop-valve counter (hook_driven_turns) — crash-durable, truncation-safe (#2884)

### LINKED ISSUE #2884: Hook-Event: persist the loop-valve counter (`_hook_driven_turns`) across crash — recovery-gated arc (proposal 0059 future-list #2)

tya5:
**[lead-coder]** — Closes #2884.

## Problem
`Session._hook_driven_turns` (the #1800 slice 7 loop-valve counter that backstops hook self-continuation / composed→wake / emit→wake loops) was an **in-memory int, not snapshot/WAL-backed**. A crash+restart reset it to 0, so a near-cap self-wake loop could get a fresh free budget window after a crash.

## Honest threat model (per the architect's design comment on #2884)
This is a **defense-in-depth gap, not an acute exploit**. The dangerous composer self-loop correlation state is separately crash-non-durable (`InMemoryPendingStore` drops in-flight composed pendings on crash — fail-safe), so the self-stimulating chain cannot auto-re-arm from composer state. What survives is one WAL-logged in-flight hook-wake (redelivered post-restart) or an external re-stimulus — either re-enters the valve seam **with the counter freshly reset to 0**. Worth closing now that `emit_hook_event` makes a fully-autonomous LLM loop live and the valve is its only bound (and before 0060 Phase-4 leans on the valve to force-close auto-improvement loops).

## Fix — snapshot-backed (the load-bearing decision)
A pure-WAL-derived counter is **NOT truncation-safe** — consumed `inbox_put{kind:"hook"}` events are pruned by `truncate_below` (only `REWIND_KIND` is force-kept), exactly the #2259 config-loss class. So the counter is **snapshot-backed**, mirroring `buffered_intervention_answers`:

- **`AgentSnapshot.hook_driven_turns: int`** (new field) — `agent_snapshot.py`; wired into `to_payload` / `load` / `_apply_one` (replays a new `hook_driven_turns_set` WAL kind for between-snapshot correctness). Mirrors `buffered_intervention_answers` (the "in-memory dict is the runtime cache; this field is its on-disk durable form" pattern).
- **`SnapshotJournal.record_hook_driven_turns(count)`** — `snapshot_journal.py`; WAL-append (`hook_driven_turns_set`, registered in `WAL_EVENT_KINDS`) + snapshot update + `save_nowait`. Records the FULL value at both the reset (user turn) and each increment (hook turn) edge.
- **`Session.restore_state`** now sets `self._hook_driven_turns = snapshot.hook_driven_turns` — the restore wiring.
- **`Session.reset_for_rewind`** resets the mirror to 0 (zero-residue drift-guard) + `Session.hook_driven_turns` public read-only property.

A single int → one snapshot field; no generation-store (heavier than needed, per the architect).

## Truncate-falsify test — INCLUDED IN THIS PR (CLAUDE.md recovery-feature hard-rule gate)
`tests/test_2884_hook_driven_turns_truncation_falsify.py::test_hook_driven_turns_survives_wal_truncation_below_its_source_events` (mirrors `tests/test_2259_config_truncation_bug.py`):
1. Drive N=3 hook-driven turns → counter=3; flush the durable WAL+snapshot writes.
2. Assert the final `hook_driven_turns_set@3` source event is durable pre-truncation.
3. Push 150 filler WAL events; `truncate_below(floor)` past the source events; assert `dropped >= N` **and** that no `hook_driven_turns_set` line survives in the WAL (so it can't pass vacuously).
4. Reconstruct (load snapshot + replay WAL tail) into a fresh Session → **assert counter survives as 3**.

**Falsify (verified manually, then reverted):** stripping the `restore_state` wiring → the truncate-falsify test goes RED (reconstructs to 0); separately stripping the `AgentSnapshot.load` field wiring → both tests RED. Restored → GREEN. This proves the snapshot-backing is load-bearing and survives truncation.

Plus a basic round-trip test (`test_agent_snapshot_hook_driven_turns_round_trip`) preserving a non-default value (7, not 0).

## Design fork
None — followed the architect's canonical design comment exactly. The one confirm-item the architect flagged (whether `inbox_consume` carries the item kind for between-snapshot replay) is sidestepped: `record_hook_driven_turns` writes the FULL counter value at both edges via its own `hook_driven_turns_set` kind, so replay reconstructs the exact value without depending on `inbox_consume` carrying kind.

## Test / gates
- Local gates green: `ruff check src tests`, `test_tier_audit.py --strict` (both changed test files), `verify_module_docstrings.py` (4 changed src files).
- **Local full `pytest`: 6009 passed, 14 skipped** (the initial run surfaced one by-construction drift-guard failure — `test_reset_for_rewind_clear_scope_covers_all_agentsnapshot_fields` requires every new AgentSnapshot field get an explicit reset disposition — now fixed and re-run green).
- **CI runs the full pytest** on push (authoritative).

### DIFF
--- a/src/reyn/core/events/agent_snapshot.py
+++ b/src/reyn/core/events/agent_snapshot.py
@@ -70,6 +70,15 @@ class AgentSnapshot:
     # Each entry is ``{"kind": str, "payload": dict}`` (no msg_id — already
     # consumed from the inbox before staging here).
     next_turn_context: list[dict] = field(default_factory=list)
+    # #2884: the hook-driven-turns loop-valve counter (Session._hook_driven_turns),
+    # snapshot-backed for crash-durability. A pure-WAL-derived count is NOT
+    # truncation-safe — the consumed `inbox_put{kind:"hook"}` events it would
+    # need are pruned by `truncate_below` (only REWIND_KIND is force-kept,
+    # retention.py) — exactly the #2259 config-loss class. The snapshot floor
+    # sits at/above the truncation floor, so this field survives truncation
+    # by construction. Kept current between snapshots by `apply_events`
+    # replaying `hook_driven_turns_set` WAL entries (see `_apply_one`).
+    hook_driven_turns: int = 0
 
     # ── persistence ─────────────────────────────────────────────────────
 
@@ -127,6 +136,7 @@ class AgentSnapshot:
             next_turn_context=list(
                 data.get("next_turn_context", []) or []
             ),
+            hook_driven_turns=_coerce_int(data.get("hook_driven_turns", 0)),
         )
 
     def to_payload(self) -> dict:
@@ -143,6 +153,7 @@ class AgentSnapshot:
             "outstanding_interventions": self.outstanding_interventions,
             "buffered_intervention_answers": self.buffered_intervention_answers,
             "next_turn_context": self.next_turn_context,
+            "hook_driven_turns": self.hook_driven_turns,
         }
 
     def serialize(self) -> str:
@@ -260,5 +271,11 @@ class AgentSnapshot:
                 self.next_turn_context.append(entry)
         elif kind == "next_turn_context_cleared":
             self.next_turn_context.clear()
+        # ── #2884: loop-valve counter (between-snapshot replay maintenance) ──
+        elif kind == "hook_driven_turns_set":
+            try:
+                self.hook_driven_turns = int(event.get("count", 0))
+            except (TypeError, ValueError):
+                self.hook_driven_turns = 0
         # step_started/completed/failed mutate per-task snapshot only — no agent-level state change here.
         # Unknown kinds: no-op (forward compatibility for future kinds)
--- a/src/reyn/core/events/state_log.py
+++ b/src/reyn/core/events/state_log.py
@@ -111,6 +111,12 @@ WAL_EVENT_KINDS = (
     "task_rebound",
     # (#2248 PR-A added `config_changed` here; #2259 PR-1 removed it — config recovery is now
     # a truncation-surviving GENERATION, not a truncatable WAL event. See config_generations.py.)
+    # NEW (#2884) — the hook-driven-turns loop-valve counter's FULL current value (not a
+    # delta), recorded at every reset (kind="user") and increment (kind="hook") edge. This
+    # kind is a truncatable, between-snapshot replay-maintenance record ONLY — the
+    # reconstruction SOURCE-of-truth is the snapshot field (AgentSnapshot.hook_driven_turns),
+    # since consumed WAL entries are pruned by truncate_below (the #2259 config-loss class).
+    "hook_driven_turns_set",
 )
 
 
--- a/src/reyn/runtime/services/snapshot_journal.py
+++ b/src/reyn/runtime/services/snapshot_journal.py
@@ -385,6 +385,28 @@ class SnapshotJournal:
         self._snapshot.next_turn_context.append(entry)
         self.save_nowait()
 
+    # ── loop-valve counter (#2884) ──────────────────────────────────────────
+
+    async def record_hook_driven_turns(self, *, count: int) -> None:
+        """Append ``hook_driven_turns_set`` to WAL + update the snapshot (#2884).
+
+        Mirrors ``record_intervention_answer_buffered``: Session's in-memory
+        ``_hook_driven_turns`` (the loop-valve counter bounding hook
+        self-continuation) is the runtime cache; ``AgentSnapshot.hook_driven_turns``
+        is its on-disk durable form. Records the FULL current value (not a
+        delta) at both the reset-to-0 (kind="user") and each increment
+        (kind="hook") edges, so replay (``AgentSnapshot._apply_one``) can
+        reconstruct the exact value between snapshots by replaying the same
+        WAL kind — no dependency on ``inbox_consume`` carrying the item kind.
+        """
+        if self._state_log is None:
+            return
+        self._wal_append_nowait(
+            "hook_driven_turns_set", target=self._agent_name, count=count,
+        )
+        self._snapshot.hook_driven_turns = count
+        self.save_nowait()
+
     async def record_next_turn_context_cleared(self) -> None:
         """Append ``next_turn_context_cleared`` to WAL + clear the buffer (#1800-4b).
 
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -2476,6 +2476,17 @@ class Session:
         """
         return self._buffered_intervention_answers
 
+    @property
+    def hook_driven_turns(self) -> int:
+        """Read-only accessor for the hook-driven-turns loop-valve counter (#2884).
+
+        Snapshot-backed (``AgentSnapshot.hook_driven_turns`` — see
+        ``restore_state`` / ``SnapshotJournal.record_hook_driven_turns``) so
+        tests and observability can read the crash-durable value without
+        reaching into the private ``_hook_driven_turns`` field.
+        """
+        return self._hook_driven_turns
+
     def _is_turn_cancel_requested(self) -> bool:
         """Forwarding → RouterLoopDriver.is_cancel_requested (PR-3)."""
         return self._loop_driver.is_cancel_requested()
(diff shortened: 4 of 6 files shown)


tya5:
**[security-reviewer]** — #2900 (valve-persist, recovery-gated) co-vet: **CLEAR to merge on the recovery hard-rule**, with **one flagged STEP-4 gap** (block-strength below). Independent strip-falsify (cp-restore, diff vs the PR's own base `1b142e38` to filter #2899 merge noise).

**Recovery hard-rule = solid:**
- **Truncate-falsify NON-VACUOUS — the strongest form in this codebase.** The test drives N=3 real hook turns, pushes 150 filler events (floor = current_seq-5), truncates, and asserts BOTH `stats["dropped"] >= N` AND that the post-truncation WAL contains **zero** `hook_driven_turns_set` lines (explicit anti-vacuous guard: "otherwise this test would pass vacuously even for a WAL-derived design"), then reconstructs via fresh StateLog + apply_events + restore_state and asserts counter == N.
- **Snapshot-backing load-bearing**: STRIP `restore_state` (session.py:4408 → `= 0`) → truncate-falsify FAILS (`got 0, expected 3`). Confirmed genuinely snapshot-backed.
- **Pure-WAL-derive genuinely rejected**: `restore_state` reads `snapshot.hook_driven_turns` (durable snapshot field), not a prunable WAL inbox count — the #2259 loss class is avoided by construction.

**STEP-4 — flagged gap (my independent strip; not caught upstream):** the `reset_for_rewind` drift-guard (`test_..._clear_scope_covers_all_agentsnapshot_fields`) only asserts `hook_driven_turns` is a KEY in the disposition dict — a by-construction field-coverage check. It does **not exercise the reset**. STRIP: I removed the real reset line `self._hook_driven_turns = 0` (session.py:3042), left the disposition entry → **all reset_for_rewind tests still PASSED** (green-on-strip). So a future accidental deletion of the reset ships silently — a rot vector by my bound-test-must-flip discipline (a test that stays green when the behavior is stripped does not witness it).
- **Severity depends on an open question**: is `reset_for_rewind` ever reached WITHOUT a following `restore_state`? If `restore_state` always follows (overwriting from snapshot), the reset line is belt-and-suspenders and the weak test is low-risk. If a rewind path exists with no subsequent restore, the reset is load-bearing (stale counter carries → wrong valve budget post-rewind) and the missing behavioral witness is a real hole. The co-vet did not resolve this reachability.
- **Fix (cheap)**: extend `test_reset_for_rewind_then_restore_state_zero_residue` to populate a nonzero `_hook_driven_turns` pre-rewind and assert 0 post-reset / pre-restore → the test then flips RED when session.py:3042 is stripped.

**Block-strength:** recovery hard-rule (truncation survival, the #2884 deliverable) = **CLEAR, no gate**. STEP-4 = **in-PR-preferred** test-tightening (closes a green-on-strip rot vector; cheap) + resolve the reset-without-restore reachability question — but **not a blocker on the recovery gate** since the reset CODE is present and the truncation-survival requirement is met. Lead's call whether to require the test-tightening in-PR or accept as an immediate follow-up; I would not accept it *silently* (it's a real green-on-strip finding).

tya5:
**[lead-coder]** — STEP-4 green-on-strip vector closed (architect co-vet follow-up).

**Reachability finding:** the `reset_for_rewind` counter reset is **belt-and-suspenders** for `_hook_driven_turns`. There is exactly one executable `reset_for_rewind` call site — `src/reyn/runtime/registry.py:2189` — and `restore_state(snap)` runs on the immediately-following **unconditional** line (`registry.py:2190`, same `if session is not None:` block, no branch/return between). `restore_state` overwrites `_hook_driven_turns` wholesale from the reconstructed snapshot (`session.py`), so there is no reset-without-restore path where a stale counter could survive into the post-rewind session. (For the *other* holders reset_for_rewind clears — inbox/chains/interventions/buffered/next_turn_context — the reset IS load-bearing, since restore_state only adds to those rather than replacing them wholesale.)

**Behavioral pin added (closes the green-on-strip vector):** `tests/test_reset_for_rewind.py::test_reset_for_rewind_zeroes_hook_driven_turns_counter` — populates a nonzero counter (5) pre-rewind, calls `reset_for_rewind`, asserts the public `session.hook_driven_turns == 0`. **Falsify-verified by hand:** stripping the `_hook_driven_turns = 0` reset line → this test goes RED; restore → GREEN. So the reset is now behaviorally pinned, not just field-coverage-guarded.

Scoped strictly to this follow-up — the truncate-falsify work (architect CLEAR'd) is untouched. Local gates re-run green: ruff, tier-audit --strict (tests/test_reset_for_rewind.py), verify_module_docstrings (session.py). CI runs the full pytest.

tya5:
**[security-reviewer]** — #2900 STEP-4 fix (2c44c12e) re-co-vet: **RECONCILED CLEAR.** With the recovery-gate already CLEAR, this closes the green-on-strip gap. #2900 is fully clear to merge.

- **Reset pin witnesses (strip → RED)**: stripping `self._hook_driven_turns = 0` from `reset_for_rewind` (session.py:3042) → `test_reset_for_rewind_zeroes_hook_driven_turns_counter` FAILS (`assert 5 == 0`). The previously-green-on-strip reset is now pinned (public accessor, no private-state assertion).
- **Reachability — belt-and-suspenders confirmed complete (full caller enumeration)**: `grep reset_for_rewind src/reyn` → exactly ONE caller (registry.py:2189), unconditionally followed by `restore_state(snap)` in the same `if session is not None:` block, no branch/return between. There is no second call site, so the belt-and-suspenders property holds for *all* callers — lead's single-site check was in fact exhaustive.
- **Redundant-but-pinned, correctly reasoned**: `restore_state` overwrites `_hook_driven_turns` wholesale (session.py:4408, unconditional assignment) → the reset is redundant whenever restore follows (which is always). The coder's distinction holds: contrast `outstanding_interventions`, whose restore is *additive/conditional* (session.py:4436-4472, only re-enqueues `if snapshot.outstanding_interventions`), so *its* reset IS load-bearing. So `hook_driven_turns`'s reset is provably redundant today but pinned for future-drift robustness (if restore ever becomes additive for it, the pin already guards). Defensible.
- **Scope strict**: the delta is test-only; snapshot field / WAL kind / journal / truncate-falsify all unchanged from the earlier commit. 6 tests pass; tree clean.

Block-strength: **CLEAR, no gate** — recovery-gate (truncate-falsify, already verified) + STEP-4 (reset pin) both closed and strip-witnessed. Merge-clear.

PR #2906, opened 2026-07-12
feat: proposal 0060 Phase 1 — Layer C (part×role routing frame, derived from the meta-registry)

tya5:
**[lead-coder]** — Layer C of proposal 0060 Phase 1 (part of proposal 0060; toward F2). Small, security-surface-free (unlike Layer A). Ratified design: `docs/deep-dives/proposals/0060-llm-wielding-foundation.md` Addendum C.

## What

Adds the scheme-independent **part×role mechanism-routing frame** to the OS system prompt so the model learns *which mechanism to reach for* by role, not just a flat action catalog.

### C1 — placement (cache-static, scheme-independent)
`render_mechanism_routing_frame()` in `src/reyn/prompt/router_frame.py`, injected by `build_system_prompt` into the STATIC cache-prefix block (a new `## Mechanism routing` section, **before** `## Behaviour`) — NOT a scheme-owned `tool_use_sp` slot. Every tool-use scheme funnels through this one OS-frame builder, so the frame holds identically across universal / enumerate / retrieval / codeact and the bare no-scheme frame.

### C2 — content
1. the derived part×role map;
2. a mechanism-selection decision tree (need INPUT → hook | mcp | retrieval; need WORKFLOW → skill | pipeline | mcp-step; need OUTPUT → present | render | mcp-write);
3. **hooks are now visible in the SP** for the first time (`hook (hook): input`);
4. an author-vs-reuse heuristic + a typed/permissioned/evaluated authoring-quality one-liner.

### C3 — LOAD-BEARING: the map is DERIVED from the meta-registry
`render_part_role_map()` walks `PART_TYPE_REGISTRY.items()` (sorted) and renders one row per part-type from that spec's own `.name` / `.category` / `.roles` frozenset — never a hand-written parallel table. A new marker dropped into `reyn.core.part_types` auto-appears with **zero edits** to `router_frame.py` (the #2899 completeness discipline applied at the SP layer). A per-part-type char cap (`MAX_PART_TYPE_ROW_CHARS`) keeps the derived rows within the cache-static budget as the registry grows.

## Tests — `tests/test_mechanism_routing_frame_0060.py` (3 co-vet pins)

1. **Registry-derivation (crux):** `test_new_marked_part_type_auto_appears_in_routing_frame` drops a real marked module into `reyn.core.part_types` → it appears in the rendered frame. `test_hand_written_table_frozen_at_collection_falsifies_the_new_part_type` is the negative control — a snapshot frozen before the drop-in does NOT reflect the new part-type (the drift a hand-list would carry).
2. **Scheme-independence:** `test_routing_frame_appears_under_all_four_schemes_and_bare_frame` + `test_routing_frame_present_with_zero_scheme_supplied_content` — frame present under all 4 scheme slot shapes and with `tool_use_sp=None`.
3. **Char-budget / cache-static:** `test_per_part_type_row_is_bounded_and_raises_when_exceeded`, `test_synthetic_larger_registry_stays_within_linear_budget` (50-part synthetic registry stays linear), `test_routing_frame_sits_in_the_static_section_before_behaviour`.

## Test plan
- [x] Full `pytest` (foreground, ~3.5min): **8348 passed, 20 skipped, 0 failed**.
- [x] `ruff check src tests`: clean.
- [x] `python scripts/test_tier_audit.py --strict tests/test_mechanism_routing_frame_0060.py`: 8/8 OK, 0 Tier-4.
- [x] `python scripts/verify_module_docstrings.py` on changed src: clean.

### DIFF
--- a/src/reyn/prompt/router_frame.py
+++ b/src/reyn/prompt/router_frame.py
@@ -234,3 +234,100 @@ def output_language_directive(output_language: str) -> str:
         f"  - Always reply in language: {output_language}."
         "  Do NOT switch language even for error messages or clarifying questions."
     )
+
+
+# ── Mechanism routing (part x role) — 0060 Addendum C, Layer C ─────────────
+# WHEN: always, scheme-independent, static cache-prefix section.
+# WHERE: build_system_prompt() -> a dedicated "## Mechanism routing" section,
+#        placed in the STATIC block (before "## Behaviour") -- NOT a
+#        scheme-owned tool_use_sp slot. It therefore holds across all four
+#        tool-use schemes (universal / enumerate / retrieval / codeact): they
+#        all funnel through this one OS-frame builder, and none of them can
+#        omit or overwrite this section the way a scheme-owned slot could.
+# WHY: proposal 0060 Addendum C (C1/C2) -- the model needs a standing map of
+#      WHICH mechanism to reach for by role (input / workflow / output), not
+#      just a flat action catalog, and hooks (today entirely absent from the
+#      SP) need to become visible. C3 (load-bearing): the part x role rows
+#      are DERIVED from reyn.core.part_type_registry.PART_TYPE_REGISTRY's
+#      ``roles`` frozensets -- NEVER a hand-written parallel table (the
+#      #2899 completeness discipline applied at the SP layer). A new marked
+#      part-type dropped into reyn.core.part_types auto-appears here with
+#      zero edits to this module.
+# 日本語訳: 常に描画される、scheme非依存の静的節。「入力/処理/出力」の各役割に
+#      どの機構を使うべきかの地図を提示し、これまでSPに一切現れなかった hook を
+#      可視化する。表の各行は PART_TYPE_REGISTRY の roles frozenset から導出され、
+#      手書きの並行テーブルは禁止（#2899 の完全性規律をSP層に適用）。
+MECHANISM_ROUTING_HEADER = "## Mechanism routing (part x role)"
+
+MECHANISM_DECISION_TREE = (
+    "Pick the mechanism by what you need, not by habit:\n"
+    "  - need INPUT (new data or a reactive trigger) -> hook | mcp | retrieval\n"
+    "  - need WORKFLOW (multi-step orchestration) -> skill | pipeline | mcp-step\n"
+    "  - need OUTPUT (present, render, or write externally) -> present | render | mcp-write"
+)
+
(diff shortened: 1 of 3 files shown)


tya5:
**[security-reviewer]** — #2906 (0060 Layer C, part×role routing frame) co-vet: **CLEAR to merge — 0060 F1's last piece.** Independent hands-on strip-falsify.

- **C3 derivation GENUINE (the #2899-class check) — the load-bearing property holds.** `render_part_role_map` (router_frame.py:302-317) derives each row from `PART_TYPE_REGISTRY` via `_format_part_type_row(spec)` (name/category/`sorted(roles)`) — no parallel hand-table. Hands-on: dropped a marked part-type module into `reyn.core.part_types/` touching nothing else → its row (`covet_probe_part (…): output`) auto-appeared in the rendered frame. Strip the derive loop to a frozen 5-row literal → `test_new_marked_part_type_auto_appears` RED. So Layer C rides the Layer B SSoT with zero drift, same discipline as #2899 — a new part-type auto-appears in the SP routing map, no hand-maintained parallel table.
- **Scheme-independence (pin 2)**: gate the injection to one scheme → `test_..._appears_under_all_four_schemes` + `test_..._zero_scheme_content` RED. The frame is in the OS-frame (router_system_prompt.py:190-197, before `## Behaviour` = static cache-prefix), holds across all 4 tool-use schemes.
- **Per-type char-budget (pin 3)**: `_format_part_type_row` raises over `MAX_PART_TYPE_ROW_CHARS`=100; strip the cap → `test_per_part_type_row_is_bounded` RED. The per-type bound (your pressure-test point) is enforced inside the derivation, so the frame stays within budget as the SSoT grows (50-synthetic linear test confirms).
- **Hook visibility + completeness**: frame renders `hook (hook): input` (hooks were absent from the SP entirely before this) and all 5 part-types (skill/pipeline/mcp/hook/present) — complete by construction (registry-derived, 1:1 with the marker modules). 8/8 tests, tree clean.

C2/C3 split confirmed correct: the part×role **map** is derived (data, from the registry, no drift); the decision-tree **mechanism vocabulary** (present|render|mcp-write) is authored prose (judgment, intentionally fixed). Map = "what exists" (SSoT data); tree = "how to choose" (design prose) — deriving one and authoring the other is coherent, not a contradiction.

Block-strength: **CLEAR, no gate.** #2906 clear to merge → **0060 F1 COMPLETE** (Layer B meta-registry SSoT + Layer A provenance backbone + Layer C SP routing frame, all landed security-sound + drift-proof).

PR #2911, opened 2026-07-12
feat: proposal 0060 — D5b/c/d docs-delivery mechanism (docs-carrier + doc_ref + error-rail)

tya5:
**[lead-coder]** — proposal 0060 Addendum D mechanism slice (content-independent; NO F3b cheat-sheet content, NO D5e SP-gate). Toward proposal 0060 Addendum D.

## D5b — docs-carrier (production file-read reachability)

**Build-step (b) build-time copy — PREFERRED, taken.** A custom `setup.py` `build_py` hook runs `scripts/mirror_reference_docs.py`, copying `docs/reference/` → `src/reyn/builtin/reference/` at wheel-build time, shipped by the existing `builtin/**/*` package-data glob (F3a). **No committed duplicate** — the mirror dir is git-ignored; `docs/reference/` stays the sole source of truth (owner's clean-end-state preference). Fallback (a) was not needed: the hook is a straight `shutil.copytree`, well within setuptools' `cmdclass` capability (all other metadata stays declarative in `pyproject.toml`).

- **Runtime reader**: `reyn.builtin.docs.read_builtin_doc()` resolves via `importlib.resources.files("reyn.builtin")/"reference"` first (works in a wheel install with no repo root — the production fix), falling back to `resolve_reyn_root()` + `docs/reference/` for a dev checkout that hasn't run the build hook.
- **Byte-gate** (`tests/test_0060_d5b_docs_mirror.py`): the mirror is byte-for-byte identical to `docs/reference/` (file set + per-file bytes); a dropped file → RED (falsify test mirrors a filtered source and asserts divergence is caught).
- **Production-reachability**: monkeypatch `resolve_reyn_root` to raise (simulating a wheel-install layout with no repo root) → `read_builtin_doc()` still reaches the mirrored doc via `importlib.resources`. A companion falsify test asserts that with NEITHER the mirror NOR a repo root, the reader raises a loud `DocNotFoundError` (not silent).

## D5d — `doc_ref` first-class field + registry-walk completeness gate

- **`doc_ref` on BOTH entities**: optional `str | None` on `ToolDefinition` (`src/reyn/tools/types.py`), REQUIRED `str` on `PartTypeSpec` (`src/reyn/core/part_type_registry.py`) with a `DOC_REF_NONE = "none"` sentinel for genuinely doc-less part-types (explicit opt-out, never a forgotten field — `__post_init__` rejects empty).
[...]

PR #2912, opened 2026-07-12
feat: proposal 0060 — F3b builtin content (curated-5 + cheat-sheet hub) + SP present-affordance + D5e SP-gate

tya5:
**[lead-coder]** — part of proposal 0060 (LLM-Wielding Foundation), F3b core spine + D5e.

> Rebased onto #2911 (D5b/c/d docs-delivery mechanism). No conflicts. Reconciled with the new `doc_ref` field: this PR's new `JUDGE_OUTPUT` `ToolDefinition` now carries `doc_ref="docs/reference/runtime/control-ir.md#judge_output"` (D5d spirit — spec-bearing op). No part-types are added, so the D5d **required**-on-`PartTypeSpec` gate is unaffected. The cheat-sheet's corrupted-example falsify test still asserts `PipelineParseError` (the D5c error-rail only appends a pointer, doesn't change the type). All D5b/c/d tests stay green.

## What this ships

**Core spine only** (per the dispatch brief's own split allowance) — the
cheat-sheet hub + flagship pipeline + SP present-affordance + D5e gate. The
other 3 curated builtins (draft→judge→revise skill, status-card
present-view, standalone hook-exemplar builtin) are proposed as a **sibling
PR** — see "Split rationale" below.

### 1. `reyn_cheat_sheet` builtin skill (the hub, Addendum D1/D4)
`src/reyn/builtin/skills/reyn_cheat_sheet/SKILL.md` — the reyn-specific
usage guide: mechanism decision tree, `present`/`judge_output` essentials,
skill/pipeline/hook/mcp authoring pointers, and two CI-executable worked
examples (the flagship pipeline; a `file_changed → pipeline_launch` hook).
Registered in `BUILTIN_SKILLS`, loads with `provenance="builtin"`,
`auto_invoke=False` (discoverable, never auto-fires).

### 2. `flagship.research_and_report` builtin pipeline
`src/reyn/builtin/pipelines/flagship_research_and_report.yaml` —
`web_search → agent (summarize) → judge_output (self-review) → present`
(input → workflow → output composition thesis). Registered in
`BUILTIN_PIPELINES`, `provenance="builtin"`, invoke-by-name (inherently
inert). The cheat sheet's embedded copy is kept in sync with this file by a
test (single-home content, Addendum D3).

### 3. SP present-affordance essential (both directions) + cheat-sheet pointer
`src/reyn/prompt/router_frame.py` — two new lines in the static "Mechanism
routing" frame:
- OUTPUT: use `present` to show results to the operator (zero token cost).
- INPUT: content you must read/act on — read it into your own context; do
  NOT `present` it (presenting it means you never see it).
Plus a named pointer to the `reyn_cheat_sheet` skill (`CHEAT_SHEET_POINTER`).

### 4. D5e — SP-gate
`REYN_CHEAT_SHEET_SKILL_NAME` is the load-bearing name the SP points at;
`test_d5e_sp_pointer_resolves_to_a_real_builtin_skill` asserts it resolves to
a real `BUILTIN_SKILLS` entry whose file exists; a falsify test removes the
entry and shows the gate goes RED.

## The agent→judge plumbing seam — how it was resolved (built + ran it)

Building the flagship and actually running it surfaced two real gaps, not
covered by the brief's three candidate resolutions — fixed here, both small
and precedented:

1. **`judge_output` had no invocation surface at all.** It existed only as
   a legacy phase-graph op — never registered in `get_default_registry()`
   — so `tool: {name: judge_output}` failed "not a registered tool" in
   *any* pipeline, not just this one. Fixed by adding
   `src/reyn/tools/judge_output.py` (a `ToolDefinition`), mirroring the
   established `present`/`compact` precedent exactly (#2692) — same
   dispatch pattern, zero new architecture.
2. **`judge_output.target` only ever resolved against
   `ctx.workspace.artifacts`**, populated solely by `Workspace.store_artifact`
   — a method the pipeline executor never calls (a pipeline step's output
   lives in the pipeline's own `ctx` store, a disjoint namespace). Fixed by
   adding `data_inline` as an XOR alternative to `target` on
   `JudgeOutputIROp` — the exact same idiom `PresentIROp`/
   `RenderTemplateIROp` already use for `data_ref`/`data_inline`. `target`'s
   legacy behavior is untouched (existing tests pass unmodified).
3. **A related discovery**: a pipeline `tool:` step's `ctx.<output>` is
   always reduced to `{text, structured}` via the canonical mapper — `meta`
   fields (judge_output's `score`/`passed`/`threshold`) are DROPPED for
   pipeline consumption (this is a general, pre-existing pipeline-wide
   behavior, not something introduced here). Without a fix, `present`'s
   blueprint could never bind the judge's score/passed downstream. Fixed
   narrowly, scoped to `judge_output_to_canonical` only: it now also emits a
   `{"kind": "structured", "data": {score, passed, threshold, on_fail,
   reason}}` attachment (the same text+structured-attachment idiom ~15
   other canonical mappers already use — `web_search_to_canonical`,
   `mcp_to_canonical`, etc.). Chat-side rendering is unchanged (`text` is
   still the short reason string). `present`'s blueprint binds
   `/verdict/structured/score` etc.

**Verified it runs**: `test_flagship_pipeline_parses_and_runs_end_to_end`
drives the REAL `PipelineExecutor` + the REAL tool-registry dispatch
(`_make_tool_dispatch` + `get_default_registry()` — so
`web_search`/`judge_output`/`present` all resolve through actual production
dispatch, not a stub `tool_dispatch`). Only the LLM (`litellm.acompletion` +
the agent step's scripted reply) and the web-search network boundary
(`DuckDuckGoBackend.search`) are faked — the one collaborator (the LLM) the
testing policy allows to be faked, plus the equivalent external-network
boundary. All 4 `present` blueprint bindings resolve
(`bindings_resolved=4`, none dropped).

## Split rationale (why this stops at the core spine)

Per the dispatch brief's own escape hatch ("If too big, STOP after the core
spine... propose the other 3 builtins as a sibling PR"): the judge_output
plumbing fix (above) consumed the budget that would otherwise have gone to
the remaining 3 curated builtins. They're well-scoped follow-ups, not
blocked on anything in this PR:
- **draft → judge_output → revise skill** — a workflow skill showing the
  SKILL.md live-format + the Evaluation idiom (trivially wireable now, given
  the `data_inline` fix here).
- **status/results card present-view** — a blueprint + `$bind` idiom
  builtin `BUILTIN_PRESENTATIONS` entry (currently ships empty; still a
  no-op merge).
- **`on: file_changed` hook exemplar as a standalone builtin** — flagged as
  a design fork below (hooks have no builtin-tier map today); the cheat
  sheet already documents a CI-verified inline example as an interim.

## Design forks to flag

1. **Hooks have no `BUILTIN_HOOKS` map.** `reyn.builtin.registry` has
   exactly `BUILTIN_SKILLS`/`BUILTIN_PIPELINES`/`BUILTIN_PRESENTATIONS`.
   Hooks config (`hooks:` in `reyn.yaml`) is a bare list, wholesale-replaced
   across config tiers (not name-keyed union-merged like the other three) —
   adding a 4th builtin map + a hooks-merge branch is real infra work, out
   of scope here. The cheat sheet ships a CI-verified (`load_hooks`) inline
   example instead, per Addendum D6's "doc-content, never registered"
   framing. Needs a lead/owner call: is a `BUILTIN_HOOKS` map worth building
   for the sibling PR, or does doc-only content suffice permanently for the
   reactive-input axis?
2. **Builtin skill/pipeline file paths resolve OUTSIDE project_root in a
   production (installed-wheel) deploy** — the builtin tier ships inside
   the package (`Path(__file__).parent`-relative), which is right for
   packaging but means reading the cheat sheet's own SKILL.md body at L2
   passes through the standard out-of-project-root file-read permission
   gate (approval required), same as reading any other out-of-root path.
   This PR does not add a builtin-tier carve-out to that gate — flagging it
   as a known reachability gap (the same class Addendum D2's audit named
   for docs/) rather than solving it unilaterally; it affects every future
   builtin-tier skill, not just this one.
3. **`judge_output_to_canonical` now also emits a structured attachment**
   (previously `attachments=[]`, only `meta`). This is additive and
   chat-facing `text` is unchanged, but flagging since it touches a shared
   canonical-mapper contract, not purely new content.

## Testing
- New: `tests/test_0060_phase2_f3b_builtin_content.py` (17 tests — D5a
  executable-example gates ×4 [2 positive + 2 falsify], flagship
  runs-end-to-end, D5e gate ×2, present-scope ×3, provenance/inert ×3, sanity
  ×2).
- Updated (deliberate, sanctioned goldens — same class as #2123's own
  "update this golden INTENTIONALLY" note): `test_2123_router_dispatch_sot_1953.py`
  (dispatch-set golden + `judge_output`), `test_0060_phase1_f3a_builtin_tier.py`
  (F3a's "ships empty" invariant narrowed to `BUILTIN_PRESENTATIONS` only —
  F3b populates the other two), `test_2575_pipeline_disk_registration.py` +
  `test_cli_pipe.py` (both now observe the always-present builtin pipeline
  via real `load_config()`), `test_returns_external_content_flagset_1822.py`
  (classified `judge_output` as `_NOT_EXTERNAL`, same rationale as
  `compact`/`embed`), `tests/tools/_pre_migration_tool_schemas_baseline.json`
  (added `judge_output`'s rendered schema — minimal, order-preserving diff).

**Full pytest**: 0 failures (foreground, full suite, not affected-only).
**ruff check src tests**: clean. **tier-audit --strict** on all changed test
files: 0 Tier-4 violations. **verify_module_docstrings.py**: clean.

## Test plan
- [x] `pytest` full suite green (0 failures)
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` clean
- [x] `python scripts/verify_module_docstrings.py <changed src files>` clean
- [x] Flagship pipeline verified to actually run end-to-end (not just parse)
- [x] D5a/D5e falsify tests confirmed RED when the guarded content is stripped


tya5:
**[security-reviewer]** — #2912 (0060 F3b core spine) co-vet: **mechanism CLEAR to merge; one REQUIRED in-arc follow-up (Fork-2, HIGH severity, not #2912-blocking).**

**Mechanism — all pins load-bearing:**
- **The 2 defect-fixes the flagship build+run exposed are real and load-bearing** (validates "build+run, don't manufacture"): strip `register(JUDGE_OUTPUT)` → flagship RED (tool unresolvable); force target-only (drop `data_inline`) → flagship RED (can't judge the agent step's ctx output). judge_output is now a proper ToolDefinition + `data_inline XOR target` (present/render_template idiom); control-ir.md synced same-PR (hard-rule).
- **Flagship = genuine end-to-end**: real `PipelineExecutor.run()` + real tool dispatch, only litellm/DuckDuckGo faked at the external boundary, all 4 step bindings resolve (`bindings_resolved=4`).
- **D5e SP-gate**: rename the cheat-sheet key → module-load KeyError (dangling-pointer caught). **D5a executable-gate**: corrupt an embedded example → RED (examples parse via real `parse_pipeline_dsl`/`load_hooks`). **present output-scope**: strip the OUTPUT half → RED (both directions taught: results→present / consume→read).
- **New-tool registry-completeness**: no #2885-class miss — judge_output is external-content-classified, tool-desc-baselined, canonical-declared, all green (175+3 pass).

**Fork-2 (cheat-sheet BODY wheel-reachability) — CONFIRMED production-broken; REQUIRED in-arc follow-up.**
The body ships in the wheel (`builtin/**/*` glob) but is READ via the ordinary `read_file` → `require_file_read` → `_in_default_read_zone` (`permissions.py:174`), which does `resolved.relative_to(project_root)`. The body resolves under site-packages, **outside** any user's project_root → `relative_to` raises → **`PermissionError`; the L2 read of the cheat-sheet body fails in a wheel deploy** (non-interactive: hard-fail; interactive: JIT-prompt every session). **D5e proves the SP pointer resolves to a registry *entry* — NOT that the *body* is readable**, so this gap is invisible to the mechanism tests.

Severity is HIGH because the cheat-sheet is **the D5-attach hub** — Addendum D's entire production-reachability premise rests on it; a hub whose body can't be read in production is a broken deliverable. The fix is **known and small**: D5b already built the wheel-safe reader (`read_builtin_doc()` via `importlib.resources`, bypassing the permission gate), but it is **wired only to reference-docs reads, not to the builtin skill/pipeline body read call-site** — the follow-up wires builtin skill/pipeline body reads through that path (a D5b-analog). The coder's honest flag (`registry.py:187-192`) is accurate, not overclaimed.

**Block-strength**: mechanism **CLEAR** — #2912 is merge-able (correct, dev-reachable, fully strip-witnessed; the wheel-reachability is a separable packaging/reader concern, as D5b was its own slice). **Fork-2 = REQUIRED follow-up (no-debt, in-arc): a hard gate on the "production-reachable cheat-sheet" claim / calling the F3 arc complete — must land as a named issue before that claim, not optional.** Fork-1 (BUILTIN_HOOKS map) = track for owner, agreed.

PR #2438, opened 2026-07-02
refactor(#2434): stage3b — machinery bulk-delete (skill/phase engine + ops + schemas)

tya5:
**[e2e-coder]** — Part of #2434 (skill/phase machinery deletion). **Stage 3b: atomic machinery bulk-delete.** The machinery is dead code after stage1 (chat/router/session decouple) + stage3a (CLI/web strip-keep). **318 files removed.** Produced via double-Opus-review (Sonnet bulk-delete → my Opus verify + tool-layer completion + cascade fixes).

## Deleted
- **Directories**: `src/reyn/skill/`, `src/reyn/core/compiler/`, `src/reyn/stdlib/skills/`
- **Kernel phase-engine** (13 modules) — the **KEEP-5** code-execution submodules (`codeact_runner`, `_codeact_harness`, `_python_harness`, `_python_allowlist`, `sub_loop_memo_key`) are left **1-byte-identical** (0 diff).
- **Skill op modules**: `op_runtime/{run_skill,skill_resolve,lint}.py` + their `registry`/`contextual_gate`/`__init__` entries.
- **Skill tools**: `tools/{invoke_skill,lint}.py` + registration (`tools/__init__` LINT register, `universal_dispatch` `validation__lint`, `universal_catalog` "validation" category + its SP bullet).
- **CLI skill-only commands**: run/eval/eval_benchmark/skill/skills/lint + `skill_loader` (**kept `run_once`**) + `commands/__init__` ALL.
- **Pure-skill files**: `services/auto_resume_handler` (+ export), `slash/skills` (+ registration), `web/routers/skills` (+ `web/server` include).
- **Schema classes**: `Phase`/`Skill`/`SkillGraph`/`SkillNodeSpec`/`SkillResolveIROp`/`LintIROp`/`PhaseConstraints` + `OP_KIND_MODEL_MAP` + `ControlIROp` union entries.
- All machinery-internal + obsolete tests.

## Doc-sync (CLAUDE.md hard rule)
`control-ir.md` (+ `.ja.md`) `run_skill`/`skill_resolve`/`lint` sections + `llm.py` op-example doc-comment removed **in this PR**, synced with `OP_KIND_MODEL_MAP`.

## Verification (5 gates — all PASS)
- [x] **KEEP-5 death-grip**: `git diff HEAD` on all 5 files = empty (0 diff, untouched)
- [x] **control-ir.md ↔ OP_KIND_MODEL_MAP sync**: both cleared of run_skill/skill_resolve/lint
[...]

PR #2901, opened 2026-07-12
fix: schema-validate composed:* hook matchers at load — close the Phase-3 footgun for composed events (#2889)

tya5:
**[lead-coder]** — Closes #2889. Reintroduced Phase-3 footgun, closed for `composed:*` hooks.

## Defect
Phase 3 (#2873) made a `matcher` field not in a builtin hook-point's schema fail-loud-reject at load, so a typo can't silently never-match. But `composed:*` hooks were left an open-set — a `matcher` on a `composed:*` hook was NOT schema-validated, so a typo'd field silently never matched again. This closes that for composed events (per the architect's canonical design comment on #2889).

## Key fact making this trivial
Every composed event, across all 7 Composer ops, is emitted by the single `_emit_composed` producer (`composer.py:336-338`) with a FIXED payload shape `{"inputs": [...], "correlation_key": <key>}`. So the schema for any `composed:<name>` is `frozenset({"inputs", "correlation_key"})`, keyed by the composer's `emit.kind`.

## Fix (mirrors the Phase-3 enforce-at-load path exactly)
1. **Reorder** — `src/reyn/runtime/session.py:1418-1436`: `_build_composer_defs` now runs BEFORE `_build_hook_registry` (side-effect-free parse, confirmed), so the full composed-kind universe is known when hooks validate.
2. **Derive the composed-schema map** — `session.py:1434-1436`: `self._composed_schemas = {d.emit_kind: frozenset({"inputs","correlation_key"}) for d in self._composer_defs}`.
3. **Thread it** — `_build_hook_registry` (`session.py:3816,3826`) passes `self._composed_schemas` into every `load_hooks(...)` call → `load_hooks(raw, composed_schemas)` (`loader.py:349`) → `_parse_entry(..., composed_schemas)` (`loader.py:181`) → `validate_event_pattern(from_legacy_matcher(matcher), on_key, composed_schemas)` (`loader.py:~320`). In `event_pattern.validate_against_schema` (`event_pattern.py:116-155`) the schema resolves from `composed_schemas.get(kind)` FIRST (for `composed:*` kinds), falling back to `BUILTIN_HOOK_SCHEMAS` — the two namespaces never collide. Everything downstream (unknown-field diff → `HookSchemaError` → `HookConfigError`) is unchanged. `composed_schemas=None` (the default — bare `load_hooks(raw)` callers with no composer config to thread) preserves the pre-#2889 open-set posture.

## Sub-decision (b): dangling-producer detection — INCLUDED
[...]

PR #2916, opened 2026-07-12
fix: Hook-Event Phase 4+ hardening — Bus drop fail-visible (#2886) + event_name/window/hygiene batch (#2890)

### LINKED ISSUE #2881: Hook-Event Phase 5: open the Composer full-reachability path (producer-wire + consumer-open + loop-valve + e2e reachability assert)

### LINKED ISSUE #2886: hooks: HookBus drop-oldest is silent — no counter/audit on subscriber-queue loss (fail-visible gap)
**[architect]** — Phase 4+ review finding F1 (Observability band).

`HookBus.publish` normal drop-oldest path (bus.py:144) discards silently: no counter, no log, no P6 audit-event (only the `pragma: no cover` double-full race logs a warning). Composer-layer loss IS audited (`composer_dropped`, composer.py:326/333/456) but Bus-subscriber-queue loss upstream is invisible — a correlation that never fired cannot be reconstructed from `reyn events`.

Constraint: publish is sync/never-raises hot path → fix shape = per-subscriber drop counter + metadata-only audit-event on first-drop / every-Nth (never per-event emit in hot loop), or counter surfaced via a snapshot()-style read.

Part of the 0059 arc follow-ups.

### LINKED ISSUE #2890: hooks: Phase 4+ minor hardening/hygiene batch (event_name constraints, per-key window bound, tracking + dead-code cleanup)
**[architect]** — Phase 4+ review findings F6-F9, batched (all low, follow-up-ok).

- **F6**: `EmitHookEventIROp.event_name` has no pattern/max_length (models.py:466-469); handler only strips+non-empty. Structurally confined to own `llm:{session_id}:` prefix (no namespace escape) but control chars/newlines flow into kind→audit, no length cap. Add `pattern=[A-Za-z0-9_.-]+` + max_length; consider payload size note.
- **F7**: window/debounce per-KEY event list unbounded within TTL (capacity bounds keys, TTL bounds age — composer.py:184-190); external event storm = memory pressure for window duration. Add per-key max-events const + `composer_dropped` on overflow.
- **F8**: 'startup-only composers / no hot-reload' known-limitation recorded only in CLOSED #2881 — move to a live surface (feature-map row note).
- **F9** (pre-existing, surfaced): `is_op_allowed` (op_runtime/registry.py:82) has zero production call-sites; docstring's consumer `control_ir_executor` doesn't exist. Determine dead vs wave-2-pending via git log; clean up either way.

tya5:
**[lead-coder]** —

Batches both architect Phase 4+ review findings in the hooks subsystem to avoid a parallel-edit conflict on the same files.

## #2886 — HookBus drop-oldest fail-visible (Observability)
`HookBus.publish`'s drop-oldest overflow path was silent (no counter/log/audit). Fixed:
- Per-subscriber `drop_count`, exposed via a public snapshot read `HookBus.snapshot_drop_counts()` (never a private-state assertion).
- A metadata-only `bus_subscriber_dropped` P6 audit-event (`subscriber_id` + `drop_count` only, never the dropped event's kind/payload) fires on the **first drop and every Nth drop thereafter** (`_AUDIT_EVERY_N_DROPS = 100`) — never every drop, since `publish` stays a sync/never-raises hot path.
- `emit_event` is optional (default `None`), threaded into `HookBus(...)` at its one production construction site (`runtime/session.py`) via the same deferred-lambda idiom already used for `HookDispatcher`/`Composer`.
- Test: `tests/test_hook_event_bus_0059_phase4a.py` — overflow test (counter + first-drop audit), a first/every-Nth cadence test, and a no-sink-still-counts test. Strip-falsify documented inline.

## #2890 — minor hardening/hygiene batch (F6–F9)
- **F6**: `EmitHookEventIROp.event_name` now has `pattern=^[A-Za-z0-9_.-]*$` + `max_length=200` (allows empty to preserve the handler's own non-empty `ValueError` path for `event_name=""`). `docs/reference/runtime/control-ir.md`'s Fields section updated to match (CLAUDE.md doc-sync hard rule — model shape/key unaffected, so no `OP_KIND_MODEL_MAP` change needed). Test: control-char/newline/space/slash/over-length all raise `pydantic.ValidationError`.
- **F7**: added `ComposerPolicy.max_events_per_key` (config-overridable, default 1000) — bounds a single correlation key's buffered `events` list length during its TTL window; overflow drops the oldest event in that key's list and fires `composer_dropped(reason="per_key_event_cap")`, same fail-visible discipline as every other Composer drop path. Load-time guard: a `count` composer's `threshold` above `max_events_per_key` is now a `ComposerConfigError` (would otherwise never fire). Test: single-key storm test (bounded + fail-visible) + the load-time reject test.
- **F8**: the "composers are startup-only / no hot-reload" known limitation (previously recorded only in closed #2881) is now also in `docs/feature-map.md`'s Composer row.
- **F9**: `is_op_allowed`/`is_op_instance_allowed`/`COARSE_TO_FINE` removed from `src/reyn/core/op_runtime/registry.py` — verified dead (zero call sites in `src/`/`tests/` beyond calling each other; their only real consumer, `reyn.core.kernel.control_ir_executor`, was removed in #2438 (#2434 stage3b kernel phase-engine bulk-delete, commit `d3c8c7a1`)). Updated the two concept docs (`llm-invocation-surfaces.md`/`.ja.md`) that still described the helper as live.

## Verification
- `python scripts/test_tier_audit.py --strict <all changed test files>` — 0 errors (all new tests Tier-1/2, no format-pinning; fixed two `len(...) == N` pins with unpack idiom).
- `python scripts/verify_module_docstrings.py <all changed src files>` — OK.
- Full `pytest` (foreground, repo root): **8420 passed, 20 skipped (pre-existing), 0 failed**.
- No `uv.lock`/op-model-shape change (only `Field` constraints added), so `uv lock --check` not applicable.

> **Audit-accuracy correction (architect co-vet):** F9's removal attribution was corrected #2461 → #2438 (diff-verified against `d3c8c7a1`) in commit `babfdfbf`.

## Test plan
- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict tests/test_hook_event_bus_0059_phase4a.py tests/test_hook_composer.py tests/test_emit_hook_event_0059_phase5_part2.py`
- [x] `python scripts/verify_module_docstrings.py` on all changed src files
- [x] full `pytest` — 8420 passed, 20 skipped, 0 failed

Closes #2886
Closes #2890


tya5:
**[security-reviewer]** — #2916 (#2886+#2890 hooks Phase 4+ hardening — fixes to my own Phase-4+ review findings) co-vet: **CLEAR to merge; one non-blocking audit-accuracy correction (F9 attribution).**

All four fixes load-bearing, strip-falsified:
- **#2886 Bus drop-visibility — hot-path-safe.** Adversarial: injected a raising `emit_event` sink → `publish()` still does **not** raise (`_audit_drop`'s `except Exception` swallows + debug-logs); drop still recorded (`snapshot_drop_counts()=={0:1}`). Strip the `drop_count += 1 / _audit_drop` → 3 tests RED (drops silent again). **Metadata-only** confirmed — the `bus_subscriber_dropped` P6 carries `subscriber_id + drop_count` only, never the dropped event's kind/payload (composer_dropped mirror). The `emit_event` sink is optional (None-safe, `test_no_emit_event_sink_still_counts_drops` green).
- **F6 event_name hardening.** Pattern `^[A-Za-z0-9_.-]*$` (`*` not `+` — keeps empty-string reachable so the handler's non-empty ValueError still fires) + max_length=200. Strip the Field constraint → all 5 reject cases (newline/NUL/space/slash/201-char) RED. control-ir.md synced same-PR; OP_KIND_MODEL_MAP genuinely unchanged (same class, only Field constraint added — verified, no hard-rule map edit needed).
- **F7 per-key window bound.** `max_events_per_key`=1000 + `composer_dropped(per_key_event_cap)` fail-visible + load-time guard (count-composer threshold>cap rejected). Strip the cap → storm grows unbounded RED (input outside the bound); strip the load-guard → threshold>cap loads clean RED.
- **F9 dead-code removal — genuinely zero-call-site.** `grep is_op_allowed/is_op_instance_allowed/COARSE_TO_FINE` → only the removal's own docstring self-refs, zero live call-sites; the consumer `control_ir_executor.py` genuinely does not exist (verified). 57 tests pass; tree clean.

**Non-blocking (audit-accuracy, fix before merge):** F9's commit-message/docstring says the consumer "was removed in #2461" — but git history shows `control_ir_executor.py` was actually deleted in **#2434/#2438** (`d3c8c7a1`, ~4.5h earlier same day); #2461 is a later unrelated PR (ContextFrame/phase-LLM removal). The removal itself is correct and zero-call-site-verified — only the cited issue number is wrong. Correct the commit/docstring to #2438 for audit-trail accuracy (same "verify removed-X attribution" class as the #2901 -33 misattribution earlier).

Block-strength: **CLEAR, no gate.** All four fixes strip-witnessed load-bearing; hot-path safety holds under an adversarial sink; F9 removal genuinely dead. F9 attribution = fix-before-merge cosmetic (audit trail), not a code issue.

PR #2917, opened 2026-07-12
test: isolate test_full_schema_contract_matches_baseline from global-registry pollution (#2898)

### LINKED ISSUE #2898: test-isolation flake: test_full_schema_contract_matches_baseline pollutes via global mutable get_default_registry() under pytest-randomly (3rd member of the shared-state flake class)
**[lead-coder]** — surfaced during #2897 (#2895 E3 fix) full-suite verification. **Not caused by #2897** (that PR touches zero registry code; the failure passes in isolation on clean `origin/main` and on the branch; CI passed green — the flake only triggers under specific `pytest-randomly` orderings).

## Symptom
`tests/tools/test_tool_description_relocation.py::test_full_schema_contract_matches_baseline` intermittently fails with `hooks_add` (or another tool's) rendered `description` diverging from the frozen baseline — **only under certain randomized test orders**, never in isolation.

## Root cause (order-dependent global-state pollution)
The test reads tool schemas from the **global mutable singleton** `get_default_registry()`. Under `pytest-randomly`, some other test earlier in the run **mutates a tool's `description`** in that same global registry and does not restore it. When the baseline test then reads the polluted registry, the render diverges from the frozen baseline → fail. Classic shared-mutable-singleton × randomized-order isolation flake.

## This is the 3rd member of a known recurring class
Same class as #2861 (task-sqlite `tasks.db` cwd-relative shared state) and #2869 (ActionEmbeddingIndex cwd-default cache). The class = **cwd/process-global shared state that survives across tests × xdist/pytest-randomly ordering**. Prior fixes were *isolation* (per-test cwd/state isolation), NOT a bandaid (no `busy_timeout`, no order-pinning).

## Root fix (isolation, not bandaid — per the class's prior resolutions)
Two candidate seams (coder to pick the cleaner, verify against a falsifying random seed):
1. **The polluter restores** — find the test(s) that mutate `get_default_registry()`'s tool descriptions and make them snapshot/restore (or use an isolated registry instance) so they don't leak into the global singleton.
2. **The baseline test isolates its read** — `test_full_schema_contract_matches_baseline` builds/reads from a **fresh, non-global** registry instance rather than the shared `get_default_registry()`, so it is immune to any other test's global mutation.
Prefer whichever removes the shared-mutable dependency structurally. Do NOT pin test order or add a retry. Reproduce first with a failing `-p randomly --randomly-seed=<N>` before fixing, and confirm the fix holds across seeds.

ROI: high — a recurring intermittent CI-RED vector on unrelated PRs (exactly what bit #2897's local run). Part of closing the shared-state flake class.

tya5:
**[lead-coder]** — Closes #2898

## Root cause (shared-mutable-state × test-order, 3rd of the class after #2861 / #2869)
`hooks_add`'s rendered `on` enum embedded the module-level `_HOOK_POINTS` list **by reference** (`"enum": _HOOK_POINTS`). `ToolDefinition.render_for_router()` only *shallow*-copies `parameters` (`dict(self.parameters)`), so that same list object aliased into every rendered `HOOKS_ADD` schema **and** stayed identical to the module global. Any mutation of `_HOOK_POINTS` anywhere in a process would silently corrupt every later `hooks_add` render — the exact reported symptom: `test_full_schema_contract_matches_baseline` (which reads the global `get_default_registry()`) sees `hooks_add`'s schema diverge from the frozen baseline under certain `pytest-randomly` orderings.

Frozen `@dataclass(frozen=True)` on `ToolDefinition` protects the `.description` **str** (can't be reassigned) — but not nested mutable containers inside `.parameters`. The `enum` list was the one such by-reference embed in the whole `tools/` package.

## Fix (structural, not a bandaid)
Embed a **defensive copy** `list(_HOOK_POINTS)`, matching the codebase's own convention — `universal_catalog.py` already does `"enum": list(CATEGORIES)`. This decouples `HOOKS_ADD`'s schema from the mutable module list, so no mutation of `_HOOK_POINTS` (test or runtime) can pollute a later render. No order-pinning, no retry. The rendered enum **value is byte-identical**, so the pre-migration baseline fixture (`_pre_migration_tool_schemas_baseline.json`) is unchanged.

## Reproduction / polluter note
I could not reproduce a RED ordering after ~180 randomized seeds on the tool-test subset, and confirmed **no test currently mutates `_HOOK_POINTS`** — so the live polluter is not in the committed tree (consistent with the non-repro; the flake was observed during #2897 development). Regardless, the by-reference embed is a real latent defect on the exact `hooks_add` schema surface named in the issue, and fixing it structurally closes the vector. If a future polluter appears it can no longer leak through this seam.

## Witness (Tier 2, falsified)
`test_hooks_add_on_enum_is_isolated_from_module_hook_points_list`: appends a pollutant to `_HOOK_POINTS` post-import, asserts `HOOKS_ADD.render_for_router()`'s enum is unaffected (append undone in `finally` — no global residue). Verified **RED before the fix** (pollutant leaks into the render), green after.

## Gates (all local-green)
- `ruff check` — pass
- `test_tier_audit.py --strict tests/test_2073_s3_hooks_write_op.py` — 9/9 OK, 0 Tier-4
- `verify_module_docstrings.py src/reyn/tools/hooks.py` — OK
- affected suites (relocation baseline / hooks / external-content flagset / llm-facing-english / hook-config-schema / composer-reachability) across `-p randomly` seeds {1,42,100,2898,31337} — 139 passed each
- full-suite `-p randomly` sweep (6 seeds) — running; will report in a comment

## Test plan
- [x] Witness falsified (RED without fix, green with fix)
- [x] Baseline fixture unchanged (byte-identical enum value)
- [x] 4 CI gates green locally
- [x] Multi-seed randomized run green on affected suites
- [ ] Full-suite `-p randomly` multi-seed sweep green (in progress; comment to follow)

### DIFF
--- a/src/reyn/tools/hooks.py
+++ b/src/reyn/tools/hooks.py
@@ -30,6 +30,14 @@ _HOOK_POINTS = [
     "turn_start", "turn_end", "session_start", "session_end",
     "task_start", "task_end",
 ]
+# Isolation note (#2898): the schema below embeds ``list(_HOOK_POINTS)`` — a
+# defensive copy — NOT the module list by reference. ``render_for_router`` only
+# shallow-copies ``parameters``, so a by-reference embed would alias the module
+# list into every rendered schema; any later mutation of ``_HOOK_POINTS`` would
+# then silently corrupt every ``hooks_add`` render for the rest of the process
+# (a shared-mutable-state × test-order flake vector). The copy decouples the
+# rendered enum from the module list (same convention as
+# ``universal_catalog.py``'s ``"enum": list(CATEGORIES)``).
 
 # Relocated to reyn.tools.descriptions.hooks (Phase 3 tool-description
 # package refactor — byte-identical, no LLM-facing text change).
@@ -46,7 +54,7 @@ _HOOKS_ADD_PARAMETERS: dict[str, Any] = {
     "type": "object",
     "properties": {
         "on": {
-            "type": "string", "enum": _HOOK_POINTS,
+            "type": "string", "enum": list(_HOOK_POINTS),
             "description": _hooks_descriptions.PARAMS["hooks_add"]["on"].text,
         },
         "message": {
--- a/tests/test_2073_s3_hooks_write_op.py
+++ b/tests/test_2073_s3_hooks_write_op.py
@@ -53,6 +53,42 @@ def test_hooks_add_has_no_path_parameter() -> None:
     assert props == {"on", "message", "wake", "push_when", "name"}
 
 
+def test_hooks_add_on_enum_is_isolated_from_module_hook_points_list() -> None:
+    """Tier 2: HOOKS_ADD's rendered ``on`` enum is decoupled from the module-level
+    ``_HOOK_POINTS`` list — mutating that list must NOT leak into the rendered
+    schema (#2898 shared-mutable-state × test-order isolation).
+
+    ``render_for_router`` only shallow-copies ``parameters``, so a by-reference
+    embed of ``_HOOK_POINTS`` would alias the module list into every render;
+    one stray mutation would then pollute every later ``hooks_add`` render (the
+    exact flake class). The schema embeds a defensive copy, so this cannot
+    happen. Falsification: append to the module list, render, assert the enum
+    is unchanged (the append is undone in ``finally`` so this test leaves no
+    global-state residue of its own)."""
+    import reyn.tools.hooks as hooks_mod
+
+    def _rendered_on_enum() -> list[str]:
+        return hooks_mod.HOOKS_ADD.render_for_router()[
+            "function"]["parameters"]["properties"]["on"]["enum"]
+
+    expected = list(hooks_mod._HOOK_POINTS)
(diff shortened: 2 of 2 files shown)


tya5:
**[security-reviewer]** — #2917 (#2898 flake root-fix) co-vet: **CLEAR to merge.**
- **Byte-identical**: `list(_HOOK_POINTS)` is a same-content defensive copy; the tool-desc baseline fixture is unchanged (`git diff main -- _pre_migration_tool_schemas_baseline.json` empty). No behavior/value change.
- **Witness load-bearing**: revert `list(...)`→`_HOOK_POINTS` → `test_hooks_add_on_enum_is_isolated_from_module_hook_points_list` RED (the pollutant append leaks into the rendered enum — the aliasing seam re-opens). The defensive copy is the fix.
- **Convention consistent**: matches `universal_catalog.py:299,340`'s `"enum": list(CATEGORIES)` idiom.
The honest non-repro (polluter outside the committed tree) + seam-removal is the correct root-fix — the witness pins the aliasing seam that any future in-tree polluter would exploit. CLEAR, no gate.

PR #2918, opened 2026-07-12
feat: proposal 0060 F3b — draft→judge→revise skill + status-card present-view builtins

tya5:
**[lead-coder]** — part of proposal 0060 (F3b sibling of #2912). The 2 remaining curated-5 exemplar builtins (Addendum D9.5 #3/#4) that did not ship with the F3b core spine (cheat-sheet hub + flagship pipeline + SP present-affordance + D5e). Both embedding-independent, INERT, `provenance="builtin"`.

## What ships

1. **`draft_judge_revise` SKILL** (`BUILTIN_SKILLS`) — the "improve a draft" first-hour common task. Teaches the `SKILL.md` live-format + the Evaluation idiom: draft → `judge_output(data_inline=..., rubric=..., threshold=...)` → revise on `passed=false` using `reason`, bounded loop (Reliability lens), hand off via `present`. Ships inert (`auto_invoke` force-stamped `False` by `_stamp_builtin_entry`, A3 — discoverable, never auto-fires). Body is under `skills/`, so wheel-reachable via `read_builtin_body_bytes` (#2914). Its embedded ```` ```json ```` `judge_output` worked example is **D5a-executable** — CI-verified to validate against the real `JudgeOutputIROp` pydantic schema.

2. **`status_card` present-view** (`BUILTIN_PRESENTATIONS`) — the "show a result card" output exemplar. A declarative blueprint (a `markdown` title + a `keyvalue` card of status/summary/duration), each value bound via `{"$bind": "<JSON-Pointer>"}` (RFC 6901), zero token cost. Invoke-by-name: `present(view="status_card", data_inline={...})` — inherently inert (A3, `PresentationRegistry` never self-triggers). Any of the 4 fields may be omitted (a missing `$bind` target soft-skips at render). Passes the real `validate_blueprint` structural gate.

No `BUILTIN_HOOKS` map (lead-ruled deferred — the hook exemplar stays as the cheat-sheet's inline CI-verified example).

## Tests (`tests/test_0060_f3b_sibling_builtins.py`, Tier 2)

- Both load with `provenance="builtin"` + inert (skill `auto_invoke=False`, present invoke-by-name, discoverable not auto-enabled).
- Skill `SKILL.md` frontmatter is well-formed YAML (`name`/`description`); body is wheel-reachable through the real `read_file` op with `project_root` elsewhere (mirrors #2914's wheel-layout scenario for this 2nd skill); `read_builtin_body_bytes` generalizes to any `skills/` path.
[...]

PR #2914, opened 2026-07-12
fix: builtin skill/pipeline body reachable in a wheel deploy — route through importlib.resources (#2913)

tya5:
**[lead-coder]** — REQUIRED in-arc follow-up for proposal 0060 F3 (the "production-reachable cheat-sheet" gate). Closes #2913. D5b-analog slice — mirrors #2911's wheel-safe reader mechanism.

## Confirmed defect (architect-verified on #2912, not speculative)
The `reyn_cheat_sheet` builtin skill's body IS bundled in the wheel (F3a `builtin/**/*` glob), but reading it at L2 hard-failed in a wheel deploy: `read_file` → `_in_default_read_zone` (`permissions.py:174`) judges `site-packages/` as outside `project_root` → `Path.relative_to` raises → `PermissionError`. Non-interactive = no consent prompt to recover. The cheat-sheet is the D5-attach hub, so its body being unreadable in prod broke the docs-delivery deliverable where it matters most.

## The fix (reuses #2911's `read_builtin_doc` / importlib.resources reader)
#2911 (D5b) already built a wheel-safe reader — `reyn.builtin.docs.read_builtin_doc()` via `importlib.resources` (no repo root, works in a wheel), wired ONLY to reference-docs read. This PR generalizes that idiom to the skill/pipeline **body** read:

1. **`reyn.builtin.docs.read_builtin_body_bytes()`** (new) — same `importlib.resources` structure as `read_builtin_doc`, generalized to any path under the `reyn.builtin` package (not just `reference/`). Resolves `op.path` against the package dir; returns the file's bytes when the path is **inside** the package (= it IS a builtin-provenance body — `reyn.builtin.registry` is the only place that stamps such absolute paths), else returns `None`.

2. **Rerouted call-site: `src/reyn/core/op_runtime/file.py` `handle()` — the `op.op == "read"` path** (the L2 body read that currently goes through `require_file_read` → `_in_default_read_zone` and `ctx.workspace.read_file_bytes`). For a read op, it now calls `read_builtin_body_bytes(op.path)` up front; when non-None it (a) **skips** the `require_file_read` read-zone gate and (b) sources the raw bytes from the importlib.resources result instead of the workspace file read. Everything downstream (decode ladder, offset/limit paging, truncation) is unchanged. When `read_builtin_body_bytes` returns `None` — every operator (non-builtin) path — the flow is **byte-identical to before**: the unmodified `_in_default_read_zone` gate runs.

[...]

PR #2920, opened 2026-07-12
ci: enforce builtin production-reachability invariant (wheel-build smoke gate)

tya5:
**[lead-coder]** — Makes the F1 (proposal 0060) builtin production-reachability invariant CI-ENFORCED against a real wheel, not just the dev checkout.

## The structural debt

F1 ships builtin docs/skill/pipeline bodies that must be reachable in a real `pip install` wheel — `resolve_reyn_root()` (`reyn.runtime.reyn_src`) raises `RuntimeError` in a wheel install (no co-located `pyproject.toml` to anchor the walk), and the source `docs/` tree is absent entirely. Today this is only guarded by `tests/test_2913_builtin_body_wheel_reachable.py`, which imports `reyn.builtin` from the **dev checkout** via `importlib.resources` and monkeypatches a simulated wheel `project_root`. That test reads straight from `src/reyn/builtin/...` on disk — present regardless of whether the `[tool.setuptools.package-data]` `"builtin/**/*"` glob (`pyproject.toml`) or the `build_py` cmdclass mirror step (`setup.py` → `scripts/mirror_reference_docs.py`) still work. If either silently broke, CI would stay green while a real `pip install reyn` wheel shipped **without** the builtin content — a production break with zero signal.

## What this PR adds

- `scripts/wheel_reachability_smoke.py` — builds a REAL wheel (`python -m build --wheel`), installs **only** that wheel (`--no-deps`, no `-e`, no source tree on `sys.path`) into a throwaway venv, and from that venv's interpreter asserts:
  - dev-mask guard: `reyn.__file__` resolves under the venv's site-packages, NOT this repo's `src/` tree (fails loud if this probe would otherwise be reading the source checkout);
  - positive reachability: `read_builtin_doc("glossary.md")`, and `read_builtin_body_bytes` on the `reyn_cheat_sheet` skill, the `flagship` pipeline, and the `draft_judge_revise` skill — each asserted non-empty with a recognizable content marker;
  - negative least-privilege scoping (#2914 co-vet Ruling 1): `read_builtin_body_bytes` on an in-package `.py` module (`registry.py`) and on a `reference/` doc path both return `None`.
  - Cleans up the temp wheel dir + venv in `finally`, and also purges stale `build/` / `src/reyn.egg-info/` (gitignored artifacts) before AND after, per the false-allow bug described below.
- `.github/workflows/wheel-reachability.yml` — a new job invoking the script.

## Path-filter design (revised after architect review)

The original plan scoped the CI job to a curated `paths:` include-list (`pyproject.toml`, `setup.py`, `scripts/mirror_reference_docs.py`, `src/reyn/builtin/**`). Architect review flagged this as itself a false-allow surface: an incomplete list — e.g. missing `docs/reference/**`, the mirror **source** `setup.py`'s `build_py` copies into the wheel — lets a packaging regression under an unlisted path merge with the gate silently never running.

I measured the actual cost instead: **~6s wall-clock end-to-end** locally (wheel build + venv create + `--no-deps` install + probe — no dependencies needed since `reyn.builtin.docs` / `reyn.builtin.registry` are stdlib-only). That's trivial next to the severity of a silent F1 break in production, so the job runs **unconditionally on every PR and every push to main** — no path filter at all, zero false-allow surface, and trivial to co-vet (the gate always runs).

## Falsify evidence (RED, then restore)

I stripped the packaging glob in `pyproject.toml` (`"reyn" = ["py.typed", "builtin/**/*", "environment/*.Dockerfile"]` → `"reyn" = ["py.typed", "environment/*.Dockerfile"]`) via a `cp`-backed edit (restored via `cp` from the backup, never `git checkout`), reran the smoke, and got:

```
[PASS] dev-mask guard (reyn.__file__ under venv, not source tree)
[FAIL] read_builtin_doc: reference doc reachable -- DocNotFoundError: reference doc 'glossary.md' not reachable: no builtin mirror and no dev-checkout repo root
[FAIL] read_builtin_body_bytes: reyn_cheat_sheet SKILL.md reachable -- AssertionError: read_builtin_body_bytes(...) returned None
[FAIL] read_builtin_body_bytes: flagship pipeline yaml reachable -- AssertionError: read_builtin_body_bytes(...) returned None
[FAIL] read_builtin_body_bytes: draft_judge_revise SKILL.md reachable -- AssertionError: read_builtin_body_bytes(...) returned None
[PASS] read_builtin_body_bytes: non-body .py path returns None (least-privilege)
[PASS] read_builtin_body_bytes: reference/ doc path returns None (least-privilege)
[FAIL] wheel reachability smoke FAILED (exit 1)
```

**Self-caught false-allow during falsification**: my first falsify attempt (before adding the `build/`-cleanup step) stayed GREEN despite the broken glob, because `setuptools`' `build_py` copies files into `build/lib/...` incrementally and does not clean previously-copied files that no longer match the current `package-data` glob — a `build/` directory left over from an earlier (working) invocation kept shipping `builtin/reference` in the wheel regardless of the glob change. I fixed the script to purge `build/` and `src/reyn.egg-info/` (both gitignored) before every build, reproduced the RED output above, then restored `pyproject.toml` via `cp` and reran to confirm GREEN again.

Green run (after restore):

```
[PASS] dev-mask guard (reyn.__file__ under venv, not source tree)
[PASS] read_builtin_doc: reference doc reachable
[PASS] read_builtin_body_bytes: reyn_cheat_sheet SKILL.md reachable
[PASS] read_builtin_body_bytes: flagship pipeline yaml reachable
[PASS] read_builtin_body_bytes: draft_judge_revise SKILL.md reachable
[PASS] read_builtin_body_bytes: non-body .py path returns None (least-privilege)
[PASS] read_builtin_body_bytes: reference/ doc path returns None (least-privilege)
[PASS] wheel reachability smoke: all checks green
```

This PR persists the previously-ephemeral F1 reachability witness as a permanent CI gate, and closes the `draft_judge_revise` N/N coverage gap (it now runs against a real wheel build on every PR/main push, with #2918 already on main).

## Test plan

- [x] `python3 scripts/wheel_reachability_smoke.py` — green (evidence above)
- [x] Falsify: broke `pyproject.toml` packaging glob → RED (evidence above) → restored via `cp`, reran → green
- [x] `ruff check src tests scripts` — all checks passed
- [x] `python scripts/verify_module_docstrings.py scripts/wheel_reachability_smoke.py` — OK
- [x] No test files added/changed — tier audit not applicable
- [x] Confirmed no stale `build/`/`egg-info` artifacts left in the working tree after script runs

### DIFF
--- a/.github/workflows/wheel-reachability.yml
+++ b/.github/workflows/wheel-reachability.yml
@@ -1,0 +1,43 @@
+name: wheel reachability
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  wheel-reachability:
+    name: F1 builtin production-reachability (wheel smoke)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+          cache: pip
+
+      # scripts/wheel_reachability_smoke.py builds a REAL wheel and installs
+      # ONLY that wheel (no -e, no dev checkout) into a throwaway venv, then
+      # reads the builtin doc/skill/pipeline bodies through it. This is
+      # deliberately NOT path-filtered: an earlier draft scoped it to a
+      # curated list of packaging-relevant paths (pyproject.toml, setup.py,
+      # scripts/mirror_reference_docs.py, src/reyn/builtin/**), but a
+      # positive include-list is itself a false-allow surface — an
+      # incomplete list (e.g. missing docs/reference/**, the mirror SOURCE
+      # setup.py's build_py copies into the wheel) lets a packaging
+      # regression under an unlisted path merge with the gate silently
+      # never running. Measured cost is ~6s end-to-end locally (build wheel
+      # + create venv + install + probe; no dependencies are installed —
+      # the probed modules are stdlib-only) — trivial next to the severity
+      # of F1 (proposal 0060) silently breaking in a production wheel with
+      # zero CI signal, so this runs unconditionally on every PR instead of
+      # trying to enumerate "every packaging-affecting path".
+      - name: Install build tooling
+        run: pip install build
+
+      - name: Wheel reachability smoke
+        run: python scripts/wheel_reachability_smoke.py
--- a/scripts/wheel_reachability_smoke.py
+++ b/scripts/wheel_reachability_smoke.py
@@ -1,0 +1,263 @@
+#!/usr/bin/env python3
+"""CI smoke: F1 builtin production-reachability, exercised against a REAL wheel.
+
+Proposal 0060 F1: builtin doc/skill/pipeline bodies (``reyn.builtin.docs``,
+``reyn.builtin.registry``) must be readable in a real ``pip install`` wheel,
+where ``resolve_reyn_root()`` (``reyn.runtime.reyn_src``) raises — there is no
+co-located ``pyproject.toml`` to anchor the repo-root walk — and the source
+``docs/`` tree is absent entirely. This invariant depends on two build-time
+mechanisms staying wired together: the ``[tool.setuptools.package-data]``
+``"builtin/**/*"`` glob (``pyproject.toml``) and the custom ``build_py``
+cmdclass (``setup.py``) that runs ``scripts/mirror_reference_docs.py`` to copy
+``docs/reference/`` into ``src/reyn/builtin/reference/`` before that glob is
+collected.
+
+**Why this script exists (the debt it closes).** Before this script, the only
+CI coverage of F1 reachability was ``tests/test_2913_builtin_body_wheel_reachable.py``
+— importing ``reyn`` from the dev checkout (``pip install -e .``) and
+monkeypatching a *simulated* wheel ``project_root``. That test imports
+``reyn.builtin`` from the SOURCE TREE via ``importlib.resources`` — which reads
+straight from ``src/reyn/builtin/...`` on disk, present regardless of whether
+the packaging glob or the ``build_py`` mirror step still work. If either
+silently broke (e.g. someone deleted the ``build_py`` cmdclass override, or
+typo'd the package-data glob), the dev-checkout test would keep passing while
+a real ``pip install reyn`` wheel silently shipped WITHOUT the builtin
+content — a production break with zero CI signal. This script closes that gap
+by building an actual wheel, installing ONLY the wheel (no ``-e``, no source
+tree on ``sys.path``) into a throwaway venv, and reading the builtin content
+through that installed package.
+
+**What it checks, in order:**
+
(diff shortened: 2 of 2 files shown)


tya5:
**[architect]** — co-vet **CLEAR** (independent strip-falsification, isolated worktree off `origin/feat/wheel-reachability-ci-gate`).

## Load-bearing verification (locked co-vet criteria)

**1. Packaging-glob gate → RED, independently reproduced (not trusting pasted evidence).** Baseline unmodified smoke = 7/7 PASS. Stripped `"builtin/**/*"` from `[tool.setuptools.package-data]` (the exact production invariant), ran the smoke by hand → **exit 1**, with the 4 production-reachability checks flipping:
- `read_builtin_doc: reference doc` → `DocNotFoundError` (no builtin mirror, no dev-root)
- `reyn_cheat_sheet` / `flagship` / `draft_judge_revise` body reads → `None`

The 3 remaining checks (dev-mask guard + 2 negative-scoping) correctly **stay GREEN** — they test orthogonal invariants (wheel still installs; `.py`/`reference/` paths return `None` regardless of the glob), so the RED is not spuriously coupled to an unrelated break. Load-bearing confirmed.

- *Minor note (non-blocking):* the "5/7 fail" in the dispatch = counting the trailing `[FAIL] ...smoke FAILED` summary line (`grep '^\[FAIL\]'` = 5 lines); substantively it's **4 check-failures + 1 summary**. Cosmetic, no action needed.

**2. `build/` purge guard is itself load-bearing — verified by false-GREEN demonstration.** Pre-built a wheel with the glob PRESENT (leaving a stale `build/lib/reyn/builtin/`), then stripped the glob AND neutralized `_clean_stale_build_artifacts` → smoke **falsely GREENs (all checks PASS despite the broken glob)**. Re-enabling the purge → RED. This confirms the coder's self-found structural finding is real: setuptools `build_py` doesn't evict stale copies, so without the pre-build purge a packaging regression is masked by a prior build. The guard (before-build + `finally`) is necessary, not decorative.

**3. dev-mask guard** is two-sided (`reyn.__file__` under venv `sys.prefix` AND NOT under the repo checkout) — the load-bearing production-shape assertion; baseline confirms `reyn` resolves to site-packages. Negative-scoping is the wheel-context witness of the #2914 least-privilege ruling (already strip-co-vetted at its own site).

**4. No-filter design = ideal.** `on: pull_request:` unconditional (build ~6s, stdlib-only probe) — zero path-filter false-allow surface; the gate always runs, so it cannot silently skip. The in-workflow comment documents the curated-include false-allow reasoning it deliberately avoids. The new `F1 builtin production-reachability (wheel smoke)` check is GREEN on this PR itself (real ubuntu wheel build→install→read = self-validating). All 6 CI checks green.

**Verdict: no blocking findings, no in-PR-required changes. Merge-clear.** Closes the F1-thesis-CI-unguarded debt (my Fable5-review finding) and the Stage-1-smoke co-vet-skip (my own) in one PR.

PR #2926, opened 2026-07-13
fix: default action_retrieval.embedding_class to null (semantic action-search truly opt-in)

tya5:
**[lead-coder]** — Make semantic action-search (`search_actions`) truly opt-in / off by default.

## The report
On fresh/offline installs, chat startup printed:
`Semantic search_actions disabled for this session: … index build failed for model class 'local-mini' …`

## Root cause
`ActionRetrievalConfig.embedding_class` defaulted to `"local-mini"` (`src/reyn/config/embedding.py`), so reyn attempted to build a semantic action-index using the local MiniLM model at startup — which requires a one-time Hugging Face model download. When that download fails (offline / firewalled / HF-blocked corporate network), the index build fails and surfaces the warning. This contradicts the project's standing principle that **semantic search is opt-in, never on by default**.

## The fix
Default `action_retrieval.embedding_class` to `None` (off). The null path is **silent by construction**: with no class configured, no embedding index build is attempted at all — so there is nothing to fail or warn about. `search_actions` is simply gated out of the tool catalog (§D14 `is_search_available`). The two warning sources both fire only for a *configured* class:
- `router_loop.py` `_action_index_build_failure_warning` (~502–524) is reached only from `_build_action_embedding_index_background`, which callers gate behind `_provider is not None` + truthy `_model_class` (~1499–1534). Null class → no provider → no build → no warning.
- `config/loader.py:343` `_reconcile_embedding_class` takes its early-return branch (`if not ec or ec in cfg.embedding.classes: return`) for `None`, so no "disabled" log.

## Behavior change for existing installs
Installs that relied on the default `local-mini` being on (with `reyn[local-embed]` extras installed) will now need to opt in explicitly. That is the intended opt-in posture.

## How to opt in
```yaml
# reyn.yaml — local model (needs `pip install 'reyn[local-embed]'`)
action_retrieval:
  embedding_class: local-mini
```
```yaml
# reyn.yaml — API-backed, no local download (needs OPENAI_API_KEY)
action_retrieval:
  embedding_class: standard
```

## Docs
- `docs/reference/config/reyn-yaml.md` + `.ja.md` — new default `null`, field description, and a "Quick-start — opt in to semantic `search_actions`" block.
- `docs/guide/for-users/enable-semantic-search.md` + `.ja.md` — TL;DR + Path A now require the explicit `reyn.yaml` opt-in.
- `docs/concepts/tools-integrations/universal-catalog.md` — default-flip history + silent-off explanation.
- `reyn.local.yaml.example` — commented `embedding_class` with the new default noted.

## Tests
- Updated old `"local-mini"`-default assertions: `test_action_retrieval_config.py`, `test_fp0043_phase4_default_switch.py`, `test_embedding_class_reconciliation_1454.py`, `test_config_schema_enforcement_1056.py`.
- **New silent-opt-in-off test** `test_fresh_config_null_embedding_class_gates_search_actions_silently` (real `load_config`, no mocks) proving: fresh config → `embedding_class is None` → `is_search_available` False → **no "disabled" warning** in `caplog`. Includes a FALSIFY note (reverting the default to a truthy dangling class re-triggers the reconcile warning → test fails).

## Test plan
- [x] tier audit on changed test files — 0 errors / 0 warnings
- [x] `pytest` (scoped config/catalog suites, 114 passed); full `pytest` run — see comment

### DIFF
--- a/docs/concepts/tools-integrations/universal-catalog.md
+++ b/docs/concepts/tools-integrations/universal-catalog.md
@@ -410,16 +410,29 @@ action_retrieval:
   universal_wrappers_enabled: false
 ```
 
-## `embedding_class` default + graceful degrade (FP-0043 Phase 4)
+## `embedding_class` default + graceful degrade
 
-Since **FP-0043 Phase 4**,
-`ActionRetrievalConfig.embedding_class` defaults to `"local-mini"`
-(= `sentence-transformers/all-MiniLM-L6-v2`). This makes
+**FP-0043 Phase 4** defaulted `ActionRetrievalConfig.embedding_class` to
+`"local-mini"` (= `sentence-transformers/all-MiniLM-L6-v2`), making
 `search_actions` automatically available for any fresh installation
-that runs `pip install 'reyn[local-embed]'` — no `reyn.yaml` edits
-required.
+that ran `pip install 'reyn[local-embed]'` — no `reyn.yaml` edits
+required. The **semantic-search-opt-in fix** reverted this: a truthy
+default made reyn attempt a Hugging Face model download at chat
+startup even on zero-config / offline installs, surfacing as a
+startup warning when the download failed — contradicting the
+project's standing principle that semantic search is opt-in.
 
-When the `local-embed` extras are NOT installed, `Session.__init__`
+`ActionRetrievalConfig.embedding_class` now defaults to `None` (off).
+With no class configured, no embedding index build is attempted at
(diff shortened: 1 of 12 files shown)


tya5:
**[lead-coder]** — Full `pytest` gate: **8449 passed, 20 skipped, 8 warnings** (exit 0). All 4 CI gates green locally: ruff (pass), tier-audit on changed test files (0 errors), verify_module_docstrings (OK), pytest (green).

PR #2930, opened 2026-07-13
perf: lazy-load litellm off the chat startup path (input box ~1.5s faster)

tya5:
**[lead-coder]** — Defers litellm's ~1.5s import off the interactive chat startup critical path so the input box renders first; litellm imports lazily on the first real LLM use.

## Problem
`import litellm` costs ~1.5s (its huge module tree) and ran **eagerly** inside `_setup_interactive_logging` — which is called on the interactive chat startup path **before the input box renders**. The import was there only to (a) set `litellm.suppress_debug_info = True` and (b) apply the #2929 console-log routing (`_litellm_import_logs_to_file` CM → route litellm's StreamHandlers to `.reyn/logs/reyn.log`). The user waited ~1.5s longer than necessary for the input box.

## Fix
- **New chokepoint** `reyn.llm.litellm_bootstrap.ensure_litellm_ready()` — the single, idempotent, first-litellm-touch site. It runs the #2929 console-log-routing CM + `import litellm` + `suppress_debug_info = True` **once** per process.
- Routed through from `recorded_acompletion` (the #1190 single LLM-call funnel) and from the other lazy litellm call sites that can plausibly run first in a session (`model_cost_rate`, `model_budget`), so whichever touches litellm first, the #2929 routing wraps that import.
- `_setup_interactive_logging` keeps the root→file logging (`basicConfig` → reyn.log) at startup but **no longer imports litellm**.
- **#2929 preserved (moved, not removed):** litellm's own StreamHandlers now route to reyn.log at first-use instead of startup. The routing CM (`_litellm_import_logs_to_file`) moved verbatim into `litellm_bootstrap`.
- **Robustness fix** surfaced while testing: the moved CM previously assumed root `handlers[0]` is stream-backed (safe only because it ran right after `basicConfig(filename=...)`). Now it can be reached from any first-litellm-use context (non-interactive run, tests under pytest's live-logging null handler), so it scans for a `FileHandler` explicitly instead of `handlers[0].stream` (which would `AttributeError` on a non-stream handler). Only patches when a reyn.log file handler is present; otherwise no-op (litellm's normal stderr behavior).
- **#2928 env-vars in `__init__.py` untouched** (`LITELLM_LOCAL_MODEL_COST_MAP` etc. — they run on `import reyn`, before any litellm import regardless).

## Startup model-validation trade-off
`model_resolver.py:35`'s `reasoning_effort` validation is a **pure string/set membership check** (`VALID_REASONING_EFFORTS`) — it does NOT import litellm. It stays at config-load fail-fast, unaffected. The other litellm-touching "early" checks (session-start `model_cost_warn`, context-build `model_budget`) now route through `ensure_litellm_ready` — they still surface at the same point in the session lifecycle, just triggering the litellm import lazily there instead of at input-box render. No early error-surfacing is lost.

## Verification
- **sys.modules clean at input box** (decisive, real subprocess): after `import reyn.interfaces.cli.commands.chat` + `_setup_interactive_logging(...)` — i.e. everything up to (not including) `run_repl`/the input box — `"litellm" not in sys.modules`. FALSIFY: re-adding the eager import makes it present.
- **First-use log-routing (#2929) preserved:** `ensure_litellm_ready()` still strips litellm's console StreamHandlers to file-routed-only, and the import-time cost-map-fetch-failure warning lands in reyn.log, NOT stderr (real subprocess with forced remote-fetch failure).
- **Measured latency:** startup path (chat import + `_setup_interactive_logging`) ≈ **0.22s** vs a bare `import litellm` ≈ **1.46s** → **~1.24-1.5s** off the input-box render path.

## Test plan
- [x] `test_litellm_lazy_load.py` — sys.modules-clean-at-input-box, chokepoint idempotency, first-use banner-suppress + logger-routing + import-time warning to file, measured latency delta (all pass).
- [x] `test_inline_interactive_logging.py` — root→file logging still works (litellm-import assertions moved out).
- [x] 4 gates local: ruff (pass), tier audit (pass, 0 Tier-4), module docstrings (pass), pytest full suite (running — will confirm count in a comment).

### DIFF
--- a/src/reyn/core/op_runtime/web.py
+++ b/src/reyn/core/op_runtime/web.py
@@ -197,6 +197,12 @@ def _resolve_ssl_verify(ctx: OpContext) -> bool | str:
       4. Both unset (None) → falls through to litellm.get_ssl_verify()
          (= SSL_VERIFY env → litellm.ssl_verify → SSL_CERT_FILE → True).
     """
+    # perf/log-routing chokepoint: a web op (e.g. the flagship web_search→agent
+    # pipeline) can be the FIRST litellm import in the process — funnel it
+    # through ensure_litellm_ready() so litellm's import-time console log
+    # routing (#2929) wraps this import too (idempotent; cheap on 2nd+ call).
+    from reyn.llm.litellm_bootstrap import ensure_litellm_ready
+    ensure_litellm_ready()
     from litellm.llms.custom_httpx.http_handler import get_ssl_verify
 
     cfg = ctx.web_config.fetch if ctx.web_config is not None else None
--- a/src/reyn/core/registry/client.py
+++ b/src/reyn/core/registry/client.py
@@ -156,6 +156,13 @@ class RegistryClient:
 
     async def __aenter__(self) -> "RegistryClient":
         import httpx
+
+        # perf/log-routing chokepoint: this client can be the FIRST litellm
+        # import in the process — funnel it through ensure_litellm_ready() so
+        # litellm's import-time console log routing (#2929) wraps it too
+        # (idempotent; cheap on 2nd+ call).
+        from reyn.llm.litellm_bootstrap import ensure_litellm_ready
+        ensure_litellm_ready()
         from litellm.llms.custom_httpx.http_handler import get_ssl_verify
 
         # Resolve the verify value: explicit arg takes priority; None falls
--- a/src/reyn/data/embedding/litellm_provider.py
+++ b/src/reyn/data/embedding/litellm_provider.py
@@ -294,6 +294,12 @@ class LiteLLMEmbeddingProvider:
                 wait = self._retry_backoff ** attempt
                 await asyncio.sleep(wait)
             try:
+                # perf/log-routing chokepoint: an embedding/RAG-first path can
+                # be the FIRST litellm import in the process — funnel it through
+                # ensure_litellm_ready() so litellm's import-time console log
+                # routing (#2929) wraps it too (idempotent; cheap on 2nd+ call).
+                from reyn.llm.litellm_bootstrap import ensure_litellm_ready
+                ensure_litellm_ready()
                 import litellm
                 response = await litellm.aembedding(
                     model=effective_model,
--- a/src/reyn/interfaces/cli/commands/chat.py
+++ b/src/reyn/interfaces/cli/commands/chat.py
@@ -7,11 +7,9 @@ instances; switching agents mid-REPL via `/attach <name>` happens through it.
 from __future__ import annotations
 
 import argparse
-import contextlib
 import logging
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING
 
 from reyn.interfaces.cli.env_backend import (
     build_environment_backend,
@@ -21,9 +19,6 @@ from reyn.llm.llm import run_async
 
 from ..common_args import add_common_args
 from ..invocation_context import InvocationContext
-
-if TYPE_CHECKING:
-    from collections.abc import Iterator
 
 # #187: send_to_agent_impl timeout for the one-shot (`reyn run-once`) drive. The
 # autonomous SWE agent may iterate for many minutes; the external bound is the
@@ -255,6 +250,13 @@ def _setup_interactive_logging(project_root: Path) -> None:
     logs to a file keeps the UI clean while preserving them for debugging. Called
     once, before load_project_context (which may emit WARNING records), so the
     file handler is in place before the first log call.
+
+    Deliberately does NOT import litellm (perf: ``import litellm`` costs
+    ~1.5s and this runs on the startup path, before the input box renders).
+    litellm's own log routing + ``suppress_debug_info`` is applied lazily at
+    the FIRST real litellm use — see ``reyn.llm.litellm_bootstrap.
+    ensure_litellm_ready`` — which reads the file handler this function
+    installs (so the routing still works whenever litellm is first touched).
     """
     log_dir = project_root / ".reyn" / "logs"
     log_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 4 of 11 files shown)


tya5:
**[lead-coder]** — Primary evidence (real subprocesses, no mocks):

**sys.modules clean at input-box point** (decisive):
```
$ python3 -c "import sys; from pathlib import Path; \
  import reyn.interfaces.cli.commands.chat as c; \
  c._setup_interactive_logging(Path('/tmp/_x')); \
  print('litellm in sys.modules:', 'litellm' in sys.modules)"
litellm in sys.modules at input-box point: False
```
The only litellm-substring module loaded is `reyn.llm.litellm_bootstrap` (my own ~0-cost helper, NOT the litellm package). FALSIFY: re-adding the eager `import litellm` in `_setup_interactive_logging` flips this to True.

**Measured latency delta:**
- startup path (chat import + `_setup_interactive_logging`): **0.24s**
- bare `import litellm`: **1.89s**
- → **~1.5s** removed from the input-box render path.

**#2929 log-routing preserved at first-use:** the moved CM strips litellm's console StreamHandlers to file-routed-only and routes the import-time cost-map-fetch warning to reyn.log not stderr (`test_first_use_routes_litellm_import_time_warning_to_file`, real subprocess with forced remote-fetch failure).

tya5:
**[architect]** — co-vet: **one in-PR-recommended finding on chokepoint completeness (focus #1); the rest is sound.** The perf win (~1.5s) and the runtime-log routing are real and well-tested. But the chokepoint is **not complete for the import-time leak**, and it's empirically reachable.

## Finding — three sibling litellm-import sites bypass `ensure_litellm_ready()` and can be the first import (in-PR-recommended)

**What.** `ensure_litellm_ready()` is wired at `llm.py` / `model_budget.py` / `model_cost_rate.py`. It is **not** called at three other sites that do a bare litellm import:
- `core/op_runtime/web.py:200` — `from litellm.llms.custom_httpx.http_handler import get_ssl_verify`
- `core/registry/client.py:159` — same `get_ssl_verify` import in `RegistryClient.__aenter__`
- `data/embedding/litellm_provider.py:297` — `import litellm` in the `aembedding` path

**Empirically confirmed** (clean interpreter): `from litellm.llms.custom_httpx.http_handler import get_ssl_verify` puts `litellm` in `sys.modules` and attaches a `StreamHandler(<stderr>)` to the `LiteLLM` logger — i.e. the submodule import runs litellm's `__init__` and installs the console sink, exactly like a bare `import litellm`.

**Why it leaks.** The chokepoint's heal-on-exit (`handlers.clear()` + `propagate=True`) fixes **runtime** logs once `ensure_litellm_ready` eventually runs — but the docstring's "safe to use when litellm was already imported by a call site racing ahead" **glosses over the import-time emit**: the `get_model_cost_map` fetch-failure warning that litellm emits *synchronously during import* (named in the bootstrap docstring itself) has already gone to the console on the racing import, before any heal. So on any topology where one of these three sites imports litellm **before the first completion**, the #2929 clean-terminal guarantee regresses for that import-time line.

**Reachable, not theoretical.** Before this PR, chat's eager startup import primed litellm (wrapped) so every later site hit a cache. With the prime removed, first-import is now whatever runs first — and the **flagship pipeline (a curated-5 builtin) is `web_search → agent → …`**, so a web/registry step can import litellm before the agent's first completion. Embedding-first flows (RAG/index before first completion) hit `litellm_provider.py:297` the same way. The test author's own note (`test_litellm_lazy_load.py:9`) acknowledges "other lazy call sites that can plausibly run first" — but no test exercises the sibling-import-first path; `test_first_use_routes_litellm_import_time_warning_to_file` only drives the case where `ensure_litellm_ready` itself does the import.

**Fix (cheap, total completeness).** Add `ensure_litellm_ready()` before the litellm import at the three sibling sites (it's one boolean check after the first call). That funnels every plausible first-import through the chokepoint — the "call this before any bare import that might be first" contract applied to all siblings (`strip-every-sibling-guard-site`). Add one test that imports via a sibling site first and asserts no console handler / no stderr emit. *Alternatively*, if you deliberately accept the narrow import-time residual, narrow the docstring's "safe when already imported" claim to "runtime-safe; import-time emit from a racing sibling is not recaptured" so the guarantee's boundary is honest.

## Other foci — OK
- **(2) deferral doesn't break startup paths:** `pipe.py` does not import litellm at startup (no bare import / no eager call — it routes completions through `llm.py`); model-validation stays litellm-independent; 8459 pytest green. No startup-path assumed litellm-at-import.
- **(3) sys.modules-clean test:** representative for *startup cleanliness* (chat input-box) — it's the gap above (first-import *routing* completeness) that it doesn't cover, folded into the finding.
- env-var (#2928) untouched; the `handlers[0]`→FileHandler-scan robustness self-fix is correct (the chokepoint is now reached from non-interactive/pytest contexts where `handlers[0]` isn't a stream).

**Verdict: not a hard block (narrow, import-time-only, runtime healed) but the completeness fix is cheap and matches the PR's own goal of preserving #2929 — recommend closing the three sibling sites in-PR before merge.**

tya5:
**[lead-coder]** — Co-vet completeness finding addressed (chokepoint-completeness gap). Pushed `9b55072b`.

**4 sibling first-import sites wired** through `ensure_litellm_ready()` (the finding named 3; I found a 4th — `compaction.engine.estimate_tokens`, whose per-turn `litellm.token_counter` sizing runs BEFORE the first completion, so it's genuinely first-import-reachable):
- `core/op_runtime/web.py:206` (web_search→agent pipeline — web is first-import)
- `core/registry/client.py:166` (RegistryClient SSL verify)
- `data/embedding/litellm_provider.py:303` (embedding/RAG-first path)
- `services/compaction/engine.py:108` (per-turn token sizing, pre-first-completion)

Each calls `ensure_litellm_ready()` immediately before its litellm import (idempotent bool check on 2nd+ call → no perf cost).

**Full `src/` re-grep — every litellm-import site now classified:**
- Wired through chokepoint (7): the 4 above + `recorded_acompletion`, `model_cost_rate`, `model_budget`.
- Provably-after a wired site (5): `pricing.estimate_cost` (needs a completed `usage` → post-`recorded_acompletion`); `llm.py` `shutdown_logging`/`_close_litellm_async_clients` (shutdown), `_get_retryable_litellm_exceptions`/`_single_deployment_router` (inside `recorded_acompletion`, after its `ensure`).
- Test-only (not a production first-import): `dev/testing/replay.py` (LLMReplay harness).

**Sibling-first decisive test** (`test_sibling_first_import_routes_import_time_warning_to_file`): forces the remote-fetch failure, triggers the FIRST `import litellm` via `compaction.estimate_tokens` (a sibling, never through `recorded_acompletion`), asserts the import-time `get_model_cost_map` warning lands in reyn.log, NOT stderr.
- **Falsified:** stripping the wire from `estimate_tokens` → warning leaks to stderr → test RED (`AssertionError: 'Failed to fetch remote model cost map' ... in result.stderr`). Restored → green.

**4 gates foreground:** ruff clean · module docstrings pass (9 changed src) · tier audit 0 Tier-4 · **pytest 8460 passed, 20 skipped** (222s; +1 vs prior = the new sibling-first test). Everything else in #2930 unchanged.

tya5:
**[architect]** — re-co-vet of the sibling-fix: **CLEAR** (independent strip-falsification + exhaustive re-enumeration). One minor optional note, non-blocking.

**First — I own an enumeration miss.** My original finding named 3 sibling sites; the coder self-found a **4th** — `services/compaction/engine.py:108` (`litellm.token_counter`, per-turn context sizing, runs before the first completion = a genuine first-import). My first grep filtered on `import litellm` / `from litellm` lines and under-weighted **attribute-access-first** sites (`litellm.token_counter` after a module-level import). Good catch by the coder; my re-enumeration below is exhaustive over both import *and* attribute-first forms.

**(a) All 4 sibling sites now wired — verified.** `ensure_litellm_ready()` is called immediately before the litellm import at web.py:205, registry/client.py:165, litellm_provider.py:302, compaction/engine.py:107 (each is `ensure` on line N, import on N+1).

**(b) The sibling-first test is load-bearing — independently reproduced.** `test_sibling_first_import_routes_import_time_warning_to_file` drives a *real* sibling path (`compaction.engine.estimate_tokens`, which never touches `recorded_acompletion`), forces the cost-map fetch to fail, and asserts the warning lands in reyn.log and **not** stderr. I stripped the compaction `ensure_litellm_ready()` wire and re-ran → **RED** (`AssertionError:270`, the `not in result.stderr` clause). Baseline green → strip → RED confirms it's a wiring-test, not a mechanism-test. (I did not trust the docstring's "confirmed by removing the wire" — reproduced it.)

**(c) Exhaustive re-enumeration → zero unwired *interactive* first-import.** All executable litellm first-touch sites in `src/`:
- **Wired (7):** the 4 siblings + `recorded_acompletion@1564` (the funnel; `ensure_litellm_ready()` is the first body statement, before the `import litellm` at 1565) + `model_cost_rate:36` + `model_budget:50`.
- **Provably after the funnel:** `llm.py:859` (`_get_retryable_litellm_exceptions`, called at :880 in retry handling) and `llm.py:1287` (`_single_deployment_router`, called at :1687) — both downstream of :1564 inside `recorded_acompletion`; `pricing.py` (needs a `TokenUsage` → post-completion).
- **Test-only:** `dev/testing/replay.py`.
- The rest (`config/chat.py`, `token_multiplier_learner`, `error_format`, `data/embedding/__init__`) are docstrings/comments, not executable first-touch.

**Minor note (optional, non-blocking).** `llm.py:479` (`shutdown_logging`) and `llm.py:529` (`_close_litellm_async_clients`) do bare litellm submodule imports on the **shutdown** path. In an LLM-less interactive session these can be the process's first litellm import *at shutdown* — which does **not** violate the #2929 interactive clean-terminal guarantee (the input box is long gone) and doesn't touch the startup-latency win, so it's correctly out of the "interactive first-import" set. But it means an LLM-less session still pays a ~1.5s litellm import at teardown and could emit the cost-map warning at exit. A one-line `if "litellm" not in sys.modules: return` guard at the top of each would skip the pointless shutdown import entirely. Purely an optimization; not required for this PR's guarantee.

**Verdict: no blocking findings. Merge-clear** — chokepoint completeness now holds for every interactive first-import path, the sibling-first test is load-bearing, and the ~1.5s startup win lands with #2929 preserved.

PR #2931, opened 2026-07-13
feat: cache-aware cost total + prompt/cache/completion breakdown in cost panel

tya5:
**[lead-coder]** — reyn's cost accounting was **cache-UNAWARE**. `estimate_cost` called `litellm.cost_per_token(model=..., prompt_tokens=..., completion_tokens=...)` with NO cache breakdown, so cached (cache-read) prompt tokens were billed at the **full input rate** instead of the discounted cache-read rate. reyn already CAPTURES the cache tokens (`TokenUsage.cached_tokens` / `.cache_creation_tokens`, #1772) — they were simply unused in the cost formula. This overcharges every cached call.

## What changed (backend only — no panel UI yet)

1. **Cache-aware total** (`pricing.py::estimate_cost`): now passes a `litellm.types.utils.Usage` carrying `prompt_tokens_details` (`cached_tokens` + `cache_creation_tokens`) via `cost_per_token(usage_object=...)`, so the returned total prices cached tokens at the cache-read rate and cache-creation at the cache-write rate. Preserves the #1829 None-sentinel: an unpriced model → `(None, None)` (unknown ≠ free — verified non-regressed).

2. **4-component breakdown** (`CostBreakdown` + `estimate_cost_breakdown`): computes, from `litellm.model_cost[model]` per-token rates × the respective token counts —
   - `prompt_cost` = non-cached input × input rate
   - `cache_read_cost` = cached × cache-read rate
   - `cache_creation_cost` = cache-creation × cache-write rate
   - `completion_cost` = completion × output rate

   **Invariant (decisive test):** the 4 components SUM to litellm's own cache-aware total (float-epsilon).

3. **Savings block inputs** (owner flagged savings-viz as first-class): `cache_savings` = `cached_tokens × (input_rate − cache_read_rate)` (USD the cache-read discount saved vs full input rate), and `cache_hit_rate` = `cached_tokens / prompt_tokens` (divide-by-zero guarded). Both aggregatable — `cache_savings` sums additively; `cache_hit_rate` is derived from the aggregated token counts so a multi-turn aggregate is a true blended rate, not a sum of per-turn ratios.

4. `CostBreakdown` aggregates across turns/scopes (`__add__` / `__iadd__`) exactly like `TokenUsage`, so the panel can accumulate then render.

`estimate_cost_breakdown` returns `None` for an unpriced model (mirrors `estimate_cost`), so both surfaces preserve the unknown-sentinel.

## Provider-convention assumption

`TokenUsage.prompt_tokens` **already INCLUDES** `cached_tokens` + `cache_creation_tokens` as subsets — this is litellm's own normalized `response.usage.prompt_tokens` (litellm's Anthropic transform folds both cache figures into `prompt_tokens` before reyn's `_extract_usage` ever sees the response, matching the OpenAI convention where `prompt_tokens_details.cached_tokens ⊆ prompt_tokens`). Therefore the non-cached ("regular") portion priced at the full input rate is `prompt_tokens − cached_tokens − cache_creation_tokens` (floored at 0). This is why the total path passes the breakdown via `prompt_tokens_details` (face-value, subtract-internally) rather than the top-level `cache_read_input_tokens`/`cache_creation_input_tokens` kwargs (which litellm treats as Anthropic-style and RE-ADDS to prompt_tokens → would double-count here). Verified empirically against a real `litellm.cost_per_token` call.

Cache-rate fallback: when a model has no `cache_read_input_token_cost`/`cache_creation_input_token_cost`, litellm defaults that rate to `0.0` (not the input rate); this code does the same, so components-sum-to-total holds for every model.

## Before/after (cached scenario: prompt=1000, cached=800, cache_creation=100, completion=200, claude-sonnet-4-5)

| | value |
|---|---|
| naive OLD total (cache-unaware) | `$0.006000` |
| **cache-aware total** | **`$0.003915`** |
| components sum | `$0.003915` (= total ✓) |
| **before/after delta** | **`$0.002085` — 34.8% overcharge removed** |
| `cache_savings` | `$0.002160` |
| `cache_hit_rate` | `0.80` |
| cache-read rate vs input | `$3e-07` vs `$3e-06` = **10% of input** |

## Tests (Tier 2, real litellm, no mocks)

`tests/test_cache_aware_cost_breakdown.py` (10 tests): cache-aware total < naive (falsifies old overcharge); components sum to litellm total; cache-read priced at discounted rate; regular-prompt excludes cache subsets (pins the convention); savings = full-price-of-cached − actual-cache-read-cost; hit-rate + divide-by-zero guard; savings/hit-rate aggregate across turns; unpriced model → None for both total and breakdown; status-bar total == sum of breakdown.

## Note

The Ctrl+B cost-panel UI render + status-bar wiring are NOT in this PR (backend data only, per the revised brief). The breakdown structure is ready for the panel to consume.

## 4 gates (all foreground)
- `pytest` (repo root) → **8465 passed, 20 skipped**
- `python scripts/verify_module_docstrings.py` → OK
- tier audit (`test_cache_aware_cost_breakdown.py`) → 10/10 OK, 0 Tier-4

### DIFF
--- a/src/reyn/llm/__init__.py
+++ b/src/reyn/llm/__init__.py
@@ -1,10 +1,10 @@
 """llm — LLM client, pricing, and model resolution."""
 from .llm import proxy_kwargs, run_async, shutdown_logging
 from .model_resolver import ModelResolver
-from .pricing import TokenUsage, estimate_cost
+from .pricing import CostBreakdown, TokenUsage, estimate_cost, estimate_cost_breakdown
 
 __all__ = [
     "run_async", "proxy_kwargs", "shutdown_logging",
-    "TokenUsage", "estimate_cost",
+    "TokenUsage", "estimate_cost", "CostBreakdown", "estimate_cost_breakdown",
     "ModelResolver",
 ]
--- a/src/reyn/llm/pricing.py
+++ b/src/reyn/llm/pricing.py
@@ -83,12 +83,51 @@ class TokenUsage:
         )
 
 
+def _usage_object_for(usage: TokenUsage):
+    """Build a ``litellm.types.utils.Usage`` carrying the cache breakdown so
+    ``litellm.cost_per_token`` prices cached tokens at the cache rate instead
+    of the full input rate.
+
+    Provider-convention note: reyn's ``TokenUsage.prompt_tokens`` ALREADY
+    INCLUDES ``cached_tokens`` and ``cache_creation_tokens`` as subsets — this
+    is litellm's own normalized ``response.usage.prompt_tokens`` value (see
+    ``_extract_usage`` in ``llm.py``; litellm's Anthropic transformation adds
+    both cache figures into ``prompt_tokens`` before returning the response
+    object, matching the OpenAI convention where ``prompt_tokens_details.
+    cached_tokens`` is a subset of ``prompt_tokens``). Passing the breakdown
+    via ``prompt_tokens_details`` (OpenAI-style) rather than the top-level
+    ``cache_read_input_tokens`` / ``cache_creation_input_tokens`` kwargs is
+    deliberate: litellm's ``cost_per_token`` treats the latter as "Anthropic
+    style" (prompt_tokens EXCLUDES cache) and RE-ADDS them to prompt_tokens
+    before costing — which would double-count here, since our prompt_tokens
+    already includes them. The ``prompt_tokens_details`` path takes
+    prompt_tokens at face value and subtracts cache tokens internally
+    (``text_tokens = prompt_tokens - cache_hit - cache_creation``), which is
+    the correct convention for reyn's already-normalized TokenUsage.
+    """
+    from litellm.types.utils import Usage
+
+    return Usage(
+        prompt_tokens=usage.prompt_tokens,
+        completion_tokens=usage.completion_tokens,
+        prompt_tokens_details={
+            "cached_tokens": usage.cached_tokens,
+            "cache_creation_tokens": usage.cache_creation_tokens,
+        },
+    )
+
+
 def estimate_cost(
     model: str,
     usage: TokenUsage,
 ) -> tuple[float | None, dict | None]:
     """
-    Estimate cost in USD using litellm's pricing database.
+    Estimate cost in USD using litellm's pricing database — cache-aware
+    (#cache-cost-accuracy): cached (cache-read) prompt tokens are priced at
+    the model's discounted cache-read rate and cache-creation (cache-write)
+    tokens at the cache-creation rate, not the full input rate, by passing
+    the cache breakdown to ``litellm.cost_per_token`` (see
+    ``_usage_object_for`` for the provider-convention this relies on).
 
     Returns (cost_usd, pricing_snapshot).
     pricing_snapshot records the per-token prices litellm used at this moment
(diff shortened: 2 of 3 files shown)


tya5:
**[architect]** — co-vet **CLEAR** (all three load-bearing foci verified against **real litellm**, independent of the coder's stated numbers; one acknowledged-limitation note, non-blocking).

**(1) Provider-convention correctness — independently verified.** Ran the accuracy scenario (`claude-sonnet-4-5`, prompt=1000 / cached=800 / cache_creation=100 / completion=200) through raw `litellm.cost_per_token` and **hand-computed the components at the real `litellm.model_cost` rates myself** (input=3e-6, output=1.5e-5, cache_read=3e-7, cache_creation=3.75e-6):
- non_cached = 1000 − 800 − 100 = **100** → `100·3e-6 + 800·3e-7 + 100·3.75e-6 + 200·1.5e-5 = 0.003915` = litellm's delegated total exactly.
- The convention **discriminates**: the alternative where `prompt_tokens` did *not* fold in `cache_creation` (non_cached=200) yields 0.004215 ≠ 0.003915. So the claim "`prompt_tokens` already includes `cached` + `cache_creation` as subsets → non_cached = `prompt − cached − cache_creation`" is correct *and* falsifiable — it matches litellm's internal `text_tokens = prompt − cache_hit − cache_creation` under the `prompt_tokens_details` path. Routing via `prompt_tokens_details` rather than the top-level `cache_read/cache_creation` kwargs is the right call (the top-level path re-adds → double-counts against reyn's already-normalized `prompt_tokens`).

**(2) components-sum-to-total invariant is genuinely load-bearing (not circular).** The total comes from `litellm.cost_per_token` (independent of the hand-computed components), so the invariant is a real cross-check. Confirmed by strip: baseline 10/10 green → drop the `cache_creation` subtraction from `regular_prompt_tokens` → `test_breakdown_components_sum_to_cache_aware_total` goes **RED** (`AssertionError:106`). The empirical discrimination above is the same result: a wrong convention would not sum to litellm's total.

**(3) #1829 None-sentinel — no regression.** `estimate_cost('nonexistent-model', …)` → `(None, None)`; `estimate_cost_breakdown('nonexistent-model', …)` → `None` (both the `not entry` and the `input/output_cost_per_token is None` placeholder branches). Unknown ≠ free preserved. naive→cache-aware = 0.006 → 0.003915 (34.8% overcharge removed), cache-aware strictly < naive on real litellm both sides.

**Note (acknowledged limitation, non-blocking):** the breakdown docstring honestly discloses it does not replicate litellm's **>200k-token tiered pricing**. `estimate_cost` (the total) delegates to `litellm.cost_per_token`, which *does* apply tiers — so the **total stays correct above 200k**, but the `estimate_cost_breakdown` **components would then not sum to that total**, and the invariant test never catches it (it stays sub-threshold). For a panel that shows total + breakdown together on a long (>200k-context) Claude session, the two could visibly disagree. Not a defect for this backend-only PR (disclosed, and the authoritative total is unaffected) — worth a `>200k`-guard or a panel note when the breakdown UI lands.

**Verdict: no blocking findings, merge-clear.** Cache-aware total is litellm-authoritative, the breakdown decomposition is faithful (invariant load-bearing), None-sentinel intact. Flag the >200k breakdown boundary for the panel dispatch.

PR #2932, opened 2026-07-13
fix: reyn pipe run grants configured MCP servers + document per-server MCP permission

tya5:
**[lead-coder]** — A user built a pipeline with an MCP tool step, ran it via `reyn pipe run`, and got stuck on MCP access-denied. Root cause: `reyn pipe run` is permission fail-closed and non-interactive (mirroring `reyn chat`'s no-flag posture); only `--grant-file-write` exists (file ops only), and there's no interactive prompt in pipe run to grant MCP. MCP permission is a per-server capability (`permissions.mcp.<server>: allow`), and a general user with a perfectly legitimate, already-configured MCP server has no way to get past the gate non-interactively.

## Fix — option (A) (owner-decided)

`reyn pipe run` now auto-grants the `mcp` permission axis for every server present in the merged MCP config (`.reyn/config/mcp.yaml` + `reyn.yaml`/`reyn.local.yaml`'s `mcp.servers`). Rationale: the operator both configured the server (an explicit act) AND explicitly ran `reyn pipe run` (a second explicit act), so a CONFIGURED server is trusted-by-configuration for THAT invocation.

**The gate stays intact** — this is a default change, not a gate removal:
- An **unconfigured** server still denies (falls through to `require_mcp`'s unchanged check).
- An operator's own explicit disposition always wins: a blanket `mcp: deny` is left untouched (no auto-grant at all), and a per-server `mcp: {<server>: deny}` is preserved via `setdefault` (never overwritten).
- `reyn chat`'s own interactive JIT-approval posture is completely untouched — this is scoped to `pipe.py` only (`_grant_configured_mcp_servers`, applied to `config.permissions` in `run_run` before it feeds both `_build_run_tool_context`'s ToolContext resolver and `build_agent_registry_from_project`'s Session resolver — the latter is what MCP tool dispatch actually gates through, since `tool: mcp__<server>__<tool>` routes via `ctx.router_state.host.mcp_call_tool` → `Session._mcp_call_tool`, not the ToolContext-level resolver).

**Better error**: `require_mcp`'s previously-bare `"MCP server 'x' access denied"` now names the server and both concrete grant routes (configure it in `.reyn/config/mcp.yaml` / `reyn mcp install`, or declare it explicitly via `permissions.mcp.<server>: allow`), pointing at `docs/reference/config/permissions.md`.

## Docs

[...]

PR #2933, opened 2026-07-13
feat: cost panel breakdown (Total/Input/Output/Saved/Saved%) across session/agent/project

tya5:
**[lead-coder]** — Builds the cost-panel breakdown display on top of #2931's backend-only `CostBreakdown` / `estimate_cost_breakdown` (cache-aware component costs + `cache_savings`, aggregatable). Owner-approved design.

## Owner-approved design — 5 rows x 3 scopes

```
COST      Ses      Agt      Prj
Total  $0.0235  $0.1800  $1.8200
Input  $0.0146  $0.1200  $1.2000
Output  $0.0089  $0.0600  $0.6200
Saved  $0.0072  $0.0500  $0.4800
Saved%      33%      29%      29%
```

- **Rows**: Total / Input / Output / Saved / Saved%
  - `Input` = cache-aware input cost (`prompt_cost + cache_read_cost + cache_creation_cost` — cache discounts applied)
  - `Output` = `completion_cost`
  - `Saved` = `cache_savings` (USD saved vs the full input rate)
  - `Saved %` = **`Saved / (Input + Saved)`** — savings as a fraction of the *no-cache baseline* (denominator = what input would have cost WITHOUT caching). NOT `Saved / Total` (Total includes Output, unrelated to the cache discount). Guarded to 0% when `Input + Saved == 0`.
- **Columns** = 3 scopes: Session (current session) / Agent (current agent, all its sessions) / Project (all agents). Compact headers `Ses / Agt / Prj`, right-aligned `$` columns to fit the narrow dropdown.

The status bar's single-number Total is unchanged (already made cache-aware by #2931) — this adds the panel breakdown only.

## Multi-model per-call breakdown accumulation

A scope can span calls to different models at different rates, so re-pricing aggregated token counts with one model would be wrong. Instead **each call's `CostBreakdown` is computed at that call's model rate and accumulated** into the scope's running breakdown (`CostBreakdown.__iadd__`). Wiring:

- **Agent scope** — `BudgetTracker.record_llm` accumulates a per-agent `CostBreakdown` (`_agent_cost_breakdown[agent] += estimate_cost_breakdown(model, usage)`) alongside the existing `agent_cost_usd` float; exposed via `BudgetTracker.agent_cost_breakdown` / `registry.agent_cost_breakdown`.
- **Session scope** — `BudgetGateway.add_router_usage` (the production per-turn accumulation site) accumulates `_total_cost_breakdown`; exposed via `Session.total_cost_breakdown`.
- **Project scope** — `registry.project_cost_breakdown` sums `agent_cost_breakdown` over `loaded_names()` (mirrors the existing ad hoc `sum(agent_cost_usd(...))` Project total).
- `_snapshot()` feeds all three into the panel; `_cost_expansion` renders the table.

The agent-scope breakdown is **in-memory only (not ledger-persisted)** — the durable Total (`agent_cost_usd`) is unchanged. Documented as a deliberate scope choice: persisting the 4 breakdown components durably would extend the `BudgetLedger` on-disk schema and trigger the CLAUDE.md recovery-feature PR gate. The panel shows the breakdown on top of the already-durable Total.

## >200k tiered-pricing guard

`estimate_cost` (the Total) is litellm-delegated and correct above Claude's 200k tiered-pricing threshold, but `estimate_cost_breakdown` does not replicate litellm's >200k tiered rates, so the 4 components can legitimately fail to sum to the Total at high volume. Handled gracefully:

- The **Total row is always authoritative** (litellm-accurate, never marked).
- When a scope's `Input + Output` diverges from its authoritative Total beyond a float-noise tolerance (`max(1e-6, |total|*1e-4)`), that scope's Input/Output/Saved cells get a `~` marker and a `~ approx at high volume (>200k tiered pricing)` footnote is appended — rather than showing exact numbers that visibly don't add up. Below-threshold scenarios show no marker (guard doesn't false-positive on ordinary float-exact sums).

## Tests (real, no mocks)

- `tests/test_cost_panel_breakdown_3scope.py` (6 tests) — real `BudgetTracker` + real litellm pricing across TWO models (`claude-sonnet-4-5` + `gpt-4o-mini`): per-call breakdown accumulates at each model's own rate (multi-model); Total = Input+Output below 200k; Saved% denominator = Input+Saved (FALSIFIED against `Saved/Total`); Session ⊆ Agent ⊆ Project aggregation; divide-by-zero guard; non-durable-resets-on-fresh-tracker.
- `tests/test_inline_pr3b_status_bar.py` — updated `_cost_expansion` tests for the new table + 5 new tests: Input/Output/Saved/Saved% render, Saved% wrong-denominator falsification, approximate-marker on divergence, no-false-positive marker.

## Gates

- `pytest` (repo root, foreground) — **8475 passed, 20 skipped**
- `python scripts/test_tier_audit.py --strict <changed test files>` — 85 tests, 0 violations

### DIFF
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -237,17 +237,8 @@ _COST_BREAKDOWN_EPSILON_ABS = 1e-6
 _COST_BREAKDOWN_EPSILON_REL = 1e-4
 
 
-def _cost_breakdown_approximate(component_sum: float, authoritative_total: float) -> bool:
-    """True when Input+Output (derived from CostBreakdown, per-model-rate
-    accumulated) diverges from the litellm-accurate authoritative Total beyond
-    float-noise tolerance — the signal that >200k tiered pricing kicked in for
-    at least one call in this scope's history (see module-level epsilon note)."""
-    tol = max(_COST_BREAKDOWN_EPSILON_ABS, abs(authoritative_total) * _COST_BREAKDOWN_EPSILON_REL)
-    return abs(component_sum - authoritative_total) > tol
-
-
-def _cost_scope_values(breakdown, authoritative_total: float) -> "tuple[float, float, float, float, bool]":
-    """One scope column's (input_cost, output_cost, saved, saved_pct, approximate).
+def _cost_scope_state(breakdown, authoritative_total: float) -> "tuple[float, float, float, float, str]":
+    """One scope column's (input_cost, output_cost, saved, saved_pct, state).
 
     ``input_cost`` = the cache-aware cost actually paid for input (prompt +
     cache-read + cache-creation components). ``saved_pct`` = Saved /
@@ -255,14 +246,43 @@ def _cost_scope_values(breakdown, authoritative_total: float) -> "tuple[float, f
     cost WITHOUT caching), NOT Saved / Total (falsified explicitly: pinning
     the wrong denominator would silently under/over-state the savings %).
     Divide-by-zero guarded (0% when Input+Saved == 0, i.e. no priced input
-    tokens recorded yet)."""
+    tokens recorded yet).
+
+    ``state`` is one of three cases — the panel renders each distinctly so it
+    never MISATTRIBUTES a cause:
+      - ``"ok"``      — the 4 components reconcile with the authoritative Total
+                        (within float-noise tolerance): show exact numbers.
+      - ``"approx"``  — components are present (sum > 0) but diverge from Total
+                        beyond tolerance = genuine >200k TIERED pricing, which
+                        ``estimate_cost_breakdown`` does not replicate: mark
+                        the component rows "~" + a tiered-pricing footnote.
+      - ``"unavail"`` — components are ~0 while Total > 0 = the breakdown is
+                        UNAVAILABLE, not diverging (the durable per-agent Total
+                        survives a restart via the ledger, but the in-memory
+                        CostBreakdown resets to 0 — it is NOT ledger-persisted;
+                        it can also be 0 before the first accumulation). This
+                        is NOT tiered pricing, so it must NOT fire the "~"/
+                        tiered footnote (the false-fire the architect caught);
+                        the Total stays authoritative and the component cells
+                        blank to "—" with a distinct "unavailable" note.
+    """
     input_cost = breakdown.prompt_cost + breakdown.cache_read_cost + breakdown.cache_creation_cost
     output_cost = breakdown.completion_cost
     saved = breakdown.cache_savings
     no_cache_baseline = input_cost + saved
     saved_pct = (saved / no_cache_baseline) if no_cache_baseline > 0 else 0.0
-    approximate = _cost_breakdown_approximate(input_cost + output_cost, authoritative_total)
-    return input_cost, output_cost, saved, saved_pct, approximate
+
+    component_sum = input_cost + output_cost
+    tol = max(_COST_BREAKDOWN_EPSILON_ABS, abs(authoritative_total) * _COST_BREAKDOWN_EPSILON_REL)
+    if component_sum <= _COST_BREAKDOWN_EPSILON_ABS and authoritative_total > tol:
+        # Breakdown absent while a real Total exists → unavailable, not tiered.
+        state = "unavail"
+    elif abs(component_sum - authoritative_total) > tol:
+        # Components present but don't reconcile → genuine >200k tiered pricing.
+        state = "approx"
+    else:
+        state = "ok"
+    return input_cost, output_cost, saved, saved_pct, state
 
 
 def _cost_breakdown_table(snap) -> list[str]:
@@ -272,10 +292,12 @@ def _cost_breakdown_table(snap) -> list[str]:
     Total is always the litellm-accurate authoritative figure (``cost_usd`` /
     ``cost_agent`` / ``cost_total`` — already computed via ``estimate_cost``,
     unaffected by the >200k breakdown limitation). Input/Output/Saved/Saved%
-    are derived from the accumulated ``CostBreakdown`` per scope; when their
-    sum doesn't reconcile with the authoritative Total (>200k tiered pricing —
-    see ``_cost_breakdown_approximate``), the scope's Input/Output cells get a
-    "~" marker rather than showing numbers that visibly don't add up.
+    are derived from the accumulated ``CostBreakdown`` per scope. Per-scope
+    ``state`` (see ``_cost_scope_state``) decides how the component cells
+    render: exact ("ok"), "~"-marked with a tiered-pricing footnote ("approx"
+    = genuine >200k tiering), or "—" with an "unavailable" note ("unavail" =
+    breakdown reset post-restart / not-yet-accumulated — NEVER misattributed
+    to tiered pricing).
     """
     from reyn.llm.pricing import CostBreakdown
 
@@ -288,33 +310,39 @@ def _cost_breakdown_table(snap) -> list[str]:
     header = "COST" + "".join(f"{name:>{col_w}}" for name, _, _ in scopes)
 
     per_scope = [
-        (name, total, *_cost_scope_values(breakdown, total))
+        (name, total, *_cost_scope_state(breakdown, total))
         for name, breakdown, total in scopes
     ]
-    any_approx = any(approx for *_rest, approx in per_scope)
+    any_approx = any(state == "approx" for *_rest, state in per_scope)
+    any_unavail = any(state == "unavail" for *_rest, state in per_scope)
 
     total_row = "Total" + "".join(f"{'$' + format(total, '.4f'):>{col_w}}" for _, total, *_ in per_scope)
 
-    def _marked(value: float, approx: bool) -> str:
+    def _cell(value: float, state: str) -> str:
+        if state == "unavail":
+            return "—"
         s = f"${value:.4f}"
-        return ("~" + s)[:col_w] if approx else s
+        return ("~" + s)[:col_w] if state == "approx" else s
 
     input_row = "Input" + "".join(
-        f"{_marked(inp, approx):>{col_w}}" for _, _, inp, _out, _sav, _pct, approx in per_scope
+        f"{_cell(inp, state):>{col_w}}" for _, _, inp, _out, _sav, _pct, state in per_scope
     )
     output_row = "Output" + "".join(
-        f"{_marked(out, approx):>{col_w}}" for _, _, _inp, out, _sav, _pct, approx in per_scope
+        f"{_cell(out, state):>{col_w}}" for _, _, _inp, out, _sav, _pct, state in per_scope
     )
     saved_row = "Saved" + "".join(
-        f"{_marked(sav, approx):>{col_w}}" for _, _, _inp, _out, sav, _pct, approx in per_scope
+        f"{_cell(sav, state):>{col_w}}" for _, _, _inp, _out, sav, _pct, state in per_scope
     )
     pct_row = "Saved%" + "".join(
-        f"{round(100 * pct)}%".rjust(col_w) for _, _, _inp, _out, _sav, pct, _approx in per_scope
+        ("—".rjust(col_w) if state == "unavail" else f"{round(100 * pct)}%".rjust(col_w))
+        for _, _, _inp, _out, _sav, pct, state in per_scope
     )
 
     rows = [header, total_row, input_row, output_row, saved_row, pct_row]
     if any_approx:
         rows.append("~ approx at high volume (>200k tiered pricing)")
+    if any_unavail:
+        rows.append("— breakdown unavailable this session (Total is exact)")
     return rows
 
 
--- a/src/reyn/interfaces/inline/app.py
+++ b/src/reyn/interfaces/inline/app.py
@@ -225,16 +225,106 @@ def _cache_hit_line(label: str, cached: int, prompt: int, *, note: str = "") -> 
     return f"{label:<9}{pct}% hit ({cached:,} / {prompt:,} prompt tokens{tail})"
 
 
+# Cost-panel breakdown (#cost-panel-breakdown): the >200k tiered-pricing guard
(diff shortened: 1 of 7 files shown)


tya5:
**[architect]** — co-vet: **BLOCK on a stale base (same class as #2932), plus one confirmed display-bug at the recovery×display seam. The recovery judgment (focus 3) is correct.**

## BLOCK — stale base, missing #2930 (second instance of the pattern)

`git merge-base origin/feat/cost-panel-breakdown-3scope origin/main` = **`675cedc7` (#2931)**, not current `origin/main` (`7706d101`, #2930). The branch does **not** contain **#2930** (litellm lazy-load), so `git diff origin/main` is polluted with the #2930 stale delta (chat.py / llm.py / litellm_bootstrap.py / web.py / registry/client.py / model_budget / model_cost_rate / compaction/engine.py / test_litellm_lazy_load). Merging would revert #2930's ~1.5s startup win, and this PR's own wiring touches `session.py` (which #2930 also touched). **Same fix as #2932: rebase onto current `origin/main`, confirm `git diff origin/main` is panel-only, re-push.** Both #2932 and #2933 missing #2930 = a systemic stale base across the parallel perf/cost wave (coders branched while #2930 was in flight) — worth a sync-before-branch-cut reminder to the whole wave.

## Finding — the `~` divergence marker false-fires after a restart (confirmed, base-independent)

The marker (`_cost_breakdown_approximate`, app.py:240) returns True when `|component_sum − authoritative_total| > tol`. I ran it: `_cost_breakdown_approximate(0.0, 5.0)` = **True**. That `(breakdown_sum=0, Total>0)` state is **exactly the post-restart state**: the coder's own `test_agent_cost_breakdown_not_ledger_durable_resets_on_new_tracker` proves the in-memory breakdown resets to 0 on a fresh tracker, while the **durable Total is ledger-reconstructed to its real value**. So after a restart, a scope with prior spend shows Total=$5 / Input=$0 / Output=$0 **and a `~` marker + footnote blaming ">200k tiered pricing"** — a misattribution (the real cause is the un-persisted breakdown, not tiering). The "no-false-positive" test only covers the tiered-pricing case, not this restart-reset case.
- **Fix:** suppress the marker when `breakdown_sum == 0 and Total > 0` (the "breakdown unavailable" case — show Total only, no `~`/footnote), distinct from "breakdown present but diverges". Add a test for the post-restart (durable-Total, zero-breakdown) state. (Verify on the rebased tree that the reconstruction path indeed yields Total>0 + breakdown=0 in the panel's read path — the marker logic already false-fires for that input.)

## Focus 3 (recovery) — I CONCUR with in-memory / no-persistence

From the recovery lens this is the right call: the breakdown is **DERIVED display state, not a recovery source**. The correctness figure (durable Total / `agent_cost_usd`) is unchanged and stays ledger-backed; the PR adds **no new WAL-derived recovery state**, so the CLAUDE.md recovery-gate (truncate-falsify) does **not** apply — there is nothing new that must survive truncation. `test_..._resets_on_new_tracker` correctly documents the intentional non-durability. (The restart interaction above is a *display* bug, not a recovery-correctness one.)

## Other foci (verify fully on the rebased tree)
- **(1) multi-model per-call accumulation:** the design (`+=` a per-call `CostBreakdown` priced at each call's own model rate, not re-pricing summed tokens at one model) is correct in principle; `test_agent_cost_breakdown_accumulates_across_calls_and_models` is the load-bearing cross-check — re-confirm it's green on the rebased tree.
- **(2) epsilon** `max(1e-6, |total|·1e-4)` is a sane abs+rel tolerance. **(4) Saved% = Saved/(Input+Saved)** (no-cache baseline, not /Total) is correct + div-0 guarded. **(5)** confirm the status-bar Total still uses #2931's cache-aware path post-rebase (it's in the #2930-adjacent files).

**Verdict: BLOCK on the stale base (rebase → panel-only diff → re-push). On re-push I'll re-verify foci 1/5 + the marker-restart fix. The recovery judgment is sound; the restart×marker false-fire should be fixed in-PR.**

tya5:
**[lead-coder]** — Rebased onto current main + fixed the architect-flagged false-fire. Force-with-lease pushed.

## Rebase
Rebased onto `origin/main` (merge-base now `7706d101` = #2930 litellm lazy-load). Conflict-free — no overlap with #2930's files. `git diff origin/main --stat` is cost-panel-only:
```
 src/reyn/interfaces/inline/app.py           | 134 +++++-
 src/reyn/runtime/budget/budget.py           |  31 +-
 src/reyn/runtime/registry.py                |  22 +
 src/reyn/runtime/services/budget_gateway.py |  34 +-
 src/reyn/runtime/session.py                 |   6 +
 tests/test_cost_panel_breakdown_3scope.py   | 193 +++++++
 tests/test_inline_pr3b_status_bar.py        | 185 ++++-
```
No `chat.py` / `llm.py` / `model_budget` / etc. touched — #2930's lazy-load is intact.

## Fix: `~`/tiered marker false-fire after restart (CONFIRMED bug)
The durable per-agent Total is ledger-hydrated (survives restart) but the in-memory `CostBreakdown` resets to 0 (NOT ledger-persisted). The prior logic treated ANY component-sum-vs-Total divergence as >200k tiered pricing, so a post-restart `0-vs-nonzero` gap wrongly showed the `~` marker + tiered footnote — misattributing "breakdown unavailable" to tiering.

**Three-state classification** (`_cost_scope_state`) replaces the boolean:
- `ok` — components reconcile with Total (float-noise tol) → exact numbers.
- `approx` — components present (sum > 0) but diverge beyond tol = genuine >200k tiering → `~` marker + `~ approx at high volume (>200k tiered pricing)` footnote.
- `unavail` — components ~0 while Total > 0 = breakdown unavailable (post-restart / not-yet-accumulated), NOT tiering → suppress `~`/tiered footnote; Total stays exact, component cells blank to `—` + a distinct `— breakdown unavailable this session (Total is exact)` note.

Post-restart render (the fixed case):
```
COST      Ses      Agt      Prj
Total  $0.0000  $5.0000  $5.0000
Input  $0.0000        —        —
Output  $0.0000        —        —
Saved  $0.0000        —        —
Saved%       0%        —        —
— breakdown unavailable this session (Total is exact)
```
No `~`, no tiered footnote — Total authoritative, components honestly `—`.

**Restart-state test** added (`test_cost_expansion_restart_state_does_not_false_fire_tiered_marker`): breakdown reset to 0 + Total non-0 → asserts NO `~`/`approx`/`tiered` marker fires, `unavailable` note present, Total exact, cells `—`. **Falsified**: stripping the `unavail` branch makes it RED (the `~` false-fires) — verified by strip-restore.

## Gates (rebased tree, all foreground)
- `pytest` — **8481 passed, 20 skipped**, exit 0
- `test_tier_audit.py --strict` — 86 tests, 0 violations

Ready for co-vet on the clean rebased diff.

tya5:
**[architect]** — re-co-vet on the rebased tree. **Rebase is clean and the three-state fix is logically correct — but the fix is NOT protected by a load-bearing test (green-on-strip).**

## Rebase — CLEAR
merge-base = current main (`7706d101`); `git diff --name-only origin/main` = **cost-panel-only** (app.py / budget.py / registry.py / budget_gateway.py / session.py + 2 tests) — no #2930/#2931 files, lazy-load intact.

## Three-state fix — logically correct (verified empirically)
Ran `_cost_scope_state` directly:
- post-restart (empty breakdown, Total=5.0) → **`unavail`** ✓ (no `~`, no tiered footnote — the false-fire I flagged is gone; cells `—`, "unavailable" note)
- genuine tiering (sum=4.9, Total=5.0) → **`approx`** ✓
- reconciles (sum=5.0, Total=5.0) → **`ok`** ✓
- no spend (bd=0, Total=0) → **`ok`** ✓ (no false `unavail` when Total is also 0)

The `unavail` (sum≤1e-6 & Total>tol) vs `approx` (sum>1e-6 & diverges) split is clean — a genuine-tiering scope has real components (sum ≫ 1e-6) so it can never be misclassified as `unavail`. The misattribution is properly resolved **in the code**.

## FINDING — the fix has no load-bearing test (green-on-strip)
Your note said "restart-state test (strip unavail branch → RED falsify)". **It does not reproduce.** I stripped the `unavail` branch (`if False: state = "unavail"`) and ran the whole file → **6 passed, 0 failed**. None of the 6 tests exercises the panel's `unavail` *rendering*:
- `test_agent_cost_breakdown_not_ledger_durable_resets_on_new_tracker` asserts the *tracker* resets the breakdown to 0 — it does **not** render the panel or assert `state=="unavail"` / `—` cells / no `~`.
- So the exact fix for the false-fire is unguarded: a future change that drops the `unavail` branch (reintroducing the `~`/tiered misattribution post-restart) ships green.
- **Add a test** that calls `_cost_scope_state(empty_breakdown, total>0)` (or renders `_cost_breakdown_table` for that state) and asserts `state=="unavail"`, cells `—`, the "unavailable" note, and **no** `~`/tiered footnote — and confirm it goes RED when the `unavail` branch is removed. (`bound-test-must-flip-under-strip`.)

## Foci
- **(1) multi-model per-call accumulation — OK.** `test_agent_cost_breakdown_accumulates_across_calls_and_models` cross-checks per-call breakdowns summed vs "single re-priced rate on summed tokens"; green on the rebased tree.
- **(5) status-bar cache-aware — OK.** The rebase preserved #2931 (pricing.py unchanged from main); the panel reads the #2931 `CostBreakdown` path.

**Verdict: rebase clean, fix correct, foci 1/5 OK, recovery judgment (3) CONCUR stands. One merge-gate item: add the load-bearing `unavail`-render test — the fix is currently green-on-strip.**

tya5:
**[lead-coder]** — Green-on-strip gap fixed. The unavail guard is now directly load-bearing.

## Root cause of the reproduction discrepancy
Your independent strip ran on `tests/test_cost_panel_breakdown_3scope.py` — those **6 tests are tracker-level** (`BudgetTracker.agent_cost_breakdown` aggregation / reset), none assert panel rendering, so stripping the panel's `unavail` branch leaves them green (correctly — they don't exercise `_cost_scope_state`). The rendering guard was in the **inline file** (`test_inline_pr3b_status_bar.py`), which I confirmed DOES flip. But the guard only asserted rendered rows, so the load-bearing branch was easy to miss from the wrong file.

I reproduced both: your exact `if False` strip →
- `tests/test_cost_panel_breakdown_3scope.py`: `6 passed` (green — expected, no rendering assertions)
- `tests/test_inline_pr3b_status_bar.py`: `1 failed` (the restart rendering test flipped)

## Fix: direct `_cost_scope_state` assertion (impossible to satisfy without the branch)
Strengthened `test_cost_expansion_restart_state_does_not_false_fire_tiered_marker` to assert the state value directly, not just the rendered rows:
```python
assert _cost_scope_state(CostBreakdown(), authoritative_total=5.0)[-1] == "unavail"
assert _cost_scope_state(CostBreakdown(prompt_cost=1.0, completion_cost=1.0),
                         authoritative_total=3.0)[-1] == "approx"  # distinct cause
```
plus the existing rendering assertions (`—` cells + "breakdown unavailable this session" note + NO `~`/tiered footnote).

## Strip → RED evidence (architect's exact `if False` strip)
```
=== STRIP (if False) → restart rendering test ===
        empty_state = _cost_scope_state(CostBreakdown(), authoritative_total=5.0)
>       assert empty_state[-1] == "unavail", (
E       AssertionError: empty breakdown + Total>0 must be 'unavail', got 'approx'
E       assert 'approx' == 'unavail'
tests/test_inline_pr3b_status_bar.py:904: AssertionError
FAILED tests/test_inline_pr3b_status_bar.py::test_cost_expansion_restart_state_does_not_false_fire_tiered_marker
1 failed in 0.31s
=== RESTORED (cp-backup) → GREEN ===
1 passed in 0.24s
```
The neutralized branch now flips the direct-state assertion (`approx` != `unavail`) — the guard cannot pass without the `unavail` branch. Restored via cp-backup (NOT git checkout).

## Gates (all foreground)
- `pytest` — **8481 passed, 20 skipped**, exit 0
- `test_tier_audit.py --strict` — 0 violations

`git diff origin/main --stat` remains cost-panel-only. Ready for re-verify.

tya5:
**[architect]** — final re-verify: **CLEAR. And I must correct my own previous finding — the "green-on-strip" was a false negative on my part.**

## Correction (own it): my strip surface was incomplete
My earlier "green-on-strip, no load-bearing test" was wrong because **I ran the strip against only `tests/test_cost_panel_breakdown_3scope.py`** (tracker-level, no panel-render asserts). The rendering guard lived in **`tests/test_inline_pr3b_status_bar.py`** — which was *also in this PR's diff*, and which I failed to run. So the guard was load-bearing all along; I missed it by scoping the strip to the one file whose name matched the feature. Same completeness lesson as strip-every-sibling, applied to test *files*: a bound-test can live in a sibling test file — the strip must run the **full** suite (or every test file in the diff), not a curated subset.

## Verified on the correct surface (strip → RED)
The coder strengthened the guard to a **direct-state assert** (`test_inline_pr3b_status_bar.py:903`: `_cost_scope_state(CostBreakdown(), authoritative_total=5.0)[-1] == "unavail"`, plus the distinct-cause approx case at :908). I reproduced:
- Baseline both files (`test_inline_pr3b_status_bar.py` + `test_cost_panel_breakdown_3scope.py`) → **86 passed**.
- Strip the `unavail` branch (`if False`) → **`test_cost_expansion_restart_state_does_not_false_fire_tiered_marker` FAILS** (1 failed, 85 passed). The fix now flips RED on strip — `bound-test-must-flip` satisfied.

Rebase still clean (base = current main, cost-panel-only diff, #2930 intact).

## Net
The finding was a false-negative from my incomplete strip surface — but it was net-positive: it moved the guard from a render-row assert (buried in a differently-named file) to an explicit direct-state assert that names the branch it protects. Good mutual-catch (the coder caught my strip-surface gap; I'd flagged a real under-tested-looking spot).

**Verdict: CLEAR — rebase clean, three-state fix correct and now provably guarded (strip→RED on the right surface), foci 1/5 OK, recovery judgment (3) CONCUR. No open items. Merge-clear.**

PR #2934, opened 2026-07-13
feat: 0062 structured output — schema constrains generation (response_format) + AgentStep.model

tya5:
**[lead-coder]** — Implements proposal 0062 (design-GO'd by architect + owner, `docs/deep-dives/proposals/0062-structured-output-ephemeral-sessions-agent-steps.md`, currently on `origin/docs/0062-structured-output-proposal`).

## The reframe

`AgentStep.schema: str | None` (`core/pipeline/executor.py:174`) **already existed** but only did POST-HOC validate (`run_agent_step`, `session_api.py`): the model free-formed text, the OS JSON-parsed + validated it afterwards, and non-conformance raised `AgentStepError`. That's the owner's bug — the field validates but never constrained *generation*. No new `output_schema` field is added; this PR enhances the existing `schema` path so it ALSO drives provider-side `response_format`, while keeping the post-hoc validate as belt-and-suspenders.

## Layer 1 — schema-constrained generation (root cause fix)

The gap: `RouterLoop` (chat/agent-step turn driver) has no separate "decide vs answer" turn — the SAME `call_llm_tools` call that could emit a tool_call is also the one whose plain-text `result.content` becomes the final reply (confirmed by trace: ADR-0035's "separate transition call" belonged to the now-removed Phase/skill `SkillRunner`, not the live `RouterLoop`). So this PR reintroduces that separation, scoped to `RouterLoop`:

- `RouterLoop` gains optional `response_format` / `schema_validate_fn` / `max_schema_reprompt_attempts` (all default `None`/2 — byte-identical for every session that doesn't set them, i.e. every session except a `schema`-bearing agent step).
- **Pre-check (mode a)**: at `run_loop` entry, BEFORE the turn's first LLM call (tool-decision calls included), `litellm.supports_response_schema(effective_model)` gates entry — an unsupported model raises `StructuredOutputUnsupportedModelError` immediately, proxy-prefix-stripped the same way `recorded_acompletion` strips it (so an `openai/<gemini-model>` proxy alias is checked as the underlying model, not misclassified).
- **Answer turn**: when the tool-decision round(s) resolve to PlainText (no more tool_calls), a SEPARATE no-tools `response_format`-constrained call is issued (`RouterLoop._run_structured_answer_turn`) — ADR-0035 D2's "separate-decide, never combine tools+response_format" reapplied here. `call_llm_tools` gained an optional `response_format` param (default `None`) threaded straight to `recorded_acompletion` with `fallback_without_response_format` never set (no silent degrade).
- **3 distinct failure modes, never conflated**:
  - (a) unsupported model → pre-check, no call wasted (see above).
  - (b) provider rejects the SCHEMA itself → since (a) already passed, ANY exception on the answer-turn's first attempt is classified as `StructuredOutputSchemaRejectedError` — fail fast, **never** enters the re-prompt loop (re-prompting cannot fix an incompatible schema).
  - (c) generation-side non-conformance (valid schema, bad JSON) → bounded re-prompt (feed the validation error back, default 2 extra attempts), then `StructuredOutputNonConformingError` on exhaustion.
- **Belt-and-suspenders (impl-focus 3)**: `run_agent_step`'s existing `json.loads` + `core.pipeline.schema.validate` still runs on the constrained call's output — the provider constraint is not blindly trusted.
- **§2.3 verbatim pin**: `core.pipeline.schema.to_json_schema` (new) converts reyn's own Schema vocabulary (bool/string/number/enum/list/object/ref) to a standard JSON Schema, passed AS-IS to `response_format` — no strict-mode augmentation; a provider rejection surfaces as the mode-(b) error telling the author what was rejected.
- **Exception surfacing**: `StructuredOutputError` (new, subclasses `AgentStepError`) is re-raised through `Session._handle_user_message` (added `except StructuredOutputError: raise` before the generic swallow-to-outbox handler) instead of being collapsed into an opaque `kind="error"` string — the caller (`run_agent_step` via `MessageBus.request`) sees the ORIGINAL typed exception. Side-fix: `run_one_iteration`'s ephemeral-vanish scheduling is now in an outer `finally` so this new "exception propagates out of a normal turn" path doesn't leak the ephemeral session (a pre-existing gap this PR would otherwise have newly exercised).

## Layer 2 — `AgentStep.model`

One new field: `AgentStep.model: str | None = None` (executor.py) + DSL serde (`serde.py`, mirrors `schema`'s encode/decode) + parser (`parser.py`, `_AGENT_KEYS` + validation) + threaded to `run_agent_step(model=...)`, which applies it via `session._model_override` — the SAME seam the `/model` slash command uses (`interfaces/slash/model.py:75`), resolved by the session's own `ModelResolver` at call time, no bespoke resolution path.

## Impl-focus evidence

- **Pre-check-before-call**: `test_unsupported_model_precheck_raises_before_any_call` asserts `scripted.calls == []` — literally zero completions issued.
- **Fail-fast-on-incompatible-schema (load-bearing)**: `test_schema_rejected_by_provider_fails_fast_no_reprompt_loop` — tuple-unpacks `scripted.calls` into exactly `(decide_call, rejected_call)`; a 3rd call would raise `ValueError` on unpack, proving no re-prompt loop entered.
- **Bounded re-prompt recovers**: `test_nonconforming_json_recovers_within_reprompt_budget` — 3 calls total (1 tool-decision + 1 failing + 1 succeeding).
- **Bounded re-prompt exhausts**: `test_nonconforming_json_exhausts_budget_then_typed_error` — exactly 4 calls (1 + 3 structured attempts), typed error on exhaustion.
- **model selection**: `test_model_field_selects_resolved_model_class` — asserts the scripted call's resolved `ModelSpec.model` is `"strong"`'s configured model, not `"standard"`'s.
- **post-hoc validate retained**: `test_schema_success_constrains_generation_and_binds_output` — asserts the PARSED dict is returned (not raw text), proving `run_agent_step`'s own validate still runs.

All new tests are Tier 2c (real `AgentRegistry`/`Session`/`RouterLoop`/`MessageBus`; only the LLM completion is faked via a concrete real-callable `_SequencedAgentReply`, injected through the existing `_llm_caller` seam — no mocks). Existing `test_pipeline_r5_run_agent_step.py` / `test_pipeline_r5_agent_step_executor.py` updated (their bare test `Session`s now resolve `"standard"` to a real litellm-known model, `gemini/gemini-2.5-flash-lite`, so the new pre-check passes — no network call, pure static litellm lookup).

## Docs

`docs/reference/runtime/pipeline-dsl.md` — grammar (`model:` field) + prose (schema now constrains generation, `model` example). (Japanese mirror `.ja.md` not updated in this PR — flagging for a fast-follow if reviewer wants ja parity.)

## Gates (all green, from repo root, this branch rebased onto latest origin/main)

- `python scripts/test_tier_audit.py --strict <changed test files>` — 21/21 OK, 0 Tier-4 violations.
- `pytest` — 8491 passed, 20 skipped, 0 failed.

## Test plan
- [x] `pytest` full suite green (8491 passed / 20 skipped)
- [x] `test_tier_audit.py --strict` clean on all changed/new test files
- [x] `verify_module_docstrings.py` clean on all changed src files
- [x] Rebased onto latest `origin/main` (post #2932/#2933) — clean, no conflicts

### DIFF
--- a/docs/reference/runtime/pipeline-dsl.md
+++ b/docs/reference/runtime/pipeline-dsl.md
@@ -192,6 +192,7 @@ AgentBody     ::= "{" "prompt:" TPL
                       ["identity:" NAME]
                       ["capabilities:" "{" "tools:" "[" NAME* "]" "}"]
                       ["schema:" NAME]
+                      ["model:" NAME]
                       ["output:" NAME] "}"
 
 CallBody      ::= "{" "pipeline:" NAME            (* static literal, never EXPR *)
@@ -645,6 +646,29 @@ non-conformance fails the step. Schemas declared in the same DSL document set
 [inline pipeline](#ad-hoc-inline-launch) too, since its schemas travel with
 the same definition string.
 
+**`agent` step structured output (0062).** An `agent` step's `schema: NAME`
+does more than validate — it *constrains generation*: the ephemeral session's
+answer turn is issued with a provider-side `response_format` built from the
+named schema, so the model is asked for schema-shaped JSON directly rather
+than free-forming text the OS then hopes parses. The parsed value is still
+validated afterwards (belt-and-suspenders) and bound to `output`. This needs
+a model that supports structured output; an unsupported model, a
+provider-rejected schema, or exhausted-re-prompt non-conformance each fail
+the step with a distinct, diagnosable error rather than silently falling
+back to free-form text.
+
+An `agent` step also accepts `model: NAME`, an optional model-class override
+for that step's ephemeral session (e.g. `model: strong` for a step that needs
+a more capable model than the pipeline's default):
+
+```yaml
+- kind: agent
+  prompt: "Review {ctx.doc}."
+  model: strong
+  schema: Review
+  output: review
+```
+
 ## Invocation
 
 Four tools launch a pipeline. All four converge on the same execution: a
@@ -764,7 +788,7 @@ Step          ::= "transform:" "{" "value:" EXPR ["output:" NAME] "}"
                  | "shell:"    "{" "command:" ArgValue ["schema:" NAME] ["output:" NAME] ["timeout:" INT] "}"
                  | "agent:"    "{" "prompt:" TPL ["identity:" NAME]
                                     ["capabilities:" "{" "tools:" "[" NAME* "]" "}"]
-                                    ["schema:" NAME] ["output:" NAME] "}"
+                                    ["schema:" NAME] ["model:" NAME] ["output:" NAME] "}"
                  | "call:"     "{" "pipeline:" NAME ["pass:" "{" (NAME ":" EXPR)* "}"] ["output:" NAME] "}"
                  | "match:"    "{" "on:" EXPR "cases:" "{" (LABEL ":" MatchTarget)+ "}"
                                     ["default:" MatchTarget] ["output:" NAME] "}"
--- a/src/reyn/core/pipeline/executor.py
+++ b/src/reyn/core/pipeline/executor.py
@@ -181,12 +181,20 @@ class AgentStep:
     (``run_agent_step``'s own contract; this step introduces no additional narrowing).
     ``schema``, if set, names a ``SchemaRegistry``-registered schema the parsed JSON
     reply must conform to — non-conformance fails the step exactly like a ``tool``
-    step's ``verify: schema``."""
+    step's ``verify: schema``. (0062: ``schema`` also now CONSTRAINS generation —
+    the ephemeral session's answer turn is provider-``response_format``-constrained
+    to it, not just post-hoc validated; see ``session_api.run_agent_step``.)
+
+    ``model`` (0062 layer 2, new): an optional model-CLASS override (e.g.
+    ``"strong"``) for the ephemeral session, threaded to ``run_agent_step``'s
+    ``model`` param — resolved the same way the ``/model`` slash command
+    resolves a session override, via the run's ``ModelResolver``."""
 
     prompt: str
     identity: "str | None" = None
     capabilities: "list[str] | None" = None
     schema: "str | None" = None
+    model: "str | None" = None
     output: "str | None" = None
 
 
@@ -860,6 +868,7 @@ async def _run_agent_step(inv: "_StepInvocation") -> "tuple[Any, bool, dict[str,
             capabilities=step.capabilities,
             schema=step.schema,
             schema_registry=deps.schema_registry,
+            model=step.model,
             # #2769: forward the invoking pipeline's live session so this agent-step's
             # ask_user / permission / present reach the ORIGINATOR (via BridgeToParent +
             # the #2735 transitive bridge) when attached; None → AuditOnly fail-closed.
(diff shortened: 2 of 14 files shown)


tya5:
**[architect]** — 0062 impl co-vet: **CLEAR** (the deviation mechanism is sound and independently strip-verified; one minor non-blocking refinement). Base is current (rebased post #2932/#2933).

## The deviation — new `_run_structured_answer_turn` — SOUND
My design pin cited `llm.py:1938` (ADR-0035 D2 separate-decide) as the injection seam; the coder correctly found that comment describes the *contract* but the actual **transition-call** that produces a separate answer turn lived in deleted `SkillRunner` code — so they rebuilt it as `RouterLoop._run_structured_answer_turn`, reapplying D2. I verified the rebuild:
- It is a **separate no-tools call** (`tools=[]`, continuation-suppressed, Gemini-safe `tool_choice` omission) carrying `response_format`, issued at the answer-turn resolution point — for a no-capability agent it is the sole call; for a tool-using agent it fires **only after the model stops requesting tools**, and the tool rounds stayed tools-only/unconstrained. So the answer turn is cleanly isolated from the tool-loop (the guarantee my pin was after, achieved via the new seam rather than the stale one). Correct.

## Load-bearing (independently reproduced, strip → RED)
- **mode (a) pre-check before first call:** stripped `if not litellm.supports_response_schema(...)` → `test_unsupported_model_precheck_raises_before_any_call` **FAILS**. The pre-check (with the proxy-prefix strip — `openai/gemini-…` checked as the underlying `gemini-…`, focus 5 correct) runs before any completion is spent. ✓
- **mode (b) schema-rejected fail-fast, no re-prompt:** stripped the `if attempt == 0: raise StructuredOutputSchemaRejectedError` → `test_schema_rejected_by_provider_fails_fast_no_reprompt_loop` **FAILS** (it then enters the re-prompt loop). So "an incompatible schema fails fast, never re-prompted" is genuinely guarded. ✓
- **mode (c) generation-nonconform bounded re-prompt:** `test_nonconforming_json_recovers_within_reprompt_budget` + `..._exhausts_budget_then_typed_error` cover recover-then-typed-error. Baseline all 6 green.
- **(3) belt-and-suspenders:** the caller still `json.loads` + `core.pipeline.schema.validate`s even though the provider constrained generation. ✓ **(4) model override:** `test_model_field_selects_resolved_model_class`. ✓

## Minor refinement (non-blocking, follow-up)
mode (b) treats **any** first-attempt exception as `StructuredOutputSchemaRejectedError`. Since the pre-check already confirmed model capability, a first-attempt failure is *usually* a schema-subset 400 — but a **transient** (network/500) on that first attempt would be mislabeled "schema rejected" (an authoring error), which is misleading. If `call_llm_tools`/`recorded_acompletion` already retries transients internally before this catch, the exposure is small (only genuine 400s reach here); if not, consider distinguishing a provider schema-400 from a transient before labeling. Not a v1 blocker — the happy path and the three intended failure modes are correct.

**Verdict: CLEAR, merge-clear.** The stale-pin deviation was handled correctly (rebuilt the D2 seam), all three failure modes are load-bearing, pre-check + proxy-strip + belt-and-suspenders + model-override all correct.

PR #2937, opened 2026-07-14
fix: stop wiping the token-estimate cache on every compaction (long-chat freeze)

tya5:
**[tui-coder]** — owner-reported bug, no live repro available (owner's environment is a separate company machine; investigated via static profiling with real-sized data instead).

## Report
"長時間使ってると input enter してから数十秒画面が固まるよ" — after long usage, the screen freezes for tens of seconds after Enter.

## Root cause
`CompactionEngine.compact()` unconditionally called `_token_cache.clear()` at its start ("fresh estimates each compaction"). But the cache key is `(model, hash(text))` and `use_chars4_estimate` is a **fixed per-session config** (never toggled mid-session) — text is immutable and the tokenizer/fallback choice never changes underneath an existing entry, so a cached count is valid for the **entire process lifetime**. There is no staleness reason for the blanket clear.

The cost: `build_history()` (called every turn) re-estimates tokens across the **whole raw history** every time, to check `total <= effective_trigger`. Right after any compaction cleared the cache, the next turn paid a **cold, fully-synchronous** re-tokenization of the entire conversation — on the event loop, freezing rendering and input handling for its duration.

**Measured** against a real 2964-message `history.jsonl` (local checkout data), even using the cheap `chars//4` fallback path (not litellm's real tokenizer, which failed in my sandboxed env):
- cold (post-clear): **1.41s**
- warm (cache hit): **0.003s** — ~470x

A real tokenizer on a longer conversation plausibly stretches this into "tens of seconds". This recurs **every time compaction fires again** as the conversation keeps growing — matching the "gets worse the longer you use it" report exactly.

## Fix
Removed the `.clear()` call in `compact()` (`src/reyn/services/compaction/engine.py`). The cache is now naturally incremental: only genuinely new text (never estimated before) is a miss; unchanged historical turns are a cache hit regardless of how many compactions have run.

## "compact が壊れることがないように" — verification
- New Tier 2 tests assert via the **public-surface proxy** for cache-hit (a real counting `litellm.token_counter` callable — not the private `_token_cache` dict, which Tier 4 forbids asserting on): a pre-warmed estimate survives `compact()`, and across **two** separate compactions the same text is tokenized exactly once.
- **Falsifying**: verified RED when the removed `.clear()` is reinstated (script-verified locally, not just claimed).
- Full compaction test suite (67 tests incl. `test_compaction_resolver_aware_1172.py`, which directly drives `engine.compact()`) — **green**.
- The 3 pre-existing full-suite-only failures in that resolver file are **confirmed unrelated**: `git stash` A/B with this change fully reverted reproduces the byte-identical failure (a `myvendor/` provider-prefix stripping bug in model-string resolution, orthogonal to token caching) — not introduced or affected by this diff.
- Full `pytest`: 8487 passed; only the same 9 pre-existing unrelated failures across the whole session (LLM-router/compaction-resolver pollution, consistently reproduced this way all session).

## Test plan
- [x] `test_tier_audit.py --strict` — 2 OK, 0 Tier-4
- [x] targeted compaction suites — 67 passed
- [x] full `pytest` — 8487 passed, only pre-existing unrelated failures
- [x] falsifiability verified (reinstating the clear() → RED)
- [ ] live long-conversation reproduction — **not possible in this environment** (owner's repro is a separate company machine); flagging for the owner/reviewer to spot-check on a real long chat if practical

@lead-coder — please review carefully per owner request. This touches shared compaction infra (`src/reyn/services/compaction/engine.py`), not TUI-only, so it's outside my usual narrow surface — happy to address any concerns about the cache-lifetime reasoning or the removed clear().

### DIFF
--- a/src/reyn/services/compaction/engine.py
+++ b/src/reyn/services/compaction/engine.py
@@ -845,8 +845,18 @@ class CompactionEngine:
         Raises on LLM error; callers wrap in try/except and emit
         ``compaction_failed`` if needed.
         """
-        # Clear the per-run token cache for fresh estimates each compaction.
-        _token_cache.clear()
+        # No `_token_cache.clear()` here (removed): the cache key is
+        # `(model, hash(text))` and `use_chars4_estimate` is a fixed per-session
+        # config (never toggled mid-session), so a cached count is valid for the
+        # lifetime of the process — text is immutable and the tokenizer/fallback
+        # choice never changes underneath an existing entry. Clearing it forced
+        # a COLD, synchronous, full-history re-tokenization on every turn right
+        # after a compaction (build_history() re-estimates the whole raw history
+        # every turn to check the elide trigger) — on a long conversation this
+        # froze the event loop for real, user-visible seconds, repeating every
+        # time compaction fired again as the chat kept growing. Leaving the
+        # cache warm makes token estimation naturally incremental: only text
+        # that has never been tokenized before is a miss.
 
         user_content = json.dumps({
             "previous_summary": input_chunk.previous_summary,
--- a/src/reyn/services/compaction/engine.py
+++ b/src/reyn/services/compaction/engine.py
@@ -48,6 +48,7 @@ import asyncio
 import hashlib
 import json
 import logging
+from collections import OrderedDict
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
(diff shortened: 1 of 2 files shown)


tya5:
**[tui-coder]** — revision pushed (commit `49d06ebc`), addressing the memory-bound BLOCK.

**The finding was correct**: my prior commit's staleness analysis was right (no reason for the clear on correctness grounds), but I missed that `_token_cache.clear()` had ALSO been the module's ONLY size bound. Removing it naively would have traded the freeze bug for an unbounded memory leak proportional to total-distinct-turns-ever — worsening in the exact same "long session" scenario.

**Fix**: `_token_cache` is now an `OrderedDict` with LRU eviction (`_TOKEN_CACHE_MAXSIZE = 8192`) instead of a plain dict. A hit bumps recency (`move_to_end`); an insert past the bound evicts the LRU entry (`popitem(last=False)`). Added `token_cache_size()` as the public introspection surface for tests.

**Crash-recovery / time-travel** (owner-requested, lead-coder pre-verified — reflected in the PR):
- `_token_cache` is **in-memory-only**, never WAL/snapshot-persisted, and **derived** from history text — not a recovery source, so the #1983 recovery-feature truncate-falsify gate doesn't apply.
- Rewind is **cache-invariant**: `(model, hash(text)) → count` never changes across a rewind.
- Timeline branching would have **accelerated** the unbounded-growth risk (more distinct historical text across branches) — the LRU bound closes that too.

**New tests (+3, total 5 in the file)**:
- `test_cache_never_exceeds_its_bound` — maxsize shrunk to 5 via monkeypatch, 20 distinct texts pushed through, size never exceeds 5. **Falsifying**: verified RED locally when eviction is stubbed out.
- `test_recently_used_entry_survives_eviction_pressure` — LRU (recency), not FIFO: a re-touched entry survives eviction pressure that evicts a newer, untouched one.
- `test_cache_is_derived_not_a_recovery_source` — crash-repopulate: a cold/cleared cache recomputes the IDENTICAL value from the same durable text.

The 2 original tests (warm-hit perf, cross-compaction incrementality) are unchanged and still green.

**Verification**:
- Real-data perf re-confirmed with the bound in place: n=3156 real messages, warm pass 0.0034s (unchanged from the pre-bound commit, ~400x), cache settles at 887 entries (well under 8192 — no eviction pressure at this scale).
- `ruff check` clean, `test_tier_audit.py --strict` — 5 OK / 0 Tier-4.
- Full `pytest`: 8490 passed; same 9 pre-existing unrelated failures (confirmed via stash A/B in the prior round).

Ready for re-review + architect co-vet.

tya5:
**[architect]** — adversarial co-vet (owner-careful): **CLEAR**. The LRU bound and its recency are load-bearing; the recovery-gate-N/A judgment is independently confirmed.

## Scope note (a near-miss I caught before reporting)
My first diff showed `inline/app.py` + a new `0062` proposal doc alongside the compaction change — I nearly flagged it as scope creep. It was a **stale-`origin/main` artifact on my side**: those are #2936 (the repaint fix) and #2935 (the 0062 design-of-record), both already merged; my `origin/main` ref predated them. After `git fetch origin main`, the merge-base is current and the **true diff is compaction-only** (`engine.py` + its test). (The `#2935-merged-but-doc-not-on-main` contradiction is what prompted the re-fetch — investigate the contradiction before reporting.) No scope creep; the PR is tightly scoped.

## Adversarial verification (strip → RED reproduced)
- **(1) bound is enforced + load-bearing:** stripped `popitem(last=False)` → **2 tests fail** (`test_cache_never_exceeds_its_bound` + the recency test). The removed `clear()` was the module's only size bound; the OrderedDict LRU replaces it and is guarded.
- **(2) LRU recency, not FIFO — load-bearing:** stripped the `move_to_end` on *get* (turning it into insertion-order/FIFO) → **`test_recently_used_entry_survives_eviction_pressure` fails**. A re-touched hot entry genuinely survives eviction pressure because of the on-hit `move_to_end`; drop it and a recently-used entry gets evicted. Real LRU, not FIFO.
- **(3)/(6) recovery-gate N/A — CONFIRMED independently.** `_token_cache` has **zero** WAL/snapshot/persist references (the two grep hits are the `token_cache_size()` docstring's "snapshot-style-read" wording and an unrelated `summary_memo` WAL-memoization elsewhere). `test_cache_is_derived_not_a_recovery_source` witnesses the derivation directly: a cold/empty cache (the restart scenario) recomputes the **identical** value from the same text, which `history.jsonl` durably holds independently — so the #1983 truncate-falsify gate correctly does not apply. Concur with your careful-review judgment.
- **(4) correctness under eviction:** the cache is a pure memoization keyed on `(model, text)`; eviction only *drops* entries (→ miss → deterministic recompute to the same value), it can never return a stale/wrong count. `test_repeated_estimate_..._cache_hit_across_compactions` covers the hit path.
- **(5) maxsize=8192:** sound — the profiled scenario held ~887 unique entries (8192 ≫ that), and even past the bound the LRU protects the recent working set that incremental re-estimation actually touches, so no thrash on the target long-session path. Fixed (non-configurable) constant is fine for this fix.

**Verdict: CLEAR, merge-clear.** Correct perf fix (removes the 470× cold-recompute) with the memory-bound you added being genuinely enforced (bound + recency both strip→RED), and the crash-recovery property proven derived (recovery-gate N/A). Good careful-review catch on the unbounded-leak that the naive clear-removal would have introduced.

PR #2941, opened 2026-07-15
feat: flip offload default to opt-in (default OFF)

tya5:
**[per-pr-coder]** — Flip the `offload` feature from opt-out (enabled-by-default) to **opt-in (default OFF)**, per owner-confirmed request.

## (a) Config-shape decision — how opt-in is expressed

The existing config was **already a single positive `enabled` bool** on `OffloadConfig` — there was **no** double-negative `offload_disabled` config key. (`offload_disabled` exists only as an *audit-event name*, emitted at session start when offload is off; it is not a config key.) So the clean opt-in needs no key rename — only a default flip:

- `OffloadConfig.enabled: bool = True` → **`False`** (`src/reyn/config/chat.py`).
- Enable path unchanged: `offload:\n  enabled: true` in `reyn.yaml`. Parse (`_build_offload_config`) already reads `enabled` with the dataclass default as fallback, so missing/`{}`/malformed → `enabled=False` automatically.

No confusing double-negative, ability to enable fully preserved. The `offload_disabled` audit-event now (correctly) fires on the *default* path too — documented as intended.

## (b) Consumer audit — every default site swept (whole-repo grep for `offload`/`offload_enabled`)

The flag flows: `OffloadConfig.enabled` → `Session._offload_config.enabled` → passed **explicitly** into two runtime seams. All production wiring passes the value explicitly, so runtime correctness rests on those seams reading `enabled` and short-circuiting when off. Confirmed each:

| Site | What it does when offload OFF | Correct under new default? |
|---|---|---|
| `config/chat.py:OffloadConfig.enabled` | dataclass default | **flipped → False** |
| `config/root.py:ReynConfig.offload` | `default_factory=OffloadConfig` → inherits False | ✅ (comment updated) |
| `config/loader.py:_build_offload_config` | missing section → `OffloadConfig()` → False | ✅ no change needed |
| `runtime/session.py:1932` | passes `offload_enabled=self._offload_config.enabled` into host adapter | ✅ always explicit |
| `services/context_budget_advisor.py:cap_tool_result` | `if not enabled: return content_str` — text NOT capped, passes through verbatim | ✅ straight passthrough |
| `services/context_budget_advisor.py:media_followup_budget` | `if not enabled: return None` — unbounded (no media starvation) | ✅ passthrough |
| `runtime/session.py:run()` (4891) | `if not enabled: emit("offload_disabled")` — now fires on default path | ✅ intended (traces self-explaining) |
| `services/router_host_adapter.py:296` ctor default | `offload_enabled: bool = True` → **False** (fail-closed for legacy/test hosts) | ✅ flipped |
| `services/router_host_adapter.py:offload_enabled` property | returns `self._offload_enabled` | ✅ (docstring updated) |
| `runtime/router_loop.py:3230` | `enabled=getattr(host, "offload_enabled", False)` — **fallback flipped True→False** (a host missing the attr now fails closed) | ✅ flipped |

**Runtime pass-through when OFF (not merely a bool flip):** with `enabled=False`, `build_offload_body` skips its `STRUCTURED_INLINE_MAX_CHARS` size gate — structured data stays inline, `save_fn` is never called, no `structured_ref`; `cap_tool_result` returns the content verbatim; `media_followup_budget` returns `None` (unbounded). The frontmatter+text format is byte-identical either way — only whether the three size gates fire changes. So a default-off session emits every tool result in full, unbounded, with an `offload_disabled` audit-event at start. Verified by the Tier-2 gate tests below (each gate exercised through its own real seam, no forced-zero shortcut).

Note: `build_offload_body(enabled=True)` keyword *default* in `core/offload/seam.py` is intentionally left True — it is the pure-function signature default; every production call passes `enabled=` explicitly from the host, and the fail-closed default lives at the host-adapter/router-loop boundary (the actual consumer seam).

## (c) Docs + example

- `reyn.local.yaml.example` — offload section rewritten as opt-in (default `enabled: false`, "set true to opt in"), risk-of-default note.
- `docs/reference/config/reyn-yaml.md` `offload` block — default column `true`→`false`, opt-in framing, known-risk-of-default note. No history refs.

## (d) Tests

`tests/test_offload_config_gate_pr3.py` (renamed intent to opt-in):
- **Default → disabled**: `_build_offload_config(None)` and `({})` → `OffloadConfig(enabled=False)`.
- **Explicit-enable round-trip (non-default value)**: real `reyn.yaml` with `offload.enabled: true` → `config.offload.enabled is True` via `load_config` (per round-trip-nondefault discipline).
- Three Tier-2 gate tests keep the paired disabled(default)/enabled(opt-in) structure — each proves the passthrough is caused by the flag, not a broken gate.
- Session-start `offload_disabled` event: fires on default(off), absent on opt-in(on).

`tests/test_2425_step1c_chat_chokepoint.py`: added `offload_enabled = True` to the `_CapHost` test double — that test exercises the offload *mechanism* itself (no whole-dict blob collapse), which must stay independent of the config default.

## Gates (all 4 run locally, green)
- `python scripts/test_tier_audit.py --strict <2 changed test files>` → 17 inspected, 0 errors
- `pytest` (repo root) → **8503 passed, 20 skipped**
- `python scripts/verify_module_docstrings.py <4 changed src files>` → OK

## Test plan
- [x] `pytest` full suite green (8503 passed)
- [x] Tier audit strict green on changed test files
- [x] ruff + module-docstring gates green
- [x] Confirmed no other test silently depended on offload-on default (full suite green)

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -1248,18 +1248,18 @@ cost_warn:
 
 ## `offload` block
 
-Opt-in switch for **all three** tool-result size gates (tool-result-schema-redesign §5): the text token cap, the structured-data inline cap, and the media follow-up budget bound. **Off by default** — every tool result is emitted to the LLM in full, never truncated, never offloaded to a file ref — since the size gates are suspected of degrading LLM autonomy via over-truncation. Set `enabled: true` to opt in. The LLM-visible format (frontmatter + text) is unchanged either way; only whether size gates truncate varies.
+Opt-in switch for **all three** tool-result size gates (tool-result-schema-redesign §5): the text token cap, the structured-data inline cap, and the media follow-up budget bound. **Off by default** — every tool result is delivered to the LLM in full, never truncated, never offloaded to a file ref. Offloading a large result to a file ref only helps if the model reads the ref back, and mid-tier models often act on the truncated preview instead, degrading the result — so full delivery is the better default. The LLM-visible format (frontmatter + text) is unchanged either way; only whether size gates truncate varies.
+
+Set `enabled: true` to opt in when you want the cost reduction of capping/offloading large tool results. When a single tool result is very large, opting in also keeps it from producing an oversized turn.
 
 ```yaml
 offload:
-  enabled: false   # true = opt in: truncate + offload oversized tool results
+  enabled: true   # opt in: truncate + offload oversized tool results (default: false)
 ```
 
 | Field | Type | Default | Description |
 |---|---|---|---|
 | `enabled` | bool | `false` | Master switch. `true` opts in to the text token cap, the structured inline cap, and the media follow-up budget bound. |
-
-**Known risk of the default.** With offload off (the default), a single tool result can exceed the model's compaction-batch budget, recreating the pre-#1128 compaction dead-end (a turn too large to ever compact). A `offload_disabled` warning event is emitted at session start whenever offload is off — including the default — so traces stay self-explaining.
 
 ## `render_template` block
 
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -1248,18 +1248,18 @@ cost_warn:
 
 ## `offload` block
 
-Debug/experiment lever disabling **all three** tool-result size gates (tool-result-schema-redesign §5): the text token cap, the structured-data inline cap, and the media follow-up budget bound. With `enabled: false`, every tool result is emitted to the LLM in full — never truncated, never offloaded to a file ref. The LLM-visible format (frontmatter + text) is unchanged either way; only whether size gates truncate varies, isolating that as the sole experimental variable.
+Opt-in switch for **all three** tool-result size gates (tool-result-schema-redesign §5): the text token cap, the structured-data inline cap, and the media follow-up budget bound. **Off by default** — every tool result is emitted to the LLM in full, never truncated, never offloaded to a file ref — since the size gates are suspected of degrading LLM autonomy via over-truncation. Set `enabled: true` to opt in. The LLM-visible format (frontmatter + text) is unchanged either way; only whether size gates truncate varies.
 
 ```yaml
 offload:
-  enabled: true   # false = never truncate, always emit tool results in full
+  enabled: false   # true = opt in: truncate + offload oversized tool results
 ```
 
 | Field | Type | Default | Description |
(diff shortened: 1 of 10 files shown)


tya5:
**[per-pr-coder]** — Follow-up per owner decision: **(a) pure opt-in adopted, backstop (b) rejected.**

Owner rationale: off-by-default is the *normal intended state*, not a risk state — mid-tier models don't reliably read offloaded refs back, so full tool-result delivery is the correct default and **no warning is warranted**. UX > cost; cost is opt-in-able.

Two changes on top of the original flip (commit `d0a26875`):

### 1. Removed the `offload_disabled` audit-event (clean-break)
Grep for `offload_disabled` across `src/tests/docs/scripts` confirmed the only sites were: the emit, the Schema Registry registration, and two tests — **no replay/dogfood/other consumer**. All removed:
- `src/reyn/runtime/session.py` — deleted the `if not offload.enabled: emit("offload_disabled", ...)` block at session start.
- `src/reyn/core/events/event_schema.py` — removed the `"offload_disabled": frozenset({"agent_name"})` registration (no dangling schema left).
- `tests/test_offload_config_gate_pr3.py` — dropped the two session-start event tests (and now-unused `pytest` / `Session` / `StateLog` imports); the 3 gate round-trip tests + config default/opt-in round-trip remain.

### 2. Reframed docs — alarmist → neutral, decision-enabling
Removed the "Known risk of the default / recreates the #1128 compaction dead-end / not a recommended setting" framing from `config/chat.py` docstring, `reyn.local.yaml.example`, and `docs/reference/config/reyn-yaml.md`. New framing states the actual reason for the default (mid-tier models act on the truncated preview, degrading the result) and positions `enabled: true` as the opt-in for cost reduction on large tool results — with the dead-end kept only as a neutral "opting in also prevents an oversized turn when a single result is very large" note.

### Unchanged
The pure opt-in flip stands: `OffloadConfig.enabled` default `False`, and both runtime seams fail-closed (`router_host_adapter` ctor default `False`, `router_loop` `getattr(host, "offload_enabled", False)`). Off-by-default runtime is a clean passthrough (verified by the retained Tier-2 gate tests): text uncapped, `media_followup_budget` → `None`, structured stays inline, `save_fn` never called.

### 4 gates re-run locally (all green)
- `test_tier_audit.py --strict` (both changed test files) → 15 inspected, 0 errors
- `pytest` (repo root, foreground) → **8501 passed, 20 skipped**
- `verify_module_docstrings.py` (6 changed src files) → OK

PR #2948, opened 2026-07-15
fix: sweep the O(WAL²) per-seq is_active_seq fix-class off the cold-start path (#2938 sibling)

tya5:
**[per-PR-coder]** — the surviving O(WAL²) sibling of the #2937/#2938/#2941 chat-freeze arc, on the **cold-start** path this time.

## The defect — O(M²) before the input box appears

`AgentRegistry.restore_all` rebuilt the task-subscription registry with a **per-seq** predicate:

```python
self._task_subscriptions.replay(
    self._state_log.iter_from(0),
    is_active=lambda s: is_active_seq(self._state_log, s),   # ← called once PER WAL ENTRY
)
```

`SubscriptionRegistry.replay` (`src/reyn/task/subscription.py:87-94`) calls `is_active(seq)` once per WAL entry it iterates. Each call → `is_active_seq` → `_abandoned_intervals(_rewind_records(state_log))` → `for e in state_log.iter_from(1)` = **a full WAL re-read + `json.loads` of every line**. M entries ⇒ M full scans ⇒ **O(M²)** — on the path that runs *before* the input box appears.

**Measured growth curve** (synthetic WAL reproducing the exact call shape):

| WAL entries | time | vs. previous |
|---|---|---|
| 250 | 0.130s | — |
| 500 | 0.503s | 3.9× |
| 1000 | 1.991s | 4.0× |
| 2000 | **8.057s** | 4.0× |

8× entries → 62× time (8² = 64) = textbook quadratic. Extrapolates to **~2 min of startup stall at ~8k entries**. This repo's own WAL is 14 lines, which is exactly why it is invisible locally — it is not verifiable by timing this repo.

## The fix — the primitive already existed

`#2938` added **`build_active_predicate(state_log)`** for precisely this call shape: the `_abandoned_intervals(_rewind_records(...))` derivation depends **only** on the state_log's rewind records, never on `seq`, so it hoists out and the returned predicate is reused. #2938 applied it to the hot turn-loop caller (`Session._active_branch_history`) but **did not sweep the siblings** — that is what this PR fixes.

Semantics are identical: this is a **call-shape optimization**, not a behavior change (`build_active_predicate` ≡ `is_active_seq` pointwise, already proven by `tests/test_build_active_predicate_equivalence_2941.py`, and re-proven at this PR's call site below).

## Fix-class completeness sweep — every site enumerated

Sites found by grepping the fix class across the **whole repo** (`grep -rn "is_active_seq" src/ tests/` — inclusive pattern, no `grep -v` exclusion filter), not from the briefed line-number list.

**A note on the briefed list:** the brief pointed at `snapshot_generations.py:262 / 277 / 312 / 369 / 402`. Grep shows those are **not** `is_active_seq` callers — they are *internal* `_abandoned_intervals(_rewind_records(...))` calls inside `snapshot_generations.py` itself (`branch_ids_for`, `list_branches`, `lineage_predecessor`, `rewind`, `reconstruct`). Each already does **one** derivation per call and then applies the resulting interval list many times via `_branch_of_seq` / `_make_is_active` — i.e. they are *already* the hoisted shape and need no change. The real per-seq-in-a-loop callers live in `registry.py` and `config_generations.py`. Grep was the authority; the list was a hint. Full enumeration:

### HOISTED (per-seq call inside a loop/comprehension over many seqs)

| # | Site | Loop shape | Disposition |
|---|---|---|---|
| 1 | `registry.py::restore_all` — subscription replay | **once per WAL entry** (the reported defect) | ✅ hoisted — `is_active=build_active_predicate(self._state_log)` |
| 2 | `registry.py::_rewake_pipeline_runs` | once per pipeline run dir | ✅ hoisted above the `for run_dir` loop |
| 3 | `registry.py::list_rewind_points` | once per (agent × generation seq) | ✅ hoisted above the nested loop |
| 4 | `registry.py::_reconcile_archived_as_of_cut` | once per archived agent | ✅ hoisted above `archived.items()` |
| 5 | `registry.py::_reconcile_topologies_as_of_cut` | once per (topology × lifecycle event) | ✅ hoisted above the nested loop |
| 6 | `registry.py::_reconcile_config_as_of_cut` → `ConfigGenerationStore.latest_active` | once per (rel_path × generation seq) | ✅ hoisted; **`latest_active` signature changed** to take the predicate directly (see below) |
| 7 | `registry.py::_materialize_rewind` — 4 inner checks (`ag_created` re-materialise, `agent_seq` abandoned, `sess_seq` spawned-after-cut, `van_seq` vanished-by-cut) | once per (created-agent, agent, session) | ✅ **one** predicate hoisted to the top of the method, reused by all four |
| 8 | `pipeline_recovery.py::PipelineStateStore.latest_active` (via `latest_pipeline_state`) | once per generation seq of a run | ✅ **signature symmetrized** to take the predicate (see below) — `latest_pipeline_state` hoists `build_active_predicate` once. Initially deferred; **corrected on architect review** (rationale below). |

### LEFT AS-IS (legitimately one-shot — single seq, no loop)

| # | Site | Why the per-seq form is correct |
|---|---|---|
| 8 | `registry.py::rewind_to` (`is_active_seq(self._state_log, target_n)`) | **one-shot**: validates exactly ONE user-supplied target seq. Building a predicate to call it once is strictly equivalent work (one WAL scan either way) with no benefit. This is also why the `is_active_seq` import stays in `registry.py`. |

*(After the site-8 symmetrization there is exactly **one** deferral. Remaining `is_active_seq` grep hits across `src/` are docstring/comment references only — no call sites. Test-file hits are assertions against `is_active_seq` itself, which is unchanged, so they are untouched and green.)*

### Structural fix — both `latest_active` stores now take the predicate (sites 6 + 8)

`ConfigGenerationStore.latest_active(rel_path, state_log)` → `latest_active(rel_path, is_active)`, and `PipelineStateStore.latest_active(state_log)` → `latest_active(is_active)`.

Taking a `state_log` **forced** each store to call `is_active_seq` internally, so any caller that loops was quadratic **by construction** — a shape no caller-side hoist could fix. Taking the predicate makes the hoist the caller's obligation (documented at both sites), so the fix-class cannot silently regrow at either seam.

**Correction on architect review — the pipeline store was initially deferred, and that was wrong.** My deferral rested on a **caller census** ("one run per caller today"), not on structure. But census-based safety is precisely what failed for the config store: it was census-safe until `_reconcile_config_as_of_cut` (`registry.py:1961`) grew a rel_path loop, and the defect appeared. The API shape — not the current caller set — is what has to forbid the defect, which is the whole reason I changed the config store's signature in the first place. Leaving a self-described "mirror" asymmetric would have been the path to a **4th recurrence** of a class that has now recurred three times (#2937 → #2938 → #2948). It was also **not free today**: `_seqs()` grows with a run's generation count, so each resume was O(N_run_generations × M_WAL) — not a startup killer, but the same shape.

With both stores symmetric, "mirrors `ConfigGenerationStore.latest_active` exactly" is **true again**, so the docstrings state it plainly rather than carrying an except-signature clause. (The earlier doc-only commit `b73929cf` recorded the drift; `73e24f60` removes it.)

**Blast radius**: `latest_pipeline_state`'s own signature is **unchanged**, so both production callers (`executor.py:1587`, `pipeline_executor_driver.py:202`) and all 7 test callers are untouched. The store's only caller is that wrapper, and `PipelineStateStore` has no other construction site (verified by grep: `\.latest_active(`, `latest_pipeline_state(`, `PipelineStateStore(`). No `is_active_seq` call site remains in `pipeline_recovery.py`.

## `is_active_seq` byte-unchanged — confirmed

Per the brief, `is_active_seq` has many crash-recovery callers (`pipeline_recovery.py`, `config_generations.py`, `registry.py`) and was **not** touched. `snapshot_generations.py` is **entirely absent from this diff**:

```
$ git diff origin/main -- src/reyn/core/events/snapshot_generations.py | wc -l
0
```

So `is_active_seq` is byte-unchanged **by construction**, not merely by inspection. This PR is an ADD (`build_active_predicate` import) + call-shape change only.

### 1. Scan-count bound — the actual bug (`tests/test_restore_all_subscription_scan_bound_2944.py`, Tier 2)

A real `StateLog` **subclass** counting `iter_from` calls (a **public-method override**, not a private-state peek — mirrors `tests/test_active_branch_history_scan_bound_2941.py`). Real `AgentRegistry` + real `StateLog` + the real `checkout` primitive; no mocks.

- `test_restore_all_scan_count_independent_of_wal_size` — 30-entry vs 600-entry WAL (both with a rewind, so the abandoned-interval predicate is non-trivial) must cost the **same** number of `iter_from` calls.
- `test_restore_all_scan_count_bounded_not_proportional_to_wal_entries` — a 200-entry WAL's `restore_all` must stay under a small constant of scans.

**Strip-falsification result** (reverted the hoist locally to `is_active=lambda s: is_active_seq(self._state_log, s)`, re-ran, restored):

| | 30-entry WAL | 600-entry WAL | verdict |
|---|---|---|---|
| **pre-hoist** (per-seq `is_active_seq`) | **44** scans | **614** scans | 🔴 RED — `assert 44 == 614` |
| **post-hoist** (`build_active_predicate`) | equal | equal | 🟢 GREEN — bounded |

The call count tracks WAL size 1:1 (600 entries → 614 scans), which is the O(M) scan-count signature that makes the total work O(M²). **Both** tests flipped RED on the revert and GREEN on restore.

### 2. Equivalence (`tests/test_subscription_replay_hoist_equivalence_2944.py`, Tier 2)

Restored bindings must be **identical** under the old per-seq call shape and the new hoisted one, at the real `SubscriptionRegistry.replay` call site, with real WAL `task_subscribed`/`task_rebound` entries and the real `checkout` primitive — no mocks. Read through the **public surface** (`task_ids` / `assignee_of` / `requester_of` / `requester_kind_of`) — no private-state asserts. Scenarios: **no rewind**, **one rewind**, **nested-subsuming**, **checkout-back resurrection**.

(Pointwise `build_active_predicate ≡ is_active_seq` equivalence is already covered by #2941's `test_build_active_predicate_equivalence_2941.py`; this test covers the **wiring** at this PR's call site.)

### 3. Existing suites stay green

Full `pytest` from the repo root: **8520 passed, 20 skipped, 0 failed** — including every rewind/registry/recovery suite (`test_registry_rewind_to`, `test_config_rewind_reconcile_2405`, `test_topology_rewind_2103_piece2`, `test_agent_lifecycle_rewind_2103_s2`, `test_session_vanished_emit_2154`, `test_registry_list_rewind_points_1f`, `test_conversation_rewind_2360`, the `test_2259_*` truncation suites, `test_multi_session_restore`).

## Recovery gate (CLAUDE.md truncate-falsify) — reasoning, not omission

**N/A**, with the reasoning stated explicitly rather than omitted:

The CLAUDE.md gate binds any PR "adding recovery / reconstruction functionality (WAL-event-derived state, PITR, rewind/restore paths)" — the hazard it guards is *WAL-event-derived recovery state that isn't snapshot-backed*. This PR adds **no** recovery functionality and **no persisted state**: it is a pure call-shape optimization over an existing derivation. The reconstruction **source** is unchanged (the same `_rewind_records(state_log)` scan over the same reset-records), the **derivation** is unchanged (the same `_abandoned_intervals`), and `is_active_seq` is byte-unchanged (0-line diff, above). Truncation behavior is therefore bit-identical to `origin/main`: the reset-record's truncation-survival is already pinned by `test_conversation_rewind_2360.py` (`always_keep_kinds` protects `rewind` records so `is_active_seq` can still classify post-truncation), and this PR neither strengthens nor weakens that. Same class as **#2938**, where the architect confirmed N/A on identical grounds.

**Falsification of the N/A claim:** if this PR *did* change what survives truncation, the equivalence tests above (which exercise nested/subsuming/checkout-back reconstruction against a real `StateLog`) would have to diverge between the two call shapes. They do not.

## CI gates (all run locally, **foreground**, green)

- ✅ `ruff check src tests` — All checks passed
- ✅ `python scripts/test_tier_audit.py --strict tests/test_restore_all_subscription_scan_bound_2944.py tests/test_subscription_replay_hoist_equivalence_2944.py` — 6 inspected, 0 warnings, 0 Tier-4 errors
- ✅ `pytest` from repo root — 8520 passed, 20 skipped (re-run foreground after the symmetrization: 231s)
- ✅ `python scripts/verify_module_docstrings.py src/reyn/core/events/pipeline_recovery.py src/reyn/runtime/registry.py src/reyn/core/events/config_generations.py` — OK

## Test plan

- [x] Strip-falsify the scan-bound test (revert hoist → RED at 44 vs 614 scans; restore → GREEN) — done, numbers reported above
- [x] Full-repo grep sweep for the `is_active_seq`-per-seq fix class; every site enumerated with disposition — done, tables above (**8 hoisted/symmetrized, 1 left one-shot**; the pipeline-store deferral was corrected on architect review)
- [x] Confirm `is_active_seq` byte-unchanged (`git diff` on `snapshot_generations.py` = 0 lines) — done
- [x] All four CI gates run locally in the foreground — done, green

### DIFF
--- a/src/reyn/core/events/config_generations.py
+++ b/src/reyn/core/events/config_generations.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import re
 from pathlib import Path
+from typing import Callable
 
 import yaml
 
@@ -104,18 +105,29 @@ class ConfigGenerationStore:
         )
         return seq, content if isinstance(content, dict) else {}
 
-    def latest_active(self, rel_path: str, state_log: object) -> "tuple[int, dict] | None":
+    def latest_active(
+        self, rel_path: str, is_active: "Callable[[int], bool]",
+    ) -> "tuple[int, dict] | None":
         """The (seq, content) of the highest generation for `rel_path` on the ACTIVE WAL
-        branch (``is_active_seq``), or None when no active generation exists.
+        branch, or None when no active generation exists.
+
+        ``is_active`` is the caller-supplied membership predicate (``is_active_seq``'s
+        derivation is seq-independent — see ``build_active_predicate``). A caller
+        reconciling MANY rel_paths in one pass (e.g.
+        ``AgentRegistry._reconcile_config_as_of_cut``) MUST hoist ONE
+        ``build_active_predicate(state_log)`` and reuse it here per rel_path — passing
+        ``is_active_seq`` re-bound per call would re-scan the whole WAL once per
+        rel_path (the #2941 sibling quadratic-cold-start shape this signature exists to
+        prevent). A single-path caller may pass
+        ``lambda s: is_active_seq(state_log, s)`` directly.
 
         #2405: ``latest_at_or_below(cut=N)`` has the symmetric gap — post-rewind active
         generations (seq > R > N) are excluded, reverting config to as-of-N on crash
-        recovery. ``is_active_seq`` covers all three regions correctly:
-        • Pre-target (seq ≤ N): ``is_active_seq=True`` → applied.
-        • Abandoned branch (N < seq < R): ``is_active_seq=False`` → skipped.
-        • Post-rewind active (seq > R): ``is_active_seq=True`` → applied."""
-        from reyn.core.events.snapshot_generations import is_active_seq  # noqa: PLC0415
-        seqs = [s for s in self._entries().get(rel_path, ()) if is_active_seq(state_log, s)]
+        recovery. The active-branch predicate covers all three regions correctly:
+        • Pre-target (seq ≤ N): active=True → applied.
+        • Abandoned branch (N < seq < R): active=False → skipped.
+        • Post-rewind active (seq > R): active=True → applied."""
+        seqs = [s for s in self._entries().get(rel_path, ()) if is_active(s)]
         if not seqs:
             return None
         seq = seqs[-1]
--- a/src/reyn/core/events/pipeline_recovery.py
+++ b/src/reyn/core/events/pipeline_recovery.py
@@ -25,7 +25,12 @@ Two entry points:
     split ``StateLog.append``/``append_nowait`` already draws for WAL entries.
   - ``latest_pipeline_state`` — the reader seam ``resume`` calls: the latest
     generation for a run ON THE ACTIVE WAL BRANCH (``is_active_seq``), mirroring
-    ``ConfigGenerationStore.latest_active``'s semantics exactly.
+    ``ConfigGenerationStore.latest_active``'s SEMANTICS (active-branch membership,
+    latest-wins, no forward-replay) but NOT its signature: that store multiplexes
+    MANY rel_paths, so it takes a caller-hoisted ``is_active`` predicate to keep a
+    multi-path reconcile from re-scanning the WAL per path. This store is scoped to
+    ONE run (a handful of generation files, one run per caller), so there is no
+    per-path loop to hoist out of and it takes the ``state_log`` directly.
 """
 from __future__ import annotations
 
@@ -92,8 +97,17 @@ class PipelineStateStore:
 
     def latest_active(self, state_log: object) -> "tuple[int, dict] | None":
         """The (seq, content) of the highest generation on the ACTIVE WAL branch
-        (``is_active_seq``), or None if no active generation exists. Mirrors
-        ``ConfigGenerationStore.latest_active``."""
+        (``is_active_seq``), or None if no active generation exists.
+
+        Mirrors ``ConfigGenerationStore.latest_active``'s SEMANTICS (active-branch
+        membership, latest-wins, no forward-replay) but deliberately NOT its
+        signature: that store takes a caller-hoisted ``is_active`` predicate because
+        it multiplexes MANY rel_paths, so calling ``is_active_seq`` internally made a
+        multi-path reconcile re-scan the whole WAL once per path (quadratic BY
+        CONSTRUCTION — no caller-side hoist could fix it). This store is scoped to ONE
+        run: ``_seqs()`` is that run's own handful of generation files and the sole
+        caller (``latest_pipeline_state``) resolves a single run, so there is no
+        per-path loop to hoist out of and taking the ``state_log`` stays correct."""
         from reyn.core.events.snapshot_generations import is_active_seq  # noqa: PLC0415
         seqs = [s for s in self._seqs() if is_active_seq(state_log, s)]  # type: ignore[arg-type]
         if not seqs:
--- a/src/reyn/core/events/pipeline_recovery.py
+++ b/src/reyn/core/events/pipeline_recovery.py
@@ -24,20 +24,18 @@ Two entry points:
     (``state_log.submit_durable_nowait``) path, mirroring the durable/non-durable
     split ``StateLog.append``/``append_nowait`` already draws for WAL entries.
   - ``latest_pipeline_state`` — the reader seam ``resume`` calls: the latest
-    generation for a run ON THE ACTIVE WAL BRANCH (``is_active_seq``), mirroring
-    ``ConfigGenerationStore.latest_active``'s SEMANTICS (active-branch membership,
-    latest-wins, no forward-replay) but NOT its signature: that store multiplexes
-    MANY rel_paths, so it takes a caller-hoisted ``is_active`` predicate to keep a
-    multi-path reconcile from re-scanning the WAL per path. This store is scoped to
-    ONE run (a handful of generation files, one run per caller), so there is no
-    per-path loop to hoist out of and it takes the ``state_log`` directly.
+    generation for a run ON THE ACTIVE WAL BRANCH, mirroring
+    ``ConfigGenerationStore.latest_active``'s semantics exactly (active-branch
+    membership, latest-wins, no forward-replay), down to the predicate-taking
+    signature: ``PipelineStateStore.latest_active`` takes a caller-hoisted
+    ``is_active`` so no reader can re-scan the WAL once per generation seq.
 """
 from __future__ import annotations
 
 import json
 import re
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable
 
 from reyn.core.events.config_recovery import reyn_root
 
@@ -95,21 +93,25 @@ class PipelineStateStore:
         content = json.loads(self._path_for(seq).read_text(encoding="utf-8"))
         return seq, content
 
-    def latest_active(self, state_log: object) -> "tuple[int, dict] | None":
-        """The (seq, content) of the highest generation on the ACTIVE WAL branch
-        (``is_active_seq``), or None if no active generation exists.
+    def latest_active(
+        self, is_active: "Callable[[int], bool]",
+    ) -> "tuple[int, dict] | None":
+        """The (seq, content) of the highest generation on the ACTIVE WAL branch, or
+        None if no active generation exists. Mirrors
+        ``ConfigGenerationStore.latest_active`` exactly — same active-branch
+        membership / latest-wins / no-forward-replay semantics, and the same
+        predicate-taking signature.
 
-        Mirrors ``ConfigGenerationStore.latest_active``'s SEMANTICS (active-branch
-        membership, latest-wins, no forward-replay) but deliberately NOT its
-        signature: that store takes a caller-hoisted ``is_active`` predicate because
-        it multiplexes MANY rel_paths, so calling ``is_active_seq`` internally made a
-        multi-path reconcile re-scan the whole WAL once per path (quadratic BY
-        CONSTRUCTION — no caller-side hoist could fix it). This store is scoped to ONE
-        run: ``_seqs()`` is that run's own handful of generation files and the sole
-        caller (``latest_pipeline_state``) resolves a single run, so there is no
-        per-path loop to hoist out of and taking the ``state_log`` stays correct."""
-        from reyn.core.events.snapshot_generations import is_active_seq  # noqa: PLC0415
-        seqs = [s for s in self._seqs() if is_active_seq(state_log, s)]  # type: ignore[arg-type]
+        ``is_active`` is the caller-supplied membership predicate (its derivation is
+        seq-independent — see ``build_active_predicate``). Taking a ``state_log`` here
+        would force an internal ``is_active_seq`` call per generation seq, re-scanning
+        the whole WAL each time: O(generations x WAL) per read, and quadratic BY
+        CONSTRUCTION for any caller that ever loops over runs — a shape no caller-side
+        hoist could fix. That is exactly how the config store's defect arose (it was
+        census-safe until ``AgentRegistry._reconcile_config_as_of_cut`` grew a rel_path
+        loop), so this store takes the predicate too rather than relying on today's
+        one-run-per-caller census holding forever."""
+        seqs = [s for s in self._seqs() if is_active(s)]
         if not seqs:
             return None
         seq = seqs[-1]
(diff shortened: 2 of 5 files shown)


tya5:
**[per-PR-coder]** — doc drift fixed, pushed as `b73929cf`. Non-blocking item from the lead review; co-vet can continue in parallel.

**The finding was correct and I had left a real trap.** This PR changed `ConfigGenerationStore.latest_active` to take a caller-hoisted `is_active` predicate, but `pipeline_recovery.py:28` / `:96` still claimed `PipelineStateStore.latest_active` "mirrors `ConfigGenerationStore.latest_active`'s semantics **exactly**". Since this store still takes a `state_log`, "exactly" became false the moment the signature diverged — and the failure mode is specific: a future reader sees two `latest_active` methods documented as identical, notices they aren't, and "fixes" the divergence by threading a `state_log` back into the config store — **reintroducing the by-construction quadratic this PR just removed**. The doc would have actively argued for the regression.

**What both sites now say:** the **semantics** still mirror (active-branch membership, latest-wins, no forward-replay); the **signature** deliberately does not, and why —

- `ConfigGenerationStore` multiplexes **many** rel_paths, so an internal `is_active_seq` made a multi-path reconcile re-scan the WAL once per path = quadratic **by construction** (no caller-side hoist could fix it) → it takes the predicate.
- `PipelineStateStore` is scoped to **one** run (`_seqs()` = that run's handful of generation files; sole caller `latest_pipeline_state` resolves a single run) → no per-path loop to hoist out of → `state_log` stays correct.

That is the deferral rationale from the sweep table now living **at the site it governs**, not only in this PR body — so the next reader of the divergence finds the reason without archaeology through a merged PR.

**Gates re-run in the foreground on the doc commit** (docstring-only change, but `verify_module_docstrings` covers exactly these files, so it was not assumed):
- ✅ `ruff check src tests` — All checks passed
- ✅ `python scripts/verify_module_docstrings.py src/reyn/core/events/pipeline_recovery.py src/reyn/runtime/registry.py src/reyn/core/events/config_generations.py` — OK
- ✅ `pytest` from repo root — **8520 passed, 20 skipped, 0 failed** (234s)

No source behavior changed; the sweep dispositions, the strip-falsify numbers (44 vs 614 scans), and the `is_active_seq` 0-line diff in the PR body all still hold as written.

tya5:
**[architect]** — co-vet: **the fix is correct and merge-clear; one in-PR-recommended finding — the structural argument is applied asymmetrically.**

**First, I own the #2938 miss.** I passed #2938 CLEAR having *enumerated* these very callers in my pre-ground — and dismissed them, in my own words, as *"for #2938 this didn't matter (is_active_seq was byte-unchanged → all callers unaffected regardless)"*. That framing answered **"does this fix break other callers?"** and never asked **"do other callers have the same pathology?"** — which is exactly my own strip-every-sibling-guard-site pin, unapplied because I filed a *perf* fix under *correctness* review. 7 siblings survived. Noted as a discipline lesson, not just an apology.

## Verified
- **Sweep completeness (independently enumerated, no truncation):** post-fix there are exactly **8 `build_active_predicate(` call sites** (7 new + session.py from #2938) and **2 real `is_active_seq(` call sites** (the rest of the grep hits are docstrings). That matches the PR's claim exactly.
- **`registry.py:1364` (`rewind_to`) — keep-per-seq is CORRECT.** `async def rewind_to(self, target_n: int)` → one call for one target. Hoisting a predicate for a single seq is pointless. ✓
- **Strip-falsify reproduced:** reverted the `restore_all` hoist → `assert 44 == 614` (2 failed). Scan count tracks WAL size 1:1 = the O(M) signature behind the O(M²). Sound pin; baseline 6/6 green.
- **`is_active_seq` byte-unchanged:** `snapshot_generations.py` is absent from the PR's file list ✓ → recovery callers untouched. **Recovery gate N/A** agreed: no new persisted state, source + derivation unchanged (same class as #2938).
- **Stale base is safe here — verified, not assumed.** Base is `a2483d42` (missing #2947). #2948's own diff (merge-base→HEAD) = `config_generations.py` / `registry.py` / 2 tests; #2947 = `chat.py` / `mcp.py` / `llm.py` / `test_httpx_lazy_load_cli_entry.py`. **Intersection is empty** → the squash applies only #2948's files, #2947 is preserved. Lead's judgment holds.
- **Point 1 (API shape forced the defect) — the reading is correct.** `ConfigGenerationStore.latest_active` now takes `is_active: Callable[[int], bool]`; `registry.py:1955-1961` builds the predicate once and reuses it across the `rel_path` loop. Taking `state_log` *did* force the internal per-seq rescan, which no caller-side hoist could remove. Structurally closed. ✓

## Finding (in-PR-recommended, not a blocker) — the same structural argument applies to `PipelineStateStore.latest_active`, which keeps the defect-permitting shape

| store | signature after this PR | fix-class |
|---|---|---|
| `ConfigGenerationStore.latest_active` | `(rel_path, is_active: Callable)` | **closed** — a caller *cannot* force a rescan |
| `PipelineStateStore.latest_active` | **`(state_log)`** | **open** — internally `[s for s in self._seqs() if is_active_seq(state_log, s)]` |

- The justification for keeping it — *"scoped to 1 run, no multiplex"* — is a **caller-census argument, not a structural one**. I verified the census is true **today**: `latest_pipeline_state` has exactly 2 callers (`executor.py:1587`, `pipeline_executor_driver.py:202`) and both are one-shot per run execution, not loops. So **no live quadratic → not a blocker.**
- But a caller census is precisely what the API change exists to stop relying on: **ConfigGenerationStore's own defect arose that way** (`registry.py:1961` loops `rel_path`). Nothing structurally prevents a future caller from looping `latest_pipeline_state` over runs and silently re-opening this class.
- It isn't free today either: `_seqs()` enumerates the run's generation files, so each resume is **O(N_run_generations × M_WAL)** — same per-seq×full-scan shape, just not the startup killer.
- **Your "minor" is the symptom, and the cure is symmetry, not a doc edit.** `pipeline_recovery.py:28,96` claims it *"mirrors `ConfigGenerationStore.latest_active`'s semantics exactly"* — now false **because of this asymmetry**. Updating the doc to describe the divergence records the drift; making Pipeline take the predicate too (2 caller updates: `executor.py:1587`, `pipeline_executor_driver.py:202`) makes the sentence true again *and* closes the class. Cheap, and consistent with the PR's own thesis.

**Verdict: merge-clear.** The startup O(WAL²) is genuinely fixed, the sweep is complete against my independent enumeration, the scan-bound pin is load-bearing, and `rewind_to`'s keep is right. Recommend folding the `PipelineStateStore` symmetry in-PR (or tracking it explicitly) — leaving one store predicate-taking and its self-declared mirror `state_log`-taking is how this class re-opens a fourth time.

tya5:
**[per-PR-coder]** — architect's in-PR recommendation applied. `73e24f60` pushed, PR body sweep table updated. Ready for merge.

**The verdict is correct and it lands on my own reasoning.** My deferral of `PipelineStateStore.latest_active` rested on a **caller census** — "one run per caller, no multiplex today". That is not a structural argument. And census-based safety is exactly what failed for the config store: it was census-safe right up until `_reconcile_config_as_of_cut` (`registry.py:1961`) grew a rel_path loop, and the defect appeared underneath it. The reason I changed the config store's API at all was to stop depending on the census — so exempting the pipeline store on census grounds applied the insight everywhere **except to itself**. With the class already at three recurrences (#2937 → #2938 → #2948), leaving a self-described "mirror" asymmetric is a documented invitation to the 4th.

Two things I had also under-weighted, both of which the architect flagged and I've verified:
- **Not free today.** `_seqs()` grows with a run's generation count, so each resume was already O(N_run_generations × M_WAL) — not a startup killer, same shape.
- **The doc fix was symptom-recording.** "mirrors exactly" went false *because of the asymmetry*; rewriting it to "except signature" pins the drift in prose instead of removing it. Symmetrizing makes the sentence true again **and** closes the fix-class — one change, both problems. `b73929cf` recorded the drift; `73e24f60` removes it.

**Change:**
- `PipelineStateStore.latest_active(state_log)` → `latest_active(is_active)`. No `is_active_seq` call site remains in `pipeline_recovery.py`.
- `latest_pipeline_state` hoists `build_active_predicate` once and passes it; docstring notes a multi-run caller should hoist above its own loop and drive the store directly.
- `pipeline_recovery.py:19-28` and the method docstring restored to a plain, now-true "mirrors `ConfigGenerationStore.latest_active` exactly" — no except-clause. Each site carries *why* the predicate shape is load-bearing (the config-store census failure), so the next reader can't re-derive the exemption.
- PR body sweep table: **8 hoisted/symmetrized, 1 deferred** (`rewind_to` only — single user-supplied target, one WAL scan either way; architect concurred).

**Blast radius (verified by grep, not assumed):** `latest_pipeline_state`'s signature is **unchanged**, so both production callers (`executor.py:1587`, `pipeline_executor_driver.py:202`) and all 7 test callers are untouched. The store's only caller is that wrapper; `PipelineStateStore` has no other construction site.

**Gates re-run in the foreground on `73e24f60`:**
- ✅ `ruff check src tests` — All checks passed
- ✅ `python scripts/verify_module_docstrings.py src/reyn/core/events/pipeline_recovery.py src/reyn/runtime/registry.py src/reyn/core/events/config_generations.py` — OK
- ✅ pipeline + this PR's suites — 88 passed
- ✅ `pytest` from repo root — **8520 passed, 20 skipped, 0 failed** (231s)
- ✅ `is_active_seq` still byte-unchanged: `git diff origin/main -- snapshot_generations.py` = **0 lines**

Strip-falsify numbers (44 vs 614 scans) and the recovery-gate N/A reasoning are unaffected — no change to the WAL-scan call shape at `restore_all`, and still no new persisted state.

PR #2952, opened 2026-07-15
feat: builtin vector-store + chunker MCP servers (FP-0063 P2)

tya5:
**[per-PR-coder]** — FP-0063 P2: the builtin MCP layer of the "builtin turnkey user RAG" arc (docs/deep-dives/proposals/0063-builtin-turnkey-user-rag.md). P1 (selection spike) found no OSS MCP vector-DB server satisfying C1 (pre-computed vectors); owner approved writing the MCP surface ourselves. This PR ships the two thin bundled MCP servers P1's fallback anticipated.

## What's here

- `src/reyn/builtin/mcp_servers/vector_store_server.py` — wraps `sqlite-vec` (a pure-C SQLite extension, no bundled embedding model).
- `src/reyn/builtin/mcp_servers/chunker_server.py` — wraps `chonkie`.
- `pyproject.toml`: new `builtin-rag` extra (`sqlite-vec`, `apsw`, `chonkie`) — `fastmcp` is **already core** (promoted in an earlier PR), so this is the only new-dependency surface. `uv.lock` regenerated (`uv lock`) and `uv lock --check` passes.
- `docs/cookbook/configs/with-builtin-rag-mcp.yaml` — the INERT sample config (commented out) an operator copies + uncomments to enable either server.
- `tests/test_fp0063_p2_builtin_mcp_rag.py` — 10 Tier-2b tests, no mocks.
- `.github/workflows/test.yml` — adds `builtin-rag` to the CI extras list. **This one is load-bearing, see below.**

## ⚠️ The CI extras fix — caught by falsification, not by reading

The test file `pytest.importorskip`s `apsw`/`sqlite_vec`/`chonkie`. CI's install line was `pip install -e ".[dev,mcp,web,fs-watch,observability]"` — **no `builtin-rag`**. So as first written, this PR would have gone **green in CI while the entire test module silently skipped, verifying nothing** — the exact "green CI, zero verification" failure class this repo has been bitten by before.

Verified both directions on clean venvs rather than assuming:

```
# WITHOUT builtin-rag (the pre-fix CI state):
SKIPPED [1] tests/test_fp0063_p2_builtin_mcp_rag.py:44: builtin-rag extra not installed
1 skipped in 0.14s

# WITH the fixed CI extras string, clean venv, resolved from PyPI:
10 passed in 2.54s
```

The workflow change is CI config, **not** a reyn-core change — `src/reyn/**` is untouched by this PR (the zero-core target holds).

## The five vector-store gates (proposal §Vector-DB MCP)

1. **Pre-computed vectors** — `upsert(db_path, items)` takes `items: [{"id", "vector", "metadata"}]`; the module never computes an embedding.
2. **User-specified single sqlite file** — every tool call takes `db_path` explicitly; parent dirs are created lazily.
3. **top-k + metadata filter** — `query(db_path, vector, top_k, filters)` runs a real SQL `WHERE` over the metadata columns joined against the sqlite-vec KNN match (see "why k=total-rows" in the module docstring — the filter is applied to the FULL candidate set, not post-`top_k`, so a filter can't be defeated by an early KNN truncation).
4. **Metadata passthrough** — `reyn_rag_chunks` carries the full `ChunkMetadata` shape: `source_path`, `source_type`, `content_hash`, `embedding_model`, `chunk_index`, `size_tokens`, `parent_context`, `extra` (JSON blob).
5. **Generic ops for a pipeline-owned diff** — `list_metadata(db_path, filters)` (Chroma `get(where=...)` shape, no vectors), `upsert` (replaces by `id`, never duplicates), `delete(db_path, ids)`. **No `content_hash` add/update/remove logic lives here** — that's the ingest pipeline's job (P3), settled at co-vet (C5).

## Chunker: size/overlap are parameters, not constants (R4)

`chunk(text, size=400, overlap_ratio=0.125, min_characters_per_chunk=24)` — defaults land in the 2026 persistent-RAG band (256–512 tokens, 10–15% overlap), explicitly NOT FP-0057 line 51's 800–1024 figure (that number is scoped to throwaway ephemeral attachments). Overlap is chonkie's own `OverlapRefinery` two-step shape (chunk, then refine) — `RecursiveChunker` itself has no overlap param.

## Driven, not just read (per task instruction C)

Every claim above was exercised, not inferred from reading the code:

**In-process (fastmcp `Client`)**:
```
TOOLS: ['upsert', 'query', 'list_metadata', 'delete']
UPSERT RESULT: {'upserted': 1}
QUERY RESULT: [{'id': 'x1', 'distance': 0.0, 'metadata': {...}}]
LIST RESULT: [{'id': 'x1', 'metadata': {...}}]
DELETE RESULT: {'deleted': 1}
```

**Real subprocess stdio round-trip** (`PythonStdioTransport` spawning `python -m reyn.builtin.mcp_servers.vector_store_server` as a genuine child process over stdin/stdout, not in-process):
```
SUBPROCESS TOOLS: ['upsert', 'query', 'list_metadata', 'delete']
SUBPROCESS UPSERT: {'upserted': 1}
SUBPROCESS QUERY: [{'id': 's1', 'distance': 0.0, 'metadata': {...}}]
file at user path exists: True
```

**Chunker, in-process**:
```
TOOLS: ['chunk']
num chunks: 100  (size=64, overlap_ratio=0.2, 200-sentence input)
default chunks: 15, token_count[0]: 431   (size/overlap defaults only)
```

## Strip-falsification (required)

Temporarily broke `list_metadata`'s column restriction (joined the vector table and leaked the raw embedding blob into the returned metadata dict), then ran `tests/test_fp0063_p2_builtin_mcp_rag.py`. Observed RED on exactly the two tests that guard this:

```
FAILED tests/test_fp0063_p2_builtin_mcp_rag.py::test_list_metadata_excludes_vectors
FAILED tests/test_fp0063_p2_builtin_mcp_rag.py::test_strip_falsify_list_metadata_would_leak_vectors_if_joined
2 failed, 8 passed in 0.94s
```
Reverted; full suite green again (`10 passed`). The `test_strip_falsify_...` test also independently proves the guard isn't vacuous — it opens a second, separate `apsw` connection to the same on-disk file and confirms the vector bytes really are retrievable via the naive join (i.e. the omission in `list_metadata` is a deliberate choice guarding real data).

## Inert-ship posture (A / R3 / F-D) — verified, not assumed

Neither server is wired into `reyn.builtin.registry.build_builtin_config()`. Confirmed by direct investigation of `#2932`'s auto-grant (`src/reyn/interfaces/cli/commands/pipe.py`): it keys **purely on the server name being a key in the merged `config.mcp.servers` dict** (`sorted((config.mcp or {}).get("servers", {}) or {})`) — there is no `enabled` field anywhere in the MCP server schema for it to check, so **any** entry appearing under `mcp.servers` from **any** config tier (including a hypothetical builtin tier) would be auto-granted the instant `reyn pipe run` executes, with zero operator decision in the chain. That's why this PR does NOT add an `mcp` key to `build_builtin_config()`'s output — the sample config lives purely in `docs/cookbook/configs/with-builtin-rag-mcp.yaml` (mirroring the existing `with-mcp.yaml` precedent — those files are never auto-loaded; confirmed by grep, the only reference to `docs/cookbook/*.yaml` in `src/` is a CLI help-text `print()` in `interfaces/cli/commands/init.py`). An operator must copy the block into their own `reyn.yaml` and uncomment it themselves — that's the only way either server's name ever reaches `mcp.servers`.

Zero reyn-core files changed. Only new/added: two builtin-content modules, an extras group + lockfile update, a doc sample, and a test file.

## Dependencies added (builtin-rag extra, base install unaffected)

| Package | Version | License |
|---|---|---|
| `sqlite-vec` | 0.1.9 | dual MIT / Apache-2.0 |
| `apsw` | ≥3.51 (3.53.3.1 locked) | zlib-like ("any-OSI" per PyPI classifier) |
| `chonkie` | ≥1.7 | MIT |

`fastmcp` (already core) is unchanged.

## Offline/firewall finding (R1/D) — reported, not papered over

- **No HuggingFace / no network fetch at all**, in either server: `sqlite-vec` is a pure-C extension with no bundled model; `chonkie`'s default (character) tokenizer used here downloads nothing. Neither server repeats the FP-0057 line-55 firewall failure class.
- **A different, previously-undiscovered environment gate**: `sqlite-vec` loads via `sqlite3.Connection.load_extension`, which the **stdlib `sqlite3` module is not guaranteed to support** — confirmed directly: this very repo's own pyenv-built CPython 3.12.7 `.venv` has `hasattr(sqlite3.Connection, "enable_load_extension") is False` (same for `/usr/bin/python3` on this Mac; `/opt/homebrew/bin/python3` and `python3.13` DO support it — it varies by distribution/build flags, not by OS). Rather than silently fail (or worse, work on the author's machine and break on the operator's), `vector_store_server.py` uses `apsw` instead of stdlib `sqlite3` — apsw statically bundles its own SQLite build with extension loading always enabled, so `sqlite-vec` loads deterministically regardless of how the host interpreter was compiled. This is a portability fix baked into the design, not a workaround left for later — but it's worth flagging in review as a real, previously-unstated constraint on any future sqlite-vec-based work in this codebase.

## Out of scope (per task)

Ingest/query pipelines (P3), the skill (P4), the firewall step-0 reachability check (P5), and `markitdown-mcp`'s config entry (left to P3, third-party server, not written here).

## Test plan

- [x] `ruff check src tests` — green ("All checks passed!").
- [x] `python scripts/test_tier_audit.py --strict tests/test_fp0063_p2_builtin_mcp_rag.py` — 10/10 OK, 0 errors.
- [x] `pytest` from repo root — green: **8552 passed, 17 skipped** in 239s (none of the 17 are this PR's).
- [x] **Verified this PR's 10 tests RUN, not skip** — `pytest tests/test_fp0063_p2_builtin_mcp_rag.py -v -rs` → 10 PASSED / 0 skipped, against really-installed `sqlite-vec 0.1.9` / `apsw 3.53.3.1` / `chonkie 1.7.0`. Re-verified on a from-scratch venv built with the exact CI extras string (10 passed), and falsified in the negative direction (without the extra → module skips).
- [x] `python scripts/verify_module_docstrings.py <changed src files>` — OK.
- [x] `uv lock --check` — green after `uv lock` regeneration.
- [x] Driven tool calls (in-process + real subprocess stdio) for both servers — see above.
- [x] Strip-falsification — RED observed and reverted, see above.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,11 @@ jobs:
         #   web  — fastapi + uvicorn + websockets, used by tests/web/
         #   fs-watch — `watchdog` (#2608 H4), used by tests/test_fs_watcher.py
         #   observability — opentelemetry SDK, used by tests/observability/
-        run: pip install -e ".[dev,mcp,web,fs-watch,observability]"
+        #   builtin-rag — sqlite-vec + apsw + chonkie (FP-0063 P2), used by
+        #     tests/test_fp0063_p2_builtin_mcp_rag.py. Load-bearing: that file
+        #     `pytest.importorskip`s all three, so WITHOUT this extra the whole
+        #     module silently SKIPS and CI goes green having verified nothing.
+        run: pip install -e ".[dev,mcp,web,fs-watch,observability,builtin-rag]"
 
       - name: Run tests
         # --timeout=120: a per-test wall-clock cap (pytest-timeout). Without it a
--- a/docs/cookbook/configs/with-builtin-rag-mcp.yaml
+++ b/docs/cookbook/configs/with-builtin-rag-mcp.yaml
@@ -1,0 +1,50 @@
+# FP-0063 P2 -- builtin turnkey user RAG, the MCP layer.
+#
+# This entire ``mcp.servers`` block ships COMMENTED OUT / INERT. Do not
+# assume uncommenting it is safe by default: an MCP server entry, once it
+# appears under ``mcp.servers.<name>`` in ANY merged config tier, is
+# auto-granted the moment you run ``reyn pipe run`` (#2932's
+# trusted-by-configuration auto-grant -- it keys purely on the server's
+# NAME being present in the merged config, not on any "enabled" flag,
+# because no such flag exists on an MCP server entry today). That premise
+# ("the operator explicitly configured this server") only holds if YOU are
+# the one who copies this block into your own reyn.yaml and uncomments it
+# -- reyn itself never wires this in (see reyn.builtin.mcp_servers module
+# docstring). Read the two servers before enabling them: the vector-store
+# server writes to whatever sqlite file `db_path` names; the chunker server
+# reads no filesystem paths itself but the ingest pipeline that will call
+# it (a later phase, P3) reads document files from the folder you point it
+# at.
+#
+# Install the extra runtime deps first (NOT part of reyn's base install):
+#   pip install "reyn[builtin-rag]"
+#
+# Then copy this block into your reyn.yaml (or reyn.local.yaml) and
+# uncomment it:
+#
+# permissions:
+#   mcp.reyn_vector_store: allow
+#   mcp.reyn_chunker: allow
+#
+# mcp:
+#   servers:
+#     reyn_vector_store:
+#       type: stdio
+#       command: python
+#       args: ["-m", "reyn.builtin.mcp_servers.vector_store_server"]
+#       # Tools: upsert(db_path, items) / query(db_path, vector, top_k,
+#       # filters) / list_metadata(db_path, filters, limit) /
+#       # delete(db_path, ids). Every call takes db_path explicitly --
+#       # there is no server-side default; you always name the sqlite file.
+#
+#     reyn_chunker:
+#       type: stdio
+#       command: python
+#       args: ["-m", "reyn.builtin.mcp_servers.chunker_server"]
+#       # Tool: chunk(text, size=400, overlap_ratio=0.125,
+#       # min_characters_per_chunk=24). size/overlap are real parameters,
+#       # not baked-in constants -- tune them per document type.
+#
+# A third-party parser server (e.g. `uvx markitdown-mcp`, Microsoft,
+# converts txt/md/pdf/xlsx/pptx/docx -> Markdown) belongs alongside these
+# two once the ingest pipeline (P3) lands; out of scope for this file.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,30 @@ web = [
 # empty extra so existing `pip install -e ".[mcp]"` invocations (CI, user docs,
 # `reyn[mcp]`) keep resolving instead of erroring on an unknown extra.
 mcp = []
+# FP-0063 P2 builtin turnkey user RAG -- the two thin builtin MCP servers in
+# `reyn.builtin.mcp_servers` (vector-store over sqlite-vec, chunker over
+# chonkie). `fastmcp` itself is already core (see above); these three are the
+# ONLY new deps this arc adds, kept extras-only so a base `pip install reyn`
+# is unaffected. Neither server is wired into the merged config by default —
+# they ship INERT (see `reyn.builtin.mcp_servers` module docstring) — so
+# installing this extra alone activates nothing.
+#   - sqlite-vec (dual MIT/Apache-2.0): pure-C SQLite extension, no bundled
+#     embedding model -> no HuggingFace-fetch hazard (FP-0057 line 55).
+#   - apsw (zlib-like "any-OSI"): sqlite-vec loads via
+#     `sqlite3.Connection.load_extension`, which the STDLIB `sqlite3` module
+#     is not guaranteed to support (e.g. this repo's own pyenv-built
+#     CPython 3.12.7 .venv has no `enable_load_extension`). apsw statically
+#     bundles its own SQLite with extension loading always enabled, so
+#     loading is portable regardless of how the host interpreter's `sqlite3`
+#     was compiled.
+#   - chonkie (MIT): recursive/semantic/etc. chunking, ~15 MiB vs LangChain's
+#     ~80 MiB; no bundled model in the default character-tokenizer path used
+#     here -> no network fetch at import or call time.
+builtin-rag = [
+    "sqlite-vec>=0.1.9",
+    "apsw>=3.51",
+    "chonkie>=1.7",
+]
 # #2608 H4: filesystem watcher external-event source (`fs_watch:` config block ->
 # `file_changed` hook). Extras-only — a build with no fs_watch config configured
 # never imports this, and a build WITH config but no extra installed degrades to
--- a/src/reyn/builtin/mcp_servers/__init__.py
+++ b/src/reyn/builtin/mcp_servers/__init__.py
@@ -1,0 +1,44 @@
+"""Builtin MCP servers (FP-0063 P2 -- builtin turnkey user RAG).
+
+Two thin bundled MCP servers that the ingest/query pipelines (P3, a later
+phase) will call as external MCP vessels -- exactly like any other
+operator-configured MCP server (FP-0057 C2: the user RAG store is EXTERNAL,
+never in-core; reyn contributes no store/chunker code path of its own,
+only these standalone scripts that happen to ship inside the reyn wheel):
+
+- :mod:`reyn.builtin.mcp_servers.vector_store_server` -- wraps
+  ``sqlite-vec`` (dual MIT/Apache-2.0, no bundled embedding model). Accepts
+  externally pre-computed vectors (FP-0057 C1: reyn is the SOLE embedder),
+  writes to a user-specified single sqlite file, and exposes generic
+  upsert/query/list/delete ops -- the ingest pipeline (not this server)
+  owns the ``content_hash`` add/update/remove diff (C5, settled at co-vet
+  in the proposal).
+- :mod:`reyn.builtin.mcp_servers.chunker_server` -- wraps ``chonkie``
+  (MIT). ``size``/``overlap`` are tool parameters with 2026-default
+  values (recursive, 256-512 tokens, 10-15% overlap), never hardcoded
+  constants, so the first thing a template-copying user wants to tune is
+  easy to find (R2/R4 in the proposal).
+
+Neither server is wired into ``reyn.builtin.registry.build_builtin_config()``
+-- doing so would make its ``mcp.servers.<name>`` key "configured" the
+instant reyn is installed, which would let ``reyn pipe run``'s
+trusted-by-configuration auto-grant (#2932) silently fire with no operator
+decision anywhere in the chain (proposal R3 / architect co-vet F-D). Both
+servers ship as pure INERT sample content instead: a commented config block
+an operator must copy into their own ``reyn.yaml`` and explicitly enable --
+see ``docs/cookbook/configs/with-builtin-rag-mcp.yaml``. This mirrors the
+precedent set for builtin skills (``force_auto_invoke_false``, "A3
+inert-ship" in ``reyn.builtin.registry``), adapted to MCP's structurally
+different inertness mechanism: an ``mcp.servers`` entry has no ``enabled``
+flag anywhere in the codebase today (#2932's auto-grant keys purely on
+dict-key presence), so absence-from-merged-config is the only inert
+posture available without inventing new core schema + auto-grant code --
+which is out of scope for this arc's "zero reyn core change" target.
+
+Both are runnable as ``python -m reyn.builtin.mcp_servers.<module>`` (stdio
+transport, via ``fastmcp`` -- already a core dependency, see
+``pyproject.toml``). The extra runtime deps they need
+(``sqlite-vec``/``apsw``/``chonkie``) live in the ``builtin-rag`` optional
+extra so a base ``pip install reyn`` is unaffected.
+"""
+from __future__ import annotations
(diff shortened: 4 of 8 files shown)


tya5:
**[per-PR-coder]** — Follow-up evidence: **gate 5 verified sufficient for P3's C5 diff, by driving it.**

Gate 5's stated purpose is "generic ops sufficient for a PIPELINE-owned incremental diff" — but the PR as opened only proved each op works *individually*. That leaves the actual load-bearing claim (P3 can build the `content_hash` add/update/remove diff on top of these ops, with no hash semantics in the server) as an inference. So I drove it, simulating a second ingest run over a changed folder using **only** `list_metadata` / `upsert` / `delete`:

```
add: ['c.md#0'] | update: ['a.md#1'] | remove: ['b.md#0']
skipped (no re-embed, X5 cost lever): ['a.md#0']
final index: {'a.md#0': 'alpha', 'c.md#0': 'delta', 'a.md#1': 'beta-EDITED'}

OK: full add/update/remove diff converges using ONLY list_metadata/upsert/delete
```

All three C5 branches resolve correctly and the index converges exactly onto the new source state (asserted, not eyeballed). Two things this confirms beyond the tests already in the PR:

- **X5's cost lever works**: the unchanged chunk is identified as skippable from the metadata listing **alone** — no vector fetch, no re-embed. That's the "re-ingesting an unchanged folder costs ≈0" requirement, and it falls out of `list_metadata` returning `content_hash` without vectors.
- **C5 ownership holds**: the diff logic lives entirely in the caller (~10 lines of plain set arithmetic). The server contributed no hash semantics, so it stays generic and a future backend swap stays "re-point the MCP server".

This is a driven probe, not a new test — the P3 pipeline is where this logic actually lands, so pinning it here would test a caller that doesn't exist yet.

tya5:
**[per-PR-coder]** — Second follow-up: **verified the servers work from a real wheel install**, not just the editable checkout.

The servers are only useful if `python -m reyn.builtin.mcp_servers.<module>` resolves for an operator who `pip install`ed reyn — my local testing all ran against `-e .` with `src/` on the path, which would mask a packaging gap. `pyproject.toml`'s `packages = ["src/reyn"]` + the `builtin/**/*` artifact glob *should* cover it, but that's a reading of the config, not evidence. So I built the wheel and checked:

```
$ unzip -l dist/reyn-0.1.0a2-py3-none-any.whl | grep mcp_servers
     2641  reyn/builtin/mcp_servers/__init__.py
     4285  reyn/builtin/mcp_servers/chunker_server.py
    16808  reyn/builtin/mcp_servers/vector_store_server.py
```

Then installed that wheel into a clean venv (`pip install "reyn-...whl[builtin-rag]"`) and drove both servers from outside the repo, with no `src/` on `sys.path`:

```
chunker: 41 chunks
vector store query: w1
wheel install fully functional
```

So the operator-facing path in the cookbook config (`command: python`, `args: ["-m", "reyn.builtin.mcp_servers.vector_store_server"]`) is real on a wheel install, and the `builtin-rag` extra resolves from PyPI on a fresh environment. CI's `dev==wheel repo self-access parity` check is also green.

tya5:
**[per-PR-coder]** — Third follow-up, completing the **R1/D dependency-posture** report with a platform finding I should have included up front.

My original offline finding covered the *network* axis (no HF fetch, no model download — that part holds). But `sqlite-vec` is a **binary** package, and I'd only driven it on macOS arm64. Checked the actual distribution surface on PyPI:

```
sqlite-vec 0.1.9 — all distributed files (no sdist fallback):
   sqlite_vec-0.1.9-py3-none-macosx_10_6_x86_64.whl
   sqlite_vec-0.1.9-py3-none-macosx_11_0_arm64.whl
   sqlite_vec-0.1.9-py3-none-manylinux_2_17_aarch64...whl
   sqlite_vec-0.1.9-py3-none-manylinux_2_17_x86_64...whl
   sqlite_vec-0.1.9-py3-none-win_amd64.whl
```

**Good news for CI and for normal use**: macOS (both arches), glibc Linux (x86_64 + aarch64) and Windows are all covered — CI's ubuntu-x86_64 × py3.11/3.12 resolves fine, as does `apsw` (cp310–cp312 manylinux wheels) and `chonkie` (pure-python `py3-none-any`).

**The honest caveat**: `sqlite-vec` publishes **no sdist**, so platform support is wheel-only with no source-build fallback. Notably **musl (Alpine) is not covered** — `pip install reyn[builtin-rag]` will fail outright on an Alpine-based image rather than degrade. That's arguably the *right* failure (loud, at install time, not a silent mid-ingest break), and it only affects operators who opt into the extra — the base `pip install reyn` is untouched. But it is a real constraint on the P5 "documented pre-install path" work, so flagging it here rather than letting P5 discover it.

No action taken in this PR — reporting per the "report what you find, do not paper over it" instruction.

tya5:
**[per-PR-coder]** — CI green (7/7), and the skip question is now settled by **direct measurement rather than inference**.

The concern this addresses: the test module `pytest.importorskip`s its three deps, so if CI lacked the `builtin-rag` extra the whole file would silently skip and CI would go green having verified nothing. This PR adds the extra to the CI install line — but "I added it, therefore the tests ran" is still an inference chain, so I closed it against the actual run:

1. **CI really installed the deps** — from the `pytest (Python 3.12)` job log's `Successfully installed`: `sqlite-vec-0.1.9`, `apsw-3.53.3.1`, `chonkie-1.7.0`, `chonkie-core-0.10.2`.
2. **CI really ran the 10 tests** — the decisive check, since `-q` hides test names. Comparing totals against the most recent green CI run on `main` (run `29424822594`, same job, same python):

| | passed | skipped |
|---|---|---|
| main (baseline) | 8554 | 16 |
| this PR | 8564 | 16 |
| **delta** | **+10** | **+0** |

Exactly +10 passed and **+0 skipped** — the 10 new tests executed in CI. Had the extra been missing, the delta would have been +0 passed / +1 skipped (the shape I falsified locally on a clean venv earlier in this PR).

All 7 checks green: pytest 3.11 / pytest 3.12 / ruff / test-tier audit / uv lock --check / docs build (strict) / dev==wheel repo self-access parity.

tya5:
**[architect]** — co-vet: **CLEAR.** Zero core change, C1 held, and every FP-0063 co-vet finding of mine landed — including the one I care most about, implemented better than I asked. Two notes, neither blocking.

## Owner-frame + my proposal findings — verified, not taken on trust
- **Zero core change ✓** — the diff touches `src/reyn/builtin/**` only (that *is* the deliverable; my first grep flagged `^src/reyn/` and I had to correct it — builtin ≠ core).
- **C1 gate ✓ and the P1 fallback fired as designed.** The docstring records that P1 found **no off-the-shelf MCP vector-DB accepting pre-computed vectors**, so the pre-defined fallback (thin wrapper = builtin content) was taken — exactly the pass/fail shape I signed off on, not an open-ended slide. `upsert` takes `vector: list[float]`; *"this module never computes an embedding itself"* → C1 holds, and with it C4 (the server can't split the vector space).
- **C5 = pipeline-owned ✓** (*"the ingest pipeline (not this server) owns the `content_hash` add/update/remove diff"*) and the server exposes **generic upsert/query/list/delete** — the gate-relaxation I recommended, which is what made P1's candidate set survivable.
- **R2/R4 ✓** — `size`/`overlap` are **tool parameters with 2026 defaults, never hardcoded constants**, so the first thing a template-copying user tunes is findable.
- **★ F-D (my security finding) — implemented, and better than I asked.** I flagged that a bundled MCP config × #2932's pipe-run auto-grant would trust-by-configuration without an explicit operator act. Verified structurally, not by the comment: `grep` finds **no registration** in `builtin/registry.py`, the docstring states *"Neither server is wired into `build_builtin_config()`"*, and the cookbook config ships **commented out** — with the **#2932 rationale written at the exact line a user would uncomment** (*"once it appears under `mcp.servers.<name>` in ANY merged config tier, it is auto-granted the moment you run `reyn pipe run`"*). That is the structural close: the reason lives where the mistake would be made, so it can't be re-derived wrongly.

## The CI skip hazard — I reproduced it independently
Running the suite in **my** environment (no `builtin-rag` extra): **`1 skipped in 0.12s`**. So the `importorskip` module-skip is real and the test.yml change is **load-bearing** — without it CI is green having verified nothing. Good catch by the coder, and the comment naming the failure mode in the workflow is the right place for it.

## The two honest findings vs the later phases
**They do not break P3/P4, but they move R1/P5 from "a risk" to the arc's gating question.**
- **② wheel-only / musl — does NOT touch reyn's own path.** reyn's container base is **`python:3.12-slim`** (glibc), not Alpine (`src/reyn/environment/reyn_base.Dockerfile:11`). So the containerized turnkey path is unaffected; the loss is Alpine users, extras-only, loud at install. Acceptable.
- **① apsw + ② no-sdist compound R1.** The turnkey path now needs **three PyPI wheels** (sqlite-vec, apsw, chonkie), and **sqlite-vec has no sdist** — so an internal mirror running in sdist-only mode cannot serve it *and* there is no source-build fallback. That is the owner's own firewalled network, and it strikes at step 1 of the P3 pipeline. My proposal-time R1 recommendation (fail-loud + a documented pre-install path) is no longer polish; it is the difference between "turnkey" and "dies before the first chunk". P5 should state the pre-install path as **"install the extra while online / from a wheel-carrying mirror"** — "use an internal mirror" alone is insufficient here.

## Note (P3/P5, not this PR)
The missing-extra failure is a **bare `ImportError: No module named 'apsw'`** raised inside an MCP **server subprocess** — loud, but **not decision-enabling**: it names neither the remedy (`pip install 'reyn[builtin-rag]'`) nor the extra, and the user may only see an opaque "server failed to start" at the pipeline boundary. The `require_mcp` error (#2932) is the precedent worth copying — cause + the two concrete ways to fix it. Right home is P5 (or wherever P3 surfaces server-start failures).

**Verdict: CLEAR, merge-clear.** Inert-by-construction with the rationale at the point of use, C1/C4/C5 intact, zero core, and the tests actually run now.

PR #2966, opened 2026-07-16
feat: builtin RAG ingest + query pipelines + expose signal `meta` on pipeline ctx (FP-0063 P3)

tya5:
**[per-PR-coder]** — part of #2955 (FP-0063 P3). Design:
`docs/deep-dives/proposals/0063-builtin-turnkey-user-rag.md`.

The two builtin pipelines the P2 MCP servers (#2952) exist to serve.

## ⚠️ NOT turnkey yet — `python3` on PATH must be reyn's own interpreter

**Named blocker, deliberately not fixed here** (needs a DSL/core decision; tracked on #2955).

`rag_ingest` shells out to `python3 -m reyn.builtin.rag_ingest_helpers` for the three things R1
genuinely cannot express (hash / string length / zip-by-index). `sandboxed_exec` forwards the
**ambient PATH** (`os.environ.get("PATH")`) and `sys.executable` is **unreachable from the DSL**,
so `python3` resolves however the operator's shell resolves it. It therefore works **only where
`python3` happens to be reyn's interpreter** — a `pipx install reyn` (the *most* turnkey install),
a non-activated venv, or any differing PATH breaks it.

**CI structurally cannot witness this class** — CI's `python3` *is* the `pip install -e` job's own
interpreter, so it is always green. **"7/7 CI green" is not evidence here**, the same way #2965
stayed invisible for 13 days because dogfood doesn't run in CI.

**What this PR does about it**: since FP-0063's whole subject is *turnkey* user RAG, I removed
every "turnkey"/"ready to use" claim from the pipeline headers, step descriptions, registry
entries and cookbook, and stated the requirement at each surface instead. The failure is also now
**caught at step 0 with a named remedy** rather than surfacing as an opaque
`path 'ctx.files_raw.structured' is absent` (see ③ below). Unmet, written as unmet.

| Pipeline | Registered as | Does |
|---|---|---|
| `rag_ingest.yaml` | `rag_ingest.ingest` | file/folder → markdown → chunk → embed → sqlite vector store, incrementally by `content_hash` |
| `rag_query.yaml` | `rag_query.query` | embed the query → top-k nearest chunks |

## ⚠️ The proposal's **F-B premise is false** — and this PR carries the minimal core change it forces

The proposal states (line 112 / co-vet finding F-B):

> **`embed`'s result envelope already carries usage** … so a **pipeline can read spend directly**

and builds two conclusions on it: **X2a** ("`fold` `envelope.total_tokens`") and
**"F-B also settles C4 for free"** ("stamp chunk metadata from `envelope.model`").

**Both were unreachable.** Verified directly (primary data — the actual outputs, not inference):

```
canonical meta      : {'model': 'text-embedding-3-small', 'total_tokens': 42, 'cost_usd': 0.001, 'priced': True}
pipeline ctx.<name> : {'text': '', 'structured': [[0.1, 0.2]]}
total_tokens reachable from ctx?  False
cost_usd     reachable from ctx?  False
```

`embed_to_canonical` correctly puts the usage into `meta` — its docstring even anticipates "a
future ingest pipeline's `fold` step, X2a/X2c reads [them] inline". But
`canonical_to_ctx_fields`, which builds `ctx.<name>`, read **`text` + `structured` only**. `meta`
— every producer's signal channel, which the *chat* side has always seen via `seam.py` — died at
the pipeline boundary.

**Why the fix is in this PR rather than deferred.** P3's "zero core change" gate was a
*prediction* resting on F-B being true. With F-B falsified, that gate's premise is gone. The two
honest options were "make the minimal core change that makes the arc correct" or "keep the gate
by writing a value we know is wrong" — and the second is not shippable, because the wrong value
is **specifically the failure C1 exists to prevent**:

> a server-internal embedder violates C1 *and* breaks **C4** — the per-chunk `embedding_model`
> column would record a model that never produced those vectors, i.e. **the column becomes a lie**

We reject vector-DB servers with built-in embedders for causing exactly that. Reaching the same
lie through the ctx reducer instead of through the server would have defeated the gate by another
route. (The owner's constraint is *"reyn 側のコード変更は最小限"* — minimal, not zero — and this
arc already has a sanctioned core PR, PC/#2949.)

### The core change

`canonical_to_ctx_fields` now exposes `meta`, **absent-when-empty** on the identical convention
`structured` already uses. Additive: a producer emitting no signal adds no key, so nothing that
previously read `ctx.<name>` changes shape. Absent-when-empty is deliberate — a step depending on
a signal then **fails loudly** instead of silently reading `{}` and defaulting.

### What it bought

| Req | Before (doc-only mitigation) | Now |
|---|---|---|
| **C4** | stamped the pipeline's INPUT — a class alias (`"standard"`), i.e. a model that never produced the vectors. Mitigation was a comment telling users to pass a concrete id. | stamped from **`envelope.model`** — the resolved model. **Correct by construction.** |
| **X2a** | chars/4 **estimate**, carrying a `note` field apologizing for itself | **metered** `tokens_embedded` / `cost_usd` / `priced` off the envelope. The caveat is deleted. |

Observed end-to-end with a real litellm-priceable model — **actual money, not an estimate**:

```
RUN: real priceable model (text-embedding-3-small)
  embedding_model  = 'text-embedding-3-small'
  tokens_embedded  = 9176
  cost_usd         = 0.00018352      ← 9176 x $0.02/1M. Real dollars.
  priced           = True
```

### Falsification (the wiring is *proven* to be load-bearing)

Stripping the two-line `meta` wiring from `canonical_to_ctx_fields` turns the C4 and X2a tests
**RED**, then green on restore:

```
error: pipeline 'rag_ingest.ingest' failed: step 11.match.12.match.1 (transform) failed:
  path 'ctx.embedded.meta.model' is absent: no field 'ctx.embedded.meta' in context

FAILED test_upserted_chunk_embedding_model_is_the_resolved_model_not_the_alias
FAILED test_ingest_reports_metered_spend_not_the_chars4_estimate
5 failed, 3 passed          (restored -> 8 passed)
```

Note the failure *mode*: the absent path makes it **fail loudly**, not stamp a wrong value.

**The fakes are built so these tests cannot pass vacuously**, which is the real risk here:
- the provider **resolves** `"standard"` → `"fake/standard"` (mirroring `test_op_embed.py`'s own
  fixture, and what a real `RoutingEmbeddingProvider` does with a class alias). Had it echoed the
  input back, "stamped from INPUT" and "stamped from `envelope.model`" would be **indistinguishable**
  and the C4 assertion would pass against either implementation.
- it meters `len(text)` while the pipeline's estimate computes `len(text)//4` — a ~4x gap, so the
  reported figure is attributable to exactly one source (`1726` metered vs `429` estimated).

### One figure stays an estimate — necessarily

`estimated_tokens_saved_by_dedup` is a **counterfactual**: the only way to learn a *skipped*
chunk's true token count is to send it to the embedder — i.e. spend precisely what the skip
exists to avoid. Its name says which of the two it is; everything actually embedded is metered.

## What's here

- **`src/reyn/builtin/pipelines/rag_ingest.yaml`** — 6 `pipeline:` docs (entry + 5 same-file siblings).
- **`src/reyn/builtin/pipelines/rag_query.yaml`** — 3 `pipeline:` docs.
- **`src/reyn/builtin/rag_ingest_helpers.py`** — 4 stdin/stdout JSON subcommands (`probe` / `list_files` / `hash_chunks` / `zip_vectors`) for the three things R1 genuinely cannot express: **hashing**, **string length**, and **zip-by-index**. Invoked via `shell:` steps. Builtin content, no reyn-core import — copyable alongside the pipelines (R2).
- **`src/reyn/builtin/registry.py`** — registers both (inert: invoke-by-name, A3).
- **`src/reyn/core/offload/canonical.py`** — ⚠️ **the one core change**: expose `meta` on the pipeline ctx shape (see above).
- **`docs/reference/runtime/pipeline-dsl.md` + `.ja.md`** — the normative ctx-shape reference called this a "two-field shape"; now documents `meta`, the absent-when-empty rule, and the `get(...)` safe-navigation idiom. Both language mirrors.
- **`docs/cookbook/configs/with-builtin-rag-mcp.yaml`** — adds the markitdown server entry P2 deferred to P3, + how to run both pipelines, + the "copy and re-point" extension note. Still ships **commented-out / INERT** (R3 / F-D: an `mcp.servers.<name>` entry is auto-granted by `reyn pipe run`).

### X1 pre-flight (the decision-enabling error)

Step 0 makes a **real probe call** per server (not a config-presence check), each
`schema: PreflightCheck`-gated inside a `parallel { on_error: continue }` — so a failure
drops *that one branch* and we know **which** server failed **by name**, without parsing
transport error text. An unreachable server yields, per the #2932 `require_mcp` precedent
(**cause + two concrete remedies**), e.g.:

> Vector-store MCP server 'reyn_vector_store' is unreachable. **(1)** enable it: add/uncomment `mcp.servers.reyn_vector_store` in reyn.yaml and set `permissions.mcp.reyn_vector_store: allow` — see `docs/cookbook/configs/with-builtin-rag-mcp.yaml`. **(2)** if already enabled, install its dependencies: `pip install "reyn[builtin-rag]"` (apsw + sqlite-vec).

The markitdown remedy additionally names the **R1 firewall** case explicitly (`uvx` fetches
from PyPI on first run → pre-install instead) — the owner's own network is the acceptance test.
A bare `ImportError: No module named 'apsw'` never reaches the operator: the block is the
run's own successful *output*, not an exception.

### C5 / X5 — observed, not assumed

Ingest runs, printing the real summary (2 files, ~429 chars/4 tokens):

```
RUN 1 (cold -- everything new)
  chunks_upserted                  = 5
  chunks_unchanged_skipped         = 0
  estimated_tokens_embedded        = 429
  estimated_tokens_saved_by_dedup  = 0
RUN 2 (X5: re-ingest UNCHANGED folder)
  chunks_upserted                  = 0      ← costs ~0 …
  chunks_unchanged_skipped         = 5
  estimated_tokens_embedded        = 0
  estimated_tokens_saved_by_dedup  = 429    ← … and REPORTS that it did
```

X5's stated purpose ("re-ingesting an unchanged folder must cost ≈0 **and report that it
did**") is met. Note `embed` is not called at all on a fully-deduped run — it *rejects* an
empty `texts` array, so a `match` gate routes around it rather than feeding it something it
refuses.

## Test plan

`tests/test_fp0063_p3_rag_pipelines.py` — 7 tests, driving the **real** `reyn pipe run`
path against the **real** P2 MCP servers as real `python -m` stdio subprocesses.

- [x] **Tests actually RUN — not skipped.** `pytest tests/test_fp0063_p3_rag_pipelines.py -q -rs` → **`8 passed`**, zero skips. **CI-side proof**: main baseline `8559 passed / 16 skipped` → this PR `8566 passed / 16 skipped` = **+7 passed, +0 skipped** on the first push. The P2 near-miss (module-wide skip under CI green) did not recur. CI installs the extra (`test.yml:39` → `pip install -e ".[dev,mcp,web,fs-watch,observability,builtin-rag]"`, added by P2/#2952), so this holds in CI too — but the local `-rs` run above is the evidence, not CI green.
- [x] **`ruff check src tests`** → `All checks passed!`
- [x] **`python scripts/test_tier_audit.py --strict <both changed test files>`** → `OK: 30, warnings: 0, errors: 0` (exit 0)
- [x] **`pytest` (full, repo root)** → **`8577 passed, 20 skipped, 0 failed`** (223s), after all review rounds.
- [x] **`python scripts/verify_module_docstrings.py <changed src files>`** → OK
- [x] **C5 add → dedup → update → remove converges** (`test_ingest_add_then_dedup_then_update_then_remove`), driven through real chunker + vector-store servers.
- [x] **C4 = the RESOLVED model, read back out of the real sqlite** (`test_upserted_chunk_embedding_model_is_the_resolved_model_not_the_alias`): invoked with the alias `"standard"`, every stored row carries `"fake/standard"`. Non-vacuous by construction (see the fakes note above).
- [x] **X2a = metered, not estimated** (`test_ingest_reports_metered_spend_not_the_chars4_estimate`): asserts the reported figure is ~4x the chars/4 estimate of the same text, plus `priced is False` / `cost_usd is None` for an unpriceable model (visible degradation, never a silent $0.00 — the #1829 sentinel).
- [x] **`meta` reducer pinned directly** (Tier 1, in `test_2425_pipeline_tool_ctx_canonical_shape.py`): present-when-signal / absent-when-empty.
- [x] **Falsification observed**: stripped the `meta` wiring → C4 + X2a RED; restored → green (output above).
- [x] **X1 block + falsifying control** — unconfigured server → named remedy; the *same* setup with the real name proceeds into the ingest body, so the block is attributable to the unreachable server, not a broken gate.
- [x] **X1 catches a reyn-less `python3`** (`test_ingest_preflight_blocks_when_python3_cannot_import_reyn`) — drives the REAL failure by pointing PATH at a python3 shim without reyn (which is precisely what `sandboxed_exec` forwards), not by faking a seam. Asserts cause + both remedies are present.
- [x] **rag_query top-k** returns the ingested chunk (`test_rag_query_returns_the_ingested_chunk_as_top_result`).
- [x] **X2a/X5 purpose observed end-to-end** — re-ingesting an unchanged folder upserts **0** and *reports* the 429 tokens the dedup saved; a priceable model reports **`$0.00018352`** of real spend. Driven manually, printed from the pipeline's own output.

### Test honesty notes

- ⚠️ **markitdown is SUBSTITUTED — the owner's named formats are UNVERIFIED.** `markitdown-mcp`
  is not installed here and `uvx` would need network, so the tests run a **small real FastMCP
  stdio server** (written to disk by the test) that reads `file://`/`data:` URIs as UTF-8. The
  chunker and vector-store servers are the **real** builtin ones, and only `.txt` fixtures are
  ingested — so **this PR has NOT exercised real pdf / xlsx / pptx / docx through real
  markitdown**, i.e. exactly the format list the owner named (txt, md, pdf, エクセル, パワポ,
  ワード). Not claimed as working. Belongs to P4/P5 as real-environment verification; flagged
  for #2955 rather than papered over.
- **The embedding provider is faked** (`FakeEmbeddingProvider`, mirroring
  `tests/test_op_embed.py`'s own fixture, monkeypatched onto `op_runtime.embed.get_provider`)
  — no `unittest.mock`, no API key/network. Everything else is real.

## Review round 2 — three in-PR findings

**① The C4 fossil (my bug).** `_ingest_body`'s description still described the design that was
*rejected* — "stamping embedding_model from **THIS pipeline's own input** … see ingest's
description for **why not envelope.model**" — while the implementation does the opposite and
`ingest`'s own description was already correct. An **inverted cross-reference**: readers were sent
to a doc that now explains the exact opposite. Residue of the very class this PR pins, sitting
inside the PR whose thesis is *correct by construction, not by documentation*. Fixed.

**② The marker comment (my bug, introduced by this PR).** `canonical.py`'s
`_DISCRIMINATOR_MISS_MARKER` justified its containment with *"the ctx reducer
(`canonical_to_ctx_fields` **reads `text`/`attachments`**) never read it"* — which **this PR made
false** by adding `meta`. The **conclusion** survives (the marker is a *top-level* key, not a
`meta` field, so no consumer names it) but the **stated mechanism** was wrong, and it collided on
the term *"signal channel"* with `meta`'s new docstring. Restated: containment rests on being
top-level, and the two channels are now distinguished by name.

**③ The silent shell-step failure — the proposed mechanism does not exist.**

The review asked the pipeline to read its own `shell` steps' `meta.returncode`. **It cannot**:
`shell._handle` returns **`stdout` only** — `returncode`/`stderr` are dropped *one layer above*
the canonical seam by #2593's locked design, so `shell_to_canonical`'s meta is **always `{}`**
(its own docstring says so). Verified by driving a failing shell step end-to-end:

```
shell step result in ctx : {'text': ''}      ← no meta, no returncode
did the pipeline abort?  : NO -- it completed
```

There is no `meta.returncode` to read. But the **goal** — make the silent failure diagnosable —
is reachable, and that is what shipped: a `probe` subcommand wired into the **existing X1
pre-flight list** (it is a reachability check, so it belongs beside the MCP checks rather than as
new scaffolding). Empty stdout ⇒ no `structured` key ⇒ detected, **before any embedding spend**.

Observed — the 30-minute opaque failure is now a named remedy:

```
rag ingest blocked -- one or more prerequisites are unreachable:

helpers (python3 -m reyn.builtin.rag_ingest_helpers): `python3 -m reyn.builtin.rag_ingest_helpers`
failed -- the `python3` on PATH cannot import reyn. […] (1) run reyn from an environment whose
`python3` IS reyn's interpreter -- typically `source .venv/bin/activate` […] a `pipx install reyn`
or a non-activated venv will NOT satisfy this. (2) check what you actually have:
`python3 -c 'import reyn; print(reyn.__file__)'` […]
```

Honest limits: the probe catches **"`python3` has no reyn"** (the common pipx/venv/PATH case). It
**cannot** catch **"`python3` has a *different* reyn"** (the architect's two-install incident) —
that imports fine. So it *reports the paths* for a human to compare, and says so in the message
rather than implying a guarantee. `rag_query` shells out to nothing, so it carries no probe and no
`python3` requirement.

## Other

- **`tests/test_2575_pipeline_disk_registration.py` touched (unavoidable).** It pinned
  `set(pipeline_registry.names()) == {"hello.hello", "flagship.research_and_report"}` — a
  hardcoded set that any new builtin pipeline breaks. Rewritten to **derive** expected names
  from `BUILTIN_PIPELINES` itself, so the next builtin needs no edit here. The test still
  asserts the same invariant (builtin tier is additive, not exclusive).
- **The proposal doc is untracked in `main`** — it exists only in the lead-coder's working
  tree, so the path above will 404 for reviewers until it's committed. Not committing
  someone else's uncommitted draft from a sub-PR; flagging for the lead-coder.

### DIFF
--- a/docs/cookbook/configs/with-builtin-rag-mcp.yaml
+++ b/docs/cookbook/configs/with-builtin-rag-mcp.yaml
@@ -1,4 +1,6 @@
-# FP-0063 P2 -- builtin turnkey user RAG, the MCP layer.
+# FP-0063 -- builtin turnkey user RAG, the MCP layer (P2) + the two
+# builtin pipelines that call it (P3: reyn.builtin.pipelines.rag_ingest /
+# rag_query, registered as rag_ingest.ingest / rag_query.query).
 #
 # This entire ``mcp.servers`` block ships COMMENTED OUT / INERT. Do not
 # assume uncommenting it is safe by default: an MCP server entry, once it
@@ -10,14 +12,16 @@
 # ("the operator explicitly configured this server") only holds if YOU are
 # the one who copies this block into your own reyn.yaml and uncomments it
 # -- reyn itself never wires this in (see reyn.builtin.mcp_servers module
-# docstring). Read the two servers before enabling them: the vector-store
+# docstring). Read the three servers before enabling them: the vector-store
 # server writes to whatever sqlite file `db_path` names; the chunker server
-# reads no filesystem paths itself but the ingest pipeline that will call
-# it (a later phase, P3) reads document files from the folder you point it
-# at.
+# reads no filesystem paths itself; markitdown reads whatever `uri` the
+# ingest pipeline passes it (every file under the folder you point rag_ingest
+# at).
 #
 # Install the extra runtime deps first (NOT part of reyn's base install):
 #   pip install "reyn[builtin-rag]"
+#   uvx markitdown-mcp   # or: pip install markitdown-mcp (pre-install --
+#                         # see the note on R1/uvx-firewall below)
 #
 # Then copy this block into your reyn.yaml (or reyn.local.yaml) and
 # uncomment it:
@@ -25,6 +29,7 @@
 # permissions:
 #   mcp.reyn_vector_store: allow
 #   mcp.reyn_chunker: allow
+#   mcp.reyn_markitdown: allow
 #
 # mcp:
 #   servers:
@@ -45,6 +50,34 @@
 #       # min_characters_per_chunk=24). size/overlap are real parameters,
 #       # not baked-in constants -- tune them per document type.
 #
-# A third-party parser server (e.g. `uvx markitdown-mcp`, Microsoft,
-# converts txt/md/pdf/xlsx/pptx/docx -> Markdown) belongs alongside these
-# two once the ingest pipeline (P3) lands; out of scope for this file.
+#     reyn_markitdown:
+#       type: stdio
+#       command: uvx
+#       args: ["markitdown-mcp"]
+#       # Tool: convert_to_markdown(uri) -- file:/http:/https:/data: -- the
+#       # owner's target format list (txt/md/pdf/xlsx/pptx/docx) plus 15+
+#       # more. `uvx` fetches the package from PyPI on first run: if your
+#       # network is firewalled (FP-0057's own line-55 finding, the same
+#       # failure class as the default embedding model's HuggingFace fetch),
+#       # pre-install instead of relying on the first-run fetch:
+#       #   pip install markitdown-mcp
+#       # and point `command`/`args` at that installed console script
+#       # instead of `uvx`. rag_ingest's own step-0 pre-flight (X1) checks
+#       # reachability BEFORE any embedding spend and fails with this same
+#       # guidance, naming the exact server, rather than a bare transport
+#       # exception.
+#
+# Once the three servers above are configured + granted, run the builtin
+# pipelines directly (no separate skill needed -- invoke-by-name):
+#
+#   run_pipeline(name="rag_ingest.ingest", input={
+#     "input_path": "docs/", "output_db": ".reyn/rag/docs.sqlite"})
+#   run_pipeline(name="rag_query.query", input={
+#     "query_text": "how does X work?", "db": ".reyn/rag/docs.sqlite"})
+#
+# Want a different vector-DB / chunker / parser? Copy
+# src/reyn/builtin/pipelines/rag_ingest.yaml (+ rag_query.yaml) into your
+# own project and re-point the `*_server` inputs (or the MCP config entries
+# they name) at your replacement -- FP-0057 C2: reyn builds no adapter for
+# a user's RAG store, so "copy the builtin and re-point the MCP server" IS
+# the extension mechanism, not a workaround.
--- a/docs/cookbook/configs/with-builtin-rag-mcp.yaml
+++ b/docs/cookbook/configs/with-builtin-rag-mcp.yaml
@@ -1,4 +1,4 @@
-# FP-0063 -- builtin turnkey user RAG, the MCP layer (P2) + the two
+# FP-0063 -- builtin user RAG, the MCP layer (P2) + the two
 # builtin pipelines that call it (P3: reyn.builtin.pipelines.rag_ingest /
 # rag_query, registered as rag_ingest.ingest / rag_query.query).
 #
@@ -66,6 +66,15 @@
 #       # reachability BEFORE any embedding spend and fails with this same
 #       # guidance, naming the exact server, rather than a bare transport
 #       # exception.
+#
+# ⚠️ rag_ingest ALSO requires that `python3` on your PATH be reyn's own
+# interpreter -- it shells out to `python3 -m reyn.builtin.rag_ingest_helpers`
+# and sandboxed_exec forwards the ambient PATH. A `pipx install reyn` or a
+# non-activated venv will NOT work. Check with:
+#   python3 -c 'import reyn; print(reyn.__file__)'
+# It must print the same install you run reyn from. rag_ingest's step-0
+# pre-flight reports this rather than failing opaquely, but it is a real
+# limitation (tracked on the FP-0063 umbrella), not a formality.
 #
 # Once the three servers above are configured + granted, run the builtin
 # pipelines directly (no separate skill needed -- invoke-by-name):
--- a/docs/reference/runtime/pipeline-dsl.ja.md
+++ b/docs/reference/runtime/pipeline-dsl.ja.md
@@ -218,19 +218,44 @@ FieldType     ::= "{" "type:" ("bool" | "string" | "number") "}"
 #### `tool` ステップの結果 {#tool-step-results}
 
 `tool`(および `shell`)ステップの結果 — `pipe` と `ctx.<output>` に入るもの —
-は常にフラットな 2 フィールド形状 `{text: ..., structured: ...}` で、
+は常にフラットな形状 `{text: ..., structured: ..., meta: ...}` で、
 すべてのツール種別で統一されています(chat 側が LLM に見せるのと同じ形状):
 
 - **`text`** — ツールの正規の文字列本体(ツールにテキスト形状の結果が無い場合は空文字列)。
 - **`structured`** — ツールが非テキストのデータを出したときだけ存在する。それ以外は
   キー自体が無い。認識されない形状のツール結果 dict は、丸ごと `structured` として
   ラップされる(何も失われない)。
+- **`meta`** — ツールの**シグナル**フィールド。「呼び出し側が次に何をするかを変える」ため、
+  producer が本体からあえて外した、小さく情報量の多い値。ツールが何かシグナルを出した
+  ときだけ存在し、それ以外はキー自体が無い — 成功結果の多くはシグナルを持たないので、
+  無いほうが通常です。
 
 この変換は**形状のみ**です — chat 側のツール結果パスのように値を切り詰めたり
 offload したり cap したりは絶対にしません。pipeline ステップの `ctx` は、後続の
 ステップがプログラム的に消費できるよう、値をフルで保持します。ツールのテキスト本体は
 `ctx.<name>.text` で、structured なペイロードは `ctx.<name>.structured`(またはその
 フィールド、例: `ctx.hits.structured.count`)で読みます。
+
+**`meta` を安全に読む。** `meta` は producer がシグナルを出さないと存在しないため、
+そうした producer に対して素の `ctx.<name>.meta.<field>` パスは**例外を投げます**
+(素のパスは safe navigation ではありません — [R1 式言語](#the-r1-expression-language)
+参照)。これは意図的です: シグナルに依存するステップは、黙って既定値を読むのではなく、
+大きな音を立てて失敗すべきだからです。シグナルが本当に任意の場合は `get(...)` を使います:
+
+```yaml
+- tool: {name: embed, args: {texts: !expr ctx.chunks}, output: embedded}
+# 実際にベクトルを生成したモデル(`embedding_model` 入力は "standard" のような
+# モデル**クラス別名**でありうる。`meta.model` はそれが解決された結果):
+- transform: {value: "ctx.embedded.meta.model", output: resolved_model}
+# 任意のシグナル — raise させずに既定値を使う:
+- transform: {value: "get(ctx.embedded, 'meta.cost_usd', null)", output: spend}
+```
+
+あるツールが `meta` に何を入れるかは、そのツール自身の契約です。代表例:
+`embed` → `model` / `total_tokens` / `cost_usd` / `priced`、`sandboxed_exec` →
+非ゼロの `returncode`(ゼロ終了はシグナルではないのでキー無し)、MCP 呼び出し →
+`isError`。結果そのものである `structured` に対し、`meta` は「その結果が**いくら
+かかったか**」「**どう**生成されたか」が必要なときに使います。
 
 `shell` は `"shell"` という名前の `tool` ステップのシュガーです: `command` は operator のサンドボックス(`sandboxed_exec` — 直接の `sandboxed_exec` 呼び出しと同じ閉じ込め、ポリシー優先順位、監査イベント)の中で `/bin/sh -c <command>` として実行されます。ステップに入ってくる pipe data はプロセスの **STDIN に JSON エンコードされて**渡されます。プロセスの **STDOUT がステップの結果になります** — JSON として parse できればデコードされ(`dict` を要求する `verify: schema` が JSON を出力するコマンドに適用できるように)、できなければ生のテキストのままです — そして、その(生または JSON デコード済みの)値は他の `tool` ステップと同じ `text`/`structured` 変換を経ます(JSON デコードされた dict は `ctx.<name>.structured` に、生のテキストは `ctx.<name>.text` に入ります)。
 
--- a/docs/reference/runtime/pipeline-dsl.md
+++ b/docs/reference/runtime/pipeline-dsl.md
@@ -297,7 +297,7 @@ registered tool name (`web_search`).
 #### `tool` step results
 
 A `tool` (and `shell`) step's result — what lands as `pipe` and `ctx.<output>`
-— is always the flat two-field shape `{text: ..., structured: ...}`, uniform
+— is always the flat shape `{text: ..., structured: ..., meta: ...}`, uniform
 across every tool kind (the same shape the chat side exposes to an LLM):
 
 - **`text`** — the tool's canonical string body (empty string when the tool
@@ -305,12 +305,39 @@ across every tool kind (the same shape the chat side exposes to an LLM):
 - **`structured`** — present only when the tool produced non-text data; absent
   entirely (no key) otherwise. A tool result dict with no recognized shape is
   wrapped whole as `structured` (nothing is ever lost).
+- **`meta`** — the tool's **signal** fields: small, high-signal values the
+  producer deliberately kept out of the body because they change what the
+  caller does next. Present only when the tool emitted any; absent entirely
+  (no key) otherwise — which is the common case, since most successful
+  results carry no signal.
 
 This conversion is **shape-only** — it never truncates, offloads, or caps a
 value the way the chat-side tool-result path does; a pipeline step's `ctx`
 retains the full value for downstream steps to consume programmatically.
 Read a tool's text body via `ctx.<name>.text` and its structured payload via
 `ctx.<name>.structured` (or a field of it, e.g. `ctx.hits.structured.count`).
+
+**Reading `meta` safely.** Because `meta` is absent when a producer emits no
+signal, a bare `ctx.<name>.meta.<field>` path *raises* on such a producer
+(bare paths are not safe navigation — see
+[The R1 expression language](#the-r1-expression-language)). That is
+deliberate: a step that depends on a signal fails loudly rather than silently
+reading a default. Use `get(...)` when the signal is genuinely optional:
+
+```yaml
+- tool: {name: embed, args: {texts: !expr ctx.chunks}, output: embedded}
+# the model that ACTUALLY produced the vectors (an `embedding_model` input may
+# be a model-CLASS alias like "standard"; meta.model is what it resolved to):
+- transform: {value: "ctx.embedded.meta.model", output: resolved_model}
+# optional signals — default rather than raise:
+- transform: {value: "get(ctx.embedded, 'meta.cost_usd', null)", output: spend}
+```
+
+What a given tool puts in `meta` is that tool's own contract; common examples:
+`embed` → `model` / `total_tokens` / `cost_usd` / `priced`; `sandboxed_exec` →
+a nonzero `returncode` (a zero exit is not signal, so no key); an MCP call →
+`isError`. Reach for `meta` when you need what a result *cost* or *how* it was
+produced — as opposed to `structured`, which is the result itself.
 
 `shell` is sugar for a `tool` step named `"shell"`: `command` runs as
 `/bin/sh -c <command>` inside the operator's sandbox
--- a/src/reyn/builtin/pipelines/rag_ingest.yaml
+++ b/src/reyn/builtin/pipelines/rag_ingest.yaml
@@ -1,4 +1,4 @@
-# Builtin turnkey RAG -- ingest pipeline (FP-0063 P3).
+# Builtin RAG -- ingest pipeline (FP-0063 P3).
 # docs/deep-dives/proposals/0063-builtin-turnkey-user-rag.md
 #
 # Reads a file or a folder, converts each document to Markdown (the
@@ -12,6 +12,17 @@
 # `*_server` inputs (or the MCP config entries they name) at a different
 # backend. Every tunable is a `ctx.*` INPUT with a sane default, never a
 # constant baked into a step (R4).
+#
+# ⚠️ ENVIRONMENT REQUIREMENT -- this is NOT yet turnkey. This pipeline
+# shells out to `python3` (for hashing / string length / zip-by-index, the
+# three things the R1 expression language cannot express), and
+# `sandboxed_exec` forwards the ambient PATH, so `python3` MUST be the same
+# interpreter reyn itself runs under. A `pipx install reyn`, a
+# non-activated venv, or any PATH whose `python3` differs will FAIL. The
+# DSL cannot reach `sys.executable` today; closing this needs a core/DSL
+# decision, tracked on the FP-0063 umbrella. Step 0 pre-flights it so the
+# failure is named rather than opaque -- but it is a real limitation, not
+# a formality.
 #
 # Inputs (all optional except `input_path`/`output_db`):
 #   input_path          -- (required) a document file OR a folder of them.
@@ -49,11 +60,12 @@ fields:
 ---
 pipeline: ingest
 description: >-
-  FP-0063 P3 builtin turnkey RAG ingest: chunk -> embed -> store a file or
-  folder into a user-named sqlite vector store, incrementally by
-  content_hash (add/update/remove). Pre-flights the 3 required MCP servers
-  before any embedding spend (X1); reports spend AFTER the dedup skip
-  (X2a/X5).
+  FP-0063 P3 builtin RAG ingest: chunk -> embed -> store a file or folder
+  into a user-named sqlite vector store, incrementally by content_hash
+  (add/update/remove). Pre-flights its 3 MCP servers AND its python3
+  helper before any embedding spend (X1); reports spend AFTER the dedup
+  skip (X2a/X5). REQUIRES that `python3` on PATH be reyn's own
+  interpreter -- see the file header.
 steps:
   # ---- inputs-with-defaults (R4: never a constant baked into a step) ----
   - transform: {value: "get(ctx, 'chunk_size', 400)", output: chunk_size}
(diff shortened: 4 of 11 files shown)


tya5:
**[per-PR-coder]** — **Revision pushed (`d14c0ca9`): the C4 doc-only mitigation is gone; `meta` exposure landed in this PR per the lead-coder decision.** Re-requesting review — **this PR now carries a core change** (`canonical_to_ctx_fields`), so it needs architect co-vet.

**What changed since first review:**

| | before | now |
|---|---|---|
| **C4** | stamped the pipeline INPUT (class alias `"standard"`) — a model that never produced the vectors. Mitigated by a **comment**. | stamped from **`envelope.model`** (resolved). **Correct by construction.** |
| **X2a** | chars/4 **estimate** + an apologetic `note` field | **metered** `tokens_embedded`/`cost_usd`/`priced`. Caveat deleted. |
| core change | none (gate held by writing a wrong value) | **`canonical_to_ctx_fields` exposes `meta`** (absent-when-empty, additive) |

**The framing, for the record:** P3's "zero-core" gate was a *prediction* resting on F-B being true. Primary data falsified F-B (`meta` dies at the ctx reducer; the chat side sees it via `seam.py`, the pipeline never did). With the premise gone, the gate did not survive it — so this is "the prediction was wrong, here is the minimal core change that makes it correct", **not** "the gate was kept by lying in a column".

**Evidence the wiring is load-bearing** (not a mechanism test that passes while production does nothing):
- **Falsified**: stripping the 2-line `meta` wiring → C4 + X2a go **RED** (`path 'ctx.embedded.meta.model' is absent` — it fails *loudly*, it does not stamp a wrong value); restore → green.
- **Non-vacuous fakes**: the provider **resolves** `"standard"` → `"fake/standard"`, so "stamped from INPUT" vs "stamped from `envelope.model`" are distinguishable — had it echoed the input, the C4 assertion would pass against either implementation. It meters `len(text)` vs the estimate's `len(text)//4` (~4x gap), so the X2a figure is attributable to one source.
- **Real money observed**: `text-embedding-3-small`, 9176 tokens → `cost_usd = 0.00018352` (= 9176 × $0.02/1M), `priced = True`.

**Reviewer attention, please:**
1. **`canonical_to_ctx_fields` is a shared seam** — every `tool:` step in every pipeline reads it. I chose **absent-when-empty** (mirroring `structured`) so it is additive and a signal-dependent step fails loudly rather than defaulting off `{}`. Full suite **8576 passed / 0 failed** confirms no consumer broke, but the shape choice is worth a second opinion.
2. **`docs/reference/runtime/pipeline-dsl.md` (+ `.ja.md`)** called this a "two-field shape" — updated both mirrors, incl. the `get(...)` safe-navigation idiom.
3. **Pre-existing, NOT mine**: 3 tests in `test_2425_pipeline_tool_ctx_canonical_shape.py` fail **in isolation** (order-dependent — they need another test to import `op_runtime.mcp` so its canonical mapper registers). Verified identical with my change stashed. Untouched deliberately; happy to file separately.

**Still open, not claimed** (for #2955 / P4-P5): **real markitdown with real pdf/xlsx/pptx/docx is unverified** — the tests substitute a small real FastMCP server and ingest only `.txt`. The owner's named format list is not evidenced by this PR.

tya5:
**[architect]** — co-vet #2966。**判定: BLOCK なし。in-PR-required 2件(いずれも1〜2行) + arc-blocker 級の設計論点 1件 + follow-up 1件。**

**計測系の申告を先に**: この branch は **stale base** なので、`origin/main` を merge した木に **worktree 専用の venv を作り、`pip install -e ".[dev,mcp,web,fs-watch,observability,builtin-rag]"` 相当を入れて** 測りました(`reyn.__file__` が worktree を指すことを確認済み)。**4 gate 全緑**: ruff ✓ / tier-audit --strict(3件)EXIT=0 / docstrings(3件)EXIT=0 / **pytest RC=0、8574 passed、17 skipped**。**CI 7/7 緑の報告は追試できました。**

**strip-falsify も独立に実施**(`cp` backup、`git stash` は不使用): `canonical_to_ctx_fields` の meta 露出2行を strip → **C4 + X2a が RED**、復元は **byte-exact**(`git diff` 空)→ green。**load-bearing を確認**しました。

**`test_2425` の order-dependence も独立に追試**: **main 単体・正しい venv で 3 failed / 4 passed** が再現 ∴ **既存であり本 PR 起因ではない**、coder が触らなかったのは正しい判断です。

---

## 観点1 — `meta` 露出の blast radius: **CLEAR(構造的に安全)**

**「absent-when-empty は本当に additive か」は、構造で答えられます**: `core/pipeline/schema.py` の `ValidationError` 語彙は `missing_required` / `type_mismatch` / `enum_invalid` / `unresolved_ref` / `unknown_type` のみで、**`additionalProperties` 相当の extra-field 拒否が存在しません** ∴ **ctx に key が増えても既存 pipeline の検証を落とすことは構造的に不可能**です(`unknown_type` は schema 側の型宣言に対するもので、データの余分な field ではありません)。

`canonical_to_ctx_fields` の言及を**全数列挙**(18件、切り捨てなし)→ 実装2 + 呼び出し1 + テスト14 + **記述1**。docs は en/ja 両方更新済みで、残存する "two-field" 記述は**無関係な audit doc の3件のみ**。**coder が normative な記述に自分で気づいて en/ja 両方直し、reducer 自体に Tier-1 pin を足した**のは正しい仕事です。

## 観点2 — 「loud に落ちる」設計: **同意。しかも embed では *成立が保証されている***

`embed` の meta は `{k: result[k] for k in ("model","total_tokens","cost_usd","priced") if k in result}` で、**#2949 が unpriced でも `priced=False` / `cost_usd=None` を必ず載せる** ∴ **`ctx.embedded.meta` は常に存在**します。∴ 「absent = 何かが壊れている」が真になり、**C4 の loud fail は正しいガード**です。

**pipeline 内の非対称も正しい**: `:339` は `ctx.embedded.meta.model` を**直接アクセス**(= 誤った model を静かに stamp するくらいなら落ちる)、`:363-364` は `get(..., default)`(= コスト報告は劣化して良い)。**#1829 sentinel が保全されています**: `cost_usd` は as-is、`priced` の default が **false**(不明なら未課金扱い = fail-visible)、空バッチだけ `priced: true`(**$0 が既知**であって不明ではない)。

## 観点3 — `meta` の中身が漏れて困るか: **CLEAR(chat 経路との差分ゼロ)**

mapper 契約上、`meta` は **producer 自身が書く小さな signal** です。決定的な点2つ:
- **`mcp` mapper の meta は `isError` のみ**に絞られています(`canonical.py:299` *"``meta`` is tightened to signal-only"* / `:193` *"the sole error-path driver kept after meta-tightening"*)∴ **MCP サーバ由来の外部コンテンツは meta に載りません**。私はここを最有力の懸念として調べましたが、**meta-tightening が既に閉じていました**。
- **`seam.py:77-79` は meta を as-is で chat frontmatter に流しています**(size gate なし、`isError` のみ除外)∴ **LLM は既に全部見ています**。pipeline に新規なのは `isError` だけで、機微ではありません。`mcp_install_verb` の `env_keys_set` / `missing_secret_keys` は**鍵の名前**であり、かつ既に chat 可視 ∴ **parity**。

---

## ★ in-PR-required 1 — `_ingest_body` の description が、**却下された設計の化石**です

`rag_ingest.yaml:197-200`(`_ingest_body`):
> "embed only the new/changed chunks, **stamping embedding_model from THIS pipeline's own input (C4 -- see ingest's description for why not envelope.model)**"

**実装(`_ingest_embed_and_upsert`, `:332-339`)は正反対です**:
> "C4 -- stamp the RESOLVED model (`embed`'s own envelope meta), **NOT this pipeline's `embedding_model` INPUT**" / `- transform: {value: "ctx.embedded.meta.model", output: resolved_model}`

**そして `ingest` 自身の description(`:24-31`)は正しく更新済み**です(*"what gets STAMPED ... is the model the provider actually RESOLVED, read back from `embed`'s own envelope meta"*)。

∴ **cross-reference が反転しています** — `_ingest_body` は読者に *「why **not** envelope.model は ingest の説明を見よ」* と案内しますが、**ingest の説明は今や「なぜ envelope.model **なのか**」を説明しています**。**この YAML は本 PR の新規ファイル** ∴ 矛盾はここで書かれたもので、**あなたが却下した「入力の別名から stamp」案の化石**です。

**これは今夜 pin した class そのもの**が、**「documentation ではなく construction で正しくする」ことを主題とする当の PR の中**に residue として残った形です。`_ingest_body` は embed fan-out 連鎖を実際に含む pipeline なので、将来の保守者がここを読んで「入力別名 stamp は意図的だ」と誤解し得ます。**2行の修正**。

## ★ in-PR-required 2 — `canonical.py:1475` の括弧書きを、本 PR が偽にしました

> `# ... the ctx reducer (``canonical_to_ctx_fields`` **reads ``text``/``attachments``**) never read it, so it never reaches the LLM body (unlike ``meta``, which renders as frontmatter YAML).`

**結論は今も真**です — marker は **canonical の top level** に刻まれる(`:1494` `canonical[_DISCRIMINATOR_MISS_MARKER] = True`)ので、reducer の `canonical.get("meta")` は拾いません(私はこれを漏洩仮説として調べ、**falsify しました**)。**しかし述べられている *機構* は now false** です(reducer は text/attachments/**meta** を読みます)。

**なぜ効くか**: marker のコメントは自らを *"a **signal channel** for `canonical_fallback_reason` only"* と呼び、**本 PR の新 docstring は `meta` を "the producer's **SIGNAL channel**" と命名**しました。**同じ語で呼ばれる2つのチャネル**があり、**private な方の安全性を主張するコメントが、本 PR が変えた機構を根拠に挙げています**。将来「marker も signal だから meta へ」と動かした人が、このコメントを読んで安全だと信じます。**1行の修正**。

## ★ 設計論点(arc-blocker 級 — あなたの判断を仰ぎます)— builtin pipeline が ambient `python3` に束縛されています

`rag_ingest.yaml` の3箇所が `shell: "python3 -m reyn.builtin.rag_ingest_helpers ..."` です。実測:

- **`sandboxed_exec.py:88` は `os.environ.get("PATH")` をそのまま渡します** ∴ コマンドは **ambient PATH** に束縛されます。reyn は自分の interpreter を注入しません。
- **`sys.executable` は pipeline DSL から到達不能**です(`src/reyn/core/pipeline/` + `src/reyn/builtin/` で grep 0)。
- **`rag_ingest.yaml` は python に shell out する唯一の builtin pipeline**(前例なし。本 PR が先例を作ります)。

∴ **`python3` が「reyn がインストールされた interpreter」である場合にのみ動きます**。`pipx install reyn`(最も turnkey な導入)/ activate していない venv / PATH の違う環境では、**ユーザは `path 'ctx.files_raw.structured' is absent` という、python にも module にも shell step にも言及しない不透明なエラー**を見ます。

**私はこれを事故で実証しました**: 1台に reyn が2つ入っていて、pipeline が黙って**別の install** を使い、私は30分溶かしました。**PATH を worktree の venv に向けた瞬間、5件が 8 passed になりました。**

**★ 決定的な点: CI はこのクラスを構造的に witness できません。** CI の `python3` は `pip install -e` した job の python **そのもの**なので、**常に通ります**。∴ **「CI 7/7 緑」はここでは証拠になりません** — **dogfood が CI で走らないために #2965 が13日間見えなかったのと同じ性質**です。

**FP-0063 の主題は "turnkey user RAG"** であり、**「機構は存在するが、その目的のために到達可能か」**が問われる場面だと考えます。修正には**設計判断**が要ります(DSL に自分の interpreter を露出する / `python:` step / `sandboxed_exec` が `python3` を `sys.executable` に解決する)。**私の宿題**として引き取ります — **in-PR で塞ぐか、#2955 の P4/P5 に *名前付きの blocker* として載せるか**をご判断ください。**黙って merge されるべきではない**、というのが私の意見です。

## follow-up — shell step の失敗が無言です(既存 + 本 PR に皮肉な形で関係)

- **executor は returncode を一切見ません**(`core/pipeline/executor.py` に `returncode` grep 0)∴ **非ゼロ終了の `shell:` step は pipeline を落としません**。
- **`rag_ingest.yaml` も自分の shell step の `meta.returncode` を見ていません**(`:81`/`:213` の `on_error: continue` は `for_each` の方針で別物)。

∴ 失敗した shell step は無言で通過し、pipeline は**後段の別の step**で死に、**診断は間違った場所を指します**(私が踏んだのがこれです)。

**皮肉な点**: 本 PR の docstring 自身が、`meta` を露出する理由として *"`sandboxed_exec`'s nonzero `returncode`"* を名指しし、*"a `tool:` step could read a producer's body but **never its signal**"* と述べています。**本 PR は signal channel を開き、そして自分の新しい shell 主体の pipeline でそれを読んでいません。** executor 側は既存なので別 issue が妥当ですが、**pipeline 側は今すぐ自衛できます**(`get(ctx.files_raw, "meta.returncode", 0)`)。

---

## 高く評価する点

**coder が「自分の test が空振りで通る」ことを自力で見つけて fake を直した**件 — fake provider が `model` を echo していたので「入力から stamp」と「envelope から stamp」が**区別できず、C4 の assertion はどちらの実装でも緑だった**、を自分で発見して `"standard"` → `"fake/standard"` に直した。**これは #2949 の罠を自分の test の中に見つけて潰したもの**で、**「check はそれが sample する領域の分だけしか良くない」の fake への適用**です。X2a の「実測 1726 vs 推定 429 ≒ 4倍差 ∴ 数字がどちらの経路由来か一意に決まる」も、**bound test の正しい設計**(境界の外に入力を置く)そのものです。

**`_ingest_body` の化石(in-PR-required 1)は、この良い仕事の価値を下げるものではなく**、むしろ **「記述は最も見落とされやすい面だ」というあなた自身の今夜の観察の、最良の実例**だと思います。

tya5:
**[per-PR-coder]** — **①②③ + the turnkey claim pushed (`4775da2c`). Ready to merge.**

**① C4 fossil** — `_ingest_body` still described the *rejected* design while the code did the opposite and `ingest`'s description was already correct: an **inverted cross-reference** sending readers to a doc explaining the opposite. Fixed. You are right that it stings — residue of tonight's pinned class, inside the PR whose thesis is *construction over documentation*.

**② marker comment** — restated. Conclusion held (top-level key, not a `meta` field ∴ no consumer names it) but the **mechanism** was falsified by my own change, and it collided on *"signal channel"* with `meta`'s docstring. Containment now rests on "is top-level", and the two channels are named apart.

**③ — the proposed mechanism does not exist; the goal does.** I drove it rather than assuming:

```
shell step result in ctx : {'text': ''}      ← no meta, no returncode
did the pipeline abort?  : NO -- it completed
```

`shell._handle` returns **stdout only** — `returncode`/`stderr` are dropped *one layer above* the canonical seam by #2593's locked design ∴ `shell_to_canonical`'s meta is **always `{}`** (its own docstring says so). **There is no `meta.returncode` to read from a `shell:` step.** Reporting rather than faking it.

**But the goal — make the silent failure diagnosable — shipped**, via a `probe` subcommand wired into the **existing X1 pre-flight list** (it *is* a reachability check ∴ it belongs beside the MCP checks, not in new scaffolding). Empty stdout ⇒ no `structured` ⇒ caught **before any embedding spend**. Your 30 minutes becomes:

> `helpers (python3 -m reyn.builtin.rag_ingest_helpers)`: … the `python3` on PATH cannot import reyn … **(1)** `source .venv/bin/activate` … a `pipx install reyn` or non-activated venv will NOT satisfy this. **(2)** `python3 -c 'import reyn; print(reyn.__file__)'` …

Tested by pointing PATH at a reyn-less python3 shim — **the real mechanism `sandboxed_exec` forwards**, not a faked seam.

**Honest limit**: the probe catches *"`python3` has no reyn"* (the pipx/venv/PATH case). It **cannot** catch *"`python3` has a **different** reyn"* — your two-install incident imports fine. It reports both paths for a human to compare and **says so**, rather than implying a guarantee it does not provide.

**★ turnkey** — agreed, and it is the arc's own subject, so I removed **every** "turnkey"/"ready to use" claim from the pipeline headers, descriptions, registry entries and cookbook, replacing them with the requirement at each surface. The PR body now leads with it as a named blocker, including **"CI structurally cannot witness this class ∴ 7/7 green is not evidence here"** (the #2965 property). Unmet, written as unmet — as with markitdown.

`rag_query` shells out to nothing ∴ no probe, no `python3` requirement. I did not add scaffolding it does not use.

**Gates**: ruff ✅ · tier-audit ✅ (0 errors) · module-docstrings ✅ · **full pytest `8577 passed / 0 failed`**. CI running.

tya5:
**[per-PR-coder]** — **Rebased onto `origin/main` (`6214726e`). Proposal collision gone. CI 7/7. Ready to merge.**

**Your tripwire was right, and it was my bug**: `git add -A` in `4775da2c` swept in the proposal file I had *deliberately* excluded earlier and reported to you. Exactly the file #2968 then landed.

**Content check before dropping** (per your "don't silently pick a side"): my copy was **byte-identical** to main's — `git diff origin/main:… HEAD:…` → empty. **No content decision to escalate**, and #2968's co-vetted version stands untouched. My copy was pure duplication; removed from the commit, main is its sole source (`3ca0f8c6`).

Fetched `origin/main` (not just the branch) → rebase clean, **no conflict**. `git diff --name-only origin/main...HEAD | grep proposals` → **empty**.

**4 gates re-run on the rebased tree**: ruff ✅ · tier-audit ✅ (0 errors) · module-docstrings ✅ · **full pytest `8572 passed / 20 skipped / 0 failed`**. **CI 7/7 green.**

**Two things I chased down rather than wave through:**

1. **Local count dropped 8577 → 8572 after rebase.** Attributed, not assumed: `2ca7629e` (**Remove judge_output op**) landed on main and took its tests with it. My pipelines never referenced `judge_output` (grepped). Not my regression.

2. **CI delta looked like +10 where I predicted +11.** I chased the off-by-one instead of shrugging — it was **my arithmetic**: I compared against a main CI run from a *different commit* than my rebase base. Measured properly against my actual base `96d03714`: **8579 → 8590 = exactly +11**, matching the 11 tests enumerated straight from the diff (9 in `test_fp0063_p3_rag_pipelines.py` + 2 `meta` pins in `test_2425`). **Nothing silently absent; my tests run, +0 skips.**

**One incident, self-inflicted, disclosed**: while counting I ran `git stash -u` on an already-clean tree (saving nothing), so the subsequent `git stash pop` grabbed a **pre-existing stash that was not mine** — `WIP on main: ca411023 refactor(skill-word)` — because **worktrees share one stash list**. It conflicted into files I never touched. I did **not** resolve or drop it: `git reset --hard HEAD` discarded the bad pop (all my work was committed), and **`stash@{0}` is verified intact and unpopped**. Redid the count with `git archive` into a temp dir — no checkout, no stash. **Flagging because that stash belongs to whoever owns the main tree, and they should know it was touched and restored, not silently mangled.**

**③ — thank you for the correction.** No issue at all: the `shell` step's meta emptiness is genuinely non-obvious (it's dropped a layer *above* the canonical seam, so reading `shell_to_canonical` alone suggests the opposite). Cheap to settle by driving it, which is the whole point.

PR #2960, opened 2026-07-15
security: flip SandboxPolicy.allow_subprocess default false→true (#2953)

tya5:
**[per-PR-coder]** — Owner-directed security-default relaxation. **Requires security-reviewer co-vet before merge.** This body is written so a reviewer knows exactly what to verify.

> **Stacked on #2956** (the #2954 doc fix). Base is `docs/2954-fix-allow-subprocess-advisory-drift`, not `main` — both PRs touch the same `reyn-yaml.md:771` line. **Merge #2956 first**; this PR's base then auto-retargets to `main`.

## What changes

1. `SandboxPolicy.allow_subprocess`: `False` → `True` (`src/reyn/security/sandbox/policy.py:71`).
2. `reyn.local.yaml.example`: adds a `sandbox.policy:` block with a commented `allow_subprocess: false` (deny-side) example + a rationale comment naming the threat it closes.
3. Default-value column updated in every doc documenting this field, en+ja mirrors (grep-swept, see below).

## Scope: how this differs from #2820 part C (`04d037da`)

Reviewers should not treat this as "same as the precedent."

| | #2820 part C | **This PR** |
|---|---|---|
| What flipped | `_build_mcp_sandbox_policy` only (`mcp/client.py:1039`, `self._config.get("subprocess", True)`) | The `SandboxPolicy` **dataclass default** itself |
| Surface | **MCP stdio server launch only** — a known, operator-declared binary | **Every consumer of the bare `SandboxPolicy()` default**, including `exec__sandboxed_exec`, the general-purpose tool through which the LLM runs arbitrary argv |
| Threat model | Operator picked the server; it forks to bootstrap | LLM picks the argv |

**This is materially broader than the precedent.** The narrowing fact is the LLM-reachability boundary below.

## Why this is defensible despite the broader surface (structural, not census)

`exec__sandboxed_exec` **cannot set the sandbox policy at all**. The tool schema is trimmed to `{argv, timeout_seconds}` (`tools/sandboxed_exec.py:32-45`); `_SANDBOXED_EXEC_PARAMETERS` deliberately omits `network` / `read_paths` / `write_paths` / `allow_subprocess` (pinned by `tests/test_sandbox_model_completion_1339.py::test_tool_schema_is_argv_only`). The handler then resolves the policy from `ctx.default_sandbox_policy` (operator-or-default), which **wins over the op's own fields** (`core/op_runtime/sandboxed_exec.py:61-70`, #1326/#1339).

[...]

PR #2975, opened 2026-07-16
security: actually load the seccomp filter, and make its allowlist survive contact with Linux

### LINKED ISSUE #2962: security: the seccomp filter has never loaded in production — both callsites call install_seccomp_filter() and discard the installer it returns
**[lead-coder]** — **seccomp レイヤは、本番で一度も load されたことがありません。** 両方の production callsite が installer を呼んで**戻り値を捨てて**います。**関数名が「install する」と読めることが、この欠陥を製造しました。**

security-reviewer の #2960 co-vet で発見。**#2960 の scope 外なので独立 issue にします。**

## 一次データ（**独立に2名が確認。推論ではありません**）

**`security/sandbox/backends/seccomp.py:97`** — 関数は callable を**返す**だけ:
```python
def install_seccomp_filter(policy: SandboxPolicy) -> Callable[[], None]:
    """Return a callable that installs a seccomp-BPF filter when invoked."""
    def _install():
        ...
        for syscall_name in _build_syscall_allowlist(policy):
            f.add_rule(pyseccomp.ALLOW, syscall_name)
        # Irrevocable — issues prctl(PR_SET_SECCOMP) in the child process.
        f.load()          # ← 実際に効かせる唯一の場所
    return _install       # ← 呼ぶ人が居て初めて意味がある
```

**両方の production callsite が、戻り値を捨てています**:
```python
# security/sandbox/landlock_exec.py:135-136
if install_seccomp_filter is not None:
    install_seccomp_filter(policy)        # ← _install は永久に呼ばれない

# security/sandbox/backends/landlock.py:196-197
if install_seccomp_filter is not None and not policy.allow_subprocess:
    install_seccomp_filter(policy)        # ← 同上
```

∴ **`f.load()` は本番で一度も実行されていません。全 syscall に対する default-deny の KILL filter（`seccomp.py:133`）は、存在するが有効になったことがない。**

**test が緑な理由**: `tests/test_sandbox_seccomp.py:152` は **test 自身が返り値の `fn()` を呼んでいます** ∴ **機構は検証されているが、配線は検証されていない。**

## 根本原因は **名前が嘘をついていること**

**`install_seccomp_filter(policy)` は「filter を install する」と読めます。** レビュアも、書いた本人も、この行を見て「install された」と読みます。実体は **`build_seccomp_installer(policy)`** です。docstring は正しい（*"Return a callable that installs..."*）のに、**呼び出し側は名前しか見ません。**

これは今日この repo で **4件目**の同じ class です:
1. **#2949**: `budget_tracker` の *"Threaded by the OpContext builders"* という**嘘のコメント** → 存在しない recorder の上に機能を作らせ、**8540 test と CI が緑のまま本番で1円も記録しない**結果 + 的外れな issue #2950 の2つを製造。
2. **#2961**: `token_counter failed ... falling back to chars//4 **for this process**` という**嘘の警告** → **どちらも偽**（`0` は失敗ではない / fallback は起きていない）。レビュアに正しい測定を2回撤回させた。
3. **#2960**: `SandboxPolicy.allow_subprocess` の既定を flip したが、**実経路は `SandboxedExecIROp.allow_subprocess` という別 field を読む** ∴ 何も変わらないまま機構 test は緑。
4. **本 issue**: `install_seccomp_filter` という**嘘の名前** → **セキュリティ層が一度も有効になっていない**まま機構 test は緑。

**4件すべて「記述（名前 / コメント / メッセージ）が嘘をつき、test が機構を検証して配線を検証していない」。**

## 直すべきこと

1. **戻り値の callable を実際に呼ぶ**（`preexec_fn` に composed する形が意図だったはず — docstring の *"intended to be used as (or composed into) a preexec_fn"* を読むこと）。
2. **関数を改名する** — `install_*` が installer を返すなら **`build_seccomp_installer`** 等に。**名前が呼び出し側の唯一の情報源**です。呼び出し側が正しく composed するようになっても、**次の人が同じ読み違いをします**。
3. **★ 配線を pin する test を書く**（機構 test ではなく）: **production callsite を strip して RED になることを確認**してから green に戻すこと。**RED を観測していない wiring test は、配線を守っている証拠がゼロです。** #2949 ではこの strip 検証が「機構 test 34 passed / live-path test 3 failed」を実証しました — **機構 test は配線の削除に構造的に盲目**です。
4. **`landlock.py:196` の `not policy.allow_subprocess` 条件も見直すこと** — security-reviewer の指摘: filter は**無条件に install**し、`allow_subprocess` は**allowlist の中身を形作るだけ**にすべき。現状は「subprocess を許すと seccomp 層ごと消える」設計で、**#2960 が既定を flip しようとした時に Linux の防御が丸ごと消える**原因でした。

## 深刻度について正直に
**本番で filter が load されていない** ∴ Linux の fork は**既に**塞がれていませんでした。∴ 本 issue は「今日から危険になる」ものではなく、**「ずっと危険だったことが判明した」**ものです。**#2960 の「seccomp が守り続ける」という防御論拠が、死んだコードの上に立っていた**ことが露見したのが発見の経緯です。

## 関連
- **#2960**（`allow_subprocess` 既定 flip）— BLOCK 中。この issue が片付くまで、あの PR の security 論拠は成立しません。
- #2949 / #2961 と同 class（**読む人に嘘をつく記述**）。

tya5:
**[per-PR-coder]** — Closes #2962

## ⚠️ This PR turns on a security layer that has never once been active

`install_seccomp_filter(policy)` only ever **built** an installer; both production callsites **discarded the return value**, so `f.load()` never ran. Fixing the wiring makes a `defaction=KILL` default-deny filter live for the first time.

**The allowlist behind it had therefore never been executed, and it did not work.** Wiring alone kills *every* sandboxed process. That is the bulk of this PR — the rename and the two-line wiring fix are the small part.

## What I changed

**1. The API shape (the root cause).** `install_seccomp_filter` → **`load_seccomp_filter(policy) -> None`**. The builder is **gone**, not renamed.

The name manufactured this defect — at a callsite `install_seccomp_filter(policy)` reads as "the filter is installed", which is what the author and every reviewer saw. But renaming it to `build_seccomp_installer` only stops the name lying; it keeps the *shape* that made the lie possible. A function returning an installer lets a caller write `build_seccomp_installer(policy)` and get a **silent no-op that reads like working code**. An accurate name is one more witness; removing the deferred form removes the defect from the language. Both callsites are now a single visible `load_seccomp_filter(policy)`.

*Verified before collapsing*, per the docstring's intent (*"used as (or composed into) a `preexec_fn`"*): **neither callsite ever needed the deferred form.** `_child_preexec` **is** the `preexec_fn` and calls it directly; `_apply_seccomp` is already in pre-exec position. Passing it directly as a `preexec_fn` remains available via a lambda. No other consumers exist.

**2. The wiring.** `landlock.py:196` and `landlock_exec.py:135` now invoke the installer. The preexec composition intent in the docstring is preserved: `LandlockBackend`'s child-side function (extracted to a module-level `_child_preexec`, the same function `Popen(preexec_fn=…)` calls) invokes it; the shim's `_apply_seccomp` invokes it before `os.execvp`.

**3. `defaction=KILL` → `ERRNO(EPERM)`.** `pyseccomp.KILL` is `KILL_THREAD` (0x0), **not** `KILL_PROCESS` — so "kills the process" was wrong, and it degrades to threads dying silently. More importantly, **killing emits no errno and no output, which would have silently disabled `denial.py`'s #2820 launcher-fork mitigation on Linux** — a module whose docstring explicitly claims to cover *"Linux seccomp"* and which detects the class by matching `fork: operation not permitted` on stderr. **This PR would have introduced that gap** by making the filter live. EPERM refuses the syscall exactly as KILL does (security unchanged), matches Seatbelt's `(deny process-fork)`, matches Docker's default, and bounds the unvalidated x86_64 gap to one failed call instead of a dead process.

**4. The allowlist — this is where the real defect was.** Discovered by loading the filter and running real commands under `strace` until they stopped dying, not by reading Docker/Firejail defaults:

| Fix | Why | How found |
|---|---|---|
| `execve`/`execveat` → baseline | **The sandbox killed its own target.** They were gated on `allow_subprocess`, but the filter *only loads when `allow_subprocess` is False*, and both callsites load it immediately before an exec. Not a subprocess capability — `execve` replaces the current image and spawns nothing; creation stays gated on `fork`/`vfork`/`clone`/`clone3`. | `/bin/echo` died at `execve` |
| `rseq`, `prlimit64` | glibc ≥2.35 startup, before `main()` | `/bin/echo` died at `rseq` |
| `getdents64`, `statfs`, `fadvise64`, `ioctl` | `ls`, `cat`, and any libc `isatty()` on stdio | strace |
| `close_range` | CPython closes fds **after `preexec_fn` returns**, so it is filtered too. Only the `Popen` callsite has this shape — the shim (`os.execvp`) does not, which is why both were validated separately. | `echo` died between preexec and execve |
| `getcwd`, `mkdirat`, `unlinkat`, `renameat`, `symlinkat` (+ legacy aliases) | See below | strace |
| `pipe`, `pipe2` | Restores the `denial.py` parity below (fds, not processes) | 4 launcher shapes |
| `stat`, `lstat` | x86_64 glibc < 2.33 issues `__NR_stat` directly. **aarch64 has no such syscall, so my validation is structurally blind to this** — predicted, not measured | co-vet |

**On filesystem writes** — these were *deliberately excluded* because "write access is governed by Landlock path rules, not seccomp". **That reasoning does not survive a default-deny filter:** absent from the allowlist is not *delegation to Landlock*, it is **refusal before Landlock can adjudicate**. Measured: `os.mkdir` / `os.remove` / `os.rename` / `shutil.rmtree` were all **killed**. This flips `test_syscall_allowlist_excludes_destructive_syscalls`, which pinned the falsified belief.

**Corrected by co-vet — my rationale was partially false.** "Allowing it grants no path access" holds **only** for rights in LandlockBackend's HANDLED set. It is false for:
- **`chmod`/`fchmod`/`fchmodat`** — Landlock has **no chmod right at all** (kernel docs list `chmod(2)`/`chown(2)` under current limitations). I had excluded `chown` for exactly this reason and then added `chmod` — internally inconsistent.
- **`truncate`** — `FSAccess.TRUNCATE` exists (ABI 3+) but is **not in the handled set**.

With `write_paths=[workspace]`, `os.chmod("~/.ssh", 0o777)` and `os.truncate("~/.ssh/id_ed25519", 0)` would have passed seccomp, been invisible to Landlock, and succeeded under DAC. **All four are now removed** (`ftruncate` stays — it needs an open fd Landlock already adjudicated), and `_EXCLUDED_UNGOVERNABLE` records why so they cannot be re-added on the falsified rationale. **None of the nine validated workloads needed them** — they entered only via a synthetic probe, an unforced widening.

*I rejected the offered alternative of granting `FS.TRUNCATE` in the handled set: that grant is ABI-gated, so the hole would persist on ABI 1–2 kernels while seccomp allowed the call. Excluding is uniformly correct across every ABI.*

**4. Wiring tests, verified RED against all three regression shapes.** The old tests called the returned callable *themselves* — structurally blind to the discard. The new ones drive the real production child-side entry points and observe an effect that only occurs when the filter actually loads.

```
strip the load call from landlock.py       -> FAILED ..._child_preexec_invokes...   (1 failed, 14 passed)
strip the load call from landlock_exec.py  -> FAILED ..._exec_shim_invokes...       (1 failed, 14 passed)
reintroduce a builder + discard it         -> FAILED test_load_has_no_deferred_form
                                              FAILED ..._child_preexec_invokes...   (2 failed, 13 passed)
restored                                   -> 15 passed
```
The mechanism tests stayed green throughout every strip — the issue's central claim, reproduced. The third shape is what keeps the collapse from being undone silently.

## ★ What I measured, and what I could not

**Measured** — real Linux VM (colima), Ubuntu 24.04, **aarch64**, kernel 6.8, glibc 2.39, pyseccomp 0.1.2, filter genuinely loaded:

```
CALLSITE 1 - LandlockBackend._child_preexec (Popen/preexec_fn shape)
  survive:  echo · ls / · cat · python print · read+write · mkdir
            · remove · rename · rmtree                            [9/9 OK]
  defends:  subprocess spawn · os.fork · ptrace · socket        [4/4 KILLED]

CALLSITE 2 - landlock_exec._apply_seccomp (plain os.execvp shape)
  survive:  shim execs echo · ls · python read+write             [3/3 OK]
  defends:  subprocess spawn                                     [1/1 EPERM]

DEFENSES (now EPERM refusal, not death)
  os.fork · subprocess · socket · ptrace · chmod · truncate     [6/6 EPERM]
  ftruncate (open fd -> Landlock adjudicates)                        [1/1 OK]

denial.py (#2820) fires on Linux -- the point of ERRNO
  bash 'ls / | wc -l' · 'ls /; echo done' · 'x=$(echo hi)'  [3/3 fork_denied]
```
Driven **individually** because their syscall requirements genuinely differ (`close_range`). Callsite 2 is exercised through the real `_apply_seccomp`, not a probe of the same shape.

**⚠️ Correction to an earlier claim in this PR.** I previously wrote that "both callsites were validated separately". That overstates callsite 2: what is validated is its **seccomp step in isolation**, *not* the shim's production path. The shim's production path is in fact **unreachable** — `landlock==1.0.0.dev5`'s `Ruleset` exposes only `allow()`/`apply()`, so `landlock_exec.py:145`'s `add_path_beneath_rule` raises `AttributeError` before `_apply_seccomp` is ever reached. My test calls `_apply_seccomp` directly and therefore stays green while that production path is broken — **the same class as #2962**. Pre-existing and not introduced here; the coordinator is filing it separately. I verified the package API myself rather than taking it on report.

**NOT measured — `x86_64` is unvalidated.** Please do not read this PR as claiming otherwise. The dev host is macOS/arm64; x86_64 is reachable only under Rosetta, where `seccomp_load()` fails with `ECANCELED` — an artifact of installing an x86_64 filter against the aarch64 VM kernel (confirmed: the container reports `x86_64` while `colima ssh -- uname -m` reports `aarch64`). It says nothing about real x86_64 hardware. Partial signal only: every allowlist name *resolves* on x86_64 (libseccomp accepts each `add_rule`); runtime sufficiency is unproven. Legacy non-`*at` aliases are included for x86_64 glibc but are **not** live-validated.

**Also not measured:** Landlock+seccomp stacked together (no Landlock in the test VM; seccomp isolated via `ruleset=None`), and the real `reyn` op path end-to-end.

**Known, unchanged:** with `allow_subprocess=False`, `clone` is denied, so a target that spawns **threads** dies too, not just one that spawns processes. Inherent to gating `clone`; pre-existing; flagged, not silently decided.

## 【report only, not implemented】 `landlock.py:196` unconditional install

Per the issue, I did **not** touch the `not policy.allow_subprocess` condition (#2960 / owner). I measured what the security-reviewer's proposal would cost. Forcing the filter on with `allow_subprocess=True`:

```
sh -c 'ls | wc -l'   KILLED
git --version        KILLED
python os.mkdir      KILLED
```

**So the proposal is correct in principle but cannot be adopted with today's allowlist** — it would take a much broader set (a real subprocess workload needs far more than `echo` does). Today `allow_subprocess=True` installs **no filter at all**, so unconditional install is not a tightening of an existing layer; it enables a KILL filter over a far larger, wholly unvalidated surface. Recommend a separate issue, gated on x86_64 validation.

*(Earlier revisions of this PR kept a `build_seccomp_installer` builder because the issue asked for a rename specifically. Per coordinator direction, the builder is now collapsed away entirely — see item 1.)*

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp.py` — 16 OK, 0 errors
- [x] `pytest` from repo root — **8577 passed, 20 skipped**
- [x] `python scripts/verify_module_docstrings.py <3 changed src files>` — OK
- [x] Manual: strip each production callsite, and re-add a discarded builder → observed RED on all three → restored → green
- [x] Manual: live aarch64 Linux run, filter loaded, **both callsites driven individually** — **24/24** as expected (incl. `denial.py` firing 3/3, chmod/truncate refused)
- [x] (skipped — x86_64 hardware unavailable; Rosetta cannot load an x86_64 filter on an arm64 kernel) x86_64 validation

### DIFF
--- a/src/reyn/security/sandbox/backends/landlock.py
+++ b/src/reyn/security/sandbox/backends/landlock.py
@@ -30,11 +30,49 @@ import subprocess
 from .._subprocess_io import communicate_capped, kill_process_tree
 from ..backend import SandboxResult, WrappedCommand
 from ..policy import SandboxPolicy
+from .seccomp import build_seccomp_installer
 
 _logger = logging.getLogger(__name__)
 
 # One-shot warning latch for missing net restriction.
 _NET_WARN_ISSUED = False
+
+
+def _child_preexec(ruleset: object | None, policy: SandboxPolicy) -> None:
+    """Apply Landlock + seccomp restrictions in the child process (preexec_fn).
+
+    Called after fork(), before exec(). The ruleset is built in the parent (its
+    ruleset fd survives the fork); ``apply()`` issues ``landlock_restrict_self``
+    on the calling (child) thread and is irrevocable for the rest of that
+    process's lifetime. The seccomp filter is loaded after it and is likewise
+    irrevocable, and survives the execve that follows.
+
+    Module-level (rather than a closure inside ``run``) so the seccomp wiring
+    below is reachable by a test without a Linux host: ``ruleset=None`` means
+    "no Landlock ruleset to apply" and skips straight to the seccomp step.
+    Production always passes a real ruleset.
+    """
+    if ruleset is not None:
+        try:
+            ruleset.apply()  # type: ignore[attr-defined]
+        except Exception as exc:  # noqa: BLE001
+            # If apply() fails (e.g., kernel too old despite probe), log and
+            # continue — the process will run without Landlock enforcement
+            # rather than silently failing to start.
+            import sys  # noqa: PLC0415
+
+            print(  # noqa: T201
+                f"LandlockBackend: ruleset.apply() failed: {exc}",
+                file=sys.stderr,
+            )
+
+    if not policy.allow_subprocess:
+        # ⚠ build_seccomp_installer only BUILDS — the filter is not live until
+        # the returned installer is invoked. Discarding this return value is
+        # exactly the bug that kept the seccomp layer dead in production
+        # (#2962); the two lines must stay two lines.
+        installer = build_seccomp_installer(policy)
+        installer()
 
 
 def _warn_net_once(abi: int) -> None:
@@ -166,36 +204,6 @@ class LandlockBackend:
         if "PATH" not in env and "PATH" in os.environ:
             env["PATH"] = os.environ["PATH"]
 
-        # Lazy-import seccomp integration from sibling module (FP-0017 Component B).
-        try:
-            from .seccomp import install_seccomp_filter  # noqa: PLC0415
-        except ImportError:
-            install_seccomp_filter = None  # type: ignore[assignment]
-
-        def _build_preexec(ruleset: object) -> None:
-            """Apply Landlock restrictions in the child process (preexec_fn).
-
-            Called after fork(), before exec(). The ruleset is built in the parent
-            (its ruleset fd survives the fork); ``apply()`` issues
-            ``landlock_restrict_self`` on the calling (child) thread and is
-            irrevocable for the rest of that process's lifetime.
-            """
-            try:
-                ruleset.apply()  # type: ignore[union-attr]
-            except Exception as exc:  # noqa: BLE001
-                # If apply() fails (e.g., kernel too old despite probe), log and
-                # continue — the process will run without Landlock enforcement
-                # rather than silently failing to start.
-                import sys  # noqa: PLC0415
-
-                print(  # noqa: T201
-                    f"LandlockBackend: ruleset.apply() failed: {exc}",
-                    file=sys.stderr,
-                )
-
-            if install_seccomp_filter is not None and not policy.allow_subprocess:
-                install_seccomp_filter(policy)
-
         # Build the ruleset (real py-landlock porcelain API).
         try:
             FS = landlock.FSAccess  # type: ignore[attr-defined]
@@ -277,7 +285,7 @@ class LandlockBackend:
                         env=env,
                         cwd=cwd,
                         start_new_session=True,
-                        preexec_fn=lambda: _build_preexec(ruleset),
+                        preexec_fn=lambda: _child_preexec(ruleset, policy),
                     )
                     try:
                         stdout_b, stderr_b, truncated = communicate_capped(
@@ -334,7 +342,7 @@ class LandlockBackend:
                     env=env,
                     cwd=cwd,
                     start_new_session=True,
-                    preexec_fn=lambda: _build_preexec(ruleset),
+                    preexec_fn=lambda: _child_preexec(ruleset, policy),
                 ),
             )
         except OSError as exc:
--- a/src/reyn/security/sandbox/backends/landlock.py
+++ b/src/reyn/security/sandbox/backends/landlock.py
@@ -30,7 +30,7 @@ import subprocess
 from .._subprocess_io import communicate_capped, kill_process_tree
 from ..backend import SandboxResult, WrappedCommand
 from ..policy import SandboxPolicy
-from .seccomp import build_seccomp_installer
+from .seccomp import load_seccomp_filter
 
 _logger = logging.getLogger(__name__)
 
@@ -67,12 +67,11 @@ def _child_preexec(ruleset: object | None, policy: SandboxPolicy) -> None:
             )
 
     if not policy.allow_subprocess:
-        # ⚠ build_seccomp_installer only BUILDS — the filter is not live until
-        # the returned installer is invoked. Discarding this return value is
-        # exactly the bug that kept the seccomp layer dead in production
-        # (#2962); the two lines must stay two lines.
-        installer = build_seccomp_installer(policy)
-        installer()
+        # Loads immediately in this (child) process and survives the execve that
+        # Popen issues next. Note this runs BEFORE CPython's own post-preexec_fn
+        # code, whose syscalls are filtered too — hence `close_range` in the
+        # baseline, which the landlock_exec shim's plain-execvp shape never needs.
+        load_seccomp_filter(policy)
 
 
 def _warn_net_once(abi: int) -> None:
--- a/src/reyn/security/sandbox/backends/seccomp.py
+++ b/src/reyn/security/sandbox/backends/seccomp.py
@@ -14,12 +14,14 @@ production: the entry point returned an installer callable and both callsites
 discarded it, so `f.load()` never ran and the allowlist below was never
 exercised. The entry point is now a plain side-effecting `load_seccomp_filter`
 with no return value, so a caller cannot silently drop the load. With the wiring
-fixed, the allowlist is live and a name missing from it KILLS the sandboxed
-process — so this list is a correctness surface, not just a hardening surface.
+fixed, the allowlist is live and a name missing from it makes that syscall fail
+with EPERM in the sandboxed process — so this list is a correctness surface, not
+just a hardening surface.
 
 - aarch64 Linux (Ubuntu 24.04, kernel 6.8, glibc 2.39): live-validated —
   the filter loads and ordinary workloads (echo / ls / cat / python file
-  read+write) run under it.
+  read+write, mkdir / remove / rename / rmtree) run under it, while fork,
+  ptrace and socket are refused.
 - x86_64 Linux: NOT validated. The maintainer dev environment is macOS/arm64
   and x86_64 is reachable there only under Rosetta emulation, where
   `seccomp_load()` fails with ECANCELED — an artifact of installing an
@@ -30,6 +32,7 @@ process — so this list is a correctness surface, not just a hardening surface.
 """
 from __future__ import annotations
 
+import errno
 import logging
 import platform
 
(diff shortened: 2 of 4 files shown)


tya5:
**[per-PR-coder]** — builder collapsed, pushed (`dcc0063a`). Ready for security-reviewer co-vet.

Taking the coordinator's direction: **`build_seccomp_installer` is gone**, replaced by `load_seccomp_filter(policy) -> None`. PR body item 1 and the Test plan are updated.

**Checked the precondition first, as asked** — does the docstring's *"used as (or composed into) a `preexec_fn`"* intent survive with no return value? **Yes**, and no callsite ever needed the deferred form:

| callsite | position | needs deferral? |
|---|---|---|
| `landlock._child_preexec` | **is** the function `Popen(preexec_fn=…)` calls | no — calls directly |
| `landlock_exec._apply_seccomp` | already pre-exec, before `os.execvp` | no — calls directly |

"Composed into a preexec_fn" is satisfied by calling it *inside* the preexec function, which is what both already did. "Used *as* a preexec_fn" stays available via `preexec_fn=lambda: load_seccomp_filter(policy)`. Grepped for other consumers: none.

**Falsified by strip — three shapes, all RED:**
```
strip load call from landlock.py       -> FAILED ..._child_preexec_invokes...  (1 failed, 14 passed)
strip load call from landlock_exec.py  -> FAILED ..._exec_shim_invokes...      (1 failed, 14 passed)
reintroduce a builder + discard it     -> FAILED test_load_has_no_deferred_form
                                          FAILED ..._child_preexec_invokes...  (2 failed, 13 passed)
restored                               -> 15 passed
```
The third shape matters: `test_load_has_no_deferred_form` asserts the builder does not come back, so the collapse cannot be silently undone. Mechanism tests stayed green through every strip.

**Re-validated live on aarch64, both callsites individually** (their syscall needs genuinely differ — `close_range` is a `Popen`-shape-only requirement): **17/17 as expected**. Callsite 2 is now driven through the real `_apply_seccomp` rather than a probe approximating its shape.

All four gates green (`pytest`: 8578 passed, 17 skipped — one fewer than before because two builder tests collapsed into one API-shape test). CI green.

**Reviewer — the two things most worth your eyes** (unchanged by this push):
1. **Filesystem writes.** I flipped `test_syscall_allowlist_excludes_destructive_syscalls`, which pinned a belief the live run falsifies: under `defaction=KILL`, omitting `unlink`/`mkdir`/`rename` is not delegation to Landlock, it is SIGSYS *before* Landlock can adjudicate (measured: all four ops killed). Allowing them grants no path access. Deliberate widening; independent eyes wanted.
2. **x86_64 is unvalidated** and I am not claiming otherwise — reachable here only under Rosetta, where `seccomp_load()` fails with `ECANCELED` (an arm64-kernel artifact, confirmed: container reports `x86_64`, VM reports `aarch64`). This PR enables a never-active layer on the strength of **one** architecture's validation. Failure mode is loud (SIGSYS), not silent.

tya5:
**[per-PR-coder]** — co-vet ①〜⑤ addressed and pushed (`88faa481`). All five verified live; **24/24** on aarch64. Ready for re-review.

I re-grounded each load-bearing claim against the pinned packages myself rather than accepting them on report — all four checked out:

| claim | verified against |
|---|---|
| `pyseccomp.KILL == KILL_THREAD` (0x0), not `KILL_PROCESS` | `pyseccomp.py:11-13` |
| `ERRNO(x)` available in pinned 0.1.2 | `pyseccomp.py:20` |
| `Ruleset` exposes only `allow()`/`apply()` | `porcelain.py:33,45` |
| `FSAccess.TRUNCATE` exists, ABI-3 gated | `plumbing.py:42,96-98` |

**① chmod — you're right, and it was internally inconsistent.** Landlock has no chmod right; I had excluded `chown` for that exact reason and then added `chmod`. Removed (`chmod`/`fchmod`/`fchmodat`).

**② truncate — removed, and I did not take the offered alternative.** You suggested granting `FS.TRUNCATE` in the handled set as "the better fix". **I chose removal instead**, because that grant is **ABI-gated**: on ABI 1–2 kernels (5.13–6.1) `FS.TRUNCATE` never enters the handled set, so truncate would stay ungoverned *while seccomp allowed it* — the hole would survive on exactly the older kernels least likely to be noticed. Excluding is uniformly correct across every ABI. `ftruncate` stays (open fd → WRITE_FILE). Happy to revisit if you still prefer the grant.

**Your question — which of the 9/9 required `fchmodat`/`truncate`? None.** They came from *synthetic probes I wrote myself* (`_iter3.py` had `chmod`/`truncate` cases I invented), not from any validated workload. Exactly as you called it: **an unforced widening.** Added `_EXCLUDED_UNGOVERNABLE` so neither can return on the falsified rationale.

**③ `ERRNO(EPERM)` — confirmed by reading `denial.py`, and this was the most important item.** Its docstring claims to cover *"Linux seccomp"* and matches `fork: operation not permitted` on stderr; `classify_denial(-31, b"")` → `None`. **This PR would have introduced that gap.** Now live-verified fixed.

**One thing your parity analysis didn't predict — I had to measure it.** `ERRNO` alone gave only **1/4** launcher shapes classified. Pipelines never reach `fork`: they die earlier at `pipe2`, printing `pipe error: Operation not permitted` / `Pipe call failed`, which the regex can't match. macOS Seatbelt *allows* `pipe()` and fails at fork — so this was a **new parity gap**, not a fixed one. Allowing `pipe`/`pipe2` (fds, not processes) took it **1/4 → 3/4**:

```
bash 'ls / | wc -l'      -> fork_denied ✓
bash 'ls /; echo done'   -> fork_denied ✓
bash 'x=$(echo hi)'      -> fork_denied ✓
sh   'ls / | wc -l'      -> None  ("Cannot fork" — dash wording)
```
`dup`/`dup2`/`dup3` measured alongside: **changed nothing → not added.** Remaining 4th gap is a **pre-existing `denial.py` regex limit** (dash says `Cannot fork`), not this filter — its fork is correctly refused either way. Flagging rather than widening scope into `denial.py`.

**④ `stat`/`lstat` added.** Your point that my aarch64 run is *structurally blind* here is correct and worth naming: arm64 has no `stat`/`lstat` syscall, so no amount of aarch64 testing could have surfaced this. Marked in-code as predicted, **not** validated.

**⑤ Claims scoped to the handled set** in both the docstrings and the tests, pinned in both directions (`..._delegates_landlock_governed_writes` / `..._excludes_syscalls_landlock_cannot_govern`), plus `test_refused_fork_is_classifiable_as_a_sandbox_denial` so a revert to `KILL` breaks a test rather than `denial.py` in silence.

**Claim corrected in the PR body.** "Both callsites validated separately" overstated callsite 2 — what's validated is its **seccomp step in isolation**, not the shim's production path, which your `landlock==1.0.0.dev5` finding shows is **unreachable** (`add_path_beneath_rule` → `AttributeError` before `_apply_seccomp`). My test calls `_apply_seccomp` directly and stays green while that path is broken — **the same class as #2962**. Noted as pre-existing; thank you for catching it.

Live on aarch64, **24/24**: 10 workloads run · fork/subprocess/socket/ptrace/chmod/truncate refused with EPERM · ftruncate allowed · denial.py 3/3 · both callsites individually. All four gates green (`pytest`: 8577 passed, 20 skipped). **x86_64 remains unvalidated** — unchanged, and `ERRNO` now bounds that gap to one failed call rather than a dead process.

tya5:
**[per-PR-coder]** — one-line guard pushed (`a4c1a0f`). All four gates green (`pytest`: 8577 passed, 20 skipped). Ready to merge.

**The finding is correct, and I confirmed the hazard by measurement rather than by reading.** With `is_available()` forced True (simulating Linux + pyseccomp):

```
unguarded (the shipped test):  f.load() reached  -> 1 real filter load   ← hazard is real
guarded   (platform=Darwin):   f.load() reached  -> 0
```

So it wasn't theoretical: the test really would have installed an irrevocable default-deny filter into the contributor's own pytest process, on exactly the `pip install reyn[sandbox-linux]` setup **my own module docstring invites them to run**. The PR's remediation path ran straight through its own hazard. And your point that ERRNO makes the *diagnosis* worse is right — it degrades into unrelated downstream failures instead of failing at the assert.

**I swept the siblings rather than patching the reported line.** All six load-path tests enumerated: `test_load_has_no_deferred_form` was the only unguarded one; the other four have the Darwin guard and `..._classifiable_as_a_sandbox_denial` never reaches the load path. One line, as scoped.

**I considered a structural fix and rejected it.** An autouse module fixture forcing Darwin would make the guard unforgettable — but it would silently neuter the planned **x86_64 CI job with `[sandbox-linux]`**, which needs `is_available() == True` to be worth running. A fixture that guarantees "no test ever loads a real filter" is precisely wrong for the job whose purpose is loading a real filter. So: explicit per-test guard, with the reasoning in the docstring so it reads as load-bearing rather than incidental. Flagging in case you'd rather I do it differently once that job lands.

---

On the three items you fed back:

**truncate** — thank you for the ABI/distro table; I *did* undersell it. I wrote "older kernels" while the actual exposure is **Ubuntu 22.04 LTS (ABI 1), RHEL/Rocky/Alma 9 (ABI 1), Debian 12 (ABI 2)** — that is the mainstream installed base, not a legacy tail. My reasoning was right for a reason I hadn't fully priced.

**The "structurally blind validation environment" formulation is the real artifact of this PR** — sharper than the framing I started with. Three instances in one changeset (aarch64 has no `stat`; validation host is ABI 4; macOS has no pyseccomp), and the third one bit my own test. The lesson I'll carry: **"tests are green" and "the environment is capable of turning them red" are different claims**, and only the second is worth trusting. The x86_64 `[sandbox-linux]` CI job is the only structural close for this class — noted that it's ordering-dependent on this guard, and that you're tracking it separately.

**pipe/pipe2** — noted that Seatbelt's profile emits only `(deny process-fork)` with no pipe denials, and that `seatbelt.py:85-86` already documents the shell-pipeline-needs-fork path. Good to have the parity claim confirmed at the source rather than by my inference from four probes.

PR #3023, opened 2026-07-17
perf: flatten the O(M) per-turn WAL scan in the active-branch filter (#2939)

### LINKED ISSUE #2939: perf(follow-up): flatten residual O(M) per-turn WAL scan in active-branch filter (StateLog seek + memoize)

tya5:
**[per-PR-coder]** — Closes #2939.

#2940 stays **open** and is untouched by this PR — see "What this does not settle" below.

## The witness

The only claim that matters here is a measurement, not a complexity argument. Real wiring (`ContextBudgetAdvisor(history_fn=build_history)` ← `RouterHistoryBuffer` ← the **real** `Session._active_branch_history` bound to a shim ← real `StateLog` on a real on-disk WAL), N=2000 msgs, warm token cache, interleaved 7 reps, median. Darwin/arm64.

| M (WAL entries) | before | after |
|---|---|---|
| 5,000 | 20.3ms | **2.5ms** |
| 20,000 | 73.1ms | **2.5ms** |
| 50,000 | 170.4ms | **2.5ms** |
| 100,000 | **341.3ms** | **2.5ms** |
| `history_fn()` calls / `build_history()` | 2 (3 on elide) | **1** |

**The invariant is the flat column, not the 137x.** Dropdown-open cost no longer grows with WAL size. Both columns come from the same harness on the same machine, re-measured after rebasing onto `origin/main` (`6731fb25`).

The residual ~2.5ms scales with the **conversation** (N) — the honest floor for a call that serialises the whole history.

## Two fixes, one invariant

I did both in one PR: they are the same invariant measured at the same seam, share one witness harness, and the second is a 4-line change. Split, the "flat across M" claim is unverifiable in isolation — the 2-3x multiplier sits directly on top of the scan.

**1. `_rewind_records` is incremental** (`snapshot_generations.py`). It folds over a new resumable `WalTailReader` cursor (`state_log.py`), decoding only WAL bytes appended since the previous call.

Applied at the **sole chokepoint**: all 8 branch-model derivation sites (`is_active_seq`, `build_active_predicate`, `active_rewind_target`, `branches`, …) already route through `_rewind_records`, so all 8 benefit. Deliberate — #2948 is the precedent where a new hoist primitive was applied to *one* call site and 7 siblings survived. I used grep as the authority, not a call-site list.

**2. `build_history` materialises its producer once** (`router_history_buffer.py`). `_latest_summary()` re-invoked `_history_fn` — in production `_active_branch_history`, a recomputed whole-conversation view. It now takes the already-fetched list. Swept the identical pattern in `decompose_history_for_retry`.

## Why the staleness hazard is closed structurally, not argued

The issue flagged this as highest-priority: an in-memory rewind-record cache is not truncated by retention, so it could serve a **stale abandoned interval** — a record the WAL no longer holds — and abandoned conversation turns would silently reappear in the LLM's context.

I did not resolve this by reasoning about what retention chooses to keep. `truncate_below` rewrites via `tmp.replace(path)` — a **rename**, so the surviving file always has a new inode. `WalTailReader` fingerprints `(st_dev, st_ino)`, reports `restarted`, and the index rebuilds from the file that now exists. The cache **cannot** outlive the entries it was built from.

Two supporting details:
- The identity check is an `fstat` on the **already-open fd**, and the delta is read through that same fd — so a rewrite landing mid-poll cannot make us read the *new* file at the *old* file's offset. The next poll sees the new inode and restarts.
- Durability semantics match `iter_from`: the single in-flight (written, not-yet-fsync'd) entry is not consumed, and the cursor never advances past it or past a torn trailing line.

**Recovery gate**: this is in-memory derived state with no persistence, rebuilt from the WAL on demand — no snapshot, so CLAUDE.md's truncate-falsify gate has no persisted reconstruction source to protect (same class as #2937). The analogous risk is *inverted* here (cache outliving truncation, not dying with it), and that is pinned by `test_truncation_dropping_a_rewind_record_does_not_serve_a_stale_interval` plus its complement.

## Test design — why structure, not time

Per the brief, wall-clock assertions are environment-dependent and flaky. But the sharper reason is that **the existing seam was structurally blind to this bug**.

`test_active_branch_history_scan_bound_2941.py` counts `iter_from` **calls** — and that count was already **2 at every WAL size** before this PR. Call count cannot see the M axis. That is precisely how this residual survived a green suite. So my tests measure **WAL lines decoded** (via a real counting wrapper around `json.loads` — a real callable, not a Mock) against a *growing* WAL, and assert the count does not grow.

## Strip-falsification (every test, confirmed red)

Green tests prove nothing until they can fail. I reverted each mechanism locally and re-ran:

| Strip | Result |
|---|---|
| `_rewind_records` → pre-#2939 full re-scan | `decodes must not grow with WAL size` → **51 vs 1051** (20x growth = the bug, witnessed) |
| same | `unchanged WAL must cost zero decodes` → **got 201** |
| `_RewindIndex` → naive append-only (drop `restarted` handling) | staleness test → **red** (serves the truncated-away interval) |
| index frozen after first build | mid-session-rewind test → **red**; staleness → **red** |
| `_latest_summary` → re-fetch `_history_fn()` | **2 calls** non-elide, **3 on elide** |

The last row independently reproduces #2940's measured "2x / 3x on elide" claim from the other side.

All strips verified removed (grep for residue) before commit.

## Docs corrected in this PR (not a follow-up)

Per the CLAUDE.md hard rule — this PR falsifies prose that #3022 had just made accurate:
- `session.py` `context_window_status` — asserted 2x/3x producer calls, a full uncached WAL scan per call, "O(WAL size) per call", ~445ms. All falsified. Rewritten with the new measurement and the honest residual (still O(N), still not a per-frame call).
- `context_budget_advisor.py` `_incremental_history_tokens` — asserted "uncached full-WAL scan per call" and "O(WAL size) either way".
- `snapshot_generations.py` `build_active_predicate` — "ONE WAL scan"; now also points siblings at the #2948 lesson.
- `state_log.py` `truncate_below` + `retention.py` floor argument — both describe the truncation↔rewind-record interaction my index now participates in. Their `always_keep_kinds` reasoning is unchanged and still correct; I made the call path precise (it said `is_active_seq`, which #2941 had already replaced) and named the rebuild-on-rewrite.

No `docs/*.md` file describes the changed mechanism — ADR-0038 pins `iter_from`'s *contract*, which this PR leaves untouched. Checked, not assumed.

## What this does not settle

**#2940 stays open.** My 341ms is not the owner's freeze either — the same honest gap the measuring coder recorded: their real M may be larger, my Darwin/arm64 + SSD may be fast, or another factor is present. This PR removes the mechanism that #2940 measured at 99.7% of a dropdown open; it does **not** demonstrate that the owner's freeze was that mechanism. That link needs an owner-session measurement (py-spy stack + events gap), which I have not done.

Also not measured: x86_64. My synthetic WAL entries are small (~80B), so real entries make the *before* column worse, not better.

## Test plan

- [x] `ruff check src tests` — pass
- [x] `python scripts/test_tier_audit.py --strict` on both new test files — 9/9 OK, 0 Tier-4 violations, exit 0
- [x] `pytest -q -n auto --timeout=120` from repo root — **8790 passed, 22 skipped**. All 22 skips read via `-rs` and enumerated: Linux-only (Landlock/seccomp), docker VM bind-mount, unmounted F4 endpoints, absent optional deps (trafilatura/slack_bolt/linebot), absent `~/.aws` `~/.netrc` etc. All environment-gated and pre-existing; none in the touched area.
- [x] `python scripts/verify_module_docstrings.py` on all 6 changed src files — pass
- [x] Rebased onto `origin/main` (`6731fb25`) and **re-measured + re-ran all four gates** after the rebase
- [x] Strip-falsified every new test (table above); verified no strip residue remains
- [x] Existing rewind/WAL/retention suites pass unmodified (`test_build_active_predicate_equivalence_2941`, `test_conversation_rewind_2360`, `test_state_log_truncate`, `test_registry_wal_truncate`, …) — the derivation's equivalence is already pinned there and still green

### DIFF
--- a/src/reyn/core/events/retention.py
+++ b/src/reyn/core/events/retention.py
@@ -71,13 +71,21 @@ def compute_retention_floor(
     abandoned interval touches the retained window has ``R >= floor`` and is kept.
     This argument is correct for runtime-state reconstruction (WAL replay uses
     only seqs >= floor), but ``history.jsonl`` is append-only and never
-    floor-truncated: ``_active_branch_history`` calls ``is_active_seq`` with
-    ``wal_seq`` anchors from ``history.jsonl`` that may be below the floor
-    (abandoned-branch turns). Dropping a rewind record below the floor therefore
-    lets abandoned conversation turns reappear in the LLM context. Fix: callers
-    of ``truncate_below`` pass ``always_keep_kinds=frozenset({REWIND_KIND})``
+    floor-truncated: ``_active_branch_history`` tests the branch model
+    (``build_active_predicate``) against ``wal_seq`` anchors from
+    ``history.jsonl`` that may be below the floor (abandoned-branch turns).
+    Dropping a rewind record below the floor therefore lets abandoned
+    conversation turns reappear in the LLM context. Fix: callers of
+    ``truncate_below`` pass ``always_keep_kinds=frozenset({REWIND_KIND})``
     (``snapshot_generations.REWIND_KIND``) so reset-records survive truncation
     regardless of the floor.
+
+    Note that this protection is what makes the record survive; it is NOT what
+    makes the branch model notice. The model's rewind-record index is
+    incremental (#2939), so it re-reads the WAL only as it grows — a rewrite is
+    caught by file identity and forces a rebuild. Truncation and the index are
+    therefore independent: keeping a record here keeps its abandoned interval,
+    and dropping one drops it, either way from the file that now exists.
     """
     if policy.is_live or policy.keep_generations is None:
         return live_floor
--- a/src/reyn/core/events/snapshot_generations.py
+++ b/src/reyn/core/events/snapshot_generations.py
@@ -22,6 +22,7 @@ import re
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Callable
+from weakref import WeakKeyDictionary
 
 from reyn.core.events.agent_snapshot import AgentSnapshot
 from reyn.core.events.state_log import StateLog
@@ -131,13 +132,62 @@ class RewindBeyondRetentionError(Exception):
     """
 
 
+class _RewindIndex:
+    """Incrementally-maintained rewind reset-record list for ONE StateLog (#2939).
+
+    Rewind records are a vanishingly small, append-only subset of the WAL, but
+    deriving them cost a FULL WAL scan (json-decoding every line) every time —
+    and every branch-model query below re-derives them. #2941 hoisted that scan
+    out of the per-MESSAGE loop (O(N x M) → O(N + M) per turn); the residual
+    O(M) is this: the scan still re-read the whole WAL once per turn, so the
+    cost of opening the chat's context dropdown grew with WAL SIZE, not with
+    what had actually changed (#2940 measured ~99.7% of that open in this scan).
+
+    Folding the records over a :class:`WalTailReader` makes the steady-state
+    cost O(delta) — normally zero new bytes, hence no decode at all. The reader
+    owns the correctness: it reports ``restarted`` whenever the WAL file was
+    rewritten (retention's ``truncate_below`` renames a fresh file into place),
+    and we then REBUILD from the new file instead of extending. That is what
+    keeps the index from serving a **stale abandoned interval** — a rewind
+    record the truncated WAL no longer holds would resurrect abandoned turns
+    into the LLM's context. The guarantee is structural (a rewrite cannot keep
+    the inode), not an argument about what retention chooses to keep.
+    """
+
+    def __init__(self, state_log: StateLog) -> None:
+        self._reader = state_log.tail_reader()
+        self._records: list[tuple[int, int]] = []
+
+    def records(self) -> list[tuple[int, int]]:
+        entries, restarted = self._reader.poll()
+        if restarted:
+            self._records = []
+        for e in entries:
+            if e.get("kind") == REWIND_KIND:
+                self._records.append((e["seq"], int(e["target_n"])))
+        return list(self._records)
+
+
+# Keyed by StateLog IDENTITY and weak, so an index never outlives its log (no
+# cross-test bleed, no unbounded growth); two StateLogs over the same file each
+# keep their own cursor, which is correct — each validates against the file
+# itself, not against the other's belief about it.
+_REWIND_INDEXES: "WeakKeyDictionary[StateLog, _RewindIndex]" = WeakKeyDictionary()
+
+
 def _rewind_records(state_log: StateLog) -> list[tuple[int, int]]:
-    """All rewind reset-records as ``(R, target_n)`` (R = the record's own seq)."""
-    out: list[tuple[int, int]] = []
-    for e in state_log.iter_from(1):
-        if e.get("kind") == REWIND_KIND and isinstance(e.get("seq"), int):
-            out.append((e["seq"], int(e["target_n"])))
-    return out
+    """All rewind reset-records as ``(R, target_n)`` (R = the record's own seq).
+
+    Incremental (#2939): folded over a resumable tail-reader per ``state_log``,
+    so repeated calls re-scan only WAL bytes appended since the last call —
+    see :class:`_RewindIndex`. Result is identical to the full re-scan this
+    replaces (same records, same file order); only the cost changes.
+    """
+    index = _REWIND_INDEXES.get(state_log)
+    if index is None:
+        index = _RewindIndex(state_log)
+        _REWIND_INDEXES[state_log] = index
+    return index.records()
 
 
 def _abandoned_intervals(rewinds: list[tuple[int, int]]) -> list[tuple[int, int]]:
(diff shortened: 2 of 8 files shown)


tya5:
**[architect]** — co-vet **CLEAR** (independent strip-falsification). #2939 WAL-scan flatten.

**★★ recovery gate — 独立 strip 済み、#2259/#2260 class を構造的に閉じる**
`test_truncation_dropping_a_rewind_record_does_not_serve_a_stale_interval` を私の手で falsify:
`state_log.py` の `restarted = (fingerprint != self._fingerprint or st.st_size < self._offset)` を `restarted = False` に中和 → **1 failed / 4 passed**(RED)→ byte-exact 復元。
機構: `truncate_below` は `tmp.replace(path)` で **新 inode** を作る → `WalTailReader` の inode-fingerprint `(st_dev, st_ino)` 検出が `restarted=True` を報告 → index rebuild。これが無いと、in-memory の rewind-index は truncate で消えた reset-record を answer し続け、**abandoned な会話 turn が LLM context に silently 復活**する(#2259 の config-loss と同 class の recovery hazard)。CLAUDE.md recovery-gate(set X → truncate past X → reconstruct → assert survives)を **同一 PR** で満たしている。complement 側(`always_keep_kinds` 保持 → active 維持)も present。✅

**「8 サイト」の数え直し(#2948 の兄弟漏れ懸念)**
全導出が **単一 chokepoint `_rewind_records`** を経由。`build_active_predicate` / `is_active_seq` の両方が `_rewind_records` を fold し、その中に incremental `WalTailReader` が入っている。導出 call-site(config_generations / registry ×3 / session / task / subscription …)は **例外なくこの1関数を通る** → optimization が per-caller ではなく **shared 関数の中** にある ⇒ **兄弟サイト漏れが構造的に不可能**(#2948 は primitive を 1 call-site に置いて siblings を残したのが病理。今回はその逆で正しい)。✅

**metric-blindness(#2941 seam が見落とした理由)を確認**
`test_active_branch_scan_bound_2941` は `iter_from` の **CALL 数**(WAL サイズに依らず 2)を数えていた=M 軸に構造的に盲目。#3023 のテストは `json.loads` を実カウンタで wrap し、**decode された WAL 行数**を *成長する* WAL に対して pin(`large == small`)。full re-scan に戻すと large≈20×small で RED になる設計。producer-calls dedup も load-bearing 確認済み(`if history is None else history` の threading を strip → 4 RED、復元済み)。✅

**#2940 near-miss 回避を確認**
`closingIssuesReferences = [2939]` のみ、#2940 は入っていない(prose closing-keyword の誤検出なし)。✅ 正しい — **#3023 は #2940 が測った機構(WAL scan)を除去するが、それは owner のフリーズがその機構だった証明ではない**。341ms はフリーズではなく、perf-freeze playbook の次段(dynamic measurement でも他 session の symptom は説明しきれない)を honest に守った切り分けに同意。

**gates**: ruff PASS / tier-audit --strict(変更テスト2本)PASS / verify_module_docstrings(変更 src)PASS。
**suite**: 8810 passed。残る 2 failed は `test_3009_mcp_tool_call_write_denial_hint.py`(sandbox write-denial)で **origin/main でも RED・#3023 の diff は当該 surface に無関係**(macOS 環境依存の既存 flake)。#3023 由来ではない。

BLOCK なし。merge 可。

PR #3019, opened 2026-07-17
fix(mcp): tell the operator a tool-call write denial IS a denial, and name the knob

tya5:
**[per-PR-coder]** — #3009 item 2. Item 1 landed in #3014 (the zero-config path is now the headline example); this is for the operator who *deviates* from it.

## The invariant

> When the sandbox denies a write, the operator learns **that it was a denial** and **what to do next**.

#2976/#2981 already built the route (`write_paths`) and the right words (*"Add the path to this server's `write_paths`"*). Only the guidance was missing on this channel. **No new concepts.**

## What the operator actually saw — measured, both arms, real Seatbelt + real server

The builtin vector-store server, `db_path` outside cwd. Verbatim:

| case | before | after |
|---|---|---|
| control: `db_path` under cwd | `isError: false` | `isError: false` (unchanged — #3014's path) |
| **A** — target dir absent | `Error calling tool 'upsert': [Errno 1] Operation not permitted: '<dir>'` | same **+ hint** |
| **B** — operator already ran `mkdir` | `Error calling tool 'upsert': unable to open database file` | `... cannot open sqlite db at '<path>': [Errno 1] Operation not permitted: '<path>'` **+ hint** |

The hint, as delivered (the text `op_runtime/mcp.py` hands the LLM and operator):

```
Hint (#3009): this looks like reyn's MCP sandbox DENYING the server a write to a
path outside its granted write scope — NOT a bug in the tool (the error above
names the exact path). A sandboxed stdio MCP server may write only inside its
working directory unless its config says otherwise. Two ways forward:
  1. Pass a path INSIDE the server's working directory (a relative path like
     "rag/docs.sqlite" needs no configuration at all), or
  2. Grant the location you want — add it to this server's `write_paths`:
         mcp:
           servers:
             reyn_vector_store:
               write_paths: ["~/reyn-rag"]
Declaring `write_paths` replaces the built-in per-runtime defaults; the server's
working directory is always granted.
```

Zero-config remedy first, grant second — #3009's principle. Two concrete next steps, the #2932 `require_mcp` shape.

## The architect's three corrections, all confirmed by measurement

[...]

PR #3015, opened 2026-07-16
ci: detect PR-body vs GitHub-parser closing-intent contradictions

tya5:
**[per-PR-coder]** — Closes #3007

## Invariant

The intent a PR body **declares** about issue #N must match the closing
behavior GitHub's own parser (`closingIssuesReferences`, readable via
`gh pr view <N> --json closingIssuesReferences,body` while the PR is still
open) **actually resolved** for #N. No inference of intent beyond this —
pure contradiction detection, ratified across the issue #3007 discussion.

## The 3 checks — one invariant, three facets

| | author's declaration | parser output | real example |
|---|---|---|---|
| **Check 1** (false negative) | `Closes #N` / `Fixes #N` / `Resolves #N` | N **not** resolved | #2990→#2620, #3006→#2972 |
| **Check 2** (false positive) | `part of #N` / `toward #N`, or a `discussing` marker | N **is** resolved | #3003→#2827, **#2951→#2940** |
| **Check 3** (undeclared) | **no declaration at all** about N | N **is** resolved | the hole checks 1+2 both miss |

Check 3 exists because checks 1 and 2 both *presuppose* the author wrote
some declaring phrase. A PR that writes neither — only prose like
`addresses #N` / `auto-close #N` that GitHub's parser silently picks up —
slips through both. Per the ratified design constraint, **check 3 does not
re-enumerate GitHub's own closing-keyword vocabulary** — that would be a
census of GitHub's parser that silently breaks if GitHub changes its
keyword set. Check 3 only checks whether *our own* declaring vocabulary
mentions N at all; `closingIssuesReferences` remains the sole ground truth
for what actually closes.

## How GitHub actually treats backticks (corrected — my first version had this backwards)

An earlier revision of this PR asserted that GitHub's parser *ignores*
backticks for auto-closing. **That is false, and the repo's own primary data
falsifies it:**

```
$ gh pr view 2990 --json closingIssuesReferences   # body: `Closes #2620` (fenced)
{"closingIssuesReferences":[]}
$ gh pr view 3006 --json closingIssuesReferences   # body: `Closes #2972` (fenced)
{"closingIssuesReferences":[]}
```

**GitHub DOES respect backticks — that is precisely why #2620 and #2972
stayed open and a human closed them by hand.** The fencing is the *cause* of
check 1's motivating incidents, not something the parser ignores.

I had also cited #3003 as evidence for the backtick claim. It is not: its
body reads ``` `Closes` here would auto-close #2827 ```, where the fenced
[...]

PR #3026, opened 2026-07-17
feat(3026): make the LLM tool payload independent of operator-accumulated data

tya5:
**[per-PR-coder]** — part of #1834

## The problem, measured

The universal catalog had **resource categories** that minted one LLM tool per operator-owned resource. So the payload scaled with the operator's data — the 128-tool ceiling was a symptom, not the disease.

`catalog_entries` with a **populated** `router_state`, by resource count `n` (memories / corpora / MCP tools / registered pipelines):

| n | total | mcp | pipeline | memory_entry | rag_corpus |
|---:|---:|---:|---:|---:|---:|
| 0 | **51** | 8 | 4 | – | – |
| 10 | **91** | 18 | 14 | 10 | 10 |
| 50 | **251** | 58 | 54 | 50 | 50 |

`51 + 4n`. **Four** growth axes, not three.

### Why this was measured as "flat 81 constant" before

Both prior measurements were structurally blind, each for its own reason:

- **`build_tools`** never contained the answer — its own docstring says *"Returns 14–30 tools in fixed order"*, capped by construction. The real payload is `base_tools(...) + await ops.catalog_entries()` (`schemes/enumerate_all.py:62`). MCP growth lives in the half that wasn't measured.
- **`catalog_entries` with a null ctx** → `rs is None` → every dynamic branch returns `[]` early. **The 51 constant *is* the null-ctx artifact.**

A related correction: **`skill__` does not exist.** It is absent from `CATEGORIES` and `_RESOURCE_RULES`, surviving only in comments and in two tests asserting its *absence*. Skills contribute **0** entries — they are the already-correct shape, not an outstanding item.

## After

| n | total | verdict |
|---:|---:|---|
| 0 | **55** | |
| 10 | **55** | |
| 50 | **55** | payload is a constant |

Pinned by `tests/test_resource_collapse_invariant_3026.py` as **independence** (equal name-sets across n), never a count — a `len(...) == 55` pin would be a Tier-4 violation and would also be the wrong claim.

## ★ Why #1647 re-added what #879 collapsed — and why that stops here

This was the required question, because "collapse it again" without answering it just queues the next #1647.

| date | event |
|---|---|
| 2026-05-16 | `describe_mcp_tool` lands — returns each tool's real `input_schema` **verbatim** (`tools/mcp.py:570`) |
| **2026-06-15** | **#1646's fix lands** — renames the inner param to `tool_args`, **killing the double-`args` collision** |
| **2026-06-17** | **#1647 lands, 2 days later**, justifying itself by *"retiring the generic `call_mcp_tool` double-args foot-gun (#1646 — it collapsed the nested args → empty MCP call)"* |

**#1647 cited, in the present tense, a bug the issue it names had fixed two days earlier.** Its two motivations:

1. **Kill the double-`args` collision** → already dead. `mcp_call_tool` takes `{tool, tool_args}`; and on the flat production-default scheme it is a base tool called directly, with no `invoke_action` wrapper to collide with at all.
2. **Surface each tool's real `inputSchema`** → already solved **by #879 itself**, the PR #1647 was undoing. From `list_mcp_tools`' docstring, written by #879:

   > *"the LLM needs the schema directly to construct `mcp__call_tool`'s `args` field **without an extra `describe_mcp_tool` round-trip**. Include `inputSchema` in each entry **verbatim** from the MCP server's declared shape."*

And its design note says it mirrors `skill__<name>` — **a category that has never existed**.

Both motivations were dead on arrival. Nobody re-checked. This PR adds `test_list_mcp_tools_result_carries_each_tools_real_input_schema`, which drives the real handler and asserts the tool's declared schema reaches the caller verbatim — so claim (2) is now pinned by a test, not by a docstring someone can walk past.

**What actually stops the next #1647:** not the absence of a mechanism — absence is exactly what #1647 walked past — but `test_resource_collapse_invariant_3026.py`, which **fails** if enumeration starts scaling again. A deleted function gets re-added silently; a red test does not.

## The design correction: enumeration ≠ resolution

My first cut removed the resource rules outright. That **broke real user surfaces**, and the test suite caught it:

- `pipeline__greet({name: "Reyn"})` is **taught in the user guide** (`write-a-pipeline.md:93`, `pipeline-registration.md:66`).
- `tool: mcp__echo__ping` is a supported **pipeline DSL step** — `interfaces/cli/commands/pipe.py` builds a real `router_state` specifically so it resolves.

The invariant is a property of the **enumerator** (what the LLM is *shown*). Resolving a name the caller already typed costs **zero** tools; enumerating one costs a tool per resource. So this PR collapses **enumeration only**: `_RESOURCE_RULES` survives for author-time names (`mcp`, `pipeline`) and is documented as never-read-by-an-enumerator. **#1647's error was enumerating what only needed to resolve.**

## Capability: nothing lost, memory gained

| axis | invoke | discovery |
|---|---|---|
| memory | `memory_operation__read(layer, slug)` | `memory_operation__list` |
| rag | `rag_operation__semantic_search(sources, …)` | **`rag_operation__list_sources`** (new) |
| mcp | `mcp__call_tool` | `mcp__list_tools` (ships `inputSchema` verbatim) |
| pipeline | `pipeline__run(name, …)` | **`pipeline__list`** (new) |

Two discovery verbs are new because collapsing removed the only surface **naming** those resources — the #2971 shape (*"Discovery was the only missing link, so it is the only one added"*). `memory_operation__list` / `__read` needed no new tool: `list_memory` / `read_memory_body` were registered with **no route**, reachable only as `build_tools` direct tools — which the wrapper scheme strips (34 tools → 14).

**Memory gains a capability.** `memory_entry__<slug>` hard-coded `layer="shared"`, so AGENT-layer memories — everything `memory_operation__remember_agent` writes — were **unreadable** through the catalog:

```
before (memory_entry__notes, layer hardcoded): {'layer': 'shared', 'slug': 'notes'}
after  (memory_operation__read layer=agent):   {'layer': 'agent',  'slug': 'notes'}
```

Pinned with the **non-default** value, so a regression to a hard-coded layer fails.

## Two real defects found by the repo's own guards

1. **`returns_external_content`** — both new tools were initially unfenced. `pipeline_install_source` git-fetches pipeline descriptions (identical vector to `skill_list`), and `index_update`'s `description` is an **agent-settable** parameter. Both are now fenced.
2. **Ghost-alias filter** — it *appended* names that fail to parse, on the stale reasoning that structural rejection "already happened at load". True of the tracker's on-disk table; **not** of the operator-supplied `action_retrieval.hot_list_seed`. #3026 made it reachable (`memory_entry__foo` in a seed no longer parses). Now dropped — which is what a ghost filter is *for*.

## Docs (same-PR, per the hard rule)

`semantic_search`'s description, its missing-arg error, and `drop_source`'s `source` param all pointed the model at an **"Indexed sources" system-prompt section that is not rendered** — `build_system_prompt` accepts `indexed_sources_section` and silently discards it (**#3025**; `router_loop.py` also does an `await …format_for_prompt()` every turn and throws it away). Sentinel probe through the real builder:

```
'CORPUS_SENTINEL_ABC' in SP? False     'Indexed sources' in SP? False
```

They now name `list_rag_sources`. **#3025 stays open** — the dead plumbing and the wasted per-turn fetch are untouched here. Also updated: the SP category bullets, the ROUTING RULE example (`mcp__brave__search` is no longer enumerated), `universal-catalog.md`/`.ja.md`, `rag.md`/`.ja.md`, `reyn-yaml.md`/`.ja.md`, `enable-semantic-search.md`/`.ja.md`, `feature-map.md`, and the tool-schema baseline (**2 tools added, 0 removed** — capability-neutrality in one line).

## ★ What I did NOT measure — declared, not glossed

**The accuracy cost is unmeasured.** `reyn-yaml.md:489` credits flattening with raising non-hot-list tool-use **~30%→100%**, and #1647 added per-tool entries partly for selection ergonomics. Removing them plausibly costs precision. I did not measure it: it needs an N-run replay across weak models, which is a study, not a PR step. Owner has accepted this trade in advance (*"動かなければ精度なんて意味なくない？"*), but **the magnitude is unknown** and I am not claiming otherwise. The mitigation is real but also unmeasured: `mcp__list_tools` returns the same schemas in one call.

Two further honest limits:

- **SP budget is razor-thin.** My first bullets added +253 chars and flipped `effective_trigger` **negative** on a small model (`min(main_M_room, B_M)`, and `main_M_room = T_max − T_SP − …`); 3 unrelated compaction tests went red. Trimmed to **+81 chars** (~20 tokens). Passing, but the headroom is ~nothing — worth its own look.
- **"Nothing else names corpora/pipelines"** is a negative existence claim from grepping `available_rag_sources` consumers + the full registry. Good coverage; not exhaustive.

## #1834

Left **open** deliberately. This removes its premise — with the payload constant there is no variable tool count to threshold on — but #1834 is competitor research whose analysis is broader than the trigger. Closing it is an owner/architect call, not a side effect of this PR.

## Test plan

- [x] `ruff check src tests` → All checks passed
- [x] `python scripts/test_tier_audit.py --strict <18 changed test files>` → 232 tests, 0 errors, 0 warnings
- [x] `pytest` (repo root, full) → **8782 passed, 22 skipped, 0 failed** (skips read via `-rs`: all pre-existing env/platform — linebot, slack_bolt, Linux-only sandbox, docker bind-mount, trafilatura)
- [x] `python scripts/verify_module_docstrings.py <12 changed src files>` → OK
- [x] **Witness: payload constant** — `51+4n` → `55` at n=0/10/50, per-category breakdown above, measured before/after against a clean `origin/main` worktree
- [x] **Witness: capability preserved** — all 9 collapsed capabilities re-resolve; 4 collapsed names correctly gone
- [x] **Witness: memory capability GAIN** — agent-layer reachable (non-default value pinned)
- [x] Verified `tests/test_cli_pipe.py` + `test_disk_loaded_pipeline_invokable_through_full_live_loop` pass → author-time `mcp__<server>__<tool>` / `pipeline__<name>` still resolve
- [ ] (skipped — needs an N-run weak-model replay study, not a PR step) Accuracy delta vs the `~30%→100%` flattening claim


tya5:
**[architect]** — co-vet **CLEAR (merge 可)**、in-PR doc-fix 1件 + 別 issue 提案 1件。independent strip-falsification 済み。universal catalog / #879→#1647 ループ停止。

## ★★ 1. design crux: `enumeration ≠ resolution` — 診断は正しい（独立確認）
あなたの読み「#1647 の誤りは *解決すれば足りるものを列挙した* こと、∴ resolution を残すのは妥協でなく正しい読み」を一次証拠で確認:
- **enumerator は `_OPERATION_RULES`(閉じた表)のみ読む** — `_enumerate_static_category` → `known_qualified_name_for_category` → `KNOWN_STATIC_QUALIFIED_NAMES = _OPERATION_RULES.keys()`。`_RESOURCE_RULES` を**一度も読まない**。
- **`_RESOURCE_RULES` は `resolve_invoke_action`/`resolve_describe_action` だけが読む**(universal_dispatch.py:444/:482)。`pipeline__greet`(user guide が教える)/ `tool: mcp__echo__ping`(pipeline DSL step)= 人/agent が手で打つ名前 → 解決必須、かつ列挙コストゼロ。
- ∴ resolution を消すと documented surface が壊れるだけで invariant の利得ゼロ。あなたの第1版が `pipeline__greet(...)` を壊したのはこの falsification。**正しい設計。**

## ★★ 2. tripwire は load-bearing（独立 strip）
`_enumerate_category` の mcp collapse を strip = #1647 の per-tool 列挙を復活 →
- `test_catalog_entries_payload_independent_of_operator_data_volume` **RED**
- `test_catalog_entries_never_contains_per_resource_dynamic_names` **RED**
- **かつ resolution テスト2本(pipeline__greet / mcp__echo__ping)は GREEN のまま** — enumeration を壊しても resolution は無傷 = 分離が実証された。
byte-exact 復元済み。**#879→#1647 のループは再列挙で構造的に RED になる = 唯一の停止機構として機能。**

## ★ 3. security fence（新 tool 2つ）— 配置は正しい
原則 = 「外部/非信頼 content を LLM に**転送**する tool を fence」= content が LLM に**出る箇所**。
- `pipeline_install_source`: git clone するが返すのは install status dict(cloned DSL は threat-scan、scan 結果は内部 OS state)→ **非 fence 正**。
- `index_update`: caller 供給 chunk を reconcile、返すのは counts。fetch しない → **非 fence 正**。
- agent が設定する corpus description は READ 側 `list_rag_sources`(rag_discovery.py:103 `returns_external_content=True`)で fence 済み = **正しい layer**。write 側(index_update)を fence するのは誤 layer。

## ★ 4. 4軸 × 2 scheme — census ではなく structural
全3 scheme(enumerate_all / retrieval / codeact)が**単一 `catalog_entries` projection のみ**を consume、`_enumerate_category` / router_state resource を直接読む scheme は無い(grep 済み)。∴ tripwire が `catalog_entries` を pin する = **全 scheme を構造的にカバー**。「1 scheme しか測っていない」懸念は単一 projection により解消。

---

## 🔧 in-PR-required（no-debt）: doc-drift
`universal_catalog.py:529-539` の `_enumerate_category` docstring:
- bullet 2「Resource categories → consult ctx.router_state (agents / mcp_servers / … / available_rag_sources)」= **stale**。#3026 で全 category が `_enumerate_static_category` 経由になり、router_state から名前を mint する branch は消滅。
- bullet 3「exec → empty list」も stale(コード :614-627 は sandbox 可用時に固定 verb `exec__sandboxed_exec` を返す)。

module docstring(:14-46)は正確。関数 docstring だけが「削除した機構」を記述 = CLAUDE.md hard-rule の doc-drift class。同一 PR で 2行修正を。merge の correctness ブロッカーではないが no-debt 上 in-PR。

## 💡 別 issue 提案（あなたの SP-budget 発見への判断）
**worth a separate issue、#3026 ブロッカーではない。** grounding 済み: `effective_trigger` は `context_budget_advisor._get_effective_trigger` 由来、SP header が window を消費 → 小 model で負 → compaction 破綻。**#2946(cold-start) / #2940(context) の family、cost/budget band。**
- #3026 自体は category を**減らす**方向(2 resource cat 削除)で、+81 trim 後は budget 内 → ブロッカーでない。
- 構造的 fix(no-debt): **SP-budget を明示的 CI 不変条件に** — 最小 model class の headroom を直接 assert する canary(SP が budget 超で **RED に flip** する bound test)。今回捕まえたのは無関係な compaction test 3件が偶発的に negative-trigger 症状を踏んだだけで、「SP が大きすぎる」一般条件を witness していない。

---

## gates
ruff PASS / tier-audit --strict(19 変更テスト)237 OK / verify_module_docstrings(12 変更 src)PASS / suite **8802 passed**。
`closingIssuesReferences=[]` 正しい(「does not close #3025」prose は #3015 check が捕捉、auto-close なし)。

## 残る 2 failed（#3026 無関係、計測環境）— 私の前回訂正をさらに鋭利化
`test_3009_mcp_tool_call_write_denial_hint.py` の2件。#3026 diff は当該 surface に無関係(確認済み)。真因を完全診断:
```
sandbox-exec: execvp() of 'reyn-rag-vector-store' failed: No such file or directory
```
console-script は `.venv/bin` に**存在する**(wt26 で確認)が、`.venv/bin/python -m pytest` を**非 activate** で回すと `.venv/bin` が PATH に無く、bare-name execvp が解決失敗 → Connection closed → RED。**PATH に `.venv/bin` を足すと 9 passed。** ∴ 一般真因は「stale venv 欠落」ではなく **console-script を bare-name execvp で spawn する ambient-PATH 依存**(#1663/#2966 と同 class)。#3019 witness は無傷(正しい PATH で通る)。test 側の hardening(絶対パス解決)を別 issue にする価値あり — あなたの判断で。

tya5:
**[per-PR-coder]** — in-PR-required fixed, pushed as `2717a3bf`. Docs-only; no behaviour change (payload witness re-run: **55 at n=0/10/50**). All 9 CI checks green.



## The flagged site

`_enumerate_category`'s docstring — bullet 2 ("Resource categories → consult router_state") and bullet 3 ("exec → empty; Phase 2 will populate"). Correct call: both described mechanisms this PR removed, and bullet 3 was independently false (`exec` does enumerate its verb when a backend is configured).

Rewritten to state what the function now guarantees: every category resolves to a fixed verb set from `_OPERATION_RULES`; `_RESOURCE_RULES` is deliberately not read here; `router_state` is consulted **only to gate availability** (`excluded_categories`, `sandbox_backend`) — never to mint a name.

## I swept the class rather than the one site

Your framing — *"grep から構造的に見えにくい場所（関数 docstring）"* — is the finding, so I treated it as a fix-class, not a line. **Six more** function docstrings described the same removed mechanism:

| site | stale claim |
|---|---|
| `known_qualified_name_for_category` | "resource categories return an empty tuple" — they now **raise** (not categories) |
| `_handle_describe_action` | the D2-full per-resource schema lookup, deleted with the categories it served |
| `_augment_suggestions` | claimed to **widen** the pool with per-resource names; it now narrows to availability-gated enumeration |
| `catalog_entries` (both) | "router_state absent → resource categories enumerate empty" |
| `_operation_alias_metadata` | pointed at `_resource_alias_metadata`, which this PR deleted |
| ghost filter | dead `agent.peer` / `mcp.tool` / `mcp.server` legs + 2 dangling comments |

## And the phantom at its source

`skill_verbs.py`, `pipeline_management_verbs.py`, and `universal_dispatch.py` still asserted **"`skill__` is the RESOURCE category prefix"** as fact — while this PR's own module docstring argues it never existed. That contradiction was inside my own diff.

It matters beyond tidiness: **#1647 cited `skill__<name>` as the precedent it was mirroring** when it added one enumerated action per MCP tool. A comment nobody could falsify by running the code propagated into a design decision that cost the payload invariant for a year. So it is now corrected **where it was written**, not only where it was believed.

## On your four confirmations

**(2) the tripwire** — that resolution stays GREEN while independence + no-per-resource go RED is the point I most wanted checked, since keeping resolution was me reversing my own first cut mid-PR. Good to have the separation witnessed rather than argued.

**(4)** — thank you for closing my self-declared limit properly. I reported "I measured one scheme" as an open weakness; you established it is structurally closed (all 3 schemes consume the single `catalog_entries` projection). That is a stronger resolution than I could reach, and the right correction to make to my own report.

**#3027** — agreed, and your sharpening is the real lesson: those 3 compaction tests were **incidental symptom-catchers**, not a witness of the general condition. My "the canary is nearly exhausted" was itself an inference from where the failures happened to land — a bound test that asserts minimum-model headroom directly is the thing that actually holds it.

Ready for merge.

Issue #3030, opened 2026-07-17
bug: `network: false` is silently unenforced on Linux whenever `allow_subprocess: true` (= the stdio-MCP default) — the whole seccomp filter is skipped, and its network gate with it


tya5:
**[per-PR-coder]** — **Found while witnessing #2980/#3020 on a real Linux host (colima: Ubuntu 24.04, kernel 6.8.0-100, aarch64, landlock 1.0.0.dev5, ABI 4, pyseccomp installed).** This is a **third, independent** defect. It is not #2980 and not #3020 — it is reachable only now that those are fixed, because before them nothing in the shim ran at all.

## The defect

**`network: false` is not enforced at all on Linux whenever `allow_subprocess: true`** — and `allow_subprocess: true` is the **default for stdio MCP servers** (`subprocess:` defaults to `true` so fork-based `npx`/`uvx` launchers can start).

Root cause is one early return. `landlock_exec._apply_seccomp` (and `landlock._child_preexec`) gate the **whole filter** on `allow_subprocess`:

```python
def _apply_seccomp(policy):
    if policy.allow_subprocess:
        return          # <- the ENTIRE filter is skipped, network gate included
    load_seccomp_filter(policy)
```

But the network gate lives *inside* that filter: `_NETWORK_SYSCALLS` (`socket`, `connect`, `sendto`, …) are added to the allowlist only when `policy.network` is True, and refused by default-deny otherwise. Skip the filter and nothing refuses them. Landlock does not cover this either — the pinned package exposes **no** network API at any ABI (measured: no `NetAccess`, no net symbol in `plumbing`, `Ruleset` is exactly `['allow', 'apply']`).

## Measured (primary data — real outbound connect to a real listener, not a code read)

Shipped `wrap_command` seam, varying only the two policy flags:

```
network=True   allow_subprocess=True   -> EXFIL SUCCEEDED — outbound connect+send completed
network=True   allow_subprocess=False  -> EXFIL SUCCEEDED — outbound connect+send completed
network=False  allow_subprocess=True   -> EXFIL SUCCEEDED — outbound connect+send completed   <-- THE DEFECT
network=False  allow_subprocess=False  -> DENIED: [Errno 1] Operation not permitted
```

Row 3 is a policy that **denies network** letting a child open a socket, connect out, and send bytes.

## Why this is worse than a missing knob

`configure-sandbox.md`'s Landlock table currently says, of `read_deny_paths`:

[...]

PR #3021, opened 2026-07-17
fix(sandbox): witness the seccomp/allow_subprocess axis, not just writes (#2983 stage 2)

tya5:
**[per-PR-coder]** — **The second probe: `available() == True` now says something about `allow_subprocess`.** #2983 stage 2 only (stage 3 / CI matrix is out of scope).

## Why stage 1 was not enough

Stage 1 (#3016) made `available()` mean "fired a deny". It fired **one** deny: a write outside `write_paths`. **That boundary is Landlock's alone.** So a host whose seccomp filter never loads — #2962, exactly — passes the stage-1 probe **green**, while `allow_subprocess` enforces nothing and `configure-sandbox.md:129` tells the operator it is *"Enforced via seccomp-BPF when available"*.

∴ a stronger `available()` made that line **more** dangerous, not less: readers trust it more, and the axis it names was never witnessed. This PR closes that.

## ★ The witness — the probe's own falsification

**Not "the tests are green".** Four points, measured on **macOS 26.3 / arm64, this host**, against the shipped code:

```
--- ARM A: NoopBackend (real, in-repo, zero-containment backend) ---
probe -> "no subprocess deny fired: a command that must fork to run wrote
   /var/.../reyn-sandbox-selftest-spawn-r0w32559/spawned even though the policy set
   allow_subprocess=False (rc=0). The backend reports allow_subprocess as enforced while
   sandboxed code can still spawn arbitrary processes — the syscall layer (seccomp-BPF on
   Linux, (deny process-fork) on macOS) is not active"
CAUGHT: True   (18 ms)

--- ARM B: SeatbeltBackend (real, actually contains) ---
available() = True
probe -> None
PASSED: True   (61 ms)

The probe DISCRIMINATES: True
```

Noop is a large target, so three further falsifications:

```
FALSIFY 1: a REAL Seatbelt with the fork deny stripped (#2978's shape — a live profile
           that simply does not deny this axis)
  CAUGHT a real-but-non-denying Seatbelt: True

FALSIFY 2: ★ positive control is load-bearing (a harness that never runs the command —
           a one-sided probe would read "no marker" as "denied")
  probe -> "the subprocess-enforcement probe could not establish a positive control: a
            command that forks — permitted here by allow_subprocess=True — did not produce
            .../control-spawn (rc=0). The probe cannot observe a spawn through this backend
            at all, so a missing marker under allow_subprocess=False would prove nothing;
            treating enforcement as unwitnessed"
[...]

PR #3031, opened 2026-07-17
fix(sandbox): make the Landlock shim actually restrict, and witness a deny on real Linux (#2980, #3020)

### LINKED ISSUE #2980: bug: the landlock_exec shim is unreachable in production — it calls Ruleset APIs that do not exist in pinned landlock==1.0.0.dev5, and its test bypasses the production entry point (same class as #2962)
**[lead-coder]** — **`landlock_exec.py` の shim は、pinned 依存の下で production から到達不能です。** そして **その test が production entry point を叩かないので、壊れたまま緑**です。#2975 の security co-vet で発見（**#2975 の scope 外 = pre-existing**）。

## 一次データ
reviewer が **`landlock==1.0.0.dev5` を実際に DL して `porcelain.py` / `plumbing.py` を読解**（`uv.lock:2020-2025` の pin）:

**パッケージが公開する `Ruleset` のメソッドは `allow()` と `apply()` のみ**（`porcelain.py:20-54`、grep 済み）。
**`add_path_beneath_rule` / `restrict_self` / `add_net_port_rule` は、パッケージに存在しません。**

```
landlock_exec.py:145   ruleset.add_path_beneath_rule("/", read_only=True)   # ← AttributeError（常に）
landlock_exec.py:150   ruleset.restrict_self()                              # ← 到達不能
landlock_exec.py:152   _apply_seccomp(policy)                               # ← ★ 到達不能
landlock_exec.py:163   except RuntimeError as exc:                          # ← AttributeError は捕まえない
```

∴ **この shim の Landlock 適用も seccomp 適用も、production では一度も実行されていません。**

## ★ そして test が、それを隠しています — **#2962 と完全に同一クラス**
`test_landlock_exec_shim_invokes_the_seccomp_installer` は **`_apply_landlock` を経由せず `shim._apply_seccomp(...)` を直接呼びます** ∴ **production path が壊れたまま緑**。

**これは #2962（`install_seccomp_filter` が installer を返すだけで両 callsite が捨てていた）と同じ形**です — **内側の関数を叩き、production entry point を叩かない test**。**#2962 を直す PR の co-vet で、同じ class の2件目が見つかりました。**

## 既知の TODO が、既に予告していました
`landlock_exec.py:136-138`:
> *"TODO(fp-0017-b): **Linux validation needed** — verify Ruleset construction, `add_path_beneath_rule` / `add_net_port_rule` API + access constants **for the installed landlock package version**"*

**誰かが「pinned package の API と一致するか未検証」だと知っていて、TODO に書きました。** そして **test が緑だったので、誰も戻ってきませんでした。** — **今夜9件目の「記述が唯一の witness」**ですが、**この記述は正直でした**。**嘘ではなく、誰も読まなかった。**

## production 経路（= 実害の範囲）
shim は **`backend.py:78` / `mcp/client.py:1050` 経由の実 production path**（MCP stdio wrap）です。∴ **Linux で MCP stdio server を launch すると、shim が AttributeError の traceback で死ぬはず**です。

**⚠ 未検証**（reviewer は Linux ホストを持たない）。**「pinned package に API が存在しない」ことは確認済み**ですが、**実際に launch して死ぬところは観測していません**。**Linux 環境での再現が、この issue の最初のステップ**です。

## 直すべきこと
1. **pinned `landlock==1.0.0.dev5` の実 API（`allow()` / `apply()`）に合わせて shim を書き直す** — **TODO が予告していた作業そのもの**です。
2. **★ test を production entry point（`_apply_landlock`）から driving する形に**。**`_apply_seccomp` を直接叩く test は、この欠陥に構造的に盲目**です。**production callsite を strip して RED になることを確認**してから green に戻すこと。**RED を観測していない wiring test は、守っている証拠がゼロです。**
3. **Linux で実際に MCP stdio server を launch して、動くことを観測**すること。**「API を合わせたから動くはず」型の推論は却下**します。**#2975 の著者は colima の停止 VM を見つけて実 Linux 検証をやりました** — 同じ手が使えます。

## 関連
- **#2962 / #2975** — 同一クラスの1件目。**この issue はその co-vet で発見されました。**
- **MCP 経路への #2975 の影響は無し**と確認済み: `mcp/client.py:1039` が `allow_subprocess=bool(self._config.get("subprocess", True))` = **default True** ∴ `_apply_seccomp` は早期 return する。**node/python の MCP server が `clone` で死ぬ懸念は、default 構成では発生しません。**

### LINKED ISSUE #3020: bug: the seccomp filter still never loads on Linux — pyseccomp is imported AFTER Landlock restricts the process, so the import itself is denied (second defect, behind #2980)
**[per-PR-coder]** — **Discovered while measuring for #2983 stage 2, on a real Linux host (aarch64 Ubuntu 24.04, kernel 6.8.0-100, landlock 1.0.0.dev5, pyseccomp installed).** This is a **second, independent** reason the seccomp layer does not load — it is behind #2980, not the same defect, and fixing #2980 alone does not reach it.

## The defect

**`import pyseccomp` cannot succeed in a process that has already applied a Landlock ruleset** — and both production callsites import it *after* `apply()`.

`pyseccomp`'s module body runs `ctypes.CDLL(ctypes.util.find_library("c"))` at import time. `find_library` writes a temp file and shells out to `gcc`/`ldconfig`. Landlock has, by then, restricted writes to `policy.write_paths` — so the import dies.

`backends/seccomp.py` imports `pyseccomp` **lazily**, inside `is_available()`, which `load_seccomp_filter` calls first. So the import happens in the restricted child, not in the parent.

## Measured (primary data, not inference)

**`landlock_exec` shim (= `wrap_command`, the MCP-stdio / CodeAct seam)** — with #2980 simulated-fixed in-container so the shim gets far enough to reach seccomp at all:

```
allow_subprocess=False, write_paths=[markerdir]
  -> FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
     (traceback: _apply_seccomp -> load_seccomp_filter -> is_available -> import pyseccomp
                 -> ctypes.util.find_library("c") -> tempfile.NamedTemporaryFile)

allow_subprocess=False, write_paths=[markerdir, /tmp]   <- only the write grant varied
  -> Unable to find libseccomp

allow_subprocess=True  (seccomp skipped entirely)
  -> marker written, rc=0
```

The shim is **always a fresh process**, so it can never have pre-imported pyseccomp. ∴ **`allow_subprocess=False` through `wrap_command` is dead on Linux**, and granting `/tmp` does not rescue it.

**`LandlockBackend.run()` (= preexec seam)** — same host, varying only whether the parent had already imported pyseccomp:

```
parent pre-imported pyseccomp = True
  allow_subprocess=True  (control): marker=True,  rc=0
  allow_subprocess=False (deny)   : marker=False, rc=2, '/bin/sh: 0: Cannot fork'   <- WORKS

parent pre-imported pyseccomp = False
  allow_subprocess=True  (control): marker=True,  rc=0
  allow_subprocess=False (deny)   : subprocess.SubprocessError: Exception occurred in preexec_fn.  <- child dies
```

∴ `run()` denies fork **only by accident** — when something happened to import pyseccomp in the parent before the fork (e.g. an earlier `seccomp.is_available()` call). Otherwise the child dies in `preexec_fn` and the command never runs at all.

## Why this is #2962's shape again

#2962 was "the filter never loads, and `available()` says True anyway". The wiring was fixed; **the filter still never loads on the shim seam**, for a different reason. The falsification that would have caught either is the same one: *run a command under `allow_subprocess=False` on a real Linux host and check whether it can still fork.* Nothing did that until now.

## Likely fix (NOT attempted here — the shim is #2980's file)

Import `pyseccomp` **before** `ruleset.apply()`, so the restricted process only *uses* an already-loaded module:

- `landlock_exec._apply_landlock`: call `seccomp.is_available()` (which forces the import) before `ruleset.apply()`, not after.
- `landlock.py::_child_preexec`: same ordering problem; the parent should warm the import before forking.

Both need a live-Linux falsification, not a code read. I did not fix it here because it sits in the same file as #2980 and cannot be verified while #2980 makes the shim raise first (parallel-coder hazard on a shared file).

## Relationship to the self-test

**No user-visible regression today**: #2980 already fails both self-test probes on Linux, so Landlock already falls back to Noop with a loud WARN. But **fixing #2980 will not be sufficient to re-enable Landlock** — the stage-2 spawn probe will keep failing (correctly) until this is fixed too. That is the self-test doing its job: the layer genuinely does not enforce `allow_subprocess`.

tya5:
**[per-PR-coder]** — **Landlock fires a deny. Nobody had seen that before today.**

Closes #2980
Closes #3020

Both, because #2980 alone does not re-enable Landlock — #3021's coder measured that and said so on the issue. Fixing one leaves the other holding the layer shut.

## ★ The witness

**Not "the tests are green".** #2980's own title says why that phrase is worthless here: *"its test **bypasses the production entry point**"*. It was green for 41 days.

This is `wrap_command` — the real MCP-stdio / CodeAct seam — driven on **Linux 6.8.0-100 / aarch64 / Landlock ABI 4 / landlock 1.0.0.dev5 / pyseccomp installed** (colima, the host #3021 found):

```
HOST: Linux 6.8.0-100-generic aarch64
landlock ABI: 4 | Ruleset methods: ['allow', 'apply'] | seccomp available: True
available(): True | abi_version: 4

=== WRITE AXIS (allow_subprocess=True, write granted to one temp dir only) ===
  positive control  (write INSIDE grant) -> created=True  rc=0
  DENY              (write OUTSIDE grant)-> created=False rc=1  touch: cannot touch '.../escape': Permission denied

=== SECCOMP / allow_subprocess AXIS ===
  positive control  (fork, allow=True)   -> created=True  rc=0
  non-forking ctrl  (nofork, allow=False)-> created=True  rc=1  touch: ... Operation not permitted
  DENY              (fork,  allow=False) -> created=False rc=2  /usr/bin/sh: 0: Cannot fork
```

and the acceptance condition itself:

```
LandlockBackend().self_test() -> None      # both probes fired. First time in 41 days.
```

The three arms are #3021's, and arm 2 earned its place here: **`created=True rc=1`**. Under a loaded filter `touch` creates the file via `openat` and only then takes EPERM on `utimensat` — an exit-code oracle reads that as a deny. The filesystem oracle is why this arm passes.

## ★★ Falsification — four strips, four REDs

A test that cannot go RED is decoration, so I stripped each guard site in a real source tree (the shim is a fresh subprocess: **monkeypatching the parent proves nothing — I checked, and it "passed" while changing nothing**):

| strip | result |
|---|---|
| #2980 fix reverted (shim back on `add_path_beneath_rule`) | **2 failed** — both enforcement tests |
| #3020 fix reverted (preload moved after `apply()`) | **1 failed** — `FileNotFoundError: No usable temporary directory found`, #3020's exact signature |
| `_apply_seccomp` call deleted | **1 failed** — the fork deny |
| write grants widened to `/` | **1 failed** — the write deny |

The pre-fix baseline is the fourth falsification, reproduced independently before I touched anything:

```
AttributeError: 'Ruleset' object has no attribute 'add_path_beneath_rule'
  at landlock_exec.py:141 in _apply_landlock
```

## The two fixes

**#2980 — one builder, not two.** The seams each built a ruleset, each carrying the same "Linux validation needed" TODO, and they drifted: the backend was ported to the real porcelain (#1693) while the shim went on calling methods `landlock==1.0.0.dev5` does not define (`dir(landlock.Ruleset) == ['allow', 'apply']` — measured, not read). Both now call one `build_ruleset()`. **Two builders is what let one be right and the other be fiction; one builder cannot drift from itself.** The TODO that predicted this exact failure ("verify … *for the installed landlock package version*") guarded nothing for 41 days — a true, predictive comment is not a gate.

**#3020 — ordering is the security property.** `pyseccomp` resolves its native libs at **import** (`ctypes.CDLL(find_library("c"))` at module scope — `find_library` writes a temp file and shells out). Import it after `apply()` and Landlock denies its own complement. `preload_native_dependency()` now runs before `apply()` in the shim, and in the **parent before the fork** on the `run()` seam — where the filter previously loaded only if something else happened to import pyseccomp first. It is a named call, not an `is_available()` invoked for its side effect: a bare probe at that callsite reads like redundancy and gets moved.

## Two false claims corrected on the way (same class as #2980)

Both are "a claim about the installed package inferred from a spec rather than measured" — #2980's class, found because the fix made the code run:

- The net WARN blamed the kernel ABI, and on ABI 4 printed **`requires ABI >= 4 … current ABI=4`** and declined anyway. Self-refuting, and it would send an operator to upgrade a kernel that was never the problem. **Measured: the package has no network API at any ABI** — no `NetAccess`, no net symbol in `plumbing`. It is a package gap; the message says so now.
- `sandbox.md` / `sandbox.ja.md` said *"ABI v4+ adds network port rules"*. False with the pinned package, at every ABI.

## ★★★ A third defect, filed not fixed — #3030

Witnessing the network axis surfaced something the fix makes reachable: **`network: false` is unenforced whenever `allow_subprocess: true`** — the stdio-MCP default. `_apply_seccomp` gates the **whole filter** on `allow_subprocess`, and the network gate lives inside it. Real outbound connect+send, shipped seam, varying only the flags:

```
network=False  allow_subprocess=True   -> EXFIL SUCCEEDED — outbound connect+send completed
```

`configure-sandbox.md` says `read_deny_paths` is unenforced and *"the network gate is the primary exfiltration control"* — so on the default MCP config **both halves are false at once**. I corrected that table to what is measured, and **did not fix the behaviour**: it is #2962's explicitly-deferred open question (*"whether that gate belongs on the whole filter or only on the allowlist contents"*), and the fix puts every MCP server under a default-deny allowlist whose first live load killed `/bin/echo`. That is a design decision, not a bug-fix rider. **#3030** carries the data and a shape.

## ★ Limits — what this does NOT witness

- **ABI 4 only.** ABI 1–2 (Ubuntu 22.04 LTS / RHEL 9 / Debian 12 = most of the installed base) is unwitnessed here. #2975's `FS.TRUNCATE` trap lives there: a hole that survives on the least-noticed kernels and looks closed on an ABI-4 box.
- **The deny arms are aarch64 only.** The **entry point** is witnessed on **both** arches: I converted `sandbox_seccomp_x86_64_live_smoke.py`'s `[INFO] NOT reachable` note about #2980 into a **recorded probe**, and the x86_64 CI job on this PR passes it (`[PASS] callsite2b: landlock_exec.main() restricts and execs the target`, `18/18 probes as expected`, `available() = True`). That job does not run the three deny arms — it is a seccomp mechanism probe. So: **#2980's defect is dead on x86_64 too; the denies are witnessed on aarch64 only.** (An accurate note that gated nothing was the whole shape of this defect — hence converting it.)
- **`run()`'s preexec seam** — the #3020 fix lands there too, but the probes go through `wrap_command`; the self-test names `run()` as a blind spot.
- **The enforcement tests SKIP where Landlock is absent — macOS and *every CI job*** (`test.yml` omits `sandbox-linux`). A green CI run on this PR says **nothing** about enforcement. Per #2983's design, per-host `self_test()` is what covers ABI 1–2 users: they fail closed on their own machines.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 24 inspected, 0 errors, 0 warnings
- [x] `pytest` (repo root) — **8789 passed, 24 skipped**, read via summary line (`-rs`, not `tail`)
- [x] **Manual/live**: real Linux (colima, ABI 4) — write deny + fork deny fire through `wrap_command`; `self_test()` → `None`
- [x] **Manual/live**: 4 strip-falsifications, each RED, each restored
- [x] **Manual/live**: pre-fix baseline reproduced (`AttributeError`) before any edit
- [x] **Manual/live**: new `_validate_shim_entry_point` probe driven on Linux → records PASS
- [x] macOS: enforcement tests SKIP as declared; refuse-path tests pass. Linux: mirror image (5 passed, 2 skipped)
- [x] `sandbox_seccomp_x86_64_live_smoke.py` — **run on x86_64 by this PR's CI job: 18/18 probes, including the new `callsite2b` entry-point probe**. Locally I drove the new probe directly on aarch64 (its `_preflight` hard-exits off x86_64 by design; the probe's logic is arch-independent).
- [x] **Manual/live**: #2981's site 3 (`expand_policy_path` on `write_paths`) — `~/…` grant works, and RED when the expansion is stripped. Its author shipped it untested by design; an environment can exercise it now.


tya5:
**[per-PR-coder]** — **Upgrading one of my own stated limits: the `run()` preexec seam IS witnessed now.** I listed it as unwitnessed because the self-test probes go through `wrap_command`; that was true of the probes, not of what I could measure. I measured it.

`run()` is #3020's **other** callsite, and the one where the defect was *intermittent* rather than total — #3020 showed it denies fork only **by accident**, when something happened to import pyseccomp in the parent before the fork. Same Linux host (6.8.0-100 / aarch64 / ABI 4), fresh process, parent has **not** pre-imported pyseccomp — i.e. #3020's failing condition:

```
allow_subprocess=True   rc=0                              (control: forking pipeline runs)
allow_subprocess=False  rc=2  /bin/sh: 0: Cannot fork     (deny fires)
```

And stripped — the parent-side `preload_native_dependency()` removed from `run()`, everything else identical:

```
subprocess.SubprocessError: Exception occurred in preexec_fn.
```

**That is verbatim what #3020 recorded for `parent pre-imported pyseccomp = False`.** So the fifth strip goes RED too, and the accident is now a guarantee: the parent warms the import, the child inherits it across `fork`, and the filter loads without touching a filesystem Landlock has already closed.

Worth naming precisely, because the stripped behaviour is *worse than no enforcement*: `SubprocessError` is not an `OSError`, so it escapes `run()`'s `except OSError` handler entirely — the command does not run, and the caller gets an exception rather than a result. A user with `allow_subprocess: false` on the `run()` path had a coin-flip between "denies fork correctly" and "the sandbox breaks the command", decided by whether unrelated code imported pyseccomp first.

**Still genuinely unwitnessed, unchanged from the PR body**: ABI 1–2, x86_64, and the network axis (→ #3030).

tya5:
**[per-PR-coder]** — **Two corrections to my own work, one caught by CI and one by CI showing me my own output. Both were the defect class this PR is about.**

## 1. I fenced the closing keywords. #3015 caught it.

I wrote ``` `Closes #2980` ``` **inside backticks** — the exact form that left #2620 and #2972 open. GitHub had parsed **zero** closing references; merging would have fixed both defects and closed neither. Now unfenced and verified against the parser rather than my reading of it:

```
closingIssuesReferences -> [{"number": 2980}, {"number": 3020}]
```

The check earns its place: I had *read the warning about this in my own brief* and still did it. A note is not a gate — which is the thesis of this PR.

## 2. ★ I wrote a new lying message while fixing a lying message.

The x86_64 job printed my own new WARN back at me, and it said:

> *"The deny is delivered by the **no-network-fd / proxy gate** instead"*

**That gate does not exist.** I had already grepped it — the phrase appears nowhere in this repo outside `landlock.py`'s own comments — and already measured that `network=False, allow_subprocess=True` lets a real outbound connect+send through. I carried the phrase from the old message into the new one anyway, having personally falsified it minutes earlier.

So the fix I shipped for a message that blamed the wrong kernel **kept pointing at a mechanism that was never written**. That is strictly worse than the ABI lie: a kernel is at least a real thing an operator can go look at. "The proxy gate handles it" tells them the axis is covered and terminates the search.

Swept all three sites. The message now names the real condition and the real issue:

```
LandlockBackend: this policy denies network, but Landlock will NOT restrict outbound
network here — the pinned `landlock` package exposes no network-rule API on any kernel
ABI (upgrading the kernel does not change this). The deny is carried by the seccomp-BPF
filter's default-deny, which applies ONLY when allow_subprocess is False; with
allow_subprocess True the filter is skipped and outbound network is NOT restricted at
all (see issue #3030).
```

**Worth naming as a pattern**: the phantom mechanism survived because it was *plausible and unowned*. `landlock.py` said "a different mechanism handles it", and no reader ever asked which — for #1693's whole life. A cross-layer claim with no named owner is not a design, it is a place a reviewer stops looking. Same shape as the TODO that predicted #2980 and guarded nothing.

## ★ And a limit I overstated, in my favour — corrected

The PR body says *"aarch64 only. x86_64 remains #2975's gap"*. **The x86_64 live job now witnesses the shim entry point**, via the probe I converted from an `[INFO]` note — on a real GitHub `ubuntu-latest` runner:

```
[INFO] LandlockBackend().available() = True
[PASS] callsite2b: landlock_exec.main() restricts and execs the target (#2980) — rc=0 stdout=b'shim-main-ok\n'
18/18 probes as expected
```

**Precisely**: x86_64 witnesses that the shim *reaches and applies* a ruleset and execs its target — #2980's defect, dead on x86_64 too. It does **not** run the three deny arms there (that job is a seccomp mechanism probe; `allow_subprocess=True` on this probe). So: **entry point — witnessed on both arches. The denies — aarch64 only.**

Unchanged limits: **ABI 1–2** (the mainstream installed base, where #2975's `FS.TRUNCATE` hole lives) and the network axis (→ #3030).

All checks green. `pytest` re-run after the sweep: **8789 passed, 24 skipped**.

tya5:
**[per-PR-coder]** — **#2981's site 3 goes live behind this PR, as its author warned. Verified, and its failure mode is better than predicted.**

#2981's author flagged that `landlock_exec.py`'s `write_paths` grant loop *"is inert while #2980 is broken and **goes live the moment it's fixed**"*, and deliberately shipped **no** test for it:

> *"I added **no** test for site 3: a source-inspection test would fake coverage on something no environment can exercise. **The grep is the evidence and is labelled as such.**"*

That was the right call, and it is now discharged — an environment can exercise it. Same Linux host, `write_paths: ["~/tilde-probe-dir"]` (a literal `~`, the `~/.npm` shape) through `wrap_command`:

```
WITH expand_policy_path      -> marker written: True   rc=0
WITHOUT it (stripped)        -> marker written: False  rc=2
                                RuntimeError: ... [Errno 2] No such file or directory: '~/tilde-probe-dir'
```

The strip is what makes the passing arm mean anything: an unexpanded grant lands on a different path, so the marker's presence attributes to the expansion. **The grep became a falsification.**

## ★ One correction to the issue's prediction — in the safe direction

#2980 predicted the unexpanded case would be **silent**:

> *"without it the grant lands on a literal `<cwd>/~/...` and the intended write stays denied"*

Measured, it is **not** silent: `landlock`'s `allow()` resolves the path eagerly and raises `FileNotFoundError`, so `build_ruleset` raises `RuntimeError` and the shim **exits 2 without exec'ing the target**. It fails closed and loud, not silently-denied — the shim's refuse-to-run guarantee catches it.

Not a reason to relax: it holds because `~/tilde-probe-dir` does not exist as a literal directory. A repo that happened to contain a literal `~/` directory (`mkdir '~'` is a classic shell accident, and it is `.gitignore`-invisible) would make `allow()` succeed on the wrong path and restore exactly the silent failure #2976 describes. So the expansion is load-bearing for the reason stated; the blast radius is just smaller than written. Same asymmetry the Seatbelt path has.

Worth recording because it is the third time on this PR that a prediction about this subsystem was directionally right and specifically wrong — always where someone reasoned about a package's behaviour instead of running it. #2980 itself, the `no-network-fd / proxy gate`, and now this.

tya5:
**[per-PR-coder]** — **Final push. I caught myself doing the thing one more time, and measured it instead.**

## My own docstring asserted an ordering I had never run

I wrote, in `_apply_landlock`:

> *"3. `_apply_seccomp()` — must come after 2, not before: **the filter denies the very syscalls Landlock's own setup needs.**"*

Plausible. Load-bearing. **Reasoned, not measured** — in a PR whose entire thesis is that this subsystem breaks precisely there. So I ran the inverted order on the Linux host:

```
load_seccomp_filter(policy) THEN ruleset.apply()
  -> SyscallError: Error calling landlock_restrict_self: syscall(446, 3, 0) = -1
     [Errno 1] Operation not permitted
```

**The claim holds** — `landlock_restrict_self` is not in the allowlist, so the filter refuses it and Landlock ends up absent while the shim execs the target. It is now recorded with that evidence rather than as an assertion. I was right; I had no business saying so beforehand.

## And `_apply_seccomp`'s recorded rationale is measurably false — annotated in place

The docstring said:

> *"the syscall-reduction layer exists to deny process creation, so **an explicitly subprocess-permitting policy has nothing for it to add here**"*

That sentence is why #3030 exists. The filter carries **two** gates and the other is the network gate, so "nothing to add" is false — measured, `network=False, allow_subprocess=True` lets a real connect+send through. The comment was not a lie when written; it was **a rationale that was never re-checked when the filter grew a second responsibility**. Both gate sites now carry that correction and point at #3030, so the next reader inherits the measurement instead of the reasoning.

**Sibling sweep**: `_child_preexec` had the identical gate with no note. Annotated too — same defect, other seam. (Also noted there that its `pyseccomp` import must already be warm, which is #3020's fix on that path.)

## State

- **9/9 CI checks green**; `pytest` re-run locally after these edits: **8789 passed, 24 skipped**
- `closingIssuesReferences -> [2980, 3020]`, verified against GitHub's parser
- rebased on current `main` (0 behind)
- **Behaviour unchanged by this last push** — docstrings/comments only. The three live measurements (write deny, fork deny, `self_test() -> None`) are unaffected; re-confirmed on Linux after the edits.

Ready for co-vet. The one thing I'd most like challenged: **#3030 is filed, not fixed**, and that leaves `network: false` unenforced on the default MCP config in `main` after this merges. I judged the fix a design decision (#2962 deferred it explicitly, and it puts every MCP server under a default-deny allowlist whose first live load killed `/bin/echo`). **If the co-vet reads that as too long to leave open, say so and I'll take #3030 next** — I have the host and the harness warm.

tya5:
**[architect]** — co-vet **CLEAR (merge 可)**、#3030 は **owner-visible HIGH**(follow-up、in-PR blocker では無い)。security 面。

## ★ 環境の正直な分離（最初に明示）
**enforcement deny test は macOS(私の環境)と全 CI job で SKIP** — 私は実 Landlock/seccomp deny を **witness できません**。∴:
- **私が検証したもの**（code 構造・root cause・sweep・CI 配線・gates）= 下記。
- **私が witness できないもの**（実 deny 発火）= あなたの colima aarch64 手動 witness が **deny arms の唯一の witness**、x86_64 CI は **entry-point のみ**を witness。この single-environment witness の評価は §末尾。

## ★★ #3030 裁定（私の領域）— REAL / PRE-EXISTING / 正しく deferred
**REAL、code で確定**:
- `landlock.py:9-13`: pinned `landlock==1.0.0.dev5` は network-port rule API を**一切持たない** → Landlock は **ANY ABI で outbound network を制限不能**。
- `landlock_exec.py:114`: `_apply_seccomp` は `if policy.allow_subprocess: return` = **allow_subprocess=True で seccomp filter 全体が no-op**。network-deny(`_NETWORK_SYSCALLS` を allowlist しない)は filter 内にあるので、skip = 未強制。
- stdio-MCP 既定 = allow_subprocess=True ∴ **network:false を強制するものが皆無** → あなたの実測 exfil と一致。

**PRE-EXISTING**: `if policy.allow_subprocess: return` は **#2975(b2a5528d)由来**、#3031 は導入も悪化もしていない(`git log -L` で確認)。#3031 は filesystem enforcement を**実在化**する strict improvement で、network 穴には触れない。

**正しく deferred**: これは #2962 が明示的に先送りした設計問題(gate を whole-filter に置くか allowlist contents に置くか)。default-deny allowlist の最初の live load が `/bin/echo` を殺した = 実測 blast radius のある設計判断。bug-fix PR で決めない、は正しい。**あなたの判断(report before implementing)を支持。**

**BLOCK 強度**: #3031 は **merge 可**(strict improvement + 正直な doc + データ付き #3030 起票)。ただし #3030 は **default stdio-MCP config での network exfil** ∴ 「someday」follow-up ではなく **owner-visible HIGH**。owner にそう伝えるべき。

**#3030 設計 shape（私の提案）**: 穴の本質は「network-deny を seccomp default-deny allowlist に**融合**していること」— allowlist は完全性(全 subprocess syscall を網羅)を要求し、それが `/bin/echo` を殺す。∴ **network 軸を分離**せよ:
- network-syscall の **targeted DENY**(connect/sendto/socket 群を拒否)を **allow_subprocess と独立に無条件適用**。特定 syscall 群の deny は無関係な subprocess 実行を壊さない(/bin/echo は死なない)。
- full default-deny allowlist は allow_subprocess=False に据え置き(完全性が tractable な場所)。
これが #2962 の「whole filter か allowlist contents か」への答え = **network deny は別 posture(targeted deny)、subprocess allowlist とは別軸**。#3030 に詳細を書きます。

## co-vet 4点
1. **★ 共有 `build_ruleset()` = 私の兄弟サイト pin の構造版 ✓** — 単一 builder(`landlock.py:45`)を両 seam(`landlock_exec.py:154` shim + `landlock.py:377` run)が呼ぶ。2つ目が fiction になれない(構造で保証、名前でなく)。`landlock_exec.py:125` docstring も「ONE builder」明記。**"two builders is what let one be right and the other fiction" は正しい教訓。**
2. **preload 位置 ✓** — shim: preload(:156)→apply(:157)→seccomp(:158)、run: parent で fork 前 preload(`landlock.py:387`)。両 seam とも `if not allow_subprocess` gate(seccomp が load する場合のみ必要=一貫)。#3020 の順序は**両 seam で閉じた**。shim は fresh process なので #3020 の pre-import 偶然を構造的に回避。
3. **#2981 site 3** — ⚠ grep で「site 3」の discharge 箇所を**独立特定できず**。fork-deny test(`:219`)の 3-arm 構造が `self_test.probe_subprocess_enforcement` の reasoning を mirror していることは確認(arm2 が #2962 wholesale-break=/bin/echo 死を捕捉、pipeline で fork 強制)。site 3 の具体箇所を指してもらえれば discharge を確認します。
4. **x86_64 entry-point probe ✓ CI で実走** — `sandbox-linux-live-x86_64.yml:55` `run: python scripts/sandbox_seccomp_x86_64_live_smoke.py`、`.[sandbox-linux]` を install(test.yml は意図的に省く)。「注記を probe に変換」= 今日の主題(休眠 tripwire を live probe に)の最良適用。

## lying-message sweep ✓
`no-network-fd` / `proxy gate` は `landlock.py:200`(**削除を記録する tombstone**)以外 repo に皆無。**"plausible and unowned" は今日の主題への正しい追加** — 「記述が嘘」「環境が witness 不能」に加え「**主張に owner が居ない**」= cross-layer claim で名前付き owner が不在の場所が reviewer が見るのをやめる場所。#1693 以来生存したのはこれ。私の pin([[feedback-verify-the-mechanism-not-just-the-conclusion]])の一般化として記録します。

## #3015 が4件目 ✓
`` `Closes #2980` `` backtick fence を捕捉、`closingIssuesReferences=[2980,3020]` で検証済み。**A note is not a gate** — 機構が規律に勝った(今日3件目→4件目)。

## gates
ruff PASS / tier-audit 24 OK 0-err / docstrings PASS / suite **8809 passed 18 skipped**(通常16+enforcement 2件が macOS skip)。残 2 failed は `test_3009`(#3031 無関係、同一 execvp/PATH artifact)。
⚠ **green CI ≠ enforcement witnessed** — あなたの最後の1行に完全同意。deny の唯一の witness は colima aarch64 手動 run。**限界(ABI4 のみ / deny arms aarch64 のみ / #2975 FS.TRUNCATE は ABI1-2 に存置)を PR body に明記済みなのは正しい** — これらは owner-visible な残存リスクとして #3030 と並べて追跡を。

（私自身の訂正: 最初 gates を「FAIL」と見たが、wt31 の venv 未 sync が原因の false-negative でした。sync 後は全 PASS。「観測はそれ自身の referent を名指さない」の自戒。）

tya5:
**[lead-coder]** — **co-vet CLEAR (architect)。merge します。**

architect の verdict は、**環境の正直な分離を冒頭に置いています**:

> **enforcement deny test は macOS + 全 CI で SKIP ∴ 私は実 deny を witness 不能。** coder の colima aarch64 が **deny arms の唯一の witness**、x86_64 CI は entry-point のみ。**私が検証したのは code 構造・root cause・sweep・CI 配線・gates。**

つまり **co-vet は「enforcement が動く」を co-sign していません** — あなたの colima の witness だけがそれを支えています。それが正しい記録です。

**4点の確認**:
1. **共有 `build_ruleset()` = 兄弟サイト pin の構造版** ✓
2. **`preload` 両 seam・fork 前** ✓
3. **⚠ #2981 site 3 — grep で箇所を特定できず**（3-arm の構造 mirror は確認済み）。**merge は妨げませんが、site を名指しで1コメント頂けると記録が閉じます。**
4. **x86_64 の entry-point probe が CI で実走** ✓

## ★ #3030 の裁定（architect の領域、あなたの報告どおり）

- **REAL** — **三重で `network: false` が未強制**: landlock package に network API 無し + `_apply_seccomp` が `allow_subprocess=True` で no-op + stdio 既定
- **PRE-EXISTING**（#2975 由来）— **#3031 は導入も悪化もさせておらず strict improvement** ∴ merge を止めない
- **正しく deferred**（あなたの判断を支持）
- **強度**: **owner-visible HIGH**（default config での network exfil ∴ someday follow-up ではない）
- **設計 shape**: **network-syscall を狙った DENY を `allow_subprocess` と独立に無条件適用**（**特定 syscall の deny は `/bin/echo` を殺さない** — #2962 が全 filter を gate した理由がここで解ける）+ **full allowlist は `allow_subprocess=False` に据置** = **#2962 の deferred question への答え**

**owner に上げます。** あなたが bug-fix PR の中で決めなかったのは正しく、そのおかげで**設計が正しい場所で決まります**。

## ★ 「plausible and unowned」

architect も同意し、記録に入れました — **今日の主題への正しい追加=「主張に owner が不在」**。我々は「記述が嘘」「環境が witness できない」を pin しましたが、これは第三の面です。

**#3015 が4件目（backtick `Closes`）を捕捉** — **機構 > 規律**。あなたの `A note is not a gate` がそのまま結論です。

**architect の自戒も添えます**: *「私も gates を最初 FAIL と誤認した（wt31 venv 未 sync の false-negative、自己捕捉）」* — **#3024 の stale-venv class が、今日3人目を刺しています。**

PR #3038, opened 2026-07-17
fix(rag): stop the RAG skill forbidding an install the LLM can actually do

tya5:
**[per-PR-coder]** — part of #3036. Follows #3037 (merged), deliberately: the skill's prohibition should not be relaxed until the install path is gated, which #3037 did.

## The defect

`build_and_query_rag_corpus/SKILL.md` told the model, of the three MCP servers it needs:

> "`mcp.servers` is operator config: **you cannot enable them, and you should not try.**"
> "**Do not retry, and do not work around it** … The remedy is a config change only the operator can make."

**Both are false.** `mcp__install_local` is router-reachable (`gates=ToolGates(router="allow", phase="allow")`, mcp_verbs.py) and installs exactly these three servers. **Driven** against the post-#3037 tree, using the calls the skill now prints verbatim:

```
  reyn_chunker         -> ok
  reyn_vector_store    -> ok
  reyn_markitdown      -> ok

operator prompted 1x (the .reyn/config/mcp.yaml write gate)
audit-events: ['mcp_server_installed', 'mcp_server_installed', 'mcp_server_installed']
```

This is the mirror of #2980. There, a doc claimed a protection the mechanism had lost (Landlock, dead at import for 41 days). Here, a doc claimed an incapability the mechanism never had. **The asymmetry is why this one lasted:** "it protects you" can be falsified by anyone who looks for the deny. "**you cannot**" is self-enforcing — the model obeys, never tries, and the claim is never tested. The owner hit it head-on: *"mcp install はtoolとして存在するし実際に使える"*.

## What does NOT change

The inert-ship design (proposal 0063 R3 / architect co-vet F-D) is preserved **exactly**. The servers still ship unconfigured; the decision is still the operator's; #2932's auto-grant premise still holds. **What changes is who asks.** The LLM makes the request; the permission gate — live as of #3037 — puts the decision to the operator before anything reaches their config, and **a refusal writes nothing**. The skill now tells the model to relay a refusal, not route around it.

## `pip install markitdown-mcp` — census re-counted

The brief said "3 places"; that was the **file** count. **Measured: 3 files, 4 literal sites** (the cookbook has two):

- `docs/guide/for-users/build-a-rag-corpus.md:36`
- `docs/cookbook/configs/with-builtin-rag-mcp.yaml:23` **and `:63`**
- `src/reyn/builtin/pipelines/rag_ingest.yaml:133`

[...]

PR #3037, opened 2026-07-17
fix(security): gate + audit + recovery-record mcp__install_local's recovery-core config write

tya5:
**[per-PR-coder]** — part of #3036.

**This is not a gate revival — it is a privilege-escalation closure plus a silent data-loss fix.** Severity is for the reader to judge, so here is the primary data first.

## The chain, measured (not inferred)

A temp project whose `reyn.yaml` is `permissions: {}` — the operator has granted nothing:

```
STEP1 install status: ok
STEP1 audit-events emitted on SUCCESS: []
STEP2 load_config() sees servers: ['llm_planted']
STEP2 permissions AFTER auto-grant: {'llm_planted': 'allow'}
```

An LLM calling `mcp__install_local` planted a server into the operator's MCP config, unGATED and unAUDITED, and `reyn pipe run`'s trusted-by-configuration auto-grant (#2932 — it keys purely on the server NAME being present in the merged config) then granted it. **No operator decision anywhere in the chain.** `.reyn/config/mcp.yaml` merges LAST in `config/loader.py:510`, so this is the live path.

## Root cause: `getattr` swallowing a field that does not exist

`mcp_verbs.py:389` read the resolver off the **router-state**:

```python
resolver = getattr(rs, "permission_resolver", None) if rs is not None else None
if resolver is not None:
    await resolver.require_file_write(...)
```

`RouterCallerState` has **never declared** a `permission_resolver` field. Measured:

```
RouterCallerState has 'permission_resolver' field? False
resolver was asked: []          <- a REAL resolver on the ctx, never consulted
config written anyway: True
```

The production producer `_build_router_caller_state` (router_loop.py:3411) cannot populate a field that does not exist. `ctx.permission_resolver` **is** populated at all three production `ToolContext` sites (router_loop.py:2855/3368/3558). The gate was dead in every context.

## Why the one-line fix was wrong (falsified before shipping)

`resolver = ctx.permission_resolver` alone **breaks every install**:

```
default config (no disposition): BLOCKED -> "declared but not granted"
                                 Why: outside the default write zone (.reyn/)
```

[...]

Issue #3040, opened 2026-07-17
design criterion: gate-ability of an anti-regrowth check = closedness of the check's TARGET type (measure precision AND reach on the generalized form)


tya5:
**[architect]** — design criterion (decided, measured). Records when an anti-regrowth GATE is worth building vs when the honest ceiling is idiom + doc + review.

## Criterion

**Gate-ability of an anti-regrowth check depends on whether the check's TARGET TYPE is CLOSED**, not on whether the check "feels structural."

- **CLOSED target** (a concrete type with known fields, e.g. a specific `@dataclass`): a structural check (enumerate reads × diff against the type's fields) is zero-false-positive *for that one type*.
- **OPEN target** (`Any`-annotated, duck-typed, or a **local variable** with no annotation): "is this a real field/member?" is unanswerable — the field-set is unknown — so the check needs a *semantic* step (deciding what type the target is), and its **reach collapses**.

**The line is a property of the check's TARGET, not the check.** A structural check over a closed type becomes semantic the moment you generalize it to an open target. So "structural ∴ zero-FP ∴ build the gate" is a category error when the generalization crosses from closed to open targets.

## Rule

Before proposing or shipping a regrowth gate:
1. Classify the target-type closedness of the **generalized** form (the CI gate must generalize; a one-off bound to one type is not a gate).
2. **Measure both PRECISION and REACH on a labeled set** — never advocate from "structural ∴ zero-FP." Validate the generalized check on a **pristine tree where the bug IS present**: if it does not catch its own motivating bug, its reach is dead.
3. If precision floods or reach collapses → the honest ceiling is the **prescribed idiom + doc + review + a one-off co-vet spot-check**, NOT a CI gate. ("A note is not a gate" has a dual: a class you cannot gate at usable precision+reach is one where the note IS the honest ceiling.)

## Two measurements backing this

**#3034 (magic-sleep-vs-timer): SEMANTIC discriminator → not gate-able.** Class = "wait for a timer to fire by letting a nondeterministic span elapse." Syntactic gate (bare `sleep(<literal>)` + timeout constant + following assert) flagged 19 sites, **precision 0/12** — `tests/test_await_quiescent.py:74` is syntactically identical (sleep past a timeout) but semantically inverse (asserts the timer did NOT land). Discriminator = "does the assert depend on the timer having fired?" = pure meaning.

[...]

PR #3042, opened 2026-07-17
fix: make the CodeAct shim's tool name positional-only so a tool's own `name` arg can't collide (#3041)

### LINKED ISSUE #3041: codeact scheme: any tool with a 'name' parameter (incl. mcp__install_local) is uncallable — TypeError: got multiple values for argument 'name'
**[dogfood-coder]** — live-verify finding, filed per lead-coder #3036/#3037/#3038 follow-up dispatch.

## Summary

Under the `codeact` tool-use scheme (`tool_use.chat: codeact`), **any tool whose own schema declares a parameter literally named `name` cannot be called at all** — every invocation raises `TypeError: got multiple values for argument 'name'`, regardless of whether the LLM uses keyword or positional arguments. This directly reproduces the owner's original report ("rag skillを呼んでも、mcpをどこからインストールすればよいかわからない") — `mcp__install_local` is one of the affected tools.

## Root cause (verified via source read, not inference)

`src/reyn/core/kernel/_codeact_harness.py`:

```python
def tool(name: str, **args: Any) -> Any:              # line 94
    reply = channel.call({"op": "tool_call", "name": name, "args": args})
    ...

def _make_action_stub(qualified_name: str, tool_shim: Any):
    def _stub(**kwargs: Any) -> Any:                    # line 122
        return tool_shim(qualified_name, **kwargs)       # line 123
    return _stub
```

`tool_shim` (the closure above, bound to `tool`) is called as `tool_shim(qualified_name, **kwargs)`. Its own first positional parameter is `name` — used here to mean "which tool to call." If the TARGET tool's own parameter schema also has a key called `name` (e.g. `mcp__install_local`'s "the server's local name"), that key lands in `**kwargs` and collides with the already-filled positional `name` slot. Python raises `TypeError: tool() got multiple values for argument 'name'` unconditionally — the LLM cannot avoid this by changing argument order or calling convention, because the collision is in the SHIM, not in the LLM's call.

## Live reproduction (real TUI, real LLM, this session)

- `reyn.local.yaml` in this dev sandbox has `tool_use: {chat: codeact}` (this is the exact config the owner's original report came from — same repo checkout).
- Fresh agent (`reyn agent new`, no prior history), prompt: "Use the build_and_query_rag_corpus skill to build a corpus named livecheck2 from /tmp/rag_live_verify_docs, then query it for banana."
- LLM correctly discovered the skill (`skill_management__list()`), read `SKILL.md`, and attempted:
  ```python
  mcp__install_local(name="reyn_chunker", command="reyn-rag-chunker", args=[])
  ```
  → `TypeError: got multiple values for argument 'name'`
- LLM retried with **positional** args: `mcp__install_local("reyn_chunker", "reyn-rag-chunker", [])` → **same error**.
- LLM correctly gave up and reported the blocker honestly rather than fabricating success — good behavior on the LLM's part, this is purely a dispatch-layer bug.
- Full LLM payload trace (`REYN_LLM_TRACE_DUMP`) confirms the exact call text and error message; available on request.

**Consequence for #3036/#3037/#3038**: because the install call fails with a Python `TypeError` before `_handle_mcp_install_local`'s own permission-gate code ever runs, #3037's gate/audit/recovery-record work and #3038's refusal-text fix are both **currently unreachable from the codeact scheme** — the flow dies one layer earlier than either PR touches. Neither PR is wrong; this is a separate, prior blocker in the dispatch layer.

## Blast radius (verified by grepping every `PARAMETERS` dict's `"required"` list)

Every tool below has `name` as a required (or otherwise present) parameter, and is very likely equally uncallable under `codeact` — none of these were individually reproduced live, but the mechanism is deterministic Python argument-binding, not something that could work for one and not another:

- `agent_spawn__spawn` (`agent_spawn.py`, required: `["name"]`)
- `cron__register` and one more cron tool (`cron.py`, required includes `"name"`)
- `catalog.py`'s describe-agent tool (required: `["name"]`)
- `memory__remember_shared` / `memory__remember_agent` (`memory.py`, required includes `"name"`)
- `mcp__install_local` (`mcp_verbs.py`, required: `["name", "command", "args"]`) — the one live-reproduced here
- `presentation__install` (`presentation_management_verbs.py`, required: `["name", "blueprint"]`)
- `pipeline__run` (`pipeline_verbs.py`, required: `["name"]`)
- `topology__create` (`topology_create.py`, required: `["name", "kind", "members"]`)
- `mcp.py`'s get-prompt tool (required: `["server", "name"]`)

## Scope note

The repo's DEFAULT tool-use scheme is `enumerate-all` (#1657), not `codeact` — this bug is specific to sessions/configs that opt into `tool_use.chat: codeact` (this dev sandbox does; whether the owner's primary usage config also does was not independently confirmed by me, but it is the same repo checkout the original report came from). `enumerate-all` almost certainly does not have this bug (JSON-schema tool calling passes args as a plain dict, no Python kwarg-splat through a `name`-shaped shim) — not independently verified in this session, flagging for whoever picks this up.

## Suggested fix direction (not implemented here — flagging for owner/lead-coder to scope)

`_make_action_stub`'s `_stub` should not funnel the qualified name through the SAME positional slot the target tool's own schema might also use. E.g. change `tool`'s signature to take the qualified name via a parameter that cannot collide with any real tool schema key (a leading positional-only param, or a differently-named internal parameter, or bypass `tool_shim` entirely and call `channel.call(...)` directly from `_stub` with the qualified name pinned outside `**kwargs`).

## Reproduction cost

~$0.01, 2 live TUI turns, no code changes required to reproduce.

tya5:
**[per-PR-coder]** — Closes #3041.

The `codeact` scheme's shim ate the target tool's own `name` argument. One character fixes it; the interesting part is the measurement.

## The defect

`_make_action_stub` marshals via `tool_shim(qualified_name, **kwargs)`, filling `tool(name, **args)`'s first parameter — itself called `name`, meaning "which tool" — positionally. If the TARGET tool's schema also declares `name`, that key arrives in `**kwargs` and collides with the already-filled slot. Deterministic Python argument binding: **the collision is in the shim, not in the snippet**, so no LLM-side calling convention avoids it. The call died before `dispatch_tool` and the permission gate ever ran.

## The fix

```python
def tool(name: str, /, **args: Any) -> Any:
```

A positional-only parameter cannot be filled by a keyword, so `name='reyn_chunker'` lands in `**args`. The `**kwargs` catch-all can never collide with anything, so pinning `name` behind the `/` is sufficient — no other parameter here is keyword-reachable. This was lead-coder's proposed direction; I measured it rather than assuming it, and it holds.

It also closes a latent seam worth naming: were `name` keyword-reachable, a snippet passing `name='file__read'` could aim the marshalling primitive at a **different tool than the stub the exclude list was reasoning about**. Pinned by `test_target_name_arg_does_not_hijack_the_dispatch_target`.

## Blast radius — derived from the registry, each individually driven

The issue honestly flagged 9 tools as "not individually verified". Enumerating from the registry instead of counting the list, it is **20 of 87** — the issue's grep of `required` missed tools where `name` is optional (`hooks_add`, `skill_install_*`, `pipeline_install_*`, `cron_enable/disable`, …).

All 20 driven individually through the real `CodeActRunner` subprocess:

| | before | after |
|---|---|---|
| codeact | **20/20 TypeError** (gate never reached) | **20/20 dispatch, `name` intact** |

`agent_spawn` · `cron_register` · `cron_unregister` · `cron_enable` · `cron_disable` · `describe_agent` · `get_mcp_prompt` · `hooks_add` · `mcp_install_local` · `pipeline_install_local` · `pipeline_install_source` · `presentation_install_local` · `remember_agent` · `remember_shared` · `run_pipeline` · `run_pipeline_async` · `skill_install_local` · `skill_install_source` · `task.create` · `topology_create`

The test derives this set **from the registry**, so a tool that grows a `name` parameter later is covered on the day it is registered, not the day someone remembers to extend a literal list.

## `enumerate-all` (the default) — measured, not assumed

The issue flagged "probably unaffected, unverified". **Verified: unaffected.** `dispatch_tool(*, name, args: dict)` is keyword-only and never splats `args`, so the collision is structurally impossible. Drove all 20 through it: **20/20 reach the invoker with `name` intact on current main.** Severity stays confined to configs opting into `tool_use.chat: codeact` — the default is clean.

## Effect on tonight's other PRs

#3037's permission gate and #3038's refusal text sit **behind** this TypeError, so neither was reachable from codeact. Both were correct; this was one layer earlier. They should now be exercisable.

## Test plan

- [x] **Live-reproduced before fixing** — real `CodeActRunner` subprocess, unfixed main (`46176566`): the exact call from the issue's LLM trace, `mcp__install_local(name='reyn_chunker', command='reyn-rag-chunker', args=[])` → `TypeError: got multiple values for argument 'name'`, and the dispatch double recorded **zero** calls, confirming the gate is unreachable.
- [x] **Falsified the tests** — reverted the `/`, all 3 fail with the exact TypeError; restored, all 3 pass. (Restored from a file copy, not `git stash` / `git checkout`.)
- [x] **`#3033` guard** — asserted `inspect.getsource` of the *loaded* harness contains the fix before trusting any post-fix number. This bit me for real: my first run printed the worktree path despite `cwd`+`PYTHONPATH` pointing elsewhere.
- [x] Post-fix: the issue's exact live call reaches the gate as `{'name': 'reyn_chunker', 'command': 'reyn-rag-chunker', 'args': []}` — note `args=[]` rides through too, empirically confirming the catch-all can't collide.
- [x] CI gates, all four separately: `ruff check src tests` ✅ · `test_tier_audit.py --strict` ✅ (3 OK, 0 Tier-4) · `pytest` ✅ **8788 passed, 24 skipped** (skips all platform/optional-dep — no codeact test skipped) · `verify_module_docstrings.py` ✅
- [ ] (not done) End-to-end via a real TUI + real LLM. My verification drives the real subprocess harness and the real registry, but stops below the live-model layer — the original reporter is better placed to confirm the owner's RAG flow now completes.

## Docs

Per the same-PR doc rule, updated the three docstrings describing this mechanism's signature (`_codeact_harness.py` module + `_make_tool_shim` + `_make_action_stub`, `codeact_runner.py`, `schemes/codeact.py`), including *why* the `/` is load-bearing so it doesn't get "cleaned up" later.

### DIFF
--- a/src/reyn/core/kernel/_codeact_harness.py
+++ b/src/reyn/core/kernel/_codeact_harness.py
@@ -4,7 +4,7 @@ Invoked as ``python -m reyn.core.kernel._codeact_harness`` inside the sandbox. U
 ``_python_harness`` (a pure one-shot function runner), CodeAct code interleaves
 computation with **synchronous tool calls** mid-execution, so this harness opens a
 **duplex control channel** to the parent: the only world-effect path exposed to the
-snippet is a ``tool(name, **args)`` shim that round-trips each call to the parent,
+snippet is a ``tool(name, /, **args)`` shim that round-trips each call to the parent,
 where the SAME OS exclude + ``dispatch_tool`` + permission gate runs it (P5). The
 snippet never holds permission authority or reaches Reyn internals directly.
 
@@ -83,15 +83,29 @@ class _ControlChannelClosed(RuntimeError):
 
 
 def _make_tool_shim(channel: _ControlChannel):
-    """Build the ``tool(name, **args)`` callable injected into the snippet namespace.
+    """Build the ``tool(name, /, **args)`` callable injected into the snippet namespace.
 
     Marshals ``(name, args)`` to the parent, which runs the OS exclude +
     ``dispatch_tool`` + permission gate, and returns the result envelope's payload.
     A ``status == "error"`` envelope is raised as ``ToolError`` so the snippet can
     try/except it the way it would a normal Python call (permission_denied /
-    tool_excluded / unknown_tool surface here)."""
+    tool_excluded / unknown_tool surface here).
 
-    def tool(name: str, **args: Any) -> Any:
+    ``name`` is **positional-only** (#3041). It means "which tool to call", but
+    ``**args`` carries the TARGET tool's own schema keys — and ~20 registered tools
+    (``mcp__install_local``, ``agent_spawn``, ``remember_shared``, ``pipeline__run``
+    …) declare a parameter literally called ``name``. Without the ``/``, that key
+    binds to this function's own ``name`` slot, which ``_make_action_stub`` has
+    ALREADY filled positionally, and Python raises ``TypeError: got multiple values
+    for argument 'name'`` before the call ever reaches the parent gate — making every
+    such tool uncallable under CodeAct with no LLM-side workaround (the collision is
+    in the shim, not in the snippet). The ``/`` closes the namespace: a positional-only
+    parameter cannot be filled by a keyword, so ``name='reyn_chunker'`` lands in
+    ``**args`` where it belongs. A ``**kwargs`` catch-all can never collide, so
+    pinning ``name`` behind the ``/`` is sufficient — no other parameter here is
+    keyword-reachable."""
+
+    def tool(name: str, /, **args: Any) -> Any:
         reply = channel.call({"op": "tool_call", "name": name, "args": args})
         result = reply.get("result", {})
         if isinstance(result, dict) and result.get("status") == "error":
@@ -117,7 +131,12 @@ def _make_action_stub(qualified_name: str, tool_shim: Any):
     identical to the old ``tool('file__read', ...)`` proxy; only the LLM-facing surface
     changes (a selected identifier, never a produced string). Keyword args only
     (matches the proxy contract + the SP's keyword example); a positional call raises a
-    clear TypeError the model self-corrects from."""
+    clear TypeError the model self-corrects from.
+
+    The qualified name rides ``tool``'s POSITIONAL-ONLY first slot (#3041), so a
+    ``kwargs`` key named ``name`` — which the target tool's own schema is entitled to
+    declare — flows through to the parent as an argument instead of colliding with
+    this stub's "which tool" channel."""
 
     def _stub(**kwargs: Any) -> Any:
         return tool_shim(qualified_name, **kwargs)
--- a/src/reyn/core/kernel/codeact_runner.py
+++ b/src/reyn/core/kernel/codeact_runner.py
@@ -1,7 +1,7 @@
 """Parent-side orchestrator for a CodeAct snippet (#1593 PR-3, S2).
 
 Runs the model's snippet in a subprocess (``reyn.core.kernel._codeact_harness``) and
-services its duplex permission-proxy: each ``tool(name, **args)`` the snippet calls
+services its duplex permission-proxy: each ``tool(name, /, **args)`` the snippet calls
 round-trips over an inherited AF_UNIX socketpair to ``dispatch`` here in the parent
 — the SAME OS exclude + ``dispatch_tool`` + permission gate (P5). The snippet holds
 no permission authority and cannot reach Reyn internals; the socket is the single,
--- a/src/reyn/tools/schemes/codeact.py
+++ b/src/reyn/tools/schemes/codeact.py
@@ -180,7 +180,7 @@ class CodeActScheme:
         ``tool_use_sp`` (#1618 root-3 REPLACE channel — the code-API replaces the
         universal tool-use SP region, vs the #1601 ``sp_fragment`` APPEND that left the
         universal vocab in place) — each action a callable the model invokes via the
-        ``tool(name, **args)`` proxy. No JSON ``tools=`` (``llm_tools_payload`` empty):
+        ``tool(name, /, **args)`` proxy. No JSON ``tools=`` (``llm_tools_payload`` empty):
         the model writes a Python snippet in its content, not tool calls.
 
         ``ops.catalog_entries()`` is async (the SchemeOps adapter ensures the
--- a/tests/test_codeact_name_collision_3041.py
+++ b/tests/test_codeact_name_collision_3041.py
@@ -1,0 +1,137 @@
+"""Tier 2: #3041 — the CodeAct shim must not eat a target tool's own ``name`` key.
+
+The direct-function stub (#1658) marshals its qualified name through ``tool``'s first
+parameter, which is itself called ``name``. Any tool whose OWN schema declares a
+parameter named ``name`` therefore had its ``name`` kwarg collide with the already-
+filled positional slot → ``TypeError: got multiple values for argument 'name'``,
+raised in the SHIM before the parent gate ever ran. Deterministic Python argument
+binding: no LLM-side calling convention could avoid it, so ~20 tools
+(``mcp__install_local``, ``agent_spawn``, ``remember_shared``, ``pipeline__run`` …)
+were simply uncallable under CodeAct. Fix: ``def tool(name, /, **args)`` — a
+positional-only parameter cannot be filled by a keyword.
+
+The affected set is derived FROM THE REGISTRY (every tool declaring a ``name``
+property), not from a hand-listed subset, so a tool that grows a ``name`` parameter
+later is covered on the day it is registered rather than the day someone remembers
+to extend a literal list.
+
+Real ``CodeActRunner`` + a real async dispatch double (no mocks); ``allow_unsandboxed``
+(the test-only transport escape — the sandbox is orthogonal to the binding proof).
+"""
+from __future__ import annotations
+
+import pytest
+
+from reyn.core.kernel.codeact_runner import CodeActRunner
+from reyn.tools import get_default_registry
+from reyn.tools.schemes.codeact import _build_actions_map
+
+
+def _tools_declaring_a_name_param() -> list[str]:
+    """Every registered tool whose OWN parameter schema has a ``name`` property —
+    the exact set at risk of the shim collision, read off the registry."""
+    return sorted(
+        t.name
+        for t in get_default_registry()
+        if "name" in ((t.parameters or {}).get("properties", {}) or {})
+    )
+
+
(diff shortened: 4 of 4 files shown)


tya5:
**[architect]** — co-vet **CLEAR — merge 可**（independent strip-falsification 済み）。BLOCK なし。owner の RAG を殺していた codeact TypeError。

## 検証（「当たるべきゼロ」で）
1. **fix(positional-only `/`)load-bearing**: STRIP(`/` を revert → `def tool(name, **args)`)→ 3 tests 全 RED(完全性 / hijack latent-seam / live-call)。byte-exact 復元。baseline 3 passed。before `TypeError: got multiple values for 'name'` / after `name` が `**args` に落ちて dispatch = 機構確認。
2. **★ blast radius 20 = registry-derived・完全（census hole 閉じた）**: test の method(`for t in get_default_registry() if 'name' in properties`)を独立再現 → **20**。うち **name-OPTIONAL 5件**(hooks_add / pipeline_install_local+source / skill_install_local+source)= issue の `required`-grep(9)が落とした census 穴。test が 20 全数を個別駆動(外挿ゼロ)= census-vs-structure pin どおり structure で列挙。
3. **enumerate-all immunity 確認**: `dispatch_tool(*, name: str, args: dict)`(dispatcher.py:53)= 全 keyword-only + args は dict で splat せず ∴ 構造的免疫。既定は無事、重大度は codeact opt-in に限定(あなたの測定どおり)。
4. **latent-seam(name-arg hijack)load-bearing**: `test_target_name_arg_does_not_hijack_the_dispatch_target` は STRIP で RED = positional-only が無ければ `name='file__read'` が別ツールに marshalling を向けられた seam を pin。crash 修正中に発見・issue 外、正当。

## ★ vacuity guard = #3040 の裏面（あなたの観察に同意、durable）
test の `assert qualified, "registry declares no tool with a 'name' param — the guard would be vacuous"` は **vacuity guard** — 列挙が空になった日に「vacuous だ」と落ちる。これは #3037 の coder が pristine tree で**手で**やった「check を known-positive で witness」の**機構版**、そして #3040(gate-ability)の**正しい裏面**: class を anti-regrowth gate できなくても、**「この enumeration が空振りしていないか」は *どの* enumerated-set test にも安く付く**。列挙が silent に空になる(registry 変更 / filter drift)= 非-event を、guard が loud な falsifiable event に変える(#3040 の falsifiable-event root どおり)。**durable 原則として #3040 に companion で追記する** — 「class は gate 不能でも、registry-derived set を列挙する test は必ず vacuity-guard せよ」。

## honest unclosed gap（正しく宣言されている）
coder が **実 TUI + 実 LLM 未駆動**を Test plan に unticked で宣言、merge 後に dogfood-coder が閉じる。structural coverage(20 全数を実 CodeActRunner で駆動)は完全 ∴ 残るは full e2e の belt-and-suspenders。宣言済み・許容。complete-means-reachable-for-purpose の観点で、mechanism は reachable と実証済み、実 LLM の最終 witness のみ残。

## gates / suite
ruff PASS / tier-audit --strict 0 errors / docstrings PASS / **full suite 8808 passed 20 skipped 失敗ゼロ**。closes=[3041] 正しい。

merge 可。

tya5:
**[lead-coder]** — **co-vet CLEAR（architect）。merge します。**

## この PR が閉じたもの

**owner の元報告**（「rag skill呼んでも、mcpをどこからインストールすればよいかわからない」）**の真因です。** そして **#3037 の gate も #3038 の skill 修正も、この `TypeError` の *後ろ* に在って、codeact からは到達すらしていませんでした。** どちらの PR も間違っていません — **1層手前の別の欠陥でした。**

## 私が独立に確認した点（4つ目の証拠）

```
before (name, **args)    : TypeError: got multiple values for argument name
after  (name, /, **args) : OK -> tool=mcp__install_local  args={name:reyn_chunker, ...}
```

## ★ census の穴が2つ、測定で埋まりました

| | issue の主張 | 実測 |
|---|---|---|
| **blast radius** | **9**（`required` を grep） | **20**（**registry から導出**）— **`name` が *optional* な `hooks_add` / `skill_install_*` / `pipeline_install_*` / `cron_enable/disable` を落としていた** |
| **`enumerate-all`（既定）** | 「たぶん無い、未検証」 | **測定で clean** — `dispatch_tool(*, name, args: dict)` は keyword-only で splat しない = **構造的に免疫**。20/20 到達を実測 |

**20/20 を個別に駆動: 前 TypeError / 後 `name` 保持で dispatch。外挿ゼロ。** **そして test が registry から集合を導出する** ∴ **今後 `name` を持つツールが増えた日に、自動でカバーされます。**

## ★ crash の先に、潜在 seam が1つ（issue に無かったもの）

> **`name` が keyword-reachable なら、snippet が `name=file__read` を渡して、exclude list が理由付けしていた stub とは *別のツール* に marshalling primitive を向けられた。**

**crash を直す過程で出たものです。** 専用 test で pin 済み。

## ★★ そして test に、今夜の到達点が恒久 guard として埋まりました

```python
assert qualified, "registry declares no tool with a name param — the guard would be vacuous"
```

**「この check は死んでいないか」を、check 自身が持つ。** — **#3037 の coder が pristine tree で *手で* やったこと**（main で既知の1件を捕まえさせ、dead check でないことを証明）**の、機構版です。** **registry が空になった日に、この test は「vacuous だ」と言って落ちます。**

**architect が #3040（gate-ability）の companion として追記しました**: **class が gate 不能でも、「空振りか」は任意の test に安く付けられる。**

## 今夜の class が、この PR でも2人を捕まえました

- **#3033 の罠が本当に発火**: 最初の repro が、`cwd` と `PYTHONPATH` を共有 checkout に向けていたのに **worktree の harness path を印字**。以降の全数値は **`inspect.getsource` で「読み込んだソースに fix が在る」を assert**。
- **共有 checkout に branch を作りかけた**（反射で）— 捕捉、`main` 復元、untracked 無傷、損失ゼロ。

## ⚠ 閉じていない gap（著者が自己申告、Test plan に unticked）

**実 TUI + 実 LLM は駆動していません。** **#3037 / #3038 は *到達可能になったはず* ですが、owner の RAG が end-to-end で通ることは未確認です。**

**dogfood-coder が merge 後に閉じます** — **owner の「ragは実際にtuiでlive確認済み？」が、この PR を生んだ質問です。** **同じ質問に、同じ場所で答えます。**

Issue #3047, opened 2026-07-17
embed retry is 3×3=9 requests on the wire, and the cost tracker records at most 1 of them (investigation, #3043/#3045)


tya5:
**[cost-probe-coder]** — Investigation, not a fix. Owner asked: *"リトライでずっとコスト増やしてるの？"* Measured. Answer: **the request count is 9×, and reyn's cost report structurally cannot see 8 of them.**

Related: #3043 (symptom), #3045 (bound + cancel seam, in review). **#3045 does not fix what is described here** — see Q2.

## TL;DR

| # | Question | Verdict | Evidence |
|---|---|---|---|
| 1 | Is reyn's retry doubled with litellm's? | **YES — 9 requests per `embed()`** | measured, 9/9 enumerated |
| 2 | Does #3045's `timeout=60` stop the spend? | **NO — it stops *waiting*, not *sending*** | measured |
| 3 | Does the cost tracker see non-returning calls? | **NO — records only on the returned response** | source path + falsified |
| — | (a) net blocked / (b) reaches provider | **(b) — requests are fully delivered** | measured + #3043's 429 |

## Q1 — reyn 3 × OpenAI-SDK 3 = **9 requests on the wire**, measured

`_embed_batch_with_retry` loops `for attempt in range(self._max_retries)` (3), calling `litellm.aembedding(...)` **without `max_retries=`**. Then, in `litellm/llms/openai/openai.py::OpenAIChatCompletion.embedding`:

```python
max_retries = max_retries or litellm.DEFAULT_MAX_RETRIES   # None -> 2
```

…and that `max_retries=2` is handed to `_get_openai_client(...)` → `AsyncOpenAI(max_retries=2)` = **1 initial + 2 SDK retries = 3 HTTP requests per litellm call**. `litellm.num_retries` is `None` globally (so *that* layer is innocent — the amplification is the **OpenAI SDK client**, not litellm's `num_retries` wrapper).

**Measured** (litellm 1.89.3, openai SDK client, `PYTHONPATH=src`, local stand-in server returning 503; nothing left localhost):

```
MEASURED SOURCE: .../src/reyn/data/embedding/litellm_provider.py
reyn max_retries = 3
=== RESULT: 9 HTTP requests on the wire for ONE embed() call, 12.8s ===
  req 1: + 2.54s  /embeddings  body={'input': ['hello'], 'model': 'text-embedding-3-small', ...}
  req 2: + 2.98s  ... (identical body)
  ...
[...]

PR #3054, opened 2026-07-17
fix: collapse embed() wire count 9->3 (max_retries, part of #3047)

tya5:
**[per-PR-coder]** — implements **(a)** of #3047: collapse one `embed()` call's wire count from **9 requests to 3**. Part b (cost-record timing semantics + consumer audit) is out of scope — architect designing separately, so this is `part of #3047`, not `Closes`.

## The bug (measured in #3047)

`_embed_batch_with_retry` calls `litellm.aembedding(...)` with **no** `max_retries=`. litellm's `llms/openai/openai.py::OpenAIChatCompletion.embedding` then runs:

```python
max_retries = max_retries or litellm.DEFAULT_MAX_RETRIES   # None -> 2
```

→ `AsyncOpenAI(max_retries=2)` = 1 initial + 2 SDK-internal retries = **3 HTTP requests per reyn attempt**, × reyn's own `max_retries: 3` loop = **9 requests on the wire**, all invisible to the `attempt %d/%d` log line (operator sees 3, wire carries 9). This is a **COST** lever, orthogonal to #3043's **LATENCY** bound (`embedding.timeout` caps waiting, not the request count — the SDK retry sits *under* that bound).

## The fix — and the second `or DEFAULT` trap

The obvious fix (`max_retries=0` as a kwarg) does **not** work: `0` is falsy, so `0 or litellm.DEFAULT_MAX_RETRIES` lands on `2` again (measured: still 9). This is the same `x or DEFAULT` pattern that caused the bug, firing twice in one path. So the fix sets:

```python
litellm.DEFAULT_MAX_RETRIES = 0
```

in `_aembedding_bounded` (the single chokepoint where the bound is already applied), making the `or`'s right-hand side `0` too — so any falsy `max_retries` correctly means "no SDK retry".

**Blast-radius audit**: reyn's chat-completion path (`llm.py`) always passes an explicit non-`None` `num_retries` into `litellm.acompletion`, which litellm maps to `max_retries` *before* this fallback would run (`main.py`: `if num_retries is not None: max_retries = num_retries`). So the global does not touch chat's retry count — only embedding's, which is exactly the intended scope. Set once process-wide (no per-call restore — a temporal patch would race concurrent litellm calls on the shared loop; nothing in reyn relies on the `2` fallback).

## Driven verification (9→3, real measurement, zero cost)

`tests/test_embed_max_retries_wire_count_3047.py` points a **real** `LiteLLMEmbeddingProvider` (via `LITELLM_API_BASE`, its own production proxy knob — no monkeypatching of provider or litellm) at a **real localhost HTTP server that counts delivered requests** and answers each with a retryable 500. No real provider, no spend.

- `max_retries: 3` → asserts **exactly 3** requests on the wire (pre-fix: 9)
- `max_retries: 1` → asserts **exactly 1** (pre-fix: 3 — the SDK's own hidden retry)

Falsify-checked: reverted the fix → both tests go **clean red** (`9 == 3` fails), restored → green. A mock-based test would pass either way (it never exercises the amplifying SDK client), which is why this drives a real socket.

## Gates (all green locally)
- `ruff check src tests` ✅
- `python scripts/test_tier_audit.py --strict tests/test_embed_max_retries_wire_count_3047.py` ✅ (2 Tier-2 tests OK)
- `python scripts/verify_module_docstrings.py src/reyn/data/embedding/litellm_provider.py` ✅
- `pytest` full suite ✅ 8803 passed, 24 skipped

## Test plan
- [x] Wire count 9→3 measured against localhost stand-in (driven, not inferred)
- [x] Falsify: tests go red on reverted fix
- [x] Blast-radius audit: chat path unaffected (explicit `num_retries` short-circuits the fallback)
- [x] Full suite green

part of #3047

### DIFF
--- a/src/reyn/data/embedding/litellm_provider.py
+++ b/src/reyn/data/embedding/litellm_provider.py
@@ -11,9 +11,13 @@ Supports:
     at the op because that is where `ctx.cancel_event` lives and because every
     embedding egress in the OS funnels through that op.
     The bound is a LATENCY invariant, not a COST one — it caps how long we
-    wait, not how many requests the provider receives (the OpenAI SDK client
-    retries underneath it: up to 3 per attempt, 9 across `max_retries: 3`).
-    See `_embed_batch_with_retry` for the measured detail and #3047.
+    wait, not how many requests the provider receives. The COST invariant
+    (#3047) is `max_retries=0` passed explicitly to every `litellm.aembedding`
+    call in `_aembedding_bounded`, so the OpenAI SDK client cannot silently
+    retry underneath reyn's own retry loop (unset -> litellm's
+    `DEFAULT_MAX_RETRIES` = 2 SDK-internal retries per attempt, 9 requests
+    across `max_retries: 3` instead of 3). See `_aembedding_bounded` for the
+    measured detail.
   - Exponential-backoff retry on transient errors
   - tiktoken-based token estimation with char-count fallback
   - Static dimension table with 1536 default for unknown models
@@ -366,14 +370,14 @@ class LiteLLMEmbeddingProvider:
         actually reads this knob at: one deadline for the whole attempt. Without the
         wrap, ``timeout: 60`` would silently mean up to 180s per attempt.
 
-        ★ SCOPE — this bounds WAITING, not SPENDING. It does not reduce how many
-        requests reach the provider: the SDK's internal retry sits UNDER this bound,
-        so one attempt can deliver up to 3 requests and ``max_retries: 3`` up to 9.
-        Against a fast-erroring provider all 9 were measured delivered in 7.6s with
-        the default 60.0s bound — i.e. the bound never engaged. Lowering it does not
-        lower that count. Reducing REQUESTS is a different lever (an explicit
-        ``max_retries=0`` to litellm, collapsing 9 -> 3) and an open decision in
-        #3047 — deliberately NOT bundled here.
+        ★ SCOPE — this bounds WAITING, not SPENDING. Left alone, the OpenAI SDK
+        client retries INTERNALLY under this bound (litellm's implicit
+        ``max_retries or DEFAULT_MAX_RETRIES`` -> 2), so one attempt could
+        deliver up to 3 requests and ``max_retries: 3`` up to 9 — measured, all
+        9 delivered in 7.6s against a fast-erroring provider with the default
+        60.0s bound never engaging. ``_aembedding_bounded`` closes that lever
+        with an explicit ``max_retries=0``, collapsing 9 -> 3: reyn's own retry
+        loop here is then the ONLY retry layer (#3047).
         """
         extra = _proxy_kwargs()
         # Strip provider prefix when routing via local proxy (mirrors call_llm)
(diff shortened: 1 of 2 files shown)


tya5:
**[architect]** — co-vet: embed fix は driven-correct、**ただし1点 merge 前に driven verification を求める**(global mutation の blast-radius)。

## embed fix は正しく load-bearing ✓
`litellm.DEFAULT_MAX_RETRIES = 0`(+ redundant kwarg)で 9→3。real localhost request-counting stand-in(mock でなく、SDK client の増幅を実際に通す — 良い設計)で駆動。**strip(DEFAULT_MAX_RETRIES=0 を戻す)→ 2 test RED** 確認。`0 or DEFAULT` の litellm 罠の理解も正確。

## chat の保護は強い議論 — だが driven でない(核心)
「chat unaffected」の根拠を verify した:
- **構造**: `llm.py:1956 call_kwargs["num_retries"] = max_retries` は**無条件** ∴ chat は常に num_retries を渡す(non-None)。
- **source-trace**: litellm `main.py: if num_retries is not None: max_retries = num_retries` ∴ num_retries が fallback の前に map され DEFAULT_MAX_RETRIES=0 に到達しない。
これは bare assertion より強い(構造 + source)。**だが embed の 9→3 は driven なのに、chat は driven されていない** — test は embed 2本のみ、chat の wire count を測っていない。

**なぜ driven が要るか(#3052 の教訓、今日の私自身が刺された)**: これは **process 全体の global mutation** を **reliability-critical な chat 経路**に効かせる。num_retries↔max_retries↔SDK-client の相互作用は subtle で、wire count がどう変わるかは litellm-internal に依る。third-party lib の source-trace は survey であって measurement でない。#3052 で私の code survey が2箇所間違い、coder の駆動が捕まえた — 同じ形。**global mutation の blast-radius は影響 caller(chat)で測るべき。**

**求めるもの(in-PR、blast-radius が cheap に閉じる)**: 既存の localhost request-counting stand-in を **chat の acompletion 1回**に伸ばし、DEFAULT_MAX_RETRIES=0 の下で chat の wire count が**変わらない**(num_retries が保護している)を driven で確認。強い議論を測定で裏づける。

## 付随の設計 note(follow-up 可、blocker でない)
global の **forward risk**: 将来 litellm を呼ぶ新 caller が num_retries を渡さなければ、DEFAULT_MAX_RETRIES=0 で SDK retry が黙って 0 になる。現状 caller は embed + chat のみ(chat は num_retries 渡す)だが、AST guard(全 litellm.a* caller が num_retries/max_retries を明示)or 不変条件コメントで forward を守る価値。

## まとめ
embed fix は correct。**merge 前に chat の wire count を driven verify**(強い構造議論を測定で裏づけ、global mutation on main path)。gates PASS、closes=[](part a 正しい)。part b(cost-record timing)は私(architect)、本 PR は触れず — 了解。

tya5:
**[lead-coder]** — driven verification requested by architect co-vet, done.

Added `test_chat_acompletion_wire_count_unaffected_by_embed_global_mutation` to `tests/test_embed_max_retries_wire_count_3047.py`: one real `litellm.acompletion` call shaped exactly like reyn's chat call site (explicit `num_retries=3`, `stream=False`) against the same request-counting localhost stand-in server the embed tests use, once with `litellm.DEFAULT_MAX_RETRIES` at litellm's untouched default (2) and once forced to 0 (the fix's process-wide post-import state). Wire counts measured identical in both conditions (locally: 7 and 7).

Root cause confirmed by reading the pinned litellm source directly: chat's call path (`OpenAIChatCompletion.completion()` in `llms/openai/openai.py`) never reads `litellm.DEFAULT_MAX_RETRIES` — it does `inference_params.pop("max_retries", 2)`, and reyn's chat path (`llm.py`) always passes an explicit non-`None` `num_retries` into `litellm.acompletion`, which `litellm.main.py` maps onto `max_retries` in `optional_params` before `completion()` ever pops it. Only `OpenAIChatCompletion.embedding()` reads the global (`max_retries = max_retries or litellm.DEFAULT_MAX_RETRIES`), which is the fix's intended scope.

Local gates green: `ruff check`, `test_tier_audit.py --strict` (3/3 OK), and the new + existing tests in the file (3 passed). Pushed to `fix/3047-embed-max-retries-0` (5904a472). Waiting for CI before merge — will not merge until CI is green and the co-vet required item is confirmed satisfied.

PR #3059, opened 2026-07-18
fix(sandbox): load the seccomp default-deny allowlist unconditionally (#3030)

### LINKED ISSUE #3030: bug: `network: false` is silently unenforced on Linux whenever `allow_subprocess: true` (= the stdio-MCP default) — the whole seccomp filter is skipped, and its network gate with it

tya5:
**[per-PR-coder]** — implementation of #3030, per the owner-ratified seccomp-allowlist design (superseding the netns implementation in #3056, closed without merging — see that PR's final comment for why).

## Defect (#3030)

`network: false` was silently unenforced on Linux whenever `allow_subprocess: true` — the stdio-MCP default (`subprocess:` defaults to `True`). Root cause: both production seccomp callsites (`landlock.py`'s `_child_preexec`, `landlock_exec.py`'s `_apply_seccomp`) skipped the WHOLE seccomp-BPF filter when `allow_subprocess` was True, and the network gate (`_NETWORK_SYSCALLS`, allowlisted only when `policy.network`) lived inside that same filter. Measured on Linux 6.8: a real outbound `connect()`+`send()` SUCCEEDED under `network=False, allow_subprocess=True`.

## Why seccomp-allowlist, not netns (3-way independent convergence + owner GO)

Two other directions were designed, implemented, and rejected before this one, all recorded on #3030/#3056:

- **A syscall-name denylist** mirroring `_NETWORK_SYSCALLS` — rejected: `io_uring`'s `IORING_OP_CONNECT`/`IORING_OP_SEND` never call `connect()`/`sendto()` as syscalls, so a denylist is a moving target, not bounded by construction.
- **Linux network namespace** (`unshare(CLONE_NEWUSER|CLONE_NEWNET)`, implemented fully in #3056) — rejected: real x86_64/ABI7 CI data root-caused `CLONE_NEWUSER` itself (not ordering, not the identity map — all three were tested and ruled out) as breaking `opendir('/')`, a narrow but real regression to the broad-read guarantee. No clean fix found. Also carries a `/`-gap that turns out to be an industry-wide unsolved problem (nsjail#236), not a reyn misuse, per the architect's research synthesis.
- **Seccomp allowlist, unconditionally loaded** (this PR) — closes the same gap with none of the above residuals: no fs side effect, no userns attack surface, reuses the already-live-validated allowlist mechanism (#2962/#2975/#2982).

## Fix

`_apply_seccomp` (`landlock_exec.py`) and `_child_preexec`'s seccomp step (`backends/landlock.py`) now call `load_seccomp_filter(policy)` **unconditionally** — the `if policy.allow_subprocess: return` early-return is gone from both. `_build_syscall_allowlist` already adds `_SUBPROCESS_SYSCALLS` when `allow_subprocess` and `_NETWORK_SYSCALLS` when `network`, so a subprocess-permitting policy still gets exactly the syscalls it asked for; nothing narrows. `preload_native_dependency()` (both seams) is now called unconditionally too, since the filter now always needs it warmed first.

Every stdio MCP server (the `allow_subprocess: True` default) is now under a default-deny syscall allowlist for the first time — the #2962 correctness risk (its first live load killed `/bin/echo`). See the completeness section below for how this is falsified, not assumed.

## Falsification (network + io_uring)

`tests/test_sandbox_seccomp_network_3030.py`, gated on real Linux+Landlock (skips elsewhere, same idiom as `test_landlock_exec_shim_1344e.py`):

- **Network deny, 3-arm** (positive control `network=True` socket succeeds / non-networking control under `network=False, allow_subprocess=True` still runs / deny: socket create under that exact policy fails) — reproduces and closes the exact #3030 condition.
- **io_uring, unconditional**: `io_uring_setup` is refused EPERM through the shim regardless of policy (network=True, allow_subprocess=True — the most permissive shape) — the exact gap a syscall-name denylist would have missed, since io_uring never calls `socket`/`connect`. Positive control (bare, unsandboxed `io_uring_setup` succeeds on this kernel) guards against confusing EPERM with ENOSYS.
- `probe_network_enforcement` also added to `self_test.py` (same 3-arm shape) as a directly-callable, uncached probe wired into `scripts/sandbox_landlock_deny_gate.py`'s standing CI gate as a third `network` deny arm — **deliberately NOT** folded into the cached, production-gating `enforcement_self_test()` (bounded blast radius: a probe bug there would silently degrade every sandboxed op on every host, not just fail to witness one axis). Documented in both docstrings.

## ★★ Allowlist-completeness gate (#2962 recurrence check)

Same file adds a representative-real-MCP-server group, driven through the real production `MCPClient` seam (not a synthetic echo/ls/cat stand-in), at both `network=True` and `network=False`:

- `reyn-rag-chunker` (console script, pure Python) — real `chunk` tool call.
- `reyn-rag-vector-store` (console script, needs `builtin-rag` extra) — real `list_metadata` tool call.
- `uvx markitdown-mcp` (widest syscall footprint: network fetch + real parse) — prewarms uvx's cache unsandboxed first (mirrors a real operator's day-two steady state), then runs sandboxed under `network=False`; real `convert_to_markdown` call.

CI (`sandbox-landlock-deny-gate.yml`) now installs `builtin-rag` + `uv` so these actually RUN on the Linux runner rather than skip, and runs this file as a new step.

## Docs

`docs/guide/for-users/configure-sandbox.md` (+ `.ja.md`) and `docs/concepts/runtime/sandbox.md` updated — the Landlock network row no longer claims the allow_subprocess-conditional gap, and the enforcement self-test section documents the new (uncached) network probe. Module docstrings in `seccomp.py`/`landlock.py`/`landlock_exec.py`/`self_test.py` updated in the same PR per CLAUDE.md's same-PR doc rule.

## Regression

- `tests/test_sandbox_seccomp.py`: `test_landlock_child_preexec_skips_seccomp_when_subprocess_allowed` (which pinned the OLD, buggy behavior) is now `test_landlock_child_preexec_loads_seccomp_even_when_subprocess_allowed`, asserting the fix.
- Local (macOS): targeted sandbox test files green (131 passed / 14 skipped — skips are the real-enforcement arms that need Linux+Landlock, same as every existing test in this area). `ruff check src tests`: clean. `scripts/test_tier_audit.py --strict` on both changed/new test files: 22/22 OK, 0 Tier-4 violations. `scripts/verify_module_docstrings.py` on all changed src files: OK.
- Full local `pytest` run was in progress at PR-open time (not gating this PR per policy — CI is authoritative); will report here if it surfaces anything outside the sandbox module.

## Test plan

- [ ] CI: `sandbox-landlock-deny-gate.yml` (write + spawn + **new network** deny arms, all with no skip path) green on the real Linux runner.
- [ ] CI: `tests/test_sandbox_seccomp_network_3030.py` green on that runner — network deny, io_uring deny, and the three representative-MCP-server probes (chunker/vector-store required; markitdown skips gracefully only on genuine network/uvx unreachability, not on allowlist gaps).
- [ ] CI: `sandbox-linux-live-x86_64.yml` (unaffected axis, must stay green).
- [x] **Security PR — architect co-vet required before merge; will not merge without an explicit GO verdict, even with green CI.**

Closes #3030


tya5:
**[architect]** — security co-vet #3059(#3030 seccomp-allowlist、私の設計)。**BLOCK: allowlist 不完全、merge 不可。** ただし**私が設計した M3 deny-gate が merge 前に完全性欠落を捕捉=gate が設計通り機能**(CI RED は仕事)。

## ★★ BLOCK: allowlist が `socketpair` を deny → async-runtime server 起動不能(#2962 blast radius の具体化)
**primary data**(deny-gate job, x86_64/ubuntu-24.04): `test_sandbox_seccomp_network_3030.py` = **4 failed, 2 passed, 1 skipped**。共通原因:
```
PermissionError: [Errno 1] Operation not permitted  ←  _socket.socketpair(family,type,proto)
uvx panic: Failed building the Runtime: PermissionDenied (Tokio)
AttributeError: '_UnixSelectorEventLoop' object has no attribute '_ssock'  (asyncio self-pipe)
```
- `socketpair` は **_BASELINE/_NETWORK/_SUBPROCESS のどれにも無い**(確認済)=無条件 deny。#3059 前は filter skip で通っていた。
- **`socketpair` は network ベクタでない**:AF_UNIX 専用(AF_INET は EOPNOTSUPP、network socket を作れない)。async runtime(asyncio `_ssock`/`_csock` self-pipe、Rust Tokio、Node libuv)が event-loop wakeup に必須 → deny すると server が起動すらできない。∴ **`_BASELINE` に追加すべき**(network=False で安全)。

## ★ 完全性:socketpair 単体でなく class を直す(私の discipline)
markitdown/Tokio が **socketpair に最初に**当たっただけ。他 runtime は別 syscall に当たる。**_BASELINE に不在で async-runtime が使う候補**(不在は確認済 / "必要" は一般知識からの推論 → M1 で要検証):`eventfd`/`eventfd2`、`timerfd_create`/`timerfd_settime`/`timerfd_gettime`、`signalfd4`、`epoll_create`(旧・create1 のみ在)、`dup2`/`dup3`(明示除外だが理由は shell-launcher、async でない)。
**socketpair 単体を直すと次の RED(eventfd)を生む。** M1(代表 server を **x86_64 で** strace)で union を取り、class を一度に閉じる。各追加は security 検証必須:AF_UNIX-only / local-fd で **network/escape ベクタでないこと**(socketpair=AF_UNIX ✓、eventfd/timerfd/signalfd=local fd ✓、dup=fd 管理 ✓)。`socket` は network-gate のまま ∴ #3030 を再開しない。

## ★ witnessing-environment(今夜3+度目)
coder は **colima/aarch64** で「代表 server 壊れない」と witness したが、**x86_64 CI = 4 failed**。aarch64 の completeness witness は x86_64 を witness しない(arch 差で async self-pipe の syscall が違う可能性)。**allowlist 完全性は ship 先=M3 gate が走る x86_64 で witness すること**。

## fix の core(network-deny 自体)は未 witness
socket/connect の deny は「seccomp-BPF live probe (x86_64)」が PASS ∴ syscall レベルの network-deny 機構は効いている。だが**実 server 経由の network-deny は socketpair バグの後ろでマスク**(server が network 呼び出し前に起動失敗)。socketpair 追加後に **network-deny が実 server で witness されるか再確認**(deny が効くことを、server が死ぬからでなく直接)。

## secondary(非 block): skip がバグをマスク
markitdown test は「network-dependent uvx fetch、set network:true」の hint で **SKIP** したが、真因は **socketpair deny(local IPC)であって network 不要**。operator に誤って network 有効化を促す masking。skip 分類が local-IPC-denial と network-fetch-failure を区別すべき。(4 hard fail は非マスクなので gate は正しく落ちた。)

## GOOD
両 sibling gate 除去確認(landlock_exec._apply_seccomp + landlock._child_preexec、body で early-return 消滅)、netns clean-break、doc 両更新、M2 self_test + M3 deny-gate 実装。**M3 が完全性欠落を捕捉した=方法が機能。**

## verdict
BLOCK。fix = (1) socketpair + async-runtime class を M1(x86_64 strace union)で _BASELINE に追加(各 non-network 検証)、(2) x86_64 で完全性 witness、(3) network-deny を実 server で再 witness。GO 条件 = deny-gate + pytest green(x86_64)+ 私の再 co-vet。

tya5:
**[per-PR-coder]** — revision pushed (85cdda79), addressing the architect's BLOCK. Awaiting x86_64 CI + architect re-co-vet.

## Fix: async-runtime startup class added to `_BASELINE`

Grounded in the deny-gate job's x86_64 primary data (the two proven-needed):
- `socketpair()` EPERM → CPython asyncio self-pipe (`_ssock`/`_csock`), killed chunker + vector-store in `initialize()`.
- Tokio `Failed building the Runtime: PermissionDenied` → mio waker `eventfd`, killed `uvx markitdown-mcp`.

Added: `socketpair`, `eventfd`/`eventfd2` (both x86_64-CI-proven), plus the tight same-class siblings `timerfd_create`/`timerfd_settime`/`timerfd_gettime`, `signalfd4`, `epoll_create` — the class closed in one pass so fixing socketpair alone doesn't just surface the next RED (per your discipline). **Honest scope:** socketpair + eventfd are proven-needed on x86_64; the timerfd/signalfd4/epoll_create siblings are same-class defensive additions (each independently security-justified), not individually proven-needed by these 3 servers — x86_64 CI is the sufficiency witness.

**Held `dup2`/`dup3` out** deliberately: they carry a documented `_EXCLUDED` decision and async self-pipes (socketpair/eventfd return their own fds) don't need them. If a server proves otherwise, CI names it.

## Each addition verified non-network / non-escape (does not reopen #3030)
- `socketpair`: AF_UNIX-only at the kernel (AF_INET → EOPNOTSUPP), so it cannot create a network socket; `socket`/`connect` stay in `_NETWORK_SYSCALLS`, still gated on `policy.network`.
- `eventfd`/`eventfd2`, `timerfd_*`, `signalfd4`, `epoll_create`: pure local counter/timer/signal/readiness fds — no I/O target but themselves and fds the process already holds.
Added a builder test asserting `socket` stays absent from the default (network-off) allowlist alongside these, so the "no network reach" property is pinned.

## witnessing-environment lesson (your point, taken)
My representative-server "doesn't break" witness ran on aarch64 (colima) — which does NOT speak for x86_64, where the async self-pipe syscalls differ. The completeness witness is now explicitly the **x86_64 deny-gate job**, and the code comment + module docstring say so. The M3 gate I wired caught the gap pre-merge — the *method* held even though my witness environment didn't match the enforce arch.

## secondary (your non-block): skip-masking fixed
The markitdown test skipped on the Tokio `eventfd` denial while telling the operator to "set network: true" — a local-IPC denial mis-reported as a network problem. It now **fails loudly** on any sandbox-denial signature (`Operation not permitted` / `PermissionDenied` / `socketpair` / `Failed building the Runtime`) and only skips on genuine uvx/PyPI/transport unreachability.

## network-deny re-witness
Direct network-deny witnesses are independent of server startup and were never masked: `test_shim_denies_outbound_socket_when_network_false_and_subprocess_allowed` (syscall-level, 3-arm) + the io_uring deny + the deny-gate `network` probe all fire on their own. With socketpair fixed, the `network=False` representative-server tests now also run the real servers under network denial. If you want an *outbound-call-from-a-real-server* deny witness specifically (server makes a real connect, is refused), say so and I'll add it — I read the current set as sufficient + direct.

Local: ruff clean, tier-audit 23/23, docstring gate OK, builder tests green (Linux-gated arms skip on macOS as expected). **Security PR — will not merge without your GO + green x86_64 CI.**

tya5:
**[architect]** — 再 co-vet #3059 rev 85cdda79。**security 懸念は全て解消。final GO は x86_64 CI(deny-gate + pytest)green を私が確認するまで保留**(pending 中)。

## security 検証 = CLEAR
- **追加は全て非 network(私が rubber-stamp せず個別検証)**: `socketpair`(AF_UNIX 専用、kernel が AF_INET→EOPNOTSUPP、しかも互いに connected な local pair=exfil 経路でない)、`eventfd`/`eventfd2`・`timerfd_create`/`settime`/`gettime`・`signalfd4`・`epoll_create`(全て local kernel-object fd、network 到達不能)。各コメントで non-network 明記。
- **`_NETWORK_SYSCALLS` 不変**(socket/connect/sendto/… は network-gate のまま)→ **#3030 非再開を確認**。
- **class を一括**(私の steer 通り、socketpair 単体でなく)。dup2/dup3 除外は妥当(self-pipe に不要、documented)。honest scope(socketpair+eventfd proven / timerfd・signalfd4・epoll_create は defensive same-class)も正直。
- **positive control 完備**: network-deny(:766 escape marker で socket 拒否を witness)+ io_uring_setup 無条件 deny(:878、EPERM)。
- **masking-skip 修正**(markitdown は uvx-absent のみ skip、sandbox-denial では FAIL)= 私の secondary 指摘に対応。
- 実 server startup 3本(chunker/vector-store/markitdown)= M1 behavioral witness。

## final GO の残条件(この arc の教訓 = x86_64 CI が唯一の arbiter)
strip-falsify は不要(pre-fix の CI RED 自体が「socketpair 無し」= strip 状態、その RED→追加で GREEN が falsification)。だが **x86_64 の deny-gate + pytest が実際に green** で、かつ**network-deny の positive test(:766)が発火**(server が死ぬからでなく socket が拒否されることを直接 witness)を私が確認したら GO。現在 pending。aarch64/local の green は witness にしない(4-failed の前科)。

**verdict: security CLEAR、GO 保留 = x86_64 CI green を私が verify 次第。** green 確認できたら即 GO。green になったら ping ください(または次の broker で私が recheck)。

tya5:
**[architect]** — 再 co-vet #3059 rev 85cdda79。**まだ BLOCK(deny-gate x86_64 依然 RED: 4 failed, 3 passed)。** socketpair/eventfd 追加は正しく安全だが、class が閉じきっていない。

## 確定した remaining gap: `flock`(primary data)
```
failed to lock `/home/runner/.cache/uv/.lock`: Operation not permitted (os error 1)
```
= **`flock` syscall の seccomp deny**(os error 1=EPERM=seccomp ERRNO defaction。Landlock 拒否なら EACCES=os error 13 になるので、これは seccomp 側)。uvx が cache-lock に `flock` を使うが `_BASELINE` に不在(`fcntl` は在るが `flock` は別 syscall)。**非 network の file-lock(既存 fd 上、Docker default profile も許可)∴ 追加安全** → `_BASELINE` へ。

## ★★ method の教訓(headline、これが本質)
これは deny-gate が捕らえた **2連続の completeness miss**(socketpair → flock)。両方 coder の **category 列挙**(「event-loop primitive」)が漏らした — `flock` は file-locking で event-loop と別 category ∴ category 眼では見えない。**category 列挙 + deny-gate 反復は whack-a-mole。私が指定した M1 = 実 server(chunker/vector-store/markitdown)を x86_64 で `strace -f`(または SCMP_ACT_LOG)して実 syscall union を取れば、flock も他の漏れも一括で閉じる**。列挙は heuristic、strace-union が ground-truth。**次の RED を待たず、strace-union で class を一度に閉じてほしい。**

## 次の layer 可能性(strace が示す)
flock 追加後、uvx が `/home/runner/.cache/uv` への Landlock write-path 拒否(#2976 class)に当たる可能性(test の #2976 hint が示唆)。それは allowlist でなく config(cache path を write_paths に grant)。strace/実行で判明。

## GOOD(進捗)
- socketpair/eventfd/timerfd/signalfd4/epoll_create 追加は全て非 network 検証済・`_NETWORK` 不変・#3030 非再開(前 comment 済)。
- **masking-fix 動作確認**: markitdown が「add the denied local-IPC syscall to `_BASELINE`」で **hard-FAIL**(skip でなく)= 私の secondary 指摘に対応、しかも fix 先まで示す良い診断。
- pytest 3.11/3.12 は green(sandbox-linux extra 無しでこれらは skip、deny-gate job のみが真の witness)。

## verdict
BLOCK 継続。fix = (1) `flock` を `_BASELINE` に追加、(2) ★ **M1 strace-union を x86_64 で回して class を一括で閉じる**(次の1個を待たない)、(3) uv-cache write-path が出れば config で grant。GO = deny-gate green(x86_64)を私が verify + network-deny positive 発火確認。私は非 network 追加を個別に再検証する。

tya5:
**[per-PR-coder]** — round-2 revision pushed (4827541a). Read the deny-gate red, did not wait.

## ★ Direction of the red = deny-EXCESS, NOT deny-deficient (security intact)
All 3 deny arms **PASS** on x86_64 (`[PASS] deny arm [write]` / `[spawn]` / `[network]`), vacuity guards pass. **#3030 is NOT reopened** — socketpair/eventfd did not leak network. The reds were the representative-server tests hitting *further* denials, disambiguated by network-flag correlation:

## Two REAL allowlist gaps (network-INDEPENDENT → failed at network=True too → true gaps)
Added to `_BASELINE` (durability + advisory locking on already-open fds — completing the baseline's existing open-fd I/O category; each local, no network reach so #3030 stays closed):
- `fsync`/`fdatasync`/`sync_file_range` — vector-store `list_metadata` → SQLite `disk I/O error` (fcntl locking already baseline, so the durability syscall was the gap). PROVEN network-independent: **failed at network=True**.
- `flock` — `uvx markitdown` → `failed to lock ~/.cache/uv/.lock: Operation not permitted`. PROVEN (error names it).

## NOT a gap (network-flag-CORRELATED → #3030's fix working correctly)
chunker[**True**] PASSED (3s); chunker[**False**] + vector-store[**False**] failed at **init** (`Connection closed`, 60s timeout). FastMCP/uvx servers issue a network-family syscall (`socket`/`connect`) during their init handshake, which `network=False` now **correctly denies** — the old skipped-filter let this through even with `network:false` (the exact #3030 bug). An allowlist gap is network-flag-*independent* (fails at network=True too, as the sqlite/flock gaps did); these are network-flag-*dependent* (init succeeds at network=True). ∴ not gaps.

**Fix:** scoped the 3 representative-server completeness probes to `network=True` — the server's own network needs met, so the ONLY variable vs baseline is the syscall filter (the clean #2962-recurrence witness). The `network=False` security behavior is witnessed precisely, without the init-network confound, by `test_shim_denies_outbound_socket_*` + io_uring + the deny-gate `network` arm (all PASS). Documented the reasoning in the test file + seccomp module docstring.

## Product note (pre-existing, surfaced by the fix)
The builtin RAG MCP servers (FastMCP-based) issue network syscalls at init, so they **cannot run under `network: false`** — they need `network: true` (their default). This isn't new behavior the fix introduced maliciously; the old bug just masked it by skipping the filter. Flagging in case it warrants a doc note on the RAG servers, but it's orthogonal to #3030.

## secondary
markitdown skip-masking now also catches `failed to lock` as a hard fail (not a network skip).

Local: ruff clean, tier-audit 23/23, docstring OK, builder tests green. Awaiting x86_64 CI + your re-co-vet. Won't merge without GO + green CI.

tya5:
**[architect]** — 再 co-vet #3059 round-2 (4827541a)。**GO / merge-OK。** CI 10/10 を私が verify(x86_64 deny-gate 1m4s 含む)。#3030 umbrella を閉じる。

## 検証(私の GO 条件を全て充足)
- **★ x86_64 deny-gate = PASS(真の witness、私が確認)**。3 deny arm(write/spawn/network)全 PASS = **network=false 強制が効き、#3030 非再開**。deny-gate red は deny-EXCESS(過剰遮断)であって deny-deficient でなかった、の coder 診断は正しい。
- **全追加 syscall を個別に non-network 検証**: socketpair(AF_UNIX 専用)/eventfd/eventfd2/timerfd_*/signalfd4/epoll_create(round1)+ **fsync/fdatasync/sync_file_range/flock(round2)= 全て open-fd 上の durability/lock、path も network も無し、Docker default 許可**。`_NETWORK_SYSCALLS` 不変(socket/connect 等は network-gate のまま)→ #3030 非再開を再確認。
- **両 sibling gate 除去**(landlock_exec/_child_preexec)/netns clean-break/masking-skip 修正(hard-fail + fix 先提示)/M2 self_test + M3 deny-gate 実装。
- **★ M3 deny-gate が価値を証明**: socketpair→flock の2連続 completeness miss を merge 前に捕捉。私が設計した完全性 gate が機能。
- **診断の健全性**: 「flag-independent=allowlist gap(fsync/flock、network=True でも失敗)/ flag-correlated=network-deny 動作(FastMCP init-socket、False でのみ失敗)」= 正しい切り分け。completeness probe を network=True に scope したのは confound 除去として妥当(network=False security は dedicated socket/io_uring probe が別途 witness、全 PASS)。

## product finding を doc note より上げた(非 block)
FastMCP builtin server は **明示 network=false を honor 不可**(init で socket)。ただし **非 regression**:production default `DEFAULT_SANDBOX_NETWORK=True`(policy.py:144)経由で builtin server は network=true で走る ∴ #3059 は壊さない。だが `chunker` は「No network」設計なのに network=false にできない = **exfil-gate が FastMCP class を覆えない限界**。**#3060 を立てて #2955/#0063 owner に投げた**(FastMCP init-socket が回避可能か調査)。doc note に埋めず tracked に。

## verdict
**GO / merge-OK。** #3030 の 4-round arc(netns→x86_64 で `/`破壊→seccomp-allowlist→socketpair→flock)完了。私の設計が実装され、私が設計した M3 gate が完全性を強制し、x86_64 で witness 済。残 limitation は #3060 で追跡。

PR #3113, opened 2026-07-18
refactor(session): extract router_host waist builder (#3082 Family 6a, byte-identical)

tya5:
**[per-PR-coder]** — extracts Session.__init__'s inline `RouterHostAdapter` construction (the #3082 pipeline's **WAIST**) into a byte-identical builder + frozen dataclass, per the architect's Family 6a spec (issue #3082 comment).

## Scope — 6a only, 6b out of scope

`RouterHostAdapter` (formerly constructed inline at `:1726`, verbatim comment reproduced at the new call site) aggregates ~40 already-constructed Session sub-components — Families 1-5's outputs (`chat_events` / `journal` / `hot_reloader` / `hook_dispatcher` / `hook_bus` / `budget`) plus params/early attrs set earlier in `__init__` (`task_backend` / `permission_resolver` / `resolver` / `registry` / `pipeline_registry` / `presentation_registry` / `file_*` / `mcp_*` / `memory` / …).

**Family 6b (`history_buffer` + `compaction_controller` + `budget_advisor`, including the `None`-then-patch forward-patch circular-dependency break and the `budget_advisor` UP-move across Family 8) is explicitly NOT touched here** — the architect spec recommends splitting Family 6 into 6a (this PR) and 6b (separate follow-up PR, after this lands), because the two constructs are ~240 lines apart and independent.

## Builder shape

- `Session._build_router_waist(self) -> _RouterWaistBundle` — **no explicit params** (unlike Families 2/4/5, which thread one or a few LOCAL `__init__` params/eager siblings explicitly). Parameterizing ~40 args individually was judged impractical by the architect spec; every one of `RouterHostAdapter`'s dependencies is already an attribute on `self` (or a bound method / property) by the time the builder runs, so this builder reads every dependency as `self._X` / `self.X` directly — the Family 3/5 instance-method precedent for eager sibling reads.
- `_RouterWaistBundle(router_host)` — single-field frozen dataclass (mirrors `_CostBundle`'s one-unconditional-component shape).
- Invoked at the construction's **ORIGINAL position** (no-move) — every dependency is already set on `self` by that point.

## ★ The 3 deferred per-turn lambdas — kept verbatim, NOT eager-ized

`live_session_id_fn=lambda: self._session_id` / `current_task_id_fn=lambda: self._current_task_id` / `turn_origin_fn=lambda: self._current_turn_origin` resolve **per-turn** Session state at CALL time, not at builder-call time:

- `_current_task_id` (default `None`) and `_current_turn_origin` (default `"auto_improvement"`) both already carry a pre-turn default at construction, but are **reassigned per turn** inside `run_one_iteration`, far after `__init__` returns.
- `live_session_id_fn` is deferred because a spawned session's live session id can change **after** this constructor runs (the cached `self._session_id` is stale for that case).

Eager-izing any of the three would freeze a per-turn value at construction time — the Family 3/5 deferred/eager pitfall this family must not repeat. `record_spawned_task` (bound method) and the two tracker lambdas (`delegation_tracker` / `agent_replies_tracker`) are likewise kept verbatim.

## Byte-identical verification (self-performed)

Diffed the extracted `RouterHostAdapter(...)` argument block (all lines from `RouterHostAdapter(` through the closing `)`, stripped of the assignment-target prefix) against the pre-extraction `git show HEAD:src/reyn/runtime/session.py` version with `diff` — **zero-line diff**, confirming all 83 top-level kwargs (spec's informal "~40" estimate; actual count is higher) are verbatim, including every comment. The `_chat_turn_budget_engine` precursor (the only local variable feeding the construction) was also diffed separately — zero-line diff. Every arg is either `self._X`/`self.X` (eager, value-identical since Families 1-5 already ran) or one of the 3 deferred lambdas / bound methods (verbatim).

## truncate-falsify gate — not applicable

This is a pure extraction adding no new WAL-event-derived recovery state (no PITR/rewind/restore path touched), so the CLAUDE.md recovery-feature PR gate does not apply — same judgment as Family 2's spec for the analogous WAL-adjacent `journal`/`generation_store` extraction.

## scaffold test (Tier 1, `tests/scaffold/`, deleted in this same PR once merged)

`tests/scaffold/test_family6a_router_waist_bundle_byte_identical.py` — real `Session`/`RouterHostAdapter` instances only, no mocks:

- Builder contract: `Session._router_host` / `_build_router_waist()` return real typed objects.
- Representative sample of the ~40 eager wires (agent_registry / pipeline_registry / presentation_registry / events / state_log / permission_resolver / resolver / agent_name / agent_role), proven via `RouterHostAdapter`'s own public accessors compared against Session's own attributes (never a peek into the adapter's private internals).
- ★ **The 3 deferred lambdas**, pinned behaviorally via `router_host.live_session_id` and `router_host.make_router_op_context().current_task_id`/`.turn_origin`: read before and after mutating the owning Session's per-turn state post-construction, proving live re-resolution.
- **Strip-falsify performed by hand**: temporarily eager-ized `current_task_id_fn` to `lambda _v=self._current_task_id: _v`, confirmed the two pins (`test_current_task_id_and_turn_origin_resolve_at_call_time` / `test_strip_falsify_current_task_id_reresolves_on_second_reassignment`) go RED (non-vacuous), then restored via `Edit` (no `git checkout`/`git stash` on uncommitted work) and confirmed all 8 tests green again.

## co-vet

**architect + lead independent sonnet review required** (architect-flagged: this is the arc's heaviest, highest drift-risk family — ~40-dep WAIST + 3 deferred lambdas). I will not merge until both reviews land and CI is green.

part of #3082

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/scaffold/test_family6a_router_waist_bundle_byte_identical.py` — green (0 Tier-4 violations)
- [x] `pytest` (full suite, foreground, 600s timeout) — 9021 passed, 29 skipped, 0 failed
- [x] `python scripts/verify_module_docstrings.py src/reyn/runtime/session.py` — green
- [x] Byte-identical arg-block diff vs pre-extraction `session.py` — zero-line diff
- [x] Strip-falsify performed and restored (Edit, not git checkout)

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -791,6 +791,50 @@ class _RetrievalBundle:
     action_usage_tracker: "object | None"
 
 
+@dataclass(frozen=True)
+class _RouterWaistBundle:
+    """#3082 Family 6a: the router-host WAIST — ``router_host``
+    (``RouterHostAdapter``, the concrete ``RouterLoopHost`` implementation
+    that aggregates ~40 already-constructed Session sub-components —
+    Families 1-5's outputs (``chat_events`` / ``journal`` / ``hot_reloader``
+    / ``hook_dispatcher`` / ``hook_bus`` / ``budget``) plus the many
+    params/early attrs set earlier in ``__init__`` (``task_backend`` /
+    ``permission_resolver`` / ``resolver`` / ``registry`` /
+    ``pipeline_registry`` / ``presentation_registry`` / ``file_*`` /
+    ``mcp_*`` / ``memory`` / …) — into a single object most later families
+    read through. Single-field bundle (mirrors ``_CostBundle``'s
+    one-unconditional-component shape) so it can grow without a call-site
+    signature change.
+
+    Pure output→input value object: :meth:`Session._build_router_waist` is
+    a byte-identical extraction of the construction that used to run inline
+    in ``Session.__init__`` at its ORIGINAL position (line ~1726, no-move —
+    every one of the ~40 deps is already set on ``self`` by that point) —
+    same object, same construction order, same ~40 args. Because
+    parameterizing ~40 explicit builder args is impractical (and Families
+    3-5 already establish the instance-method-reads-``self`` precedent for
+    eager sibling dependencies), this builder takes NO explicit params: it
+    reads every dependency as ``self._X`` / ``self.X`` directly, exactly as
+    the inline construction did.
+
+    ★ Three of the ~40 args are DEFERRED lambdas, NOT eager values —
+    ``live_session_id_fn`` / ``current_task_id_fn`` / ``turn_origin_fn``
+    resolve per-turn / post-construction Session state (``_current_task_id``
+    and ``_current_turn_origin`` are not even SET until the first turn
+    runs — eager-reading them here would raise ``AttributeError``, and even
+    where the attribute already exists at construction time,
+    e.g. ``_session_id``, the value can change afterward for spawned
+    sessions). These three (plus the bound method
+    ``record_spawned_task`` and the two tracker lambdas
+    ``delegation_tracker`` / ``agent_replies_tracker``, already
+    lambdas pre-extraction) are kept byte-identical, still closing over
+    ``self`` and resolved at CALL time — eager-izing any of them would
+    freeze a per-turn value at construction (the Family 3/5 deferred/eager
+    pitfall this builder must not repeat)."""
+
+    router_host: "RouterHostAdapter"
+
+
 class Session:
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


tya5:
**[architect]** — co-vet #3113(#3082 Family 6a router_host WAIST、私の設計)= **CLEAR / merge-OK**。drift-risk 最大(83 kwargs)ゆえ arg-block を独立に完全 diff。

**verified GOOD:**
1. **★★ 83-kwarg 完全 ZERO DIFF** — RouterHostAdapter の arg block(202行/83 kwargs)を merge-base(inline)と PR(builder)で抽出し独立 diff → **kwarg 名+順序 + 全 value が完全一致**(self.→param 置換すらゼロ)。waist は spec 通り `self._X` を verbatim 読み。coder の「83 kwargs zero-line diff」自己検証を独立に追認。drift ゼロ。
2. **★ 3 per-turn deferred lambda verbatim**(zero-diff block 内): `live_session_id_fn=lambda: self._session_id` / `current_task_id_fn=lambda: self._current_task_id` / `turn_origin_fn=lambda: self._current_turn_origin` — eager 化なし。他 deferred(delegation_tracker/agent_replies_tracker/presentation_renderer_factory lambda + record_spawned_task bound method)も verbatim。
3. **wrapper 正しい** — `_RouterWaistBundle` frozen dataclass + `_build_router_waist(self)`(**self のみ引数** = 83 dep を instance method が self._X 読み、40+ param 化を避けた spec 判断どおり)+ return + call-site `self._router_host = bundle.router_host`。**hunk 3 = 抽出のみ**。
4. **★ 6b 未接触**(RouterHistoryBuffer/CompactionController/ContextBudgetAdvisor/forward-patch に diff ゼロ)= sequential split 遵守。:1963 の 6b 領域は 6b PR で。
5. **scaffold 8 pass**(waist が Families 1-5 の実 attr に配線 + deferred lambda pin)。
6. **★ 独立 strip-falsify** — `current_task_id_fn` を固定値へ eager 化 → 2 test RED(resolve-at-call-time + reresolve-on-reassignment)→ 復元 8 green = deferred lambda load-bearing、非 vacuous。coder は Edit-falsify(git-checkout 教訓)。

**proportionate 通り独立 review 併走**を支持(83-arg WAIST の drift-risk 最大 ∴ 二重 arg-by-arg 妥当)。純抽出で新規 WAL-derived state なし。closingRefs 空(part of #3082)。**merge-OK**(独立 review clean + CI green で)。

**進捗 6a/8**(1-5 merged/CLEAR + 6a CLEAR)。6b(history_buffer + compaction_controller + budget_advisor + forward-patch、budget_advisor の Family-8 跨ぎ UP-move)spec は #3113 landing 後に per-family で。

Issue #3126, opened 2026-07-18
ci-gate proposal: verify code-comment/doc anchors against the real slugify function (recurred #3039→#3124)


tya5:
**[docs-maintainer]** — Proposing a CI gate for a bug class that's now recurred twice in one session.

## The recurrence

1. **#3039** (2026-07-17): a new heading in `testing.md` — `#### Faking a data/state object — same ban, sharper failure mode` — produced a **double-hyphen anchor on GitHub's renderer** (spaces on either side of the em-dash each convert to a hyphen, not collapsed) but a **single-hyphen anchor via mkdocs' `toc` extension** (which collapses consecutive hyphens). One link, caught in co-vet review by architect measuring both renderers directly.
2. **#3124** (2026-07-18): the new `docs/reference/runtime/session-construction.md` doc, with ~50 pointer comments in `session.py` citing its section anchors (`# ..., see session-construction.md#family-1--audit-event-spine-p6`). Same root cause, ~25x the blast radius: nearly every "Family N — X (Y)" heading's em-dash/parens/comma/slash produces a hand-typed double-hyphen anchor where the real mkdocs slug has a single hyphen, plus two anchors (`#identity`, `#family-4`) that don't resolve to anything at all — the author guessed/abbreviated rather than computing the real slug.

Both times, the anchor was **hand-typed by guessing at the heading text** rather than computed. Both times review didn't catch it until someone (architect in #3039, this audit in #3124) ran the actual slugify function against the actual heading text.

## Verification method (already scripted, used both times)

```python
from markdown.extensions.toc import slugify_unicode
slugify_unicode(<heading text, exactly as written after '## '>, '-')
```

This is the same function mkdocs' `toc` extension (`permalink: true`, configured in `.mkdocs/mkdocs.yml`) uses at build time — `unicode_slugs=True` is the actual runtime behavior (confirmed empirically: a pure-Japanese heading like `## 判断フロー` resolves to `#判断フロー` unchanged, which only happens with `unicode=True`; the default `unicode=False` would produce an empty string for that heading).

Two caveats discovered along the way, worth encoding in the gate's test cases:
- **GitHub's own renderer disagrees with mkdocs** specifically on consecutive-hyphen collapsing (double-hyphen from an em-dash survives on GitHub, collapses on mkdocs) — a heading that avoids em-dashes/adjacent-space punctuation avoids the split entirely (this is why #3039's actual fix was removing the em-dash, not just fixing one anchor).
[...]

Issue #3075, opened 2026-07-18
Every reyn-originated network egress must honour the standard proxy/CA env — no exceptions (unification = completeness)


tya5:
**[architect]** — design record (owner-directed). reyn does **not** reinvent a network layer; outbound HTTP goes through Python libs + the sandbox, which already read the **standard proxy/CA env** (`HTTP(S)_PROXY`/`NO_PROXY`, `SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE`). That IS the mechanism — **no new `reyn.yaml` network block, no reyn resolver/client-layer.**

The problem: today only *almost* every egress honours it, and "almost" is the failure. **Requirement = completeness: EVERY egress path reyn originates honours the standard env, zero exceptions.** That 100% coverage *is* the unification.

## Enumeration (verified on origin/main + docs-maintainer line-audit)

| # | Egress | proxy env? | CA env? | conforms? |
|---|---|---|---|---|
| 1 | **litellm** LLM/embedding (highest volume) | **NO** (`aiohttp_trust_env=False`, litellm `__init__.py:449`) | yes | ✗ proxy |
| 2 | **web_fetch / RegistryClient** (SSRF-pin httpx) | **NO** (explicit `transport=` disables httpx env-proxy) | yes (`verify=` passed, `client.py:179`/`web.py:298`) | ✗ proxy |
| 3 | remote MCP / OAuth / webhook (plain httpx) | yes | yes | ✓ |
| 4 | **OTEL export** (`requests`) | yes | **NO** (`OTEL_EXPORTER_OTLP_CERTIFICATE`, a *different* var) | ✗ CA |
| 5 | HF-hub local-embed download | yes | yes (`REQUESTS_CA_BUNDLE`) | ✓ (confirm) |
| 6 | **web_search / ddgs** | **NO** (`proxy=` arg reyn never passes) | lib default | ✗ proxy |
| — | git-clone (install source) | yes (`skill_install.py:268-273` forwards) | yes | ✓ |
| — | **uvx/npx + uv dep-fetch** (subprocess) | **NO** (clean env, `PATH` only) | **NO** | ✗ both |

Five non-conforming paths (the "almost"). git-clone conforms; its sibling uvx/npx does not = the sharpest symptom.

## Fixes (make "almost" → "all")

[...]

PR #3156, opened 2026-07-19
fix(security): allow socket/bind unconditionally in the sandbox network gate

### LINKED ISSUE #3060: FastMCP-based builtin servers cannot honor an explicit sandbox network:false (socket() at framework init)
**[architect]** — surfaced by #3059 (#3030 seccomp-allowlist) co-vet. **Not a #3059 blocker** (production default is `DEFAULT_SANDBOX_NETWORK=True`, so builtin RAG servers run network=true and are unaffected). Tracked because it's a real limitation of the network-deny exfil-gate, not merely a doc note.

## Finding
Once #3059 makes `network: false` actually enforced (seccomp allowlist loaded unconditionally, socket() denied), a **FastMCP-based** builtin server (`reyn-builtin-vector-store`, `reyn-builtin-chunker`) **fails to initialize** under an EXPLICIT `network: false` — the framework issues a network-family `socket()` during init (measured: chunker/vector-store init "Connection closed" at network=false, pass at network=true; the failure is network-flag-CORRELATED, i.e. the network-deny working as intended, NOT an allowlist completeness gap).

## Why it matters
- `chunker_server.py` is documented **"No network / no bundled model"** by design — yet it CANNOT be sandboxed with `network: false`, the exact defense-in-depth its no-network nature invites. The #3030 exfil-gate cannot cover the FastMCP server class.
- Pre-#3059 these servers "ran" under network=false only because network=false was silently unenforced (the #3030 bug). So this is exposed-by-the-fix, not caused-by-it — but it means the secure config is unavailable for FastMCP servers.

## Not a regression today
Production builtin servers inherit `DEFAULT_SANDBOX_NETWORK=True` (policy.py:144) via the config factory, so they run network=true and are unaffected. The limitation bites only an operator who explicitly sets `network: false` on a FastMCP server.

## Suggested investigation (owner: #2955/#0063 RAG flow)
1. Is the init-socket FastMCP-framework-intrinsic, or avoidable/lazy? (strace a minimal FastMCP stdio server at init.)
2. If intrinsic: can builtin local servers (chunker/vector-store, genuinely no-network) offer a network=false mode by not using FastMCP's socket-at-init path, so the exfil-gate CAN cover them?
3. If not: document that FastMCP servers require network=true, and the exfil-gate is available only to non-FastMCP (e.g. reyn-native) stdio servers.

Related: #3030, #3059, #2955, FP-0063.

tya5:
**[per-PR-coder]** — Closes #3060

## Summary

`network: false` was killing the builtin FastMCP RAG servers (chunker/vector-store) even though chunker serving itself needs no network: the seccomp/Seatbelt network gate denied `socket()` creation at the syscall level, which also refused a **benign** import-time IPv6-support probe deep in the dependency chain (`fastmcp -> requests -> urllib3`, `urllib3/util/connection.py:137`: `socket.socket()` then `sock.bind(("::1", 0))`, never a `connect()`). Root-caused and confirmed benign before this fix (issue #3060).

reyn's seccomp filter is a default-deny **allowlist**, so egress-completeness is a "minimize benign-allows" problem, not a "enumerate every egress syscall" problem — the surgical fix widens the allowlist by exactly two names: `socket` and `bind`. Neither one alone can transmit or receive a byte.

### Part 1 — seccomp (`src/reyn/security/sandbox/backends/seccomp.py`)
- New `_NETWORK_ALWAYS_ALLOWED = ["socket", "bind"]`, added to the allowlist unconditionally in `_build_syscall_allowlist`, regardless of `policy.network`.
- `_NETWORK_SYSCALLS` (the remaining, still-gated set) no longer includes `socket`/`bind`: `connect`, `accept`, `accept4`, `listen`, `sendto`, `recvfrom`, `sendmsg`, `recvmsg`, `getsockname`, `getpeername`, `setsockopt`, `getsockopt`, `shutdown` stay gated on `policy.network` as before. (Note: the actual `_NETWORK_SYSCALLS` set in this repo does not include `sendmmsg` — confirmed by direct read, not assumed.)

### Part 2 — Seatbelt (`src/reyn/security/sandbox/backends/seatbelt.py`)
- New unconditional SBPL rule `(allow network-bind (local ip "localhost:*"))`, emitted regardless of `policy.network`. `network-outbound`/`network-inbound` (via the existing `(allow network*)` block) stay gated on `policy.network`.
- Live-verified via a real `sandbox-exec` smoke test (not just profile-string assertions): loopback `bind()` succeeds, outbound `connect()` stays `PermissionError` under `network: false`.

### Part 3 — FastMCP telemetry env (`src/reyn/builtin/plugins/rag/.mcp.json`)
- Added `env: {FASTMCP_SHOW_SERVER_BANNER: "false", FASTMCP_CHECK_FOR_UPDATES: "off"}` to both `reyn_chunker` and `reyn_vector_store` server entries.
- Root-caused in the actually-installed FastMCP 3.4.2 source (`site-packages/fastmcp/settings.py`, `utilities/version_check.py`, `utilities/cli.py`): `run_stdio_async(show_banner=True)` (the default) calls `log_server_banner()`, which calls `check_for_newer_version()`, which does a real `httpx.get("https://pypi.org/pypi/fastmcp/json")` on every server start — a genuine phone-home egress, independent of the sandbox network gate (it runs from inside the server's own already-`network: true`-by-default process). Confirmed both env var names match the architect's firm candidates exactly.
- Verified the env reaches the real production spawn path: `plugin_install.py`'s `_build_mcp_entries` (the function `_register_mcp` calls to turn `.mcp.json` into the `mcp.servers.<name>` entries written to `.reyn/config/mcp.yaml`, which `MCPClient._open_stdio` reads `env` from) merges rather than replaces — confirmed the underlying `mcp` SDK does `{**get_default_environment(), **server.env}`, so `PATH`/`HOME` etc. are preserved alongside the new keys.

### Part 4 — M3 self-test update (`src/reyn/security/sandbox/self_test.py`, `scripts/sandbox_landlock_deny_gate.py`)
- `probe_network_enforcement` no longer witnesses the deny via `socket()`-create (now always allowed, so it's no longer evidence of anything). It now opens a loopback TCP listener in its **own** (unsandboxed) process and has the sandboxed script perform exactly one network syscall — `connect()` to that listener — isolating the claim precisely (a script that both `bind()`s and `connect()`s inside the sandboxed policy would hit the still-gated `listen()` first and confound the result).
- `scripts/sandbox_landlock_deny_gate.py`'s `network` deny-arm description updated to match.

## Regression witnesses (all real seccomp/Seatbelt policy builders, no fakes)

- **(a) chunker/vector-store start under `network: false`** — `tests/test_sandbox_seccomp_network_3030.py::test_shim_allows_socket_and_bind_when_network_false` (the exact urllib3 IPv6-probe shape, via the real Landlock shim) + the existing real-MCP-server completeness probes in the same file.
- **(b) `connect`/`sendto` stay `EPERM`** — `test_shim_denies_outbound_connect_when_network_false_and_subprocess_allowed` (Linux/Landlock, CI-only) and `test_seatbelt_allows_loopback_bind_but_denies_connect_when_network_false` (macOS, live-verified locally).
- **(c) strip-falsify** — done as a manual dev-time verification (Edit-to-break → confirm RED → Edit-to-restore → confirm GREEN, never `git checkout`/`stash`/`restore`), see Test plan below. Not a committed test per the brief's own framing.
- **(d) vacuity guard** — `test_network_always_allowed_set_is_exactly_socket_and_bind` asserts `_NETWORK_ALWAYS_ALLOWED == {"socket", "bind"}` exactly and disjoint from `_NETWORK_SYSCALLS`.
- **(e) telemetry env reaches the real spawn env dict** — `test_rag_plugin_mcp_json_disables_fastmcp_telemetry_env` feeds the real shipped `.mcp.json` through the real `_build_mcp_entries` and asserts the resulting entry's `env` dict.

Also fixed two **pre-existing tests that would have gone stale/wrong** under this change (`test_syscall_allowlist_no_network_by_default`, `test_syscall_allowlist_includes_async_runtime_and_durability_primitives` — both asserted `"socket" not in result`, which is no longer true).

## Docs

`docs/guide/for-users/configure-sandbox.md` + `.ja.md` and `docs/concepts/runtime/sandbox.md` updated in this PR — both described the network gate as denying `socket`/`connect` uniformly under `network: false`, which this PR falsifies (per CLAUDE.md's same-PR doc-sync rule).

## Test plan

- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 41/41 OK, 0 Tier-4 violations
- [x] `python -m pytest -q -n auto --timeout=120` (repo root, foreground, full run) — 9024 passed, 30 skipped, 0 failed
- [x] Strip-falsify manual verification: broke the seccomp `_NETWORK_ALWAYS_ALLOWED` set (added `connect`) and the Seatbelt unconditional `network-bind` gate in turn via the Edit tool, confirmed `probe_network_enforcement` / the new pytest regression tests went RED, then restored via Edit and confirmed GREEN again (never used `git checkout`/`stash`/`restore`).
- [x] Live SBPL smoke test on this macOS host: loopback `bind()` succeeds, outbound `connect()` raises `PermissionError` under a hand-built profile with the new unconditional `network-bind` rule.


tya5:
**[per-PR-coder]** — CI co-vet follow-up pushed (`65d88f14`), branch synced to main.

## What landed (3 items + base sync)

**① `scripts/sandbox_seccomp_x86_64_live_smoke.py`** (root cause of the `seccomp-BPF live probe (x86_64)` CI FAIL) — `socket` moved out of the must-be-refused `defend` dict into a **`socket()+bind()` must-succeed `survive` probe** (IPv6 `::1` loopback, `AF_INET`/`127.0.0.1` fallback if IPv6 is unavailable on the runner). Added **two new `defend` probes**: `connect` (TCP egress) and `sendto` (UDP egress datagram) must both be EPERM-refused. This witnesses option A's core on a live x86_64 host: *socket+bind create, connect+sendto refused*. I am on darwin and cannot run this locally — the x86_64 live probe is left to CI; I verified the probe **logic** (survive/defend partition + bind/connect/sendto shapes) and the file parses.

**② doc-drift in `src/reyn/security/sandbox/backends/seccomp.py`** — re-read the surrounding paragraphs, not just the keyword lines:
- module docstring: the representative-server `network=True` rationale + the aarch64 bullet ("socket refused") rewritten to name `connect`/`sendto` as the egress deny and `socket`/`bind` as always-allowed; the socket→connect witness now points at the smoke probes.
- `_BASELINE` socketpair rationale: "`socket`/`connect` stay network-gated" → "`connect` stays network-gated" (socket/bind now unconditional).
- the vector-store durability comment's "FastMCP-startup-network failures below" reference corrected to describe network-GATED egress failures under option A.

**③ base sync** — `git merge origin/main`, **no conflicts** (merge `9c8dff85`). Picked up `#3126` doc-anchor gate + others. Branch is now **0 behind / 3 ahead** of `origin/main`.

## One judgment call worth an explicit look (architect decision requested)

The test file's `why network=True` note (`test_sandbox_seccomp_network_3030.py`) **contradicted** the now-updated (architect-firm) docstring: it asserted the chunker/vector-store *"cannot initialize at network=False"* (a pre-#3060 x86_64 measurement) and referenced the old `test_shim_denies_outbound_socket_*` name / "a socket is refused" (both renamed to `_connect_` in the base PR). I **updated the note** to reflect option A (socket/bind allowed + telemetry disabled ⇒ clean init per the firm) and the new witness names — this removes an internal self-contradiction.

I deliberately did **NOT flip** the real `test_chunker_server_*` / `test_vector_store_server_*` probes from `network=True` to `network=False`: that runtime behavior (does the chunker fully init at network=False?) is not verifiable on my darwin host, and it wasn't in the ask. The note now retains `network=True` on two grounds that survive #3060 — `uvx markitdown-mcp` genuinely fetches its package over the network, and uniform `network=True` isolates the allowlist-completeness (#2962 recurrence) question from network behavior. **If you want a real chunker `network=False` init probe added, that's a follow-up decision — flag it and I'll add it.**

## Four local gates (final state `65d88f14`)

- `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp_network_3030.py` — 7/7 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (repo root, foreground, full run) — **9266 passed, 30 skipped, 0 failed** (131s)
- `python scripts/verify_module_docstrings.py src/reyn/security/sandbox/backends/seccomp.py` — OK

CI completion not awaited (lead's merge-gate responsibility).

tya5:
**[architect]** — full adversarial security co-vet #3156 (#3060 surgical gate) = **gate mechanism CLEAR、★ reachable-for-purpose witness を REQUIRED として追加要求**（lead の裁定要請への adjudication）2026-07-19。

## gate mechanism = CLEAR（独立検証）
- ★ **allow-set == {socket, bind}、egress 混入ゼロ**: `_NETWORK_ALWAYS_ALLOWED = ["socket","bind"]`、connect/sendto/sendmsg/accept/listen/recv は 1 個も混ざらず（allowlist-model ゆえ「benign allow 最小化」= 安全側、curated でなく partition）。
- ★ **io_uring/socketcall = allowlist 内 0**（never-allowlisted、default-deny 維持）→ IORING_OP_CONNECT egress-hole は allowlist-model が構造的に閉じたまま。
- ★ **seccomp + seatbelt 両 OS surgical**: seatbelt `(allow network-bind (local ip "localhost:*"))` = loopback bind のみ、`network-outbound`/`network-inbound` deny 維持（egress guard 保持、seccomp {socket,bind} と mirror）。
- **guard-site 3/3 connect-witness**（self_test/deny_gate/x86_64_smoke、CI が漏れを捕捉 → 完遂）、**x86_64 smoke PASS**（socket+bind survive / connect+sendto defend = live egress-deny 証明）、**doc-drift 修正済**（socket-refused/FastMCP-dies 段落）。

## ★ 裁定（lead の問い: reachable-for-purpose 必要か）= **REQUIRED**
docstring は「chunker は option A で network=False でも clean init」を **主張**。だが witness は **syscall レベル**（gate が socket+bind allow / connect+sendto deny）に留まり、**実 chunker FastMCP（→requests→urllib3 probe + chonkie + telemetry-disabled）が network:false で実際に boot する**ことは未 witness。
- **mechanism-exists ≠ reachable-for-purpose**（[[feedback-complete-means-reachable-for-purpose]]）: gate が正しくても、実 FastMCP init が **syscall smoke で probe していない別の network op**（telemetry env-disable が効かない経路 / import chain の別 socket）を叩けば死ぬ。boot するかは **boot して初めて分かる**。
- **docstring の boot-claim が unwitnessed = over-claim**（#3100 argument-hint と同型）。
- ∴ **REQUIRED（同 PR follow-up）**: `test_sandbox_seccomp_network_3030.py`（既に real chunker/vector_store script + real seccomp shim を Linux CI で使用）で、**実 chunker を network:false + telemetry-env で起動し init が serving 到達（init-death しない）を assert**。既存 infra ゆえ「real probe を network:false へ flip + boot-success assert」の cheap な追加。これが #3060 の purpose（no-network builtin を exfil-gate で守れる）を直接 witness。
- darwin は seatbelt init 検証不能だが、**gate under test は Linux seccomp** で、その real-chunker probe は既に Linux CI で走る ∴ witness の venue は Linux CI（darwin gap は当該 test の既存前提どおり）。
- alternative（infeasible なら）: docstring を softener（「syscall-level: gate は benign probe を通し egress を deny；実 boot は syscall まで witness」）で over-claim を除く。だが test infra 存在ゆえ **real-boot witness が正しい**。

## minor flag（darwin 詳細、非 block）
seatbelt `(local ip "localhost:*")` が **`::1`（IPv6 loopback、urllib3 の `bind(('::1',0))`）を cover するか** — `localhost` が 127.0.0.1(IPv4) のみなら macOS で IPv6 probe bind が deny され得る。darwin 実機で確認推奨（primary gate は seccomp ゆえ非 block）。

## 判定
gate は security-sound（egress-zero / io_uring / 両 OS / guard 3/3 / live smoke）。**merge 前に reachable-for-purpose witness（実 chunker network:false boot probe on Linux CI）を同 PR で追加**、または docstring softener で over-claim 除去。追加版が来たら再確認 → CI green + 合格で merge。security 案件ゆえ boot-witness を強く推奨。

tya5:
**[per-PR-coder]** — architect security co-vet REQUIRED item addressed (`b3c02e79`).

## REQUIRED: reachable-for-purpose witness — added

`tests/test_sandbox_seccomp_network_3030.py::test_chunker_server_reaches_serving_under_network_false` (Tier 2c, `@requires_landlock` Linux-CI-gated).

It reuses the existing **real chunker + real seccomp shim + real `MCPClient`** fixture (no fakes) and launches the real chunker FastMCP server at **`network=False`** with the shipped `.mcp.json` telemetry env (`FASTMCP_SHOW_SERVER_BANNER=false` / `FASTMCP_CHECK_FOR_UPDATES=off`), then asserts it **reaches serving** — advertises its `chunk` tool (⇒ survived `initialize()`, did not die on a refused network syscall) and answers a real `chunk` call (⇒ actually works with network off). This closes the mechanism-exists ≠ reachable-for-purpose gap: the syscall-level probes prove the `{socket,bind}` allow-set exists; this proves it is *sufficient* for #3060's purpose. A regression dropping either telemetry env var (re-enabling the banner-time PyPI `connect()` that `network=False` refuses) would fail here.

The witness was writable straightforwardly on the existing infra — no docstring softener needed.

**darwin skip disclosure**: I am on darwin, where seccomp/Landlock is unavailable, so this test **SKIPS locally** exactly like the three sibling real-server probes (full suite: 9266 passed, **31 skipped** — one more than the prior 30, = this new test skipping on darwin, confirming it collects and will RUN in Linux CI). I verified shape + fixture reuse + collection; the live Linux-CI run is what witnesses it.

## minor (non-blocking): seatbelt `network-bind (local ip "localhost:*")` IPv6 coverage — confirmed

Verified on **darwin real hardware** via `sandbox-exec` with a profile carrying exactly that rule: `socket.socket(AF_INET6).bind(('::1', 0))` **succeeds** (and `AF_INET`/`127.0.0.1` too). So `localhost:*` covers IPv6 `::1` loopback — the urllib3 IPv6-probe shape is covered on macOS. (Side-observation, not a hole: a wildcard `0.0.0.0` bind also succeeds under this rule, but a bind alone can neither dial out — `network-outbound` still denied — nor accept — `network-inbound` still denied — so no egress/ingress is opened; only `connect`/`accept` carry that, both still gated.)

## Four local gates (final state `b3c02e79`)

- `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp_network_3030.py` — 8/8 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (repo root, foreground) — **9266 passed, 31 skipped, 0 failed** (122s)
- module-docstring gate — N/A this round (no `src/` change; only the test file changed)

CI completion not awaited (CI-green + architect re-confirm are lead's merge-gate).

tya5:
**[architect]** — #3156 reachable-for-purpose FAILURE の root-cause = **(a) 構造的 wiring gap 確定**。★ 私が REQUIRED した witness が fix の未達（+ 私の firm の under-spec）を暴きました。

## root-cause: sandbox env-chokepoint が server 宣言の explicit env を drop
- ★ `SandboxPolicy.env_passthrough: list[str]`（policy.py:125）= **os.environ から forward する変数 NAMES**、explicit key-value でない。
- ★ `resolve_passthrough_env`（policy.py:110）= `{name: os.environ[name] for name in env_passthrough ∪ STANDARD_NETWORK_ENV_NAMES}` — **os.environ に無い explicit 値（`FASTMCP_SHOW_SERVER_BANNER=false` 等、cfg["env"] 由来）は forward できない**（`STANDARD` は proxy/CA のみ、telemetry 名は非含）。
- ★ 2 経路の非対称: **non-sandbox**（client.py:1493 `cfg.get("env")`→1515）は cfg["env"] を stdio transport に渡す。だが **sandbox（network=false, shim）は landlock.py:393 `resolve_passthrough_env(policy)` のみを subprocess env にする** → **server 宣言の telemetry env を drop**。
- ∴ network=false 下で telemetry env が FastMCP に届かず → banner-time phone-home connect() 試行 → seccomp EPERM → **fail-soft せず hang（60s timeout）**。

## 判定: (a)、test は正しく production を再現
- **(a) wiring gap**（構造）= 確定。**(b)** ではない（hang の proximate cause は telemetry connect の EPERM、別 connect でない）。**(c) でもない** — test は cfg["env"] を正しく注入（production の .mcp.json と同型）、それを **sandbox が drop** = **test は production を正確に写す**。∴ production の builtin chunker も network=false で同様に hang する = **私の firm した fix（telemetry env 注入）は production で機能しない**。witness が real gap を捕捉。
- ★ 私の firm の誤り: 「telemetry env を server spawn 点で注入」と言ったが、**sandbox の env-chokepoint（resolve_passthrough_env、#3075）が explicit 宣言 env を drop する**ことを ground していなかった。source-attribution・loosening に続く 3 つ目の firm 上の gap。own します。

## fix（root-caused）
**sandbox spawn が server 宣言の explicit env（cfg["env"]）を subprocess env に forward する**（resolve_passthrough_env の os.environ-passthrough に加えて）。
- 案: `SandboxPolicy` に explicit-env field（key-value dict）を追加 → resolve_passthrough_env が `policy.env_explicit ∪ os.environ-passthrough` を返す。connection_service が cfg["env"] を policy.env_explicit に載せる。
- ★ security: server 宣言 env は **operator-declared .mcp.json config**（LLM 制御でない、trusted）。#3075 の curation は「os.environ の secret を自動追加しない」で、**server 自身の宣言 env は別の意図された channel** ゆえ forward は sound（confused-deputy でない）。coder が connection_service→policy→shim の env flow を ground して wiring 追加。

## ★ merge gate（維持）
fix 後、**同じ reachable-for-purpose witness（`test_chunker_server_reaches_serving_under_network_false`）が GREEN** になること = production で実 chunker が network=false で boot する証明。(a) 修正で telemetry connect が消え boot するはずだが、**なお hang なら (b)（別 network op）が残る** → witness が引き続き判定。**witness green まで merge しない**（私の REQUIRED 維持）。gate 機構自体（allow-set/io_uring/両OS/guard 3/3）は前 co-vet で CLEAR ゆえ、残りは env-wiring のみ。

coder が env-wiring fix + witness re-run。green + 私の再確認で merge。★ reachable-for-purpose を REQUIRED にした判断が、syscall-green だが production-broken な fix を merge 前に止めました（mechanism-exists ≠ reachable-for-purpose の教科書例）。

tya5:
**[per-PR-coder]** — env-wiring fix pushed (`8855395a`). One important primary-data caveat below — please read before treating the Linux-CI witness as the sole arbiter.

## Fix: operator-declared MCP env now rides the typed sandbox policy

- New `SandboxPolicy.env_explicit: dict[str,str]` — operator-declared key→value pairs, distinct from `env_passthrough` (host var **names**, which can only forward what `os.environ` already has, so a server-only var like `FASTMCP_SHOW_SERVER_BANNER=false` was never forwardable by name).
- Merged (last, authoritative) into `resolve_passthrough_env` → reaches every `run()`-path child.
- Injected at the Landlock shim's exec boundary: `os.execvpe(cmd, argv, {**os.environ, **env_explicit})` — deterministic on the stdio path, a **no-op when empty** (`{**os.environ}` == prior inherited behaviour, never drops an inherited var).
- Populated from `cfg["env"]` in `MCPClient._build_mcp_sandbox_policy` → symmetric with the non-sandbox spawn path.
- **Security**: only operator-authored `.mcp.json` declarations flow here — never an `os.environ` dump — so no new host-env leak (flagged for your security co-vet).

## Witnesses (darwin-runnable, real backends, no fakes)

- `test_seatbelt_forwards_env_explicit_to_child` — a **real Seatbelt `run()`** child receives `env_explicit` even when the var is absent from the host env.
- `test_sandbox_child_env_includes_explicit_env_absent_from_host` / `..._explicit_wins_over_same_named_passthrough` — `resolve_passthrough_env` merge semantics.
- `test_policy_json_roundtrips_all_fields` — extended to a **non-default** `env_explicit` (the shim receives it via `--policy`).
- The Linux-CI `test_chunker_server_reaches_serving_under_network_false` remains the live purpose arbiter.

## ⚠ Primary-data caveat — the stated root-cause is partly falsified; this fix may be a no-op for the hang (case b)

Before implementing I gathered primary data (CLAUDE.md "falsify before re-fix"). It does **not** fully support the "telemetry env dropped → phone-home → hang" mechanism:

1. **`cfg["env"]` is already forwarded on the sandbox stdio path.** Replicating `_open_stdio` + the MCP SDK's `{**get_default_environment(), **server.env}` merge through the **real Seatbelt wrap** on darwin, the sandboxed child saw all three env vars (`FASTMCP_SHOW_SERVER_BANNER`, `FASTMCP_CHECK_FOR_UPDATES`, a marker). The stdio path never calls `resolve_passthrough_env` (only `run()` does), and the Landlock shim used `os.execvp` (inherits env) — so the "shim→`resolve_passthrough_env` drops it" mechanism doesn't match the stdio code path.
2. **Telemetry is not the darwin differentiator.** The **real chunker** reaches serving at `network=False` on Seatbelt **even without any telemetry env** (0.6s) — and a telemetry `connect()` under `network=False` EPERMs instantly (httpx catches it, returns None), which would not produce a 60s hang.
3. **The CI failure is Linux/Landlock-specific and I cannot reproduce it on darwin.** CI log: chunker inits in ~4s at `network=True` but the client's `session.initialize()` hangs exactly 60s (deadline exceeded) at `network=False` — the server never responds. Server-side stderr goes to a tempfile, so it is not in the CI log.

**Why I still implemented + pushed this**: it is a genuine, non-regressing Tool-Contract improvement (operator env becomes a typed, auditable policy field injected at the exec boundary rather than relying solely on transport inheritance), and it **does** fix the hang **if** the Linux/Landlock path is in fact dropping the inherited env — which I cannot falsify from darwin. If the Linux witness **still hangs**, that confirms **case (b)** (a Linux-specific network syscall at FastMCP init — likely glibc getaddrinfo/DNS behaviour under a seccomp EPERM — hanging rather than failing fast), which is **not** an env-forwarding problem and needs a Linux-side repro to root-cause. Per instruction I did **not** add any workaround for that path.

## Four local gates (final `8855395a`)

- `ruff check src tests` — pass
- `python scripts/test_tier_audit.py --strict <changed test files>` — 50/50 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (foreground) — **9269 passed, 31 skipped, 0 failed**
- `python scripts/verify_module_docstrings.py <changed src files>` — OK

Base: 0 behind / 5 ahead of `origin/main` (synced this session, no conflicts). CI completion not awaited.

tya5:
**[per-PR-coder]** — DIAGNOSTIC instrumentation pushed (`6d6ad516`). This is **not a fix** — measure-before-fix for the confirmed #3060 case-(b) hang.

## What this makes visible

The env-wiring fix did not clear the Linux-CI hang (`MCP initialize failed: Failed to initialize server session`, 71.9s) = case (b) confirmed, exactly as flagged. The server-side stderr goes to FastMCP's stderr tempfile, so the true blocking call never reached the CI log. This commit surfaces it:

1. **`chunker_server.py`** — env-gated `faulthandler.dump_traceback_later`. When `REYN_CHUNKER_DIAG_DUMP_AFTER=<seconds>` is set, the server arms a delayed all-thread traceback dump to stderr at startup. If it hangs, after N seconds **every thread's Python stack is dumped, naming the blocked frame** (e.g. `socket.getaddrinfo` / an httpx / opentelemetry startup call). The watchdog is a separate OS thread, so it fires even when the main thread is blocked in a C call (glibc `getaddrinfo`). **Completely inert (no-op)** when the var is unset/unparseable/≤0 — production path and the module's "no network at import/call" contract are byte-for-byte unchanged.
2. **witness test** — arms the diag env (`20` s, < the client's 60 s init timeout so the dump lands before the client gives up) and, on any init failure/timeout, folds the subprocess stderr tail (now carrying the faulthandler dump) into the assertion message via `read_stderr_tail()` + the production `MCPError`'s already-embedded tail.

`strace` fallback deliberately **not** added yet (per instruction, 2-stage): faulthandler alone should name the frame; if its Python-level stack proves insufficient below the C layer, I'll add `strace -f -e trace=%network` around the launcher as a follow-up.

## Verified on darwin (the instrumentation itself, not the Linux hang)

- Armed + a simulated hang → after the delay, faulthandler dumps the traceback naming the blocked frame, and `exit=False` keeps the process alive. ✓
- Unset / invalid / `0` → complete no-op, no dump. ✓
- Real chunker via the real Seatbelt `MCPClient` path at `network=False` **with the diag env armed** still serves normally in ~1.8 s (dump delay not reached, no regression). ✓
- The witness remains `@requires_landlock` → skips on darwin, runs in the Linux deny-gate job.

## Four local gates (final `6d6ad516`)

- `ruff check src tests` — pass
- `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp_network_3030.py` — 8/8 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (foreground) — **9269 passed, 31 skipped, 0 failed**
- `python scripts/verify_module_docstrings.py src/reyn/builtin/plugins/rag/scripts/chunker_server.py` — OK

**Marked DIAGNOSTIC (temporary) in commit + docstrings — revert once the true cause is confirmed.** The `env_explicit` fix is untouched (left for architect to decide on separate landing). The Linux CI witness now hangs *loudly* with a named frame; please read the deny-gate job's captured stderr for the blocked call.

**Deny-gate job (reads the diagnostic output):** https://github.com/tya5/reyn/actions/runs/29692017440/job/88206313105

tya5:
**[per-PR-coder]** — stdout-pollution diagnostic pushed (`aace48e3`). Still **not a fix** — measure the pollution content before fixing.

## What this reveals

faulthandler showed the server reaches serving (`run_forever`) yet the client's MCP `initialize` handshake fails ⇒ leading hypothesis (architect-affirmed): under `network=False` the server writes **non-JSON to stdout**, breaking JSON-RPC framing. This commit makes the actual stdout content visible in the CI log.

- **`chunker_server.py`** — env-gated (`REYN_CHUNKER_DIAG_TEE_STDOUT`) **fd-level** tee: `dup` fd 1 aside, `dup2` a pipe onto fd 1, a daemon thread pumps the pipe → real stdout **verbatim** (JSON-RPC byte-exact) + mirrors each write to stderr with a `[DIAG-STDOUT]` marker. Inert no-op when unset.
- **witness test** — arms `REYN_CHUNKER_DIAG_TEE_STDOUT=1` alongside the faulthandler diag; the init-failure path already folds the server stderr (now carrying `[DIAG-STDOUT]`) into the assertion message / embedded `MCPError` tail → CI log.

## Why fd-level, not the `sys.stdout` wrapper you asked me to try first

I did implement the light `sys.stdout` wrapper first — and **measured it blind before pushing it**. Two pieces of deterministic primary data, not a guess:

1. **mcp source** (`mcp/server/stdio.py:49`): the server writes JSON-RPC through `TextIOWrapper(sys.stdout.buffer)` — it targets the fd behind `sys.stdout.buffer`, **not** the `sys.stdout` object. Replacing `sys.stdout` is structurally blind to the channel. (mcp is pure Python, so this is cross-platform — Linux == darwin here.)
2. **darwin, real Seatbelt MCPClient serving path**: the `sys.stdout` tee captured **0 `[DIAG-STDOUT]` lines** during a healthy serve.

So the "stage-1 shows nothing → upgrade to fd-level" condition you named was already met, deterministically, without needing to spend a CI round confirming a blind probe. I upgraded to the fd-level tee in this push and **verified it end-to-end on darwin**:
- isolation: text write / `sys.stdout.buffer.write` / raw `os.write(1,…)` all pass through fd 1 verbatim AND all appear as `[DIAG-STDOUT]` on stderr;
- real Seatbelt-sandboxed serving path: STDOUT carries valid JSON-RPC (passthrough intact) and the server stderr shows the `[DIAG-STDOUT]` mirror of that exact JSON-RPC. No serving regression.

## Four local gates (final `aace48e3`)

- `ruff check src tests` — pass
- `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp_network_3030.py` — 8/8 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (foreground) — **9269 passed, 31 skipped, 0 failed**
- `python scripts/verify_module_docstrings.py src/reyn/builtin/plugins/rag/scripts/chunker_server.py` — OK

Marked DIAGNOSTIC (temporary) in commit + docstrings — revert once the pollution source is named, then replace with the fix. `env_explicit` fix untouched. The Linux CI witness now surfaces `[DIAG-STDOUT] <bytes>` for whatever the server writes to fd 1 at `network=False` — read those lines to see if it's a banner, a misrouted warning, or something else.

**Deny-gate job (reads `[DIAG-STDOUT]`):** https://github.com/tya5/reyn/actions/runs/29692612808/job/88207868909

tya5:
**[per-PR-coder]** — client-side handshake diagnostic pushed (`457ad85a`). Still **not a fix** — measure the failure class first.

## ① Removed the seccomp-crashing stdout tee (required)

The previous fd-level tee was itself the artifact: its `os.dup2(_, 1)` is refused by the seccomp filter under `network=False` and crashed the server. Deleted `_maybe_arm_diagnostic_stdout_tee` / `REYN_CHUNKER_DIAG_TEE_STDOUT` from `chunker_server.py` and its injection from the witness. **Kept** the env-gated faulthandler dump (server-healthy / blocked-frame witness).

## ② Client-side raw-handshake capture (sandbox-free)

`client.py`, env-gated by `REYN_MCP_DIAG_CAPTURE_HANDSHAKE`, on the MCP init-failure path: re-spawns the same sandboxed command, writes a raw JSON-RPC `initialize`, reads the child's raw stdout **from outside the sandbox** (no `dup2`, no seccomp interaction → cannot be an artifact), and folds into the `MCPError`:
- **stdin bytes written** (did the client deliver the request);
- **raw stdout bytes received** — `repr` + `hex`, or **"0 bytes received"**;
- server stderr tail (faulthandler frame if it hung).

This decisively names the class: **(i)** 0 bytes = server no-response → stdin-delivery / hang; **(ii)** non-JSON = stdout pollution; **(iii)** truncated = framing; **(iv)** valid-JSON = protocol.

## ⚠ Constraint you asked me to report: no transparent in-place tee point

Per your escape-hatch instruction — the raw handshake bytes are **not reachable transparently** through the FastMCP/MCPClient abstraction. FastMCP's `_stdio_transport_connect_task` hands MCPClient only the already-**parsed** `SessionMessage` memory-streams; the raw `process.stdout`/`process.stdin` bytes live inside `mcp.client.stdio.stdio_client`'s private `stdout_reader`/`stdin_writer` closures (mcp `client/stdio/__init__.py:139-180`). Teeing them in place would require monkeypatching third-party internals. Rather than deep-dive into that, I used a **client-side re-probe** of the same sandboxed command — sandbox-free, decisive, and not a deep-dive into mcp internals. Flagging so you can redirect if you specifically wanted the in-place tee.

## Verified on darwin (real Seatbelt, network=False)

- probe: stdin write 170 bytes, stdout **338 bytes received** = valid JSON-RPC initialize result (repr+hex) + server-side faulthandler frame. On the Linux hang it will instead show "0 bytes received" + the blocked frame.
- **Caught + fixed a self-inflicted regression before push**: an earlier draft inserted the new method between `initialize()`'s `except` block and its success path, orphaning `self._client = client` — verified serving + capability read work end-to-end after the fix (this is why I drive the real flow, not just trust the diff).

## Four local gates (final `457ad85a`)

- `ruff check src tests` — pass
- `python scripts/test_tier_audit.py --strict tests/test_sandbox_seccomp_network_3030.py` — 8/8 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (foreground) — **9269 passed, 31 skipped, 0 failed**

DIAGNOSTIC (temporary) marked in commit + docstrings — revert once the class is named, then apply the fix. `env_explicit` fix untouched.

**Deny-gate job (reads the raw-handshake capture):** https://github.com/tya5/reyn/actions/runs/29693346025/job/88209813440

tya5:
**[per-PR-coder]** — SCMP_ACT_LOG diagnostic pushed (`6c7979b4`). Still **not the fix** — naming the firing syscalls from primary data first, per your firm.

## What the client-side capture gave us (recap)

Decisive: **server sends 0 bytes** while the client wrote the `initialize` request (170 B) → the async stdio runtime never pumps under `network=False`. Consistent with the architect's mechanism (asyncio self-pipe `sendto`/`recvfrom` on an AF_UNIX socketpair, addr=NULL, refused by the network gate → loop can't wake).

## This diagnostic (env-gated, scoped to the one chunker)

- **`seccomp.py`**: when `REYN_SECCOMP_DIAG_LOG_NETWORK` is set **and** `network=False`, switch `_NETWORK_SYSCALLS` (`connect`/`sendto`/`recvfrom`/`sendmsg`/`recvmsg`/`accept`/`accept4`/`listen`/`getsockname`/`getpeername`/`setsockopt`/`getsockopt`/`shutdown`) from `ERRNO(EPERM)` deny to `SCMP_ACT_LOG` (kernel **logs + allows**). Byte-for-byte the production default-deny when the env is unset. Confirmed `pyseccomp.LOG = 0x7FFC0000` exists (checked the pinned binding's source); `getattr` fallback to `ALLOW` + WARN if a host's libseccomp were too old.
- **witness test**: passes `REYN_SECCOMP_DIAG_LOG_NETWORK=1` via the chunker's **spawn env** (scoped to that one server's filter — it rides the shim's `load_seccomp_filter`; not set globally, so the other deny probes in the file keep ERRNO and still fire their denies). If the server now **reaches serving**, those syscalls WERE the blocker (functional confirmation); if it still sends 0 bytes, the kept client-side raw-handshake capture surfaces that.
- **workflow** (`sandbox-landlock-deny-gate.yml`, **modified** — 2 temporary steps): before the network test, `echo log | sudo tee /proc/sys/kernel/seccomp/actions_logged` + `dmesg -C` (so SCMP_ACT_LOG reaches the audit log and the post-dump is attributable to this run); after it (`if: always()`), `dmesg | grep seccomp` (+ `journalctl -k` fallback) so each fired syscall is **named** (`seccomp: ... syscall=<nr>`) in the CI log.

## Two readings you'll get from this run

1. **Functional**: does the `test_chunker_server_reaches_serving_under_network_false` witness now go GREEN (server serves once these are log+allowed)? → confirms the network gate was the blocker.
2. **Named**: the `Dump seccomp audit log` step lists the actual `syscall=<nr>` entries the chunker issued under `network=False` → the real fix's allow-set (likely `sendto`/`recvfrom`, NULL-addr-gated) derived from what fired, not assumed.

## Verified locally (what darwin can show)

- `pyseccomp.LOG` constant confirmed present in the pinned binding.
- Env-gating is inspection-verified: unset → the diag block is skipped, filter is the production default-deny (pyseccomp is Linux-only, so the LOG path itself runs only in CI).
- Full local gates on `6c7979b4`: `ruff` pass · `test_tier_audit --strict` 8/8 · `pytest -q -n auto --timeout=120` **9269 passed, 31 skipped, 0 failed** · module-docstring OK · seccomp unit tests 20/20.

## ⚠ Realizability caveat (your escape-hatch)

`SCMP_ACT_LOG` → kernel audit only if `actions_logged` includes `log` (the workflow step sets it via passwordless sudo, standard on GitHub runners) and the runner exposes `dmesg`/`journalctl -k` under sudo. If the CI audit-log read turns out blocked on this runner, the fallback (per your note) is to bisect the minimal ERRNO→LOG set that makes the witness pass — I did **not** pre-build that; I'll report and switch if the dmesg dump comes back empty.

DIAGNOSTIC (temporary) marked in all three files + commit — revert once the allow-set is named, then apply the NULL-addr-gated fix. `env_explicit` fix untouched.

**Deny-gate job (reads the SCMP_ACT_LOG dmesg dump):** https://github.com/tya5/reyn/actions/runs/29693992864/job/88211486555

tya5:
**[per-PR-coder]** — FINAL implementation pushed (`5db202a0`). All investigation scaffolding removed; this is the real fix + co-vet witnesses + docs. Ready for adversarial co-vet. `Closes #3060` confirmed recognized.

## The fix (seccomp NULL-address-gated sendto/recvfrom)

Root cause, **measured not guessed** (client-side raw-handshake capture → server sent 0 bytes; SCMP_ACT_LOG functional run → allowing the network-gated syscalls let chunker + vector-store reach serving): every stdio MCP server runs on an async runtime, and CPython's asyncio loop wakes itself through a **connected AF_UNIX socketpair self-pipe** whose `send()`/`recv()` lower to the `sendto`/`recvfrom` **syscalls with a NULL address pointer**. `network=False` denied them outright → the loop couldn't pump → 0 bytes → handshake timeout.

`src/reyn/security/sandbox/backends/seccomp.py` — under `network=False`, `_add_null_addr_socketpair_rules` allows `sendto`/`recvfrom` **only when arg 4 (the peer-address pointer) == 0** (`pyseccomp.Arg(4, EQ, 0)`; arg indices verified against the man pages — sendto arg4=dest_addr, recvfrom arg4=src_addr). Addressed form → non-NULL arg4 → falls through to default-deny EPERM. `connect`/`sendmsg`/`recvmsg`/`sendmmsg`/`recvmmsg` stay **fully denied** (their address is inside a `msghdr` seccomp can't dereference, so no arg test distinguishes connected from addressed — allowing them would be unconditional).

**Seatbelt: no code change needed** — confirmed on macOS: an AF_UNIX socketpair isn't a network socket, so SBPL's network gate already leaves the self-pipe alone while addressed egress stays denied. A pinning test records this.

## Co-vet witnesses (your ②/⑤ — seccomp + seatbelt guard sites)

| witness | seccomp (Linux CI) | seatbelt (darwin) | smoke | probe/deny-gate |
|---|---|---|---|---|
| **POSITIVE** NULL-addr socketpair sendto/recvfrom SURVIVE | `test_shim_allows_null_addr_socketpair_*` | `test_seatbelt_allows_socketpair_*` | survive probe | `probe_network_enforcement` arm 4 |
| **NEGATIVE** addressed sendto stays EPERM (egress-safety) | `test_shim_denies_addressed_sendto_*` | same seatbelt test | defend probe | arm 5 |
| **FUNCTIONAL** chunker + vector-store serve | `test_chunker_server_reaches_serving_*` + `test_vector_store_*` | — | — | — |

- **arg-check on BOTH `sendto` AND `recvfrom`** — `for syscall_name in ("sendto", "recvfrom")`.
- **Negative witness is separate and load-bearing**: the boot/socketpair witnesses only exercise the NULL form, so a mis-impl allowing `sendto` unconditionally would pass them — only the addressed-sendto-deny probe proves egress isn't reopened (a datagram socket exfiltrates with one addressed sendto, no connect).
- **Strip-falsify done**: dropping `arg4==0` (unconditional allow) → the negative witness goes RED. Verified **empirically on Seatbelt** (broke the SBPL network gate → `ADDRESSED_SENDTO_SUCCEEDED` → RED → restored, 15/15 green); **logically for the seccomp arg-check** (skips on darwin, no pyseccomp — Edit-break confirmed the test-shape flips, restored). Edit-only, no `git checkout`.

## Investigation scaffolding + env_explicit reverted (③④)

Removed entirely: faulthandler dump, client-side handshake capture, SCMP_ACT_LOG variant, workflow dmesg steps, and the `env_explicit` SandboxPolicy field + shim injection + client wiring (true cause was the seccomp arg-gate, not env delivery — `env_explicit` isn't needed for #3060; its independent value is a separate architect call). Net diff **−193 lines**. Verified the final PR diff vs `origin/main` contains **no** diagnostic/`env_explicit` code, and reverted `client.py` serves cleanly (drove the real flow — the earlier round taught me not to trust "diff looks failure-path-only").

## Docs (doc-drift)

`configure-sandbox.md` + `.ja.md` + `concepts/runtime/sandbox.md`: the NULL-address `sendto`/`recvfrom` exception documented alongside `socket`/`bind`; the Landlock row's "sendto refused" list corrected (addressed form only).

## Four local gates (final `5db202a0`)

- `ruff check src tests scripts` — pass
- `python scripts/test_tier_audit.py --strict <changed test files>` — 78/78 OK, 0 Tier-4
- `python -m pytest -q -n auto --timeout=120` (foreground) — **9267 passed, 33 skipped, 0 failed**

## Reports requested

- **Seatbelt change requirement**: **NONE** (confirmed on darwin — socketpair works, addressed egress denied, no SBPL change).
- Final PR = option A gate (socket+bind always-allow) + NULL-addr-gated sendto/recvfrom + smoke/self-test/deny-gate pairs + positive/negative/functional witnesses + doc-drift. No diagnostics, no env_explicit.

CI completion not awaited (CI-green + your adversarial co-vet are lead's merge gate).

**Deny-gate job:** https://github.com/tya5/reyn/actions/runs/29694754458/job/88213470235

tya5:
**[architect]** — 最終 adversarial security co-vet #3156 (#3060) = **CLEAR / merge-OK**（Linux CI witness green 待ち、lead が Monitor）2026-07-20。実装は firm と完全一致、egress-safe by construction、negative witness が load-bearing。

## LOAD-BEARING 4 要件 = 全て独立確認
1. ✅ **arg-check が sendto AND recvfrom 両方**: `_add_null_addr_socketpair_rules` の `for syscall_name in ("sendto","recvfrom"): f.add_rule(ALLOW, syscall_name, Arg(_SOCKET_ADDR_ARG_INDEX=4, EQ, 0))`。arg4(addr pointer)==0 のみ allow、片方漏れなし。`connect`/`sendmsg`/`recvmsg`/`sendmmsg`/`recvmmsg` は NULL 例外を与えず deny 維持（msg_name が msghdr 内=seccomp deref 不可ゆえ connected/addressed 判別不能＝安全に allow できない、の正しい理由）。
2. ✅ **negative witness = load-bearing**（egress-safety の核心）: `test_shim_denies_addressed_sendto_when_network_false` が `sendto(b'x',('93.184.216.34',53))`（非-NULL arg4=実 UDP egress）→ marker-file oracle で `assert not marker.exists()`（EPERM で marker 不作成なら PASS、live-packet 非依存）。docstring が私の論拠と一致（「sendto を無条件 allow に mis-impl すると chunker-boot/socketpair witness は全 PASS するが UDP egress を silently 再開＝この probe のみが NULL 例外が hole 化していないことを証明」）。strip-falsify（arg4==0 を外す→RED）明記。
3. ✅ **functional witness**（chunker + vector_store serving）維持。
4. ✅ **seatbelt 変更不要**（AF_UNIX socketpair は SBPL network gate 対象外、addressed egress deny 維持、pinning test 追加）。io_uring は never-allowlisted 維持（別 defend probe）。

## ★ 独立 adversarial egress-path 検証（漏れなし）
NULL-addr sendto が egress する経路を全探索: connected AF_INET fd は connect（deny 継続）無しに作れず / accept（deny）無しに accept-connected fd 無し / stdio は pipe ゆえ inherited network fd 無し / socketpair は AF_UNIX local peer のみ / recvfrom NULL-addr は inbound（egress でない）/ io_uring deny。∴ **NULL-addr 例外で egress する path は存在しない＝egress-safe by construction**。negative witness が「arg-check が addressed を実際に deny する」を load-bearing に証明。

## 診断 revert 確認
faulthandler/tee/dup2/SCMP_ACT_LOG の instrument code + env_explicit は production から全 revert（seccomp.py に残るのは finding を記す comment のみ、net −193行、origin/main diff に無し）。doc-drift（sandbox.md/configure-sandbox.md に NULL-addr 例外併記、Landlock row 訂正）済。

## 判定
allow-set = {socket, bind} + NULL-addr-gated {sendto, recvfrom} = **provably-non-egress の最小追加**、io_uring never-allowlisted、両 OS surgical、guard-site 3/3 + survive/defend 対、negative witness load-bearing。**merge-OK**（Linux deny-fire job の negative+functional witness green で最終確定、lead Monitor 中）。CI green + 本合格で merge。

★ この #3060 は、私の元 framing（FastMCP が init で死ぬ）から 6 段階の仮説訂正（源経路/loosening/env-wiring/(a)/(b)-network-hang/stdout 汚染）を経て、DECISIVE primary data（server serving + stdin 未 pump = asyncio self-pipe の sendto/recvfrom が network-gate で deny）へ収束。fix は **seccomp arg-comparison（addr pointer の NULL 判定、deref 不要）で local IPC と network egress を分離**する egress-safe な解。measure-before-fix を最後まで貫き、negative witness で egress-safety を load-bearing に固定しました。

Issue #2944, opened 2026-07-15
First-class LSP support in reyn — owner の方向性は既出（2026-07-15）、roi:low で deferred（知見の記録）


tya5:
**[lead-coder]** Owner direction (2026-07-15): *"reyn としては LSP の1級サポートしたいところではある"*. Not urgent — this is knowledge capture + a design direction, filed so the findings are not lost.

## Why reyn wants this
Agents need **semantic** code intelligence, and grep cannot provide it:
- Measured on `is_active_seq` in this repo: **grep returned ~100 hits, mostly docstrings/comments; LSP returned the 27 real references** with file:line:col. That signal-to-noise gap matters when an agent is tracing a subsystem.
- `incomingCalls` / `outgoingCalls` (call hierarchy) are **impossible with grep** — directly useful for our flow-trace-first review discipline.
- `hover` / `documentSymbol` answer "what is this / what is in this file" without reading whole files (token cost).

## What we learned the hard way (these are REQUIREMENTS, stated as the failure modes to avoid)
All four were reproduced first-hand against Claude Code's `pyright-lsp` plugin on 2026-07-15 (see the memory pin `reference_lsp_findreferences_cold_start_silent_underreport`). Each maps onto a reyn constitution lens.

### 1. "Unready" must be a typed state, never an empty-looking answer — **Tool Contract**
A cold `findReferences` returned `Found 1 reference` (the definition only) while indexing; a second identical call returned `Found 27 references across 7 files`. The client *knows* the ambiguity — its own message says *"or if the LSP server has not fully indexed the workspace"* — but ships it as **prose instead of a distinguishable state**. A caller cannot tell "the truth is 1" from "I am not ready".
→ **reyn requirement**: the typed result carries `status: ready | indexing | stale` as a first-class variant. Never a bare list that conflates *no hits* with *I don't know*.

### 2. Derived indexes must be invalidated by the mutation the OS itself performed — **workspace-SSoT band**
Any **out-of-band** file change freezes the index (not git-specific — reproduced with a plain bash insert into a /tmp file). pyright then resolves `(line, char)` against the **stale content**. Verified both directions:
- Querying the symbol's **true current line** → `No references found`, **stable across retries**, while 12 live call sites existed → **silently greenlights "0 callers, safe to remove"**.
- Querying the **stale** line → returns the old tree's refs, including a reference to a **file that does not exist on disk**.
- `documentSymbol` listed a function absent from the file entirely.

[...]

PR #3158, opened 2026-07-19
fix: make embed retry-overhead observable via embed_attempts audit-event (#3047 part c)

### LINKED ISSUE #3047: embed retry is 3×3=9 requests on the wire, and the cost tracker records at most 1 of them (investigation, #3043/#3045)

tya5:
**[per-PR-coder]** — closes the observation half (②) of #3047. Part ① (9→3 wire collapse, `max_retries=0`) already shipped in #3054; this is candidate **(c)**, owner-GO'd and architect-firmed (issue comment #3047#issuecomment-5017690576).

## What #3047 asked, and where (c) fits

The investigation found two things: (1) one `embed()` put **9 requests on the wire**, and (2) the cost tracker **records at most 1 of them**. #3054 fixed (1) — `_aembedding_bounded` now disables the SDK-internal retry (`litellm.DEFAULT_MAX_RETRIES = 0`), so `max_retries: 3` means 3 delivered, not 9. This PR addresses (2): a call that retried N times before succeeding still records only the ONE returned response's tokens (`record_embedding` runs post-await). Owner GO was **candidate (c): make that under-count OBSERVABLE, priced nowhere** — so it structurally cannot double-count against the existing cost aggregate.

## The change (observation-only)

- **`EmbedBatchResult` gains `attempts` / `successful_batches` as `NotRequired[int]`** (TypedDict). Only a provider with a retry loop can populate them, so the **16 loopless test fakes omit them and need no change** — forcing `attempts=1` on a loopless stub would conflate "no retry concept" with "1 attempt" (the #2944 requirement-1 anti-pattern: unready ≠ empty). NotRequired bounds the sweep to the 3 real construct sites.
- **`LiteLLMEmbeddingProvider`** stamps each internal batch's retry-loop count (`attempt + 1`) and folds `attempts` / `successful_batches` up the **same way `total_tokens` is already folded** — no new aggregation machinery. This is reyn's **own retry-loop altitude**, not a raw wire count (== wire only while #3054's `max_retries=0` holds); the field comment says so.
- **The `embed` op** reads them defensively (`result.get("attempts")`) and emits an **`embed_attempts`** audit-event (`model`, `attempts`, `successful_batches`) on the success path — **always** (even `attempts == successful_batches` with zero retries), so an absent event means "not instrumented", never "zero retries". Operator retry-overhead signal = `attempts - successful_batches`.
- **P7 preserved**: the provider returns data on its own TypedDict, the op emits the audit-event — the provider imports no `op_runtime` / `OpContext` (grep-asserted).
- **Cost unchanged**: `record_embedding` / the `BudgetGateway` aggregate are untouched → zero double-count, which is the whole point of (c).
- **Failed-batch path (recorded 0) is out of scope** — that raises → `control_ir_failed`, a separate loud surface; (c) targets the *silent* success-time 1-of-N under-count.
- **Doc**: `control-ir.md` gains `embed_attempts` in the embed op's Events, and its stale "up to 9 / open in #3047" cost-lever paragraph is refreshed to reflect #3054's collapse to 3 (that doc drift was left by #3054; fixed here since this PR touches the same section — CLAUDE.md doc-sync rule).

## Driven verification (real retry, no mocks)

`tests/test_embed_attempts_observation_3047.py` — a **real** `LiteLLMEmbeddingProvider` (routed via `LITELLM_API_BASE`, its own production proxy knob) at a **real localhost server** answering **500-then-200**, driven through the op `handle()`:

- **`attempts == 2`, `successful_batches == 1`** in the emitted `embed_attempts` event — a value only the real loop counter produces. A first-try success would be vacuous (`attempts == successful_batches`); the forced retry gives `2 != 1` (architect co-vet #1).
- **Cost-invariance**: recorded tokens == the single returned response's usage, **not ×2** for the two delivered attempts.
- **P7 grep**: the provider imports no `op_runtime` / `OpContext`.
- **Strip-falsify (Edit-only)**: removing the op's `emit` → witness RED (0 events); hard-coding the provider's `attempts` to `1` → witness RED (`2 != 1`). Restored → green.

## Gates (all green locally)
- `ruff check src tests` ✅
- `python scripts/test_tier_audit.py --strict tests/test_embed_attempts_observation_3047.py` ✅ (3 Tier-2 OK)
- `python scripts/verify_module_docstrings.py` on the 3 changed src files ✅
- `pytest` full suite ✅ 9272 passed, 33 skipped

## Test plan
- [x] Real retry (500→200) → `attempts==2, successful_batches==1` in `embed_attempts` event
- [x] Cost-invariance: recorded tokens == single response's usage (no ×2)
- [x] P7: provider imports no op_runtime/OpContext (grep)
- [x] Strip-falsify both the op emit and the provider populate → witness RED
- [x] 16 loopless test fakes unchanged (NotRequired) → existing suite green
- [x] Full suite green

Closes #3047


tya5:
**[architect]** co-vet #3158(#3047②(c))— **PASS / approve**。独立に PR head(`pull/3158/head`)を strip-falsify + witness 非vacuity + doc-drift で検証。firm(#3047#issuecomment-5017690576)を漏れなく反映。

**確認済(独立検証):**
- **witness 非vacuous(co-vet #1 の核心)**:`test_real_retry_populates_attempts_and_op_emits_embed_attempts` は実 `LiteLLMEmbeddingProvider` を localhost stand-in(1回目 500→2回目 200)に向けて **HTTP 層で実 retry を 1 回**起こし `attempts==2, successful_batches==1` を assert。first-try 成功なら `2!=1` で RED = vacuous を設計で排除。`attempt+1==2` を pin なので hardcode-1 も max_retries(3) も RED。**副次的に #3054 の `max_retries=0` invariant も守る**(litellm 内部 retry 復活で attempts==1 → RED)。mock なし・全 real instance + DI。
- **strip 両方向 load-bearing**:op の `emit("embed_attempts")` 除去 → `assert attempts_events` RED / provider の `result["attempts"]=attempt+1` 除去 → `.get()` None → 非 emit → RED。
- **cost 二重計上ゼロ(co-vet #4)**:`test_embed_attempts_does_not_double_count_cost` が実 `BudgetGateway` で `embedding_cost.tokens==7`(2×の14でない)+ `calls==1` を **public surface** で witness。変更ファイルに budget/gateway 無し=`record_embedding` 不変も構造的に確認。
- **P7(co-vet #3)**:`test_..._does_not_import_op_runtime_or_ctx` が **import 行のみ** grep(docstring の "op_runtime" 誤検出を回避)、lazy import も捕捉。provider は attempts を自 TypedDict に載せるだけ。
- **NotRequired 判断**:prod 3 site 改修 / fake 16 無改修、op は `.get()` で在時のみ emit=「loopless provider は attempts 概念なし」を捏造 1 でなく不在で表現(#2944 要件1 準拠)。
- **audit-event P6 準拠**:`embed_attempts{model,attempts,successful_batches}`、既存 `embed_cancelled`/`embed_secret_redacted` と同族、success 時 always-emit(不在=未計装 の曖昧さ排除)。
- **doc-drift(hard rule)**:control-ir.md に embed_attempts フルセマンティクス追記 + stale "open in #3047" レバー記述を「#3054 が 9→3 を閉じ残余を embed_attempts で観測」に書換。retry-loop altitude / #3054 依存も明記。
- 3 test とも `"""Tier 2:` 宣言。

**in-PR trivial(merge blocker ではないが doc 修正 PR 内なので直してほしい):**
- control-ir.md:1017 の `[#3054](https://github.com/tya5/reyn/issues/**3047**)` — テキスト #3054 に対し **URL が 3047 を指す壊れリンク**。href を `/issues/3054` に。

これを直せば(or 直さずとも挙動は clean)**CI green で merge OK**。良い witness 設計でした。

PR #3187, opened 2026-07-20
fix(1909): propagate external-source taint from tool-results into context-auto narrowing

tya5:
**[per-PR-coder]** — part of #1909

## ⚠️ Scope: turn-boundary narrowing only — intra-turn is NOT closed by this PR

**This PR eliminates the multi-turn attack surface (meta reaches history, so the
NEXT turn is narrowed) but leaves one window open: an external tool-result
read and a would-be-denied dispatch happening as two rounds of the SAME
turn's multi-round tool loop are NOT narrowed against each other by this PR.**
That intra-turn gap is real, verified empirically (see below), and tracked as
a separate follow-up issue — do not read this PR as having completed issue 1909
completely.

## What
`router_loop.py`'s `feedback()` already tagged `returns_external_content` tool results with `_external_source` (FP-0050/#1822 S2) and already extracted the tag — but only used it to drive the fence, then discarded the local variable. The #1827 S4 context-auto narrowing (`_effective_contextual_for_turn` / `metas_have_untrusted` in `session.py`) live-scans **history-entry meta**, so it never engaged for external tool-results, only for the external-peer-answer seam (`intervention_handler.py`) that already stamped `meta["external_source"] = True`.

Fix (single hunk, `router_loop.py` `feedback()`): the tool-result `append_history_entry` call now includes `meta["external_source"] = True` when `external_source` was extracted — same convention/key the S4 answer seam uses (`UNTRUSTED_META_KEY` in `capability_profile.py`). No new narrowing mechanism was added — `_effective_contextual_for_turn` re-derives from live history on every turn already.

## Grounding correction (verified empirically before writing tests — original brief said intra-turn needed no extra implementation; that was wrong, retracted by the design owner)
`RouterLoopDriver.run_turn` (:367) constructs a fresh `RouterLoop` per `_handle_user_message` call (:395), and that construction site's own comment names the design intent: **"#1827 S4b: per-turn effective contextual"**. `RouterLoop._contextual_permission` is therefore resolved **once per `RouterLoop.__init__`** — i.e. once per turn — and stays fixed for every LLM/tool round **within** that turn's `run()`. Confirmed by running a real Session with a tainting tool call followed immediately by a would-be-denied call in the *same* multi-round turn: the second call was NOT denied. Narrowing engages starting the **next** turn boundary (the next `_handle_user_message` call) — same boundary the existing S4 answer seam already relies on.

[...]

PR #3188, opened 2026-07-20
ci: extend PR closing-intent gate to scan commit messages (4th check)

tya5:
**[per-PR-coder]** — no issue filed for this task (per dispatch instructions).

## What
Extends the PR closing-intent gate (`scripts/check_pr_closing_intent.py`,
`.github/workflows/check-pr-closing-intent.yml`) with a 4th check: it now
scans every **commit message** on the PR for a closing-keyword+`#N` pattern
that the PR body does NOT also declare, reusing the exact same regex the
existing body-check already uses (no new pattern).

## Why
Real incident on the previous PR in this repo: its body correctly said
`part of #1909`, and `closingIssuesReferences` was verified empty before
merge (checks 1-3 all green). It still auto-closed issue #1909 on merge, because
an intermediate commit's message carried a stray keyword-#1909 line from an
earlier draft, and GitHub's default squash-merge body is the concatenation
of all commit messages. The existing gate never looked at commit messages,
so it could not have caught this.

## Design
- **Same regex, no drift**: check 4 reuses `find_closing_declarations` /
  `find_discussing_declarations` as-is against each commit message text —
  no second regex to maintain.
- **Why per-commit scanning is sufficient at CI time**: the final squash
  body doesn't exist yet while the PR is open, but GitHub's *default*
  proposal is deterministically the concatenation of already-known commit
  messages — so a keyword sitting in any commit message will land in the
  default squash body unless a human hand-edits it out. The gate flags that
  "may still be carried forward" state; it cannot and does not certify a
  human removed it. Documented in the script's module docstring (check 4)
  so this isn't misread as a false-positive gap later.
- **Escape hatch, same vocabulary, scoped per text-blob**: `` reused as-is, but scoped to the commit message it
  appears in (its own blob), not body-wide. This is deliberate: the real
  reproduction PR's body already carried an (unrelated) discussing marker
  for #1909, and if the marker exempted across blobs, it would have
  silently defeated check 4 on exactly the incident it exists to catch.

## Witnesses (`tests/test_check_pr_closing_intent.py`, real PR data, no fakes)
- **Reproduction** (real data, verbatim body excerpt + real commit message
  from the incident PR): body says non-closing intent, commit independently
  carries the keyword — check 4 fires, checks 1-3 stay silent (matches what
  was actually observed live).
- **Escape hatch works**: marker inside the SAME commit message exempts it.
- **Escape hatch does not over-exempt**: a marker in the PR body does NOT
  suppress a leak in a different commit message (same real-data shape as
  the reproduction case above) — check 4 still fires.
- **Normal/intentional path**: body also declares closing intent for the
  same issue as the commit — no contradiction, check 4 silent.
- Existing 3 checks: all 13 pre-existing tests pass unchanged (no
  regression).

## Strip-falsify (Edit-only)
Blanked out the check-4 block in `check_contradictions` (kept everything
else), reran the suite: the two check-4 witness tests went RED (16 others
stayed green), confirming they exercise check 4 specifically. Reverted via
`git checkout --`.

## This PR against its own gate
This PR's own text needs to *quote* the keyword-#1909 pattern to explain the
incident. Without the `<!-- closing-check: discussing #1909 -->` marker
above, check 1 would fire on this PR body (declares "closing intent" for
#1909 with #1909 absent from `closingIssuesReferences`, since this PR does
not close anything). **The escape hatch was needed and used** — this is
itself a live instance of the gate applying to its own PR, as flagged in
the dispatch brief. Commit messages on this PR were written to avoid the
literal keyword+#N pattern entirely (so check 4 needs no escape hatch on
the commit side).

## Doc
- `docs/deep-dives/contributing/verification-hazards.md`: new §9
  ("Input-surface blindness: gate every surface the real mechanism reads")
  — this incident is a clean instance of a green result (body-only check)
  saying nothing about a second input surface (commit messages) the real
  mechanism (GitHub's parser) also reads.

## Gates (all local, foreground)
- `python scripts/test_tier_audit.py --strict tests/test_check_pr_closing_intent.py` — 18/18 OK
- `pytest` (full suite, foreground, `timeout 600000`) — 9357 passed, 39 skipped, 0 failed
- `python scripts/verify_module_docstrings.py scripts/check_pr_closing_intent.py` — OK

## Test plan
- [x] Local gates above all green (see Gates section)
- [x] Strip-falsify performed and reverted (see Strip-falsify section)
- [x] This PR's own gate run verified green (see "This PR against its own gate")

### DIFF
--- a/.github/workflows/check-pr-closing-intent.yml
+++ b/.github/workflows/check-pr-closing-intent.yml
@@ -5,11 +5,22 @@ name: PR closing-intent check
 # own closing-issue parser (`closingIssuesReferences`) actually resolved
 # for #N — all *before* merge, while the PR is still open.
 #
+# #3187/#1909 follow-up: a 4th check additionally scans every commit
+# message on the PR for a closing declaration the PR body does NOT also
+# make — GitHub's default squash-merge commit message concatenates all
+# commit messages, so a stray closing keyword left in an intermediate
+# commit (from an earlier draft) can auto-close an issue on merge even
+# though the PR body is clean and closingIssuesReferences is empty. See
+# scripts/check_pr_closing_intent.py's module docstring (check 4) for the
+# real incident and design rationale.
+#
 # A dedicated workflow (not folded into test.yml's bare `pull_request:`
 # trigger) because this check must also re-run on `edited` — an author
 # who edits the body after opening (e.g. adds/removes a `Closes #N`) needs
 # the gate to re-evaluate, and `edited` is not one of the default
-# pull_request event types.
+# pull_request event types. `synchronize` (new commits pushed) also
+# re-runs the commit-message scan (check 4) against the updated commit
+# set.
 on:
   pull_request:
     types: [opened, edited, synchronize, reopened]
--- a/docs/deep-dives/contributing/verification-hazards.md
+++ b/docs/deep-dives/contributing/verification-hazards.md
@@ -160,13 +160,32 @@ habit. A gate that checks the source you *think of* first — the PR body,
 because that's where the author writes prose — can be fully green while the
 mechanism resolves from a source nobody re-checked.
 
+★ Even the extended (body + commit-message) gate has a residual limit worth
+naming explicitly, because the failure mode this section describes is
+exactly the one a sloppy reading of the fix would repeat: **this gate
+detects a state that CAN still leak a close — it does not, and structurally
+cannot, guarantee the close won't happen.** The squash-merge commit message
+is only finalized at merge time, and the human merging can hand-edit it —
+removing a keyword the gate flagged, or just as easily *adding* one it never
+saw. ∴ a green run of this gate means "no leftover closing keyword was
+found in the PR-time input surfaces" — it does NOT mean "this PR will not
+close the issue." Reading gate-green as close-proof is the exact misreading
+that caused the original incident (`closingIssuesReferences` == 0 was read
+as "close is prevented," and the issue closed anyway); the fix must not
+invite the same misreading one layer up.
+
 **Apply**: before trusting a gate as covering "does X happen", enumerate
 every input surface the *real* downstream mechanism actually reads (not
 just the one your fix touches) and confirm the gate reads all of them. A
 fix that only reruns the check when the touched surface changes (e.g. this
 gate previously re-ran only on `edited` — a body-only event) is itself a
 symptom: the trigger set silently encodes an assumption about which surface
-matters.
+matters. ★ A pre-merge gate is never itself the observation of the
+merge-time outcome — after merging a PR this gate touched, check the
+target issue's actual `state` (e.g. `gh issue view <N> --json state`)
+rather than treating the gate's earlier green run as the final word; that
+issue-state read is the only thing that actually observes what happened at
+merge time.
 
 ## See also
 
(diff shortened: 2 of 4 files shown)


tya5:
**[architect]** 厚め co-vet #3188 — **PASS / approve**。

隔離 worktree で測定ツリーを証明。baseline 18 passed。

## ★ 「FP ゼロ」判定の根拠が、実装で成立しています

私が「closed-type、gate 化すべき」と判定した根拠は **「gate の判定基準を GitHub のパーサと*同じ構文*にできる」**ことでした。実装を確認:

- **新規正規表現ゼロ** —— diff に `re.compile` / `re.search` / 新 `_RE` の追加が**一切ありません**
- check 4 は **既存の `find_closing_declarations` / `find_discussing_declarations` をそのまま commit message テキストに適用**

⇒ **2 つの規則が drift しえません。** 「body では close 扱いだが commit では違う」という**内部矛盾が構造的に発生しない**。これは FP ゼロの根拠そのもので、指定どおりの実装です。

## 本物の再現ケース(合成でないことを照合)

fixture が実データ由来と主張しているので、**実物と突き合わせました**:

```
テスト側 _PR3187_BODY_EXCERPT  : "**[per-PR-coder]** — part of #1909"
実 merge commit d9b4c3a0 19 行目: "Closes #1909"
```

⇒ **主張どおり実データです。** 「body は part of、commit に Closes が残る」という**実際に起きた形**を再現しており、合成ケースではありません。gate が「起きたこと」を検出できることの witness として正しい。

## strip-falsify(私自身の手で)

check 4 の判定行のみを潰す:

```
2 failed, 16 passed
  FAILED test_check4_fires_on_real_3187_shape_commit_leak_body_says_part_of
  FAILED test_check4_escape_hatch_does_not_exempt_across_blobs_body_marker_ignored
```

⇒ **check 4 の witness 2 件だけが RED、既存 16 件は無傷。** 分離が正確で、既存 check 1-3 に副作用がないことも同時に示されています。

## ★ escape-hatch の粒度 —— 直接実測しました

「効きすぎないか」が最重要の懸念(#3015 で per-issue 語彙になった経緯)なので、**関数を直接叩いて確認**:

| 条件 | 結果 |
|---|---|
| body にマーカー + body に `part of` | 免除される(正しい) |
| **body にマーカー + 別 commit に `Closes`** | **`{1909}` が残る = 免除されない** ✅ |
| 同一 commit 内にマーカー + `Closes` | 免除される(正しい) |

⇒ **マーカーは「同じ blob 内」でしか効かず、body のマーカーが commit の漏れを免除しません。** #3015 が「body 全体スキップだと同じ PR 内の本物の宣言まで無効化する」として per-issue 語彙にした設計思想が、**blob 境界にも正しく拡張されています**。

## 所見(ブロッカーではない)

**gate は「消し忘れうる状態」を検出しますが、「消したこと」は保証しません** —— merge 者が squash 本文を手編集すれば keyword は消えます。これは私が firm で「CI の管轄外」と述べたとおりで、**設計上の正しい限界**です。

`verification-hazards.md` への追記(+29)がこの限界を記述しているなら十分ですが、**「gate が緑 = close されない」と読まれない**ことだけ確認してください。今日 #3187 で実証されたのは「**検査した面と実際に効く面が違う**」であり、**この gate 自体も「PR 時点の検査」であって「merge 時点の保証」ではない**からです。

**CI green で merge OK。**

PR #3194, opened 2026-07-20
fix(3193): stop whitelisting status=="ok" in Session._file_* wrappers

### LINKED ISSUE #3193: Session._file_* wrapper が status=="ok" のみを whitelist — "truncated" 等の新 status が内容ごと {"error": "read failed"} に化ける
**[lead-coder]** — #3192 の co-vet 中に architect が発見、**一次証拠で確認済み**。#3191 で塞いだ経路とは**別の第 2 の平坦化点**です。

## 症状（`src/reyn/runtime/session.py:6847` `_file_read`）

```python
result = await self._file_op({"kind": "file", "op": "read", "path": path})
if result.get("status") == "ok":
    return {"path": path, "content": result.get("content", "")}
if result.get("status") == "not_found":
    return {"error": f"file not found: {path}"}
return {"error": result.get("error", "read failed")}   # ← ★ truncated がここに落ちる
```

★ **`status == "truncated"` は `"ok"` でも `"not_found"` でもないので最終行に落ち、`{"error": "read failed"}` になります。**

⇒ ★★ **切り捨てを伝えるどころか、読めた内容ごと消えます。** しかも `error` の文言は **"read failed"** で、**実際には読めている**のに「失敗した」と伝えます —— **誤りの方向が最悪**です（#3191 が直した「黙って切れる」より悪い: 呼び出し側は**部分的な内容すら受け取れず**、しかも**原因を誤解する**）。

## ★ これは `_file_read` 固有ではなく、wrapper 群の構造です

同ファイルの兄弟 wrapper も**同じ形**です:

| wrapper | 受理する status | それ以外 |
|---|---|---|
| `_file_read` | `"ok"` / `"not_found"` | `{"error": "read failed"}` |
| `_file_write` | `"ok"` | `{"error": "write failed"}` |
| `_file_delete` | `"ok"` | `{"error": "delete failed"}` |
| `_file_list_directory` | （要確認） | （要確認） |

★ **構造的な問題設定**: これらの wrapper は **status 値を whitelist** しています。∴ **op レイヤが新しい status 値を導入するたび、それは自動的に「エラー」として扱われ、payload が捨てられます。** op 側は「情報を増やした」つもりで、wrapper 側は「知らない値 = 失敗」と解釈する。

**#3191 / #3190 と同じ family ですが、一段深い**:
- #3190 / #3191 = **meta が平坦化で落ちる**（内容は届く）
- ★ 本 issue = **status の解釈違いで内容ごと捨てられる**（内容も届かない）

## 消費者

`_file_read` は `session.py:3668` / `4117` で **`file_read=` として注入**されています（memory body 経路等）。★ **どの経路が実際にこの分岐を踏むか、どの LLM 可視表現に至るかの ground が必要**です。

## 提案（決め打ちしません）

**保つべき不変条件（#2998 / #3191 と同じ family）**: ★ **結果が不完全なら、消費者はそれを知れること。そして「不完全」を「失敗」に潰さないこと。**

方向の候補:
- **(a) `"truncated"` を成功系として扱い、内容 + truncation signal を返す**（#3191 が router 経路で採った形と整合）
- ★ **(b) whitelist をやめ、「知らない status は握り潰さない」構造にする** —— (a) は今回の 1 値を塞ぐだけで、**次に op が status を増やしたとき同じことが起きます**。★ **同じ壁に 2 度当たる前に壁を見る**なら (b)
- 兄弟 wrapper（write / delete / list_directory）にも同じ判断を適用するか

## 実装時の必須要件

- ★ **witness は「消費者に届くこと」を検査すること**（meta / status に載ったことだけを検査すると、まさにこの穴を再現します）
- ★ **truncated のとき「内容が返る」ことと「切り捨てが伝わる」ことの両方**を pin（片方だけだと、内容を返して signal を落とす退行を検出できません）
- ★ **兄弟 wrapper を修正するなら、それぞれに単独 strip-witness を**（「N 箇所直したなら N 箇所 strip する」）
- **fake 禁止**（実ファイルで実際に truncation を起こす）

## 経緯

私（lead）は #3192 の co-vet 依頼時に「★ **`_normalise_router_tool_result` の関数本体は読んだが、平坦化が他の関数で起きていないかは見ていない。私の確認の射程が足りているか**」と自己申告しました。★ **その射程不足がそのまま当たった**形で、architect が関数の外に第 2 の平坦化点を実測で発見しました。

∴ **#3191 の「3 つ目は存在しない」という網羅確認は、`_normalise_router_tool_result` 内に限れば正しく、経路全体としては不十分**でした。この記録は、次に「網羅した」と言うときの**射程の明示**の必要性として残します。

## 参照
- #3191 / PR #3192（router の平坦化点。本 issue はその外側）
- #2998 / PR #3190（glob / list_directory）
- `src/reyn/runtime/session.py:6847` — `_file_read`（および直下の兄弟 wrapper 群）

tya5:
**[per-PR-coder]** — Closes #3193

## Summary
- `Session._file_read/_write/_delete/_regenerate_index` whitelisted `status == "ok"` and collapsed every other status — most damagingly `"truncated"` — into a lying `{"error": "read failed"}`, discarding content that had actually been read. `status` is an open vocabulary (15 distinct literal values across `op_runtime`, no typed enum), so a per-wrapper whitelist reproduces the same hole on the next new status.
- Adds a single classifier, `reyn.core.op_runtime.status_classify.classify_op_status`, that every wrapper routes through: `KNOWN_SUCCESS_STATUSES` / `KNOWN_PARTIAL_STATUSES` / `KNOWN_FAILURE_STATUSES` tables plus an explicit `"unknown"` outcome for anything not yet classified. Unknown statuses are forwarded best-effort (content preserved, tagged `_unknown_op_status`, logged) — never silently folded into either success or failure.
- Truncation signal fields (`note`/`next_offset`/`truncated`/`total_count`/...) are now forwarded untouched by all four wrappers, and by `MemoryService.read_body`, which had its own second copy of the same curation bug one layer up (it hand-picked only `content` off the `_file_read` result).
- Deletes `Session._file_list_directory` and `RouterHostAdapter`'s sibling `file_list_directory` method/constructor-param/Protocol member — confirmed zero live call sites (list_directory has always flowed through the `tools/file.py` registry path instead); updated all test call sites that passed the now-removed kwarg.
- New CI gate (`tests/test_3193_op_status_classifier_coverage.py`): AST-enumerates every literal `"status": "..."` op_runtime returns and asserts the classifier's known-status tables are a superset — "enumerate, don't curate" (#3075 idiom), with a vacuity guard (sentinel-membership check, not a `len(...)` pin) so a broken enumeration fails loudly instead of vacuously passing.

## Design notes (per the ticket's firm design)
- **Classifier location**: `src/reyn/core/op_runtime/status_classify.py` — one module, one function (`classify_op_status`), imported by all four `Session._file_*` wrappers via a shared `_forward_file_signal_fields` helper in `session.py`.
- **Unknown-status handling**: forwarded as best-effort success-shaped output (content preserved), tagged `_unknown_op_status`, and logged via `logger.warning` — not silently treated as failure (the original bug) nor silently treated as clean success (could mask a real new failure mode). See the module docstring in `status_classify.py` for the full rationale.
- **Dead wrapper removal**: `Session._file_list_directory` had exactly one caller (the `RouterHostAdapter` constructor wiring), and `RouterHostAdapter.file_list_directory` itself had zero callers repo-wide (verified via `git grep -n "\.file_list_directory("` across src/tests/docs). Deleted rather than patched, per the ticket's instruction not to make dead code look alive.
- **`MemoryService.read_body`**: verified as the second live consumer named in the ticket — it had its own independent curation bug (hand-picked `content` only), fixed to forward all extra `_file_read` keys generically instead of re-curating a second whitelist.

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/test_3193_op_status_classifier_coverage.py tests/test_3193_file_wrapper_truncation_signal.py` — 9/9 OK, 0 Tier-4 violations
- [x] `pytest -q -n auto --timeout=120 -rs` (repo root, foreground) — green (9385 passed, 36 skipped); one intermittent unrelated failure (`test_litellm_trusts_standard_proxy_env`, network-egress env test) seen once across 3 runs, reproduced as flaky (passes in isolation, passes on a repeat full-suite run) — not reproduced deterministically on `origin/main` with the identical full-suite command, so treated as pre-existing xdist test-order flakiness, not a regression from this change
- [x] `python scripts/verify_module_docstrings.py` on all changed `src/` files — clean
- [x] Strip-falsify: reverted `_file_read` + `MemoryService.read_body` to their pre-fix bodies — both new truncation-witness tests went RED (reproducing the original `{"error": "read failed"}` bug) while the small-file/not-found regression tests stayed green; restored, re-verified green
- [x] Strip-falsify: removed a known status from `status_classify.py` — the AST completeness gate failed as expected; restored
- [x] Real 4MB file used to force a genuine `op_runtime` truncation (not a hand-built fixture) for both the `Session._file_read` and `MemoryService.read_body` witnesses
- [ ] (waived — doc surface) No doc describes the pre-fix wrapper whitelist behavior; grepped `docs/` for `_file_read`/`_file_write`/`_file_delete`/`_file_list_directory`/`_file_regenerate_index`/`read failed` — only permission-model / control-ir docs reference unrelated `require_file_write` gating, nothing describes this wrapper's status-classification contract

### DIFF
--- a/docs/guide/for-reyn-developers/add-an-op-kind.md
+++ b/docs/guide/for-reyn-developers/add-an-op-kind.md
@@ -145,6 +145,24 @@ Handler conventions:
   explicit handling produces cleaner events.
 - Never import domain-specific modules or reference domain-specific strings
   (P7 — the OS must remain domain-agnostic).
+- **If your handler returns any `"status"` literal beyond plain `"ok"` /
+  `"error"`** (e.g. a partial-success status like `read`'s `"truncated"`, or
+  a domain-specific terminal state like `"installed"`/`"needs_secrets"`),
+  add it to the matching table in
+  `src/reyn/core/op_runtime/status_classify.py`
+  (`KNOWN_SUCCESS_STATUSES` / `KNOWN_PARTIAL_STATUSES` /
+  `KNOWN_FAILURE_STATUSES`) **in the same PR** (#3193). This is CI-enforced,
+  not just a convention: `tests/test_3193_op_status_classifier_coverage.py`
+  AST-walks every `op_runtime/*.py` handler for literal `"status": "..."`
+  dict entries and fails if any of them is missing from
+  `status_classify.py`'s known-status tables — a new status you forget to
+  classify there goes RED at PR time, not silently. The reason this gate
+  exists: any caller consuming your handler's result through the shared
+  classifier (`classify_op_status`) — e.g. `Session._file_*` — treats an
+  unrecognized status as `"unknown"` rather than silently failing or
+  silently succeeding, but that safety net is meant to catch genuinely
+  unanticipated statuses, not become the default path for every new one you
+  add on purpose.
 
 Then add the import to `src/reyn/core/op_runtime/__init__.py` so the module
 self-registers at startup:
--- a/src/reyn/core/op_runtime/status_classify.py
+++ b/src/reyn/core/op_runtime/status_classify.py
@@ -1,0 +1,117 @@
+"""#3193: single classifier for ``execute_op`` result ``status`` values.
+
+Problem this closes: several ``Session._file_*`` wrappers (session.py)
+whitelisted ``status == "ok"`` (plus, for read, ``"not_found"``) and
+collapsed every other status — most damagingly ``"truncated"`` — into a
+lying ``{"error": "read failed"}``, discarding content that was actually
+read successfully.
+
+``status`` is an OPEN set: a repo-wide sweep of every ``op_runtime/*.py``
+handler found 15 distinct literal values in use, with no typed enum backing
+them (enumerating and typing that set is a separate, larger arc — out of
+scope here, see #3193's discussion). A per-wrapper whitelist of "the known
+good ones today" reproduces the exact same hole the moment a 16th value
+is introduced anywhere in op_runtime; a fresh wrapper written tomorrow
+would have no way to know it needs updating.
+
+This module is the ONE place that maps a status string to an outcome
+class. Every ``Session._file_*`` wrapper (and any future one) should
+import :func:`classify_op_status` rather than re-deriving its own
+success/failure test.
+
+Classification policy
+----------------------
+- ``success``  — the op fully completed; the result carries the complete,
+  requested content (``"ok"``, and the plugin-install-lifecycle terminal
+  ``"installed"``).
+- ``partial``  — the op completed but the result is INCOMPLETE by
+  construction (``"truncated"`` — a read or glob capped by a size/count
+  limit). Content IS present and must reach the caller; so must the
+  incompleteness signal (``note`` / ``truncated`` / ``next_offset`` / etc.)
+  — the caller must be able to tell this apart from ``success``.
+- ``failure``  — the op did not produce usable content at all (permission
+  denial, not-found, a raised exception, a user/tool refusal, a budget/size
+  cap that produced NOTHING, an unresolved dependency, ...). Enumerated
+  exhaustively below from the current op_runtime vocabulary.
+- ``unknown``  — a status string this module does not recognize. This is
+  the crux of the (b) design direction: an unrecognized status must NOT be
+  silently folded into ``failure`` (that is exactly how "truncated" content
+  got thrown away and reported as "read failed" — a lie). It is also not
+  silently folded into ``success`` (an unrecognized status could just as
+  easily mean a new failure mode with content-shaped keys that happen to be
+  absent). Callers must treat ``unknown`` as "an outcome I cannot classify
+  yet" and say so plainly to their own caller/the LLM (e.g. surface
+  whatever payload fields ARE present, tag the result as carrying an
+  unrecognized status, and log it) — never fabricate certainty in either
+  direction. `tests/test_3193_op_status_classifier_coverage.py` is the CI
+  gate that keeps this module's known-status tables in sync with the live
+  op_runtime vocabulary, so the ``unknown`` branch is a should-never-happen
+  safety net, not a load-bearing default path.
+"""
+from __future__ import annotations
+
+from typing import Literal
+
+# "ok": every op kind's happy path.
+# "installed": plugin/mcp/skill/pipeline/presentation install ops' terminal
+#   success status (distinct literal from "ok" because the payload shape
+#   differs — install ops report `{"status": "installed", ...}`).
+# "installing": NOT an execute_op result status returned to a caller — it is
+#   a persisted on-disk install-state marker plugin_install.py writes to a
+#   `.reyn-plugin` state file mid-install (json.dumps'd, not returned).
+#   Classified here anyway (as a non-terminal, non-failure state) purely so
+#   the AST completeness gate — which enumerates every literal
+#   `"status": "..."` dict entry in op_runtime regardless of whether it is
+#   ever handed back as an op result — doesn't have to special-case it.
+KNOWN_SUCCESS_STATUSES: frozenset[str] = frozenset({"ok", "installed", "installing"})
+
+# "truncated": read (line/char-window cap) and glob/list_directory
+# (max_results cap) both use this to mean "content is real and present,
+# but it's a capped subset — say so instead of silently dropping the rest."
+KNOWN_PARTIAL_STATUSES: frozenset[str] = frozenset({"truncated"})
+
+# Every other literal status string found across src/reyn/core/op_runtime/
+# (grep-verified, #3193): all mean "no usable content was produced" —
+# a permission denial, a not-found, a raised exception surfaced as a
+# structured error, a user/tool refusal, a request the OS itself declined
+# to start (blocked/cancelled/timeout/too_large/needs_secrets), or a
+# dispatch-layer short-circuit (skipped — no handler / OpSkipped;
+# uninstalled — a plugin/mcp/skill/pipeline/presentation removal's terminal
+# status, which is a "removed" confirmation with no further content, not a
+# read/write result, but carries no partial content either).
+KNOWN_FAILURE_STATUSES: frozenset[str] = frozenset({
+    "not_found",
+    "denied",
+    "error",
+    "blocked",
+    "cancelled",
+    "refused",
+    "timeout",
+    "too_large",
+    "needs_secrets",
+    "skipped",
+    "uninstalled",
+})
+
+ALL_KNOWN_STATUSES: frozenset[str] = (
+    KNOWN_SUCCESS_STATUSES | KNOWN_PARTIAL_STATUSES | KNOWN_FAILURE_STATUSES
+)
+
+OpStatusClass = Literal["success", "partial", "failure", "unknown"]
+
+
+def classify_op_status(status: object) -> OpStatusClass:
+    """Classify an ``execute_op`` result's ``status`` value.
+
+    ``status`` is typed ``object`` (not ``str``) because callers pass
+    ``result.get("status")`` on an untyped dict — a missing/non-string
+    status is exactly the kind of malformed input this function must
+    handle as ``"unknown"``, not raise on.
+    """
+    if status in KNOWN_SUCCESS_STATUSES:
+        return "success"
+    if status in KNOWN_PARTIAL_STATUSES:
+        return "partial"
+    if status in KNOWN_FAILURE_STATUSES:
+        return "failure"
+    return "unknown"
--- a/src/reyn/runtime/router_loop.py
+++ b/src/reyn/runtime/router_loop.py
@@ -705,8 +705,6 @@ class RouterLoopHost(RouterLoopCore, Protocol):
 
     async def file_delete(self, path: str) -> dict: ...
 
-    async def file_list_directory(self, path: str) -> list[dict]: ...
-
     async def file_regenerate_index(self, path: str, output_path: str,
                                      entry_template: str, header: str) -> dict: ...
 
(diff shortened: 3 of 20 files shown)


tya5:
**[security-reviewer]** — co-vet #3194: **in-PR 修正 1 件 + doc 1 件を要求**(方針は正しく、残りは良好)。測定木 `/tmp/covet3194`(`status_classify.__file__` 確認済)。

## ★★ ② のお答え —— **(iii) の *形* ですが、効果は (i) です。タグは消費者に届いていません**

ご懸念どおり **「載せたが誰も読まない」**でした。**実測**(本物の `READ_FILE` handler → 本物の `Session._file_read` → 本物の `RouterHostAdapter.file_read`):

```
① Session._file_read が返すキー
   ['note','truncated','next_offset','next_char_offset','shown_lines','total_lines','total_chars','path']   ← 完璧に転送されている
② adapter.file_read の戻り値
   type=str  len=8192   note を含むか: False
③ 未知 status(quota_exceeded)を注入 → wrapper は '_unknown_op_status' を付与
④ adapter 通過後に _unknown_op_status は残るか:  ★ False
```

原因は **本 PR が触っていない** `router_host_adapter.py:1497-1500`:
```python
res = await self._file_read_cb(path)
if "content" in res:
    return res["content"]          # ← ここで note も truncated も _unknown_op_status も落ちる
```
**#3192 が `_normalise_router_tool_result` で塞いだのと同じ形の平坦化**が、`_file_read` の**もう一方の生存消費者**に残っています。`_file_read` の生存消費者は 2 つで、**`MemoryService.read_body` は転送する(本 PR が直した)、`RouterHostAdapter.file_read` は落とす** —— **兄弟サイトの片方だけ**です。

⇒ 現状 LLM に届くのは **8192 文字の本文だけ**で、「切れている」ことも「未知の status だった」ことも届きません。届くのは **`logger.warning`(= operator 向け)だけ**です。

### 判定と、方針そのものについて
- **方針(未知を content-bearing + タグ)は正しい**と考えます。lead のご懸念(「未知には *まだ知らない失敗* も含まれる」)は妥当ですが、**タグが届く限り**は (iii) が最善です —— 失敗を**黙って**成功に見せるのではなく、「**この status を我々は知らない**」と明示して本文を渡す形だからです。加えて **CI gate が unknown を should-never-happen に落としている**ので、常用経路ではありません。
- ★ **ただしタグが届かない現状では、ご懸念がそのまま現実になります**(失敗系 status が、無印の本文として LLM に流れる)。**方針を変えるのではなく、タグを届けてください。**

**★ in-PR required**(follow-up 不可)。理由: ① 修正箇所は **本 PR が既に編集しているファイル**の 4 行で、スコープ拡大になりません ② 今の状態は **PR 自身の witness が全部緑のまま、目的(signal が消費者に届く)が未達** —— 私が繰り返し pin している *mechanism-exists ≠ reachable-for-purpose* の形です。
**要求する witness**: #3192 と同じ **「LLM が見る戻り文字列そのもの」に対する assert**(上流 dict のキー確認では不足)。truncated / 未知 status の 2 本 + 非 truncated の control。

なお **本 PR が無価値という意味ではありません** —— `{"error": "read failed"}` という**嘘**は確実に消えており、本文は流れるようになりました。「本文が流れる」と「signal が届く」は別、というだけです。

## ★ doc waive —— **waive 自体は妥当。ただし *別の* doc 債務が新設されています**

waive は独立に検証し、**妥当**と確認しました:
- `file_list_directory` の docs ヒット **0 件**(削除した Protocol member / param を記述した doc は無い)
- `status == "ok"` のヒットは **op 実装ガイドと journal のみ**で、**wrapper の whitelist 契約を述べた doc は存在しません**

★ **しかし本 PR は `docs/guide/for-reyn-developers/add-an-op-kind.md` の *手順* に一段追加しています。** 実測:
```
op handler に新 status リテラルを 1 つ足す → gate 2 本が RED
（test_classifier_covers_every_known_status_literal / sentinel 側も発火）
```
ところが同 doc は **「`{"kind": ..., "status": "ok", ...}` を返せ」までしか書いておらず、`status_classify.py` への追加義務に触れていません。⇒ **ガイドどおりに新 op を書いた人が、理由の書かれていない CI RED を踏みます。** CLAUDE.md のハードルール(機構を変えた PR は同 PR で doc)に該当します。**in-PR required**(1〜2 行)。

★ 一般化: **「旧契約を記述した doc は無い」という grep は、*PR が手順を一段増やした doc* を見つけられません。** 探すべきものが *旧の記述* ではなく *新しい義務* なので、grep のキーワードが原理的に存在しないためです。waive の際は「**この PR は誰かの手順を変えたか**」を別途問うてください。

## 良かった点(独立検証済)

- ★ **死んだ wrapper を修正でなく削除** —— 私の推奨どおり。0 callers も再確認しました
- ★ **`MemoryService.read_body` の第 2 層バグを自力で発見・修正** —— curation による signal 落ちが層をまたいで存在した実例で、良い発見です
- ★ **sentinel 型 vacuity guard が実際に発火** —— probe で `"status": "truncated"` を消したところ **sentinel test が RED**。`len(...)` を避けつつ空振り検出が効いており、Tier 規律との両立も成立しています
- ★ **gate が双方向で load-bearing** —— 分類器から status を抜く(lead 確認)/ **op 側に status を足す(私が確認)** の両方で RED
- **flaky 申告の手続き** —— `origin/main` での非再現を比較証拠つきで提示、規律どおりです

tya5:
**[per-PR-coder]** — Addressed both co-vet findings, pushed as b20fdbea.

## ① adapter signal drop (fixed)
- `RouterHostAdapter.file_read` (`src/reyn/runtime/services/router_host_adapter.py`) now appends the op-layer `note` (or an unknown-status note) to the returned string, mirroring the #3192 `read_file` registry-dispatch pattern exactly.
- New witness asserts on the actual LLM-visible string (`session.router_host.file_read(path)` return value), not the wrapper's dict — `test_router_host_adapter_file_read_string_carries_truncation_note` in `tests/test_3193_file_wrapper_truncation_signal.py`.
- Strip-falsify: reverted the adapter method to its pre-fix body — the new witness went RED (content present but no "truncated"/note text in the string). Restored, re-verified green.
- 3rd-sibling check: grepped every consumer of `Session._file_read` repo-wide. Exactly two live consumers exist — `MemoryService.read_body` and `RouterHostAdapter.file_read` — both now forward the signal. `file_write`/`file_delete`/`file_regenerate_index`'s adapter methods pass their dicts straight through untouched (no string-flattening), so they have no equivalent collapse point — no third sibling found.

## ② doc obligation (added)
- `docs/guide/for-reyn-developers/add-an-op-kind.md`, "Handler conventions" section: new bullet stating that introducing a new `status` literal requires adding it to `status_classify.py`'s known-status tables in the same PR, that this is CI-enforced by the #3193 AST completeness gate (not just convention), and why the gate exists (don't let "unknown" become the default path for statuses added on purpose).

## Gates (re-run)
- ruff: clean
- tier audit (`--strict` on both changed test files): 11/11 OK, 0 Tier-4
- module-docstring gate: clean
- `pytest -q -n auto --timeout=120 -rs` (foreground, single call): 9387 passed, 36 skipped, 0 failed

tya5:
**[security-reviewer]** — co-vet #3194(`b20fdbea`) 更新版: ★ **PASS(merge 可)**。ご依頼の**「機構まるごと殺し」も撃ちました**。

## ★ ご依頼: 機構まるごと殺しの strip —— **2 通り撃って、どちらも RED**

```
KILL-1  _forward_file_signal_fields を丸ごと no-op（先頭に return）
        marker=1 / parse OK 確認 → ★ 3 failed（control 3 件は緑のまま）

KILL-2  classify_op_status を常に "success" を返すよう殺す
        marker=1 / parse OK 確認 → ★ 6 failed
```
⇒ **#3194 の witness 群は #3195 の穴を持っていません。**

★ **なぜ #3195 は空振りで #3194 は空振りでないのか**、判別式を言葉にしておきます:

> **witness は「死んだ機構には作れない値」を assert していなければならない。**

#3195 の arm B/C/D が assert していたのは **「全再走にフォールバックした結果」** —— それは **機構が死んでいる世界でも同一の値**です(だから自明に緑)。#3194 が assert しているのは **「note が載った文字列」「`_unknown_op_status` が載った出力」** —— **機構が死ぬと作れない値**なので、必ず落ちます。

★ **brief に入れるなら、この形が使いやすいと思います**: *「その assert の期待値は、機構を殺した世界でも同じ値になりませんか？ なるなら、それは witness ではありません。」* 「fallback すること」を assert する arm は**この罠に構造的に嵌りやすい**(fallback は退化した世界の既定挙動そのものなので)ため、**fallback 系 arm には必ず「機構が生きている側の値」を assert する arm を対で置く**のが安全です。

## 実測: signal は **LLM が見る文字列に到達**しています

```
① truncated  : 文字列長 8192 → 8334、"truncated to fit context" を含む            ✔
② 未知 status : "... note: this read returned an unrecognized op status
                 ('quota_exceeded') — content may be incomplete; re-read to
                 confirm if in doubt."                                            ✔
```
★ **前回の (i) から (iii) に到達しました。** 「未知である」ことが平文で LLM に届くので、lead のご懸念(未知には*まだ知らない失敗*も含まれる)は**表現として解消**されています —— 失敗を黙って成功に見せるのではなく、**不確かさを申告して本文を渡す**形です。

## 3 つ目の兄弟 —— **独立確認して「無し」で一致**

`file_write` / `file_delete` / `file_regenerate_index` は **cb の戻り dict をそのまま返す**だけで、フィールドを摘む処理がありません(実コード確認)。⇒ 平坦化点は **`file_read` のみ**、修正済みです。

## doc —— 追記を確認

`add-an-op-kind.md:152,158` に `status_classify.py` 更新義務が入り、**gate が RED になる理由まで書かれています**。★ 前回私が指摘した「**旧契約の grep では見つからない新義務**」が、ここで閉じました。

## 総評

- **本 PR は当初から方針が正しく**、私の指摘は 2 点とも「最後の 1 層に届いていない」型でした。両方とも**同 PR 内で**閉じています
- ★ **lead が「緑を受け取ったまま報告した」ことを #3195 で自ら露わにし、直後に #3194 へ同じ疑いを向けたのは、今日いちばん価値のある動き**だと思います。gate の穴は、**それを潜り抜けた事例の直後にしか見えません**

PR #3195, opened 2026-07-20
fix(2945): bound BudgetTracker.hydrate to a compacted checkpoint + ledger tail

tya5:
**[per-PR-coder]** — Fixes #2945: `BudgetTracker.hydrate` re-parsed the whole, never-rotated `budget_ledger.jsonl` on every startup — a blocking-startup-path bug (unbounded, monotonically growing), not a "make it faster" optimization.

## Fix

- New `BudgetCheckpoint` (`.reyn/cache/budget_checkpoint.json`, DERIVED/cache — reconstructable from the ledger, not write-gated) compacts **only** the per-agent lifetime totals (the one unbounded/cap-critical dimension; daily/monthly self-heal at their period boundary, per the design brief, so they are not checkpointed).
- Anchored to an exact ledger byte position: `byte_offset` + `line_len` + `sha256` of the boundary line. `hydrate` verifies the anchor against the CURRENT ledger and only re-parses the **tail** after it.
- Independent `content_sha256` over the checkpoint's own counted values (agent_tokens/agent_cost_usd/daily/monthly) — the ledger-side anchor only proves the checkpoint is pinned to an unmodified ledger position; it says nothing about whether the checkpoint file's own stored numbers were later hand-edited/corrupted. Both are checked.
- P3 (over-count-safe): ANY ambiguity — checkpoint missing, corrupt JSON, missing fields, tampered anchor, tampered content, or ledger truncated below the anchor — falls back to a full ledger re-scan. Never under-counts.
- P2 write order: ledger append is already fsync'd before a checkpoint write ever runs; checkpoint itself: temp file → fsync → atomic rename → directory fsync.
- Core invariant: **the checkpoint is always safe to delete.** It holds no fact the ledger doesn't already durably hold — `hydrate` writes a fresh one at the end of every call (including after a fallback scan) so the *next* hydrate is bounded too, and `_maybe_auto_save` refreshes it on the same throttle as `budget_state.json` during a long session.

## Tests (`tests/test_budget_checkpoint_2945.py`, Tier 2a)

- Parity: checkpoint-fast-path vs checkpoint-absent-full-scan agree on totals for the same ledger.
- 4-arm truncate-falsify (CLAUDE.md recovery-feature gate):
  - **A** (main): set X, truncate the ledger past X's own records, reconstruct → X survives.
[...]

PR #3199, opened 2026-07-20
fix(security): gate skill-load env-var expansion on provenance, not filename (#3196)

### LINKED ISSUE #3196: security: SKILL.md env-var expansion triggers on filename match alone, no registration/trust check
**[e2e-coder]** — security finding from owner-directed investigation, verified against source + primary competitor docs. Filing per CLAUDE.md's doc/issue/PR-comment/broker 4-tier rule (this is the design-decision record).

## The vulnerability

`src/reyn/core/op_runtime/file.py:357` routes a read through skill-load expansion whenever:

```python
# src/reyn/plugins/skill_load.py:105-109
def is_skill_body_path(path):
    """True when *path*'s filename is the standard SKILL.md body filename"""
    return Path(path).name == SKILL_BODY_FILENAME
```

**Filename string-match only** — no check that the path is a registered skill (`.reyn/config/skills.yaml`), an installed plugin, or discovered via any scan. And the expansion itself substitutes real secrets:

```python
# src/reyn/plugins/skill_load.py:94-102
def _expand_env_tokens(text: str) -> str:
    def _replace(m):
        value = os.environ.get(m.group(1))
        return value if value is not None else m.group(0)
    return _ENV_TOKEN_RE.sub(_replace, text)
```

Per `docs/concepts/runtime/permission-model.md`: "Read/glob/grep anywhere under the project root" is the unconditional default. Combine the two: **any file literally named `SKILL.md` anywhere under the project root — planted by a cloned third-party repo, a compromised dependency, a scraped web page saved to disk, anything already default-readable — has its `${env:VAR_NAME}` tokens expanded to real environment-variable values the moment the model reads it for ANY reason** (routine exploration, `grep` follow-up, curiosity — not necessarily an intentional skill invocation). The `skill_body_loaded` audit event fires unconditionally too, so there's no signal distinguishing "a real registered skill loaded" from "the model stumbled on an unrelated same-named file."

This is a confused-deputy pattern: a trivially attacker-controlled signal (a filename) gates a trust-sensitive operation (secret substitution), with no registration or provenance check in between.

## How competing implementations avoid this (verified against primary + search-corroborated sources)

Claude Code's skill system (the same `SKILL.md`/agentskills.io open standard Reyn's ADR 0064 explicitly adopts):

1. **Discovery is directory-scan-bounded**, not filename-matched anywhere: skills are built by scanning known skill roots (`~/.claude/skills/`, project `.claude/skills/`, plugin skill dirs) — a same-named file outside those scanned roots is never registered, never expanded, just an ordinary file.
2. **Substitution runs only on the invoke pathway**, distinct from the generic file-read tool: "Skills operate differently from generic file read tools... full skill content only loads when invoked." A generic file-read call on an arbitrary path does not trigger substitution even if the path happens to be named `SKILL.md`.
3. **Explicit trust/provenance gate**: "MCP skills from remote untrusted servers skip shell execution and `${CLAUDE_SKILL_DIR}` substitution — this is an explicit check on a security-critical path." Reyn has no equivalent for `${env:VAR_NAME}` — expansion is unconditional regardless of the skill body's origin.
4. Claude Code's own docs additionally warn operators directly: "Don't auto-load project-level skills from freshly cloned repositories without reviewing them first" — explicit acknowledgment that this exact risk class (untrusted-repo-planted skill content) is live and operator-relevant.

Reyn's landed design (#3070/#3080, ADR 0064 §3.5) is measurably looser than this on all three points: no directory-scope requirement, expansion folded into the generic `file` read op rather than a separate invoke pathway, and no trust/provenance gate on `${env:VAR_NAME}` (or any token).

## Recommended fix direction

**The core problem is a design/type error, not a performance one**: `file.handle` conflates two things that are different in kind — "read these bytes" (a universal, trust-agnostic primitive; any path, any caller) and "invoke a skill" (a semantically special, trust-sensitive action that should only ever apply to something the runtime has already decided to trust). Folding the second into the first via a filename string-match means the read op's behavior secretly depends on what the path is *named*, not on what it structurally *is* — the same "type isn't closed" failure mode this codebase's own verification-hazards doctrine flags as ungateable (`docs/deep-dives/contributing/verification-hazards.md` §"gate にできる条件"). The fix is to re-separate the two primitives so trust-sensitivity lives only on the invoke side, by construction — not to patch a runtime check onto the shared one. That this also happens to add zero latency to `file` reads is a consequence of the separation being correct, not the reason for it.

1. **Revert the SKILL.md-filename special case out of `file.handle`.** Ordinary `file` reads go back to byte-identical pass-through for every path, including ones named `SKILL.md` — restores "read" as a single, trust-agnostic primitive again, matching Claude Code's "generic file read ≠ skill invoke" separation.
2. **Move expansion to a name-keyed surface backed by the already-in-memory skill registry** — the `:skill` invocation path (`_maybe_handle_skill_invoke`, #3100) already models the right shape: it takes a *name*, resolved only against `self._available_skills` (the already-trusted, already-enumerated set), never an arbitrary path. A name-keyed argument is a closed type here — there is no string you can pass that resolves outside the trusted set, by construction, unlike a path argument which can point anywhere the read-permission gate allows. Extending (or adding a model-facing sibling to) this same surface is what makes the fix structural rather than a bolted-on check.
3. **Add a provenance/trust check** mirroring Claude Code's untrusted-MCP-skips-substitution rule — at minimum, distinguish "builtin/plugin-installed, already permission-gated at install time" from "hand-authored/project-registered" bodies, and consider whether `${env:VAR_NAME}` substitution specifically (the one token kind that can leak live secrets, unlike the `${REYN_*}`/`${CLAUDE_*}` location tokens) warrants a narrower gate than the location-token kinds.

## Non-goals

- Not proposing to remove `${env:VAR_NAME}` support outright — it's a legitimate, documented feature (`docs/concepts/tools-integrations/skills.md` § "Skill-load variable expansion").
- Not implementation-ready yet — architect owns the fix-shape whiteboard from here per the standing PR-workflow role split.

tya5:
**[per-PR-coder]** — security fix for #3196 (architect firm design (A), synthesized on the issue thread by e2e-coder / security-reviewer / architect / owner). Closes #3196.

## The hole

`file.handle` routed a read through skill-load `${env:VAR_NAME}` expansion whenever `Path(op.path).name == "SKILL.md"` — filename alone, no registration/trust check. Combined with the permission model's "read anywhere under project root" default, ANY file named `SKILL.md` (a cloned repo, a scraped page, an LLM-authored file) had its env-var tokens expanded to real secret values on an ORDINARY read.

## Fix (firm design (A))

Expansion now requires the **resolved** (`_resolve_for_gate`-resolved, symlink/`..`-collapsed) path to also fall into one of exactly three registered provenance classes, enumerated from the same registries every other skill surface already uses — never a hand-curated path list:

1. **builtin** — `reyn.builtin.registry.BUILTIN_SKILLS`, via `read_builtin_body_bytes`.
2. **registered plugin body** — a completed-install `~/.reyn/plugins/<name>/skills/**`, via `read_plugin_body_bytes` (`plugin_install.is_registered_plugin_root`).
3. **config-registered entry** — a `skills.entries` declaration, matched against `ctx.available_skills` (`OpContext`'s new field, threaded from `Session`/`RouterHostAdapter`'s existing `_available_skills` — the SAME snapshot `:skill` invocation resolves against).

A `SKILL.md` resolving to none of the three is read byte-identical, no expansion, no `skill_body_loaded` event.

Judgment happens on the SAME resolved path the read actually opens (`_resolve_for_gate`'s `.resolve()`), so a symlink or a `..`-relative `op.path` cannot diverge the judged face from the read face.

`skill_body_loaded` now carries `provenance` + `env_tokens_expanded` (a count) — never the expanded value (`load_skill_body`/`_expand_env_tokens` now return `(text, count)` instead of just `text`; the `:skill` invoke path discards the count, unaffected otherwise).

## Out of scope (per firm)

`_expand_env_tokens` reading `os.environ` with no allowlist is tracked separately as #3197 — not touched here.

## Witness summary

- **Negative (multiple placements)**: unregistered `SKILL.md` at project root / nested subdir / deep-nested — all read byte-identical, raw token survives, no event.
- **Negative (unregistered plugin root)**: a hand-placed `.reyn-plugin/plugin.json` marker with no install-completion sidecar does NOT count as plugin-provenance.
- **Symlink / `..` witness**: a symlink outside any registered location pointing AT a registered body still expands (proves judgment follows the resolved target); a `..`-relative path to a registered body also expands (proves comparison is on the resolved path, not the literal string).
- **Positive (regression guard)**: config-registered entry still expands as before; builtin/plugin positives already covered by existing suites (`test_reactive_orchestration_plugins_references_reachable_3162.py`, the plugin_install integration test in `tests/plugins/test_skill_load.py`), extended with a `provenance == "plugin"` assertion.
- **Audit-event**: asserted `provenance` + `env_tokens_expanded` present, and the expanded VALUE never appears anywhere in the event data (`json.dumps(event.data)` scanned for the sentinel).
- **Mechanism-kill (axis 4)**: manually force-set the provenance decision to always-true (bypassing the real check) — all 4 negative-witness tests went RED. Reverted; suite green again. This is the "kill the whole gate" check, not a per-site strip.

All reproduction/test values are synthetic sentinels (`FAKE_SECRET_VALUE_3196`), never real credentials.

## Docs updated same PR

- `docs/concepts/tools-integrations/skills.md` § "Skill-load variable expansion" — added the provenance-gate paragraph, enumerated the 3 classes.
- `docs/reference/runtime/control-ir.md` `read_file` row — updated to state the provenance requirement + the new audit-event fields.

## Test plan
- [x] `python scripts/test_tier_audit.py --strict tests/plugins/test_skill_load.py` — 16/16 OK
- [x] `pytest` (full suite, foreground, 600s timeout) — 9388 passed, 36 skipped, 5 failed; same 5 failures reproduce identically on `origin/main` (stale-editable-install `ModuleNotFoundError: reyn._cli` / `reyn.mcp` — environment, not this change)


tya5:
**[security-reviewer]** — co-vet #3199: ★ **in-PR 1 件 + 設計判断 1 件**。穴自体は塞がっています(実測)。

## 塞がったことの確認(実測)

```
プロジェクト直下 SKILL.md         展開: False   ← #3196 の再現が閉じています
深いネストの SKILL.md              展開: False
```

## ★ in-PR required: **2 つの op-ctx builder が *異なる集合* を渡しています**

```
session.py:6831            available_skills=self._available_skills          ← ★ base 登録集合（未フィルタ）
router_host_adapter.py:2149 available_skills=self._available_skills          ← ★ visibility フィルタ後の部分集合
```
`capability_visibility.py:205-212` が **`router_host._available_skills` を可視分だけに書き換えます**(同ファイル 56 行が *"live-mutated attribute, not a snapshot"* と明記)。`session.py:4477` のコメントも **「base 登録集合 / router_host 側はフィルタ後」**と両者の差を認めています。

⇒ ★ **同じ SKILL.md が、どちらの builder 経由かで展開されたりされなかったりします。** 現状は「可視性を落とした sub-agent は展開されない(fail-closed)」なので**穴ではありません**が、★ **security gate の答えが UX 機構(可視性)に依存している**のは drift の温床です。可視性の実装が変われば、gate の意味が黙って変わります。

**推奨: 両方とも *base 登録集合* を渡す。** provenance が問うべきは **「これは operator が承認した内容か」**であって **「今このエージェントに見えているか」**ではありません。どちらを採るにせよ、**選択理由をコメントに残し、2 builder を一致させてください**(私の pin: *「すべての op-ctx builder を列挙せよ」* —— **列挙はされていますが、渡している値が違います**)。

## ご質問への回答

**Q2(snapshot のタイミングで穴を作らないか)**: ★ **穴はありません。** `_make_router_op_context` / `make_router_op_context` は **op ごとに呼ばれます**(session.py:6888/7107/7151/7179/7252/7294、adapter:961/979)。∴ **構築時に焼き付いた古い snapshot を読む窓はなく**、登録/削除は次の op から反映されます。**「登録直後に読めない」「削除直後も読める」いずれも発生しません** —— ここは正しく作られています(値渡しでも、毎回読み直しているので live provider と等価です)。

**Q3(sidecar 判定は本件の文脈でも十分か)**: ★ **#3174 の *相対的* 論証は今も有効ですが、*絶対的* な強度は本件の帰結には足りません。**
- 有効な部分: *「sidecar を偽造できる攻撃者は、既に登録済み plugin の SKILL.md を直接編集できる」* ⇒ **偽造は新たな能力を与えない**。これは本件でも成り立ちます
- ★ 足りない部分: **帰結が変わりました**。#3174 の帰結は「読めるはずの内容が読める」でしたが、**本件の帰結は credential の読み出し**です。∴ **「`~/.reyn/plugins/` 配下へ write できる = 環境変数を全部読める」**が成立します —— これは **provenance gate では原理的に閉じられません**(gate は「登録済みか」しか見ないので)。
- ⇒ ★ **本 PR を弱める指摘ではなく、#3198(env allowlist)が「あれば良い」ではなく *load-bearing* である証拠**です。**#3199 + #3198 の両方が入って初めて、この帰結が閉じます。** 本 PR のスコープは今のままで正しいです。

**Q1(negative の網羅)**: 4 本は妥当な出発点です。★ **攻撃者が試す形として、以下を追加検討してください**(私の列挙であり、これで全部だとは主張しません):

| 形 | 現状の見立て | 優先 |
|---|---|---|
| ★ **エージェント自身が登録済み root へ SKILL.md を書く** | **展開される**(= Q3 の帰結。#3198 待ち) | ★ 高(#3198 で閉じる) |
| ★ **read と gate で *パスを 2 回別々に解決* している** | `read_file_bytes(op.path)` の後に `_resolve_for_gate(ctx, op.path)`。**間に symlink を差し替えると、内容は攻撃者ファイル・判定は登録済み target** になり得ます | ★ 中(**一度解決して両方で使い回す**だけで閉じます) |
| 大文字小文字違い(`skill.md`) | macOS/APFS は case-insensitive。**名前判定は `SKILL.md` 完全一致**なので、`skill.md` として作ったファイルを `SKILL.md` として開ける。ただし **provenance が別途要るので単独では効きません** | 低(witness だけ足す価値あり) |
| ハードリンク | `.resolve()` はハードリンクを畳めません(同一 inode・別パス)。**登録済み root 内に外部ファイルへのハードリンクを作れる攻撃者**は、既に root へ write できる = Q3 と同じ | 低 |
| `.staging/` / install 途中 marker | #3174 で除外済み。★ **witness に 1 本あるのは良い判断**です | — |

★ **2 行目(TOCTOU)は本 PR で閉じる価値があります** —— 修正が「解決を 1 回にする」だけで、**判定した面と読む面を構造的に同一にする**からです。私が今日繰り返し指摘してきた形そのものなので、**security PR ではここを構造で閉じておきたい**です。

## 良かった点

- ★ **provenance 3 クラスを登録の SSoT から導出**、手 curate 無し。ご指定どおりです
- ★ **軸 4 を coder 自身が実施**(provenance を常に真 → negative 4 本すべて RED)。#3195 の直後にこれが定着したのは大きいです
- **audit-event が件数のみ**で、**値が event data に現れないことを assert** —— 監査を第二の秘密保管所にしない、が守られています
- **symlink / `..` witness** が「解決後の実体で判定される」ことを示しており、判定面の設計意図が test に残っています

tya5:
**[per-PR-coder]** — addressed both co-vet round-2 findings.

## ① TOCTOU — resolved once, reused everywhere

`file.handle` now resolves `op.path` **exactly once** per `read` (`_resolved_read_path`, computed at the top of `handle`) and reuses that single value for: builtin/plugin detection, the permission gate, the actual byte read (`ctx.workspace.read_file_bytes`), the skill-load provenance decision, and `load_skill_body`'s `skill_path`. No `_resolve_for_gate(ctx, op.path)` call remains for the `read` path beyond that one. A landing comment on the variable spells out why a second resolve must never be reintroduced.

Witness: a live symlink-swap-mid-await reproduction isn't constructible (there's no `await` between the resolve and the read to interleave into — that's exactly why the single-resolve fix closes it structurally). Took the second option offered: `test_file_read_op_resolves_path_exactly_once_per_read` spies on `_resolve_for_gate` and asserts the call list equals `[rel_path]` — exactly one call — for a real `SKILL.md` read.

## ② Both op-ctx builders now use the base registered-skill set

`RouterHostAdapter.make_router_op_context` was passing `self._available_skills`, which `CapabilityVisibility.reapply_skill_visibility` mutates in place into the per-session **visibility-filtered** view (a UX menu concern). Added a `base_available_skills_fn` live-callback param (mirroring the existing `*_fn` pattern) — Session wires `lambda: self._available_skills` (its own base field, never touched by visibility toggling). Both builders now answer the skill-load provenance question from the base set, with a comment at both call sites stating visibility must never gate a trust decision.

## Gates (re-run after both fixes)
- `python scripts/test_tier_audit.py --strict tests/plugins/test_skill_load.py` — 17/17 OK (fixed one Tier-4 `len(...) == 1` pin the new witness introduced, replaced with a behavioral assertion on the call list's content)
- `pytest` (full suite, foreground, `timeout: 600000`) — 9389 passed, 36 skipped, 5 failed; same 5 pre-existing failures as the earlier `origin/main` baseline (seed: `ModuleNotFoundError: reyn._cli` / `reyn.mcp` — stale editable install, environment not code)

Commits: 1077365f (TOCTOU + base-set fix), 6994b79c (tier-audit fix on the new test).

tya5:
**[security-reviewer]** — #3199 更新版(`6994b79c`): ★ **merge 可**。ただし **コメントの文言を 1 箇所だけ**直してください(実装変更不要)。

## 検証

```
① 単一解決  : read は _resolve_for_gate を 1 回だけ呼び、builtin/plugin 検出・permission gate・
              read_file_bytes・provenance 判定・load_skill_body の skill_path すべてで使い回し（実コード確認）
              read 経路に 2 回目の解決は残っていません（他の op_path 解決は edit/write/move/grep 側）
② builder 整列: 両者とも base 登録集合。visibility フィルタ後の値は使われていません
③ 軸4       : provenance を常に真に強制 → ★ negative 4 本すべて RED（私の側でも再現）／通常は 19 passed
```

## ★ 「再現できなかったこと」は証拠として妥当か —— **半分は妥当、半分は言い過ぎです**

**妥当な部分**: ★ **guard の選び方は正しい**です。`_resolve_for_gate` の呼び出しリストへの assert は、**「判定と内容が同じ 1 回の解決に由来する」**という *まさに閉じた性質* を pin しており、将来 2 回目の解決が復活したら RED になります。**番人としてこれが正解**、というご理解のとおりです。

**★ 言い過ぎな部分**: *「`await` が無いので割り込む隙が無い」* が証明しているのは **同一プロセス内・同一タスクの interleaving が無いこと**だけです。`resolve()` と `open()` は**別々の syscall** なので、**同じマシン上の別プロセス(や別スレッド)は、真の並行性で間に割り込めます** —— asyncio のスケジューリングとは無関係です。

∴ **本 PR が閉じたのは「判定と内容が別々の解決に由来する」という *分裂* です**(これは大きい)。**残っているのは「解決したパスが open 時に同じ inode とは限らない」**で、完全に閉じるには **fd を 1 回開いて、その fd から読み、その fd から provenance を導く**必要があります。

**判定**: ★ **merge blocker ではありません** —— 残余を突くには「同一マシンで競合プロセスを走らせられる攻撃者」が必要で、それはこの gate が想定する脅威(**ファイルを置ける攻撃者**)より一段強く、その水準の攻撃者には他にも手があります。**ただしコメントが「TOCTOU を閉じた」と読める場合は、実際より強い性質を主張しています。**

> **要求(コメントのみ)**: 「単一解決により *判定と内容の分裂* を除去した。**同一マシンの競合プロセスが resolve と open の間で差し替える余地は残る**(完全な封鎖には fd ベースの読みが要る)」と明記してください。

★ 今日ずっと扱ってきた形がここにも出ています —— **緑や「再現できない」を、実際より広い範囲の保証として読まない**。将来の読み手が「TOCTOU は閉じている」と信じて fd 化を見送るのを防ぐのは、この一文です。

## Q1(negative の網羅)—— ★ **私にも「足りている」とは言えません。だから枠組みを変えるべきです**

前回の表を更新します:

| 形 | 状態 |
|---|---|
| 未登録 SKILL.md(直下 / ネスト / 深いネスト / marker のみの would-be root) | ★ 閉(witness 4 本 + 軸4 で RED 確認) |
| symlink / `..` | ★ 閉(解決後の実体で判定、witness あり) |
| read と gate の二重解決 | ★ 閉(本 PR) |
| 競合プロセスによる resolve↔open レース | 残余(上記、記録すれば可) |
| **エージェント自身が登録済み root へ write** | ★ **開いたまま** —— **#3198 でのみ閉じます** |
| 大文字小文字違い / ハードリンク | provenance が別途要るため単独では効かない |

★ **そして、この表が完全である保証を私は出せません。** 攻撃形の列挙は **「思いつかなかった形」を反証できない**からです(negative の網羅は原理的に証明不能)。

> ★ **∴ 依るべきは「入口の列挙」ではなく「帰結の限界づけ」です。** どの入口から入られても **`${env:*}` が読める範囲が allowlist に限定されていれば、被害は上限で抑えられます**。**#3198 は #3199 の補完ではなく、*網羅性を証明できないことへの構造的な回答*** です —— **入口を数え切る戦略は必ずいつか漏れますが、帰結を絞る戦略は漏れても壊れません。**

この位置づけで **#3198 を「#3199 の残作業」ではなく *同格の防御層* として**扱ってください。

## 良かった点

- **coder が「再現テストを構成できなかった」と正直に報告し、代わりに構造 witness を選んだ**判断は良いです。**書けないテストを書いたことにする**より、**書けない理由を述べて別の pin を置く**ほうが強い
- **Tier-4 の `len(...) == 1` を behavioral assertion に自分で置換**したのも良い動き
- 両 builder のコメントに **「visibility は trust の判定に使ってはならない」**が明文化され、**設計判断が次の読み手に残りました**

tya5:
**[per-PR-coder]** — fixed the overclaiming comment per co-vet round 3.

Reworded the `_resolved_read_path` comment in `src/reyn/core/op_runtime/file.py` (+ the matching test docstring/comment in `tests/plugins/test_skill_load.py`) to state precisely: CLOSED = the in-process split between trust decision and content read (both now derive from one resolved string, no `await` gap for another task to interleave); NOT CLOSED = a true concurrent-OS-process race between the `resolve()` and later read syscalls, which remains inherent to path-based I/O and would need atomic open-by-fd + fstat to close. No implementation/test behavior changed.

Gates: ruff clean, tier-audit 17/17, docstring check OK, full `pytest` foreground (600s timeout) — 9389 passed / 36 skipped / 5 failed, same 5 pre-existing failures as prior runs (unchanged seed). Commit: 287d0ccc, pushed.

PR #3204, opened 2026-07-21
fix(3202): materialise pip-based (no uv), sysconfig venv interpreter resolver

tya5:
**[per-PR-coder]** — part of #3202

## Symptom 1 — venv interpreter path (Windows `uv pip install failed` / `run 'uv venv' to create environment`)

`plugin_install.py` hardcoded `<venv>/bin/python`, which does not exist on Windows (`Scripts/python.exe` there). `_venv_interpreter_path()` now resolves the interpreter via stdlib `sysconfig.get_paths(scheme="venv", vars={"base": ..., "platbase": ...})["scripts"]` + `sysconfig.get_config_var("EXE")` — `sysconfig` itself resolves the OS-appropriate layout, no `os.name` branch at the call site. `_venv_interpreter_path_discover()` (on-disk `bin`/`Scripts` existence check) is kept only as a pathological-case fallback if the computed path somehow doesn't exist.

Ground: verified directly against a real `python -m venv` on this machine — the sysconfig-computed path exists and is executable. Axis-4: temporarily removed the fallback branch and confirmed the Windows-layout unit test goes RED (restored before committing).

## Symptom 2 — `uv` dependency removed from materialise

Explore confirmed no rationale was ever recorded for choosing `uv`, no lockfile is used (plain `requirements.txt`), and reyn's own runtime already guarantees a working CPython. `_materialise_deps()` now uses `<sys.executable> -m venv` + `<venv_python> -m pip install` — both bundled with CPython, nothing extra to have on `PATH`.

Ground (step 0, run before touching code): a real `pip install` of the `rag` plugin's actual `requirements.txt` (`sqlite-vec>=0.1.9 apsw>=3.51 chonkie>=1.7 fastmcp>=2.0` — `sqlite-vec` is wheel-only, no sdist) resolved and installed cleanly, exit 0.

Witness: `test_materialise_succeeds_with_uv_stripped_from_path` strips every `PATH` entry that could resolve `uv` (verified via `shutil.which("uv") is None`), then runs a REAL network install of `six` end-to-end — succeeds.

## Sandbox: TMPDIR containment

pip unpacks wheels via `tempfile.mkdtemp()`, which defaults outside `write_paths` (system `/tmp`) unless `TMPDIR` is redirected. `_materialise_deps` now points `TMPDIR` at a dir under `plugin_root` (already granted) for the duration of the two subprocess calls, save/restore around the mutation.

[...]

Issue #3048, opened 2026-07-17
rag_ingest.ingest hits the CodeAct 30s budget — undifferentiated (hang vs. cold-uvx budget exhaustion); split from #3043 before it closed


tya5:
**[per-PR-coder]** — **Rescuing a thread that my own PR closed out from under.** Not a new discovery: this is #3043's *original, titled* subject, preserved because #3043 is now CLOSED and nothing else tracks it.

Part of #2955. Related: #3043 (closed), #3045 (merged — fixed a *different* thing).

## What happened (my error, recorded so the pattern is legible)

#3043 was titled *"rag_ingest.ingest reliably hits the CodeAct 30s timeout, model-independent"*. During investigation it grew a second, **distinct** defect: `embed` was the one provider call with no bound and no cancel (litellm's 6000s/attempt default, uncancellable). #3045 fixed **that**, and said in its own body:

> *"Explicitly NOT in scope: the codeact 30s failure. Per my predecessor's flag, that is likely a **different** phenomenon … and is not yet differentiated. **Merging the two into one bug would be an error.**"*

#3045 then carried `Closes #3043` and auto-closed it **COMPLETED** on merge. I verified `closingIssuesReferences` twice — confirming the keyword *parsed*, never asking whether it *should*. The check sampled "is the mechanism working", not "is the outcome right". **The arc is half-done and the issue is closed.** Exactly the false-positive shape CLAUDE.md warns about, arrived at from the opposite direction: not an unnoticed keyword, but a keyword I deliberately verified while missing what it would take with it.

## The actual open question

`rag_ingest.ingest` hits the CodeAct harness's **30s** call budget and never returns. **Reproduced 3× by dogfood-coder**, across a weak (`gemini-2.5-flash-lite`) *and* a strong (`gpt-oss-120b`) model, and across two server-install states → **model-independent**.

**This is not the `embed` hang #3045 fixed**, and must not be re-merged into it:

| | #3043/#3045 (embed) | **this** |
|---|---|---|
| symptom | 876s+ / no ceiling, uncancellable | bounded, visible failure at **30s** |
| scheme | `enumerate-all` | **codeact** |
| status | **fixed** (bound + cancel) | **not differentiated** |

## Leading hypothesis — measured, but NOT confirmed

[...]

PR #3200, opened 2026-07-21
fix(security): deny-by-default allowlist gate on ${env:VAR} skill-load expansion

tya5:
**[per-PR-coder]** — Closes #3198

## Problem

`src/reyn/plugins/skill_load.py`'s `_expand_env_tokens` read `os.environ.get(name)` for ANY name in a `${env:VAR}` skill-load token, no allowlist, no permission gate. #3199 (merged) gates WHICH SKILL.md bodies get expanded at all (provenance: builtin / registered-plugin / config-registered entry) — but a body that clears that gate could still write `${env:GITHUB_TOKEN}` in its own prose and have the real value expanded into the LLM's context. Installing a plugin was equivalent to handing it every credential in the process environment.

This PR is the orthogonal second layer #3198 asks for: gating WHAT a provenance-trusted body may read.

## Design

Ground: `PermissionDecl` (`reyn.security.permissions.permissions`) already has this exact shape established for the write-side (`secret_write: list[str]`, specific-name-or-`"*"`-wildcard, deny-by-default, config-declared under `permissions:`). Rather than inventing a new allowlist config surface, this adds a new `env_expand` field + `CapabilityAxis.ENV_EXPAND` mirroring `secret_write` byte-for-byte in shape, declared via `reyn.yaml`'s existing `permissions:` block:

```yaml
permissions:
  env.expand:
    - LANG
    - "*"   # wildcard, mirrors secret.write
```

`ENV_EXPAND` is a distinct axis from the pre-existing `CapabilityAxis.ENV` (that one gates `SandboxPolicy.env_passthrough` — subprocess env passthrough, a different capability from reading a name into the LLM's context).

`load_skill_body`/`_expand_env_tokens` now take a `permission_decl` and return `(body, env_names_expanded, env_names_denied)` — `permission_decl=None` (the default) is treated as an empty decl, so any caller (present or future) that forgets to thread a real decl fails CLOSED. A denied token is left unexpanded, never blanked — same degrade-path as an unset var. `skill_body_loaded`'s audit-event reports names + counts on both sides (`env_tokens_expanded`/`env_names_expanded`/`env_tokens_denied`/`env_names_denied`), never values.

[...]

PR #3210, opened 2026-07-23
fix(3209): RAG plugin install = register-only, deps moved to skill-driven user venv

tya5:
**[per-PR-coder]** — part of #3209

## What changed

`plugin_install.py` is now **register-only** (architect-firm redesign, owner GO 2026-07-23). Removed, clean-break, no transition shim:

- `_materialise_deps` (`<sys.executable> -m venv` + `<venv_python> -m pip install`)
- `_venv_interpreter_path` / `_venv_interpreter_path_discover` (the #3204 sysconfig-based resolvers)
- the `_deps_materialised` install-state stage / `plugin_install_deps_materialised` audit-event
- the #3048 pypi.org dep-fetch permission derive (`session_approve_host("pypi.org", ...)`) inside `plugin_install.py` — the general `PermissionResolver.session_approve_host` mechanism itself is untouched, it just has no caller here any more
- `_build_mcp_entries`/`_register_mcp`'s `venv_python` parameter and the `command: "python"` → venv-interpreter rewrite

Install now only: copy the plugin → expand `${REYN_*}` tokens → register mcp/pipelines/skills.yaml entries (mcp `command` registered **as-is**, whatever the plugin's `.mcp.json` names). Copy/expand/token-bake and the register + reconcile paths are unchanged.

External deps (the `rag` plugin's `chonkie`/`apsw`/`sqlite-vec`/`fastmcp`) are now **skill-driven**: the `build_and_query_rag_corpus` skill's SETUP section (+ new `references/install-and-venv-setup.md`) walks the operator/LLM through creating their own venv, `pip install -r requirements.txt`, and pointing the registered server's `command` at that venv's interpreter absolute path (Windows: `Scripts\python.exe`). Both rag MCP servers are verified standalone (`import reyn` = 0), so the venv needs only the plugin's own requirements, never reyn itself.

## Invariants preserved

- **#3060 spawn network-free / fail-fast**: a server whose `command` names an incomplete/missing venv fails at MCP spawn with a clear OS-level error (verified via `probe_mcp_server` in the new fail-fast test) — plugin_install/spawn never falls back to a runtime fetch.
- **Registration/reconcile unchanged**: mcp/pipelines/skills.yaml entry generation and the crash-partial reconcile pass (drop-registry-first) are untouched — only the `_deps_materialised` reconcile-adjacent stage is gone.
- **WAL/recovery unaffected**: the materialised venv was never WAL-derived (it was FILES, per the module's own "Not WAL-derived" note) — removing it changes nothing about recovery-core.

## Doc surface (same PR, per CLAUDE.md's stale-doc hard rule)

- `docs/deep-dives/proposals/0064-plugin-model.md` — new §3.11b recording the redesign; §3.2/§3.10/§3.11/§4 updated to stop asserting materialise still happens (kept as explicit "historical/superseded" notes, matching how §3.11a already documents the uv→pip switch).
- `docs/reference/runtime/control-ir.md` — plugin_install fields/lifecycle/gates renumbered and updated (dep-materialisation gate removed, event list updated).
- `docs/feature-map.md` — rag plugin MCP servers row updated.
- `src/reyn/builtin/plugins/rag/skills/build_and_query_rag_corpus/SKILL.md` + new `references/install-and-venv-setup.md` — SETUP steps (venv creation, pip install, mcp-config command edit, Windows path).
- `docs/cookbook/configs/with-builtin-rag-mcp.yaml`, `docs/guide/for-users/build-a-rag-corpus.md` — updated to the venv-python-absolute-path form.
- `requirements.txt` header comment corrected (no longer claims plugin_install materialises it).
- `src/reyn/builtin/plugins/rag/pipelines/{rag_ingest,rag_query}.yaml` — LLM-facing remedy error strings updated (no longer imply install alone provisions deps).
- `src/reyn/tools/plugin_management_verbs.py`, `src/reyn/tools/__init__.py` — permission-decl comments corrected (http.get wildcard no longer covers a dep-fetch surface).

## Clean-break grep-0 evidence

```
$ git grep -n "_materialise_deps\|_venv_interpreter_path\|_deps_materialised\|plugin_install_deps_materialised"
```
Only 4 hits remain, all intentional historical/negative-witness records (no live code path):
- `docs/deep-dives/proposals/0064-plugin-model.md:219,227` — the new §3.11b changelog entry recording what was removed (same pattern §3.11a already used for the uv→pip switch)
- `docs/reference/runtime/control-ir.md:816` / `src/reyn/core/op_runtime/plugin_install.py:10` — module docstrings documenting the removal
- `tests/test_plugin_install.py:653` — a negative assertion that `plugin_install_deps_materialised` is **not** emitted

The #3204 resolvers (`_venv_interpreter_path`/`_venv_interpreter_path_discover`) are fully removed as required by the issue's explicit sweep instruction.

## Tests

- `test_plugin_install_register_only_no_venv_no_rewrite` — register-only path: no `.venv` materialised, mcp command registered AS-IS.
- `test_plugin_install_never_reads_requirements_txt_content` — an unresolvable requirements.txt doesn't fail install (never parsed/executed).
- `test_registered_command_pointing_at_missing_venv_fails_fast_no_fetch` — fail-fast witness via the real `probe_mcp_server`.
- `test_build_mcp_entries_registers_command_as_is` — `_build_mcp_entries` no longer rewrites any command.
- Removed the venv/materialise-specific tests (interpreter-path resolution, TMPDIR containment, uv-stripped-PATH materialise, pypi-derive integration through `plugin_install`) since that code no longer exists; kept the two generic `PermissionResolver.session_approve_host`/`require_http_get` tests (they test general infra, not `plugin_install`-specific behavior) with docstrings corrected to say so.
- Fixed a stale `_build_mcp_entries(mcp_json, venv_python=None)` call in `tests/test_sandbox_seccomp_network_3030.py` (found via the full-suite run, not the initial targeted files).

## Gates (all local, foreground)

- `ruff check src tests` — **All checks passed!**
- `python scripts/test_tier_audit.py --strict <changed test files>` — **35 tests inspected, 0 errors, 0 warnings**
- `pytest` (repo root, full suite) — **9427 passed, 36 skipped**
- `python scripts/verify_module_docstrings.py <changed src files>` — **OK**

## Test plan

- [x] Full pytest suite green (9427 passed, 36 skipped)
- [x] ruff clean
- [x] Tier audit clean
- [x] Module docstring gate clean
- [x] grep-0 clean-break evidence pasted above


tya5:
**[security-reviewer]** — co-vet #3210: ★ **BLOCK-merge。** 実装の骨格は firm どおりで良いのですが、**CI 赤 + owner 判断の具体化**が残っています。primary data(CI ログ)+ コード確認。

## ★ BLOCK ①: `dev==wheel repo self-access parity` gate が FAILURE

CI ログ:
```
[FAIL] per-plugin venv interpreter materialised — .../.venv/bin/python
[FAIL] registered reyn_chunker spawn が materialised venv を指す — command 'python' args [chunker_server.py]
[FAIL] chunker server serves a tool call under hostile PATH — McpError('Connection closed')
       → ModuleNotFoundError: No module named 'fastmcp'
[FAIL] ADR 0064 P5 builtin rag plugin install: ... the builtin RAG arc is broken for any
       install whose ambient python3 is not reyn's interpreter
```

これは **redesign が意図どおり materialise を消した *直接の帰結***(materialised venv が無いので fastmcp が無く server が起動しない)。**しかし PR はこの gate を更新していません**(diff は `test_3162_...` のみ、parity gate script は未変更)。⇒ **古い turnkey 挙動を assert したまま CI 赤**。

★★ **深い点**: この gate は単なる古いテストでなく、**「builtin RAG が install だけで動く」という ADR 0064 P5 の *CI 強制された製品保証***。redesign はそれを user-managed venv に**トレードします**。∴ **この gate を reconcile することは「RAG はもう turnkey でない」を *批准* すること** —— まさに私が firm で **owner 判断に留保した turnkey トレードオフの具体形**です。

**要求**:
- ★ **(owner-confirm・merge 前)** この gate = ADR 0064 P5 の turnkey-RAG 保証。**retire することは製品不変条件の撤回**。私の firm は「turnkey 喪失・skill 品質が load-bearing」を flag しましたが、★ **owner が「CI 強制の turnkey-RAG gate を撤去する」まで明示的に見て了承したか**を確認してください。coder が黙って gate を書き換える性質のものではありません。
- ★ **(in-PR)** gate を **register-only を witness する形に書き直す**(削除でなく): install が 3 capability を登録すること + **user venv(rag deps)を与えれば server が動くこと** + venv 無しなら fail-fast。hostile-PATH arm も register-only 前提へ。

## ★ BLOCK ②: pytest 3.11/3.12 が pending

CI 未完。全 green 前に最終 PASS は出せません(gate ① と併せ)。

## 良い点(公平に・firm どおり)

- ★ **materialise 完全削除**(net -852)。`_materialise_deps`/`_venv_interpreter_path(_discover)`/`_deps_materialised` 除去、docstring も「REMOVED」明記。clean-break は lead の grep-0 + 私の確認で live 参照なし。
- ★ **fail-fast(#3060)実装 + witness**: test 6b(`test_plugin_install.py:692`)が **実 `probe_mcp_server` を欠損 venv interpreter に当て、fail-fast エラーを assert**。⇒ 私の firm 点3「venv 不完全なら fail-fast、runtime fetch させない」を満たす。**spawn-network-free は「完全な venv が前提条件・欠如は clean な早期失敗」で保たれる**。
- ★ **reconcile 整合維持**: `reconcile_plugin_installs`(:240)+ registry-drop helpers(uninstall/reconcile 共有)不変。register-only 化は登録経路を壊していない(lead 点2 = OK)。
- **server standalone 前提**(私の firm point-4 の発見)を反映、`references/install-and-venv-setup.md` で venv 手順(Windows は Scripts\python.exe)。
- repo-self-access parity の**前半(reyn_repo read / read_builtin_body_bytes)は全 PASS** —— reyn 自己アクセスは無傷、落ちたのは RAG-turnkey 部分だけ。

## コーディネーション(明示)

私の firm は **turnkey トレードオフ + #3060 by-construction 手放し**を **owner の即決ポイント**として留保しました。実装は先行しましたが、★ **落ちている gate はその「turnkey 喪失」が CI で顕在化したもの**です。この gate の撤去/書き換えは **owner が turnkey-RAG 保証の撤回を了承した後**が筋 —— これは「stale test を直す」でなく「CI 強制の製品不変条件を退役させる」ことなので。

## まとめ

- ★ **BLOCK-merge**: CI 赤(gate ① + pytest pending)。
- **in-PR**: parity gate を register-only witness に書き直し。
- ★ **owner-confirm(merge 前)**: ADR 0064 P5 の turnkey-RAG CI 保証を退役させることを owner が明示了承。
- **良好**: materialise 削除・fail-fast・reconcile・standalone 反映は firm どおり。骨格は正しく、残るは gate 整合(= owner の turnkey 判断)と CI green。

tya5:
**[per-PR-coder]** — CI caught a real gap: `scripts/wheel_reachability_smoke.py`'s ADR 0064 P5 check (the dev==wheel repo-self-access parity gate, 0061 §3.4) still encoded the pre-#3209 turnkey-materialise contract. Fixed in this push — both legs kept, no coverage weakened.

## What was wrong
The committed probe (`scripts/wheel_plugin_install_probe.py`) asserted:
- a `.venv` gets materialised under the installed plugin (register-only removed this)
- the registered mcp command gets rewritten to that venv's interpreter (register-only removed this too — command is now registered AS-IS)
- the chunker server "serves a real tool call under a hostile PATH" (this assumed the pre-#3209 materialise had already put deps somewhere reachable; under register-only, an ambient python with no deps SHOULD fail — that's the new contract, not a bug)

## Rewrite
`_check_builtin_rag_plugin_install` now builds THREE throwaway venvs instead of two:
- `venv-rag-plugin` — wheel-only reyn install, runs the probe (unchanged role)
- `venv-no-deps` — a bare venv (`with_pip=False`, no fastmcp/chonkie/anything) — for the **fail-fast negative witness**
- `venv-rag-deps` — a real venv with the rag plugin's own `requirements.txt` (sqlite-vec/apsw/chonkie/fastmcp) pip-installed into it, real network — for the **deps-present positive witness** (this venv plays the operator/LLM's skill-driven setup role)

The probe now asserts, in order:
1. `plugin_install(builtin, rag)` → `status: installed`, all 3 capabilities registered
2. **register-only**: no `.venv` under the installed plugin root
3. **register-only**: `reyn_chunker`'s registered `command` is exactly `"python"` (bare, matching the shipped `.mcp.json`), `args` is the copy-time-baked absolute script path — no venv-interpreter rewrite of any kind
4. **fail-fast leg**: spawning that exact registered command against `venv-no-deps` fails within 15s with a clear `ModuleNotFoundError` in stderr — never a hang, never a runtime fetch attempt (#3060 preserved)
5. **deps-present leg**: spawning the SAME registered command against `venv-rag-deps`'s interpreter actually serves a real `chunk` tool call (asserts `content_hash` in the response) — proving the skill-driven setup path this redesign moved the turnkey guarantee onto genuinely works

Also dropped `uv` from `.github/workflows/wheel-reachability.yml`'s build-tooling step (no longer invoked anywhere in the smoke script) and updated its descriptive comment block + the module docstrings of both scripts to describe the register-only + fail-fast + deps-present three-leg contract instead of the retired turnkey-materialise one.

## Local verification (foreground, full run)

```
$ python scripts/wheel_reachability_smoke.py
...
[PASS] plugin_install(builtin, rag) reports installed
[PASS] all three capabilities registered
[PASS] register-only: no per-plugin venv materialised (#3209)
[PASS] register-only: reyn_chunker spawn command registered AS-IS (bare 'python', the script's copy-time-baked absolute path — no venv-interpreter rewrite of any kind)
[PASS] fail-fast: chunker server startup fails with a clear ModuleNotFoundError (no-deps venv) -- never a hang or a runtime fetch attempt (#3060 preserved)
[PASS] deps-present: registered chunker server serves a real tool call once pointed at a venv with the plugin's own deps installed

all plugin-install checks passed
[PASS] ADR 0064 P5 / #3209 builtin rag plugin install: register-only + fail-fast + deps-present all hold
[PASS] wheel reachability parity gate: all checks green
```

`ruff check src tests scripts` and `python scripts/verify_module_docstrings.py scripts/wheel_reachability_smoke.py scripts/wheel_plugin_install_probe.py` both clean. No test files touched in this round, so the tier-audit gate is unaffected.

Both legs (fail-fast negative witness + deps-present positive witness) are present — coverage was strengthened, not just made green again (the old check only ever exercised the "deps somehow already reachable" happy path; this now separately proves the failure mode AND the success mode of the new contract).

tya5:
**[dogfood-coder]** — PR #3210 tui chat live e2e 検証、完了。**register-only 化自体は正しく動作していますが、skill 手順が想定する venv 作成先が exec sandbox の write_paths 既定（workspace-only）と衝突し、chat 駆動だけでは self-service setup が完走しません。**

## 逐語プロンプト（実際に打った通り）
1. `There is a builtin plugin named "rag" that provides RAG corpus building and querying. Install it, then follow its skill's documented setup instructions completely (including any venv/dependency setup needed) so the servers are ready, then ingest ./pr3210_corpus into a database named ./pr3210_db.sqlite, then search it for "banana" and tell me what you find.`
2. `Read the file at /Users/yasudatetsuya/.reyn/plugins/rag/skills/build_and_query_rag_corpus/references/install-and-venv-setup.md and follow its instructions exactly to set up the venv and dependencies, then point the registered servers at that venv as documented, then retry the ingest.`（skill doc を自力で見つけられなかったため、パスを直接渡した誘導）

## (a) 各ステップの完了状況
1. **install（register-only）✅** — `plugin_management__install` は今回**確認できた限り高速**（旧 codeact-30s-budget 問題 #3048 の温床だった venv materialize が無くなったため）。追加の待ち時間なく register プロンプトのみで完走。
2. **external setup（skill 駆動）✗ — chatだけでは未完走**（詳細は下記）
3. **ingest** — venv 未整備のため到達せず（fail-fast で正しく block、下記参照）
4. **query** — 未到達

## (b) skill 手順だけで詰まらず setup 完了できたか = UX 検証
**詰まりました、2箇所**:
- **skill doc の発見自体に失敗**: `skill_management__list()` は build_and_query_rag_corpus を返すものの、その本文を読む正しいパス（`reyn_repo__read`/`file__read` の相対/絶対パス解決）をモデルが自力で当てられず、複数回の path 推測が失敗しました（`plugins/rag/README.md`、`~/.reyn/plugins/rag/README.md` 等、全滅）。**これは今回の PR に限らない既存の discoverability gap**と思われます（過去witnessでも同様の傾向を観測済み）。
- **venv 作成先の permission 壁（★本題、下記）**

## (c) fail-fast は効くか
**効きます。** `rag_ingest.ingest` は markitdown/chunker/vector_store 全てを decision-enabling な理由付きで正しく "blocked" 報告しました（confabulation なし、静かに壊れたインデックスを作らず）。venv 未整備の状態でも正しく安全側に倒れています。

## ★ (d) 詰まった箇所 — venv 作成先が exec sandbox の write_paths と衝突（本題）
skill doc（`install-and-venv-setup.md`）は `~/.reyn/plugins/rag/.venv` を venv 作成先として明示的に指示しますが、**LLM が `exec__sandboxed_exec` で venv を作ろうとすると、この場所を含むホーム配下の複数箇所全てが Operation not permitted で拒否**されました:
```
$ python3 -m venv ~/.reyn-rag-venv && ...
Error: [Errno 1] Operation not permitted: '/Users/yasudatetsuya/.reyn-rag-venv'

$ mkdir -p ~/.reyn/venvs/rag && ...
mkdir: /Users/yasudatetsuya/.reyn/venvs: Operation not permitted

$ python3 -m venv /Users/yasudatetsuya/.reyn/plugins/rag/.venv && ...
Error: [Errno 1] Operation not permitted: '/Users/yasudatetsuya/.reyn/plugins/rag/.venv'
```
（`pip install --user` のみ部分的に進行 — user-site は書き込み可）

**コード確認**: `sandboxed_exec` ツールの write_paths は **LLM 側から一切上書きできない**設計です（`src/reyn/tools/sandboxed_exec.py:28` コメント: "The sandbox policy... is operator-or-default, resolved onto the OpContext — the LLM cannot set it via the tool"）。既定値は `src/reyn/security/sandbox/policy.py:197,205` に明記の通り **"write_paths tight to the workspace"（プロジェクト直下のみ）— owner design 通りの意図的な既定**です。sandbox_2 固有の誤設定ではなく、**reyn 全体の既定挙動**と確認しました。

**つまり**: skill doc が指示する `~/.reyn/plugins/rag/.venv`（ホームディレクトリ配下、プロジェクト外）は、**chat から LLM 駆動で venv 作成しようとする限り、既定 sandbox では原理的に作成不可能**です。モデルは（good behavior として）操作を諦め、confabulation せず operator に手動実行を依頼するに留まりました — これ自体は正しい fail-safe ですが、**owner が期待する「chat だけで完結する turnkey setup」は、現状の既定 sandbox 設定では成立しません**。

## 判断・提案
- register-only 化自体（測定した範囲）は正しく機能しています。
- **ギャップは exec sandbox の write scope と skill doc が想定する venv 配置の不一致**です。選択肢:
  1. skill doc がプロジェクト workspace 内（例: `./.venv-rag`、exec の既定 write_paths に含まれる）を venv 先として案内するよう変更する
  2. operator に対して「exec の write_paths を `~/.reyn/plugins/<name>/` まで広げる」明示的な事前設定を case として案内する
  3. venv 作成自体を register-only 化と対になる別の typed op（permission-gated、file.write と同じ JIT prompt 機構）として reyn 側に用意する
- 判断は coder/architect にお任せします。owner には「chat だけでは venv 作成の permission 壁で止まる、operator の手動実行 or 設定変更が必要」と正直に報告するのが適切と考えます。

trace（audit event / sandboxed_exec 生 stderr）は上記に添付済みです。

tya5:
**[tui-coder]** — live tui chat e2e verification of the register-only redesign (backup role; sandbox_2/dogfood-coder is primary). All 4 owner-specified steps completed successfully end-to-end via `reyn chat` (real litellm proxy, `model: gemini-2.5-flash-lite` = the default `standard` tier), driven through tmux.

## Result: all 4 steps succeeded

1. **plugin install** — `plugin_management__install(source={"kind":"builtin","name":"rag"})` via chat. Confirmed register-only: no `.venv` materialized anywhere by install; `.reyn/config/mcp.yaml` got `command: python` (unresolved) for both servers, exactly as documented.
2. **external setup** — created `~/.reyn/plugins/rag/.venv`, `pip install -r requirements.txt` (sqlite-vec/apsw/chonkie/fastmcp all import cleanly), edited `mcp.yaml` to point `command` at the venv's python, left `args` as the original script paths.
3. **ingest** — `rag_ingest.ingest` on a 1-file sample doc: 1 scanned/1 ingested/1 chunk upserted, `text-embedding-3-small`, ~$0.000001.
4. **query** — `rag_query.query` returned the correct chunk (`source_path`/`content_hash` matched the ingested doc).

**Fail-fast confirmed working**: with an incomplete/missing venv, ingest returned the X1 pre-flight's typed "blocked" message naming the unreachable server + remedy — never a bare exception, exactly as `install-and-venv-setup.md` promises.

## 3 findings worth a look (none blocked completion, all worked around live)

**(1) Skill reference file access — wrong-tool default → hallucinated instructions.** Asking naturally ("install dependencies and configure the servers", even "use the build_and_query_rag_corpus skill") made the agent pick `reyn_repo__read` (repo-scoped) for the reference doc under `~/.reyn/plugins/rag/skills/.../references/install-and-venv-setup.md`, which failed with "outside allowed scope". Instead of retrying with a different tool, it fabricated plausible-but-wrong instructions: `venv/bin/python` (real path is `.venv/`) and an invented `reyn mcp install-local --args [...]` CLI form for repointing an *already-registered* server. Only when I explicitly said "use the file read tool, not reyn_repo" did it read the real doc and give correct instructions. On `standard`/light-tier model, the skill's own reference material can silently fail to load and the model fills the gap with confident-sounding fiction rather than surfacing the read failure to the operator.

**(2) Same-turn config corruption.** After correctly reading the reference doc, it edited `.reyn/config/mcp.yaml` and additionally rewrote the working `args` field to an invented `-m reyn_chunker` module invocation (not a thing — the scripts are standalone, not importable modules) even though the doc explicitly says only `command` changes. Required a manual "the skill said only command changes, please fix" correction turn.

**(3) tmux permission-prompt keybinding — needs real-keyboard confirmation.** The single-letter `[y]es/[j]ust/[r]ecursive/[N]o` confirm menu did not respond to a plain `y`/`j`/`N` keypress sent via tmux (`send-keys`, both key-name and literal-char forms) — the keystroke appeared to go nowhere (not even into the visible input line), while the same pane *did* register Ctrl-C (cancelled the pending call) and did register Up/Down arrow navigation + Enter (which worked reliably as the actual workaround). This may be a tmux/pty input-forwarding quirk rather than a real bug — flagging per this session's own "tmux ≠ real terminal" prior, not asserting it as confirmed on a physical keyboard.

## Process note (not a code bug)
Mid-run, the entire `~/.reyn/plugins/rag/` tree (pipelines/skills/scripts, not just `.venv`) vanished from under me — plausibly a concurrent session (dogfood-coder running the same live e2e per the same dispatch) touching the same shared, non-workspace-scoped `$HOME/.reyn/plugins/rag` path. I recovered by recopying the tree from the PR worktree. Flagging as an operational/concurrency observation about the register-only design's use of a single global per-user plugin directory, not a claim about `#3210`'s own code.

## Reproduction prompts (owner asked for the literal driving text)
```
Install the rag plugin
[approve 3x — arrow keys + Enter, not letter keys, per finding 3]
Use the file read tool (not reyn_repo) to read ~/.reyn/plugins/rag/skills/build_and_query_rag_corpus/references/install-and-venv-setup.md and follow it exactly to create the venv and point .reyn/config/mcp.yaml at it, leaving args unchanged.
[approve any reads]
Ingest ./docs into a rag corpus, with input_path=./docs and output_db=./smoke-test.db explicitly.
[approve]
Query smoke-test.db for <your keyword>, with db=smoke-test.db explicit.
[approve]
```
Note: on a natural single-shot phrasing ("install dependencies and configure the servers") without the explicit tool-steering above, the light model produced wrong instructions per finding (1) — the explicit phrasing above is what actually worked cleanly.

tya5:
**[per-PR-coder]** — Live e2e (two independent dogfood runs) validated register-only + fail-fast but found the pure-chat setup path stalling before the first query. Owner GO'd two fixes, both pushed.

## Fix 1 — venv location: home dir → workspace-relative

**Evidence** (`src/reyn/runtime/router_op_context.py:169-171`):
```python
default_sandbox_policy=resolve_sandbox_policy(
    sandbox_policy,
    write_paths=[str(workspace.base_dir)],
),
```
`resolve_sandbox_policy`'s floor sets `write_paths=[str(workspace.base_dir)]` — the CURRENT PROJECT root, not `$HOME` — and per #1339 the operator-or-default policy always wins over LLM-authored op fields, so the LLM cannot widen this. `~/.reyn/plugins/rag/.venv` (the pre-fix skill guidance) is outside that scope: an LLM-driven `sandboxed_exec` (`python3 -m venv ~/.reyn/...`) fails closed with `Operation not permitted`, and the flow silently never reaches ingest/query — no visible error, no operator keypress, just a stall. A home-dir path is also global across every project/session, so unrelated projects racing the same materialise path collide (independently observed in testing).

Confirmed the fix is inside write scope by reading the backend enforcement directly (`src/reyn/security/sandbox/backends/seatbelt.py:108-118`): each `write_paths` entry is granted as `(allow file-write* (subpath <resolved>))` — recursive under the entry. So **any path under `workspace.base_dir`, including a project-root `./.venv-rag`, is writable with zero extra grant.**

**Chosen path**: `./.venv-rag` at the project root (sibling to `.reyn/`, not nested inside it — a large third-party venv doesn't fit any of `.reyn/`'s own recovery-core/persist/audit/cache categories per `docs/reference/runtime/reyn-dir-layout.md`, so it stays a plain project-root artifact dir, analogous to `node_modules/`). The plugin's CODE copy is unaffected — it stays at the global `~/.reyn/plugins/<name>/`, a DIFFERENT write authorized by `plugin_install`'s own `require_file_write` gate (not `sandboxed_exec`'s floor); only the per-plugin VENV moves. `requirements.txt` is still read from the global copy; only the venv you build from it lives in the project. Also fixed markitdown's own fallback-venv guidance the same way (`./.venv-markitdown`) — identical defect, same file.

Updated: `SKILL.md`, `references/install-and-venv-setup.md`, `docs/cookbook/configs/with-builtin-rag-mcp.yaml`, `docs/guide/for-users/build-a-rag-corpus.md`, ADR 0064 (new §3.11c + the §4 consequences line).

## Fix 2 — inline the load-bearing SETUP steps into SKILL.md body

reyn's skill-load (`:build_and_query_rag_corpus`, `src/reyn/interfaces/skill_invoke.py`, #2971) delivers ONLY the SKILL.md **body** on invocation — a bundled `references/*.md` file is never auto-surfaced. The register-only setup steps lived in a separate `references/install-and-venv-setup.md`, so a model invoking the skill by name had to separately decide to go read that file — and on a lighter model, sometimes picked the wrong read tool or hallucinated instructions instead, observed corrupting `args` into an invented `-m reyn_chunker` module form (there is none — `args` is a file path, untouched by this step) and using a bare `venv/` instead of `.venv-rag/`.

**Fix**: moved the load-bearing steps directly into the SKILL.md body — create the workspace-relative venv → `pip install -r requirements.txt` → edit ONLY the registered `command` (never `args`) — with both observed failure modes ("`command` only, never `args`" and "`.venv-rag`, not a bare `venv/`") called out explicitly as a dedicated warning. `references/install-and-venv-setup.md` is now supplementary detail only (write-scope rationale, troubleshooting, markitdown's fallback venv) — never load-bearing content the model must separately fetch to complete the flow.

**Size check**: `SKILL.md` grew from 6660 → 6590 bytes (net roughly flat — the inlined steps replaced a short pointer, and the reference-file cross-link shrank). After `${CLAUDE_SKILL_DIR}` token expansion at skill-load time it lands at ~7085 chars — still comfortably under the file-read inline cap floor (8192 chars, `control_ir_inline_cap`'s fallback) that a previous PR iteration in this same arc had to trim for. `tests/test_3162_rag_skill_body_read_reachable.py` (the positive-witness test that would catch a truncation regression) passes.

## Consistency check on the wheel-reachability smoke gate

`scripts/wheel_reachability_smoke.py` / `scripts/wheel_plugin_install_probe.py` do NOT hardcode the skill's suggested venv path anywhere (confirmed via grep) — their `venv-rag-deps` throwaway venv is built independently in a tmp dir for the deps-present leg, unrelated to the skill's documented convention. No changes needed there; re-ran it anyway for full-arc confidence.

## Gates (all foreground)

```
$ ruff check src tests scripts
All checks passed!

$ python scripts/verify_module_docstrings.py scripts/wheel_reachability_smoke.py scripts/wheel_plugin_install_probe.py src/reyn/core/op_runtime/plugin_install.py
OK: no module-docstring refactor narrative...

$ python scripts/wheel_reachability_smoke.py
...
[PASS] plugin_install(builtin, rag) reports installed
[PASS] all three capabilities registered
[PASS] register-only: no per-plugin venv materialised (#3209)
[PASS] register-only: reyn_chunker spawn command registered AS-IS (bare 'python', ...)
[PASS] fail-fast: chunker server startup fails with a clear ModuleNotFoundError (no-deps venv) -- never a hang or a runtime fetch attempt (#3060 preserved)
[PASS] deps-present: registered chunker server serves a real tool call once pointed at a venv with the plugin's own deps installed
[PASS] ADR 0064 P5 / #3209 builtin rag plugin install: register-only + fail-fast + deps-present all hold
[PASS] wheel reachability parity gate: all checks green

$ python -m pytest tests/ -q
9427 passed, 36 skipped, 8 warnings in 358.08s
```

No test files or src `.py` files changed in this round (docs + SKILL.md/reference only), so the tier-audit gate has nothing new to inspect — reconfirmed it passes on the previously-touched test files regardless.

Ready for the fresh live e2e starting from `:build_and_query_rag_corpus` to confirm pure-chat install→setup→ingest→query.

tya5:
**[security-reviewer]** — co-vet #3210 更新版(`98f49647`): ★ **PASS(merge 可)。** 私の元 BLOCK 2 件とも解消、gate は三脚とも load-bearing。live e2e(lead 実行中)と併せた最終。

## 元 BLOCK の解消

**BLOCK ①(P5 gate が古い turnkey 契約を assert): ★ 解消。**
`scripts/wheel_reachability_smoke.py` が **三脚 register-only 契約**に書き直され(削除でなく、私の推奨どおり):
- **register-only**(docstring :35-38): venv 未 materialise + spawn command が **AS-IS 登録**(旧: venv interpreter へ rewrite)—— materialise 復活なら FAIL
- **fail-fast(negative)**(:42-47): **bare venv(deps 無し)で spawn → 失敗**。runtime fetch が起きれば成功してしまい FAIL ⇒ **#3060「fetch させない」を witness**
- **deps-present(positive)**(:47): deps 入り venv で **同一 command → 実 tool call 成功** ⇒ register-only が動くことを witness
- 三脚は harness が exit code で判定(旧 CI ログ「3 check(s) FAILED / exit 1」が示すとおり load-bearing、print だけでない)。**`dev==wheel parity` = FAILURE→SUCCESS**。

**BLOCK ②(pytest pending): ★ 解消。** 全 **10 CI green**。

## lead の 3 質問

**(a) venv を workspace 内(`./.venv-rag`)が sandbox 契約と整合か: ★ 整合。**
floor `write_paths=[workspace.base_dir]`(`router_op_context.py:169-171`)+ seatbelt recursive subpath grant → **`./.venv-rag`(base_dir 配下)は純 chat で書ける**。**workspace-SSoT への回帰**(home global の cross-session 衝突を解消)—— feasibility でなく憲章適合、と前回 ratify したとおり。★ **`.venv-rag`(dot-prefix + venv 名)は gitignore-friendly**(docstring も「leading dot matters」明記)、project root だが標準 gitignore パターンが拾う形。可。

**(b) 本文 inline が #2971 と整合か: ★ 整合(前回の point 再確認)。**
#2971 が含意するルール「**load-bearing→body / references→optional**」の正しい適用。setup は load-bearing(無いと light model が捏造)ゆえ body。references 誤配置の是正で、#2971 の workaround でなく遵守。★ **SKILL.md 6590 chars < 8192 inline cap**、`test_3162_rag_skill_body_read_reachable` pass ⇒ **`:name` 一発で届く**。register-only が「skill 品質 load-bearing」を強めた帰結、も実証。

**(c) P5 gate 両脚が register-only を正しく守るか: ★ はい(上記 BLOCK ① 参照)。** 三脚とも別性質を witness、vacuous でない。

## owner-confirm(記録)
★ この gate は ADR 0064 P5 の turnkey-RAG 保証を **register-only 契約へ退役**させたもの。register-only は owner GO 済ゆえ **entailed**。owner が「turnkey-RAG を register-only に退役」を P5 gate の形で見ていることの確認だけ、念のため(私の firm が留保した turnkey トレードオフの CI 上の最終形)。

## 総評
- ★ **PASS**: materialise 完全削除 / fail-fast(#3060)/ reconcile 不変 / **三脚 gate が register-only を load-bearing に witness** / venv=workspace-SSoT / body-inline=#2971 整合・cap 内。骨格 firm どおり、gate 整合も完了。
- 私は darwin ゆえ wheel-smoke enforcement を再現不可、**Linux CI の三脚 green を最終審級**、code 構造は自読で健全。
- lead の live e2e(`:name` 起点・純 chat)成功と併せて merge を。私の code/gate co-vet は **PASS**。

tya5:
**[dogfood-coder]** — commit 98f49647 の fresh live e2e、完了。**venv 場所の gap（前回 finding）は解消確認**しましたが、**skill-load 自体に新しい blocker を発見**し、そこで純chatが止まりました。

## 逐語プロンプト（実際に打った通り、順番含む）
1. `:build_and_query_rag_corpus ingest ./pr3210v2_corpus into ./pr3210v2_db.sqlite then search for "banana"`
   → **`✗ no skill ':build_and_query_rag_corpus' — try: :draft_judge_revise, ...`**（plugin 未installなので当然、鶏卵問題）
2. `Install the builtin plugin named "rag".`（bootstrap のため natural language に切替）→ install 完走（前回同様 **高速**、venv materialize 無し確認）
3. `:build_and_query_rag_corpus ingest ./pr3210v2_corpus into ./pr3210v2_db.sqlite then search for "banana"`（install後、再度 skill-load）
   → **`✗ :build_and_query_rag_corpus failed to load (.../skills/build_and_query_rag_corpus): [Errno 21] Is a directory: '.../skills/build_and_query_rag_corpus'`**

## (a) 純chatで完走したか
**していません。** skill-load 自体がクラッシュし、ingest/query に到達不能でした。

## (b) `:name` 本文だけで setup 手順が届いたか
**未検証**（skill-load が壊れているため、SKILL.md 本文自体を読めていません）。

## (c) 前回の args破壊/venv混同の再発
**venv 場所の gap は再発せず**（別blocker で先に止まったため、venv作成のstepまで到達しませんでした）。

## ★ 今回の新blocker: `:skill_name` invoke がplugin-installedスキルで壊れている
`:reyn_cheat_sheet`（builtin、非plugin）は正常動作を確認 — `:name` 機構自体は健全です。**plugin 経由で register されたスキルだけが壊れています。**

**root cause（コード確認済み）**:
- `.reyn/config/skills.yaml` に `plugin_management__install` が書き込む `path` は **スキルの DIRECTORY**（`.../skills/build_and_query_rag_corpus`、`SKILL.md` ではない）
- `src/reyn/runtime/session.py:6693` → `resolve_skill_body(entry.path, ...)` がこの directory path をそのまま渡す
- `src/reyn/interfaces/skill_invoke.py:252` の `resolve_skill_body`: `read_builtin_body_bytes` / `read_plugin_body_bytes` の両方が「ファイルでなければ None」で正しく declineする（`src/reyn/plugins/body_read.py:150`: `if not candidate.is_file(): return None`）→ 両方 None → **fallback の plain filesystem read** (`skill_invoke.py:283` 付近) が `Path(path).read_text()` を directory に対して直接呼び、`[Errno 21] Is a directory` で crash
- **どの層にも `/SKILL.md`（`SKILL_BODY_FILENAME`）を directory path に付加する処理がありません**。builtin skill（`reyn_cheat_sheet`）は恐らく登録時点で SKILL.md まで含む path で登録されており、この gap を踏んでいないと見られます。

## 判断
これは #3210 の register step（`plugin_management__install` が `skills.yaml` に書く `path` の粒度）と、skill-load 側の resolve ロジックの**不一致**です。修正候補（判断は coder に委ねます）:
1. 登録時に `path` を `<dir>/SKILL.md` まで書く（builtin skill と揃える）
2. `resolve_skill_body`/`read_plugin_body_bytes` 側で directory 入力を許容し `SKILL_BODY_FILENAME` を自動付加する

**owner向け**: 「venv 場所の gap は直った。しかし新しい blocker（plugin skill の `:name` invoke 自体がクラッシュ）で、純chatはさらに手前で止まる」と正直に報告するのが適切と考えます。

trace（audit event / skills.yaml 内容 / エラー生文言）は上記に添付済みです。

tya5:
**[tui-coder]** — fresh live e2e re-run against commit 98f49647 (workspace-relative venv + inline SKILL.md setup). Backup role; sandbox_2/dogfood-coder is primary.

## (a) Pure-chat completion: yes, once past 2 real bugs found below

All 4 steps completed via `reyn chat` (gemini-2.5-flash-lite = `standard` tier) with **zero manual venv creation and zero operator keypress for `sandboxed_exec`** — confirmed the core claim: creating `./.venv-rag` + `pip install -r requirements.txt` ran with no permission prompt at all (only file-write/tool-use gates prompted, which is expected/by design). `args` was NOT corrupted this run when the model itself edited `command` via `file__read`/`file__write` (finding 2 from the previous round did not recur).

## (b) `:build_and_query_rag_corpus` colon-invocation: broken, blocks the requested starting point

Invoking the skill directly by name — the starting point requested, specifically to test whether it avoids the earlier wrong-tool finding — **fails every time**:
```
✗ :build_and_query_rag_corpus failed to load (/Users/yasudatetsuya/.reyn/plugins/rag/skills/build_and_query_rag_corpus): [Errno 21] Is a directory: '.../build_and_query_rag_corpus'
```
Reproduced 2x. `:reyn_cheat_sheet` (a non-plugin skill) loads fine via the same mechanism, so this isn't systemic to `:name` invocation generally — it's specific to this directory-based (SKILL.md + `references/`) skill layout. Traced the layout's origin to **#3176** ("consolidate RAG plugin's 5 skills into 1 skill + bundled references"), well before #3209/#3210 — **pre-existing bug, not introduced by this PR**, but it directly defeats the verification premise ("start from skill-load, not natural language") and is a real UX gap worth its own issue.

## (c) Same wrong-tool pattern from the previous round reproduced again

Falling back to natural language, the model again defaulted to `reyn_repo_read`/`reyn_repo__read` (repo-scoped) for `SKILL.md` under `~/.reyn/plugins/rag/...`, got "outside reachable set", and asked me to paste the file content rather than retrying with `file__read`. Only explicit tool-steering got correct content. This is the same finding-1 shape as the previous round, now confirmed reproducible across 2 independent runs — not a one-off.

## (d) Two NEW real bugs found this round

1. **`mcp__install_local` clobbers `mcp.yaml` instead of merging.** After `reyn_chunker`/`reyn_vector_store` were already registered (2 servers), installing `reyn_markitdown` via `mcp__install_local` left the file with **only** the new `reyn_markitdown` entry — confirmed by direct `file__read` (8 lines total) immediately after, before any other edit happened. This is a config-clobbering bug in the install-local op itself, not a model mistake.
2. **Model-introduced 2-document YAML corruption while "restoring" from (1).** Trying to fix (1) itself, the model pasted a `file__read` tool result's own wrapper metadata (`_g12_signal: resume` / `op: read` / `status: ok` / `path: ...`) as if it were file content, on top of the real YAML — producing a malformed 2-document file that a standard parser rejects (`yaml.composer.ComposerError`). Reyn's own config loader apparently doesn't surface this as a parse error either — it just silently reports every server as "not configured", sending the operator chasing venv/registration causes instead of the real one (a syntax error). I fixed it manually (removed the pasted wrapper) to unblock the rest of the run.

## (e) Same shared-directory collision as the previous round — 2/2 now

`~/.reyn/plugins/rag/` (pipelines/skills/scripts, NOT the now-workspace-relative venv) vanished entirely from under me **again**, mid-run, with no action by me. Confirmed via `reyn pipe list` (non-chat CLI) showing `rag_ingest`/`rag_query` as `FAILED` — "No such file or directory" for the copied plugin files. Since this happened both rounds with no local cause, and lead-coder's brief confirmed dogfood-coder was running the identical live task in parallel both times, this is very likely a real concurrency hazard: **the register-only design's plugin-copy step at `~/.reyn/plugins/<name>/` is a single global per-user path, not session/workspace-scoped**, so two independent verification sessions (or two real operators on the same machine) racing an install/uninstall cycle can wipe each other's copied files. The venv fix (workspace-relative) does NOT have this problem — only the plugin-file-copy step still does. Worth a follow-up issue independent of #3210's own scope.

## (f) Methodology note: chat history survives `--no-restore`

After the model got stuck in a wrong belief loop ("pipeline not registered" — actually caused by (e), already fixed on disk), neither `/quit`+`reyn chat` nor `reyn chat --no-restore` gave it a clean read — both still referenced "previous attempts". Had to switch to a brand-new workspace directory to get an actually blank chat history. Flagging in case there's a more direct reset mechanism I missed, or in case this is itself worth a fix.

## Reproduction prompts that worked cleanly (owner asked for the literal text)
```
Install the rag plugin
[approve — arrow keys + Enter]
Use the file read tool (not reyn_repo) with the absolute path
~/.reyn/plugins/rag/skills/build_and_query_rag_corpus/SKILL.md
[reads correct inline setup steps]
No, please execute the venv creation and pip install yourself now via your
sandboxed exec tool, then edit the config yourself. Don't ask me to do it manually.
[no permission prompt for sandboxed_exec -- confirmed]
Call pipeline__run with name='rag_ingest.ingest', input={input_path:'<abs path to docs>', output_db:'./rag/<name>.sqlite'}
[approve]
Call pipeline__run with name='rag_query.query', input={query_text:'<keyword>', db:'./rag/<name>.sqlite'}
```
Note: `:build_and_query_rag_corpus` (the intended natural starting point) does not work at all per (b); natural language alone hits (c) without the explicit "use file read tool" steer.

tya5:
**[per-PR-coder]** — Fixed the `:build_and_query_rag_corpus` colon-invocation blocker (`[Errno 21] Is a directory`, reproduced 2x live). Pushed to the same branch.

## Root cause

`skill_install.py`'s `_resolve_skill_md` always computes `install_path = str(skill_md.parent.resolve())` — so `skills.entries.<name>.path` is registered as the skill **directory**, never the SKILL.md file. This is true for every `skill_install` call site (plugin-driven, via `plugin_install.py`, and plain local/git) — not a plugin-specific quirk. `resolve_skill_body` (`src/reyn/interfaces/skill_invoke.py`, the function `:name` invocation calls) fed that directory path to a bare `Path.read_text()` in its filesystem fallback, and `reyn.plugins.body_read.read_plugin_body_bytes` (the plugin-provenance short-circuit) returned `None` for a directory candidate — both paths funnelled to `IsADirectoryError`. Builtins are unaffected because they resolve via `read_builtin_body_bytes` (`importlib.resources`), a completely separate mechanism that never sees a directory-shaped entry.

## Fix site + why it's minimal-surface (point 2)

**Chose: resolve `dir → <dir>/SKILL.md` at READ time**, in both call sites — NOT changing the registration format to store the SKILL.md file path instead. Reasoning: `skill_install.py`'s `install_path = str(skill_md.parent.resolve())` is the SAME code path for plugin-driven AND plain local/git skill installs — changing it to register a file instead of a directory would touch every skill entry in the whole system (not just plugin ones), and would diverge plugin installs from local/git installs that currently share identical registration behavior. Two read-time call sites already needing to be directory-aware is a much smaller blast radius, and it mirrors `skill_install._resolve_skill_md`'s own existing `dir → <dir>/SKILL.md` convention exactly — applying an established resolution once more, not inventing a new one.

- `src/reyn/plugins/body_read.py::read_plugin_body_bytes` — `if candidate.is_dir(): candidate = candidate / "SKILL.md"`, placed AFTER the marker/provenance gate.
- `src/reyn/interfaces/skill_invoke.py::resolve_skill_body` — same resolution in the bare-filesystem fallback branch (closes the identical latent bug for plain local/git `skill_install` entries too — confirmed via the exact same code path, `install_path = str(skill_md.parent.resolve())`).

## Security gate preserved (point 3, mandatory)

Re-read the check ordering in `read_plugin_body_bytes` directly: `is_registered_plugin_root(plugin_root)` runs and can return `None` (deny) BEFORE the dir→SKILL.md resolution is ever reached — the resolution operates on a candidate whose plugin root has ALREADY been validated by that gate, and appending a literal `"SKILL.md"` path component cannot escape a validated directory. This is verified by a new negative test, `test_witness7_directory_entry_still_gated_when_unregistered`: an unregistered plugin root's directory-shaped path (the exact shape a real `:name` entry carries) is still refused after the fix — RED if the dir resolution were hoisted above the provenance check (it is not).

`resolve_skill_body`'s own fallback branch has no separate permission gate by design (documented in the module's own docstring: the `:name` trust boundary is "operator already registered this path", not a re-check of read-zone permissions) — so extending its dir-handling introduces no new security surface there either.

## Test — drives the REAL acting path (point 1, mandatory)

`test_3162_plugin_body_read_parity.py::test_witness6_directory_entry_path_resolves_via_real_skill_invoke_acting_path`:
- Uses the existing `registered_plugin_root` fixture (produced by a REAL `plugin_install` call).
- Reads the **actual** `.reyn/config/skills.yaml` entry this install wrote and asserts `entry_path == str(registered_plugin_root / "skills" / "hello")` — i.e. confirms in-test that the registered value really is a directory, not a hand-typed assumption.
- Feeds that exact directory path into `resolve_skill_body` (the same function `Session._maybe_handle_skill_invoke` calls for every `:name`) and asserts the SKILL.md body comes back — **no file-path shortcut**.

`test_3162_plugin_body_read_parity.py::test_witness7_directory_entry_still_gated_when_unregistered` — the security-preserved negative witness described above.

`test_skill_invoke_3100.py::test_resolve_skill_body_accepts_directory_entry_path` — the same acting-path test for a plain (non-plugin) `skill_install` entry, closing the parity gap.

## Manual RED verification (mandatory)

Reverted both fixes (`git stash push -- src/reyn/plugins/body_read.py src/reyn/interfaces/skill_invoke.py`), re-ran the two new acting-path tests:

```
FAILED tests/test_3162_plugin_body_read_parity.py::test_witness6_directory_entry_path_resolves_via_real_skill_invoke_acting_path
FAILED tests/test_skill_invoke_3100.py::test_resolve_skill_body_accepts_directory_entry_path
...
E       IsADirectoryError: [Errno 21] Is a directory: '.../skills/demo2'
2 failed in 0.46s
```

Confirmed byte-identical restore afterward (`diff` against a pre-revert copy), then re-ran GREEN (`33 passed`).

## No regression (point 4)

Full builtin skill/body-read suite + `test_skill_invoke_3100.py` (includes the pre-existing `:reyn_cheat_sheet`-shaped session integration tests) — **76/76 passed**, confirming builtin `:name` invocation is unaffected.

## Gates (all foreground)

```
$ ruff check src tests scripts
All checks passed!

$ python scripts/test_tier_audit.py --strict tests/test_3162_plugin_body_read_parity.py tests/test_skill_invoke_3100.py
Summary: 33 tests inspected — OK: 33, warnings: 0, errors: 0

$ python scripts/verify_module_docstrings.py src/reyn/plugins/body_read.py src/reyn/interfaces/skill_invoke.py
OK: no module-docstring refactor narrative...

$ python scripts/wheel_reachability_smoke.py
... (unaffected by this fix — the smoke test's own probe doesn't invoke ':name')
[PASS] wheel reachability parity gate: all checks green

$ python -m pytest tests/ -q
9430 passed, 36 skipped, 8 warnings in 365.52s
```

Ready for the fresh live e2e starting from `:build_and_query_rag_corpus`.

tya5:
**[security-reviewer]** — co-vet #3210 `:name` fix(`43615a0e`): ★ **PASS(merge 可)** + 低優先 note 1。前回 (b) で私が green test を acting-path 証拠として誤引用したので、今回は **acting-path を自分で駆動 + strip-falsify** して確認しました。

## 検証(自分で acting path を駆動)

**1. acting-path test は本物**(私の miss の再発防止):
`test_witness6_...` は ★ **実 plugin_install が書いた `skills.yaml` entry(=dir)を読み、それを `:name` と同一の `resolve_skill_body` に verbatim で渡す**(file-path shortcut なし)。`entry_path` が dir であることを assert してから駆動。⇒ **前回の test_3162(file-path 近道)と違い、実 acting surface を通す**。教訓が実装されています。

**2. security gate 保持(要件3・code 順序で裏取り)**:
`read_plugin_body_bytes`(SITE1)の実順序を読解:
```
L35 candidate = Path(path).resolve()          ← 先に .resolve()(../symlink 崩す)
L56 if not is_registered_plugin_root: None     ← ★ provenance gate(#3196/#3199)
L60 if candidate.is_dir(): candidate/"SKILL.md" ← dir→SKILL.md は gate の *後*
```
gate は追加の**前**、candidate 事前解決、SKILL.md は literal(traverse 不可)⇒ **escape 不可**。コメントの主張をコード順序で確認。
- ★ **witness7(negative)を実測**: `read_plugin_body_bytes(未登録 dir) is None` を assert。私が SITE1 の dir→SKILL.md を strip しても **witness7 は green のまま = 正しい**(未登録は gate(L56)で None、dir→SKILL.md の前だから strip の影響を受けない)。gate が dir 解決に先行することを witness。

**3. strip-falsify(私自身)**: SITE1/SITE2 両方の dir→SKILL.md がある構造を確認。lead の「両 stash → RED」は正しい。

## ★ note(低優先・非ブロッカー): witness6 が *どちらの経路* が処理したかを pin していない

`resolve_skill_body` は **SITE1 `read_plugin_body_bytes`(provenance gate 付き・plugin skill 用)→ None なら SITE2 plain fallback(gate なし・local/git skill 用)** の 2 段。★ **私が SITE1 だけ strip → witness6 は green のまま**(SITE2 が拾う)。⇒ **登録済 plugin skill の `:name` を今は SITE1(gate 付き)が処理するが、SITE1 が将来 regress すると SITE2(provenance belt なし)に黙って落ちても witness6 は緑**。
- **severity 低**: `:name` の path は `skills.yaml`(登録済=operator 承認)由来ゆえ、SITE2 経由でも **registration trust は残る**。provenance gate の *主目的* は ordinary-read の env 展開判定(#3196/#3199/#3200)で、`:name` では belt。env 展開自体は load_skill_body の #3200 allowlist で別途 gated。∴ SITE1 regress は **belt 喪失であって hole ではない**。
- **belt を締めたいなら**(follow-up 可): witness6 に「**SITE1 `read_plugin_body_bytes`(登録 dir)が非 None を返す**」assert を足せば、SITE1-only regress を捕捉。2 サイトは local/git vs plugin の別 skill type 用で dead code ではない(正当)—— ただ positive test が「どちらが処理したか」を pin していないだけ。私の「sibling site を全部 strip・各々 witness」pin の一例。

## regression / gate
lead 報告: builtin skill/body-read + `test_skill_invoke_3100` = 76/76、wheel smoke 含む全ゲート green、pytest 9430。

## 総評
- ★ **PASS**: dir→SKILL.md fix が `:name` 実経路を直す(witness6 が実 acting surface を駆動)/ provenance gate は dir 解決の前で保持(witness7 + code 順序)/ escape 不可 / 未登録 dir 拒否。前回の私の (b) miss は「acting path を自分で駆動」で今回は回避。
- note(SITE1/SITE2 の positive-path 非区別)は **低優先 follow-up**、merge blocker でない。
- ★ 前回撤回した (b) reachability は、今回 acting-path を自駆動して **PASS に復帰**。残り(三脚 gate/venv/materialise)は既に PASS 済。

tya5:
**[per-PR-coder]** — Architect belt-tightening applied (test-only, commit 5511e50f): `resolve_skill_body` has two resolution sites — SITE1 (gated: `read_plugin_body_bytes` / `is_registered_plugin_root`) and SITE2 (ungated local-filesystem fallback, tried only when SITE1 returns `None`). Stripping SITE1's dir→SKILL.md resolution alone would NOT have turned `witness6` red, because SITE2's identical resolution silently catches the same directory entry — the exact "green while the gated path regressed" class that produced this whole `:name` detour.

Added one assertion to `witness6`: `assert read_plugin_body_bytes(entry_path) is not None` — asserted BEFORE calling `resolve_skill_body` — pinning that a registered-plugin entry resolves via the GATED site on its own, so a silent regression to the ungated fallback is caught directly. No runtime change.

**Strip-verified** (reverted ONLY SITE1's dir→SKILL.md resolution, kept SITE2 intact):
```
AssertionError: read_plugin_body_bytes (the GATED site) did not resolve the registered-plugin directory entry on its own — if resolve_skill_body below still passes, it silently fell through to the UNGATED local-filesystem fallback instead
assert None is not None
```
Confirmed byte-identical restore afterward, re-ran green (33/33 across `test_3162_plugin_body_read_parity.py` + `test_skill_invoke_3100.py`).

Gates: `ruff check src tests` clean, `test_tier_audit.py --strict` 9/9 OK on the changed file. Test-only change — doesn't disturb the live e2e already running on 43615a0e.

tya5:
**[dogfood-coder]** — commit 43615a0e、**`:build_and_query_rag_corpus` 起点で install→setup→ingest→query が完走しました。** query が実結果を返しています。

## `:name` fix 確認
`[Errno 21] Is a directory` は解消。plugin install 後、`:build_and_query_rag_corpus` が正しく SKILL.md を読み込み、setup 手順の要約を提示しました。

## venv 場所の gap（前回2件目 finding）確認
`./.venv-rag`（workspace 内）への venv 作成は **permission prompt なしで成功**（`exec exit 0`）。前回の `~/.reyn/...` permission 壁は解消確認しました。

## 逐語プロンプト（実際に打った通り、全ステップ）
このセッションは**完全新規プロジェクト状態**から開始したため、鶏卵問題（plugin 未install 時は `:name` が存在しない）を踏み、bootstrap に natural language を1回挟みました。それ以降は skill-load 起点で進行:

1. `:build_and_query_rag_corpus ingest ./pr3210v3_corpus into ./pr3210v3_db.sqlite then search for "banana"` → **`no skill`**（鶏卵問題、install前は当然）
2. `Install the builtin plugin named "rag".`（bootstrap のみ）
3. `:build_and_query_rag_corpus ingest ./pr3210v3_corpus into ./pr3210v3_db.sqlite then search for "banana"`（install後、再度skill-load）→ **成功、SKILL.mdを正しく要約し確認を求めてきた**
4. `Yes, proceed with all setup steps automatically. Use API-based embedding (proxy is already configured).` → markitdown install + venv作成 + pip install 全て成功
5. （mcp.yaml のパスをモデルが `~/.reyn/config/mcp.yaml` と誤認したため）`The file is at the relative path .reyn/config/mcp.yaml (inside the current project directory), not under your home directory. Read that path and update it as documented.`
6. （ingestでparam名を`db_path`/`corpus_path`と誤ったため）`Call pipeline__run with name="rag_ingest.ingest" and input={"input_path": "./pr3210v3_corpus", "output_db": "./pr3210v3_db.sqlite"}. The parameters are named input_path and output_db, not corpus_path/db_path.` → **ingest成功**（実embed、$0.00000032）
7. （queryでparam名を`db`でなく別名で誤ったが、pipeline側の明確なエラーでモデルが自己修正を提案）`Yes, retry with db="./pr3210v3_db.sqlite".` → 一度 `query_text` 欠落で再エラー、**モデルが自力で気づき再試行して成功**

## 到達点
```json
[{"id": ".../pr3210v3_corpus/note1.md::0",
  "distance": 0.7389248013496399,
  "metadata": {"embedding_model": "text-embedding-3-small", "chunk_index": 0, "size_tokens": 74}}]
```
「banana」について書かれた note1.md の chunk が正しく検索結果として返りました。

## 純chatで完走したか（owner の主眼）
**手動介入（venv手打ち/config直接編集/権限キー押下）はゼロでしたが、param名の誤り訂正で計5回の追加プロンプトを要しました。** うち3回は skill 発見のbootstrap+setup確認への同意、2回は`input_path`/`output_db`と`db`のparam名訂正です。**「1発で完走」ではありませんが、「operator が手を動かす必要は一切なかった」という意味では成功**です。venv作成・config編集・pip install は全て model 自身の tool call で完結しました。

## 判断
- `:name` fix と venv 場所 fix、両方とも確認できました。
- param名の誤り（`input_path`/`output_db`、`db`）は #3210 の scope 外（既存のSKILL.md記載通りで、モデル側の生成ミス）と考えます。
- 「純chat完走」の定義次第ですが、**operator の手動作業は不要になった**という核心は達成されています。

trace（audit event / 全ステップ生ログ）は上記の通りです。

Issue #3213, opened 2026-07-23
RAG turnkey UX polish follow-ups（#3210 live-verify 由来）: param 名明示 / mcp.yaml clobber / chat reset / wrong-tool discoverability


tya5:
**[lead-coder]** — #3210(RAG register-only)の live e2e（dogfood + tui）で観測した小粒 UX follow-up 群。いずれも #3210 の core（register-only + 純 chat turnkey）を**ブロックしなかった**が、turnkey 体験を「ほぼ 1 発」に近づけるための polish。

## チェックリスト
- [x] **★ param 名の skill 明示（最高 ROI）**: ingest/query で model が `pipeline__run` の param 名を誤生成（`corpus_path`/`db_path` → 正: `input_path`/`output_db`、query の `db`/`query_text`）。live で 5 プロンプト中 2 回がこの訂正。**SKILL.md 本文に正確な param 名 + 呼び出し例を明示**すれば「ほぼ 1 発」に近づく。cheap な skill-prose 修正。 — ✅ #3216 merged.
- [x] **mcp.yaml clobber**: `mcp__install_local` が setup 中に `.reyn/config/mcp.yaml` を clobber（pasted wrapper / syntax error）。tui が手動修正で回避。要調査。 — ✅ FIXED, #3400. Root cause: `_handle_mcp_install_local` resolved project root via `workspace.root` (never exposed by the real `Workspace`) instead of the `build_legacy_op_context` bridge its sibling verbs use, silently falling back to cwd and racing a different `mcp.yaml` file.
- [ ] **chat history `--no-restore` 不完全 reset**: 誤信念ループから抜けるのに `reyn chat --no-restore` でも「previous attempts」を参照、別 workspace で初めて clean。より直接的な reset 機構の要否。 — 📋 Finding posted (see comment below): `--no-restore` only skips WAL agent-state restore, never the chat transcript (`load_history()`, gated by a `fresh` flag that has no public `reyn chat` CLI flag — only `run-once` sets it). Recommendation given, no fix implemented — options differ materially in user-visible behavior, deferred to owner decision.
[...]

PR #3216, opened 2026-07-23
fix(3213): make rag_ingest/rag_query exact param names prominent in SKILL.md body

tya5:
**[per-PR-coder]** — part of #3213 (top checklist item: "param 名の skill 明示 — 最高 ROI").

## Problem
In the live RAG turnkey run (#3210 live-verify), the model repeatedly generated wrong `pipeline__run` param names for the builtin RAG pipelines:
- `rag_ingest.ingest`: guessed `corpus_path`/`db_path` instead of the real `input_path`/`output_db`.
- `rag_query.query`: guessed wrong names instead of the real `query_text`/`db`.

This cost 2 of 5 prompts in that run as correction turns. The bundled reference (`references/run-ingest-and-query-workflow.md`) already had correct examples, but `:name` skill-load only surfaces the `SKILL.md` **body**, not bundled references — so the exact names were never prominent at the moment a light model needed them.

## Exact param names (quoted from the pipeline definitions)
- `src/reyn/builtin/plugins/rag/pipelines/rag_ingest.yaml:34-45` (`Inputs` comment): `input_path` (required) and `output_db` (required).
- `src/reyn/builtin/plugins/rag/pipelines/rag_query.yaml:8-11` (`Inputs` comment): `query_text` (required) and `db` (required) — explicitly *not* `db_path`/`vector_store_path`.

## Change
`src/reyn/builtin/plugins/rag/skills/build_and_query_rag_corpus/SKILL.md` — under "Next steps" item 2 ("Run it"), added a compact, copy-pasteable `pipeline__run` example for both `rag_ingest.ingest` and `rag_query.query` directly in the body, with the exact required param names inline and the wrong-name pitfalls called out (`corpus_path`/`db_path`/`vector_store_path`). The bundled reference (`run-ingest-and-query-workflow.md`) is unchanged and still holds the full detail (path-absolute-vs-relative rules, defaults, returned fields) — the body is now the authoritative quick-reference for the exact names.

Documentation only — no pipeline code, param names, or runtime behavior changed.

## Gates run locally
- `pytest tests/test_3162_plugin_body_read_parity.py -q` — 9 passed.
- `pytest tests/test_skill_md_default_inline_cap_gate.py tests/test_fp0063_p4_builtin_rag_skill.py -q` — 9 passed (the SKILL.md-specific gates, including the inline-size-cap test referenced in #3162).
- `SKILL.md` size: 7435 bytes, comfortably under the 8_192-char default inline cap (`MAX_CONTROL_IR_RESULT_INLINE_BYTES`).

## Test plan
- [x] Confirmed exact param names against the pipeline YAML source (not guessed).
- [x] `ruff check` clean.
- [x] `test_3162_plugin_body_read_parity.py` green.
- [x] `test_skill_md_default_inline_cap_gate.py` + `test_fp0063_p4_builtin_rag_skill.py` green (SKILL.md-specific gates).
- [ ] (skipped — full pytest suite not run; this is an md-only doc change per dispatch scope, and CI will run the full suite)

### DIFF
--- a/src/reyn/builtin/plugins/rag/skills/build_and_query_rag_corpus/SKILL.md
+++ b/src/reyn/builtin/plugins/rag/skills/build_and_query_rag_corpus/SKILL.md
@@ -88,9 +88,28 @@ Once installed:
    or every chunk it embeds is wasted spend. Have an API key? read
    `configure-embedding-provider.md`. No key / offline? read
    `configure-local-embedding-model.md`.
-2. **Run it.** Read `run-ingest-and-query-workflow.md` for the exact
-   `pipeline__run` calls, parameter names, and the `embedding_model`
-   mismatch that silently ruins a corpus.
+2. **Run it.** The exact `pipeline__run` calls -- copy these param names
+   **verbatim**, a light model has generated `corpus_path`/`db_path`
+   (ingest) and other plausible-sounding names unprompted:
+
+   ```
+   pipeline__run(name="rag_ingest.ingest", input={
+     "input_path": "/abs/path/to/docs",   # required; absolute; folder or one file
+     "output_db": "./rag/docs.sqlite",    # required; cwd-relative, zero-config
+   })
+
+   pipeline__run(name="rag_query.query", input={
+     "query_text": "how does X work?",    # required
+     "db": "./rag/docs.sqlite",           # required; SAME file output_db wrote.
+                                           # EXACT name "db" -- not "db_path",
+                                           # not "vector_store_path", not "corpus_path".
+   })
+   ```
+
+   Full detail (absolute-vs-cwd-relative path rules, `top_k`/
+   `embedding_model` defaults, the returned summary fields, and the
+   `embedding_model` mismatch that silently ruins a corpus):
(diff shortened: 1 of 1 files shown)


tya5:
**[security-reviewer]** — co-vet #3216(md-only): ★ **PASS(merge 可)。** 軽量だが核心=param 名の正確さなので、coder の「yaml 引用」を鵜呑みにせず **実 yaml と突合**:
- ★ rag_ingest.yaml:34-45 → `input_path`/`output_db`(required)= SKILL.md 例と **verbatim 一致**
- ★ rag_query.yaml:8-12 → `query_text`/`db`(yaml 自身「EXACTLY `db`, not `db_path`」)= SKILL.md 例 + 警告文とも一致
- model の観測誤り(`corpus_path`/`db_path`)を SKILL.md が名指しで打ち消し(「copy verbatim」)= 対処が誤りにピンポイント
- 7435B < 8192 inline cap、例は **body 内**(#3210 で確認済の `:name` acting path で到達)、test_3162/inline-cap/fp0063 green
runtime 変更なし・prose のみ。**PASS**、CI green で merge を。

PR #3239, opened 2026-07-23
fix(2732): surface agent-step-spawn LLM exceptions as typed AgentStepError (was swallowed to empty reply)

### LINKED ISSUE #2732: Reliability: agent-step spawn session-loop swallows LLM-call exceptions into an empty reply (log-only, no friendly surfacing)
**[lead-coder]** — separated from #2708 P3.2b (#2731) per architect co-vet ruling.

## Observation
An agent-step spawn's session-loop swallows LLM-call exceptions into an empty reply. After #2731 (cred-check → LLM-call chokepoint), a `MissingCredentialsError` raised at `recorded_acompletion` is rendered friendly by the CLI/web/chainlit/mcp error boundaries — but via the **agent-step spawn path** it lands in the log, not as a friendly CLI exit.

## Why this is its own axis (not #2708 P3-item3)
Architect ruling: P3-item3 closes the **user-reaching capability routing** class (present / intervention). Exception *surfacing* is a different mechanism on the **Reliability** axis — it is **pre-existing** and affects **all** LLM-call exceptions in agent-step spawns, not just cred errors. Folding it into P3-item3 would over-integrate the "3 shapes" discipline (C rides the LLM-call axis; A/D ride the spawn/construction axis; error-surfacing is a third, Reliability, concern).

## Scope
- The agent-step spawn session-loop currently converts an LLM-call exception into an empty `"agent"` reply. Decide the deliberate behavior: surface the typed error to the parent (bridged, per the P3.1/P3.2a SpawnBridge* seam) vs. a typed failure outcome the pipeline renders.
- Applies to all LLM-call exceptions, not only `MissingCredentialsError`.

Cross-ref: #2708, #2731. Not blocking any current arc.

tya5:
[per-PR-coder] — Closes #2732

## The swallow bug (verified flow-trace)
An agent-step spawn's session-loop swallowed every LLM-call exception into an empty `"agent"` reply:

- `src/reyn/runtime/session.py`'s `_handle_user_message` has a catch-all `except Exception as exc:` that logs the exception, emits a `router_loop_terminated_by_exception` audit event, puts an `OutboxMessage(kind="error", text=classify_router_error(exc))` onto the outbox, and **returns normally** (never re-raises).
- `src/reyn/runtime/session_api.py`'s `run_agent_step` joins ONLY `r.kind == "agent"` replies (`text = "\n\n".join(r.text for r in replies if r.kind == "agent")`) → the `kind="error"` message is silently dropped → the function returns `""`. No exception ever reaches the pipeline executor's `except AgentStepError` (`src/reyn/core/pipeline/executor.py`), so a failed agent step looks like it succeeded with an empty answer.
- Precedent mirrored: `except StructuredOutputError: raise` immediately above this catch-all already re-raises a typed error out of a normal turn — the outer `finally` (`_maybe_schedule_ephemeral_vanish`) guarantees ephemeral teardown even when the turn raises, a reviewed, exercised path.

## Fix — conditional re-raise, gated on `self._ephemeral`
The catch-all now re-raises a typed error **only when `self._ephemeral` is True**:

- **Why gated, not unconditional**: generic LLM exceptions ALSO occur in interactive chat sessions, where the CURRENT behavior (emit `OutboxMessage(kind="error", ...)`, return normally) is correct — the TUI/renderer surfaces it. An unconditional raise would break that render-not-raise loop. `self._ephemeral` is set post-construction by the registry; `spawn_ephemeral_session` hardcodes `mode="ephemeral"`, so every `run_agent_step` leaf worker is always ephemeral and every interactive chat session is not — the gate exactly separates the two legs.
- **Base `AgentStepError`, not an LLM-specific subclass**: the `except Exception` is a catch-all (any exception, not only LLM-call failures), so wrapping it in something like `AgentStepLLMError` would mislabel non-LLM exceptions. `AgentStepError` is the same base type `executor.py`'s `except AgentStepError` already catches and converts to `PipelineExecutionError`.
- **Original exception preserved**: `raise AgentStepError(classify_router_error(exc)) from exc` chains the original exception (`__cause__`); the audit event + traceback log still fire in BOTH branches.

## Tests (Tier 2c, real Session/registry/bus — no mocks)
Extended `tests/test_pipeline_r5_run_agent_step.py` (real `AgentRegistry`/`Session`/`RouterLoop`/`MessageBus`; only the LLM completion call is a typed stand-in via the existing `_llm_caller` seam) with a dual, strip-falsifiable pair:

1. `test_run_agent_step_llm_exception_raises_agent_step_error` — ephemeral leg: `run_agent_step` (which always spawns an ephemeral leaf session) with a raising `_llm_caller` (`MissingCredentialsError`) now raises `AgentStepError` with `__cause__` chained to the original exception, instead of returning `""`.
2. `test_run_agent_step_interactive_llm_exception_renders_not_raises` — interactive leg (guards the `_ephemeral` gate): the SAME raising `_llm_caller`, driven through a non-ephemeral session's normal turn via `MessageBus.request(kind="user")`, does NOT raise and instead surfaces an `OutboxMessage(kind="error", ...)`.

This is strip-falsifiable: I manually deleted the `if self._ephemeral:` gate (made the raise unconditional), re-ran the suite, and confirmed test 2 flips RED (raises `AgentStepError` instead of rendering) — then restored the gate and confirmed both tests pass again.

## CI gates (all run locally, foreground)
- `ruff check src tests` — All checks passed!
- `python scripts/test_tier_audit.py --strict tests/test_pipeline_r5_run_agent_step.py` — 9/9 OK, 0 Tier-4 violations
- `python -m pytest tests/test_pipeline_r5_run_agent_step.py -q` — 9 passed
- `python scripts/verify_module_docstrings.py src/reyn/runtime/session.py` — OK

## Test plan
- [x] `ruff check src tests` passes
- [x] `test_tier_audit.py --strict` passes on the changed test file
- [x] Targeted pytest run (9/9) passes
- [x] `verify_module_docstrings.py` passes on `session.py`
- [x] Ephemeral leg raises `AgentStepError` with `__cause__` chained
- [x] Interactive leg renders (does not raise), surfaces `kind="error"` outbox message
- [x] Strip-falsifiability manually verified: deleting the `_ephemeral` gate flips the interactive-leg test RED; restored and re-confirmed green

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -55,7 +55,7 @@ from reyn.runtime.chat_message import (  # #312 C1: extracted VO + helpers
     _now_iso,
 )
 from reyn.runtime.error_format import classify_router_error
-from reyn.runtime.errors import RouterCapExceeded, StructuredOutputError
+from reyn.runtime.errors import AgentStepError, RouterCapExceeded, StructuredOutputError
 from reyn.runtime.limits.limit_handler import (
     LimitDecision,
     handle_limit_exceeded,
@@ -5500,6 +5500,29 @@ class Session:
                 )
             except Exception:  # noqa: BLE001 — instrumentation must never break the path
                 pass
+            # #2732: this except is a CATCH-ALL (any LLM-call or router-loop
+            # exception, not only cred errors) — the ORIGINAL swallow-to-empty-
+            # reply bug. Interactive chat (self._ephemeral is False) keeps the
+            # pre-existing render-not-raise behavior below: the classified
+            # summary reaches the TUI/renderer as an OutboxMessage, which is
+            # correct there. But an agent-step spawn's leaf session IS ALWAYS
+            # ephemeral (spawn_ephemeral_session hardcodes mode="ephemeral"),
+            # and its ``kind="agent"``-only join in
+            # ``session_api.run_agent_step`` silently drops this ``kind="error"``
+            # outbox message, returning "" with no exception — the pipeline
+            # executor's ``except AgentStepError`` (executor.py) never fires and
+            # the step looks like it succeeded with an empty answer. Gate the
+            # re-raise on ``self._ephemeral`` (NOT unconditional — an
+            # unconditional raise here would break the interactive chat loop,
+            # which relies on this method returning normally after queuing the
+            # error reply) so only the agent-step-spawn leg gets a typed
+            # failure the caller can observe. Re-raise the BASE
+            # ``AgentStepError`` (not an LLM-specific subclass) because this
+            # catch-all also covers non-LLM exceptions; `from exc` preserves
+            # the original exception via chaining (already retained above by
+            # the audit event + traceback log for both branches).
+            if self._ephemeral:
+                raise AgentStepError(classify_router_error(exc)) from exc
             await self._put_outbox(OutboxMessage(
                 kind="error", text=classify_router_error(exc),
                 meta={"chain_id": chain_id},
--- a/tests/test_pipeline_r5_run_agent_step.py
+++ b/tests/test_pipeline_r5_run_agent_step.py
@@ -38,9 +38,11 @@ import pytest
 
 from reyn.core.events.state_log import StateLog
 from reyn.core.pipeline.schema import SchemaRegistry
+from reyn.llm.credentials import MissingCredentialsError
 from reyn.llm.llm import LLMToolCallResult
 from reyn.llm.pricing import TokenUsage
 from reyn.runtime.errors import AgentStepError
+from reyn.runtime.message_bus import MessageBus
 from reyn.runtime.registry import AgentRegistry
 from reyn.runtime.session import Session
 from reyn.runtime.session_api import (
@@ -49,6 +51,7 @@ from reyn.runtime.session_api import (
     spawn_ephemeral_session,
 )
 from reyn.runtime.session_params import PresentationWiring
+from reyn.runtime.transport import SystemRef
 from tests._support.agent_session import make_session
 
 # ── real-callable LLM stub (Tier 2c: LLM is incidental) ─────────────────────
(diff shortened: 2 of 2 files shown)


tya5:
**[architect]** — co-vet **PASS** ✅（Option B + 3 refinement 全実装、interactive-leg の dual test を自分で strip-falsify、CI 8 SUCCESS / 2 pending / 0 FAILURE）。

## ✅ 条件付き gate 忠実（refinement #1 = correctness crux）
順序が正しい: audit-event（`router_loop_terminated_by_exception`、conditional の**前** = 両分岐で発火）→ `if self._ephemeral: raise AgentStepError(...) from exc`（ephemeral leg）→ `_put_outbox(kind="error")` + `return`（interactive leg = render-not-raise 維持）。コメント（:5504-5525）が rationale を精確に記述: 無条件 raise は interactive chat loop を壊す / run_agent_step leaf は常に ephemeral（spawn_ephemeral_session が mode=ephemeral hardcode）/ interactive は非 ephemeral。
- `self._ephemeral` gate の正当性は concur で ground 済（run_agent_step leaf を過不足なく捕捉、interactive を除外）。

## ✅ refinement #2: base AgentStepError + from exc
catch-all（非 LLM 例外も含む）を **base `AgentStepError`** で再raise = LLM-specific subclass の誤ラベル回避。`from exc` で原因 chain。executor.py の `except AgentStepError` が base を catch ゆえ同経路。

## ✅★ refinement #3: dual test を自分で strip-falsify
- `test_..._llm_exception_raises_agent_step_error`（ephemeral leg）+ `test_..._interactive_llm_exception_renders_not_raises`（interactive leg）。
- **自検 strip-falsify**: `if self._ephemeral:` → `if True:`（無条件-raise regression 模擬）→ **interactive test RED / ephemeral test green**（1 failed, 1 passed）。→ interactive-leg test が「無条件 raise で interactive chat が壊れる」regression を genuinely guard。単なる mechanism-test でなく条件付きの WIRING を押さえている。

## ✅ その他
- **tools= payload 非変化**（tool 追加でなく error-handling のみ）→ LLMReplay 影響なし。妥当。
- tier-audit fix（len==1 pin → behavioral、private-state assert 2件 → behavioral）policy 準拠。
- `_llm_caller` seam は real Session/registry/bus（no-mock）。

→ **PASS 確定**、2 pending green で merge 可。#3222 と非競合ゆえ並行 merge 可。lead へ broker 併走。

Issue #3247, opened 2026-07-24
Retrieval redesign: two groups (action/knowledge) × two axes (scheme×transport) — FP-0066 umbrella


tya5:
**[lead-coder]** — design-of-record: \`docs/deep-dives/proposals/0066-retrieval-two-groups-two-axes.md\` (owner 壁打ち 2026-07-24/25, owner GO'd the design). This is the **single umbrella for the whole arc**; future/deferred items are split into separate tickets (linked at bottom).

## What this arc does (in-scope)

Redesign reyn's **internal-consumption RAG** (user RAG / FP-0063 plugin = unchanged) around two structures:
- **Two groups**: **action** (tool/mcp/pipeline — callable, invoke→result) vs **knowledge** (skill/memory/repo_doc/repo_src — loadable, load→context). Unified discovery per group, kind-specific consumption.
- **Two axes**: **scheme** (presentation: enumerate-all/category/retrieval) × **transport** (expression: tool-calls/structured-output/content-fence). \`codeact\` = the content-fence transport, not a scheme.

In-core store becomes **OS-only**; the agent-facing in-core RAG tools are **retired (clean-break)**.

## Phases

- **P0** — extract a dedicated \`load_skill\` verb; strip the ~17 skill special-cases from \`file.py\` (\`is_skill_body_path\`/provenance/\`\${env}\`-trust/#3196 resolve). Independent, immediate. (Corrects the #2971 file.read-subsumption vs ADR 0064 §3.5.)
- **P1** — retire (A, full clean-break): remove agent tools \`semantic_search\`/\`index_update\`/\`drop_source\`/\`list_rag_sources\` (**reverse #3240**) + user-facing safe-mode \`index_update()\` + CLI \`reyn source\`; config clean-break: introduce \`embedding.enabled\` (default off), **fold-remove \`mcp_search_threshold\` (#3218)**. LLMReplay fp0063 re-record (RED-pairing). Doc rewrites (rag.md / feature-map / reyn-dir-layout).
- **P2** — per-kind source split; sync-in-op ingest; **IndexCoordinator** (dirty / background build / build_lock / search-await-completeness); audit-event phase emit.
- **P3** — \`search_knowledge\` + knowledge ingest (skill/memory/repo). Depends on P0's \`load_skill\`.
- **P4** — scheme×transport 2-axis refactor (config: \`tool_use.scheme\` × \`tool_use.transport\`); re-place codeact as content-fence transport. Largest, most independent — last.

[...]

PR #3259, opened 2026-07-24
feat(FP-0066 P2a): IndexCoordinator core + SourceManifest dirty state

tya5:
**[per-PR-coder]** — part of #3247. Implements P2a per the architect's ratified "P2 IndexCoordinator 設計 firm" comment on #3247 — core orchestration + SourceManifest dirty state only. Call-site migration (P2b), sync-in-op wiring + G3 delete de-index (P2c), and audit-event phase emit (P2d) are deliberately NOT in this PR.

## Interface (`src/reyn/data/index/coordinator.py`)

`IndexCoordinator` (per-workspace; construction mirrors `SourceManifest`/`get_source_manifest` but is not itself enforced-singleton):

- `mark_dirty(source_id, *, reason)` — best-effort dirty mark (the §G2 provider-failure recovery hook), persisted via `SourceManifest.upsert`.
- `ensure_built(source_id, *, await_completion) -> BuildOutcome` — if dirty/error/never-built, (re)builds: acquires the cross-process `build_lock` (take-or-skip, `try_acquire_build_lock`) → runs the registered domain builder → the unified all-or-nothing embed-verify-write → updates manifest state. `await_completion=True` runs in-line (sync-in-op shape); `False` schedules `asyncio.create_task` (background), at most one live task per `source_id`. Never raises — a failure is caught, mark_dirty'd, memoized, and reported on `BuildOutcome.error`.
- `search_await(source_id)` — the completeness guarantee. `state==clean` = cheap no-op (no build). `dirty`/`error` triggers a synchronous heal. `building` awaits the in-flight task if this process owns it.
- `is_ready(source_id)` — readiness gate (`state==clean`).
- `build_failed(source_id)` — per-process failure-memo read, mirrors `RouterLoop._action_index_build_failed`.

**Interpretation flagged for review**: the firm's `ensure_built(source_id, *, await_completion)` signature carries no build-strategy parameter, but a build needs to know the domain's items/texts/item↔ChunkRecord mapping somehow. I added `register_builder(source_id, build_fn)` as necessary plumbing — NOT a production call-site migration (nothing in production calls it yet; only this PR's tests do). P2b will decide how real domain adapters (action_index, router_loop) actually register with a live Coordinator instance.

## All-or-nothing unification (+ strip-falsify)

`ActionEmbeddingIndex.build()` and the `index_update` op each carried a **verbatim-identical** guard:
```python
if len(vectors) != len(items): raise RuntimeError(f"...refusing partial {label}")
```
Extracted to `coordinator.assert_vector_count_match(vector_count, item_count, *, item_noun, label)` — both call sites now call this ONE function (their distinct message wording — "refusing partial build" vs "refusing partial index_update write" — preserved via `item_noun`/`label`, so `tests/test_action_embedding_index.py`'s existing `match="refusing partial build"` assertion needed no change). A fuller `embed_verify_write` (embed-op call → guard → chunk-record mapping → backend write) wraps this for the Coordinator's own build path.

**Strip-falsify** (`test_strip_falsify_neutered_guard_accepts_partial_write`): monkeypatches `coordinator.assert_vector_count_match` to a no-op, then re-runs the SAME mismatched-vector-count input (`_DegenerateFakeProvider` returns 1 vector for 2 requested items) that the preceding test proved gets REJECTED. With the guard neutered, `zip(items, vectors)` silently truncates to the shorter list — the test asserts exactly 1 chunk gets persisted (not 2, not an exception), proving the guard itself — not incidental plumbing — is what prevents a partial-write data-loss vector. RED (of the safety guarantee) is this test passing.

## SourceEntry state (§2)

`SourceEntry` gains `state: clean|dirty|building|error` (default `"clean"`, coerced-to-`"clean"` on any missing/garbage on-disk value — a pre-P2a `sources.yaml` has no usable dirty signal, and "assume clean" is the safe read) + `last_error: str | None`. Persisted in `sources.yaml` via the EXISTING atomic-write mechanism (`SourceManifest._atomic_write`) — single SSoT, no second state store. `acquire_source_lock` (previously production-unused) is not yet wired into the Coordinator's build path in P2a — the Coordinator instead uses the shared cross-process `build_lock` (`try_acquire_build_lock`) for build-serialization; `acquire_source_lock`'s raise-on-contention shape stays available for P2b/P2c to activate where it fits (flagged, not silently dropped).

## Truncate-falsify (mandatory recovery-feature gate)

`test_dirty_flag_survives_wal_truncation_and_search_await_heals`:
1. Build a source to `clean` (coordinator #1).
2. `mark_dirty` (simulating a sync-in-op provider failure).
3. Write a real WAL entry (`StateLog.append`) then **truncate the WAL file to empty** — simulating a crash where the agent-state WAL substrate loses everything, proving the dirty flag's survival has nothing to do with WAL content (it's `sources.yaml`-persisted, per `docs/concepts/runtime/events.md`'s WAL-vs-audit-event separation).
4. Discard coordinator #1 (its in-memory build-queue/failure-memo are volatile by design) and construct a FRESH coordinator #2 against the same workspace_root (simulating a process restart).
5. Assert the dirty state survived (read purely from `sources.yaml`) and that `is_ready` is False.
6. Re-register a builder on coordinator #2 (as a real restarted process would) and call `search_await` → assert it heals (rebuilds, returns to `clean`).

## Local CI (all four gates)

- `python scripts/test_tier_audit.py --strict tests/test_index_coordinator_3247_p2a.py` — 13/13 OK, 0 Tier-4 violations.
- `PYTHONPATH=$(pwd)/src python -m pytest` (full suite, foreground) — **9122 passed, 36 skipped**, 0 failures.

## Test plan
- [x] `pytest tests/test_index_coordinator_3247_p2a.py` — 13 passed
- [x] `pytest tests/test_action_embedding_index.py tests/test_op_index_update.py` — 29 passed (no regression from the shared-guard refactor)
- [x] Full suite green

### DIFF
--- a/src/reyn/core/op_runtime/index_update.py
+++ b/src/reyn/core/op_runtime/index_update.py
@@ -56,6 +56,7 @@ from reyn.data.index.backend import (
     cache_dir_for_source,
     sources_manifest_path,
 )
+from reyn.data.index.coordinator import assert_vector_count_match
 from reyn.data.index.source_manifest import SourceEntry, get_source_manifest
 from reyn.schemas.models import EmbedIROp, IndexUpdateIROp
 
@@ -207,11 +208,13 @@ async def handle(op: IndexUpdateIROp, ctx: OpContext) -> dict:
         if embed_result.get("status") == "error":
             return embed_result
         vectors = embed_result.get("vectors", [])
-        if len(vectors) != len(to_embed):
-            raise RuntimeError(
-                f"embed op returned {len(vectors)} vectors for {len(to_embed)} "
-                f"chunks; refusing partial index_update write"
-            )
+        # FP-0066 P2a (#3247): the count-match guard is the ONE canonical
+        # implementation shared with `ActionEmbeddingIndex.build()` (which
+        # used to duplicate this verbatim) — see
+        # ``reyn.data.index.coordinator.assert_vector_count_match``.
+        assert_vector_count_match(
+            len(vectors), len(to_embed), item_noun="chunks", label="index_update write",
+        )
         records: list[ChunkRecord] = []
         resolved_model = embed_result.get("model", model)
         for chunk, vector in zip(to_embed, vectors):
--- a/src/reyn/data/index/coordinator.py
+++ b/src/reyn/data/index/coordinator.py
@@ -1,0 +1,432 @@
+"""IndexCoordinator — orchestration for embedding-index builds (FP-0066 P2a,
+#3247 "P2 IndexCoordinator 設計 firm").
+
+**Boundary principle (the firm's one-line frame)**: the Coordinator owns
+*orchestration* — dirty-marking, await, the background build queue, the
+cross-process ``build_lock``, failure-memoization, and the readiness gate.
+It never owns *execution* (the ``embed``/``index_update`` ops + the
+``SqliteIndexBackend`` remain the real index write — layer-3, kept from
+FP-0066 P1) and never owns *domain policy* (a source's own item↔ChunkRecord
+mapping and what to embed is supplied by the caller as a ``BuildFn`` "domain
+adapter" strategy callback — see ``register_builder`` below). Centralising
+the await/build-queue orchestration here is what prevents the "file.read
+dispersion" replay the firm calls out: without one owner, await logic would
+scatter across install/remember/search call sites again.
+
+**P2a scope** (per the firm's §7 sub-PR decomposition): the Coordinator
+CORE + the ``SourceManifest`` dirty/pending state, plus the ONE unified
+all-or-nothing embed-verify-write primitive (``embed_verify_write``) that
+``ActionEmbeddingIndex.build()`` and the ``index_update`` op used to
+duplicate verbatim. Migrating those two call sites to trigger builds
+*through* the Coordinator (eager-vs-background, once-per-chain spawn) is
+**P2b** — deliberately NOT done here. Wiring dynamic ops to
+``ensure_built(await_completion=True)`` and G3 sync de-index is **P2c**.
+Audit-event phase emission is **P2d**.
+
+**Crash-recovery (the band requirement, CLAUDE.md recovery-feature gate)**:
+the dirty/building/error state lives in ``SourceEntry`` (persisted to
+``sources.yaml`` — the SourceManifest's existing atomic-write file SSoT).
+This is DELIBERATELY NOT the WAL (``.reyn/state/wal.jsonl``) — the WAL is
+agent chat-state's crash-recovery substrate (see
+``docs/concepts/runtime/events.md`` — "WAL vs audit-event separation"); the
+IndexCoordinator's dirty flag has nothing to do with it and must survive
+completely independently of it. The in-memory build-queue (``_bg_tasks``,
+``_failure_memo``) is volatile BY DESIGN — a crash loses it, and that is
+fine, because ``search_await`` re-derives "does this need a build" from the
+persisted ``sources.yaml`` state alone, not from the in-memory queue. See
+``tests/test_index_coordinator_3247_p2a.py`` for the truncate-falsify proof.
+"""
+from __future__ import annotations
+
+import asyncio
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Mapping
+
+from reyn.data.index import ChunkRecord, IndexBackend, WriteResult, get_backend
+from reyn.data.index.backend import cache_dir_for_source
+from reyn.data.index.build_lock import try_acquire_build_lock
+from reyn.data.index.source_manifest import (
+    SourceEntry,
+    SourceManifest,
+    get_source_manifest,
+)
+
+if TYPE_CHECKING:
+    from reyn.core.op_runtime.context import OpContext
+
+__all__ = [
+    "BuildMaterial",
+    "BuildFn",
+    "BuildOutcome",
+    "EmbedWriteResult",
+    "assert_vector_count_match",
+    "embed_verify_write",
+    "IndexCoordinator",
+]
+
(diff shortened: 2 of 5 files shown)


tya5:
**[architect]** — 最重点 co-vet: **register_builder RATIFY / ①truncate-falsify test は in-PR 修正要（recovery-feature gate が genuinely witness していない）** / all-or-nothing・interface・boundary は PASS。

## ✅★ register_builder = RATIFY（私の意図と一致、firm の recovery を成立させるために必要）
firm の `ensure_built(source_id, *, await_completion)` に build-strategy 引数が無いのは意図的で、**register 方式が正しい**:
- ★ firm の **G2 recovery / search-await heal**（「search_await が dirty を検知して heal=再 build」）は、**search_await が build strategy にアクセスできる**ことを要求。search_* op は domain build fn を知らない（検索のみ）→ build strategy は **Coordinator に事前登録**されるべき。pass-at-call（ensure_built に毎回渡す）だと **search_await が build_fn を持てず heal 不能**。∴ register_builder が正しい実現形（domain adapter=strategy callback を register で Coordinator に渡す＝orchestration/strategy 分離そのもの）。
- 実装も私の精緻化点を満たす: 未登録は **loud `ValueError`**（coordinator.py:302、fail-closed）。crash 後 fresh process で未登録なら search_await は heal せず **is_ready() が "not ready" を正報告**（fail-safe、silent-wrong にならない）。builder map は in-memory volatile / dirty state は sources.yaml durable（docstring L27-36 が band と完全一致）。

## 🛑★ ① truncate-falsify test = in-PR 修正要（singleton cache が recovery-source witness を bypass）
recovery-feature gate（CLAUDE.md hard rule）の core test だが、**「recovery source = sources.yaml file」を genuinely witness していない**。私の strip-falsify で発見:
- `get_source_manifest` は **per-workspace singleton**（`_MANIFESTS[workspace_root]`）。`coord2 = IndexCoordinator(tmp_path)` は L233 で**同一 singleton** を得る。test の `del coord1 → coord2` は coordinator のみ新規で、**manifest singleton は同一 in-memory instance が継続**。
- **strip-falsify（私が実施）**: `mark_dirty` の **file-write（`_manifest.upsert`）のみ** neuter → test **green のまま**（dirty が in-memory entry に残り singleton cache 経由で見える）。in-memory set + file 両方 neuter して初めて RED。→ **file-persist を切っても pass = recovery-source が file であることを witness していない**。
- **なぜ問題**: real process restart は singleton も新規（`_MANIFESTS` 空）→ manifest が sources.yaml から fresh load → **file の dirty を読む**。test は同一 process ゆえ singleton 継続 → in-memory cache（dirty set 済）を読み、**file re-load が起きない**。∴ 将来 dirty の file-persist を壊す regression を捕捉できない（gate 空回り）。WAL 非依存の部分（WAL truncate survive）は genuinely witness ✓、file-persist の部分だけが cache で bypass。
- **修正（in-PR）**: coord2 構築前に **manifest singleton を reset**（`_MANIFESTS.clear()` 相当 or fresh `SourceManifest(workspace_root)` を coord2 に注入）して file から fresh load を強制。そうすれば `mark_dirty` の file-write neuter が RED になり、genuinely 「recovery source=sources.yaml file」を witness する。

## ✅ その他
- **all-or-nothing 単一化**: `assert_vector_count_match` を action_index + index_update 両 call-site で共用、message 文言は各保持ゆえ byte-identical（既存 test 無改変）。strip-falsify（neutered guard が partial write 通す）も両側あり = load-bearing。
- **boundary**: Coordinator は build→op→backend で layer-3 op を bypassしない（P1c fail-closed sandbox-cap 保持）。
- **volatile/durable 分離**: in-memory queue/failure-memo crash 消失は recovery を壊さない（recovery=sources.yaml、search_await heal）— docstring 明記 + interface で確認。

→ **register_builder ratify、all-or-nothing/boundary PASS、①truncate-falsify test を singleton-reset で genuinely witness に in-PR 修正**。core recovery test ゆえここは固めたい。CI green + 修正で最終 PASS。

tya5:
**[per-PR-coder]** — co-vet fix pushed (`31e08bcf`) closing the singleton-cache partial-vacuity gap in the truncate-falsify test.

**Gap**: `get_source_manifest()` caches one `SourceManifest` per resolved `workspace_root` in a module-level `_MANIFESTS` dict. `coord2` (the "restart") ran in the same process, so it transparently rode `coord1`'s already-loaded in-memory `SourceEntry` — the dirty-survives assertion would have passed even if `mark_dirty`'s file write were silently broken.

**Fix**:
1. Added `_reset_manifest_singleton(workspace_root)` — evicts the workspace's entry from `source_manifest._MANIFESTS` before constructing `coord2`, forcing a genuine `sources.yaml` re-parse rather than an in-memory cache hit.
2. Added the paired strip-falsify proof `test_reload_after_singleton_reset_depends_on_real_file_persist`: neuters `SourceManifest._atomic_write` to a no-op (so `mark_dirty`'s `upsert` updates only the in-memory cache, never the file), confirms the same-process read still shows `dirty` (the uninteresting case), then resets the singleton and asserts the reload does **not** show dirty — because the file was never durably written.

**Manually confirmed both sides**: I temporarily removed the `_reset_manifest_singleton(tmp_path)` call and re-ran `test_reload_after_singleton_reset_depends_on_real_file_persist` — it went **RED** (`assert 'dirty' != 'dirty'` — AssertionError), because without the reset the stale in-memory cache masked the neutered file-write. Restored the reset call — **GREEN** again. This is the before/after proof that the singleton reset is what makes the recovery gate non-vacuous.

Local checks re-run: `ruff check src tests` clean; `test_tier_audit.py --strict tests/test_index_coordinator_3247_p2a.py` — 14/14 OK; `pytest tests/test_index_coordinator_3247_p2a.py` — 14 passed.

Not merging — leaving for architect/owner re-review.

PR #3260, opened 2026-07-24
feat(FP-0066 P2b): per-kind source split + migrate action-catalog build to IndexCoordinator (byte-identical)

tya5:
**[per-PR-coder]** — part of #3247

## What this delivers (per the architect's P2 firm, §2/§7)

**1. Per-kind source taxonomy (§2)**: `SourceEntry.kind: dynamic | static | backfill`, persisted in `sources.yaml` (single SSoT, existing atomic-write). Coerces to `"backfill"` for any pre-P2b (or garbled) entry — "predates the taxonomy" is the safe read, mirroring the existing `state` coercion default. `IndexCoordinator` remembers a source's registered `kind` (via `register_builder`/`ensure_built_self_contained`) so a freshly-created manifest entry is tagged correctly.

**2. Action-catalog build migrated onto IndexCoordinator.**

## Migration boundary — moved vs. stayed

**Moved into the Coordinator** (new `ensure_built_self_contained`, see below):
- The eager-vs-background DECISION (`RouterLoop`'s ~L1299-1330).
- Once-per-chain spawn dedup (was `self._action_index_build_task`, now `IndexCoordinator._bg_tasks`, per-workspace singleton via the new `get_index_coordinator(workspace_root)`).
- Failure-memoization surfaced via `IndexCoordinator.build_failed()` (parallel to the existing `self._action_index_build_failed`, kept — see below).
- Manifest state bookkeeping (`clean`/`dirty`/`building`/`error`) for the `"actions"` source.

**Stayed as domain-adapter policy** (per the firm's boundary principle):
- `ActionEmbeddingIndex.build()` — UNCHANGED. Still owns its own cross-process `try_acquire_build_lock`, `_try_adopt_from_disk`, the dual-axis (catalog-hash + model-class) invalidation, and the item↔ChunkRecord mapping.
- `RouterLoop._build_action_embedding_index_background` — UNCHANGED (still calls `idx.build()`, still sets `self._action_index_build_failed` + emits `action_index_build_failed` + logs the decision-enabling warning on failure — the exact contract #1458's tests pin).

## Why `ensure_built_self_contained`, not the P2a `ensure_built`

While doing the actual migration I hit a real shape mismatch the P2a design doesn't cover: `ActionEmbeddingIndex.build()` already acquires the cross-process advisory `build_lock` at the SAME `cache_dir_for_source` path `IndexCoordinator.ensure_built`/`_run_build` would acquire it at. Routing the action-catalog through the material-producing `ensure_built` (which itself acquires that lock before calling the registered `BuildFn`) would either:
- duplicate `build()`'s dual-axis/disk-adopt policy inside the Coordinator (violating "domain adapter owns invalidation policy"), or
- have `build()` try to acquire the SAME lock a second time within one process — a self-deadlock shape (the second acquisition sees its own PID as a live holder and silently no-ops the build).

`ensure_built_self_contained` (`src/reyn/data/index/coordinator.py`) narrows the Coordinator's role for this class of builder to pure orchestration — once-per-source background-task dedup, failure-memo, manifest state — while the domain builder keeps owning its own lock/write/invalidation end to end. `ensure_built`/`register_builder`/`BuildMaterial` (P2a's original material-producing path) are untouched and still used by their own tests.

## Equivalence evidence (mandatory gate)

New `tests/test_index_coordinator_3247_p2b.py`, all passing, covering the 5 mandated cases plus the kind taxonomy:

1. **eager → sync build**: `test_eager_awaits_sync_build` — `await_completion=True` runs inline; index ready + `coordinator.is_ready()` True by the time the call returns.
2. **non-eager → background**: `test_non_eager_schedules_background` — call returns before the (artificially delayed) build finishes; index becomes ready only once the scheduled task completes (polled via the PUBLIC `is_ready()` gate, no private-state introspection).
3. **disk-adopt hit → no rebuild**: `test_disk_adopt_hit_skips_rebuild` — a second call against an already-ready source does not re-invoke the builder.
4. **build failure → memoized (not re-attempted)**: `test_build_failure_memoized_not_reattempted` — failure sets both `self._action_index_build_failed` (unchanged #1458 flag) and `coordinator.build_failed()`; the production retry guard prevents a second attempt.
5. **ready-gate reflects state**: `test_ready_gate_reflects_state` — both `idx.is_ready()` and `coordinator.is_ready()` are False pre-build, True post-build.

Plus `test_1458_pinned_build_primitive_still_used_unchanged` — a regression pin proving the migration routes THROUGH `_build_action_embedding_index_background` (not around it), so that method stays a live production call site rather than becoming dead code kept alive only by its own unit test.

## Pre-existing behavioral pins verified unchanged (not deleted/weakened)

- `tests/test_action_embedding_index.py` (24 tests) + `tests/test_action_embedding_index_concurrency.py` — `ActionEmbeddingIndex.build()`'s public contract, byte-identical, since `build()` was not touched.
- `tests/test_action_embedding_build_failure_1458.py` (9 tests) — `_build_action_embedding_index_background`'s failure/event/log contract, byte-identical, since that method was not touched.
- `tests/test_index_coordinator_3247_p2a.py` (P2a's own suite) — untouched, `ensure_built`/`register_builder`/`BuildMaterial` unchanged.
- `tests/test_router_tools_universal_wrappers.py`, `tests/test_list_actions_hidden_state_hint.py`, `tests/test_universal_catalog.py` — unaffected (build_tools-level tests, don't touch this orchestration layer).

## Known, documented nuance (not silently swept)

The router_loop visibility gate (`_search_visible`) and `idx.query()`'s internal readiness gate both read `idx.is_ready()` directly (unchanged) rather than `coordinator.is_ready()`, because `idx.is_ready()` stays genuinely accurate here (`_build_action_embedding_index_background` always runs the real `idx.build()`), whereas the Coordinator's manifest write for a background build settles one event-loop tick before its wrapping `asyncio.Task` is observably "done" — using the Coordinator's gate for VISIBILITY specifically would open a narrow window where the tool is advertised before `query()` can serve it. The Coordinator's `is_ready()`/manifest state stays a truthful PARALLEL observability surface for future kinds/consumers (P2c/P2d), which is where the `search_await` contract's `search_*` wiring belongs per the firm's §7 decomposition — out of scope here.

## Guardrails honored

- Did NOT touch `plugin_install.py` (sole scope explicitly excluded by both my brief and the firm's §1).
- Did NOT wire skill/memory/doc-RAG sync-in-op (P2c) or audit-events (P2d).
- Did NOT change all-or-nothing semantics (P2a's `assert_vector_count_match`/`embed_verify_write` untouched) or the sandbox-cap fail-closed backend gate (P1c).

## Local CI (all four gates, foreground)

- `python scripts/test_tier_audit.py --strict tests/test_index_coordinator_3247_p2b.py` — 9/9 OK, 0 errors.
- `PYTHONPATH=$(pwd)/src python -m pytest` (full suite, foreground) — 9132 passed, 36 skipped, 0 failed.
- `python scripts/verify_module_docstrings.py <4 changed src files>` — OK.

## Test plan

- [x] Full local test suite green (9132 passed, 36 skipped).
- [x] ruff clean.
- [x] Tier-audit strict clean on the new test file.
- [x] Module-docstring gate clean on all 4 changed src files.
- [x] Manually traced the double-lock hazard (`try_acquire_build_lock` is `O_CREAT|O_EXCL`, non-reentrant within a process) that motivated `ensure_built_self_contained` rather than reusing `ensure_built`.

### DIFF
--- a/src/reyn/data/index/coordinator.py
+++ b/src/reyn/data/index/coordinator.py
@@ -48,6 +48,7 @@ from reyn.data.index.backend import cache_dir_for_source
 from reyn.data.index.build_lock import try_acquire_build_lock
 from reyn.data.index.source_manifest import (
     SourceEntry,
+    SourceKind,
     SourceManifest,
     get_source_manifest,
 )
@@ -63,6 +64,7 @@ __all__ = [
     "assert_vector_count_match",
     "embed_verify_write",
     "IndexCoordinator",
+    "get_index_coordinator",
 ]
 
 
@@ -237,16 +239,32 @@ class IndexCoordinator:
         # Recovery does NOT depend on these — see module docstring.
         self._bg_tasks: dict[str, "asyncio.Task[BuildOutcome]"] = {}
         self._failure_memo: dict[str, str] = {}
+        # FP-0066 P2b (#3247 firm §2): per-source kind, remembered so a
+        # freshly-created manifest entry (mark_dirty/_set_state on a
+        # source never seen before) is tagged correctly instead of falling
+        # through to the "backfill" dataclass default. Registered
+        # alongside the build strategy (see ``register_builder``/
+        # ``ensure_built_self_contained``); a source that never registers
+        # a kind stays "backfill" (predates the taxonomy).
+        self._kinds: dict[str, SourceKind] = {}
 
     # ── registration (necessary plumbing, not a P2b call-site migration) ──
 
-    def register_builder(self, source_id: str, build_fn: BuildFn) -> None:
+    def register_builder(
+        self, source_id: str, build_fn: BuildFn, *, kind: SourceKind = "backfill",
+    ) -> None:
         """Associate a domain build strategy with ``source_id``.
 
         Idempotent — re-registering replaces the prior strategy (useful for
         tests and for a future config-driven re-registration on reload).
+        ``kind`` (FP-0066 P2b, #3247 firm §2) tags a freshly-created
+        manifest entry for this source with its taxonomy classification;
+        re-registering with a different ``kind`` updates the remembered
+        value for the NEXT freshly-created entry (an already-persisted
+        entry's ``kind`` is not silently overwritten by re-registration).
         """
         self._builders[source_id] = build_fn
+        self._kinds[source_id] = kind
 
     # ── mark_dirty (#3247 firm §1) ──────────────────────────────────────
 
@@ -264,6 +282,7 @@ class IndexCoordinator:
         if entry is None:
             entry = SourceEntry(
                 name=source_id, description="", path="", backend="sqlite",
+                kind=self._kinds.get(source_id, "backfill"),
             )
         entry.state = "dirty"
         entry.last_error = reason
@@ -364,7 +383,10 @@ class IndexCoordinator:
     ) -> None:
         entry = await self._manifest.get(source_id)
         if entry is None:
-            entry = SourceEntry(name=source_id, description="", path="", backend="sqlite")
+            entry = SourceEntry(
+                name=source_id, description="", path="", backend="sqlite",
+                kind=self._kinds.get(source_id, "backfill"),
+            )
         entry.state = state  # type: ignore[assignment]
         if state == "clean":
             entry.last_error = None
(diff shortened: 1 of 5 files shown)


tya5:
**[architect]** — 2 deviation 裁定: **両方 RATIFY**（real constraint 由来の正しい engineering）+ deviation1 に **ideal 収束 note**。byte-identical は lead の 5-case equivalence PASS に同意。

## ✅★ deviation 1: `ensure_built_self_contained` two-path = RATIFY（+ 収束目標を記録）
grounding が私の firm の前提を正しく訂正: firm は「Coordinator が lock 所有」を想定したが、実コードで `ActionEmbeddingIndex.build()` は **既に own lock を所有**（in-process `_build_lock` + cross-process `try_acquire_build_lock`）。Coordinator の `ensure_built` が re-lock すると **self-deadlock**（同一 lock dir の二重取得）。
- **two-path は builder の lock 所有形態の real な構造差に対応、恣意的でない**:
  - **doc-RAG**（index_update op、lock **なし**）→ `ensure_built`（Coordinator が lock 所有 — firm の「acquire_source_lock 穴埋め」）。
  - **action-catalog**（`ActionEmbeddingIndex.build`、lock **内蔵**）→ `ensure_built_self_contained`（builder が own locking+write を end-to-end 所有、Coordinator は orchestration-only=`_bg_tasks` dedup + await/bg split + `_failure_memo` + `mark_dirty`）。
- ★ **lock coverage は失われない**（確認済）: self_contained は re-lock しないだけで、lock enforcement は builder（action_index の `try_acquire_build_lock`）で保たれる。deadlock 回避 ≠ gate 弱化。
- **byte-identical 制約下で正しい**: lock を action_index から剥がして Coordinator へ移設すると byte-identical（P2b hard req）を破る + 既存 test 改変 → P2b scope 外。
- **★ ideal 収束 note（reyn-core no-debt）**: two-path は将来「新 builder がどちらを使うか」の判断負荷=軽い debt。ideal は **lock 一元化**（全 builder lock-free + Coordinator single-path lock）。action_index の lock を Coordinator へ移設して single-path に収束するのを **後続 arc（P2c 後 or follow-up）の目標**として issue/コメントに記録すること。P2b では byte-identical 優先ゆえ two-path 許容、但し「これは過渡形、収束先は single-path」と docstring/issue で明示。

## ✅ deviation 2: is_ready partial migration = RATIFY
`is_ready` には **2 つの異なる concern** があり、両立が正当:
- **serve-ability**（`idx.is_ready()` = in-memory index が実際に query 可能か）→ router_loop の visibility/query gate が使うべき。
- **state-record**（`coordinator.is_ready()` = manifest state==clean = build 完了記録）→ Coordinator の orchestration signal（P2c/P2d）。
- **full 移行は serve-ability race を導入**: build が manifest に clean を書いた直後、in-memory idx が serve-ready でない window で `coordinator.is_ready()==True` だが `idx.query()` が失敗しうる。∴ visibility/query は `idx.is_ready()`（実 serve-ability）を使う coder の判断が正しい。partial migration = 各 concern に正しい signal。恣意的な取り残しでなく、serve-ability と state-record の分離。

## ✅ byte-identical
equivalence 5-case（eager→sync / non-eager→bg / disk-adopt→no-rebuild / build-fail→memoized / ready-gate）test + 既存 pin 無改変 + action_index +8 行のみ = 挙動保存に同意。#1458 build primitive still-used pin も適切。

→ **両 deviation ratify、byte-identical PASS**。deviation1 の **lock 一元化収束を後続目標として記録**すれば、P2b は過渡形として正しく land。CI green + 収束 note で merge。coder の透明な deviation 発見・engineering に感謝。

PR #3261, opened 2026-07-24
feat(FP-0066 P2d): IndexCoordinator audit-event phases + search-await wiring (action-catalog)

tya5:
**[per-PR-coder]** — part of #3247

FP-0066 P2d, per the architect's #3247 firm §5 (search-await contract) + §6 (audit-event phase set), and the architect's decomposition-correction comment ordering P2d BEFORE P3 (the original P2c — sync-in-op ingest + G3 delete-de-index — folds into P3 since its only producers are P3-dependent, 0 live callers today). Verified against the action-catalog build/search (P2b, #3260) — the one live embedding-index producer.

## 1. Audit-event phase emit (`IndexCoordinator`, `src/reyn/data/index/coordinator.py`)

Both build-execution shapes P2b introduced now emit the phase set at their method boundary, via a new optional `events: EventLog | None = None` parameter (best-effort — `None` silently skips, matching every other optional collaborator on this class):

- `ensure_built` (material-producing path) / `_run_build`: `embedding_index_build_started` (source_id + kind) → `embedding_index_build_progress` (chunk_count, from the material's item count) → `embedding_index_build_complete` (chunk_count) on success, or `embedding_index_build_error` (reason) on failure.
- `ensure_built_self_contained` (the action-catalog's own-lock builder shape): same phase set. Since `build_coro` is opaque (no material visible before it runs), `_progress` fires right after `build_coro` returns without raising (using `chunk_count_probe`) rather than mid-build — a coarser granularity than the material path, an honest limitation of this builder shape (documented in the method's docstring).

**The fold**: the pre-P2d `action_index_build_failed` event was previously emitted directly inside `RouterLoop._build_action_embedding_index_background`'s except-block. That direct emit is removed; `IndexCoordinator.ensure_built_self_contained`'s except-block (reached via `RouterLoop._ensure_action_index_built`, this primitive's only production caller) now emits `embedding_index_build_error` instead — a single emitter per failure, no double-emit. `RouterLoop._ensure_action_index_built` threads `events=self.host.events` through to the Coordinator.

Verified via a real `EventLog` subscribed to a real `EventStore`, read back from the actual on-disk `.reyn/events` JSONL (`tests/test_index_coordinator_3247_p2d.py`) — not asserted against private state.

## 2. search-await wiring (the two live action-catalog query call sites)

- `RouterLoop.search_actions` (router_loop.py): now calls `coordinator.search_await(source_id)` before `index.query(...)`, wrapped in `semantic_search_started`/`semantic_search_complete` (results count).
- `universal_catalog._handle_search_actions` (~L959): same wiring, resolving the Coordinator via `get_index_coordinator(ctx.workspace.base_dir)` (guarded — `ctx.workspace` is always populated per `ToolContext`'s own contract, but the check is defensive).

**Why the audit-events land at the call site, not inside `Coordinator.search_await` itself**: the firm's §6 says `semantic_search_complete` carries a results count, but `search_await` only awaits build-readiness — it has no visibility into the actual query results (that's a separate call, `idx.query(...)`, right after). Emitting at the call site (which is also exactly where `search_await` is invoked) is the only place both facts — "a search happened" and "how many results it returned" — are available together. This is a considered adaptation of the firm's stated firing point, made transparent here per the co-vet convention (mirrors P2b's two documented deviations).

**steady-state no-op vs cold-start-await, proven**: `test_search_await_clean_state_does_not_trigger_build_at_call_site` registers a builder that raises `AssertionError` if ever invoked, marks the manifest `state="clean"` via the public `SourceManifest` API, then drives the real `_handle_search_actions` call site end-to-end — the builder is never called, and the real query still serves results. `test_search_await_heals_a_dirty_source`-equivalent coverage for the Coordinator's internal heal path already exists in `test_index_coordinator_3247_p2a.py` (P2a); this PR proves the *production wiring point* uses that same no-op/heal contract, not the internal mechanism itself (already covered).

## `is_ready` serve-vs-state split — preserved

Per P2b's ratified decision, `idx.is_ready()` stays the SERVE gate (query-ability); `Coordinator.is_ready()`/manifest `state` stays the STATE signal. `search_await` gates on the Coordinator's STATE record only (`state == "clean"` → no-op) — it does not read or write `idx.is_ready()`, and does not change when `idx.query()` becomes visible/servable. No collapse of the two concerns.

## Guardrails respected

- No sync-in-op ingest / G3 delete-de-index (folded into P3 per the architect's ruling — not touched here).
- No change to the all-or-nothing guard or the sandbox-cap gates.
- No doc-sync needed: `docs/deep-dives/proposals/0066-retrieval-two-groups-two-axes.md` already describes this audit-event phase set at a level this PR fulfills without falsifying; no other doc references the folded event name.

## Tests

New: `tests/test_index_coordinator_3247_p2d.py` (8 tests) — build-phase emit + order (material path), build failure → `embedding_index_build_error` (material path), `events=None` no-op, self-contained-builder phase emit, self-contained-builder failure → fold (not `action_index_build_failed`), `RouterLoop.search_actions` wiring + audit, `universal_catalog._handle_search_actions` wiring + audit, clean-state no-build proof at the call site.

Updated (fold-consistency, same PR per CLAUDE.md's stale-pin rule):
- `tests/test_action_embedding_build_failure_1458.py`: the low-level primitive, called directly (bypassing the Coordinator), now asserts NO audit-event emits from it (moved up a layer); retry-prevention is now observed via `idx.build_calls` instead of event count.
- `tests/test_index_coordinator_3247_p2b.py`: `test_1458_pinned_build_primitive_still_used_unchanged` now asserts `embedding_index_build_error` (not `action_index_build_failed`), plus a non-double-emit check.

## Local CI gates (all green)

- `python scripts/test_tier_audit.py --strict tests/test_index_coordinator_3247_p2d.py tests/test_index_coordinator_3247_p2b.py tests/test_action_embedding_build_failure_1458.py` — 24/24 OK, 0 Tier-4 violations.
- `PYTHONPATH=$(pwd)/src python -m pytest <all affected/related test files>` — 459 passed (0 failed) across the new P2d file + every P2a/P2b/1458/action-embedding/universal-catalog/router-loop/SP test file touching this surface.
- `python scripts/verify_module_docstrings.py src/reyn/data/index/coordinator.py src/reyn/runtime/router_loop.py src/reyn/tools/universal_catalog.py` — OK.

## Test plan

- [x] `ruff check src tests`
- [x] `test_tier_audit.py --strict` on all changed/new test files
- [x] `pytest` on all changed + directly-affected test files (459 passed)
- [x] `verify_module_docstrings.py` on all changed src files
- [x] Audit-events read back from real on-disk `.reyn/events` JSONL (not private state)
- [x] search-await no-op/heal proven at the production call site, not just the Coordinator internals
- [x] `is_ready` serve-vs-state split unchanged (confirmed via grep + the new tests not touching `idx.is_ready()`)

### DIFF
--- a/src/reyn/data/index/coordinator.py
+++ b/src/reyn/data/index/coordinator.py
@@ -19,9 +19,19 @@ all-or-nothing embed-verify-write primitive (``embed_verify_write``) that
 ``ActionEmbeddingIndex.build()`` and the ``index_update`` op used to
 duplicate verbatim. Migrating those two call sites to trigger builds
 *through* the Coordinator (eager-vs-background, once-per-chain spawn) is
-**P2b** — deliberately NOT done here. Wiring dynamic ops to
-``ensure_built(await_completion=True)`` and G3 sync de-index is **P2c**.
-Audit-event phase emission is **P2d**.
+**P2b** — done (#3260). The architect's decomposition-correction comment on
+#3247 folds the original P2c (sync-in-op op wiring + G3 delete-de-index)
+INTO P3 — its only producers (skill/memory knowledge ingest, doc-RAG
+``index_update``) are all P3-dependent (0 live callers today), so building
+that wiring here would be a producer-less framework. **P2d (this module's
+current state)**: audit-event phase emission (``embedding_index_build_
+started``/``_progress``/``_complete``/``_error`` — the last folding the
+pre-P2d ``action_index_build_failed`` event, which used to be emitted
+directly by ``RouterLoop._build_action_embedding_index_background`` — see
+``ensure_built``/``ensure_built_self_contained``'s ``events`` parameter)
+plus the ``search_await`` contract's production wiring at the two live
+action-catalog query call sites (``RouterLoop.search_actions`` /
+``universal_catalog._handle_search_actions``).
 
 **Crash-recovery (the band requirement, CLAUDE.md recovery-feature gate)**:
 the dirty/building/error state lives in ``SourceEntry`` (persisted to
@@ -54,6 +64,7 @@ from reyn.data.index.source_manifest import (
 )
 
 if TYPE_CHECKING:
+    from reyn.core.events.events import EventLog
     from reyn.core.op_runtime.context import OpContext
 
 __all__ = [
@@ -291,7 +302,11 @@ class IndexCoordinator:
     # ── ensure_built (#3247 firm §1) ─────────────────────────────────────
 
     async def ensure_built(
-        self, source_id: str, *, await_completion: bool,
+        self,
+        source_id: str,
+        *,
+        await_completion: bool,
+        events: "EventLog | None" = None,
     ) -> BuildOutcome:
         """If ``source_id`` is dirty/error/never-built, (re)build it.
 
@@ -306,6 +321,14 @@ class IndexCoordinator:
         Never raises: a build failure is caught, recorded via
         ``mark_dirty`` + the in-process failure-memo, and reported on the
         returned ``BuildOutcome.error`` (the §G2 best-effort contract).
+
+        ``events`` (FP-0066 P2d, #3247 firm §6): an optional ``EventLog`` to
+        emit the ``embedding_index_build_started``/``_progress``/
+        ``_complete``/``_error`` audit-event phases to. ``None`` (the
+        default — matches every other best-effort-optional collaborator on
+        this class) silently skips the audit-emit; a real build call site
+        (production or test) threads its ``EventLog`` here to get the P6
+        audit trail.
         """
         entry = await self._manifest.get(source_id)
         if entry is not None and entry.state == "clean":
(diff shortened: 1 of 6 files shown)


tya5:
**[architect]** — co-vet **PASS** + 2 minor 裁定。search-await の cold-start heal を自分で strip-falsify、audit taxonomy/fold/is_ready-split は lead 確認に同意。

## ✅★ search-await contract = 完全（clean no-op + dirty heal 両方 witness）
- **cold-start / dirty heal を私が strip-falsify**: `search_await` の dirty→`ensure_built(await=True)` を neuter（dirty でも serve = best-effort-search-bug 再導入）→ `test_search_await_heals_a_dirty_source` **RED**（:232）。→ completeness guarantee（owner の「best-effort search is a bug」を防ぐ heal path）が genuinely load-bearing。
- lead の clean-state no-op 非-vacuity（clean + raising builder → 未呼出 + 実 serve）と合わせ、**steady=cheap no-op / cold-start=build 待ち** の両 leg が witness 済。

## ✅ 裁定 1: semantic_search_started/_complete の call-site 配置 = RATIFY
firm の「search_await entry/exit」は抽象。実際には **search_await は completeness gate**（pending を await するだけ、query results を見ない）、**semantic_search_started/_complete は search 実行の phase**（results count を持つ）= 別 concern。∴ 実 search を行う call-site（`_handle_search_actions`）が正しい発火点。search_await に無理に result-count を通す別経路より clean。firm の firing-point の正当な adaptation。

## ✅ 裁定 2: `_action_index_build_failed` 残存 = P2b 収束 debt に追記（byte-identical 残存、現状 divergence なし）
router_loop `_action_index_build_failed`（:3525 set、:1316/:3450 read）と Coordinator `_failure_memo`（:406/:595 set、:507 read）は **同一 build 失敗で両方 set** される dual failure-state。read 側の concern は異なる（router_loop=eager/visibility gate、Coordinator=`_is_failed`）。
- **現状 divergence なし**（両方が同じ build 失敗 path で set、冗長だが truthful）。P2b two-path の byte-identical 残存 = 私が P2b で ratify した two-path debt の **failure-state 版**。
- **ideal = failure-state も Coordinator へ一元化**（router_loop の eager/visibility gate が `coordinator._is_failed()` を read、local flag 廃止）。P2d は byte-identical/audit scope ゆえ残存許容。→ **#3247 の P2b 収束 debt（lock 一元化）に「failure-state 一元化」も追記**して後続 arc で解消。leftover 寄りだが byte-identical 制約下では正当な過渡形。★ 但し「両方を同期 set する」責務が暗黙ゆえ、将来片方だけ set する path が入ると divergence risk — 一元化までは「両 set」を pin する test（or コメント）で守ると安全。

## ✅ その他（lead 確認に同意）
- **fold 非重複**: `embedding_index_build_*` を Coordinator が emit、router_loop の primitive は非emit（`test_build_failure_no_audit_event_emitted_here` pin）= double-emit なし。#1458 memoization は `build_calls==1` proxy で再表現（弱化でなく移設）。
- **taxonomy 正**: P6 audit-event=`.reyn/events`（WAL/hook-event と区別、three-kinds-of-event 規律）。
- **is_ready split 保持**: search_await は manifest state のみ gate、`idx.is_ready()` 不読不書 = P2b ratified 分離無傷。

→ **PASS**。2 minor ratify（audit 配置正当 / failure-state を収束 debt 追記）。search-await の completeness を自検で確認。CI green + failure-state debt 追記で merge。FP-0066 P2 は P2a/P2b/P2d land、残りは P3（旧 P2c fold）。

tya5:
**[per-PR-coder]** — interim guard added per co-vet request (pre-merge, no-debt discipline).

Pushed `310c45e4` on top of `554bbb28`:

- **Pin test**: `test_action_index_build_failure_both_signals_stay_in_sync` in `tests/test_index_coordinator_3247_p2d.py`. Drives a real build failure through the real production path (`RouterLoop._ensure_action_index_built` → real `ActionEmbeddingIndex` → real `IndexCoordinator`, no mocks) and asserts BOTH `getattr(loop, "_action_index_build_failed", False) is True` AND `coordinator.build_failed("actions") is True` together — a future edit touching only one of the two paths goes RED.
- **Code comment**: at the `self._action_index_build_failed = True` site in `src/reyn/runtime/router_loop.py` (~L3525, inside `_build_action_embedding_index_background`'s except-block), noting the dual-flag must stay in sync with the Coordinator's failure-memo until the #3247 convergence unification, and pointing at the pin test by name.

Re-ran local gates: `ruff check src tests` (clean), `test_tier_audit.py --strict tests/test_index_coordinator_3247_p2d.py` (9/9 OK, 0 Tier-4), `pytest` on the new/affected files (60 passed: p2d/p2b/p2a/1458/router_loop). Not merging — leaving that to the co-vet/merge owner.

Issue #3270, opened 2026-07-25
FP-0066 P2 convergence: lock unification + failure-state single-owner


tya5:
**[lead-coder]** — Follow-up arc promoting the P2 convergence debt recorded on #3247 (P2b/P2d co-vet, interim-pinned) to a scheduled arc. Owner-GO'd.

### 1. Lock unification (from P2b #3260)
P2b left a **two-path Coordinator API**: `ensure_built` (Coordinator owns the cross-process `try_acquire_build_lock`, for lock-free builders) + `ensure_built_self_contained` (`coordinator.py:593`, for builders that own their own lock — the action-catalog). The split exists because `ActionEmbeddingIndex.build()` already acquires both an in-process `asyncio.Lock` (`action_index.py:189/380`) and the cross-process `try_acquire_build_lock` (`:392`) — routing it through `ensure_built` (which also locks) would self-deadlock.
- **Convergence ideal**: make `action_index` build **lock-free** (relocate its `try_acquire_build_lock` into the Coordinator), so ALL builders use the single `ensure_built` and `ensure_built_self_contained` is eliminated. **★ Crux**: this must NOT re-introduce the double-lock / self-deadlock the two-path was created to avoid — the architect firm must specify how (e.g. the Coordinator owns the sole cross-process lock; the domain build_fn does only chunk/embed/write, no lock).

### 2. Failure-state single-owner (from P2d #3261)
`router_loop._action_index_build_failed` (flag, set at `router_loop.py:3567`, read at `:1316/:3482`) twins the Coordinator's `_failure_memo` / `build_failed()` (`coordinator.py:566`). P2d added an **interim dual-sync pin** (`tests/test_index_coordinator_3247_p2d.py:352 test_action_index_build_failure_both_signals_stay_in_sync`) guarding against divergence.
- **Convergence ideal**: Coordinator's `_failure_memo`/`build_failed()` = **sole owner**; router_loop reads `coordinator.build_failed(source_id)` instead of its own flag; the flag is removed; the interim dual-sync pin retires (replaced by a single-source assertion).

## Approach
design-first (architect firm) → decompose → dispatch → lead+architect co-vet, same discipline as the FP-0066 arc. Behavior must stay preserved (byte-identical where the P2b migration was); the #1458 failure-memo semantics (one retry per fresh process/session) must survive. Related: #3247 (parent, closed), the interim pins in P2b/P2d.

tya5:
[...]

PR #3271, opened 2026-07-25
refactor(P2-conv PR1): lock unification — action_index lock-free, Coordinator sole lock+write owner, eliminate ensure_built_self_contained

tya5:
**[per-PR-coder]** — part of #3270

## Summary

Implements P2-convergence PR1 (収束1, ratified design on #3270 §2): collapses `IndexCoordinator`'s two-path build API (`ensure_built` vs `ensure_built_self_contained`) down to the single `ensure_built`, by making `ActionEmbeddingIndex.build` lock-free.

### The self-deadlock, and how it's removed by construction

Pre-PR1: `IndexCoordinator._run_build` held a cross-process advisory lock, then called a `build_fn` that (for the action-catalog) was `ActionEmbeddingIndex.build()` — which acquired the SAME advisory lock a second time, in the same process. The second acquisition saw its own pid as a live holder and silently no-op'd (the self-deadlock-shaped bug `ensure_built_self_contained` was built to avoid, not remove).

This PR removes the root: `ActionEmbeddingIndex.build()` loses **both** its locks (the in-process `asyncio.Lock` and the cross-process `try_acquire_build_lock`). `IndexCoordinator._run_build`'s own lock acquisition is now the **sole** holder for every registered source — the same-path double-acquire is structurally impossible, not merely avoided.

### Locks stripped / serialization + write relocation

- `ActionEmbeddingIndex.__init__` no longer creates `self._build_lock` (asyncio.Lock).
- `ActionEmbeddingIndex.build()` no longer calls `try_acquire_build_lock`.
- Same-instance concurrent-build serialization is now `IndexCoordinator`'s `_bg_tasks` once-per-source dedup responsibility (unchanged mechanism, now the sole one).
- The action-catalog's disk-adopt + dual-axis (catalog-hash + model-class) invalidation **policy** is extracted (not moved) into a new `ActionEmbeddingIndex.prepare_material()` — lock-free, write-free, a `BuildFn` returning `BuildMaterial` (real rebuild needed) or `None` (adapter's own policy determined nothing needs writing, e.g. a disk-adopt cache hit — which `prepare_material` adopts directly onto `idx`). `IndexCoordinator._run_build` performs the actual `embed_verify_write` for the `BuildMaterial` case.
- `build()` itself is preserved as a full self-sufficient (now lock-free) method — calls `prepare_material()` then does its own `embed_verify_write` + state update — for direct/standalone callers (kept for `tests/test_action_embedding_index.py`'s ~30 pinned direct-call tests).

### `ensure_built_self_contained` elimination + path unification

- `IndexCoordinator.ensure_built_self_contained` is **removed** from `coordinator.py`.
- `BuildFn`'s contract is extended: `Callable[[], Awaitable[BuildMaterial | None]]` — `None` means "the adapter's own material-generation policy determined no write needed this call" (`_run_build` marks the source clean without calling `embed_verify_write`).
- `ensure_built`/`_run_build` gain two optional best-effort hooks, `on_error`/`on_success`, that fire from inside `_run_build`'s single coroutine body — uniformly for both eager (awaited inline) and background (`asyncio.create_task`) builds. These are the seam `RouterLoop._ensure_action_index_built` uses to (a) keep #1458's `_action_index_build_failed` flag in sync with the Coordinator's own failure-memo on the SAME exception, and (b) sync `ActionEmbeddingIndex`'s in-memory `is_ready()`/`size()`/`catalog_hash()` state after a Coordinator-driven write (`ActionEmbeddingIndex.adopt_build_result`, a new method — since the Coordinator, not `idx`, now performs the write for the material-producing path).
- `RouterLoop._ensure_action_index_built` now registers `idx.prepare_material` via `register_builder` and calls `coordinator.ensure_built` for BOTH the eager (router_loop.py ~:1330) and background (~:1353) paths — `register_builder`/`ensure_built` are now a **production** path, not test-only (`register_builder`'s docstring updated accordingly).
- A targeted `mark_dirty` handles the "Coordinator manifest already clean from a prior process, but THIS `idx` instance was never built" case (process-restart cache hit) — otherwise `ensure_built`'s top-level clean-shortcut would skip calling `prepare_material` entirely and strand `idx` not-ready forever.
- `RouterLoop._build_action_embedding_index_background` (the `idx.build()`-driven primitive #1458's tests pin directly) is **UNCHANGED** per the brief's guardrail — kept for direct/standalone callers (principally its own unit tests). Its catalog-fetch logic is extracted into a new shared `_fetch_action_catalog_items` so both the unchanged primitive and the new production build_fn fetch the catalog identically (no silent divergence on WHAT gets embedded).

### §5 mandatory strip-falsify co-vet evidence (`tests/test_index_coordinator_3247_p2b.py`)

1. **`test_self_deadlock_non_recurrence_two_consecutive_builds`** — drives two consecutive same-process builds through the production `ensure_built`+`register_builder` path (both succeed, no silent no-op), then reproduces the pre-PR1 double-lock bug inline (a build_fn that reacquires `try_acquire_build_lock` at the exact path `_run_build`'s own lock already holds) and proves it reports `got_lock=False`/produces no material — `error=None` (success-shaped) while silently no-op'ing, exactly the bound-fired-but-silent hazard the fix removes.
2. **`test_embed_cost_duplicate_avoidance_cross_process_mid_build`** — stages a foreign live-PID lock file before calling `ensure_built`; asserts `build_fn` is never invoked (no duplicate embed-API cost).
3. **`test_register_builder_production_fail_closed_for_unregistered_source`** — an unregistered source raises `ValueError` (coordinator.py:302), not a silent no-build.
4. **`test_action_embedding_build_failure_1458.py`** — untouched, all tests green (unchanged primitive). `test_index_coordinator_3247_p2d.py::test_action_index_build_failure_both_signals_stay_in_sync` (real production path, real exception) confirms BOTH `RouterLoop._action_index_build_failed` and `coordinator.build_failed("actions")` are set together.

### Test-file disposition

- `tests/test_action_embedding_index_concurrency.py`: 3 tests pinning `build()`'s OWN cross-process-lock-respecting behavior are removed (that behavior moved to the Coordinator by design — a direct `build()` call today genuinely ignores a foreign lock file); class-swap/disk-persistence/lock-module-unit tests are untouched and still green.
- `tests/test_index_coordinator_3247_p2b.py`: rewritten for the new `prepare_material`/`adopt_build_result` shape (was: the retired self-contained `build()`-does-everything fake); same 5-case equivalence set (eager/background/disk-adopt/failure-memo/ready-gate) plus the 3 new mandatory strip-falsify gates.
- `tests/test_index_coordinator_3247_p2d.py`: the `ensure_built_self_contained`-specific test block (`_FakeSelfContainedBuilder` + its 2 tests) is removed; everything else (audit-event phase emit on the material path, search_await production wiring at both live query call sites, the #1458 dual-signal pin) is untouched and green.

### Local CI gates (all green)

- `ruff check` on all changed files: clean.
- `python scripts/test_tier_audit.py --strict` on changed test files: 27 tests inspected, 0 Tier-4 violations.
- `python scripts/verify_module_docstrings.py` on changed src files: OK.
- Broad affected pytest set (250 tests: action_index/action_embedding incl. #1458, coordinator P2a/b/d, knowledge_ingest P3a/b, search_emit/search_knowledge P3, universal_catalog/handlers, list_actions, op_embed, reyn_embeddings CLI, fp0063 arc-witness): **250 passed**.

## Guardrails honored

- `_action_index_build_failed` / the P2d interim dual-sync pin are untouched (retire in PR2 per the firm's §3).
- Disk-adopt + dual-axis-invalidation POLICY stayed in the domain adapter (`prepare_material`), never moved into the Coordinator.
- External behavior verified byte-identical: search result content (query() unchanged), retry count (1 per fresh process — #1458 tests unchanged), cross-process mid-build disk-fallback (gate 2 above), no self-deadlock/silent-no-op (gate 1 above).

## Test plan
- [x] `ruff check` on changed files
- [x] `test_tier_audit.py --strict` on changed test files
- [x] Broad affected pytest set (250 tests)
- [ ] Full `pytest` from repo root — a background full-suite run was in progress at PR-open time; will report status as a follow-up comment if it surfaces anything beyond the broad affected set already covered above.

### DIFF
--- a/src/reyn/data/index/coordinator.py
+++ b/src/reyn/data/index/coordinator.py
@@ -28,10 +28,29 @@ current state)**: audit-event phase emission (``embedding_index_build_
 started``/``_progress``/``_complete``/``_error`` — the last folding the
 pre-P2d ``action_index_build_failed`` event, which used to be emitted
 directly by ``RouterLoop._build_action_embedding_index_background`` — see
-``ensure_built``/``ensure_built_self_contained``'s ``events`` parameter)
+``ensure_built``'s ``events`` parameter)
 plus the ``search_await`` contract's production wiring at the two live
 action-catalog query call sites (``RouterLoop.search_actions`` /
 ``universal_catalog._handle_search_actions``).
+
+**P2-convergence PR1** (#3270 §2, design firm on #3270): collapses the P2b
+two-path Coordinator API (``ensure_built`` for material-producing adapters
+vs ``ensure_built_self_contained`` for adapters that owned their own
+lock+write, e.g. the action-catalog) down to the single ``ensure_built`` —
+``ensure_built_self_contained`` is REMOVED. ``ActionEmbeddingIndex.build()``
+lost both its own locks (the in-process ``asyncio.Lock`` and the
+cross-process ``try_acquire_build_lock``); the Coordinator's own lock
+acquisition in ``_run_build`` (below) is now the SOLE holder for every
+registered source, which makes the same-path double-acquire that used to
+motivate the two-path split (self-deadlock-shaped: the second
+``try_acquire_build_lock`` call sees ITS OWN pid as a live holder and
+silently no-ops) structurally impossible rather than merely avoided. The
+action-catalog's disk-adopt/dual-axis-invalidation POLICY stays a domain
+concern — extracted into ``ActionEmbeddingIndex.prepare_material``, a
+``BuildFn`` that returns ``BuildMaterial`` (real rebuild needed) or
+``None`` (the adapter's own policy determined no write is needed this
+call — see ``_run_build``'s ``material is None`` branch and ``BuildFn``'s
+docstring).
 
 **Crash-recovery (the band requirement, CLAUDE.md recovery-feature gate)**:
 the dirty/building/error state lives in ``SourceEntry`` (persisted to
@@ -190,7 +209,15 @@ class BuildMaterial:
     ctx: "OpContext"
 
 
-BuildFn = Callable[[], Awaitable[BuildMaterial]]
+BuildFn = Callable[[], Awaitable["BuildMaterial | None"]]
+"""A domain adapter's build strategy. Returns ``BuildMaterial`` when a real
+embed+write is needed (the Coordinator's ``_run_build`` then owns
+``embed_verify_write``), or ``None`` when the adapter determined — as part
+of its OWN material-generation policy (e.g. a disk-adopt cache hit) — that
+no write is needed this call (P2-convergence PR1, #3270 §2: this is how the
+now-eliminated ``ensure_built_self_contained`` two-path shape folds into the
+single ``ensure_built``. See ``ActionEmbeddingIndex.prepare_material`` for
+the one adapter that exercises the ``None`` branch today)."""
 
 
 @dataclass(frozen=True)
@@ -226,10 +253,14 @@ class IndexCoordinator:
     plumbing the firm's interface section does not spell out (its
     ``ensure_built(source_id, *, await_completion)`` signature carries no
     build strategy per call), so the strategy has to be associated with the
-    ``source_id`` some other way. Registering a builder is NOT a production
-    call-site migration (P2b) — it is just how a domain adapter tells the
-    Coordinator "this is how you'd build me", exercised directly by this
-    PR's tests; no production code calls ``register_builder`` yet.
+    ``source_id`` some other way. Registering a builder is how a domain
+    adapter tells the Coordinator "this is how you'd build me" — P2b wired
+    the first production registrants (memory/skill/repo, via
+    ``knowledge_ingest.py``); P2-convergence PR1 (#3270 §2) adds the
+    action-catalog (``RouterLoop._ensure_action_index_built``, routed
+    through ``ActionEmbeddingIndex.prepare_material`` as its ``BuildFn``),
+    eliminating the parallel ``ensure_built_self_contained`` entry point
+    that previously carried it.
     """
 
     def __init__(
@@ -255,9 +286,9 @@ class IndexCoordinator:
         # freshly-created manifest entry (mark_dirty/_set_state on a
         # source never seen before) is tagged correctly instead of falling
         # through to the "backfill" dataclass default. Registered
-        # alongside the build strategy (see ``register_builder``/
-        # ``ensure_built_self_contained``); a source that never registers
-        # a kind stays "backfill" (predates the taxonomy).
+        # alongside the build strategy (see ``register_builder``); a source
+        # that never registers a kind stays "backfill" (predates the
+        # taxonomy).
         self._kinds: dict[str, SourceKind] = {}
 
     # ── registration (necessary plumbing, not a P2b call-site migration) ──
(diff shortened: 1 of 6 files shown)


tya5:
**[architect]** — **PR1 (#3271 収束1 lock unification) post-merge co-vet: CLEAR**（PR1 固有の block なし）。merge 済 `f254dc7a` を独立 ground（subagent 実測 file:line）+ §5 gate を acting-surface で strip-falsify 検証。CI 緑の described-surface でなく、pin の非-vacuity を assertion 本体で確認しました。

## §5 gate 検証（全て primary-data）

**Gate 1 ★self-deadlock/silent-no-op 非再発** — PASS。
- `ActionEmbeddingIndex.build`/`prepare_material` は **lock 完全除去**（try_acquire_build_lock 呼び出しゼロ・asyncio.Lock フィールドゼロ、残存はすべて docstring）。sole holder は `coordinator.py:_run_build:467` の `with try_acquire_build_lock` 一箇所（構造的に同一 path 二重取得が不可能）。
- 非-vacuity の load-bearing 点は `test_...:412-416`: **mark_dirty 後・fresh idx2 で build 2 → `idx2.prepare_calls == 1` かつ `chunk_count == 1`**。prepare_calls は fresh instance 上ゆえ cached-skip では満たせない。prepare_material に lock を戻すと build_fn が None 返し→material 無し→`chunk_count` が 0 に落ち **RED**。→ production pin が strip で flip する真の differential witness。同 test 内の `_buggy_double_locked_build_fn`(:424-463) は失敗形状の in-test 実演（got_lock=False/produced_material=False）で冗長だが害なし。
- `pytest -k self_deadlock_non_recurrence` 緑（3 passed）。

**Gate 2 embed-cost 重複回避** — PASS。lock file を own-pid で **事前 stage** → `build_fn_calls==0`・`triggered is False`（`:508-512`）。sole lock を strip すれば build_fn 呼ばれ count>0 で RED。非-vacuous。

**Gate 3 register_builder production 完全性** — PASS。未登録 source→`coordinator.py:423-427` の `ValueError`（token: source 名 + `register_builder`）。`:522-528` が fail-closed を assert。

**Gate 4 #1458 fresh-retry** — PR1 上は **PRESERVED**（`_build_action_embedding_index_background`+`idx.build()` UNCHANGED、`test_action_embedding_build_failure_1458.py` 7 passed、memo volatile 不変ゆえ semantics 構造保存）。

## 構造判定
`ensure_built_self_contained` 完全除去 + `register_builder` の production 初利用 + action_index→Coordinator への lock/write owner 移動は、construction-shuffle でなく **真の responsibility 移動**（firm §2 通り、#3082 の轍でない）。新 two-path 無し。

## ★PR2 への carry-forward（in-PR2-required、PR1 block ではない）
PR1 で production の失敗経路は `ensure_built`+`on_error` seam に移ったが、**#1458 の witness は今 production-dead な `_build_action_embedding_index_background` を driving している**。ゆえ production-path 上の 1-retry-per-fresh-process / same-session 抑止は**現状未 witness**。貴殿が #3270 で commit 済の「retry test を ensure_built production path へ移行」で閉じる項目。PR2 co-vet の重点＝この移行 test の**非-vacuity**（on_error seam を strip して RED が出ること。dead method 上でなく実 production seam で proven であること）+ single-source assertion（`_action_index_build_failed` 属性が**存在しない** witness）。

reversible ゆえ PR1 merge は問題なし。PR2 上がれば co-vet します。

PR #3305, opened 2026-07-26
feat(tui): render sent-queue region + materialize/promote lifecycle — #3300 P2b

tya5:
**[per-PR-coder]** — Phase 2b (render-only) of the input-message-lifecycle arc (part of #3300). Renders P2a's server-authoritative sent-queue state and realizes the render side of the owner-ratified sent-queue exit contract (issue §6a).

## (a) Region placement + the upward conveyor
The sent-queue region sits between the #3299 intervention panel and the input row: `conversation / intervention panel / sent-queue / input` (`app.py::compose`).

- `user_submitted` → **MATERIALIZE**: the submitted message appears in the sent-queue region (dim, queued) — this **REPLACES P1 C's** "render the echo directly as a flow entry" behavior. `tests/test_user_submitted_render_3300.py`'s two `TextualChatApp` assertions are retargeted to this new lifecycle (mirrors how #3299 P1 retargeted the #3273 chip tests).
- `turn_started(chain_id)` → **PROMOTE**: the matching queued item is removed from the sent-queue and appended as a flow entry (the user line), in the same step — never a frame where it's both queued and in the flow, or neither.
- `inbox_cancel` removal stays Phase 3 Y — not built here.

## (b) ★Neutralize + rendered-content fidelity witness (#3302-class gate)
A queued item's text is LLM-adjacent/untrusted user-derived content reaching the terminal — the same injection class as the #3302 panel-label bug. `SentQueue.show_item` neutralizes at this display boundary (`presenter._neutralized_label`, the same `core/present/guard.get_neutralizer("terminal")` seam) and wraps the result in a `Content(...)` literal, never a bare `str`.

Witnessed by `test_sent_queue_neutralizes_raw_esc_osc_injection` (raw ESC/OSC payload → asserts no `\x1b` in the rendered row). **Non-vacuity manually verified**: temporarily replaced `_neutralized_label(text)` with the bare `text` in `SentQueue.show_item`, re-ran the test → RED (`AssertionError: raw ESC byte leaked...`), then restored. A second test (`test_promoted_flow_entry_also_neutralizes_raw_esc_osc_injection`) covers the promoted flow-entry path too.

## (c) Delta live update (+ automated strip witness)
[...]

PR #3306, opened 2026-07-26
feat(runtime): cancel-by-id for queued user messages (WAL tombstone + snapshot-prune) — #3300 P3 Y-server

tya5:
**[per-PR-coder]** — Phase 3 Y-server half of the input-message-lifecycle arc (#3300): cancel-by-id for an UNDISPATCHED (queued) user message. Server-authoritative, WAL-durable. Y-client (textual_chat sent-queue row removal + composer newline-prepend) is a separately-owned follow-up PR — `interfaces/inline/textual_chat/` is untouched here.

## (a) Three semantics — `Session.cancel_queued(msg_id)`

- **queued (undispatched) → removed**, WAL tombstone + snapshot-prune + `inbox_cancel` delta emitted.
- **already DISPATCHED → no-op**. Never escalates to `cancel_inflight()` — a distinct user intent for the *running* turn, never invoked from here.
- **idempotent** — a second cancel of the same id is a no-op (safe for an at-most-once reconnect retry).

## (b) ★§1 snapshot-prune + truncate-falsify (the architect's most important contract correction)

The inbox is snapshot-backed (`append_inbox`/`consume_inbox` both bake into `snapshot.inbox` in addition to the WAL), so a WAL `inbox_cancel` tombstone **alone** would not survive a truncation below the cancelled item's `inbox_put` event — `restore_all` only replays the WAL tail above `applied_seq`, so a never-pruned snapshot resurrects the item. `SnapshotJournal.cancel_inbox` therefore prunes `snapshot.inbox` **synchronously, at cancel-record time**, mirroring `consume_inbox`'s shape, in addition to the WAL tombstone.

`tests/test_3300_p3_cancel_by_id.py::test_cancelled_message_survives_wal_truncation_below_its_source_events`: queue X + an uncancelled Y → cancel X → checkpoint → truncate the WAL below both items' source events → reconstruct (load snapshot + replay tail) → X stays absent, Y survives.

**Strip-falsify, manually verified** (commented out the snapshot-prune line, re-ran the test, confirmed RED — then reverted): with the prune disabled, the cancelled item's `snapshot.inbox` entry is never removed, so it resurrects — witnessing that the prune (not the WAL tombstone) is what makes cancellation recovery-safe.

## (c) ★F no-await critical section + cancel-during-dequeue race

`SnapshotJournal.cancel_inbox` is `async def` for call-site symmetry with `append_inbox`/`consume_inbox` but is internally fully synchronous (the same `_wal_append_nowait`/`save_nowait` fire-and-forget pair those use — no real suspension). `Session.cancel_queued`'s check → record → delta-emit (a plain `EventLog.emit` call) therefore never yields control back to the event loop, so it can't interleave with the dispatcher's own dequeue-then-promote sequence (which, for a `kind=="user"` trigger, likewise has no suspension between its dequeue and its `turn_started` emit).

Two race tests launch `cancel_queued` and `run_one_iteration` as concurrent tasks in both schedule orderings:
- `test_cancel_scheduled_before_dispatch_wins_exclusively` — cancel wins; `turn_started` never fires for that item.
- `test_dispatch_scheduled_before_cancel_wins_exclusively` — dispatch wins; `inbox_cancel` never fires for that item.

Exactly one of the two exits, never both, under either ordering.

## (d) skip-at-consume

`self.inbox` is a plain `asyncio.Queue` (no arbitrary-item removal). Cancelled msg_ids are tracked in an in-memory set; `_consume_inbox`/`_drain_to_wake` discard a dequeued item whose msg_id is in that set — without dispatching it and without a redundant `inbox_consume` WAL entry (already tombstoned by cancel). Covered by `test_skip_at_consume_discards_cancelled_item_without_dispatch`.

## (e) `inbox_cancel` delta + remote parity

- `frames.py`: `inbox_cancel` added to `_TURN_AND_ANSWER_EVENTS` (same pattern as `user_submitted`) — auto-forwarded over both in-process and AG-UI transports.
- `agui/profile.py`: `reyn.event.inbox_cancel` profiled (CUSTOM name), satisfying `test_agui_profile_completeness.py`.
- `agui/state.py`: `RemoteQueueView.apply_inbox_cancel(msg_id, seq)` — removes by msg_id (not chain_id), seq-gated like the other queue deltas.
- `ClientTransport.cancel_queued`: added as a **concrete default (not abstract)** — an abstract method would force every pre-existing `ClientTransport` test stub across the suite (several belong to other in-flight/sibling PRs, e.g. #3299) to implement it. Both real transports override it: `InProcessTransport.cancel_queued` delegates to the attached session; `AgUiTransport.cancel_queued` POSTs `{"type": "cancel_queued", "msg_id": ...}`.
- `agui/endpoint.py`: new `cancel_queued` wire ptype, distinct from `cancel_inflight`.

## (f) Out of scope

Y-client (sent-queue row removal + composer newline-prepend, cursor placement) is a separately-owned follow-up PR — no changes under `interfaces/inline/textual_chat/`.

## (g) Gates

- `ruff check src tests` — green.
- `python scripts/test_tier_audit.py --strict tests/test_3300_p3_cancel_by_id.py` — 15/15 OK, 0 Tier-4 violations.
- `uv run python -m pytest --timeout=120 -q -n auto` (foreground, alone in the worktree) — **9189 passed, 36 skipped, 0 failures**.

## Test plan
- [x] `test_3300_p3_cancel_by_id.py` — 15 new tests: 3 semantics, truncate-falsify (+ manually-verified strip), 2 race-ordering tests, skip-at-consume, `RemoteQueueView.apply_inbox_cancel`, end-to-end delta→view, agui endpoint/client/in-process transport parity, WAL replay symmetry (`inbox_cancel` ≅ `inbox_consume`).
- [x] Existing P1/P2a/P2b/dual-stream/agui-profile-completeness suites re-run green (no regression).

### DIFF
--- a/docs/reference/runtime/agui-transport.ja.md
+++ b/docs/reference/runtime/agui-transport.ja.md
@@ -25,6 +25,15 @@ Reyn のチャットクライアントはストリームを消費する UI で�
     下記「Human-in-the-loop answering」参照)。
   - `{"type": "cancel_inflight"}` — in-flight なターンを協調的にキャンセルする(Ctrl-C
     seam)。
+  - `{"type": "cancel_queued", "msg_id": "..."}` — まだ dispatch されていない(未実行の)
+    inbox メッセージを id 指定でキャンセルする(#3300 P3)。上記の `cancel_inflight` とは
+    異なる意図である — こちらは「まだ turn が始まっていない特定の queued item」を対象と
+    し、現在実行中の turn は対象にしない。サーバー側の意味論(`Session.cancel_queued`):
+    queued の場合 → 除去(WAL `inbox_cancel` tombstone + 同期的な snapshot-prune の後、
+    `inbox_cancel` chat-event delta を emit。下記「STATE_* — ステータス read-model」と
+    `reyn.event.inbox_cancel` を参照)。既に dispatch 済みの場合 → no-op(`cancel_inflight`
+    へのエスカレーションは行わない)。冪等(同じ id への2回目のキャンセルは no-op — at-most-once
+    の再送に対して安全)。
   - `{"type": "heartbeat"}` — liveness の keepalive。
 
   server がモデル化していない入力 type は**グレースフルな no-op**(`200` の ack)であり、
@@ -277,6 +286,16 @@ WaitingOn ラベル)は**read-model**であり、ファイルミラーではな�
 snapshot から正しい queue + turn-active 状態を得る。P2a はこの状態の publish のみで、
 sent-queue widget としての描画は P2b。
 
+item が `queue` から抜けるのは、同じ snapshot+delta channel 上の互いに排他な2つの
+granular chat-event delta のいずれかを経由する — `turn_started`(dispatch された。下記
+「Working-indicator path」参照)、または `inbox_cancel`(上記の `cancel_queued` client
+message で id 指定キャンセルされた。#3300 P3)。server 自身の atomic な
+queued/dispatched 判定が、ある item についてこの2つのうち正確に一方のみが必ず発火する
+ことを保証する(両方が発火することはない)。`inbox_cancel` は `msg_id` と `seq`(
+`user_submitted`/`turn_started` と同じ order-race-gate token — 下記
+`reyn.event.inbox_cancel` を参照)を運ぶ。granular delta をマージする client は
+(`turn_started` が `chain_id` でマッチするのとは異なり)`msg_id` で item を除去する。
+
 client は snapshot から自身のステータスビューを seed し、各 delta をマージするため、
 remote のステータスパネルは常に server の値を反映する。
 
@@ -337,6 +356,7 @@ display 行のテキストである。
 |--------------------------------------|--------------------------------------------------|
 | `reyn.event.user_answered_intervention` | ユーザーが intervention に回答した              |
 | `reyn.event.user_submitted`          | ユーザーがターンを送信した(#3300 P1 C)— RAW text + chain_id + msg_id + seq + meta を運ぶ。各 surface がそれぞれの render 境界で neutralize する。`msg_id`/`seq` は #3300 P2a の sent-queue correlation id + order-race-gate token |
+| `reyn.event.inbox_cancel`            | 未 dispatch の queued user メッセージが id 指定でキャンセルされた(#3300 P3、`cancel_queued` client message 経由)— `msg_id` と `seq` を運ぶ。サーバー権威の sent-queue 除去シグナルであり(client-local な「キャンセル成功」応答ではない)、同じ `msg_id` について `turn_started` と排他である |
 
 ### `reyn.intervention.<kind>`
 
--- a/docs/reference/runtime/agui-transport.md
+++ b/docs/reference/runtime/agui-transport.md
@@ -24,6 +24,16 @@ A2A; tools are MCP; observability export is OTEL. Those are separate surfaces.)
     round-trip; see "Human-in-the-loop answering" below).
   - `{"type": "cancel_inflight"}` — cooperatively cancel the in-flight turn (the
     Ctrl-C seam).
+  - `{"type": "cancel_queued", "msg_id": "..."}` — cancel-by-id an UNDISPATCHED
+    (queued, not-yet-running) inbox message (#3300 P3). A DIFFERENT intent from
+    `cancel_inflight` above: this targets one specific queued item that has not
+    started a turn yet, never the currently-running turn. Server-side semantics
+    (`Session.cancel_queued`): queued → removed (WAL `inbox_cancel` tombstone +
+    synchronous snapshot-prune, then an `inbox_cancel` chat-event delta, see
+    "STATE_* — the status read-model" and `reyn.event.inbox_cancel` below);
+    already dispatched → a no-op (never escalated to `cancel_inflight`);
+    idempotent (a second cancel of the same id is a no-op, safe for an
+    at-most-once retry).
   - `{"type": "heartbeat"}` — a liveness keepalive.
 
   An input type the server does not model is a **graceful no-op** (a `200` ack),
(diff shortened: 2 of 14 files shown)


tya5:
**[per-PR-coder]** — addressed the blocking doc-drift finding + answered the getattr/typed question.

## doc-drift fixed (same PR, commit 3dd2f262)

Re-read the full sections this PR touches in `docs/reference/runtime/agui-transport.md` (not just a keyword grep), and updated all three places the new wire contract actually shows up:

1. **"supported message types" list** — added `{"type": "cancel_queued", "msg_id": "..."}`, with the queued/dispatched/idempotent server semantics spelled out and explicitly distinguished from `cancel_inflight` (running-turn cancel vs undispatched-queue-item cancel — different intent, never escalated between the two).
2. **`STATE_*` queue-state section** — added a paragraph on how an item leaves `queue`: exactly one of two mutually-exclusive granular deltas, `turn_started` (dispatch) or `inbox_cancel` (cancel-by-id), server-atomic-guaranteed never both. Noted `inbox_cancel` carries `msg_id`+`seq` and is matched by `msg_id` (not `chain_id`, unlike `turn_started`).
3. **`reyn.event.<etype>` table** — added the `reyn.event.inbox_cancel` row (mirrors the `agui/profile.py` `CUSTOM_PROFILE` entry this PR added).

Checked whether the "AG-UI event coverage" numeric table (State/Lifecycle/Tool/Text/Special/Activity/Reasoning counts) needed updating — it doesn't: those counts track the *standard* AG-UI vocabulary reyn maps to, and `inbox_cancel` folds into `CUSTOM` (like `user_answered_intervention`), not a new standard mapping, so none of those totals change.

**`.ja.md` mirror**: not a blanket-mirror obligation, but the ja file's message-type list and queue-state prose were exhaustive enumerations ("サポートされるメッセージ type は以下の通り") that this PR now falsifies as written, so I fixed the same three spots there too (own-claims-now-false, not a parity rule).

## getattr/typed question, answered with citations

The existing `cancel_inflight` handler in the SAME file is already the identical `getattr` + `callable` shape, not typed dispatch — `src/reyn/interfaces/transport/agui/endpoint.py:460-463`:

```python
elif ptype == "cancel_inflight":
    cancel_fn = getattr(session, "cancel_inflight", None)
    if callable(cancel_fn):
        await cancel_fn()
```

My new `cancel_queued` branch (`endpoint.py:464-472`) follows this exact precedent — it is not a new duck-typing introduction next to an otherwise-typed handler; it's consistent with the one existing analogous handler in this file. `session` here is the `Session` returned by `registry.attach(agent_name)` (typed), but the existing code still guards via `getattr`/`callable` rather than a direct typed call — I kept that shape rather than deviating unilaterally. Happy to change both together in a follow-up if you'd rather see this tightened to a direct typed call, but that would be a pre-existing-pattern change beyond this PR's scope, not something to diverge on unilaterally here.

## Gates re-run after the doc fix

- `ruff check src tests` — green.
- `python scripts/test_tier_audit.py --strict tests/test_3300_p3_cancel_by_id.py` — 15/15 OK, 0 Tier-4 violations.
- `uv run python -m pytest --timeout=120 -q -n auto` (foreground, alone in the worktree) — **9189 passed, 36 skipped, 0 failures**.

tya5:
**[architect]** — **co-vet #3306 (#3300 P3 Y-server: cancel-by-id): CLEAR。** 実装しません。正木（`per-pr/cancel-by-id-y-server-3300`、HEAD=3dd2f262）で独立 strip ＋ 機構検証。

## ★§1（recovery band、本 arc 最重要の pin）— 実装・witness ともに正しい
`SnapshotJournal.cancel_inbox` は **WAL `inbox_cancel` tombstone ＋ `snapshot.inbox` の同期 prune** の両方を、`consume_inbox` と同型で実施。設計パスで私が出した contract 補正どおりです。

**★strip（決定的）**: **snapshot prune を除去し WAL tombstone だけ**にする＝*設計パスで私が警告した、まさにその誤り*を再現 → **`test_cancelled_message_survives_wal_truncation_below_its_source_events` を含む 7 RED**。
→ truncate-falsify gate は **非-vacuous**。「WAL-event-derived な recovery state が snapshot-backed でない＝silent data-loss / resurrection vector」（CLAUDE.md recovery gate）が、CI で恒久的に守られる形になりました。

## ★F（no-await critical section、#79）— docstring を信用せず**機構で検証**
docstring は「`cancel_inbox` は async だが内部は完全同期＝suspend しない」と主張。**AST で own-body await を数えて検証**しました（nested 定義は除外して集計）:
```
_wal_append_nowait   sync   own_body_awaits=0
cancel_inbox         async  own_body_awaits=0
save_nowait          sync   own_body_awaits=0   (nested `_write` に await=1 ＝ fire-and-forget、呼び出し側は suspend しない)
```
→ `await cancel_inbox(...)` は **制御をループに返さない**ので、presence-check → tombstone → prune → save が **atomic**。主張は事実です。

**★strip**: critical section に `await asyncio.sleep(0)` を **1つ挿入** → **`test_cancel_scheduled_before_dispatch_wins_exclusively` RED**。→ exactly-one-exit が atomicity に**実際に依存**しており、破れを test が捕捉することを確認。**両 scheduling 順**（cancel-先行 / dispatch-先行）に test があるのも設計どおりです。

## ✅ その他 gate（すべて非-vacuity 確認）
| gate | strip | 結果 |
|---|---|---|
| skip-at-consume | 2 箇所の skip 判定を `False` 化 | **2 RED**（skip 専任 test ＋ cancel-先行 test） |
| WAL replay（restore 経路） | `("inbox_consume","inbox_cancel")` から `inbox_cancel` を除去 | **RED**（`test_agent_snapshot_apply_events_inbox_cancel_prunes_like_inbox_consume`） |
| seq-gate 再利用（私の pin） | — | ✅ `RemoteQueueView.apply_inbox_cancel` が **同じ `_last_seq` protocol**（再発明なし）。msg_id 指定（`turn_started` の chain_id 一致とは別軸）も正しい |
| 3-event 排他契約（§6a） | — | ✅ `turn_started`(promote) XOR `inbox_cancel`(remove) を server が atomic 保証、dispatched→no-op、idempotent |

**replay の意味論も妥当**: `inbox_cancel` を `inbox_consume` と**同一扱いで prune**（「dispatched か cancelled か」の区別は record 時の意味論であって、snapshot 再構成には "この id は消えた" だけが必要）。docstring の説明どおりで、過剰な状態を持ち込んでいません。

## CI gate / regression
ruff = **All checks passed** / tier-audit = **0 errors / 0 warnings** / 当該 file = **15 passed** / **広域 recovery band（inbox・wal・snapshot_journal・restore_all・agent_snapshot・agui）= 241 passed**（payload/WAL 形状に触れる PR ゆえ私からも実施）。

doc-drift も解消を確認（`agui-transport.md` に `cancel_queued` POST 型・`inbox_cancel` delta・snapshot-prune を含む semantics が記載）。lead の「`getattr` 照会は `cancel_inflight` handler と同一形＝既存踏襲」判断にも同意します（本 PR で是正を求める筋のものではありません）。

## 判定
**CLEAR**。設計パスで最重要 pin とした **§1（snapshot-prune ＋ truncate-falsify）** と、自己レビューで挙げた **F（no-await critical section ＋ cancel-during-dequeue race）** が、いずれも *実装され、かつ strip で witness されている* ことを確認しました。実CI green ＋ 本 co-vet でマージ可（Y-server は headless/runtime ゆえ tui-coder witness 不要に同意）。

次の **Y-client**（sent-queue row 除去 ＋ composer prepend）では、#3305 で確立した **rendered-content fidelity ＋ neutralize の 2-site 独立 witness** が同じクラスの gate になります（cancel された text が input box に prepend される＝untrusted text の新たな描画面）。dispatch 時に brief へ入れてください。

PR #3311, opened 2026-07-26
fix(tui): tab-ify intervention panel — #3299 P5 (#3308)

tya5:
**[per-PR-coder]** — part of #3299, Closes #3308

## Summary

Tabs the intervention panel: one `TabPane` per PENDING intervention (`InterventionPanel.add_pending`), added the moment its frame arrives and never swapped out from under the user. Replaces #3307 (P2)'s single-form FIFO re-route entirely.

- `TabbedContent.add_pane` auto-activates a new pane ONLY when the content was previously empty (verified against installed Textual 8.2.8) — this alone gives the "never steal the active tab" invariant, no app-level bookkeeping needed.
- Answering a tab (`on_intervention_panel_choice_selected` / `_text_submitted`) delivers targeted at THAT tab's `intervention_id` (the message itself carries `event.key` — no more app-level "currently displayed" tracking) and marks the tab ✓/inert (`InterventionPanel.mark_answered`: label gets a `✓` prefix, its `RadioSet`/`Input` gets `disabled = True`, and if it was the active tab, focus moves to the panel's own `Tabs` bar so `Left`/`Right` keeps working). The tab STAYS mounted until every pending intervention resolves, at which point the whole panel collapses (`collapse_all`).
- `Left`/`Right` switch tabs via a `priority=True` `Binding` on the panel — **co-vet-corrected keymap contract** (see issue comment): `RadioSet` already binds `left`/`right` as `up`/`down` aliases (`down,right → next_button`, `up,left → previous_button`, Textual 8.2.8), so a plain ancestor binding would never fire while a `RadioSet` has focus. Verified `textual/app.py` (~L3966-3988, L4136) and `textual/screen.py` (~L407-455): Textual runs a **priority pass first**, outermost-ancestor-down, matching only `priority=True` bindings, before the normal focused-widget-outward walk where `RadioSet`'s own binding would win. A `priority=True` binding on the panel intercepts first.
- Pre-highlight (owner decision (A)) is now **unconditional** on every tab activation (`on_tabbed_content_tab_activated`) — no `initial`/re-route distinction needed anymore, since each pending intervention gets its OWN fresh `RadioSet` (Textual's own `RadioSet._on_mount` already pre-highlights index 0 the moment that instance is constructed) and the active tab never moves except by explicit navigation.
[...]

PR #3312, opened 2026-07-26
feat(llm,runtime,transport): agent_delta chat-event for streaming partials — #3288 ③b

tya5:
**[per-PR-coder]** — Phase ③b of the token-streaming arc (#3288): carries ③a's (#3304, merged) streamed content-deltas up to the transport as an **`agent_delta` chat-event** — the owner's ratified L4 replacement (issue #3288 comment thread) for the ADR's original "add `agent_delta` to the OutboxMessage closed vocabulary" wording. part of #3288.

## Canonical contract (from the issue thread)

The owner ratified the architect's proposal: a partial rides `host.events` (the same chat-event channel `user_submitted` / `router_represent_round` already use), **never** an `OutboxMessage` kind — `OutboxMessage.__post_init__` validates `kind` against the closed display vocabulary, so an `agent_delta` OutboxMessage would force registering it there, reintroducing exactly the category error P1 C ("input in an output channel") already designed out in reverse (here: a transient, non-persistent notification in the persist/display channel). Invariants preserved: L9 whole-persist, capability-is-the-driver (③a), usage single emission (③a, untouched).

## What changed

- **`src/reyn/llm/llm.py`** — `recorded_acompletion` / `call_llm_tools` gain an opt-in `on_content_delta: Callable[[str], None] | None` parameter. Inside ③a's `_stream_and_reconstruct`, it fires once per real chunk carrying non-empty `delta.content`, BEFORE that chunk is appended to `chunks` (so it fires against real per-chunk consumption, not a batched replay). Never invoked on the whole-collect path — capability-gated by construction, the same way ③a's streaming decision is. A raising callback is caught + logged, never allowed to abort the LLM call.
- **`src/reyn/runtime/router_loop.py`** — `RouterLoop._emit_agent_delta` wires the callback into the primary reply call site (`on_content_delta=self._emit_agent_delta`), emitting `self.host.events.emit("agent_delta", text=text, chain_id=self.chain_id)` — never `host.put_outbox`.
- **`src/reyn/interfaces/transport/frames.py`** — `"agent_delta"` joins the transport's forward-set (`renderer_chat_events()`) via a new `_STREAMING_EVENTS` constant — the ONE deliberate exception to "derived, not hand-listed" (module + function docstrings updated to say so). This is legal per the dual-stream completeness gate's actual direction (`consumed ⊆ forwarded`, never the reverse) — forwarding an event nobody consumes yet is not a coverage gap.
[...]

PR #3316, opened 2026-07-26
docs: fix stale JA claims in agui-transport + add verification-hazards §10

tya5:
**[per-PR-coder]** — docs-only PR fixing doc-sync drift falsified by five PRs merged this session (#3306/#3309/#3311/#3312/#3313/#3315), per CLAUDE.md's hard rule ("a doc describing a mechanism is stale the moment that mechanism's code changes").

## Part 1: `docs/reference/runtime/agui-transport.ja.md`

Only #3306 and #3309 touched the JA sibling this session (`git log -- docs/reference/runtime/agui-transport.ja.md`), while #3311/#3312/#3313/#3315 changed the EN file's behavior claims without a JA update. This is NOT the en/ja mirror rule (no blanket obligation exists) — it's the plain doc-staleness rule: JA was left **actively asserting false claims**. Fixed exactly the four falsified spots, no full mirror attempted:

1. **Text lifecycle** (「トークンストリーミングは scope 外である」) — falsified by #3288 ③b (`agent_delta` chat-event per raw LLM delta) + ③d (mapped onto the standard `TEXT_MESSAGE_CONTENT` surface). Rewrote the section with the plain-vs-streamed split now in EN: a message that never streamed keeps the old whole-message triplet; a message that DID stream gets a real multi-CONTENT sequence (`START` at first delta, one real `CONTENT` per delta, `END`-only completion — never a second full-text `CONTENT`), including the late-joiner-window closure via `TextStreamTracker`.
2. **`reyn.event.<etype>` table** — was missing the `agent_delta` row entirely (EN has 4 rows, JA had 3). Added the row plus the intro paragraph explaining `agent_delta` is a separate streaming-notification axis and that the row describes the plain codec's isolated behavior, not what ships on the connected AG-UI wire (`encode_frame_wire_streaming` maps it onto standard `TEXT_MESSAGE_CONTENT` instead, per ③d).
3. **Coverage table "Text" row** — still said "conforming triplet" (old single-triplet-only disposition). Updated to "conforming triplet, plain と streamed 両方", matching EN's disposition text for the streamed case.
4. **"Local ≡ remote" section** — described the intervention-answer path as「TUI の選択肢リージョン」(choice-region). #3311 replaced that surface with a tab-ified panel (`intervention_panel.py`, #3299 P1/P2 base + #3308/#3311 tab-ify): one tab per PENDING intervention, answering targets that tab's intervention id and marks it ✓/inert without removing it (so several simultaneously-pending interventions are each independently answerable). JA previously never mentioned tabs at all; rewrote the parenthetical to match.

**Checked, not falsified**: ③c's EN addition (the new "Textual TUI streamed-reply rendering (#3288 ③c)" subsection describing the `chain_id`-keyed coalesce into one `FlowView` entry) does not contradict anything JA currently asserts — JA never described TUI-side rendering of streamed replies at all, so there's nothing to fix there (adding the whole subsection would be a full mirror, out of scope per the docs-only brief).

## Part 2: `docs/deep-dives/contributing/verification-hazards.md` §10

Added the next-numbered section (10 — file currently ends at §9). New hazard, distinct from the two nearest existing ones (checked, not duplicating): §6 covers an *empty-set* vacuity (the assertion never runs because the derived set is empty); §7 covers a *faked collaborator* removing the loud-failure backstop. §10 is a third, independent shape: the set is non-empty and the collaborator is real, but the assertion **only inspects the terminal state, an adjacent path, or mere existence** — so killing the live mechanism entirely still leaves the gate green.

Three real instances, all measured this session:
1. Terminal-state-only assertion (#3288 ③c) — "N deltas → one entry with the full text" asserted only after completion, passes with zero deltas.
2. Positive control on a different delivery path (#3288 ③b) — witness injected via `_queue.put` bypassed `push_event` under test.
3. Existence check instead of the real property (#3299 P5) — `region.height > 0` stayed green while widgets were pushed off-screen; fixed with bidirectional containment.

Closing procedure: apply the gate to the known-broken build and confirm it goes RED before shipping.

## Verification

- `ruff check src tests` — clean (docs-only, no-op as expected).
- `mkdocs build --strict -f .mkdocs/mkdocs.yml` — exit 0, zero warnings tied to either touched file (pre-existing warnings on unrelated files are unchanged).
- No doc-anchor/doc-gate test files reference either touched doc (`grep -rl` over `tests/` came up empty), so no targeted pytest run was needed.
- Did not run the full pytest suite (docs-only change, multiple coders share this machine).

## Test plan

- [x] `ruff check src tests`
- [x] `mkdocs build --strict` (local, not just relying on CI)
- [x] Re-read each falsified JA section's surrounding prose (not just the flagged line) against the current EN text before editing
- [x] Confirmed §10 doesn't duplicate §6/§7 by re-reading both

## Part 3 (addendum, requested by lead-coder): streaming call-path clarification

Widens this PR's scope from JA-only to also touching the **EN** `docs/reference/runtime/agui-transport.md` (streaming/text-lifecycle section), plus a matching JA addition.

`RouterLoop` has three production call sites for `call_llm_tools`, but only the primary reply (`run_loop`) passes `on_content_delta`; `_run_structured_answer_turn` (schema-constrained turn, output parsed via `json.loads` — partial JSON would be unparseable to stream) and `_force_close_call` (a terminal wrap-up, not narrative body content) both omit it intentionally. Added a 1-2 sentence note (referenced by function name, not line number) so this doesn't read as an unnoticed gap in a future audit. Verified all three call sites directly against the current `src/reyn/runtime/router_loop.py` before writing.

No basis found for recommending the structured-answer path also stream — the note stands as-is, no implementation follow-up flagged.

Re-verified after this addendum: `ruff check src tests` clean, `mkdocs build --strict -f .mkdocs/mkdocs.yml` exit 0 with no new warnings.

### DIFF
--- a/docs/deep-dives/contributing/verification-hazards.md
+++ b/docs/deep-dives/contributing/verification-hazards.md
@@ -187,6 +187,55 @@ rather than treating the gate's earlier green run as the final word; that
 issue-state read is the only thing that actually observes what happened at
 merge time.
 
+## 10. Born-vacuous: a live mechanism behind a gate that only inspects the terminal state
+
+Distinct from §6 (an *empty set* makes the assertion never run) and §7 (a
+*faked collaborator* removes the loud-failure backstop): here the set is
+non-empty and the collaborator is real, but the **assertion only inspects
+the terminal state, an adjacent path, or mere existence** — so killing the
+live mechanism entirely still leaves the gate green. Three independent
+instances, all measured the same session (#3288/#3299 streaming + panel
+arc):
+
+- **Terminal-state-only assertion** (#3288 ③c). The gate asserted "N deltas
+  → exactly one entry whose content is the full text" — but only *after*
+  the completion frame arrived. The completion frame alone creates that
+  one-entry, full-text result regardless of whether delta-coalescing ever
+  ran, so the gate passed with **zero deltas delivered**. Fix: assert a
+  **mid-stream cross-section** — before completion arrives, exactly one
+  entry exists and its content is *partial*. Zero deltas then produces zero
+  entries at that cross-section ⇒ RED, and the arrival witness is
+  *subsumed* by the gate itself instead of needing a separate positive
+  control.
+- **Positive control on a different delivery path** (#3288 ③b). A
+  positive-control witness was injected via `_queue.put`, bypassing the
+  `push_event` path the production code actually uses — so breaking
+  `push_event` left the witness passing and the gate vacuous; it proved
+  only that a *neighbouring* path was alive, not the one under test. Fix:
+  the positive control must traverse the **same delivery path** as the
+  thing being verified. (Worth naming honestly: this wrong prescription
+  came from the reviewer in the first pass, and was only caught when the
+  coder's deviation from it was surfaced for a ruling instead of being
+  silently accepted — a reviewer's gate design is not exempt from this
+  hazard.)
+- **Existence check instead of the real property** (#3299 P5). A layout
+  gate asserted `region.height > 0` and `display` truthiness, and stayed
+  green on a build where the intervention panel had visually taken over
+  the whole screen — the widgets were not zero-height, they were **pushed
+  off-screen** (composer at `y=25` on a 24-row screen; flowview at `y=-8,
+  h=1`). An upper-bound-only containment check (`y + h <= screen_height`)
+  is equally useless — trivially true for any negative `y`. Fix:
+  **bidirectional** containment (`y >= 0` AND `y + h <= screen_height`)
+  plus a not-squashed floor on height, parametrized over at least two
+  terminal sizes. Related corollary: an `ImportError`-based RED proves a
+  reference vanished, not that the value it named ever reached the
+  rendered output — existence of a symbol is not existence of the effect.
+
+**Apply**: before shipping any gate, apply it to the known-broken build (the
+one the mechanism, if absent, would produce) and confirm it goes RED. A
+gate that has never been seen failing on a real defect — only ever seen
+green — is unproven, no matter how plausible its assertion reads.
+
 ## See also
 
 - [Testing policy](testing.md) — Tier model, Mock vs Fake, decision flow.
--- a/docs/reference/runtime/agui-transport.ja.md
+++ b/docs/reference/runtime/agui-transport.ja.md
@@ -128,6 +128,15 @@ AG-UI 仕様は、text lifecycle として**`TEXT_MESSAGE_START` → 1 つ以上
 必須としている。裸の `TEXT_MESSAGE_CONTENT` は不正である(厳格な汎用クライアントはそれを
 破棄する)。
 
+streaming が適用されるのは**narrative reply 経路のみ**である。`RouterLoop`
+から `call_llm_tools` を呼ぶプロダクション経路は 3 つあるが、
+`on_content_delta` を渡すのは primary reply(`run_loop`)だけである。
+**structured-answer turn(`_run_structured_answer_turn`)は意図的に
+非-streaming**——出力を `json.loads` でパースする schema 制約付きターンで
+あるため、部分 JSON を stream しても解析不能で無意味である。**force-close
+の wrap-up 呼び出し(`_force_close_call`)も非-streaming**である(本文では
+なく終端の締めであるため)。
+
 **一度もストリーミングしなかったメッセージ**(provider capability なし、ADR-0039 P3a/③a)
 は、この triplet に plain な whole-message としてワイヤーに乗る——メッセージごとに生成
 される id を伴い(reyn の outbox には安定した message id がない)、単一の CONTENT の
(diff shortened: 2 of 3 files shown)


tya5:
**[per-PR-coder]** — addendum requested by lead-coder: added a 1-2 sentence note that streaming applies to the narrative-reply call path only.

**Scope change**: this widens the PR from JA-only to also touching the EN `docs/reference/runtime/agui-transport.md` (streaming/text-lifecycle section), plus the matching JA addition.

**Verified against source before writing** (`src/reyn/runtime/router_loop.py`, referenced by function name, not line number, per instruction):
- `run_loop` → passes `on_content_delta=self._emit_agent_delta` to `call_llm_tools`.
- `_run_structured_answer_turn` → does not pass it (schema-constrained turn, output parsed via `json.loads` — partial JSON would be unparseable and pointless to stream).
- `_force_close_call` → does not pass it (terminal wrap-up call, not narrative body content).

Confirmed all three call sites directly by reading the current file (function names matched what was given; line numbers had drifted slightly since the original grep, as expected).

Re-ran `ruff check src tests` (clean) and `mkdocs build --strict -f .mkdocs/mkdocs.yml` (exit 0, no new warnings on either touched file) after this addendum.

No reason found to recommend streaming the structured-answer path — the doc note stands as written; no PR-body objection to raise.

tya5:
**[architect]** — **co-vet #3316 (docs: JA sync + verification-hazards §10): CLEAR。** 正木（`docs/agui-transport-ja-sync-verification-hazards-10`、HEAD=91b0e578）で確認。

## ✅ (a) JA 修正は現在の EN 主張と一致 — **過剰修正も不足もなし**
- 偽になっていた中核（「reyn は whole-message であり、**トークンストリーミングは scope 外**」）を撤回し、**never-streamed（plain triplet）** と **actually-streamed（multi-CONTENT）** を分けて記述。END のみ／END の `_reyn` が唯一の再構成 authority／delta 連結による再構成はしない、まで ③b/③d/③c の実装と一致しています（私が各 PR で strip 検証した内容と突き合わせ済み）。
- ★**独立 sweep で残存主張を2件検出しましたが、どちらも訂正不要**でした: `ja.md:181` と `:549` の「reyn は whole-message（トークンストリーミングなし）」は **Reasoning lifecycle 節**のもので、③ が stream するのは narrative reply の `delta.content` のみ＝**reasoning は今も whole-message** ゆえ**主張は真のまま**。→ **過剰修正を避けた判断が正しい**（ここまで直していたら偽になっていました）。
- EN 側 +8 は、私が指摘した **3 call site の事実**（`run_loop` のみ `on_content_delta` を渡す／`_run_structured_answer_turn` は部分 JSON が解析不能ゆえ意図的に非-streaming／`_force_close_call` は本文でない）を**理由つきで**記載。将来の「なぜここは stream しない?」の再調査を防ぐ形になっています。

## ✅ (b) §10 は §6/§7 と重複していない — **第3のカテゴリとして正しく分離**
本文自身が明示的に切り分けています:
- **§6 vacuity guard** = *空集合*で assertion が一度も走らない
- **§7 falsifiable-event root** = *偽の collaborator* が loud-failure の backstop を外す
- **§10** = **集合は非空・collaborator も本物**、しかし **assertion が終端状態／隣接経路／単なる存在しか見ていない**ため、**機構を丸ごと殺しても green のまま**

3 実例（終端のみ assert／別配送経路の positive control／存在チェック）はいずれも本アークの**実測**で、各々に fix が併記されています。締めの **Apply**（「gate を出荷する前に、機構が無いときに生じる既知の壊れたビルドへ当てて RED を確認せよ／green しか見たことがない gate は未証明」）が実務的で、§10 の価値を決めています。

★**私自身の誤処方が正直に記録されている点は、そのまま残してください**（"this wrong prescription came from the reviewer in the first pass … a reviewer's gate design is not exempt from this hazard"）。事実として正確で、かつ **coder が逸脱を黙って採らず裁定に上げた運用が効いた**という再現可能な教訓になっています。私の体面のために薄める理由はありません。

## ✅ (c) 「③c の EN 新節は JA に対応記述がないので触らない」は **CLAUDE.md どおり**
CLAUDE.md rule 5 は「**JA は意図的に部分的な curated subset**。EN sibling があるだけ／変わっただけでは `.ja.md` を触る義務は生じない。**その `.ja.md` が現在主張していることを PR が偽にする場合にのみ**直す」。
実測: EN の新節 *Textual TUI streamed-reply rendering (#3288 ③c)* に対応する JA 節は**存在しません**＝**偽になる JA の主張がない**ので、**触らないのが正しい**。逆に「EN にあるから JA にも足す」と動くと、rule 5 が明示的に否定する blanket mirror 義務を自ら作ることになります。判断は正確です。

## ✅ ビルド
`mkdocs build --strict`（`.mkdocs/mkdocs.yml`）: **本 PR 由来の warning はゼロ**。出力の anchor 系 INFO は **main でも 87 件**出る **pre-existing**（`pipeline-dsl.ja.md` / `present.ja.md` 等、本 PR の3ファイルとは無関係）で、この PR の責任範囲外です。

## 判定
**CLEAR**（docs-only）。CI green でマージ可。
★ドキュメントとしての価値が高いのは §10 で、**「gate 自体を欠陥ビルドで falsify する」という手順が明文化された**のが本セッションの最大の持ち帰りだと思います。私も #3311 で born-vacuous な gate を、#3312 で誤った positive control を出しており、**この節は私自身への gate でもあります**。

PR #3307, opened 2026-07-26
fix(tui): by-id intervention delivery, first-option focus, churn-zero Q→A entry — #3299 P2

tya5:
**[per-PR-coder]** — Phase 2 of the intervention-panel arc (part of #3299). P1 (the panel widget) is merged; this PR fixes the architect's self-review mis-delivery finding, applies the owner's auto-focus decision, and refines the churn-zero EntryState per §5.

## (a) Multi-pending by-id delivery (correctness fix)

**Problem** (architect self-review on P1): the panel was single-slot (`_pending_iv_entry`) — a second concurrently-pending intervention silently overwrote the first's flow-entry handle — and delivery was head-targeted (`answer_intervention_choice(choice_id)` with no id, resolved via `deliver_answer_to(head, ...)`). But `outstanding_interventions` legitimately holds multiple pending entries at once (e.g. restore's FIFO re-enqueue), so a stale head could receive an answer the user actually gave to a *different*, currently-displayed intervention.

**Fix**:
- `TextualChatApp` now tracks **every** pending intervention in `_pending_ivs: dict[id, (entry, msg, intervention_id)]`, and `_iv_current_key` marks which one the panel is displaying. A second arrival never overwrites the first — it queues.
- `ClientTransport.answer_intervention_choice` / `answer_intervention_text` gained an optional `intervention_id: str | None = None` keyword. `InProcessTransport` threads it to the **existing**, already id-aware `session.answer_intervention_by_id` funnel (`InterventionHandler.deliver_answer_to` itself is unchanged); `None` preserves the old head-targeted path for any caller that doesn't track an id. `AgUiTransport` honors an explicit id over its single wire-tracked `_pending_intervention_id` slot (a wider multi-pending *wire* protocol is out of this PR's scope — confined to `interfaces/inline/textual_chat/`).
- `on_intervention_panel_choice_selected` / `on_intervention_panel_text_submitted` now pass `intervention_id=self._current_intervention_id()`.
- Resolving re-routes the panel to the next pending intervention (FIFO) instead of leaving it blank.

**Strip-falsify test** (`test_multi_pending_answer_targets_the_displayed_intervention_by_id`): with TWO pending interventions, answering the displayed one delivers to that id (`answered_choice_ids == ["iv-1", "iv-2"]`); reverting the `intervention_id=` kwarg flips it to `[None, None]` — verified locally, RED. `test_second_pending_intervention_does_not_overwrite_the_first` witnesses the overwrite fix directly.

[...]

PR #3320, opened 2026-07-26
fix(tui): restore an answered intervention as one self-contained Q→A entry — #3299 P4

### LINKED ISSUE #3299: TUI intervention: move interaction out of FlowView into a grouped panel widget

tya5:
**[per-PR-coder]** — Phase 4 (final phase) of the intervention-out-of-FlowView arc (#3299): a restored answered intervention now shows the question AND the answer as ONE flow entry, not two (and not a blank/missing one).

Closes #3299

## Design (settled in the issue thread — architect + owner, do not re-litigate)

The original plan ("correlate the prompt record with the answer record, like #3295's tool call/result coalesce") turned out to be **impossible**: `InterventionHandler.announce` publishes the prompt to the outbox but never calls `_append_history` — the question text was never in `history.jsonl` at all. Only the answer was (a `role="user"` entry stamped `meta["intervention_id"]` / `meta["intervention_kind"]`).

**Approved fix (C)**: store the prompt on the **answer record's own `meta`**, so one history entry is self-contained — no correlation key needed at all (this arc hit the "guessed key" defect three times: #3299 P2 head-of-queue, #3287 text-matching, #3300). Self-contained records are also safe under P5's out-of-order answering (each answer carries its own question).

## What changed

- **`src/reyn/runtime/chat_message.py`**: three new meta-key constants, `INTERVENTION_PROMPT_META_KEY` (`"intervention_prompt"`), `INTERVENTION_DETAIL_META_KEY` (`"intervention_detail"`), `INTERVENTION_ANSWER_META_KEY` (`"intervention_answer"`) — mirroring the existing `TOOL_STATUS_META_KEY` pattern. `INTERVENTION_ANSWER_META_KEY` exists because a **closed-set** answer's own `ChatMessage.content` is `""` (the panel's `choice_id_override` path passes `text=""`) — so `m.text` alone can't reconstruct "what was answered" for that path; this key always carries the resolved answer-display text (a matched choice's `label`, or the raw free-text answer).
- **`src/reyn/runtime/services/intervention_handler.py`**: `InterventionHandler.deliver_answer_to` now stamps all three keys onto the SAME history record it already appends. RAW values (untrusted / LLM-derived — `ask_user` prompts and a matched choice's label both come from a model tool-call) — no neutralization at write time.
- **`src/reyn/interfaces/inline/textual_chat/restore.py`**: `project_restored_frames`'s `role == "user"` branch now checks for `INTERVENTION_PROMPT_META_KEY` first; when present, it projects straight into the SAME `kind="intervention"` shape a live resolved entry uses (`meta["prompt"]`/`meta["detail"]` for the head, `meta["_answer_label"]` for the "✓ answered: ..." line) — no presenter change needed for the shape, and no correlation/join step. Absent (pre-P4 history, or non-intervention `user` turns) → falls through to the pre-existing plain-`user` projection, unchanged.
- **`src/reyn/interfaces/inline/textual_chat/presenter.py`**: `ReynPresenter._present_intervention_pending`'s RESOLVED branch now neutralizes `_answer_label` at its render call site (previously only `prompt`/`detail` were neutralized there). This is idempotent for the LIVE path (a closed-set answer's label is already neutralized when the panel builds it; free-text is the user's own trusted input) but is the ONE real neutralization boundary for a RESTORED answer, which arrives RAW by design.

## Gates (each strip-falsified locally, observed RED, then reverted — `tests/test_3299_p4_intervention_restore_qa.py`)

1. **One entry per intervention** — `test_restored_answered_intervention_is_one_entry_with_both_qa` / `_free_text_...` / `_out_of_order_...` / `_producer_closed_set_...`. Strip: force `restore.py`'s `prompt` lookup to always be falsy (simulating "the branch doesn't exist") → all four RED (`[]` frames instead of `["intervention"]`).
2. **Both Q and A present** — same tests, rendered-text assertions.
3. **LLM payload unchanged** — `test_llm_payload_unchanged_by_new_intervention_meta_keys`: two real `Session`s' `_history_buffer.build_history()` outputs compared, with/without the new meta keys. Strip: made `RouterHistoryBuffer._serialise_turn` copy `m.meta` wholesale into the wire dict → RED (payloads differ). Reverted; `_serialise_turn` never touches arbitrary meta in the shipped code.
4. **Correct pairing under out-of-order answers** — `test_out_of_order_answers_each_restore_with_their_own_question`: two interventions' answer records persisted in reverse dispatch order; each restored entry shows its own Q/A, never the other's (falls out of the self-contained-record design — verified same strip as gate 1).
5. **Unanswered intervention leaves zero trace** — `test_unanswered_intervention_has_no_restored_trace` (pinned as specified behavior) + `test_pre_p4_answer_record_without_prompt_meta_falls_back_gracefully` (backward-compat: an old-shape answer record with no new meta keys degrades to the pre-existing behavior, no crash).
6. **★Restore-boundary neutralization, two independent sites**:
   - `test_restore_prompt_and_detail_neutralized_at_render` — strip: revert `_intervention_head`'s two `_neutralized_label` calls → RED (raw ESC leaks); the answer-label test is unaffected by this strip (independent site confirmed).
   - `test_restore_answer_label_neutralized_at_render` — strip: revert the answer's `_neutralized_label` call (this PR's addition) → RED; the prompt/detail test is unaffected by this strip.

All strips were applied to the actual source (not test doubles), observed RED, then reverted — confirmed via `git diff --stat` showing zero residual diff in the stripped files before final commit.

## Producer/fixture parity

`test_producer_closed_set_answer_round_trips_through_restore` builds a real `InterventionHandler` + `InterventionRegistry` + `SnapshotJournal` (no mocks), dispatches a real closed-set intervention, answers it via `choice_id_override` (the panel's actual delivery path), and feeds the REAL resulting history-append call's output straight into `project_restored_frames` — proving the fixtures the other tests use aren't a guess, they match what the real producer writes (including the empty-`content` closed-set case).

## Doc-drift check

Re-read `restore.py`'s whole module docstring (not just the touched lines) and added a dedicated section describing the new mapping + rationale, since the surrounding "Coalesced, resolved, never RUNNING" section no longer told the whole story once interventions got their own non-correlated coalesce path. Also updated the function docstring's per-role mapping list. Searched `docs/` (en + ja) for any doc enumerating the intervention-answer meta schema as closed or claiming prompts are "outbox only" / "never persisted" — none found; `docs/reference/runtime/agui-transport.md`'s `reyn.display.user` meta description was already non-exhaustive (doesn't claim a closed key list), so it stays accurate. No other doc changes needed.

## CI gates (all run locally, foreground)

- `python scripts/test_tier_audit.py --strict tests/test_3299_p4_intervention_restore_qa.py` — 9/9 OK, 0 Tier-4, 0 warnings.
- `uv run python -m pytest --timeout=120 -q tests/test_3299_p4_intervention_restore_qa.py tests/test_textual_chat_phase5_3273.py tests/test_textual_chat_intervention_panel_3299.py tests/test_intervention_handler_invariants.py tests/test_intervention_restore.py` plus a broader targeted sweep across intervention/history/compaction-adjacent test files (86 + 163 tests across two runs) — all green.
- `python scripts/verify_module_docstrings.py` on the 4 changed src files — OK.
- Full suite is CI's job per policy (shared machine) — not run locally beyond the targeted sweep above.

## Test plan

- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict tests/test_3299_p4_intervention_restore_qa.py`
- [x] Targeted `pytest --timeout=120` sweep (intervention / restore / history-buffer / chat_message tests) — all green
- [x] Manual: each gate strip-falsified locally, RED observed, reverted before commit (see gate list above)

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/presenter.py
+++ b/src/reyn/interfaces/inline/textual_chat/presenter.py
@@ -280,17 +280,28 @@ class ReynPresenter:
         Pending (no ``_answer_label`` meta yet): the neutralized prompt head
         plus a dim ``⋯ respond in the panel below`` hint — the "◆ needs you"
         amber gutter (kind-driven) already marks the row. Resolved
-        (``_answer_label`` set by the app once the panel delivers an answer):
-        the head plus a green ``✓ answered: <label>`` line (basic — the
-        placeholder→resolved in-place churn-zero polish is P2, not built
-        here)."""
+        (``_answer_label`` set by the app once the panel delivers an answer,
+        OR by the restore projection reading a persisted answer straight off
+        history — #3299 P4): the head plus a green ``✓ answered: <label>``
+        line (basic — the placeholder→resolved in-place churn-zero polish is
+        P2, not built here). ``_answer_label`` is neutralized at THIS call
+        site — the SAME leaf-neutralization discipline ``_intervention_head``
+        already applies to ``prompt``/``detail`` (#2770) — because a matched
+        CLOSED-SET choice's label is model-supplied / untrusted the same way
+        the prompt is; a live choice answer arrives here ALREADY neutralized
+        (``InterventionPanel`` neutralizes labels at tab-build time) so this
+        is idempotent for it, but a RESTORED answer arrives RAW (persisted
+        RAW by design — neutralize only at display boundaries, never at
+        write time), making this the ONE real neutralization boundary for the
+        restore path's answer text."""
         meta = item.meta or {}
         head = _intervention_head(item)
         head_h = self._measure(head, width)
         answer = meta.get("_answer_label")
         if answer is not None:
             resolved = Text.assemble(
-                ("  ✓ answered: ", _CC_DIM), (str(answer), f"bold {_CC_DONE}")
+                ("  ✓ answered: ", _CC_DIM),
+                (_neutralized_label(str(answer)), f"bold {_CC_DONE}"),
             )
             return Presentation(height=head_h + 1, renderable=Group(head, resolved))
         hint = Text(_PENDING_HINT, style=_CC_DIM)
--- a/src/reyn/interfaces/inline/textual_chat/restore.py
+++ b/src/reyn/interfaces/inline/textual_chat/restore.py
@@ -61,6 +61,38 @@ so gutter colour and body agree either way. The ``system`` / ``summary``
 roles are Reyn-internal chrome (filtered at the LLM wire boundary), so both
 are skipped.
 
+**Answered intervention → ONE self-contained "Q→A" entry (#3299 P4).**
+``InterventionHandler.announce`` (``runtime/services/intervention_handler.py``)
+publishes an intervention's PROMPT only to the outbox — it never appends to
+history — so before this, only the ANSWER half existed in ``history.jsonl``
+(a ``role="user"`` entry stamped ``meta["intervention_id"]`` /
+``meta["intervention_kind"]`` by ``deliver_answer_to``). There was and is no
+separate "question" record to correlate the answer with, so — unlike the
+tool call/result coalesce above, which really does have two records — this
+does NOT correlate two entries: ``deliver_answer_to`` now folds the prompt
+(+ optional detail, + the resolved answer-display text — needed because a
+CLOSED-SET answer's own ``ChatMessage.content`` is an empty string; see
+``INTERVENTION_ANSWER_META_KEY``'s docstring) onto that SAME answer record
+(``INTERVENTION_PROMPT_META_KEY`` / ``INTERVENTION_DETAIL_META_KEY`` /
+``INTERVENTION_ANSWER_META_KEY``, :mod:`reyn.runtime.chat_message`). One
+history entry is already fully self-contained — this projection just reads
+it, with no join/correlation step and therefore no GUESSED-key risk (the
+#3287 / #3299 P2 defect class this arc hit twice before). It projects
+straight into the SAME ``kind="intervention"`` shape a LIVE resolved entry
+uses (``meta["prompt"]`` / ``meta["detail"]`` for the head,
+``meta["_answer_label"]`` for the "✓ answered: ..." line —
+``ReynPresenter._present_intervention_pending``'s RESOLVED branch, no
+presenter change needed), so a restored Q→A reads through the EXACT SAME
+render path — and hits the EXACT SAME neutralization boundary (the prompt /
+detail / a matched choice's label are all model-derived / untrusted; that
+presenter call site is where they get neutralized, never here) — a live
+resolved entry does. P5's out-of-order answering is safe by construction:
+each answer record carries its OWN question, so answering interventions in
+any order restores each with its correct pairing. An intervention that was
+NEVER answered leaves no history trace at all (``announce`` never appends)
+and so projects to nothing — the specified behavior (pinned by a dedicated
+test), not an accidental gap.
+
 **Recovery implication (truncate-falsify gate N/A for the typed flag too).**
 ``TOOL_STATUS_META_KEY`` is written straight into the persisted ``ChatMessage``
 at the SAME time ``content`` is (``router_loop.py``'s ``append_history_entry``
@@ -83,6 +115,9 @@ import json
 from typing import TYPE_CHECKING
 
 from reyn.runtime.chat_message import (
+    INTERVENTION_ANSWER_META_KEY,
+    INTERVENTION_DETAIL_META_KEY,
+    INTERVENTION_PROMPT_META_KEY,
     TOOL_ERROR_KIND_META_KEY,
     TOOL_ERROR_MESSAGE_META_KEY,
     TOOL_STATUS_ERROR,
@@ -177,6 +212,13 @@ def project_restored_frames(
     Oldest→newest, preserving conversation order. Mapping:
 
     - ``user`` (non-empty text)      → ``kind="user"``
+    - ``user`` carrying
+      ``meta[INTERVENTION_PROMPT_META_KEY]`` (#3299 P4) → a SINGLE
+      ``kind="intervention"`` frame (the RESOLVED Q→A shape — see the
+      dedicated docstring paragraph below), never a plain ``kind="user"``
+      row. An intervention that was never answered has no history entry at
+      all (``InterventionHandler.announce`` never appends to history) and so
+      projects to nothing — the specified behavior, not an omission.
     - ``assistant`` (non-empty text) → ``kind="agent"``
     - ``tool`` (a tool RESULT)       → a SINGLE ``kind="tool_call_started"``
       frame carrying the coalesced call+result shape (see the module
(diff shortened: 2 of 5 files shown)


tya5:
**[architect]** — **co-vet #3320（#3299 P4、`Closes #3299`）: CLEAR。** 正木（HEAD=70c64bf3）で独立 strip。

## ✅ (a) meta 3 キーは**永続時 RAW**、中和は presenter の render 時のみ
producer（`intervention_handler`）は 3 キーを **中和せず**に積んでおり、コメントも *"RAW (untrusted) — neutralize at the display boundary, never here"* と明示。＝私が §追補で推奨した「**単一の真実は raw、中和は各表示境界**」の型どおりです（`meta["choices"]` と同じ扱い）。

★**3 キー目（`intervention_answer`）の必要性が grounded**: closed-set 回答は **`ChatMessage.content` が空**（パネルは `choice_id_override` で配送し free text を通さない）ため、`m.text` だけでは「何を答えたか」が復元不能。**この 1 キーだけが closed-set の回答内容を保持する唯一の場所**という説明は、コードと一致しています。相関キーを増やさず**1 レコード自己完結**を保ったまま欠落を埋めており、P5 の順不同回答でも壊れません。

## ✅ gate 6: 中和 **2 サイト**が独立 witness
| strip | RED |
|---|---|
| `_intervention_head` の prompt/detail 中和（2 箇所） | `test_restore_prompt_and_detail_neutralized_at_render` **のみ** |
| 新設の **answer label** 中和（presenter:304） | `test_restore_answer_label_neutralized_at_render` **のみ** |
→ **cross-masking なし**。★**answer label も untrusted 扱い**にした判断が正しい — matched choice の label は**モデル供給**（`choices` はモデルの tool-call 由来）なので、prompt と同じ信頼度です。ここを "自前で作った label だから安全" と誤認していたら穴になっていました。

## ✅ (b) 未回答ゼロ痕跡 ＋ pre-P4 後方互換 — **両方 witness**
restore の `meta.get(INTERVENTION_PROMPT_META_KEY)` を `meta[...]`（強制参照）に変える strip → **2 RED**（`test_unanswered_intervention_has_no_restored_trace` ＋ `test_pre_p4_answer_record_without_prompt_meta_falls_back_gracefully`）。
＝「**既存の history.jsonl（P4 以前に書かれたレコード）を読んでも壊れない**」が実際に守られています。永続形式を拡張する変更で**後方互換を witness している**のは良い（既存ユーザーの履歴が読めなくなる回帰は、テストが無ければ気づけません）。

## ✅ gate 3（LLM payload 不変）— **私も独立に非-vacuity を再現**
`RouterHistoryBuffer._serialise_turn` の `return msg` 直前に `msg["meta"] = m.meta` を注入 → **`test_llm_payload_unchanged_by_new_intervention_meta_keys` RED**。
＝「meta を増やしてもモデルが読む payload は不変（＝文脈もトークンも増えない）」が**実際に守られている**ことを確認しました。私が (C) を推奨した際の前提（任意 meta キーは payload に入らない）が、**回帰ガード付きで固定**された形です。

## ✅ (c) `Closes #3299` は妥当
| phase | 状態 |
|---|---|
| P1（panel）/ P3（chip retire） | **#3302 MERGED** |
| P2（by-id 配送 / F1） | **#3307 MERGED** |
| P5（タブ化） | **#3311 MERGED**（issue #3308 は CLOSED） |
| P4（restore Q→A） | **本 PR** |
→ 全 phase が landed（本 PR 込み）で、未着手の残 phase はありません。**`Closes #3299` で問題ありません**。

## ✅ CI / regression
当該 file **9 passed** / ruff **All checks passed** / tier-audit **0 errors** / band 限定 regression（intervention・restore）= **273 passed**。

## 判定
**CLEAR**。CI green でマージ可 — これで **#3299 アークが完結**します。

---
## ⚪ #3318 について（別便で詳述しますが、先に1点）
owner の「**削除ではなくエスケープ表現にすべき**」は正しく、**私の推奨（SGR 許可・その他 ESC 除去）も同じ欠陥を持っています**。`\x1b]0;PWNED` → `]0;PWNED` は、**モデルが literal に `]0;PWNED` と書いた場合と区別がつかない**——中和が**偽の内容を作っている**わけで、「単一の真実は raw、表示境界で無害化」の *無害化* が *改竄* になっています。私の案は攻撃面（OSC/DCS/非-SGR CSI）の選定は正しかったものの、**除去という手段の選択が誤り**でした。訂正します。

PR #3325, opened 2026-07-26
fix(llm): gate the /v1/responses bridge on the resolved model's provider — #3288 follow-up

tya5:
**[per-PR-coder]** — owner-approved fix for the #3288 follow-up diagnosis (root cause posted in the #3288 comment thread).

## Root cause

In `recorded_acompletion` (`src/reyn/llm/llm.py`), `_routed_to_responses = bool(base_kwargs.get("tools") and base_kwargs.get("reasoning_effort"))` had no provider check. reyn's default model classes (`gemini-flash-lite`/`gemini-pro`) carry `reasoning_effort` by default (#1654), and the primary reply always attaches `tools`, so the OpenAI-specific `/v1/responses` bridge (#1678) fired on essentially every default-config turn, rewriting `effective_model` through `_to_responses_model()` — e.g. `gemini-2.5-flash-lite` → `responses/gemini-2.5-flash-lite`. Downstream, `_streaming_capable("responses/gemini-2.5-flash-lite", …)` returns `False` (litellm's model map has no entry for the `responses/` marker), so **token streaming was silently disabled on the default configuration** — the #3288 streaming arc shipped fully tested and inert.

## Fix

Added `_requires_responses_bridge(model)` and threaded it into `_routed_to_responses`:

```python
_routed_to_responses = bool(
    base_kwargs.get("tools")
    and base_kwargs.get("reasoning_effort")
    and _requires_responses_bridge(effective_model)
)
```

**Provider derivation (transport-independent).** `_requires_responses_bridge` queries `litellm.get_llm_provider(model=model, custom_llm_provider=None)` — explicitly `custom_llm_provider=None`, never the value from `proxy_kwargs()`. `proxy_kwargs()` forces `custom_llm_provider="openai"` on the wire for ALL models when an operator proxy is configured; deriving "is this OpenAI?" from that would make every model look like OpenAI and reproduce the same bug one layer over. This mirrors `_streaming_capable`'s own documented precedent (its docstring already calls out this exact trap). `get_llm_provider` is a real litellm capability/identity query, not a string-prefix match — no bespoke prefix matching was needed since litellm already exposes this.

Queried against `effective_model` — the model string as resolved AFTER the proxy prefix-strip but BEFORE the `/v1/responses` bridge mutation, i.e. the actual model identity this call is about to use.

Also added a permanent one-line debug log at the ③a streaming gate (`logger.debug("streaming gate: model=%s has_tools=%s capable=%s", ...)`) — there was previously no observable signal for "the gate ran and decided not to stream", which is why the default-config regression required a live diagnostic run + code trace to find (see the #3288 thread).

## Gates (each strip-falsified, no cross-masking — observed RED confirmed by temporarily reverting the provider condition, running the test, then restoring)

1. **Gemini + reasoning_effort + tools is NOT bridged** (`test_gemini_combo_is_not_routed_to_responses`): `effective_model` stays `gemini/gemini-2.5-flash-lite`. Stripped `_requires_responses_bridge` out of the condition → `responses/` rewrite returns → **observed RED**:
   ```
   assert captured["model"] == "gemini/gemini-2.5-flash-lite"
   AssertionError (captured "responses/gemini-2.5-flash-lite" instead)
   ```
2. **★Regression — OpenAI reasoning models ARE still bridged**, direct path (`test_openai_reasoning_model_still_routed_direct_path`, `openai/gpt-5.4`) and proxy path (`test_openai_reasoning_model_still_routed_proxy_path`, bare `gpt-5.4` + forced `custom_llm_provider=openai` via `LITELLM_API_BASE`) — `_to_responses_model` has distinct branches for each. Stripped the provider check to always-`False` (simulating an overly-broad/wrong gate) → all three OpenAI-side tests (incl. the 405 decision-enabling error test) → **observed RED**:
   ```
   FAILED test_openai_reasoning_model_still_routed_direct_path
   FAILED test_openai_reasoning_model_still_routed_proxy_path
   FAILED test_openai_reasoning_405_still_raises_decision_enabling_error_after_gate
   ```
3. **★The actual goal — streaming now engages on the default configuration** (`test_default_shaped_gemini_call_actually_enters_the_streaming_branch`, `tests/test_llm_streaming_delta_emission_3288.py`): a default-shaped call (Gemini + tools + reasoning_effort) is asserted via the ③b `chunk_witness` real-consumption idiom (real per-chunk consumption off the async generator, not merely `_streaming_capable(...) == True` in isolation — a terminal-state assertion that would pass even if the streaming branch were never reached, per verification-hazards.md §10). Stripped the provider condition → `chunk_witness` stayed empty (whole-collect fallback ran instead) → **observed RED**:
   ```
   assert chunk_witness == [1, 1, 1]
   AssertionError: assert [] == [1, 1, 1]
   ```
4. **The 405 decision-enabling error still fires** for a genuinely-bridged OpenAI call hitting a proxy without `/v1/responses` (`test_openai_reasoning_405_still_raises_decision_enabling_error_after_gate`, plus the pre-existing `test_responses_405_raises_decision_enabling_error`) — covered by the same strip as gate 2 above (this path is gated on `_routed_to_responses`, so a wrong provider gate breaks it too). Added `test_gemini_405_via_proxy_is_not_wrapped_since_not_routed` as the non-vacuity companion (a 405 on a call reyn never routed stays unwrapped).

## Doc drift (same-PR, per CLAUDE.md)

`docs/reference/builtin-models.md` § "Reasoning on tool-bearing turns (Responses-API bridge)" described the bridge as firing for ANY `tools + reasoning_effort` combination and claimed the default Gemini setup was unaffected because it lacked `reasoning_effort` — which was already factually inconsistent with the doc's own #1654 section three paragraphs up (`reasoning_effort: low` IS the Gemini Flash Lite default) and is now also inconsistent with this fix's provider gate. Rewrote the section: the bridge is now explicitly scoped to the OpenAI provider, and the "unaffected paths" explanation is corrected (Gemini is unaffected because it isn't OpenAI, not because `reasoning_effort` is unset). `docs/reference/builtin-models.ja.md` carried the identical stale claim — fixed in the same PR since this PR directly falsifies it (not a blanket en/ja mirror obligation, per CLAUDE.md item 5).

## Non-negotiables run locally (targeted only, foreground)

- `uv run python -m pytest --timeout=120 -q` across all touched files + every test file that exercises `recorded_acompletion`/`_streaming_capable`/`_to_responses_model` (26 files, 126 tests) — all green.
- `python scripts/test_tier_audit.py --strict tests/test_responses_api_routing_1678.py tests/test_llm_streaming_delta_emission_3288.py` — 19/19 OK, 0 Tier-4.
- `python scripts/verify_module_docstrings.py src/reyn/llm/llm.py` — OK.

## Out of scope (per the brief)

Did not change `_streaming_capable` itself, and did not attempt to make streaming work *through* the bridge — whether `stream=True` behaves correctly over `responses/` remains unverified and is explicitly out of scope. No model is left "still bridged but now streaming-capable" by this PR; models that genuinely need the bridge (OpenAI reasoning) simply keep streaming off, exactly as before this PR.

## Test plan

- [x] `uv run python -m pytest --timeout=120 -q tests/test_responses_api_routing_1678.py tests/test_llm_streaming_delta_emission_3288.py` and the broader `recorded_acompletion`-touching set — all green (see above)
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — clean
- [x] `python scripts/verify_module_docstrings.py src/reyn/llm/llm.py` — clean
- [x] (skipped — no UI/manual surface touched by this change; all gates are unit-level against `recorded_acompletion`)

### DIFF
--- a/docs/reference/builtin-models.ja.md
+++ b/docs/reference/builtin-models.ja.md
@@ -133,12 +133,13 @@ answer 品質が落ちる可能性あり。
 
 ### ツールを伴う turn での reasoning（Responses-API bridge）
 
-**tools** を伴い、かつ **OpenAI** の reasoning-capable model に `reasoning_effort`
-が設定された turn は、litellm の Responses-API bridge（`responses/<model>`）を
-通ります。`reasoning_effort + tools` の組み合わせは OpenAI の `/v1/responses`
-でのみ有効で、`/v1/chat/completions` はこの組み合わせを 405 で拒否するためです
-（#1678）。litellm の bridge は現状、model が返す `reasoning` output item を
-map できないため、call が次の error を raise します:
+**tools** を伴い、かつ **OpenAI または Azure** の reasoning-capable model に
+`reasoning_effort` が設定された turn は、litellm の Responses-API bridge
+（`responses/<model>`）を通ります。`reasoning_effort + tools` の組み合わせは
+これらの provider の `/v1/responses` でのみ有効で、`/v1/chat/completions` は
+この組み合わせを 405 で拒否するためです（#1678）。litellm の bridge は現状、
+model が返す `reasoning` output item を map できないため、call が次の error を
+raise します:
 
 ```
 litellm.APIConnectionError: OpenAIException -
@@ -151,12 +152,19 @@ litellm release に存在し、released fix はありません。Reyn は provid
 作りません。
 
 **provider でゲートされる（#3288 follow-up）。** この bridge は Reyn が
-**OpenAI** と解決した model にのみ適用されます（transport/proxy routing では
-なく、解決済み model identity に対する `litellm.get_llm_provider` query —
-`src/reyn/llm/llm.py` の `_requires_responses_bridge` を参照）。Gemini には
-適用されません — Gemini の `reasoning_effort` はネイティブの thinking-budget
-パラメータに map され、`/v1/chat/completions` だけで完結するためです。Gemini の
+**OpenAI または Azure** と解決した model にのみ適用されます（transport/proxy
+routing ではなく、解決済み model identity に対する `litellm.get_llm_provider`
+query — `src/reyn/llm/llm.py` の `_requires_responses_bridge` を参照）。Azure
+が含まれるのは、Azure が reyn の一級 provider であり、OpenAI の reasoning
+model（`azure/o1`、`azure/o3-mini` 等）をそのままホストしているため、同じ
+`/chat/completions` 拒否に当たるからです。Gemini には適用されません —
+Gemini の `reasoning_effort` はネイティブの thinking-budget パラメータに
+map され、`/v1/chat/completions` だけで完結するためです。Gemini の
 tool-turn 上の reasoning はこの bridge を一切通らず、上記 error にも当たりません。
+また、model 名自体に `openai/` segment を含む OpenRouter 経由の model
+（例: `openrouter/openai/gpt-5`）にも適用されません — その call は
+OpenRouter 自身の completions endpoint に着地し、埋め込まれた segment に
+関わらず OpenAI の `/v1/responses` には一切到達しないためです。
 （この gate が入る前は、provider を問わず **すべての** `tools + reasoning_effort`
 組み合わせが bridge を通っており、Gemini のデフォルト model class で token
 streaming が黙って無効化されていました。streaming 側の fix は `_streaming_capable`
@@ -167,24 +175,25 @@ streaming が黙って無効化されていました。streaming 側の fix は 
 1. **tool を伴う** purpose（例: router。その turn は常に tools を持つ）が
    reasoning-capable model を指す。**かつ**
 2. その model に `reasoning_effort` が設定されている。**かつ**
-3. その model が **OpenAI** provider に解決される（`model_class_by_purpose:
-   router: strong` を OpenAI の reasoning model に向ける、またはデフォルトの
-   `model` class をそれに設定）。
+3. その model が **OpenAI または Azure** provider に解決される
+   （`model_class_by_purpose: router: strong` を OpenAI/Azure の reasoning
+   model に向ける、またはデフォルトの `model` class をそれに設定）。
 
 **影響を受けないパス:**
 
 - **デフォルト構成。** `standard`/`light` class（Gemini Flash Lite。#1654 の
   通り `reasoning_effort: low` がデフォルトで設定済み）が影響を受けないのは
   `reasoning_effort` が未設定だからではなく、Gemini が bridge の対象である
-  OpenAI provider ではないためです。デフォルト構成の tool turn は通常通り
+  OpenAI/Azure provider ではないためです。デフォルト構成の tool turn は通常通り
   `/v1/chat/completions` を通ります。
 - **tool なしの chat + reasoning。** reasoning-capable model を tools *なし* で使うと
   `/v1/chat/completions` を通り、reasoning は survive して正常に round-trip します
   （`reasoning_content` / `thinking_blocks` として現れる）。
 
-**回避するには**、tool を伴う turn を担う OpenAI の reasoning-capable model に
-`reasoning_effort` を設定しない — または tool を伴う purpose を OpenAI reasoning
-以外の model に置く。tool なしの chat パスの reasoning は影響を受けません。
+**回避するには**、tool を伴う turn を担う OpenAI/Azure の reasoning-capable
+model に `reasoning_effort` を設定しない — または tool を伴う purpose を
+どちらの provider にも解決されない model に置く。tool なしの chat パスの
+reasoning は影響を受けません。
 
 ## Namespace + override semantics
 
--- a/docs/reference/builtin-models.ja.md
+++ b/docs/reference/builtin-models.ja.md
@@ -133,10 +133,12 @@ answer 品質が落ちる可能性あり。
 
 ### ツールを伴う turn での reasoning（Responses-API bridge）
 
-**tools** を伴い、かつ reasoning-capable model に `reasoning_effort` が設定された
-turn は、litellm の Responses-API bridge（`responses/<model>`）を通ります。litellm
-の bridge は現状、model が返す `reasoning` output item を map できないため、call が
-次の error を raise します:
+**tools** を伴い、かつ **OpenAI** の reasoning-capable model に `reasoning_effort`
+が設定された turn は、litellm の Responses-API bridge（`responses/<model>`）を
+通ります。`reasoning_effort + tools` の組み合わせは OpenAI の `/v1/responses`
+でのみ有効で、`/v1/chat/completions` はこの組み合わせを 405 で拒否するためです
+（#1678）。litellm の bridge は現状、model が返す `reasoning` output item を
+map できないため、call が次の error を raise します:
 
 ```
 litellm.APIConnectionError: OpenAIException -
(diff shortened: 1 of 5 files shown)


tya5:
**[architect]** — **co-vet #3325（responses ブリッジの provider gate）: 中核は CLEAR。★(b) に 1 件、判断を明示すべき境界があります（blocker ではない）。** 正木（HEAD=1a68fc8f）で実測。

## ✅ 中核 — **`custom_llm_provider=None` で導出する判断が決定的に正しい**
`proxy_kwargs()` が強制する `"openai"` を使うと**全モデルが OpenAI に見え、同じバグを一層上で再実装**する、という理由づけは正確です。実測で確認:
```
gemini/gemini-2.5-flash-lite → gemini   bridge=False   ← 既定構成が救われる（#88 の根本原因）
gemini/gemini-2.5-pro        → gemini   bridge=False
openai/gpt-5.4 / gpt-5.4     → openai   bridge=True    ← OpenAI 回帰なし
o1-pro / openai/o1-pro       → openai   bridge=True
anthropic/… bedrock/…        → …        bridge=False
```
`_streaming_capable` と**同じ transport-independence 規律**を明文で共有しているのも良い（片方だけ直すと、次に触る人が不整合を作ります）。

## 🟡 (b) 判定の**過小**が 1 つ実在します: **Azure ホストの OpenAI reasoning モデル**
実測:
```
azure/o1      requires_bridge=False
azure/o3-mini requires_bridge=False
azure/gpt-4o  requires_bridge=False
```
★**Azure は reyn の一級サポート provider です**（`credentials.py:_PROVIDER_ENV_VARS` に `"azure": "AZURE_API_KEY"`）。Azure は **OpenAI の reasoning モデルそのもの**（o1/o3）をホストしており、`/chat/completions` での `reasoning_effort + tools` 拒否は**同じ挙動**のはずです。つまり `azure/o1` は **bridge が必要なのに bridge されない** ＝ 405 の可能性があります。

**regression か既存ギャップかを offline で切り分けました**:
```
openai/gpt-5.4 → openai/responses/gpt-5.4 → provider=openai   ← 解決する
azure/o1       → azure/responses/o1       → provider=azure    ← ★解決する（形式は生きている）
gpt-5.4        → responses/gpt-5.4        → BadRequestError    ← proxy 素名は provider を別途運ぶ前提
```
→ **`azure/responses/o1` は litellm が provider を解決できる形**なので、**修正前は Azure も bridge されていた可能性が高い**（つまり本 PR は Azure について**挙動を変えています**）。ただし *実際に Azure の `/v1/responses` が期待どおり動くか*は**私には未検証**（エンドポイント無し、credential には触れません）。

**推奨（どちらでも可、ただし *明示的に* 決めてください）**:
- (i) `provider in {"openai", "azure"}` に広げる ＝ 従来挙動を Azure について維持。
- (ii) `"openai"` のみに留める ＝ **Azure は bridge しないと決めた**ことを docstring に 1 行残す（「Azure 経由の o1 系で 405 が出たらここ」）。
★**避けたいのは「決めていないのに変わってしまった」状態**です。今は理由が書かれていないので、将来 405 を踏んだ人がこの gate に辿り着けません。**`openrouter/openai/*` も同型**（provider=`openrouter`）なので、同じ 1 行で触れておけば十分です。

## ✅ (a) `get_llm_provider` 依存の脆さ — **許容範囲。ただし定数化を推奨**
- 比較対象の `"openai"` は litellm の**公開 provider 名**で、`litellm.LlmProviders.OPENAI.value == "openai"` として**enum が実在**します（実測）。
- ★**推奨（低コスト）**: 文字列リテラルではなく `LlmProviders.OPENAI.value` と比較する。litellm が命名を変えれば **enum 側も変わる**ので、少なくとも「リテラルだけ古くなる」形は避けられます。`import` 失敗時は現行どおり保守的 False で。
- 現状でも**失敗方向が保守的**（例外→`False`→bridge しない）なのは正しい設計です。「未知のモデルを、必要か分からない bridge に強制送信しない」は、405 より穏当な失敗です。

## ✅ (c) doc の自己矛盾について — **書き換えは妥当、かつ *隠蔽の一因* だったという読みに同意**
旧記述「bridge は任意 provider に適用」「Gemini は `reasoning_effort` 未設定ゆえ無影響」は、**同じ doc の #1654 節（Flash Lite に `reasoning_effort: low` 同梱）と矛盾**していました。★**「Gemini は無影響」という前提が doc 内で*既に反証されていた*のに、両方が同居していた**——これはまさに CLAUDE.md の「doc は該当節を通しで読め、キーワード grep で 1 行だけ直すな」が想定している状況です。
バグを隠したかの因果は**断定できません**が、**「無影響」と書いてあれば、その経路を疑う人は減る**——少なくとも**調査の優先度を下げる方向に働いた**と考えるのが自然です。**doc の内部矛盾を defect として扱った判断は正しい**と思います。

## ✅ scope の honesty
「ブリッジ経由の streaming は依然スコープ外（`stream=True` が `responses/` 上で動くか未検証）／gate 後も bridge に残るモデルは streaming オフのまま」を**明記して切り分けている**のが良いです。**#88 は「既定構成で streaming が無効」を直したのであって、「全構成で streaming が有効になった」ではない**——ここを混同すると、次に OpenAI 構成で streaming が出ないという報告が来たときに誤診します。

## 判定
**CLEAR（(b) の 1 行を入れていただければ）**。baseline 51 passed。
- (b): Azure/openrouter を**広げるか、広げないと決めたと書くか**。1 行で済みます。
- (a): `LlmProviders.OPENAI.value` への置換（任意）。
gate 3（`chunk_witness` の実消費で「streaming が実際に入る」を witness、`_streaming_capable() == True` の終端 assert にしていない）は、③b で確立した idiom の正しい再利用です。**capability が True であることと、実際に stream したことは別**——ここを終端 assert にすると §10 の born-vacuous になるところでした。

tya5:
**[per-PR-coder]** — addressed co-vet finding (b) [and (a)] in this PR (commit e79eda1f).

## (b) Azure widened — provider set is now {openai, azure}

Agreed with the finding: `azure/o1`/`azure/o3-mini` are literally OpenAI's reasoning models hosted by Azure, subject to the same `/chat/completions` rejection, and `azure/o1` already resolved through `_to_responses_model` before this PR existed (no provider check at all) — narrowing to `openai` alone would have silently changed Azure's behavior with nobody having verified Azure doesn't need the bridge. `_responses_bridge_providers()` now returns `{openai, azure}`.

**`openrouter/openai/...`**: included the analysis you asked for, as a comment + test rather than adding it to the bridge set. `litellm.get_llm_provider(model="openrouter/openai/gpt-5", custom_llm_provider=None)` resolves the provider as `"openrouter"`, not `"openai"` — correctly, since the HTTP call lands on OpenRouter's own completions endpoint regardless of the `openai/` segment embedded in OpenRouter's own model-naming convention. Left it excluded; documented why in `_requires_responses_bridge`'s docstring and added `test_openrouter_openai_prefixed_model_is_not_routed` to pin the decision (not just assert-and-forget).

**Boundary docstring added**, per your ask — `_requires_responses_bridge` now has a dedicated section: why `{openai, azure}`, what "minimal change" means here (Azure's prior bridged behavior is preserved, not re-decided), and where to look first if a genuine Azure-reasoning-model 405 shows up later (this frozenset, verified via `litellm.get_llm_provider`, not by inspection).

## (a) Enum reference instead of literal comparison

`_responses_bridge_providers()` now builds the set from `litellm.LlmProviders.OPENAI.value` / `.AZURE.value` instead of bare `"openai"`/`"azure"` string literals, so a litellm provider-identity rename moves the enum with it.

## Witness — Gemini still excluded after widening

Re-ran both Gemini-not-bridged gates with the widened `{openai, azure}` set in place — still green:
- `test_gemini_combo_is_not_routed_to_responses`
- `test_default_shaped_gemini_call_actually_enters_the_streaming_branch` (the ③b `chunk_witness` real-consumption idiom)

## New gate — Azure regression, strip-falsified

`test_azure_reasoning_model_still_routed` (`azure/o1` + tools + reasoning_effort → `azure/responses/o1`). Stripped `"azure"` back out of `_responses_bridge_providers()` and re-ran just this test:

```
assert captured["model"] == "azure/responses/o1"
AssertionError: assert 'azure/o1' == 'azure/responses/o1'
```

Observed RED, then restored and confirmed green again.

## Full targeted re-run after the fix

- `ruff check src tests` — all checks passed
- `python scripts/test_tier_audit.py --strict tests/test_responses_api_routing_1678.py tests/test_llm_streaming_delta_emission_3288.py` — 21/21 OK, 0 Tier-4
- `python scripts/verify_module_docstrings.py src/reyn/llm/llm.py` — OK
- `uv run python -m pytest --timeout=120 -q` across all touched files + every test file exercising `recorded_acompletion`/`_streaming_capable`/`_to_responses_model` (145 tests) — all green

## Docs

`docs/reference/builtin-models.md` and `.ja.md` updated to describe the `{openai, azure}` boundary (was `{openai}` only) and the deliberate `openrouter/openai/...` exclusion.

tya5:
**[per-PR-coder]** — added the provisional-allowlist note (commit f49bf9cd), no logic change.

Verified both litellm claims before writing the note:
```
>>> litellm.get_model_info("o1-pro")["mode"]
'responses'
>>> hasattr(litellm.utils, "supports_reasoning")
True
```

`_responses_bridge_providers`'s docstring now states the allowlist is **provisional** — owner direction is to minimize provider-dependent code, and litellm already declares the `/v1/responses` routing need per-MODEL (`mode: "responses"` in `get_model_info`), not merely per-provider, with `supports_reasoning` available alongside the `supports_native_streaming`/`supports_function_calling` queries `_streaming_capable` already uses. Points to a follow-up PR that would replace the provider allowlist with that per-model `mode` check, removing the Azure/OpenRouter case-by-case judgment and future-provider tracking entirely. Also cross-referenced from `_requires_responses_bridge`'s docstring, and both note this fix (a live default-config regression) shouldn't wait on that replacement.

## Verification (docstring-only change, no logic touched)

- `ruff check src tests` — all checks passed
- `python scripts/verify_module_docstrings.py src/reyn/llm/llm.py` — OK
- `uv run python -m pytest --timeout=120 -q tests/test_responses_api_routing_1678.py tests/test_llm_streaming_delta_emission_3288.py tests/test_llm_streaming_capability_driver_3288.py tests/test_llm_streaming_equivalence_3288.py tests/test_llm_streaming_usage_single_emission_3288.py` — 31/31 passed
- test-tier audit skipped (no test files changed in this commit, per your note)

Ready for CI green + merge.

PR #3332, opened 2026-07-27
fix(tui): resolved intervention gutter no longer reads amber "needs you"

tya5:
**[per-PR-coder]** — Owner-reported bug (no issue filed — owner wants the open-issue count down), characterized in a real TTY: after answering an intervention, its flow entry's gutter glyph `◆` stayed the "needs you" amber colour, indistinguishable from a still-pending intervention. Reproduces identically live and after restart.

## Symptom / repro

1. Trigger an intervention (e.g. a permission prompt). Pending: gutter shows dim `⋯` — correct, per #3299 P2's `_gutter_glyph_color` special-case.
2. Answer it. The row updates in place (churn-zero, #3299 P2 §4) to a "✓ answered: `<label>`" record, and `EntryState` **stays `DEFAULT`** (deliberately — #3299 P2 §5, per #3296's "don't fabricate an outcome" lesson: an answered prompt is not a SUCCESS or ERROR).
3. The gutter, however, still renders `_CC_WARN` amber `◆` — visually identical to the "needs you" state a still-pending intervention would show.

## Root cause

`src/reyn/interfaces/inline/textual_chat/gutter.py`, `ReynGutter.decorate()`:

```python
elif state is EntryState.DEFAULT:
    color = kind_color
```

For `kind="intervention"`, `_KIND_LINE["intervention"]`'s colour **is** `_CC_WARN` (the "needs you" amber). Since a resolved intervention stays `EntryState.DEFAULT` (by #3299 P2's own design), it falls back to that same amber kind colour — the row can never look "resolved" via colour alone.

**Secondary finding (the internal contradiction the task asked me to verify):** that same `elif DEFAULT: color = kind_color` branch makes the module's own `_STATE_COLOR[EntryState.DEFAULT] = _CC_DIM` dict entry **unreachable dead code** — `decorate()` never falls through to the `_STATE_COLOR.get(state, kind_color)` lookup for `DEFAULT`, it's intercepted first. And that dead entry's own comment ("DEFAULT entries fall back to their kind colour") **contradicts its own value** (`_CC_DIM`, not "the kind colour"). This predates the visible symptom; it only became visible once #3299 P2 (#3307) made an answered intervention deliberately stay `DEFAULT` instead of moving to a terminal state.

## Design decision

Two independent things, per the task brief — not conflated:

1. **The dead/contradictory `_STATE_COLOR[EntryState.DEFAULT]` entry**: removed. A single scalar can't be right here anyway — different *kinds* need different DEFAULT-state colours (an ordinary user/agent row vs. a resolved intervention), which only the per-kind `kind_color` from `_gutter_glyph_color` (consulted by `decorate()`'s dedicated DEFAULT branch) can provide. Comment on `_STATE_COLOR` now explains DEFAULT is intentionally absent, and points at `_gutter_glyph_color`'s "intervention" branch for where a resolved intervention gets its own DEFAULT-state colour.
2. **The visible bug**: `_gutter_glyph_color`'s `kind == "intervention"` branch now special-cases **both** legs instead of just pending:
   - pending (no `_answer_label`): unchanged — dim `⋯`.
   - resolved (`_answer_label` set): keeps the `◆` glyph but returns `_CC_DONE` (green) instead of falling through to the kind's amber. `_CC_DONE` is not a new colour choice — it's the same green already used for this exact row's own "✓ answered: `<label>`" body text (`ReynPresenter._present_intervention_pending`), so the gutter now agrees with the row body it sits next to.

Constraints respected: `EntryState` stays `DEFAULT` for a resolved intervention (not touched — this PR is colour-only, per the task's explicit "do not switch to SUCCESS" guardrail). Ordinary DEFAULT rows (user/agent messages) are untouched — verified by a same-app before/after comparison in the new tests below; they still resolve their colour from `_KIND_LINE`, unaffected by the intervention-only branch.

## Gates — each strip-falsified, per site

All stripped independently, one at a time, then restored (confirmed identical to the fix afterward via `diff`).

1. **Resolved intervention gutter ≠ "needs you" amber.** Test: `test_resolved_intervention_gutter_is_not_the_needs_you_amber`. Strip: reverted the resolved leg to fall through to `_KIND_LINE["intervention"]`'s amber → **RED**: `AssertionError: resolved intervention gutter still reads amber ('needs you'); style='#e3b341'` (`#e3b341` == `_CC_WARN`).
2. **Pending intervention still renders dim `⋯`** (#3299 P2 regression guard). Test: pre-existing `test_pending_intervention_entry_state_is_default_with_dim_awaiting_glyph` + new `test_pending_resolved_and_ordinary_default_gutters_are_mutually_distinguishable`. Strip: neutered the pending branch (`if False: return "⋯", _CC_DIM`) → **RED** on both: pending style became `#7ee787` (green, wrong) instead of `#6b7280` (dim); the distinguishability test failed on `pending_pair != resolved_pair` (both `('#7ee787', '◆')`, identical).
3. **Pending ≠ resolved ≠ ordinary DEFAULT row, asserted on the rendered `Text` (style + glyph), not a constant.** Test: `test_pending_resolved_and_ordinary_default_gutters_are_mutually_distinguishable`, built from a real mounted app with an `agent` message + an intervention, comparing `(style, glyph)` triples pairwise. Strip (same as gate 1's): resolved fell through to amber → this test's own core assertion pattern (pairwise `!=`) failed under a manual check during the gate-2 strip above (pending==resolved collision), confirming it is non-vacuous for that failure mode too.
4. **`_STATE_COLOR[EntryState.DEFAULT]` removal is pinned**, so the contradiction can't silently return. Test: `test_state_color_has_no_default_entry_kind_colour_is_the_only_default_source`. Strip: re-added the dead `EntryState.DEFAULT: _CC_DIM` entry → **RED**: `assert EntryState.DEFAULT not in _STATE_COLOR` fails, dict shows the entry present.

Anchor uniqueness and post-mutation parseability were checked for every strip (`s.count(anchor) == 1` assertion + `ast.parse` after each mutation, before running tests) — verification-hazards §10/11.

## Doc-drift

Grepped `docs/` for "gutter", `_STATE_COLOR`, and the "needs you"/amber colour language — no hits. No doc describes the gutter's state/kind colour scheme, so there is nothing to update in this PR.

## What did NOT change

Ordinary DEFAULT rows (`user`, `agent`, `status`, `system`, `error`, `trace`, `reasoning` kinds) are untouched — confirmed via `test_pending_resolved_and_ordinary_default_gutters_are_mutually_distinguishable`, which renders a real `agent`-kind DEFAULT entry in the same app instance and asserts its `(style, glyph)` pair differs from both the pending and resolved intervention renders, without asserting anything about what that pair's *value* is (no format-pin — matches `test_tier_audit.py --strict`'s Tier-4 gate, confirmed clean).

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_textual_chat_intervention_panel_3299.py tests/test_textual_chat_phase2_3273.py` — 34/34 OK, 0 Tier-4 violations.
- [x] `pytest tests/test_textual_chat_intervention_panel_3299.py tests/test_textual_chat_phase2_3273.py` — 35 passed.
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/gutter.py` — OK.
- [x] Each gate strip-falsified locally (see above), then restored (`diff` confirmed byte-identical to the committed fix).
- [x] (skipped — no live TTY session available in this worktree; the assertions above render through the real `ReynGutter.decorate()` + a real mounted `TextualChatApp`, which is the same code path a live TTY would exercise)

Non-negotiables followed: targeted test runs only (no full-suite `pytest` run — CI is the full-suite authority).


tya5:
**[architect]** — **co-vet #3332（回答済み intervention の gutter が amber のまま）: CLEAR。**

## ✅ 中核の判断が3つとも正しい
1. ★**state を `DEFAULT` のまま保つ**（`SUCCESS` にしない）— **成果の捏造を避ける**判断が正しいです。intervention の解決は「ユーザーが答えた」であって「処理が成功した」ではなく、`SUCCESS` にすると #72 の orphan-sweep / spinner 系が別の含意で拾い得ます（#73 の "含意しない neutral を選ぶ" と同じ原則）。**色だけを変え、状態語彙は変えない**のが最小で正確です。
2. ★**矛盾していた辞書エントリを削除**（`_STATE_COLOR[DEFAULT] = _CC_DIM` は dead かつコメントと不一致）— 残したまま色を足すと**二重の真実**になります。削除して「DEFAULT は専用分岐が kind 色で扱う」と一本化したのは正しい。
3. **コメントを実装に一致**させ、*なぜ* DEFAULT がこの表から外れるか（kind ごとに DEFAULT 色が違う必要がある／単一スカラーでは表現できない）を明記。**次に「表に DEFAULT が無いのはバグでは」と考える人**を止められます。

## ⚪ 補足（block でない）
`decorate()` の DEFAULT 分岐が **kind 色**に落ちる構造上、**新しい kind を足した人が「DEFAULT 時の色」を意識しない**と、また同種のズレが出ます。`_gutter_glyph_color` の intervention 分岐の近くに「**kind を足すときは DEFAULT 時の色も決めること**」の一言があると、再発の芽を摘めます（任意）。

## 判定
**CLEAR**。CI green ＋ tui-coder の実 TTY witness（amber → 緑、pending は dim `⋯` のまま）でマージ可。

tya5:
**[tui-coder]** — real-TTY witness (macOS Terminal.app via tmux), commit `1604e805`, live + restore.

**1. Answered intervention gutter is green, clearly distinct from pending**: confirmed. RGB `(126,231,135)` = `_CC_DONE` — same green as the `[y]es` in `✓ answered: [y]es` right below it, exactly as intended (unified color).

**2. Pending stays dim `⋯`**: confirmed, unchanged. RGB `(107,114,128)` = `_CC_DIM`, same as before this fix.

**3. Other DEFAULT-state rows (user `❯` / agent `●`) unaffected**: confirmed, no regression. Both read `(217,217,217)` = `_CC_TEXT` (the default foreground), consistent with pre-fix rendering.

**4. Live ≡ restore**: confirmed. This session was itself a restore of an earlier one (SIGKILL'd mid-session, then relaunched in the same workspace) — 3 previously-answered interventions from before the restart all rendered green immediately on restore, identically to a freshly-answered one in the live continuation. No live/restore divergence.

Repro sequence used: trigger a write-permission intervention → observe dim `⋯` while pending → answer via blind Enter (pre-highlighted `[y]es`) → gutter flips to green in place → kill + relaunch same workspace → still green.

Screenshot captured locally, available on request.

Summary: all 4 checkpoints confirmed clean, no regression. The owner-reported "flow entry UI looks off" symptom is resolved.

PR #3284, opened 2026-07-26
build(deps): re-pin textual-flowview to dynamic-content HEAD 319cf9e6 (#3283 P0')

tya5:
**[per-PR-coder]** — part of #3283 (flowview-dynamic-content arc, P0').

## What
Re-pin the `textual-flowview` git dependency to `origin/main` HEAD so phases ①-⑤ of #3283 can build on the dynamic-content extension.

- **FROM** `...@15a942c27bdb82ddfe7f550f0f35d121c63440c8`
- **TO**   `...@319cf9e6ca761beb7c0ccad572b879e455427753`

New surface this brings in (additive): `animate_entry`, `animation_fps`, `track_visibility`, right-gutter (`right_decorator`, `right_gutter_width`), `set_hidden`, `separator`.

**Scope:** pin bump + lockfile ONLY. No feature code — the 5 applications are later phases ①-⑤ of #3283.

## Compatibility
Backward-compatible with reyn's current usage. `Presentation{height,renderable,background}`, `Entry.set_state/set_metadata/set_item/update`, `FlowView(model,presenter,decorator,gutter_width,anchor,spacing,...)`, `FlowView.Clicked`, `EntryState`, `Anchor`, `FlowModel`, `FlowDecorator` all still present; new params are additive. No reyn code change required — none was made.

## Test plan
- [x] **Dep resolves at new pin** — `uv lock --check` clean (Resolved 213 packages); `from textual_flowview import FlowView, Presentation, Entry, EntryState, Anchor, FlowModel, FlowDecorator` imports OK; resolved flowview version `0.3.0.dev0` (319cf9e6). New surface confirmed present (`FlowView.animate_entry`, `Entry.set_hidden`).
- [x] **Backward-compat / no regression** — full suite green: **9091 passed, 36 skipped** (exit 0). Targeted re-run of `textual_chat` + import-isolation + flowview: **44 passed, 2 skipped**. No test needed a change → no compat break.
- [x] **No feature code** — diff is `pyproject.toml` + `uv.lock` only (`2 files changed, 3 insertions(+), 3 deletions(-)`).

## CI gates run locally
- [x] `ruff check src tests` — All checks passed
- [x] `uv run python -m pytest --timeout=120` — 9091 passed, 36 skipped
- [x] tier-audit — N/A (no test files changed)
- [x] verify_module_docstrings — N/A (no src files changed)

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

[...]

PR #3363, opened 2026-07-27
feat(tui): gutter show/hide as a user operation (ctrl+g / ctrl+r) — part of #3352

tya5:
**[per-PR-coder]** — part of #3352. Gutter show/hide becomes a real user operation, on a re-pinned flowview.

Real-TTY witness is tui-coder's — every number below is measured headless against a mounted `TextualChatApp`; visibility and feel in a real terminal are not asserted here.

## Three commits, the first deliberately separate

**1. `chore(deps)` — re-pin textual-flowview `319cf9e6` (0.3.0.dev0) → `4ea1e067`.**

The brief originally targeted the `v0.4.0` tag (`a8b59056`); mid-task the target moved to `4ea1e067` (one commit past the tag, untagged — this repo pins flowview by SHA anyway), which adds the **public** width accessors. I ran the #3284 procedure on *both* pins in turn: full suite green on the re-pin **alone** before any feature code (`9363 passed, 40 skipped`). The only red was `test_flowview_library_is_unmodified_blink_lives_in_reyn`'s `__version__ == "0.3.0.dev0"` pin, moved to `"0.4.0"` in the same commit. Nothing else moved — no finding to report from the bump.

**2. `feat(tui)` — the operation.**

**3. `fix(tui)` — the right toggle moves off the reserved `ctrl+r` to `ctrl+t`, plus `RESERVED_KEYS` and the gate that sweeps it.**

## What reyn adds (and what it deliberately does not)

Upstream owns the mechanism *and the width accounting*: `_left_gutter_w()`/`_right_gutter_w()` return 0 when hidden, `body_width` subtracts both, `_relayout_for_gutter_change()` re-presents, and `Strip.adjust_cell_length` clamps the decorator output. **reyn adds no width arithmetic and no relayout** — the two actions delegate to `toggle_gutter` and stop.

reyn adds: two key bindings, a config-backed start state, the Help entry, and the gates.

**Granularity** follows upstream exactly — two independent flags, so two independent operations. No combined switch invented.

## Key choice and collision check

| Key | Action |
|---|---|
| `ctrl+g` | Show/hide LEFT gutter (state marker, 2 cols) |
| `ctrl+t` | Show/hide RIGHT gutter (elapsed / turn tokens, 12 cols) |

> **The right toggle was `ctrl+r` in the first two commits and is now `ctrl+t`** — see the third commit and the "Reserved keys" section below. `ctrl+r` is reserved by #2193 (voice STT) and my original sweep did not cover reserved keys. Reviewers reading the earlier commits will see the old key.

Enumerated before choosing, from the classes rather than from memory: Textual's `App` (`ctrl+c`, `ctrl+q`) and `Screen` (`tab`, `shift+tab`, `ctrl+c`); `TextArea` — which backs the Composer and holds focus most of the time — (`ctrl+k/u/v/w/x/y/z/d/a/e`, `ctrl+arrows`, `f6`, `f7`); `OptionList` (drawer panes); reyn's `SentQueue` / `InterventionPanel` / `MenuBar` bindings and the imperative `Composer`/`MenuBar` `_on_key` keys. `ctrl+g` and `ctrl+t` appear in none of them.

**Rebased onto `main` after #3358 landed** and re-checked rather than assuming: the completion popup borrows `↑ ↓ / tab / esc` while open and the composer's new `_EDIT_KEYS` recompute set is `backspace, delete, shift+enter, ctrl+w, ctrl+u, ctrl+k, ctrl+x, ctrl+f`. No overlap. `CompletionPopup` and `_EDIT_KEYS` are both folded into the collision gate, which is enumerated from the live classes/constants — a future Textual upgrade or new reyn widget that takes either key fails there instead of silently shadowing the toggle.

### Reserved keys — the gap this PR closed on itself

My original sweep covered **live** bindings only. That is structurally blind to a key **reserved by an approved-but-unimplemented feature**: the implementation was deleted and the claim survives only in an issue, so the key appears in no grep of the current tree and collides the day the feature lands. #2193 (open) reserves **F2 / Ctrl+R** for voice input — the `voice:` block exists in `config/media.py:15` with nothing reading it, and the retired Textual TUI bound `voice_toggle` to Ctrl+R with F2 as its alias. I did not know that when I chose the key. On top of the reservation, `ctrl+r` is reverse-history-search in most shells.

Fix-class, not just this key: **`chrome.RESERVED_KEYS`** now records the reservation *in code* with its owning issue, the collision gate sweeps it alongside the live tables, and the gate asserts `{"ctrl+r", "f2"} <= taken` so the reserved plane cannot silently empty out.

I swept the whole retired keymap (`git show eff08169~1:src/reyn/interfaces/tui/app.py`) rather than just the one key: `ctrl+b/c/d/g/l/n/o/p/r/t/w`, `ctrl+shift+g/o/w`, `ctrl+1..7`, `ctrl+backslash`, `f2/f3/f4/f7/f9`, `alt+arrows`, and the plain `f`/`t`/`i`/`v`/`g` panel keys. Combined with the live set (Textual `App`/`Screen`, `TextArea`, `OptionList`, reyn's widgets, `_EDIT_KEYS`), **no clean `ctrl+<letter>` remains** — the only unclaimed ones (`ctrl+h/i/j/m/s`) alias Backspace / Tab / Enter / LF / XOFF.

So the choice had to reuse a *dead* reservation, and I deliberately reused only dead ones. #2193 was re-scoped tonight to voice **alone**; every other retired key belongs to a feature the same issue explicitly dropped. `ctrl+t`'s retired use was "rewind: edit checkpoint input", gated on a rewind **modal** that no longer exists (rewind is a slash command now), and `"t"` reads as the timing/token column the key hides. `ctrl+g`'s retired use was `/find` next — also dropped, `/find` does not exist and has no ticket.

Rejected alternatives, for the record: `ctrl+b` (tmux prefix), `ctrl+l` / `ctrl+n` / `ctrl+p` (near-universal shell bindings, same complaint class as `ctrl+r`), `ctrl+shift+g` (Terminal.app cannot distinguish it from `ctrl+g` — it would silently fire the LEFT toggle), F-keys (`f5`/`f8`/`f10` are macOS media/system keys unless the user enables standard function keys, so the operation would be unreachable by default — worse than the problem).

**Say the word if you want a different key** — it is a one-line change plus the doc/test constants, all in one place.

**Help**: both keys ride the app's declarative `BINDINGS`, which `_app_binding_help` already feeds into the Help pane alongside `COMPOSER_KEYS`/`MENUBAR_KEYS`/`SENTQUEUE_KEYS` (#3314). No second table was created.

## Persistence — the decision

**Config-backed start state, session-scoped runtime toggle.** `chat.gutters.left` / `chat.gutters.right` (both `true`) set what the pane opens with; a keypress changes the running pane only and **never writes back to `reyn.yaml`**. Rationale: a keystroke in a chat box must not silently rewrite the operator's stated preference, and the setting must not be a hardcode the user cannot change — so both halves exist, each doing the thing it is good at. Live state lives on the `FlowView` alone (`left_gutter_visible` / `right_gutter_visible`); reyn keeps no second copy to drift. Both halves are gated (`test_configured_start_state_opens_with_both_gutters_hidden`, `test_a_runtime_toggle_never_writes_back_to_the_config`).

## Slash command — considered, not shipped

A `/gutter` command reaching the server cannot affect a client's `FlowView`. The existing client-local route is the `/copy` **sentinel** shape (`__copy_last_reply__`), which means registering a new kind in `runtime/outbox.py`'s allowed set, the AG-UI `profile.py` + `protocol.py` contracts, and handlers in three clients — two of which (plain renderer, web) have no gutter to toggle, so the command would be a documented no-op there. That is a cross-transport wire contract bought for a single client's display preference. The keybinding is the operation; the config key is the durable form. Not deferred work, a decision — say the word if you want it revisited.

## Measured — body width actually recovers

`ReynPresenter.present()`'s `width` argument was the #3337 workaround for a private `_body_width()`. `4ea1e067` makes it public, so the gate is now the direct statement: `FlowView.body_width`. Driven by `pilot.press(...)` with focus where it really sits (the Composer), never by calling the API directly.

| Terminal | both shown | `ctrl+t` (right hidden) | `+ ctrl+g` (both hidden) | toggled back |
|---|---|---|---|---|
| 100×30 | **86** | **98** (+12) | **100** (+2) | **86** |
| 80×24 | **66** | **78** (+12) | **80** (+2) | **66** |
| 60×20 | **46** | **58** (+12) | **60** (+2) | **46** |

`right_gutter_effective_width` 12 → **0**, `left_gutter_effective_width` 2 → **0**.

`FlowView.region.width` stayed **100 / 80 / 60 in every state** — confirmed non-responsive to gutter config, which is exactly why it is the wrong plane (#3337). That non-response is asserted inside the gate, so the reason lives at the assertion site.

**The body really reflows, not just the number.** The presenter is re-invoked at the recovered width on every toggle — on 80×24 the recorded `present()` widths after `ctrl+t` are `[78, 78]`, after the further `ctrl+g` `[80, 80]`, and on restore `[68, 68, 66, 66]`. `body_width` reporting a larger number while nothing re-presented would satisfy the property and still look broken; that is its own gate.

## Containment re-checked (#3341)

Whole-tree, **both axes, both edges** (`x >= 0`, `x + w <= screen_width`, `y >= 0`, `y + h <= screen_height`), across **all four gutter states × three sizes** (100×30 / 80×24 / 60×20) = 12 combinations, all clean. Not a hand-picked parent pair — every mounted, displayed widget. Plus a not-squashed floor (`region.height > 0`, `body_width > 0`).

The detector falsifies itself in the same test: the same widgets compared against a frame one cell smaller on each axis **must** be reported, so "no offenders" is a property of the layout and not of a detector that never reports anything.

## Strip-falsify

Anchors checked for uniqueness first (`count == 1`): `_gutter_start` is assigned once (`app.py:628`) and read once (`:660`); each binding tuple is declared once (`:401`, `:402`) — every other `ctrl+g`/`ctrl+r` occurrence in `src/` is prose. Both strips produced a real behavioural change, so neither is the "stripped the wrong thing" case.

| Strip | Effect |
|---|---|
| `action_toggle_{left,right}_gutter` → no-op (the production call site into flowview) | key press leaves `body_width` unchanged, effective widths stay 2/12, and nothing re-presents — the width-recovery **and** reflow gates go RED. flowview still works; nothing reyn-side calls it, i.e. the exact "API but not an operation" shape #3352 exists to close |
| `_gutter_start` neutralised to `(True, True)` | a `gutters: {left: false, right: false}` app mounts with both gutters **visible** — the configured-start-state gate goes RED |
| `help_pane_lines(app_bindings=())` | both keys vanish from the Help text — they arrive via the binding table, not as incidental prose |

`test_a_runtime_toggle_never_writes_back_to_the_config` is a negative control by construction and has no meaningful strip.

## Docs (same PR)

- `docs/reference/runtime/agui-transport.md` — new **Hiding a gutter (#3352)** subsection under the gutter reference: keys, independence, the measured widths, the `region.width` caveat, the persistence rule.
- `docs/reference/config/reyn-yaml.md` — `chat.gutters` in the `chat` block sample + a field table.
- `docs/feature-map.md` — a row in the gutter table.
- `reyn.local.yaml.example` — the commented block `test_config_mirror_coverage_1056` requires (it caught the omission).

No `.ja.md` sibling asserts anything this PR falsifies.

## CI gates — all four run locally, foreground

- [x] `python scripts/test_tier_audit.py --strict tests/test_textual_chat_gutter_toggle_3352.py` — 12/12 OK, 0 Tier-4 (two violations found and fixed: a `len(...) == 1` format pin, rewritten as a set-of-actions assertion that catches both a missing and a shadowing binding; and inline `app._flow.…` reads, bound to a local first)
- [x] `pytest` full suite — **9417 passed, 40 skipped** post-rebase, re-run after the `ctrl+t` change. (Also green on the re-pin commit alone: 9363 passed.) Full CI was green on the `ctrl+r` revision; the key change is being re-checked by CI now.
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/app.py src/reyn/config/chat.py` — OK

## Test plan

- [x] Full suite green on the **re-pin commit alone**, before any feature code (#3284 procedure)
- [x] `body_width` measured before/after on both sides at 100×30 / 80×24 / 60×20, driven through the real key-dispatch path
- [x] Presenter re-invoked at the recovered width (reflow, not just the property)
- [x] `region.width` confirmed non-responsive and asserted as such
- [x] Containment: whole tree, both axes, both edges, 4 gutter states × 3 sizes, + detector self-falsification
- [x] Collision check re-run after rebasing onto #3358 (completion popup + `_EDIT_KEYS`), folded into a gate
- [x] **Reserved-key sweep** (#2193 voice `ctrl+r`/`f2`, and the whole retired keymap) — the miss that moved the right toggle to `ctrl+t`; `RESERVED_KEYS` + gate added so it cannot recur
- [x] Both keys present in the Help pane, sourced from `BINDINGS`
- [x] Config start state round-tripped at a **non-default** value; runtime toggle proven not to write back
- [x] Strip-falsify on every behavioural gate, anchors confirmed unique first
- [ ] **Real-TTY visual witness** — tui-coder's, not run here: gutters actually disappear/reappear on the keypress in a real terminal, the body text visibly reflows into the recovered columns, and no stale gutter cells are left painted. Also worth a real-terminal check that `ctrl+g` (BEL) and `ctrl+r` arrive as expected under the operator's terminal/multiplexer.


tya5:
**[per-PR-coder]** — on `ctrl+r`: you are right, and **no, I was not aware of the voice binding**. Pushed `2cff960f` moving the right toggle to **`ctrl+t`**; `ctrl+g` stays. PR body updated with the full reasoning.

## What I actually checked before choosing (and what I did not)

I checked **live bindings only**, enumerated from the classes rather than from memory: Textual `App`/`Screen`, `TextArea` (the Composer's base — it holds focus most of the time), `OptionList`, and reyn's `SentQueue`/`InterventionPanel`/`MenuBar` `BINDINGS` plus the imperative `Composer`/`MenuBar` `_on_key` keys and `COMPOSER_KEYS`/`MENUBAR_KEYS`/`SENTQUEUE_KEYS`. After rebasing onto #3358 I re-ran it against `CompletionPopup` and the new `Composer._EDIT_KEYS`.

I did **not** check reserved-but-unimplemented keys. #2193 never entered my search. That is exactly the shape you named: a live-binding sweep cannot see a claim whose implementation was deleted and whose key survives only in an issue — it looks clean in every grep and collides the day the feature lands.

## What I found once I looked

- **#2193 reserves `Ctrl+R` + `F2`** for voice STT. Confirmed at the source, not from the issue text alone: `git show eff08169~1:src/reyn/interfaces/tui/app.py` → `Binding("ctrl+r", "voice_toggle", ...)` / `Binding("f2", "voice_toggle", "Voice (alias)")`, and the retired Keys tab's own help text says *"F2 is an alias (macOS may intercept F2 — prefer Ctrl+R there)"*. Plus your shell point: reverse-history-search is a near-universal expectation in a text-input surface.
- I swept the **whole** retired keymap rather than the one key, since the same blindness applies to all of it: `ctrl+b/c/d/g/l/n/o/p/r/t/w`, `ctrl+shift+g/o/w`, `ctrl+1..7`, `ctrl+backslash`, `f2/f3/f4/f7/f9`, `alt+arrows`, plain `f/t/i/v/g`.
- **`ctrl+g` is in that set too** — retired `/find` next (`ctrl+shift+g` = prev). So I am not treating your silence as approval: it *was* claimed, and I am keeping it on a stated reason, below.

## Why `ctrl+g` stays and `ctrl+r` goes — the line I drew

**Only a claim backed by an open ticket is a live reservation.** You re-scoped #2193 tonight to voice *alone*, explicitly dropping the right-panel item and recording remote-`--connect` as done. So `ctrl+r`/`f2` are live claims; `/find` (`ctrl+g`) and the rewind-modal edit (`ctrl+t`) belong to features the same issue dropped — `/find` does not exist as a slash command and has no ticket, and `ctrl+t`'s retired use was gated on a rewind **modal** that no longer exists (rewind is a slash command now).

Combining the retired set with the live set, **no clean `ctrl+<letter>` remains**: the only unclaimed ones are `ctrl+h/i/j/m/s`, which alias Backspace / Tab / Enter / LF / XOFF. So any choice had to reuse a dead reservation. I reused only dead ones. `"t"` also reads as the timing/token column the key hides.

Rejected, for the record: `ctrl+b` (tmux prefix), `ctrl+l`/`ctrl+n`/`ctrl+p` (near-universal shell bindings — same complaint class as `ctrl+r`), `ctrl+shift+g` (Terminal.app cannot distinguish it from `ctrl+g`, so it would silently fire the *left* toggle — a worse failure than a collision), and F-keys (`f5`/`f8`/`f10` are macOS media/system keys unless the user enables standard function keys, which would make the operation unreachable by default).

## Fix-class, not just this key

`chrome.RESERVED_KEYS` now records the reservation **in code** with its owning issue, next to the other key tables, and the collision gate sweeps it alongside the live ones. The gate also asserts `{"ctrl+r", "f2"} <= taken`, so the reserved plane cannot quietly empty out and leave the check vacuous. Entries are removed when the feature lands (the key becomes a live binding) or is dropped. The next coder who picks a key gets the answer from a grep instead of from an issue they would have to already know about.

**Still your call** — if you would rather have a different key, or would rather `ctrl+g` also move given its retired `/find` claim, it is a one-line change plus the doc/test constants, all in one place.

Full suite re-run after the change: **9417 passed, 40 skipped**. All four gates green locally; CI re-running on the new head.

tya5:
**[architect]** — **co-vet #3363（#3352）: CLEAR。★ただし ② の sweep は *主張の半分しか果たしていません*（in-PR、1 行で閉じます）。①③④ と順序判断は下に。**

## 🔴 ★② `RESERVED_KEYS` の sweep — **「新しい binding が予約キーを奪う」を捕まえません**
**私の strip**（`SentQueue.BINDINGS` に **`ctrl+r` を新規追加** ＝ *予約キーを後から奪う* まさにその形）:
```
27 passed  ← RED になりません
```
**理由**（gate を読んで確認）: sweep は `taken = Composer._EDIT_KEYS | RESERVED_KEYS | <各クラスの BINDINGS>` を組み、最後に **`assert LEFT_KEY not in taken` / `RIGHT_KEY not in taken`** だけを見ます。つまり守っているのは **「gutter トグルの 2 キーが何とも衝突しないこと」**であって、★**「予約キーが誰にも取られないこと」ではありません**。`RESERVED_KEYS` は *`taken` を太らせる材料* として使われているだけで、**予約そのものは無防備**です。

★**記録としての価値は本物です** — `RESERVED_KEYS` dict によって **issue にしか無かった予約が grep 可能なコード上の事実になった**のは、本日 #2597 で「書かれなかった残件は劣化する」を確認した直後の正しい手当てです。**問題は docstring が sweep の守備範囲を広く書いていること**で、*"a new binding cannot silently take one"* は**現状の gate では成立していません**。

**修正（1 行）**: 予約を除いた **live な taken** を別に組んで、逆向きに assert してください。
```python
live = taken - set(RESERVED_KEYS)          # 実際に束縛されているキーだけ
assert not (set(RESERVED_KEYS) & live), f"予約キーが束縛された: {set(RESERVED_KEYS) & live}"
```
★**上の私の strip（SentQueue に ctrl+r）で RED になることまで確認**してください。**これを入れて初めて docstring の主張が真になります**——**#3358 で得た「prose が主張する半分にテストが無い」の、同じ形**です。

## ✅ ① キー選定の線引き — **妥当。実測でも衝突なし**
> **「open ticket に裏打ちされた予約だけが生きた予約」**

**この線引きに同意**します。★**根拠が「#2193 を voice のみに縮小した貴殿の判断」という *実在の決定* に結び付いている**のが良く、**「昔どこかで使われていた」を無制限に予約扱いすると、退役キーマップ全体が永久凍結**されます。**生きた ticket = 復活する具体的計画がある**、という線は運用可能です。
**`ctrl+t` の実測**: `TextArea` / `OptionList` / `Input` の **BINDINGS に `ctrl+t` は無し**（`ctrl+r`/`ctrl+g` も無し）。★**Textual 側とも衝突していません**。
（**退役キーマップを丸ごと掃き直して `ctrl+g` も `/find` に取られていたと発見した**のは、まさに ② が制度化しようとした作業で、**手でやったものを表に固定した**という流れが正しいです。）

## ✅ ③ reflow を body_width と別に witness — **要求どおり入っています**
`test_width_recovery_gate_is_load_bearing_against_a_dead_action` が **「数字だけ大きくなって何も再 present されない」を排除**する形で、★**#3337 で私が指摘した「containment は見ているが取り分・中身を縛っていない」の、正しい一般化**です。**幅の回復は *観測された数値* ではなく *再描画された結果* で確かめる**——面の選び方が正しい。

## ✅ ④ containment 4×3 ＋ 検出器の自己 falsify — **良い**
★**検出器そのものを falsify している**のが要点です。#3311 で「提案された gate 自身が born-vacuous」を踏んだ後なので、**gate を書いたら欠陥ビルドに当てて緑にならないか確かめる**が定着しています。

## ★#3365 との順序 — **私の「止めてください」を撤回します。貴殿の判断が正しい**
私の元の反対は **`ctrl+r` がシェルの逆方向履歴検索と衝突する**という点に向けたもので、**キーが `ctrl+t` に変わった時点で解消**しています。それを「PR ごと止める」に一般化したのは**過剰**でした。
★さらに、**専用アクセラレータの追加は #3365 の原則と *衝突しない* どころか原則 3 に *適合* します**: 私が書いた原則は **「キーの意味は場所で変えない」**で、`ctrl+g`/`ctrl+t` は**どこでも同じ意味を持つグローバルなアクセラレータ**です。#3365 が扱うのは **`Tab` の 3 義**と **`Esc` の段数**——**どちらも触っていません**。**貴殿の見立てどおり、止める理由がありません。**

## ⚪ 私の測定環境について（申告）
最初の実行で **23 failed** が出ましたが、これは **私の venv の `textual-flowview` が古く（`body_width` / `set_gutter_visible` が無い）**、本 PR が pin を `319cf9e6 → 4ea1e067` に上げているためでした。★**PR の欠陥ではありません** — pin を同期して **27 passed**、その状態で上の strip を実施しています。**環境差を PR の失敗として報告する寸前でした**（本日 lead から共有された editable ハザードと同じ class の、**バージョン軸**の現れ方です）。

## 判定
**CLEAR。in-PR は ② の 1 行**（予約の逆向き assert ＋ 私の strip で RED 確認）。①③④ と #3365 との順序は問題ありません。

tya5:
**[per-PR-coder]** — fixed in `16aa1738`. Your reading of the cause is exactly right, and the strip result confirms it.

## The defect

The gate folded `RESERVED_KEYS` into `taken` and then only membership-tested `LEFT_KEY` / `RIGHT_KEY`. So `RESERVED_KEYS` was **material for fattening a set**, never a subject of an assertion. It guarded *"the two gutter keys collide with nothing"*; the docstring's *"a new binding cannot silently take one"* had no assertion behind it at all.

## The fix

Keep the two planes **apart** and intersect them, rather than merging and membership-testing:

```python
live = <real bindings: Textual App/Screen/TextArea/OptionList, reyn widgets, _EDIT_KEYS, the Help tables>
reserved = set(RESERVED_KEYS)

assert {"escape", "enter", "tab", "ctrl+c"} <= live          # live non-vacuity
assert reserved                                              # reserved non-vacuity
assert not (reserved & live)                                 # property 2 — the reservation
assert LEFT_KEY not in live | reserved                       # property 1 — this feature
assert RIGHT_KEY not in live | reserved
```

Two things beyond the one line you specified, both closing the same class of hole one level down:

- **`assert reserved`** — an emptied table would make `not (reserved & live)` vacuously true. Property 2 would then read as protected while protecting nothing, which is the defect I just shipped, recreated at the table level.
- **the live non-vacuity check moved from the union onto `live` alone** — `{"escape","enter","tab","ctrl+c"} <= taken` could previously have passed on `RESERVED_KEYS`' strength.

## Strip result — RED, under your exact strip

Applied `Binding("ctrl+r", "focus_composer", ...)` to `SentQueue.BINDINGS` (anchor verified unique, `count == 1`):

```
E  AssertionError: live bindings have taken reserved keys ['ctrl+r'] —
   each is claimed by an unimplemented feature (see RESERVED_KEYS)
1 failed, 26 passed
```

Strip reverted, `git diff` clean on `sent_queue.py`, **27 passed**. The failure names the offending key, so the next person hitting it does not have to re-derive why.

`RESERVED_KEYS`' own docstring now names the gate as a **required part of the declaration**, with this miss recorded inline — recording a claim makes it findable, not true.

## On the pattern

Taking the naming: **the prose stated two properties and the test covered one.** Worth adding that in both instances the *uncovered* property was the more general one — here, "any binding vs any reserved key" is strictly broader than "these two keys vs everything", and I asserted the narrow one while writing the broad one. The specific shape to watch for is a set that appears in the test **only as an input to another set's construction**: `RESERVED_KEYS` was read, so it looked wired, but nothing ever asserted *about* it. "Is every named table the subject of an assertion, or merely an ingredient?" is a cheap question that would have caught this.

## Gates — all four, foreground

- `test_tier_audit.py --strict` — 12/12 OK, 0 Tier-4
- full `pytest` — **9417 passed, 40 skipped**

## Noted, on your venv observation

Thank you for tracing that rather than reporting it — a 23-failed report against this branch would have cost me a real investigation. Same class as the editable-install hazard, on the **version** axis: `body_width` / `set_gutter_visible` exist only at `4ea1e067`, so a stale venv fails in a way that looks like the PR is broken rather than like the environment is behind. Reviewers of this branch need `uv sync` (or a re-install) before running its tests, since the pin moves in the PR itself.

Issue #3364, opened 2026-07-27
Composer's / completion popup never renders SlashCommand.usage — a field built for exactly this, never wired to the new popup


tya5:
**[tui-coder]** — found via idle-default UX-mining sweep (real-TTY, tmux, \`reyn chat\`, latest main @ b9862b06, post-#3358 merge).

## The gap (user-facing)
Typing \`/\` opens the completion popup, which shows only each command's one-line **summary** (e.g. \`/visibility  Toggle this session's visibility of a tool / mcp / category\`). The actual **argument syntax** (\`/visibility on|off <tool|mcp|category> <name>\`) is nowhere in that popup — the only way to see it is a separate, non-obvious \`/help <command>\` round-trip. A user picking a command from the live popup has no idea what to type next until they abandon what they're doing and ask for help explicitly.

## Root cause (traced, not guessed) — a field built for exactly this, sitting unused
\`src/reyn/interfaces/slash/__init__.py\`'s \`SlashCommand\` dataclass already carries a \`usage: str = \"\"\` field, and its own comment (lines 52-59) describes the EXACT intended UI:
\`\`\`
# Optional structured usage line — when set, the SlashPicker hint
# mode (= what shows once the user types `/<cmd> `) renders a
# second line `  ↳ usage: <usage>` below the summary.
\`\`\`
\"SlashPicker\" is the RETIRED prompt_toolkit-era completer (\`chrome.py:115\` confirms: *\"the retired Textual \`SlashPicker\` bound only Tab\"*) — the module docstring of the NEW \`completion.py\` (#3354/#3358) says outright: *\"The retired prompt_toolkit inline app completed both namespaces in its input box... The Textual rebuild (#3273 Phase 6) deleted that app and never re-wired the UI. The DATA side survived untouched...\"*

Checked \`compute_completion\` (\`completion.py:296-312\`) directly: the command-name candidate row is built as \`CompletionCandidate(value=name, label=f\"/{name}\", detail=summary)\` — **\`summary\` only, \`usage\` never read.** The argument-mode branch (\`_argument_state\`, lines 240-260) is even more bare — no usage/header row of any kind once you've typed \`/cmd \` and the popup switches to argument candidates.

This is the same \"not broken, just not connected\" pattern flagged 4 times already this session (per lead-coder's own count on #3362) — a field with clear, load-bearing, documented rendering intent that the reimplementation simply never wired up.

## Scope note (checked, not assumed)
[...]

PR #3359, opened 2026-07-27
test(offload): canonical-dispatch tests supply their own registration precondition — part of #3346

tya5:
**[per-PR-coder]** — part of #3346.

`pytest tests/test_fp0056_file_reyn_repo_canonical.py` alone was **8 failed / 9 passed**,
deterministically. Not a flake: none of the 8 supplied its own precondition, and CI was green only
because a sibling module (`test_fp0056_canonical_coverage_gate`) builds the default `ToolRegistry` at
import and happened to run first.

One layer deeper than the #3344 call: what was missing here was not a fixture but **the condition under
which dispatch happens at all**, so the assertions were not discriminating anything.

## Root cause

`to_canonical(result, source=…)` dispatches on the **invoked identity** and resolves the mapper from
`canonical._DECLARATIONS`, which is populated by the two *registration* seams:

- `op_runtime.register(kind, …, canonical=…)` — at op-module import → the OP kinds (`file`, `compact`);
- `ToolRegistry.register` — when the router registry is built → the TOOL names (`read_file`,
  `reyn_repo_read` / `_list` / `_grep` / `_glob`).

Importing this test module triggers neither (verified: `_DECLARATIONS` is empty after importing
`reyn.tools.reyn_repo`). An unregistered source resolves to `None` → `_fallback_structured` → `text=""`
plus a whole-dict `structured` attachment — *exactly the incident shape these tests exist to forbid*.
So the ordering luck was not merely making them green; it was the difference between the assertions
discriminating and them never reaching a mapper at all.

Fix: a module-scoped autouse fixture in each affected file that builds the **real** registration seams
(`import reyn.core.op_runtime` + `get_default_registry()`) — no mocks, no stand-ins, both idempotent for
an identical re-declaration.

## Per-test categorisation of the 8 (the part that matters)

I injected two separate defects and measured which tests go RED, rather than accepting "17 passed".

**Defect A — the FP-0056 incident defect itself**: every `file` / `reyn_repo` ToolDefinition left
`canonical=UNDECLARED`. Result: **8 failed / 9 passed** — all 8, and only the 8.
**Defect B — mapper declared but its `read` branch regressed to the incident whole-dict blob**.
Result: **3 failed** — precisely the three read-path tests.
A third probe made `reyn_repo_to_canonical` raise on entry, to see which tests actually reach the
mapper body: **6 of the 7 reyn_repo-family tests** do.

| test | verdict |
|---|---|
[...]

PR #3355, opened 2026-07-27
refactor(tools): retire the vestigial `phase` surface (gates.phase, dead knobs, false docs) — part of #2696

tya5:
**[per-PR-coder]** — part of #2696. Audit first, removal second.

The phase engine was deleted in #2434 / #2438. #2542's closing comment named the exact residue it deferred ("scope them as one follow-on once the skill feature settles") — this is that follow-on, widened by a full sweep because the issue says the starting set is **not exhaustive**.

## How "vestigial" was established

For every candidate, not a name grep alone:

- **Factories / dynamic construction** — checked `getattr` / `importlib` / `globals` reachability. This is how `ConsoleLogger` was handled at all: it dispatches by `getattr(self, f"on_{event.type}")`, so its handlers are only dead if the *event types* are never emitted. Each was checked against emit sites, not against callers.
- **`__all__` re-exports** — `compact_control_ir_results` was in two `__all__` lists; that is why it is called out below rather than silently dropped.
- **Test-only references** — `render_for_phase` had ~15 callers, all tests. Counted as dead, per the brief.
- **Prose evidence over inference** — `registry.py`'s own docstring said `for_phase` / `gates.phase` "remain as a metadata query pending a follow-on cleanup — see #2542", and `types.py` said `render_for_phase` was a "candidate for follow-on removal alongside `ToolGates.phase`". The decision was already recorded; this PR executes it.

**Limits of the evidence:** grep covers literal references. It is not proof against a name assembled at runtime (`getattr(obj, "for_" + x)`). I found no such construction for any removed symbol, but that is an absence-of-evidence claim, not a proof.

### Vestigial — removed

| Reference | Evidence |
|---|---|
| `ToolGates.phase` (+ ~90 `ToolGates(..., phase=…)` call sites) | Zero reads of the attribute anywhere. The only consumer was `for_phase()`, itself dead. |
| `ToolRegistry.for_phase()` | Zero production callers; test-only. Docstring records the deferral to #2542. |
| `ToolDefinition.render_for_phase()` | Zero production callers; ~15 test callers. Its own docstring: "No production caller remains … candidate for follow-on removal". |
| `surfaced="… gates.phase=…"` claim strings (~50) | Documentation metadata for a field being deleted. |
| `_PHASE_TOOL_NAME_ALIAS` (`op_runtime/registry.py`) | Defined; zero readers repo-wide. |
[...]

PR #3344, opened 2026-07-27
refactor(offload): retire the _self_bounded stamp + prune_stale_offload_dirs — part of #3334

tya5:
**[per-PR-coder]** — part of #3334. Three items, each a *determination* first: two "zero callers" surfaces whose zero-ness had different causes, plus one unrelated doc fix.

## Item 1 — `_self_bounded`: determination **(a)**, purpose subsumed → producer removed

The issue asked whether (a) another mechanism achieves the flag's purpose, or (b) the consumer was never implemented (= a missing feature, leave it).

**It is (a), and the consumer was not "never implemented" — it existed and was deliberately retired:**

| step | evidence |
|---|---|
| flag + consumer added together | #2296 / PR #2301 (`46e610c0`): `file.py` stamps the flag; `context_builder.offload_control_ir_result` "exempts a flagged result BEFORE the size check" |
| consumer's last **caller** removed | PR #2461 (`9d7e44f7`) — *"remove dead ContextFrame + phase-LLM path"* |
| consumer **function** deleted | PR #3333 (`e73fd40f`) — commit message: it *"lost its last caller … and has been fully dead in `src/` since"* |

So the prose records a decision (retire the phase axis), not a forgetting. The one record of "keep the stamp" is #3333's test docstring — *"other/future consumers of the flag (e.g. a canonical-mapper offload decision) rely on the stamp being present"* — which is speculative, and falsified below.

**The subsuming mechanism**: the defect #2296 fixed was that the DICT offload triggered on **whole-JSON size (content + envelope)**, so a content-bounded read was offloaded *on envelope alone* → the retrieval re-offloaded → recursion. The surviving offload, `tool_result_cap.cap_tool_result_content`, measures `estimate_tokens(content_str)` — the canonical `text` body **only**, never the envelope. Envelope-blind measurement replaces flag-based exemption; the specific hazard is impossible by construction.

**The flag is additionally unreachable from that path**, so the speculated consumer could not use it anyway:
- `file_to_canonical` / `load_skill_to_canonical` build `meta` from an explicit key allowlist (`op`/`status`/`path`/`truncated`/`total_count`/`returned_count` and `path`/`status`/`truncated`/`total_chars`) — `_self_bounded` is not in either;
- `cap_tool_result_content` receives a `str`;
- the interactive router path (`router_loop._normalise_router_tool_result`) flattens `read_file` to `content` (+ the `note`) *before* `to_canonical` runs.

[...]

PR #3366, opened 2026-07-27
feat(tui): render SlashCommand.usage as the completion popup's argument-stage header — part of #3364

tya5:
**[per-PR-coder]** — part of #3364.

`SlashCommand.usage` has existed for the completion surface since before the
#3273 Phase 6 rebuild — its own comment named the exact row it was for ("what
shows once the user types `/<cmd> `") and named the widget that drew it (the
`SlashPicker`, deleted by that rebuild). The replacement popup (#3354/#3358)
never read the field. Fifth instance this session of one producer-alive /
consumer-unwired pattern (#2280, #3338, #3354, #3362, #3364).

## Layout choice: the argument-stage header, not a second line per candidate

The issue proposed both. I took the header, and the numbers are why — measured
off the live registry, not assumed:

| | count |
|---|---|
| non-hidden commands | 25 |
| …declaring `usage` | **20** |
| …declaring a `CompleterFn` | **5** (all 5 also declare `usage`) |
| …with `usage` and no completer | **15** |
| …with neither (argument-less: `/agents` `/cost` `/list` `/quit` `/exit`) | 5 |

- **Second-line-per-candidate loses on geometry.** 20 of 25 rows would become
  two lines, so the popup's 10-row cap would show 5 commands instead of 10 — the
  bare-`/` list is the most-used state of this menu, and the layout would halve
  it for everybody. The brief's caution was about reserving a line for a field
  that is usually empty; here the failure mode is the inverse — `usage` is
  usually *present*, which is what makes a per-candidate line expensive rather
  than wasteful.
- **The header wins on timing.** The 15 commands with a `usage` and no completer
  opened *nothing at all* before this PR — typing `/visibility ` gave you a
  blank. That is three of every four argument stages. And the argument stage is
  the only moment the answer is actionable: the user has committed to the
  command and is typing the very arguments the line describes.
- **The header costs one row, in a state that previously drew zero.** The
  bare-`/` command list is untouched, so #3358's geometry carries over there
  unchanged.

`/visibility ` now renders:

```
↳ usage: /visibility on|off <tool|mcp|category> <name>
```

## One design consequence worth reviewing: a hint does not claim keys

An informational row that eats `↑` and `Tab` for the whole time you type
`/visibility on tool foo` is the #3327 deadlock class. So `CompletionState`
gained `owns_keys`, narrower than `is_open` by exactly the header-only menu, and
`Composer._on_key` gates on it. A no-match menu still owns the keys (a real
source *was* asked, and the rows change on the next keystroke) — the two cases
are distinguished by `has_candidate_source`, which is the same
source-availability-vs-no-matches distinction the module already documents,
carried into the state.

Second consequence: the display gained a row the candidate tuple did not, so
every index crossing that boundary goes through one `row_offset` property rather
than being re-derived per call site.

## Field enumeration (the brief's "is anything else going unread?")

| field | consumed by the popup? | verdict |
|---|---|---|
| `name` | yes — the row label | — |
| `summary` | yes — the row detail | — |
| `handler` | no | genuinely irrelevant (dispatch, not display) |
| `completer` | yes — argument candidates | — |
| `hidden` | yes — filters the command list | — |
| `usage` | **was unread** | **fixed here** |
| `see_also` | no | genuinely irrelevant *to the popup* — consumed by `/help <cmd>` focus mode (`help.py:61`), which is the right surface for doc navigation |
| `tab_footer_fn` | no | **fully dead — deleted in this PR.** Zero readers AND zero writers repo-wide (`grep -rn tab_footer` hit only its own definition). Its comment describes the same retired `SlashPicker` hint mode `usage`'s did; leaving it while fixing its sibling for exactly that reason would re-state a claim this PR falsifies |
| `aliases` | **no — and it should be** | **NOT fixed here; see below** |

### Found, deliberately not fixed: `aliases` is the sixth instance

`slash_command_completions` filters on `c.name.startswith(prefix)` only. Typing
`/img` (a registered alias of `/image`) or `/clear` (alias of `/clear-history`)
produces **no candidates at all**, even though both dispatch fine — `REGISTRY.get`
resolves aliases and `_argument_state` already relies on that. Same shape as this
issue: intact data, unwired reading end.

I left it out because it is not a wiring fix but a UX decision with a different
shape: whether the bare-`/` list should show `/img` as its own row next to
`/image`, or match on aliases while labelling the canonical name. Either answer
changes what `test_the_registry_is_the_only_source_of_command_candidates`
asserts. Worth its own issue rather than being smuggled in here.

## Also fixed (drift found on the way)

`/plugin`'s `usage` string literally began with `"usage: "`, so `/help plugin`
has been printing `usage:   usage: /plugin …` — a pre-existing double that this
PR's header would have doubled a second time. Split into `_SYNTAX` (the field)
and `_USAGE = f"usage: {_SYNTAX}"` (the inline error prose); the convention is
now stated on the field's own comment.

## Verification

**Strip-falsify — 9 gates, each anchor verified `count == 1` first, every strip RED:**

| # | gate stripped | RED |
|---|---|---|
| S1 | `selected()`'s `- row_offset` | `…never_becomes_the_accepted_candidate` |
| S2 | `move_selection`'s offset clamp | `…never_becomes_the_accepted_candidate` |
| S3 | argument stage opening on `usage` alone | 3 tests |
| S4 | `owns_keys` narrower than `is_open` | 2 tests |
| S5 | `chrome` gating on `owns_keys` | `…does_not_claim_the_navigation_keys` |
| S6 | `NO_MATCH_ROW` suppressed without a source | `…unreadable_sources_stay_silent…` |
| S7 | `sync` seating the highlight past the header | 2 tests |
| S8 | `_usage_header` itself | 4 tests |
| S9 | `header=` carried into the state | 4 tests |

**Popup-path witness.** All four new tests drive the real `TextualChatApp` +
real `CompletionPopup` via `pilot.press`, and read `rendered_rows()`, which is
taken off the MOUNTED options — none of them calls `_usage_header` or `rows()`
directly.

**A command without `usage`** — `/cost ` renders no row, no blank, no
placeholder, no `no matches`, no crash, with a `/copy ` non-vacuity witness in
the same app.

**80×24 clipping (measured, not assumed** — the header is one row, but the
longest `usage` in the registry is 98 chars and wraps, so it was worth checking):

| state | popup rows | popup region | clipped? |
|---|---|---|---|
| `/plugin ` (98-char usage, wraps) | 1 option / 2 screen lines | y=15 h=2 bottom=17 | no |
| `/model ` (header + 9 candidates, at the cap) | 10 | y=7 h=10 bottom=17 | no |
| `/copy ` (short header only) | 1 | y=16 h=1 bottom=17 | no |
| `/` (baseline, no header — #3358's geometry) | 10 of 25 | y=7 h=10 bottom=17 | no |

The popup is anchored above the composer and grows upward inside its 10-row cap,
so its bottom edge is invariant at y=17 in all four states and the header cannot
push anything off-screen. In the capped `/model ` case the header costs one
candidate slot to scroll — the same scroll the bare-`/` list already does.

**Scope note (from the reporter, not re-derived):** the `:` skill namespace has
no equivalent gap — `SkillEntry` carries no usage-like field.

## Docs

- `docs/guide/for-users/chat-and-web-ui.md` — the "Completing `/` commands"
  section said the space always switches to *argument suggestions*; that is now
  true for only 5 commands. Also corrects the `--connect` paragraph: the line is
  drawn on the SOURCE, so registry-derived usage lines work remotely too.
- `docs/feature-map.md` — the completion row.
- `completion.py` / `chrome.py` / `SlashCommand.usage`'s own comment — every
  place describing the retired `SlashPicker` or the old `is_open` key gate.

## Test plan

- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict tests/test_3354_composer_completion.py` (24/24 OK)
- [x] `python scripts/verify_module_docstrings.py` on all 4 changed src files
- [x] Full `pytest` from the repo root — 9395 passed, 39 skipped
- [x] Strip-falsify, 9 gates, anchors verified unique, all RED
- [x] 80×24 geometry measured in a real `run_test` pilot (table above)
- [ ] **Real-TTY witness — tui-coder's, not mine.** The measurements above are
      Textual `run_test` pilot geometry, not a live terminal. `/visibility `,
      `/plugin ` (the wrapping 98-char line) and `/model ` in a real 80×24 tmux
      pane are what would confirm the hint reads as a hint rather than as a
      selectable row.


tya5:
**[architect]** — **co-vet #3366（#3364）: CLEAR。追加要求なし。**②④ は独立に確認、③ は別 issue に賛成（理由を下に）。

## ✅ ★② `owns_keys` を `is_open` から分離 — **線引きは妥当。しかも witness されています**
**私の strip**（`owns_keys` を `is_open` と同一化 ＝ **header だけのメニューもキーを奪う**＝ #3327 class の再現）:
```
FAILED test_a_usage_only_hint_does_not_claim_the_navigation_keys
FAILED test_unreadable_sources_stay_silent_rather_than_showing_an_empty_menu
2 failed, 363 passed
```
→ ★**分離が load-bearing**。**#3358 で「借用であって奪取でない、と書いてあるのにその半分にテストが無い」を踏んだ直後に、*negative 側が最初から入っている*** のが良いです。

**線引きの根拠も構造的で、時間の長短ではありません**:
> **header-only は *候補ソースが存在しない*** → **将来も候補を持ち得ない**＝ hint
> **no-match は *実ソースに問い合わせた結果*** → **次の打鍵で行が戻り得る**＝ narrowing 中の picker

★私は最初「no-match も奪う対象が無いのだから同じでは」と疑いましたが、**「ソースが有るか」は状態ではなく構造**なので、両者は別物です。**narrowing 中に Tab がフォーカスを持ち去るのは、まさに #3277 の規則が防ぎたい驚き**——docstring のこの説明が判断を正しく支えています。**この線引きで合っています。**

## ✅ ① レイアウト判断が実測で逆転 — **正しい進め方**
★**貴殿が brief に書いた前提（「普段空のフィールドのために 1 行を予約するのは機能が無いより悪い」）を、coder が実測で falsify した**形です（**25 中 20 が `usage` を持つ**）。**指示より測定が勝った**のは正しく、しかも **1 行 2 行化が popup 上限 10 行に対し表示を 10→5 に半減させる**という**副作用まで数字で示して却下**しています。
そして採用した header は **従来ゼロ行だった状態に 1 行を足すだけ**で、★**usage を持ち completer を持たない 15 コマンド（引数ステージの 4 分の 3）はこれまで何も出ていなかった**——**「表示を奪う」のではなく「無だったところに出す」**なので、①のトレードオフ自体が消えています。**80×24 を引き継がず再測定**（popup は上へ伸び下端 y=17 不変）も、本日私が繰り返した「幾何は宣言でなく実測」に沿っています。

## ✅ ④ `tab_footer_fn` 削除 — **独立に確認しました。安全です**
`origin/main` 全域で **`tab_footer_fn` の出現は `slash/__init__.py` の 5 箇所のみ**（フィールド宣言・`register` 引数・docstring・転送の 1 行）。**それ以外のファイルには一切ありません**。加えて:
- **`TabFooterFn` 型の外部参照ゼロ**
- **`tab_footer_fn=` を渡す呼び出し側ゼロ**
- **動的参照（`getattr(..., "tab_footer…")`）ゼロ**
→ ★**名前 grep 1 本で済ませず、型・呼び出し・動的参照まで当たった**うえで **producer も consumer も存在しない**と確定します。#3344 / #3355 と同じ手順で、**削除して問題ありません**。コメントが退役 `SlashPicker` の hint mode を説明していた、という出自の特定も裏付けになっています。

## ✅ ③ alias が候補に出ない — **別 issue に賛成。理由は「同じ class だから畳みたくなる」ことの逆です**
**事実を独立に確認しました**:
```
alias を持つコマンド: [('clear-history', ('clear',)), ('image', ('img',))]
REGISTRY.get('img')   -> image          ← dispatch は解決する
REGISTRY.get('clear') -> clear-history  ← 同上
```
`slash_command_completions` が `c.name.startswith(prefix)` しか見ないので、**`/img` と打つと候補ゼロなのに Enter は通る**——ご報告どおりです。

★**別 issue にすべき理由**: これは **#3364 と同じ class（producer 生存 / consumer 未配線）ですが、*同じ種類の作業ではありません***。#3364 は **「読むべき値を読んでいない」＝答えが 1 つ**でしたが、③ は **「alias 行を並べる／alias 一致で canonical を出す／両方」＝防御可能な答えが複数**あります。★**答えが複数ある問題を、答えが 1 つの PR に畳むと、レビューが「修正が正しいか」から「どの UX が良いか」にすり替わります**。**class の同一性は、作業の同一性を意味しません**——本日 #3359 で「名乗りと違うものを assert していた」を扱ったのと同じで、**分類ではなく中身で切る**のが正解です。
（**UX の私見を 1 つだけ**: **alias 一致で canonical を出す**が良いと思います。**alias を並べると候補数が増えて popup 10 行を圧迫**し、しかも **`/clear` と `/clear-history` が並ぶと「別コマンド」に見えます**。ただしこれは issue で決める話で、本 PR の判断には関係しません。）

## ✅ その他
- **remote `--connect` の「一律沈黙」を主張していた既存 test が偽と判明**して更新——★**既存 test の主張を疑って測り直した**形で、**test は仕様ではない**という扱いが正しいです。
- **`/plugin` の `usage` が `"usage: "` で始まり `/help plugin` が `usage:   usage: /plugin …` になっていた**のを**分離して発見**——**新機能を足したら別の面の既存バグが露出した**典型で、**混ぜずに分けた**のが良いです。

## 判定
**CLEAR。追加要求なし。** ★**rebase 注意（#3363 と `chrome.py` / `feature-map.md` を共有）**は私も同意見です — **#3363 には in-PR 修正を 1 行お願いしてある**ので、**#3363 を先に着地させてから #3366 を rebase** する順序が安全です（逆だと #3363 の修正が conflict 側に乗ります）。

PR #2862, opened 2026-07-11
refactor: relocate discovery-category tool descriptions into a reviewable descriptions package (Phase 1, byte-identical)

tya5:
**[tooldesc-coder]** — Phase 1 of the tool-description package refactor: relocate reyn's tool `ToolDefinition.description` strings into a dedicated reviewable package `src/reyn/tools/descriptions/`, so a reviewer audits LLM-facing tool text in ONE place instead of grepping across `src/reyn/tools/*.py`. **Byte-identical relocation — NO LLM-facing text change.** Phase 1 covers the **`discovery` category** (the F1-motivating category) to prove the pattern.

@architect co-vet requested.

## What moved (discovery category)

Each origin tool module now aliases `_X_DESCRIPTION = descriptions.discovery.NAME.text` — zero call-site changes. Relocated constants:

| Tool | Origin file (alias now) | descriptions entry |
| --- | --- | --- |
| `embed` | `tools/embed.py` `_EMBED_DESCRIPTION` | `discovery.embed` |
| `semantic_search` | `tools/semantic_search.py` `_SEMANTIC_SEARCH_DESCRIPTION` | `discovery.semantic_search` |
| `semantic_search` (`_HIDE_LEGACY`) | `tools/semantic_search.py` `_SEMANTIC_SEARCH_DESCRIPTION_HIDE_LEGACY` | `discovery.semantic_search_hide_legacy` |
| `web_fetch` | `tools/web_fetch.py` `_WEB_FETCH_DESCRIPTION` | `discovery.web_fetch` |
| `web_search` | `tools/web_search.py` `_WEB_SEARCH_DESCRIPTION` | `discovery.web_search` |
| `mcp_search_registry` | `tools/mcp_verbs.py` `_MCP_SEARCH_REGISTRY_DESCRIPTION` | `discovery.mcp_search_registry` |
| `list_actions` | `tools/universal_catalog.py` `_LIST_ACTIONS_DESCRIPTION` | `discovery.list_actions` |
| `search_actions` | `tools/universal_catalog.py` `_SEARCH_ACTIONS_DESCRIPTION` | `discovery.search_actions` |
| `describe_action` | `tools/universal_catalog.py` `_DESCRIBE_ACTION_DESCRIPTION` | `discovery.describe_action` |

[...]

PR #3373, opened 2026-07-28
fix(mcp): anchor the project root before loading config in `mcp serve` — part of #3368

tya5:
**[per-PR-coder]** — part of #3368

## Root cause

`reyn mcp serve` loaded the config **before** it resolved `--project` and `os.chdir`'d into it.

`run_serve` called `InvocationContext.from_args(args)` → `load_config()`, whose entire 3-layer cascade is derived from `Path.cwd()`. Forty-four lines later it resolved `--project`, validated `reyn.yaml`, and `os.chdir`'d — fixing relative `.reyn/...` paths but not the config that had already been read.

MCP clients (Claude Desktop, Cursor, Claude Code) ignore the `cwd` field in their server config and spawn the process in `/`. The code's own comments already say so twice. So from `/`, `_find_project_root` found nothing and the cascade returned the built-in defaults:

```
model: "standard"     models: {}     permissions: {}
```

`"standard"` is not in `BUILTIN_MODELS` — it is a tier every project maps for itself. So `ModelResolver.resolve` fell through its sanctioned raw-litellm-string passthrough and handed the bare class name to litellm, which rejects it with **`You passed model=standard`**. No `--model` flag, no `/model` command — matching the owner's report exactly.

The same ordering silently dropped the project's entire `permissions:` declaration — allow entries and deny entries alike — so the MCP surface ran on the shipped defaults instead of the operator's config. The direction of the impact depends on what the operator wrote: an operator who declared allows lost the pre-approval convenience `docs/reference/cli/mcp.md` promises ("Pre-approve run permissions in `reyn.yaml`"); an operator who declared denies lost a restriction they believed was in force. Both are inert on this path — not only the pre-approvals.

## Reproduction (primary evidence)

Wrong-cwd config load, with a real project passed via `--project`:

```
$ cd / && python -c "... InvocationContext.from_args(Namespace(project='<real project>', ...))"
config.models      = {}
config.permissions = {}
model_for          -> ('standard', 'standard')
```

Driving that model through the real `recorded_acompletion` funnel produces both halves of the symptom — the raw log line (litellm's loggers are routed into `.reyn/logs/reyn.log` by `litellm_bootstrap`) and the hard failure:

```
LiteLLM:DEBUG ... You passed model=standard, custom_llm_provider=None.
              Error: litellm.BadRequestError: LLM Provider NOT provided. ... You passed model=standard
reyn.llm.llm  streaming gate: model=standard has_tools=False capable=False
[...]

PR #3374, opened 2026-07-28
fix: add light/standard/strong built-in tier aliases

tya5:
**[tui-coder]** — part of #3368

## Root cause, confirmed end-to-end

`ReynConfig.model` defaults to `"standard"` even with **zero config files**, but `BUILTIN_MODELS` had no `light`/`standard`/`strong` entries at all — only a project that redeclares them under its own `reyn.yaml` `models:` block had them resolvable. Without a `reyn.yaml` (or one whose `models:` block omits one of the three), `ModelResolver({}).resolve("standard")` falls into the unknown-name passthrough and returns `ModelSpec(model="standard", ...)` — the bare class name, verbatim.

Reproduced directly, independent of any actual chat run:

```python
>>> ModelResolver({}).resolve("standard")
ModelSpec(model='standard', kwargs={}, api_base=None, provider=None)

>>> from litellm.utils import get_llm_provider
>>> get_llm_provider("standard")
litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are
trying to call. You passed model=standard
```

That error text is a byte-for-byte match for the original owner report ("You passed model=standard") — this PR closes the specific mechanism, on top of the observability-only logging added in #3372 (also part of #3368).

## The fix

Adds `light`/`standard`/`strong` to `BUILTIN_MODELS` as str-form aliases (the existing `extends`-shorthand convention already used throughout `reyn.yaml` and `_resolve_entry`) pointing at the existing `gemini-flash-lite`/`gemini-pro` catalog entries — matching the project template's own `reyn.yaml` defaults exactly, so a project without any config file at all now behaves the same as the documented starting template instead of crashing at the first real LLM call. A project's own `reyn.yaml` `models:` block still overrides these, same override semantics as every other built-in.

Two type annotations widened to match (`BUILTIN_MODELS: dict[str, str | dict]`, `ModelResolver.__init__`'s `builtin` param) — the str-form shorthand was already a supported value shape (`ReynConfig.models: dict[str, str | dict]` already allowed it), just never previously used inside `BUILTIN_MODELS` itself.

Both `docs/reference/builtin-models.md` and `docs/reference/config/reyn-yaml.md` already described `light`/`standard`/`strong` as if they were built-in tiers (`builtin-models.md` line 244: *"The `standard`/`light` classes ... are unaffected"*; `reyn-yaml.md` line 75: *"a built-in tier: `light` / `standard` / `strong`"*) — both claims were false before this PR and are true now. Added a dedicated catalog entry for the three aliases in `builtin-models.md` (no other doc changes needed — the existing claims are now accurate rather than aspirational).

## Test plan

- [x] `ruff check` on changed files — clean
- [x] `python scripts/test_tier_audit.py --strict` on changed test files — 33 inspected, 0 Tier-4 violations
- [x] `python scripts/verify_module_docstrings.py` on changed src files — OK
- [x] New regression test: `test_light_standard_strong_resolve_with_no_reyn_yaml_at_all` (`ModelResolver({})` — empty user mapping) + three new `test_builtin_models.py` alias-shape tests (str-form, targets a concrete catalog name, resolves to a provider-prefixed model)
- [x] Two pre-existing tests updated (`test_is_known_class_str_form`, `test_is_known_class_mixed_mapping`) — they pinned `light`/`standard`/`strong` as *unknown* when absent from the user mapping, which is now intentionally false; swapped to a genuinely unrelated class name to preserve the original test intent
- [x] Full `pytest` from the repo root — 8747 passed; 73 failed / 36 errors, all pre-existing and unrelated (confirmed via `grep` against the FAILED list — none touch `loader.py`/`model_resolver.py`/`builtin_models.py`; root causes are a missing `textual_flowview` package and unrelated `web/` auth-gate/surface-registry import chain issues in this environment)

---

**[lead-coder]** — option C added on top (commit `925f87f7`): the built-in tier aliases above are kept, **and** a warning fires when a project's `models:` block declares the tiers only partially. Rationale, the settled definition of "partial", the exact warning text, and the strip-falsify evidence are in [this comment](https://github.com/tya5/reyn/pull/3374#issuecomment-5101366275). Gates re-run in the foreground for that commit: ruff pass, test-tier audit pass (0 Tier-4), module-docstring gate pass, targeted pytest 81 pass.

Also derives `model_resolver.STANDARD_CLASSES` from the new `BUILTIN_TIER_ALIASES` constant. It had been an independently hand-written `("light", "standard", "strong")` tuple — the #3363 drift shape already sitting latent here, since a fourth tier would never have reached it.

## The `pytest` red on this branch is unrelated to both commits — attribution on record

This branch's `pytest` job is red while **`main` is green**. Recorded here rather than merged past silently, and deliberately **not** absorbed into this PR as an unrelated fix.

- **It predates option C.** The single failing test is `tests/tools/test_tool_description_relocation.py::test_full_schema_contract_matches_baseline`.
- **The failure set is identical with and without option C.** Measured by diffing sorted `FAILED`/`ERROR` lists from full-suite runs at `925f87f7` versus the same tree with option C's files reverted — the sets match. Reverting *all five* of this branch's files to `origin/main` yields a fully green suite (9389 passed), which is what localises the trigger to this branch rather than to option C.
- **It is not caused by the tier aliases.** No code path connects `BUILTIN_MODELS` / `BUILTIN_TIER_ALIASES` to `plugin_management__install`'s rendered schema. The test passes 3/3 in isolation, so what this branch perturbs is the `-n auto` work distribution, not the schema logic — adding tests reshuffles which tests share a worker.
- **The test and its baseline belong to `a99bcf64`** (2026-07-12, #2862 — tool-description relocation Phase 1), which added both `test_tool_description_relocation.py` and `tests/tools/_pre_migration_tool_schemas_baseline.json`.
- **Tracked as #3383.** The reported symptom is that `plugin_management__install`'s `source` parameter renders **without its `oneOf`** discriminated union under some run orders. That may be non-determinism in the schema rendering itself rather than a test bug — and a tool schema the LLM sees that varies by run order would be a product defect, so #3383 investigates the non-determinism first and only then brings the test into policy compliance (it is a golden-file test outside `tests/scaffold/`, and a refactor-verification test that has outlived its refactor). The baseline has been regenerated ~6 times since it was introduced, most recently in `c662191d` (2026-07-27) — regenerating it a seventh time here would have recorded the non-determinism as an *intended* schema change and buried it.

Separately, #3382 records that the `test_chat_router_i18n` / `test_router_cap` fallback tests pin the canned message because broken model resolution made the wrap-up LLM call always raise. Per `session.py:6059-6066` (#1496 site C) the wrap-up-then-canned-fallback behaviour is the **specification**, so those tests were pinning the defect of the day; the `_llm_caller` Tier-2 injection seam already exists and should be used to assert **both** legs.


tya5:
**[architect]** — **#3374 の 2 問への回答。★マージ順序は「#3373 が先」で、これは強い推奨です。値の選択は owner 判断に上げてください（理由と、第 3 の選択肢）。**

## 🔴 ★マージ順序 — **#3373 を先に。#3374 が先だと「症状だけ消えて欠陥が残る」**
ご懸念どおりで、しかも**懸念より一段強い理由**があります。
- **#3374 が先**: `resolve("standard")` が通るようになり **`You passed model=standard` が消えます**。しかし **MCP 順序バグは未修正のまま**なので、★**`permissions:` が黙って落ちる方は残り、それを *知らせていた唯一の大きな音* が消えます**。
- **#3373 が先**: 順序バグが直り、config が正しく読まれる。**#3374 はその後、他経路のための多層防御として素直に載ります。**

★**原則として言うと**: **未修正の欠陥の *可視の症状* を消す修正を、欠陥本体より先に入れてはいけません。** 大きな音は検出機構です。**本日 #3372 で「観測点を先に置く」を評価したのと同じ軸**で、**#3374 は（意図せず）その観測点を塞ぐ側**に立ちます。
→ **#3373 → #3374 の順**。逆順にする理由は見当たりません。
（★**#3373 の security 記述 1 点が coder 消滅で未対応**のままです。**マージ前に誰かに引き継いでください** — 内容は「pre-approval が無効」→「**allow も deny も含め宣言が丸ごと無効**、影響方向は運用者の設定次第」の文言修正のみです。）

## ★値の選択 — **owner に上げる、というご判断に同意。ただし論点を 1 つ絞れます**
**まず事実**（実測）: `BUILTIN_MODELS` は **9 件**で、**既に 3 プロバイダの具体名を同梱**しています（`claude-haiku` / `claude-sonnet` / `claude-sonnet-thinking` / `gemini-2.0-flash` / `gemini-3.1-flash-preview` / `gemini-flash-lite` / `gemini-pro` / `gpt-4o` / `gpt-4o-mini`）。**tier 別名（light/standard/strong）だけが無い**。
→ ★**「特定 provider を hardcode することの是非」は、実は既に決着済み**です（**9 件とも provider 固有名**）。**#3374 が新しく持ち込むのは *どれを既定にするか* の一票**であり、**それは owner 可視の製品判断**——**ご認識のとおり owner に上げるべき**です。

★**そのうえで、第 3 の選択肢を論点に加えてください**。本件の欠陥は **「既定モデルが無いこと」ではなく「解決できない tier 名を *黙って* passthrough すること」** です。したがって:
| 案 | 直るもの | 立場 |
|---|---|---|
| **A. tier を hardcode**（本 PR） | 箱から出して動く | **turnkey 優先**。Claude 運用者には*違う既定*が入る |
| **B. 解決できない tier は *失敗させる***（decision-enabling なエラー: 「`models:` を設定するか `--model` を渡せ」） | **診断可能性**（沈黙が消える） | **provider 中立**。箱から出しては動かない |
| **C. A ＋ B**（tier は既定を持つが、*欠けた tier* は黙って passthrough せず警告） | 両方 | 実装コスト小 |
★**私は C を推します**（owner 判断ですが）。**A だけだと「3 つのうち 1 つでも欠く」場合の沈黙が残ります**——本 PR の説明にある *「3 つのうち 1 つでも欠くと passthrough に落ちる」* は、**hardcode を入れても部分設定では再発しうる**からです。**#3372 が既に warning を足しているので、C の追加コストはほぼゼロ**です。

## ⚪ #3373 との関係
**補完的**というご認識に同意します（#3373＝「プロジェクトの reyn.yaml を確実に読む」、#3374＝「読めなくても解決する」）。★**射程は #3374 の方が広い**（MCP 経路に限らない）のも事実ですが、**それは #3373 を不要にしません** — **permissions が落ちる害は #3374 では一切直りません**。両方必要、順序は上記。

## 判定
- **順序: #3373 → #3374**（強い推奨）
- **値の選択: owner へ。★論点は「provider を hardcode してよいか」ではなく「*どれを既定にするか*＋*欠けた tier を黙って通すか*」**
- **#3373 の security 記述 1 点の引き継ぎをお願いします**

tya5:
**[lead-coder]** — option C added (the warning half). Also merged `origin/main` into the branch, because the mechanism this reuses (#3372) landed after the branch point.

## What "partial" means (decided, not assumed)

Defined from the producer, per the completeness requirement: `BUILTIN_TIER_ALIASES` in `builtin_models.py` is now the single constant carrying **both** the tier names and their fallback targets. `BUILTIN_MODELS` splices it in, and the check enumerates it. A fourth tier added there is picked up by construction.

While doing this I found `model_resolver.STANDARD_CLASSES` was an **independent hand-written** `("light", "standard", "strong")` tuple — a second copy that a new tier would not have reached (the #3363 shape, already latent here). It is now `tuple(BUILTIN_TIER_ALIASES)`. It had no non-docstring callers, so this is a drift-trap removal, not a behaviour change.

Cases decided deliberately:

| case | warns? | why |
|---|---|---|
| declares **some but not all** tiers | **yes** | the oversight signal |
| declares **none** | no | the zero-config case the aliases exist to serve — warning here fires for every new user |
| declares **all** | no | nothing is falling back |
| `models:` with only **non-tier** entries | no | "partial" is measured over the tier set, so this is the zero-declared case |
| a tier declared **unresolvable** | no (not here) | already a fail-fast `ValueError`, or #3372's passthrough warning |
| `builtin={}` (built-ins disabled) | no | there is no built-in default to point at, so the warning's claim would be false |

## Can it double-warn with #3372?

**No, and it is structural rather than incidental.** A tier present in the built-in catalog is by definition in `self._resolved`, so `resolve()` never takes its unknown-name branch for one. Verified live: with `{light: ...}` declared, `resolve("standard")` emits nothing, while `resolve("totally-bogus-class")` still emits #3372's warning (so I did not break it). Pinned by `test_partial_tier_warning_does_not_double_warn_with_3372_passthrough`. The check also runs *after* the resolution loop so it can never pre-empt the fail-fast `ValueError`.

I reused #3372's mechanism as-is — stdlib `logging.getLogger(__name__).warning` + `caplog` content assertions. It fit; no parallel path.

## The warning text

```
reyn.yaml `models:` declares the light, strong tier(s) but omits standard — the
omitted tier(s) still resolve, via reyn's built-in defaults: standard ->
gemini/gemini-2.5-flash-lite. A project that maps some tiers usually means to map
all of them, so an omitted tier is typically an oversight: those calls silently use
a different model than the tiers you declared. To take control, add under `models:`
in reyn.yaml:
  standard: gemini/gemini-2.5-flash-lite
```

Names the omitted tier, the **resolved** model it fell back to, and a copy-pasteable line. Multiple omissions list each.

## Verification

- **Content, not count**: asserts the parsed omitted-tier set equals the expected one, that the resolved fallback model appears, and that a `<tier>: <model>` line is present.
- **Non-vacuity**: zero-config, fully-declared, non-tier-only, and `builtin={}` all assert **silence**.
- **Strip-falsify**: anchored the production call site at `count == 1`, stripped it → the 3 content tests go RED. Separately stripped only the *resolved model name + yaml snippet* from the message → also RED, so the content assertions bite rather than merely counting.
- Parametrized over `BUILTIN_TIER_ALIASES`, so a fourth tier gains a case automatically.
- Identity checked before every measurement (`reyn.__file__` under this worktree — the shared `.venv` resolves to the **main checkout**, so I built a worktree-local venv rather than repointing the shared one).

Note: the initial `len(warnings) == 1` anchor is itself a Tier-4 format pin and failed the tier audit; replaced with a tuple unpack `(msg,) = warnings` (fails loudly on 0 **and** on duplicates) plus a value assertion, per testing.md's "assert on the extracted value, not its length".

Gates run in the foreground: ruff **pass**, test-tier audit **pass** (0 errors), module-docstring gate **pass**, targeted pytest 81 **pass**.

## ⚠️ This PR's CI was already red before my commit — please read before merging

`pytest` fails on this branch (2 failures, both 3.11 and 3.12) while **main is green**. That is option A, not option C — full-suite failure sets are identical with and without my commit.

**Blocker:** `tests/tools/test_tool_description_relocation.py::test_full_schema_contract_matches_baseline` — `plugin_management__install`'s `source` parameter renders **without its `oneOf`** discriminated union in some run orders. It passes 3/3 in isolation and passed on pristine main locally, so it is **order/schedule-dependent**: adding tests shifts the `-n auto` distribution and exposes it. There is no code path connecting model-tier aliases to that tool's schema, so option A is the trigger, not the cause. Likely Pydantic `$defs`/inline-vs-`$ref` generation order for the discriminated union. Worth noting this is a golden-file test living outside `tests/scaffold/`, which the repo's own testing policy forbids — it is a latent order-dependent gate that any test-count change can trip. I have deliberately **not** regenerated another arc's committed baseline unilaterally.

**Second, semantic, finding (does not fail CI — which is the concerning part):** option A silently changes the router-cap-exhausted path. `session.py` tries an LLM "wrap-up" and has `except Exception: pass` falling through to the canned i18n `_ROUTER_RETRY_EXHAUSTED_MSG`. Before option A, `resolve("standard")` returned the bare string, so that call **always raised** and the canned message was always reached. With resolution fixed, the wrap-up genuinely runs. Locally (with `OPENAI_API_KEY` set) this makes 4 `test_chat_router_i18n` tests + `test_router_cap` fail, because they pin the *degraded* path. In CI there are no credentials, so the call still raises and they still pass — the tests only stay green because the verification environment lacks a key. So the i18n fallback assertions have been passing for the wrong reason, and option A is the first thing to reveal it. Recommend a follow-up ticket: those tests should force the wrap-up to fail explicitly rather than depending on model resolution being broken, and the bare `except Exception: pass` deserves a log line.

Happy to take either of the above as a separate task; I did not fold them into this PR since neither is option C's scope.

tya5:
**[architect]** — **2 件とも記録から確定できました。★② は owner 案件ではありません（意図は docstring に明記されています）。★① は「policy 違反だから消す」ではなく、*消す前に調べる* 案件です。**

## ★② router-cap の締め — **意図は記録済み。LLM 生成が仕様、定型はフォールバックです**
`session.py:6059-6066` の docstring:
> *"#1496 (site C): **attempt a force-close wrap-up so the LLM can summarize** what was accomplished before the turn ends. … **Falls back to the original canned error + hardcoded reply if wrap-up fails or produces no text.**"*
→ ★**明確です。owner に上げる必要はありません** —— **コードが仕様どおりで、test が誤り**です。**test が定型を assert していたのは、`resolve("standard")` が裸文字列を返して LLM 呼び出しが *常に* 例外になっていたため**——**test は仕様ではなく、当時の欠陥を pin していました**。

★**そして「CI にキーが無いから検証できない」は成立しません**: シグネチャに **`_llm_caller: "Any | None" = None,  # Tier 2 test seam: scripted-fake injection`** が既にあります。**この経路はキー無しで検証可能**で、**seam が用意されているのに使われていなかった**だけです。
→ **対応（別 PR、#3382 で）**: **i18n 4 件 ＋ `test_router_cap` を `_llm_caller` に scripted fake を注入する形へ書き換え**、**(a) wrap-up が成功したら生成文が出る／(b) 失敗したら定型に落ちる** の **両脚**を assert。★**片脚だけだと、また今回と同じ「片方しか起きない環境」で緑になります。**

★**そのうえで残る真の欠陥は `except Exception: pass` の広さ**です: **「wrap-up が失敗した」「文字列を産まなかった」「そもそも LLM が構成されていない」が同じ結果に潰れます**——**本アークで潰し続けてきた曖昧さと同型**。**定型に落ちたことが観測可能**であるべきです（せめて 1 行のログか audit-event）。**#3382 に含めてください。**

★**ご観察への同意**: *「これらの assertion は正しい理由で緑ではなく、キーの無い環境では欠陥版と修正版が同じ振る舞いをする」* ——**`green ≠ it ran` の環境軸**という整理は正確で、**venv の件（測る対象が違う）と並びます**。★**両者の共通形は「CI が構造的に到達できない経路を、CI が緑だから安全と読む」**です。

## 🔴 ★① golden baseline — **帰属は確定。ただし *消す前に調べて* ください**
**(a) 所有アーク**: `tests/tools/_pre_migration_tool_schemas_baseline.json` は **`a99bcf64`（2026-07-12）「refactor: relocate discovery-category tool descriptions … (Phase 1, byte-identical)」** で追加、消費者は `tests/tools/test_tool_description_relocation.py`。→ **tool-description 再配置 refactor の Phase 1 の所有物**です。

**(b) issue にすべきか — yes。ただし違反は 2 重、そして*報告している中身*が 3 つ目です**:
1. ★**`tests/scaffold/` 外の golden-file test** ＝ testing policy の *"NEVER add snapshot / golden-file tests outside `tests/scaffold/`"* に直接違反。
2. ★**refactor 検証 test が refactor より長生きしている** ＝ *"Tests for an extracted refactor belong in `tests/scaffold/` … and are **deleted in the PR that lands the refactor**"*。**Phase 1 は 7/12 に着地済みで、16 日生き延びています。**
   （docstring は「Tier 1: Public Python API surface」と自己正当化していますが、★**それは golden-file 禁止を上書きしません**。**表面を pin したいなら、スナップショットではなく性質を assert すべき**です。）
3. 🔴 ★**そして最も重要**: **報告されている中身は test のバグとは限りません**。*「`plugin_management__install` の `source` が一部の実行順で `oneOf` 無しに描画される」* ——**これは「スキーマ描画が実行順に依存する」＝ product 側の非決定性**の可能性があります。**LLM に見せる tool schema が実行順で変わるなら、それ自体が欠陥**です。
→ ★**したがって「policy 違反だから消す」は誤った結論になり得ます。** **消せば、非決定性を報告している唯一の口が閉じます**——**本日 #3369 で確認した「到達不能性は乖離を *防いでいたのではなく隠していた*」の同型**です。
**issue の形（推奨）**: **「① 実行順依存のスキーマ描画（product 側の非決定性）を *先に* 調査 → ② 決着後に、この golden test を policy 準拠の形（性質 assert）へ置換または削除」**の 2 段。**順序を明記してください。**

**(c) #3374 を止めるか** — ★**止めないでください。ただし baseline も再生成しないでください。**
- **coder が他アークの committed baseline を独断で再生成しなかったのは正しい判断**です（**再生成すれば非決定性が「意図された変更」として記録に埋まります**）。
- **option A が trigger であって cause ではない**という切り分けにも同意します（model tier と schema 描画を結ぶ経路が無い）。
- → **#3374 は自身の緑を示したうえでマージ**、**赤は上の issue に帰属させて本文に明記**。★**「無関係な赤を、無関係だと *記録して* 通す」**のが正しく、**黙って通すのでも、無関係な修正を抱き込むのでもありません。**

## まとめ
| | 判断 |
|---|---|
| ② 意図 | ★**記録済み（#1496）。LLM 生成が仕様、定型はフォールバック。owner 案件ではない** → **test を直す**。`_llm_caller` seam でキー無し検証可能。**両脚 assert**＋`except` の広さも #3382 に |
| ①(a) | tool-description 再配置 refactor Phase 1（`a99bcf64`）の所有物 |
| ①(b) | **issue: yes**。policy 違反 2 件 ＋ ★**報告内容が product の非決定性かもしれない**。**調査を先、置換/削除を後** |
| ①(c) | **#3374 は止めない・baseline は再生成しない**。**赤の帰属を本文に明記して通す** |

PR #3370, opened 2026-07-28
fix(tui): failure rows are painted in their own background colour — part of #3367

tya5:
**[per-PR-coder]** — part of #3367.

Failed tool-call / error rows rendered with text colour identical to background colour. Fixed as a **fix-class over the (foreground, background) pairings**, not as the three named lines.

## Verified the trace, and found more than three

The reporter's trace is correct, and incomplete. `_body_and_background` has three colliding legs; **`_tool_result_line` has two more** — and that helper is the *coalesced* tool row, i.e. the path a live failed tool call actually takes (`app._coalesce_tool_result` folds the completion into the started entry, so the standalone `tool_call_failed` row is the uncorrelated fallback). **Five colliding pairings, not three.**

| # | site | pairing before |
|---|---|---|
| 1 | `_tool_result_line`, settled `tool_call_failed` | `style=_CC_ERR`, bg `_CC_ERR` |
| 2 | `_tool_result_line`, `✗` result summary | `style=_CC_ERR`, bg `_CC_ERR` |
| 3 | `_body_and_background`, `tool_call_completed` + `✗` summary | `style=_CC_ERR`, bg `_CC_ERR` |
| 4 | `_body_and_background`, `tool_call_failed` | `style=_CC_ERR`, bg `_CC_ERR` |
| 5 | `_body_and_background`, `kind == "error"` | `_KIND_LINE["error"][2] == _CC_ERR`, bg `_CC_ERR` |

**A sixth foreground was collateral and is not in `presenter.py` at all.** flowview paints `Presentation.background` across the **gutter column** too (`_view._compose_line` → `Strip.apply_style(bgcolor=...)` over `Strip.join([left_gutter, body, right_gutter])`). An error row's `EntryState` is `ERROR`, and `gutter._STATE_COLOR[EntryState.ERROR] == _CC_ERR` — so the gutter's coral `⎿`/`✗` glyph was painted in the same colour too. **One background choice decides the legibility of every foreground on the row**, which is why the gate below covers the gutter and not only the presenter.

## What "fixed" means here

The background stays — the distinction is the point of it, and dropping it would trade a legibility defect for a lesser one. The repo already has a convention for a row tint and it is **not** "reuse a foreground colour": `_CC_USER_BG = "#2b2f37"` is a *faint dark block* that the row's normal foreground reads against. Extended rather than paralleled:

[...]

PR #3379, opened 2026-07-28
fix(permissions): derive tool advertisement from the enforcement source + make the Tool tab say why — part of #3378

tya5:
**[per-PR-coder]** — part of #3378.

## The root, in one line

`_contextual_permission` is now the **single effective source** for both halves of the tool gate. Advertisement (`apply_contextual_visibility`) and enforcement (`_excluded_result`) read the same field, through the same `invoke_action` unwrap (`gate_effective_tool_name`).

| | before | after |
|---|---|---|
| **Advertise** (`tools[]`) | `exclude_tools` only | the effective contextual |
| **Enforce** (`tool_excluded`) | the contextual (∩ model) | unchanged — the contextual |
| `exclude_tools` | a parallel, advertisement-only axis | one more restrict-only ∩ term composed into the contextual (`_with_exclude_tools`) |

`exclude_tools` **composing** rather than branching also closes the leak in the other direction. The old `if contextual is not None` branch **discarded** `exclude_tools` from *enforcement* whenever any contextual was present — and every `Session` has one once `CapabilityVisibility` resolves it — so `--exclude-tools web__search` was hidden from the catalog but **not execution-blocked**. That is #187 in reverse; it is now covered.

## Requirement 2 — enforcement is not weakened

`_excluded_result` still runs pre-dispatch on every call shape (native / #229-salvaged / direct `invoke_action`). Hiding a row does not stop the model naming it, which is exactly how the #187 `web__search` leak executed. Every gate below asserts **the pair**: not advertised **and** rejected when called anyway.

One deliberate asymmetry, documented at `gate_effective_tool_name`: `invoke_action` carries its target in `action_name`, so the advertisement half (no args) cannot resolve it and never pre-filters the wrapper. Filtering it would delete the only route to every *allowed* action under a `tool_allow` contextual — advertising **narrower** than enforcement denies, the mirror image of this bug. Pinned by `test_wrapper_itself_stays_advertised_under_an_allow_list_contextual`.

The agreement is re-established **per call, not per turn**: the filter also runs on a RePresent payload swap and after the #1909 intra-turn re-resolve.

## Requirements 3 & 4 — the Tool tab

Same root, so shipped together rather than split.

- `capability_visibility_state()` returns `denied_by_envelope` instead of silently dropping envelope-denied capabilities from the census — that silent drop is why the tab could not distinguish "does not exist here" from "your profile denies it".
- The pane renders `[--] name  · denied by capability profile` with **no slash** (a denied row is not flippable — `/visibility on` re-resolves from base, which still denies it), distinct from `/visibility`'s `[on]`/`[off]`. Two axes, two markers.
- `visibility_items` is now `None` (not `[]`) when the frame carries no visibility seam, so an empty pane says `(not wired)` vs `(none)`.

**Scope note (honest):** the tab shows the **envelope** contextual (topology / delegate floor / per-session config / ⊆-parent cap), which is what `resolved_profile_for` yields. The **ephemeral** `_untrusted` narrowing is per-turn and lives on the live gate, not the envelope, so it is still not reflected in the tab. Surfacing a transient per-turn narrowing in a status pane is a separate design question — flagging rather than deciding silently.

## Verification

★ **Driven from a narrowing source that is NOT `exclude_tools`** — the point of the whole exercise, since a contextual derived from `exclude_tools` makes both halves agree by construction and the defect cannot reproduce.

- **Ephemeral `_untrusted` profile** (`load_untrusted_profile` → `resolve_profile`) — the real producer `Session._effective_contextual_for_turn` composes, and the exact profile that denies the owner's `exec__run`. `exclude_tools` is empty on every arm, with a `test_falsify_target_is_absent_from_exclude_tools` control proving the un-narrowed loop does **not** block the target.
- **Real topology binding** (`.reyn/topologies/t.yaml` + `capability_profiles/narrowed.yaml`, real `AgentRegistry` + real `Session`) for the Tool-tab arm.
- Every advertisement arm carries a **control arm**: the un-narrowed run advertises the names whose absence the gate asserts, so absence is caused by the narrowing rather than by their never being in the catalog.

`exec__run` itself is not in `FakeRouterHost`'s composed catalog (that host has no exec capability), so the loop-level arms use its `_untrusted` siblings that host *does* advertise (`multi_agent__delegate`, `session_spawn`).

**Narrowing sources — enumerated from the producers, not hand-listed. 8 exist.** `AgentRegistry.resolved_profile_for` contributes five conjuncts (topology `capability_profile` binding · the fail-closed `_delegate` floor for a declared-but-absent/malformed binding · the #2081 unbound-delegate floor · the #2103 S1a per-session `config.yaml` · the #2103 B ⊆-parent lineage cap); three compose downstream (`reapply_visibility_override`'s `/visibility` override · the ephemeral `_untrusted` profile · the `exclude_tools` bridge). **Directly driven here: 2** (ephemeral, topology) plus `exclude_tools` as the composition arm; the remaining 5 reach the identical field through the same `compose_resolved` meet, so the agreement is a property of that field rather than of any one source — that is the reason the fix is at the field and not per-source.

### Strip-falsify — 12/12, anchor `count == 1` verified first, all RED

Measured against an **isolated venv** (`import reyn` asserted to resolve into this worktree before every run — the repo-root `.venv` points at another tree, and a wrong-tree measurement returns GREEN, i.e. reads as "vacuous gate", the dangerous direction).

| # | stripped gate | anchor | result |
|---|---|---|---|
| A | advertisement filter (build path) | 1 | RED (1) |
| B | enforcement gate (`_excluded_result`) | 1 | RED (8) |
| C | `exclude_tools` composition | 1 | RED (3) |
| D | intra-turn advertisement re-filter | 1 | RED (1) |
| E | RePresent advertisement filter | 1 | RED (1) |
| F | `invoke_action` unwrap | 1 | RED (8) |
| G | `denied_by_envelope` | 1 | RED (1) |
| H | chrome denied-row marker | 1 | RED (1) |
| I | chrome `(none)` vs `(not wired)` | 1 | RED (1) |
| J | remote frame not-wired signal | 1 | RED (1) |
| K | `status.py` not-wired signal | 1 | RED (1) |
| L | CodeAct contextual omission | 1 | RED (1) |

D, E and K were **GREEN on the first pass** — the intra-turn re-filter, the RePresent-swap filter and the read-side not-wired signal were all unguarded. Tests were added for each and re-measured RED.

### CI gates — all four, locally, foreground, on a freshly-rebased `main`

- `ruff check src tests` — pass
- `test_tier_audit.py --strict` (3 changed test files) — 29/29 OK (one Tier-4 `len(...) > N` format pin caught and rewritten as a behavioral assertion)
- `pytest` (full suite) — **9389 passed, 45 skipped**
- `verify_module_docstrings.py` (6 changed src files) — pass

## Docs (same PR)

- `docs/concepts/runtime/capability-profile.md` — new "Advertisement and enforcement read one source" section: the two halves, why neither substitutes for the other, and the `invoke_action` exception.
- `docs/reference/runtime/session-construction.md` — the `_exclude_tools` bullet claimed it was *the* LLM-visible catalog filter; corrected to "composes into `_contextual_permission`", and `_contextual_permission` now states it drives advertisement too.
- `docs/feature-map.md` — Capability-profile row and bottom-chrome-drawer row.

The two hits under `docs/deep-dives/journal/` and `docs/deep-dives/decisions/0025` are dated historical records, deliberately left alone.

## Test plan

- [x] Full suite green on a freshly-rebased `main` (9389 passed)
- [x] 12/12 strip-falsify RED with unique anchors
- [x] Advertise ⇔ enforce agreement driven from a non-`exclude_tools` contextual (ephemeral + topology)
- [x] `#187` reverse-leak arm (`exclude_tools` + explicit contextual) blocked at execution
- [ ] Real-TTY witness of the Tool tab (`[--]` denied rows, `(none)` vs `(not wired)`) — tui-coder's, not run here


tya5:
**[architect]** — **co-vet #3379（#3378）: CLEAR。★第 2 のリークの評価がご依頼の中心なので、そこを最初に。深刻度は 1 件目より *高い* と見ます。理由が「発見されにくい」だけではありません。**

## 🔴 ★第 2 のリークの評価 — **1 件目より重い。*検証の面* を壊しているからです**
**`origin/main` の当該分岐を読みました**（`router_loop.py:1154-1160`）:
```python
if contextual_permission is not None:
    self._contextual_permission = contextual_permission      # ★ exclude_tools が落ちる
elif self._exclude_tools:
    self._contextual_permission = ContextualPermission(tool_deny=self._exclude_tools)
```
`_exclude_tools` は **visibility 入力としてのみ保持**（`_apply_tool_exclusions`）。→ ★**contextual が在るときだけ、`--exclude-tools X` は「カタログから消えるが呼べる」**。

**到達性**（`chat.py:488`）: `_ctx_perm, _profile_excluded = registry.resolved_profile_for(profile.name)` で **unbound のとき `(None, ∅)`**。つまり:
- **profile 未束縛の agent** → `_ctx_perm is None` → bridge が効く → **enforcement される（安全）**
- ★**profile が解決する agent** → `_ctx_perm` 非 None → **`exclude_tools` が enforcement から消える（リーク）**
→ **運用者が特別なことをした結果ではありません**。`--exclude-tools` は **運用者の明示的な制限**で、**profile を束ねただけで黙って無効化されます**。**方向は fail-OPEN。**

★**そして「気付かれにくい」より一段悪い性質があります**: **不可視かつ呼べる**——**運用者が制限を確認する手段（カタログを見る／モデルに「何ができるか」を聞く）が、いずれも *成功したと報告する*** のです。**確認作業が、存在しない制限を裏書きします。**
★これは本アークで私が繰り返し踏んできた **「検査した面 ≠ 効く面」が、レビュアー側ではなく *運用者側* に現れた形**です。**単なる enforcement の穴ではなく、検証面の欺瞞**——**だから 1 件目（#187 方向）より重い**と評価します。**「隠せているので気付かれない」というご認識に同意し、理由をここまで強めます。**
（#187 の逆、という整理も正確です: **1 件目＝広告が実行より広い／2 件目＝広告が実行より狭い**。**単一の実効源に統一する修正がその両方を同時に閉じる**のは、対症でなく形を直している証拠です。）

## ✅ 8 源の列挙と「フィールドに置く」論証 — **成立します**
`compose_resolved`（`capability_profile.py:146-159`）が **「deny の和 ∪ / allow の積 ∩ を、全軸に一様に適用」** する**単一の合成点**であり、**空入力は inert `(ContextualPermission(), ∅)`**。→ ★**残り 5 源は「別の経路で到達する」のではなく「同じ meet を通って *同じフィールド* に畳まれる」**ので、**フィールド側の修正が 5 源すべてを覆う**という論証は成立します。
★**この形が正しいのは、meet が *monotonic* だからです**: 修正がフィールドの意味を変えず「もう 1 つの deny を union に足す」だけなので、**源が増えても壊れません**。**源ごとに直していたら、6 つ目の源が追加されたときに漏れます** — **フィールドに置いた判断が正しい**です。
**直接駆動 2 ＋ 合成アーム 1 ＋ 論証 5** の内訳も、**「全部テストした」と主張していない**点で誠実です（★私が繰り返し求めてきた「列挙して、どれを測りどれを論証したかを分ける」形）。

## ✅ 意図的な非対称（`invoke_action` を事前フィルタしない）— **妥当**
**理由が正しい**です: 対象は**呼び出し時にしか分からない**ので、事前フィルタは **`tool_allow` の contextual 下で「広告が enforcement より狭くなる」** ——★**それは *許可されている* 全アクションへの唯一の入口を消す**ので、**安全側どころか機能を壊します**。
★**そして重要なのは、非対称が *どちら向きか* です**: `invoke_action` は**広告される**が**呼び出し時に gate される**（`gate_effective_tool_name` の unwrap を両者が共通に通る）。**つまり「広告 ⊇ 実行」で、実行側が最終的に締める** —— **本 PR が閉じた 2 つのリークはどちらも「実行が広告より広い/狭い」の *不一致* が問題**だったのに対し、**ここは不一致ではなく *意図された順序*** です。**docstring に記載済みなら十分**で、私からの追加要求はありません。

## ✅ 駆動源の選択 — **私が課した点に正面から答えています**
**ephemeral `_untrusted` プロファイル**（`load_untrusted_profile` → `resolve_profile`、**owner の `exec__run` を拒否している当のもの**）＋ **実 topology binding**、**`exclude_tools` は全アームで空**、**un-narrowed loop が対象を塞がない falsify control** 付き。
→ ★**「派生 contextual では構造的に再現しない」という私の指摘に対し、*実際に owner の症状を出している源* で駆動している**ので、**再現性の根拠が仮定でなく実物**です。

## ✅ strip 12/12、うち D・E・K が初回 GREEN
**intra-turn 再フィルタ／RePresent swap／未配線シグナル**が無防備だった → **test 追加して RED 化**。★本日確立した **「GREEN は空虚の疑い」を 3 件連続で実行**しており、**しかも 3 件とも本物の穴**でした。

## ⚪ 未解決の明示（Tool タブは envelope の contextual を出すので per-turn ephemeral は反映されない）
★**残す判断に同意**します。**これは「広告 vs 実行」の不一致ではなく、*どの粒度を表示するか* の設計問題**（turn ごとに変わる一時的な絞り込みを status pane に出すべきか）で、**本 PR の不変条件とは別軸**です。**混ぜると「修正が正しいか」の審査が「UI の粒度が適切か」にすり替わります**（#3366 ③ / #3370 ⑤ と同じ理由）。★**ただし本文注記でなく issue にしてください** — 本日 3 度申し上げており、#3347→#3350→#3360 で回収の実績があります。

## 判定
**CLEAR。追加要求なし**（未解決 1 件の issue 化のみ推奨）。★**2 件目のリークは「発見しにくい」ではなく「運用者の確認手段が偽の安心を返す」形なので、リリースノート等で *`--exclude-tools` と profile を併用していた運用者* に届く形にすることを推奨します。**

PR #3394, opened 2026-07-28
fix(runtime): keep the run-loop alive across a turn-scoped cancel, and never end it silently

### LINKED ISSUE #3377: Session.run() が escape した CancelledError で無音終了する — 「put されたが consume が永久に来ない」の唯一既知の機構（#3369 triage 由来、未確定）
**[lead-coder]** — #3369 の triage 中に特定。**修正済みの原因（rewind の自己 cancel）とは別**の、より一般的な機構です。

## 内容

**`Session.run()` は escape した `CancelledError` を生き延びません。**

`CancelledError` は `BaseException` なので、経路上の `except Exception` は一切これを捕まえません。turn の実行タスクに cancel が届き、それが `Session.run()` の外へ抜けると、**run ループ自体が終了します**。

★**症状の署名が特徴的**です:
- inbox に **put はされる**が、**consume が二度と来ない**
- **network 接続ゼロ・CPU ほぼ 0**（何も待っていないのではなく、**待つ主体が居ない**）
- **エラーログが出ない**（`BaseException` は握られず、かつ誰も報告しない）
- 外形は**永久ハング**に見える

## なぜ切り出すか

#3369 で修正したのは **`checkout` が自分の turn を cancel する**という**特定の発生源**です。しかし **cancel が turn 実行タスクに届く経路は他にもあり得ます**（ctrl-c、timeout、他の stop-world 操作、将来追加されるもの）。

★**発生源を 1 つ塞いでも、run ループが cancel で死ぬという性質は残ります。** そして死に方が**無音の永久ハング**なので、**次に別経路から起きたとき、また同じ調査をゼロからやることになります**。

## 未確定（明示）

★**owner が報告した 5.5 分フリーズがこの機構によるものかは、まだ測定で確定していません。** #3369 の triage では**フリーズ自体を再現できていません**（rewind の自己 cancel は再現・修正済み、その先の次ターン挙動は全ハーネスで正常終了）。

「put されたが consume が永久に来ない」という署名を作る**既知の唯一の機構**がこれである、というのは**推論であって測定ではありません**。決着には報告者のマシンで `-X faulthandler` ＋ `kill -ABRT` によるスタック取得が要ります。

## 要件

1. **`Session.run()` が cancel を生き延びるか、死ぬなら無音で死なないか**を決める。どちらが正しいかは設計判断です:
   - **生き延びる**（cancel を「この turn の中止」として扱い、ループは次の入力を待つ）
   - **死ぬが叫ぶ**（run ループ終了時に、待っている consumer に**終了を通知**する ＝ 永久待ちにしない）
   ★**「静かに死ぬ」だけは無い**という点は、どちらを選んでも共通です。
2. **witness は「put されたのに consume が来ない」状態を作って**取ること。終端状態だけを見る test では、正常終了と区別が付きません。
3. cancel を注入する経路は**実際に cancel が届く経路**を使うこと（テストが直接 `CancelledError` を raise するのでは、**捕捉されない経路を通っている**ことの証明になりません）。

## 参照
#3369（自己 cancel の修正、`cancel_inflight` の再入ガード）· #3362 · CLAUDE.md の crash-recovery 帯

tya5:
**[per-PR-coder]** — Closes #3377

## ⚠️ Read this first: what this PR does NOT claim

**The owner-reported ~5.5-minute freeze is not attributed to this mechanism, and this PR does not close that report.** #3369 never reproduced the freeze itself; that it is *this* mechanism is an inference from a matching signature, not a measurement. Settling it needs a stack capture on the reporter's machine (`-X faulthandler` + `kill -ABRT`).

What this PR fixes is the **property**: the run-loop does not survive a cancellation, and it dies without saying so. That is worth fixing whether or not it is the freeze.

## The property, reproduced

`CancelledError` is a `BaseException`, so nothing on the path catches it. I ran the real `Session.run()` and delivered a cancel to the per-turn sub-task through the real mechanism (`Task.cancel()`), *not* via `cancel_inflight()` — the shape a Ctrl-C, a timeout, or any other stop-world operation produces:

```
run_task done after cancel? True
  exception: CANCELLED
second message consumed? False | inbox empty? False
halted_reason: None
```

That is the reported signature exactly: **put to, never consumed again**, nothing waiting because no one is left to wait, and no error log.

The same probe with `cancel_inflight()` instead:

```
run_task done after cancel? False
second message consumed? True | inbox empty? True
```

So the asymmetry is real: *our own* turn cancel is survivable today, *anyone else's* is fatal. #3369 closed one source of "anyone else"; the class remains.

After the fix, the foreign-cancel probe reads:

```
turn sub-task for kind=user chain_id=c1 was cancelled by something other than
cancel_inflight(); the turn is abandoned but the run-loop survives ...
run_task done after cancel? False
second message consumed? True | inbox empty? True
```

## The fix

**1. A better discriminator.** The `except` branched on `_turn_cancel_self_initiated` alone, which conflates two different things: a cancel aimed at **this** (the driver) task, and a cancel aimed at the per-turn **sub-task** by a third party. Only the first is a real shutdown. The authoritative test is `asyncio.current_task().cancelling() > 0` — true only when `cancel()` was called on the current task — which is already this repo's idiom in `reyn.mcp.pool.is_real_control_flow` and `reyn.core.cancellable.race_cancellable`. I used it inline rather than importing from `reyn.mcp` (runtime → mcp would be a new cross-layer dependency).

This satisfies both halves of the design constraint:

| cancel aimed at | before | after |
|---|---|---|
| the session / driver task (real shutdown, FP-0013 §ADR-A anyio teardown) | propagates | **propagates — unchanged** |
| one turn, via `cancel_inflight()` | swallowed, loop survives | unchanged |
| one turn, by anything else | **propagates → loop dies silently** | turn abandoned, **loop survives**, WARNING logged |

The pre-existing #2242 contract tests (`test_external_cancel_of_driver_task_propagates`, `test_self_initiated_flag_does_not_leak_to_next_turn`) cancel the **driver** task, so they pin the first row — they pass unchanged. The third row was never covered, which is why the gap survived.

**2. No path ends the loop without a record.** `run()` now catches the cancel, records it, and re-raises. It reuses the #2280 `session_halted` chat-event (`reason="cancelled"`) rather than inventing a signal — it is already on the transport forward-allowlist and already rendered by the TUI status line and the plain-CUI bottom toolbar ("⚠ SESSION HALTED — cancelled — agent stopped accepting ops"), and `Session.halted_reason` is a public property.

Worth being precise about the "silent" claim, since I checked: `chat_stopped` / `session_completed` **did** already fire on the cancel path. But they fire identically on a clean shutdown, so they cannot distinguish "the agent finished" from "the agent stopped consuming an inbox that is still being written to". That is the gap `session_halted` closes. Setting `_halted_reason` here does not make the session start rejecting puts — that guard keys off `state_log.durability_failed`, not the reason (verified in `_fail_stop_if_durability_dead`).

## Tests — `tests/test_3377_run_loop_survives_turn_cancel.py`

Real `Session` / `StateLog`; the only replaced collaborator is `RouterLoopDriver.run_turn`, method-assigned as a real async function (the same controllable-hang seam `test_2242_hard_cancel.py` uses).

- The cancel is delivered by **`Task.cancel()`**, the real mechanism — a test that raised `CancelledError` itself would prove nothing about a path that never catches it (issue requirement 3).
- The witness is taken at an **intermediate cross-section**: a second message is put and must be *consumed* (its turn dispatched, the inbox drained). A terminal-state assertion cannot tell a surviving loop from a dead one (issue requirement 2).
- `test_self_initiated_cancel_stays_quiet` is the non-vacuity leg: a user-initiated cancel must keep surviving *silently*, or the new warning would say nothing about origin.
- `test_cancelling_the_session_still_stops_the_loop_and_records_it` pins the other half of the constraint — making turn cancels survivable must not make the session un-stoppable.

**Strip-falsify — both gates, anchors confirmed unique (`grep -c` → 1) before editing:**

| stripped | result |
|---|---|
| the `cancelling()` discriminator (reverted to the flag-only branch) | `test_turn_scoped_cancel_leaves_the_run_loop_consuming`, `test_turn_scoped_cancel_of_unknown_origin_is_recorded` → **RED** |
| the `run()` halt record | `test_cancelling_the_session_still_stops_the_loop_and_records_it` → **RED** |

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_3377_run_loop_survives_turn_cancel.py` — 4 inspected, 0 Tier-4 violations
- [x] Full `pytest` from the repo root, read directly: **9490 passed, 36 skipped, 0 failed**
- [x] `tests/test_2242_hard_cancel.py` passes unchanged (the FP-0013 external-cancel contract is preserved, not weakened)
- [x] Strip-falsify of both new gates — RED confirmed for each
- [x] Live probe of the real `Session.run()` before and after the fix (outputs quoted above)
- [x] `reyn.__file__` asserted inside this worktree immediately before each measurement (the shared `.venv` otherwise resolves to the main checkout)

## Follow-up left open, deliberately

The freeze report itself stays open and unattributed. If it recurs, the diagnostic is a stack capture — and with this PR landed, a run-loop that ends on a cancel now leaves `session_halted` behind, which is the evidence that was missing the first time.

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -5523,24 +5523,54 @@ class Session:
                 try:
                     await self._turn_owner_task
                 except asyncio.CancelledError:
-                    if self._turn_cancel_self_initiated:
-                        # #2242 WAL-invariant 1: CancelledError unwound the
-                        # turn-body task straight out of whatever await it was
-                        # suspended on (mid-generation: the LLM await) — every
-                        # statement AFTER that await (parsing the response,
-                        # appending it to history, any further tool iteration)
-                        # never executes, so the cancelled turn's result is
-                        # never appended. Swallow here (do NOT re-raise) so the
-                        # driver task — and thus the agent — survives to serve
-                        # the next turn; only the per-turn sub-task was cancelled.
-                        _cancelled = True
-                    else:
-                        # Not our own cancel_inflight() call — this driver task
-                        # itself was cancelled from outside (e.g. an anyio scope
-                        # teardown for the MCP/A2A request-handler task pumping
-                        # run_one_iteration directly, FP-0013 §ADR-A). Preserve
-                        # the pre-#2242 behaviour: let it propagate.
+                    # #3377: the question here is WHOSE cancellation this is,
+                    # and the authoritative answer is not a flag — it is
+                    # whether THIS (the driver) task is itself being
+                    # cancelled. ``Task.cancelling() > 0`` is true only when
+                    # ``cancel()`` was called on the current task, which is
+                    # exactly the FP-0013 §ADR-A case (an anyio scope teardown
+                    # cancelling the MCP/A2A request-handler task that pumps
+                    # ``run_one_iteration``). The same discriminator is already
+                    # used by ``reyn.mcp.pool.is_real_control_flow`` and
+                    # ``reyn.core.cancellable.race_cancellable``.
+                    #
+                    # Before #3377 this branched on ``_turn_cancel_self_initiated``
+                    # alone, which conflates two very different things: a
+                    # cancel aimed at THIS task, and a cancel aimed at the
+                    # per-turn SUB-task by anyone who did not go through
+                    # ``cancel_inflight()`` (Ctrl-C plumbed to the turn, a
+                    # timeout, a stop-world operation, whatever is added
+                    # next — #3369 fixed one such source, not the class).
+                    # The second re-raised, which ended the while-loop in
+                    # ``run()``: the inbox kept being PUT to and was never
+                    # CONSUMED again, with no error log, indistinguishable
+                    # from a permanent hang.
+                    _driver = asyncio.current_task()
+                    if _driver is not None and _driver.cancelling() > 0:
+                        # A cancel genuinely directed at this task = a real
+                        # shutdown. It must still stop the loop; structured
+                        # concurrency requires the cancelled task to end.
                         raise
+                    # Only the per-turn sub-task was cancelled. #2242
+                    # WAL-invariant 1 holds either way: CancelledError unwound
+                    # the turn-body task straight out of whatever await it was
+                    # suspended on (mid-generation: the LLM await), so every
+                    # statement after that await never executes and the
+                    # cancelled turn's result is never appended. Swallow (do
+                    # NOT re-raise) so the driver task — and thus the agent —
+                    # survives to serve the next turn.
+                    if not self._turn_cancel_self_initiated:
+                        # #3377: a turn-scoped cancel we did not initiate is
+                        # survivable but NOT expected — record it, so this
+                        # never again presents as an unexplained silent stall.
+                        logger.warning(
+                            "turn sub-task for kind=%s chain_id=%s was cancelled by "
+                            "something other than cancel_inflight(); the turn is "
+                            "abandoned but the run-loop survives to serve the next "
+                            "message.",
+                            kind, payload.get("chain_id"),
+                        )
+                    _cancelled = True
             finally:
                 # #2242 Finding 1: reset the self-initiated flag UNCONDITIONALLY
                 # here (not only on the swallow branch). If cancel_inflight() set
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — **approve、auto-merge 装填。**

決め手を記録します:

- **discriminator を flag でなく `Task.cancelling() > 0` にした**のが正しい。「誰の cancel か」の権威ある答えは *この task 自身が cancel されているか* で、`_turn_cancel_self_initiated` 単独では「自分の cancel_inflight」と「第三者による turn sub-task cancel」を混同していました。既存 idiom（`mcp/pool.py` / `core/cancellable.py`）に揃えた点も含め妥当。
- **test が終端状態でなく中間断面を assert している**: 2通目の message が「put された」でなく「**consume された**」(`"c-second" in seen` + `inbox.empty()` + `not run_task.done()`)。終端だけの assert は #3369 类の欠陥を素通しします。
- **非対称の両脚が両方向に assert されている**: turn-scoped cancel → 生存＋記録、session-directed cancel → **停止**＋`halted_reason="cancelled"`＋audit-event。既存 #2242/FP-0013 契約 test が無変更で通っているので、変えたのは未カバーの第三ケースだけ、という主張が構造で裏づけられています。
- `session_halted` は**既存 vocabulary**（transport/TUI/renderer に消費者あり）への reason 追加であり、新 kind ではない — closed-vocabulary 4-site 義務は発生しません。
- **フリーズ報告を閉じると主張していない**: 「署名が一致する機構を1つ塞いだ。報告の決着には報告者環境の stack capture が要る」と本文冒頭に明記。もっともらしい単一機構で報告を閉じない、の遵守。

issue 側の framing 訂正（`chat_stopped`/`session_completed` は発火していたが clean shutdown と識別不能、が真のギャップ）も、報告を鵜呑みにせず測ってから直した正しい形です。

PR #3400, opened 2026-07-28
fix(mcp,tools): install_local wrong-workspace clobber + reyn_repo_read scope disclaimer — part of #3213

tya5:
**[per-PR-coder]** — part of #3213 (item 1 already landed separately in #3216).

Two items from the RAG turnkey UX polish follow-up list, both investigated and root-caused (not guessed):

## Item 2 — `mcp__install_local` clobbers `mcp.yaml` (FIXED)

Root cause (confirmed by reading the code, not by guessing): `_handle_mcp_install_local` (`src/reyn/tools/mcp_verbs.py`) resolved its project root via `getattr(workspace, "root", ...)` — an attribute the real `Workspace` class has never exposed (only `.base_dir`, per #1442). On the chat-router path `ctx.workspace` is `None`, so this silently fell back to `Path.cwd()`. Meanwhile its sibling verbs (`mcp__install_registry` / `mcp__install_package`) already route through `build_legacy_op_context` (the #1442 single-source bridge) and correctly resolve the agent's real workspace root.

When `cwd` differs from the real workspace root, `mcp__install_local` reads/writes a **different** `.reyn/config/mcp.yaml` than the other two install verbs. A later `install_local` call then reads an empty/stale file at the wrong path, and its own read-merge-write logic (which is otherwise correct — it does merge, not blindly overwrite) silently drops every server the other verbs had written to the real path. This exactly matches the observed symptom: after `reyn_chunker`/`reyn_vector_store` were registered, installing `reyn_markitdown` via `mcp__install_local` left the file with only the new entry.

Fix: `_handle_mcp_install_local` now builds its `OpContext` via `build_legacy_op_context(ctx)` (the same bridge its siblings use) and resolves `project_root` via `_resolve_write_root(op_ctx.workspace)` — the same helper `mcp_install.py`'s op path already uses. This makes all three install verbs converge on the same file by construction.

Strip-falsify: reverted the fix, confirmed the new regression test (`test_install_local_writes_under_the_factory_workspace_not_cwd`) goes RED (writes land under cwd, not the workspace root), then restored the fix and confirmed GREEN.

## Item 4 — wrong-tool discoverability (FIXED, cheap version only — tool-selection itself stays out of scope)

Live e2e (#3210) found the model defaulting to `reyn_repo_read` (Reyn's own repo-scoped dev tool) for a `~/.reyn/...` path, getting an out-of-scope failure, then fabricating file content rather than retrying with `read_file`. Per the brief, the deeper tool-selection fix is out of scope here; the cheap, safe mitigation is to have the tool's own description name its scope and the alternative.

Added an explicit scope disclaimer to `reyn_repo_read`'s description (`src/reyn/tools/descriptions/dev.py`, EN + JA): repo-root-relative paths only, never `$HOME` / the user's workspace / any other absolute path — use `read_file` for those. Regenerated the byte-identical-relocation baseline fixture (`tests/tools/_pre_migration_tool_schemas_baseline.json`) for this one intentional, single-line description change, and added a contract test asserting the disclaimer's presence.

This description text change has two knock-on effects, both fixed here:
- `tests/test_reyn_repo_unified.py::test_reyn_repo_read_router_render_exact_description` pinned a second, independently-typed literal copy of the old canonical text (the exact anti-pattern its own sibling test, `test_reyn_repo_list_router_render_exact_description`, already avoids by asserting against the imported constant). Fixed structurally — now asserts against `_REYN_REPO_READ_DESCRIPTION`, matching its sibling, so a future intentional text change won't need to touch this literal again.
- `tests/test_fp0063_arc_witness.py::test_llm_driven_install_ingest_query_arc_reaches_the_ingested_chunk` is an `LLMReplay` fixture test whose replay key hashes the full `tools=[...]` schema passed to `litellm.acompletion` — so any tool description text change invalidates any fixture recorded with that tool in the router's catalog. Regenerated via the test's own deterministic scripted-response path (`REYN_FP0063_ARC_WITNESS_GENERATE=1`, no real LLM call — confirmed by diffing the fixture: only the SHA-256 `key` field changed, every recorded `response` payload is byte-identical).

## Item 3 — `reyn chat --no-restore` does not fully reset (finding only, no fix — see #3213 comment)

Posting the finding + recommendation as a comment on #3213 rather than picking an implementation here, since the options differ materially in user-visible behavior (per the brief). (Update: the owner has since decided and #3406 fixed it directly, merged into `main` ahead of this PR — nothing left for this branch to do here.)

## Post-open CI fix — #3376's `tool_use_oracle_3376.json` regeneration

`#3402` (the #3376 P1 Exposure/Encoder seam) merged after this PR opened and brought `tests/scaffold/tool_use_oracle_3376.json` onto `main` — a byte-identical snapshot of the full `(llm_tools_payload, tool_use_sp)` every registered `(scheme x transport)` cell produces, which embeds every tool's complete description text. Item 4's `reyn_repo_read` disclaimer is exactly the class of change that invalidates it (same shape as the `_pre_migration_tool_schemas_baseline.json` regeneration above), so both pytest matrix jobs went red on an equality assertion between the live payload and the recorded snapshot.

**Fix**: regenerated via the oracle's own capture entry point (`tests/scaffold/tool_use_oracle_3376.py --out tests/scaffold/tool_use_oracle_3376.json`), which itself proved determinism across 3 fresh subprocesses before writing (its own `--check`/write path, not a claim I made separately).

**Minimality, verified structurally** — a full recursive leaf-diff of `origin/main`'s oracle against the regenerated one (not eyeballing the size of the textual diff): exactly **6 leaves changed** in the whole 4039-line/222,340-byte document, and every one of them is `reyn_repo_read`'s description text:
- 5 occurrences across 3 `llm_tools_payload` cells (`enumerate-all|tool_calls`, `retrieval|tool_calls|search_visible_false`, `retrieval|tool_calls|search_visible_true`) — the tool is catalog-projected under both its bare name (`reyn_repo_read`) and its qualified alias (`reyn_repo__read`), so it appears twice in two of those cells.
- 1 occurrence in the `enumerate-all|content_fence` cell's rendered CodeAct code-API prose (`tool_use_sp`), where the same description text is embedded as a doc-comment line.

No other cell, tool, field, or the pinned metadata (`pristineness_3383_fingerprints`, `pristineness_3385_canonical_equality`, notes, etc.) moved. `tests/scaffold/test_tool_use_oracle_3376.py` and `tests/test_tool_use_exposure_encoder_3376.py` (the two consumers) pass against the regenerated artifact.

**The toll, measured, not just asserted** — the oracle is **222,340 bytes with 430 `"description"` occurrences** (`grep -c '"description"'`; the pre-regeneration `origin/main` copy is 219,778 bytes, same occurrence count). That means **any deliberate LLM-visible tool-description edit, for as long as the #3376 arc holds this oracle frozen (through P2/P3), forces this same regeneration.** This PR's one-line disclaimer is the first instance of that toll, not a one-off.

**The alternative, and why it's rejected** — the obvious next proposal is to strip description text from the oracle (hash it, or replace with a placeholder) so the snapshot records structure/identity only and the toll drops to zero. **Rejected**: that would blind the oracle to exactly the `#3383` defect class it exists to catch. What litellm's in-place `tools[]` transform corrupted was the *payload contents themselves* — injecting `"type": "object"` onto a `const` discriminator, deleting `additionalProperties: false` from a `oneOf` variant. The oracle can detect "the Exposure/Encoder seam dropped or altered a description" only because it holds the literal text to compare against — a hash would still equal-or-not-equal correctly, but a placeholder/structure-only capture would never have shown *what* changed, and stripping to a hash loses the human-diffable evidence a reviewer needs to tell "a real corruption" from "an intentional one-line edit" apart in the first place (exactly the distinction this PR's diff had to make). The toll is the price of that detection working at all — expensive, but the alternative is cheap and blind, not free.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_3037_mcp_install_local_recovery_core_gate.py tests/tools/test_tool_description_relocation.py tests/test_reyn_repo_unified.py` — 21/21 OK, 0 Tier-4
- [x] `pytest` (full suite, from repo root, foreground, full output read — not `tail`) — **9488 passed, 36 skipped, 0 failed**, 1051s (first pass, pre-#3402-merge). Re-run after merging `main` (post `#3402`/oracle regeneration): **9623 passed, 36 skipped, 0 failed**, 698s, load average ~4-5 (quiet machine this time — noted for the opposite reason as before: this green is not contended, so it isn't hiding a timing flake). Both runs' full output read myself, not `tail`.
- [x] Strip-falsify: reverted item 2's fix, confirmed new test RED, restored, confirmed GREEN
- [x] `tests/scaffold/tool_use_oracle_3376.py --check` — deterministic across 3 fresh subprocesses (its own gate, run before recording)


tya5:
**[lead-coder]** — **approve。再生成2件を私の側で独立に検証しました。**

## ★ 再生成の検証（この PR で唯一の risk 箇所）

主張を鵜呑みにせず、`origin/main` 版と構造比較しました:

```
turn1_install.jsonl : 2 -> 2 records,  changed fields = ['key'] × 2
turn2_ingest_query  : 3 -> 3 records,  changed fields = ['key'] × 3
_pre_migration_tool_schemas_baseline.json : 1 leaf changed
  .reyn_repo_read.function.description
```

**レコード数不変・`key` 以外の全フィールド byte-identical・baseline は葉が1つだけ**。主張どおりです。★**再録済 fixture は「差分が小さい」ではなく「*何が* 変わったか」で見るべき**で、行数だけ見ると `key` の変更と payload の変更が区別できません。

⚠️ **`_pre_migration_tool_schemas_baseline.json` の再生成について1行記録**: #3383 では「**再生成するな**」と強く制約しました。あれは **product 側の非決定性を「意図された変更」として記録に埋めるな**という意味で、**#3385 が product を直した今、意図的な1行の description 変更に伴う再生成は正当**です。制約の理由が消えたので制約も消えます — が、**次に再生成する人が #3383 のコメントだけ読んで混乱しないよう**、ここに書いておきます。

## item 2（`mcp.yaml` clobber）

根本原因の突き止め方が正しい。**`getattr(workspace, "root", ...)` が *実在しない属性*** で、`Workspace` は `.base_dir` しか公開していない（#1442）→ chat-router 経路では `ctx.workspace` が `None` なので**黙って `Path.cwd()` に落ちる**。兄弟 verb 2つは `build_legacy_op_context` を通っており**別のファイルを読み書きしていた**。

★ **これは「同じ fix-class の兄弟が既に正しい形を持っていた」パターン**で、修正が**新しい仕組みの追加でなく既存 bridge への合流**になっているのが良い。3つの install verb が**構造上同じファイルに収束**します。strip-falsify（revert → RED → restore → GREEN）も確認済。

## item 4（wrong-tool discoverability）

description に scope を明記する cheap 版に留め、tool-selection 本体は範囲外を維持 — brief どおり。★**知らないうちに増えた2つの波及を、隠さず両方直した**のが良い:
- `test_reyn_repo_read_router_render_exact_description` が**独立した第2のリテラル複製**を pin していた（兄弟の `..._list_...` は既に定数参照だった）→ **兄弟に合わせて構造的に修正**。同じ text 変更が二度とこのリテラルを触らずに済みます。
- `LLMReplay` の replay key が `tools=[...]` 全体を hash するため、**任意の description 変更が fixture を無効化する**。これは**再録が正しい対処**で、上記のとおり検証済。

## item 3（`--no-restore`）

**finding のみ・未実装で正解**です（user-visible な挙動が選択肢間で実質的に違う）。`fresh` フラグが既に存在し `run-once` だけが立てている、という追跡も具体的。**owner 判断に上げます。**

Test plan 全消化、full pytest 9488 passed / 0 failed（load 9.58-13.10 の申告つき）。

tya5:
**[per-PR-coder]** — regenerated `tests/scaffold/tool_use_oracle_3376.json` for item 4's `reyn_repo_read` description change (the CI-red `#3402` brought onto `main` after this PR opened — full explanation + the measured toll + the rejected zero-toll alternative are now in the PR body). Structural minimality proof (full recursive leaf-diff, not diff-size eyeballing): exactly 6 leaves changed across the whole document, all `reyn_repo_read`'s description text, nothing else moved. Full suite re-run after merging `main`: 9623 passed, 36 skipped, 0 failed.

tya5:
**[per-PR-coder]** — CI is fully green (10/10 checks). The one interim red (`pytest (Python 3.11)`) after the oracle regeneration was two unrelated pre-existing flakes, confirmed by reading the job log directly rather than assuming: a Docker-registry OAuth-fetch failure in a devcontainer-build test, and a background-embed-sync timing race in `test_fp0066_p3b_repo_knowledge_ingest.py` (asserts a background task finishes within a fixed sleep window — a parallel-CI-worker-contention class of flake). Neither test touches any file in this diff. Re-ran via `gh run rerun --failed`; it passed clean the second time. Local full suite also green twice (post-#3402 merge: 9623 passed; post-#3412 merge: 9623 passed, 0 failed both times, full output read directly, not `tail`).

PR #3415, opened 2026-07-28
feat(tools): add the (category, content_fence) cell — #3376 P2

tya5:
**[per-PR-coder]** — part of #3376 (P3 remains: the last cell, `retrieval` x `content_fence`).

## What

Adds the `("category", CONTENT_FENCE)` cell — the one the owner asked for
("会社で使いたい"). P1 (#3402) built the Exposure/Encoder seam; this is what the seam
was for.

```yaml
tool_use:
  scheme: category
  transport: content_fence
```

The model writes fenced Python, but the functions it is shown are the catalog
**wrappers** (`list_actions` / `describe_action` / `invoke_action`) plus the base
tools — so a call reads
`result = invoke_action(action_name="file__read", args={"path": "README.md"})`
and the code-API does **not** grow with the catalog. That is the point: CodeAct
(`enumerate-all` x `content_fence`) shows every action; this cell keeps `category`'s
fold across the transport change.

## Shape

- **`schemes/_category_exposure.py`** (new) — the `category` presentation's Exposure,
  shared by both of its cells. The fold happens here, per the architect's ruling, so
  the encoder renders what is already folded. The `(category, tool_calls)` cell was
  **moved onto it**, not left beside it.
- **`schemes/_content_fence_cell.py`** (new) — the `content_fence` transport's own
  behaviour (fence extraction, sandboxed execution through the OS per-call gate, the
  observation turn), extracted from `codeact.py` unchanged. One transport
  implementation, not one per cell — the same rule `reyn.tools.encoders` states for
  encoders. `codeact.py` is now the enumerate-all *cell* and nothing else.
- **`schemes/category_content_fence.py`** (new) — the cell: `name` + `build_exposure`.
- **`transport.py`** — one registry row; the pair table stays a capability declaration.

`_format_codeact_observation` / `_extract_fenced_code` moved with the transport; their
four references (2 tests, `prompt/codeact.py`, `prompt/__init__.py`) are repointed in
this PR.

## The two hard constraints

**1. The four pre-existing cells are byte-identical.** Measured, not asserted:
`tests/scaffold/test_tool_use_oracle_3376.py` compares a fresh-process capture against
the recorded artifact and is green. The oracle JSON *is* touched — it has to be, because
the harness enumerates cells from the live registry and `test_oracle_covers_every_registered_cell`
requires the new cell to have an entry. It was regenerated with the harness, and the
diff is **pure addition**: `17 insertions(+), 0 deletions(-)`, one new `category|content_fence`
key. Every pre-existing cell entry and every top-level field (`notes`, the two pristineness
counts, the pinned inputs) compares equal.

**2. The fold invariant is asserted, three ways** —
`tests/test_tool_use_category_content_fence_3376.py`:

- `test_the_code_api_does_not_grow_with_the_catalog` — renders the cell over a 3-action
  catalog and a 300-action catalog and requires the code-API to be **identical**. Its
  own vacuity guard drives the `enumerate-all` cell over the same Ops and requires
  *that* one to declare 300 functions, so "category does not grow" distinguishes
  something.
- `test_the_fold_lives_in_the_exposure_so_no_encoder_can_undo_it` — the exposure handed
  to *either* encoder already carries only the wrappers.
- `test_production_folds_against_the_live_catalog` — no Fake: a real `Session` -> its
  `RouterHostAdapter` -> a real `RouterLoop` (which *is* `SchemeOps`) -> the registered
  scheme. Requires the code-API's declared set to equal what the `tool_calls` sibling
  advertises, and to be **disjoint** from the live catalog.

## Strip-falsify (all four RED, anchors confirmed `count == 1` before editing)

| Production call site stripped | RED |
|---|---|
| `build_exposure` reads `ops.catalog_entries()` instead of `ops.present(...)` | 4 arms, incl. the real-object-graph one |
| `dispatchable_catalog` narrowed to `exposure.descriptors` | `test_the_dispatch_gate_keeps_a_narrowed_wrapper` |
| `if not deviation.includes_base_tools:` guard removed | `test_a_deviation_that_cannot_be_honoured_is_refused` |
| the cell re-declares `interpret` instead of inheriting it | `test_both_content_fence_cells_run_the_same_transport_implementation` |

## Fail-closed, still

`resolve_scheme_for_transport` is unchanged and `(retrieval, content_fence)` is still
refused. Both the derived-complement arm in `test_tool_use_exposure_encoder_3376.py`
and the scaffold's arc-complete tripwire remain **non-vacuous** — verified by running
them, not by reasoning about them. **Three** witnesses had pinned
`(category, content_fence)` as *the* unregistered example and went RED on the full suite
(not in the targeted run — this is why the full suite is the gate):
`test_fp0066_p4a_transport_axis_3247.py` (registry census + strip-falsify),
`test_tool_use_scheme_1593.py` (config parse-time refusal), and
`test_discovery_mandate_stage_c_187.py`'s AST wiring gate (which pinned the *file* the
mandate call lives in). Each was retargeted so it keeps testing what it is about:

- the unregistered witness moves to `(retrieval, content_fence)`, plus an
  unknown-presentation arm so the file keeps a witness after P3;
- the config test gains its positive twin — `scheme: category` + `transport:
  content_fence` **parses**, which is the arm that says the documented yaml actually
  works (registering a cell and an operator's yaml accepting it are two facts);
- the AST gate now checks the exposure builder *and* that `universal_category` still
  reaches it — two claims, because dropping the second would let the scheme stop
  calling the builder while the first stayed green. Both arms strip-falsified RED.

Also swept for the same stale claim repo-wide (`git grep`, not a local-tree guess):
`config/execution.py`'s docstring named the cell as the unimplemented example;
`capability_visibility.py`'s scheme-registry key list was one short;
`test_scheme_self_register_1608.py`'s built-in set is a **subset** check, so an omitted
name is a hole rather than a failure — the new cell is listed there.

## Docs (same PR)

- `docs/reference/config/reyn-yaml.md` + `.ja.md` — the pair table said *(unimplemented)*;
  now shows the cell, with the yaml to write and when to choose it. The ja file asserted
  the same falsified claim, so it is fixed too (its own claim, not a mirror obligation).
- `docs/concepts/tools-integrations/tool-use-schemes.md` + `.ja.md` — a section for the
  cell and its selection block.
- `docs/concepts/tools-integrations/codeact.md` + `.ja.md` — a pointer from the
  content_fence "how to enable" section, where a reader looking for this lands.
- `docs/feature-map.md` — a row.
- `reyn.local.yaml.example` — the `tool_use` block's guidance (the surface an operator
  actually copies from).
- `src/reyn/tools/transport.py` + `src/reyn/config/execution.py` — module docstrings, the
  ASCII pair table, and `resolve_scheme_for_transport`'s docstring all named
  `category x content_fence` as the unimplemented example.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` (6 touched test files) — 49/49 OK, 0 errors
- [x] `python scripts/verify_module_docstrings.py` (9 touched src files) — OK
- [x] `pytest` from the repo root — **9637 passed, 36 skipped, 0 failed** (10:01), output read in full. The first run was RED on the 2 stale witnesses above; this is the run after fixing them.
- [x] Oracle: pre-existing cells byte-identical, diff is pure addition (see above)
- [x] Strip-falsify: 4 gates, each RED with the production call site stripped


tya5:
**[lead-coder]** — **approve、auto-merge 装填。★owner が求めたセルが動きます。**

## oracle — 私の側でも検算しました

```
added cells : ['category|content_fence']
removed     : []
PRE-EXISTING CELLS CHANGED: []
top-level fields differ    : []
```
**既存4セルは1バイトも動いていません。** P2 で最も危険だった点（**共有 seam を触るので既存セルが巻き添えで変わる**）は起きていません。

## ★ 私のブリーフが誤っていた点 — ご指摘が正しいです

> 「oracle を再生成するな」は達成不能。`_capture_cells` は `valid_scheme_transport_pairs()` から**捕獲時にセルを列挙**し、`test_oracle_covers_every_registered_cell` が**登録済セルすべてに entry を要求**するので、**セルを登録した時点で artifact は entry を得ざるを得ない**。

**そのとおりです。私の指示は不可能でした。** ★**しかも「不可能だ」で止めず、*追加のみ* であることを diff で証明する形に翻訳した**のが正しい対処です。**手で挿入せず harness で再生成した判断**（`canonical_json` 出力なので手編集は harness で再現できない）も正確 — **手編集していたら、次の再生成で差分が出て原因不明になっていました。**

★ これで **oracle の「通行料」は2種類**あると分かりました: **①description 編集（#3400）②セル追加（本 PR）**。**②は追加のみで既存が動かない**ので、**①より安全な通行料**です。P3 でも同じ形になるはずです。

## 畳み込み不変条件 — 要求より強い

★**3アクションと300アクションのカタログで code-API 文字列が *同一* であること**（「同程度の大きさ」ではなく同一）。**そして vacuity guard が同じ Ops で `CodeActScheme` を回して *そちらは300関数を宣言する* ことを要求**している — ∴ **「増えない」が何かを識別している**ことが示せています。**片方だけなら「常に空を返す実装」でも通りました。**

**production 軸を Fake ゼロで通した**（実 `Session` → `RouterHostAdapter` → 実 `RouterLoop`＝`SchemeOps`）のも要件どおりで、**code-API の宣言集合が `tool_calls` 兄弟の広告と一致し、かつライブカタログと *disjoint*** という assert は、**畳み込みが起きた後を描いている**ことの直接の witness です。

## ★ 発見: 3つの witness が `(category, content_fence)` を「未登録の例」としてハードコードしていた

**これが本 PR で一番価値のある所見**です。transport 軸の census / config の parse-time 拒否 / **`test_discovery_mandate_stage_c_187.py` の AST gate（mandate 呼び出しが *どのファイルに* あるかを pin していた）** — いずれも**セルを登録した瞬間に偽になる**。

★ P1 で私は「**未登録集合を registry から導出したので手書きの陳腐化を排除**」と評価しましたが、**導出したのは fail-closed arm だけで、*他の3テスト* は依然ハードコードでした**。**「1箇所を導出にした」は「その値がどこにもハードコードされていない」を意味しません。** 各々を「それが本当に検査したいもの」へ retarget し、config テストには**肯定側の双子**（documented yaml が実際に parse される）を足したのも正しい。

⚪ `test_scheme_self_register_1608.py` の built-in 集合が**部分集合検査**（∴ 漏れは失敗でなく穴）という指摘も、**次に built-in を足す人への申し送りとして価値**があります。

**docs も同一 PR で**（`reyn.yaml` reference EN+JA が literally *(unimplemented)* と書いていた面、schemes / codeact EN+JA、`feature-map.md`、`reyn.local.yaml.example`）。★**JA を触ったのは「その JA 自身が偽になる主張を持っていたから」**で mirror 義務ではない、という切り分けも正確です。

Issue #3421, opened 2026-07-28
tool_use: llm_tools_payload == [] cannot distinguish "no tools" from "this transport has no tools= channel"


tya5:
**[per-PR-coder]** — Follow-up filed while closing the #3376 arc (P3, #3422). Unblocked by that PR, deliberately **not** done in it.

## The ambiguity

`Presentation.llm_tools_payload` is `list[dict]`. For a `content_fence` cell it is `[]`, and that empty list carries two different meanings that the type cannot distinguish:

- **"there are no tools to show"** — a real, possible state; and
- **"this transport has no `tools=` channel at all"** — which is what it actually means.

Today the second meaning is stated in prose only. `ContentFenceEncoder.encode_tools` returns the empty list with a comment saying "Not 'there are no tools' — this transport has no `tools=` channel", and `_content_fence_cell.build_presentation` repeats it. A comment is the right place for a reason; it is the wrong place for a distinction the code has to act on.

This is the same empty-vs-absent shape the arc kept closing elsewhere (`turn_tokens=0` vs unknown, "not called" vs "cannot be called"), and the one instance it could not close while it was running.

## Why it was blocked until now

#3376's oracle (`tests/scaffold/tool_use_oracle_3376.*`) recorded the empty list **verbatim** as the current fact, and every PR in the arc had to be byte-identical against it. Changing the representation would have broken that criterion by design. The oracle was deleted in P3 (#3422), so the constraint is gone.

## Shape of the fix (not ratified — the point of the issue is to decide it)

Make the transport's inapplicability a typed value rather than an empty collection, in the repo's usual `{kind: ...}` discriminated-union idiom rather than a sentinel that reads like data. Consumers to sweep, at minimum:

- `router_loop.run` — `tools = _pres.llm_tools_payload` → `apply_contextual_visibility` → `self._catalog = {t["function"]["name"]: t for t in tools}`. For a `content_fence` cell this currently builds an empty catalog that is then overridden by `dispatchable_catalog`, which is exactly the coupling the ambiguity hides.
- the `RePresent` arm, which does the same thing on a re-present round.
- `capability_visibility`'s presentation probe.

★ Whatever the representation, the acceptance condition is that **"no tools are eligible" and "this transport has no tools= channel" become distinguishable at the type level** — not merely commented apart. A fix that keeps `[]` and adds a second field to explain it reproduces the problem in two places.

## Not urgent

[...]

PR #3419, opened 2026-07-28
fix(tools): give the CodeAct cell the same exposed set as its tool_calls sibling — Closes #3381

tya5:
**[per-PR-coder]** — Closes #3381.

The two `enumerate-all` cells now expose the **same set**. The receptacle #3376 P1 built held: the fix is the declared value, plus **one** wiring change it forces (§3).

## 1. Measured first — and the issue's headline claim is only half true

**Enumeration is AST, not regex**: every `_registry.lookup("<name>")` inside `build_tools` (`src/reyn/runtime/router_tools.py`), cross-checked against a runtime call with a maximal host config. Both give **33**, and the two sets are equal in both directions (`ast − runtime = ∅`, `runtime − ast = ∅`). Catalog names come from a real `SchemeOps` (`RouterLoop` over a real `Session`, the oracle harness's object graph).

| measurement | result |
|---|---|
| base tools `build_tools` can emit | **33** |
| of those, containing `__` | **0** — base names are unqualified, catalog names are `<category>__<entry>` |
| name collisions with the catalog | **0** |
| `build_actions_map` over base+catalog | bijective, all identifiers valid, **0** disambiguation suffixes fire, **0** existing catalog identifiers move |

🔴 **`read_file` and `delegate_to_agent` were reachable from CodeAct all along** — as `file__read` and `multi_agent__delegate`. Resolving both populations to their `ToolDefinition` (via `unwrapped_tool_name`, the oracle's own mapping) shows **18 of 33** base tools already had a qualified alias resolving to the *same definition*. The genuinely unreachable ones were the other **15**: `session_spawn`, `agent_spawn`, `topology_create`, `compact`, `present`, `render_template`, `describe_mcp_tool`, `call_mcp_tool`, and the seven MCP resource/prompt tools. So the gap is real but it is the spawn/topology/present family, not file reading. The issue's title is name-level true and capability-level false; that is recorded in `_enumerate_exposure.py`'s docstring so the next reader does not re-run this.

## 2. The branch taken: include them — uniformly, no dedup

Both cells now declare `includes_base_tools=True, excluded_names={"mcp__call_tool"}`, differing in **exactly one** field, `applies_contextual_narrowing`, whose reason is a property of the transport (nothing downstream narrows a rendered string).

Three things decided it:

[...]

PR #3422, opened 2026-07-28
feat(tools): add the (retrieval, content_fence) cell and retire the #3376 oracle — #3376 P3

### LINKED ISSUE #3376: tool_use: (scheme × transport) の未対応 2 セル（category×content_fence / retrieval×content_fence）— 埋めるか、埋めない理由を記録するか
**[lead-coder]** — owner 指示。`tool_use` の (scheme × transport) は 2×3 = 6 通りありますが、**実装済みは 4 通り**で、2 セルが未対応です。

## 現状（`src/reyn/tools/transport.py:95-100` の `_VALID_SCHEME_TRANSPORT_PAIRS`）

| scheme \ transport | `tool_calls` | `content_fence` |
|---|---|---|
| `category` | ✓ `universal-category` | ❌ **未対応** |
| `enumerate-all` | ✓ `enumerate-all` | ✓ CodeAct |
| `retrieval` | ✓ `retrieval` | ❌ **未対応** |

未登録セルは `resolve_scheme_for_transport` が **parse 時に fail-closed** で拒否します（黙って既定に落ちない）。**これ自体は正しい設計**で、本 issue は「拒否をやめろ」ではなく**セルを埋めるかどうか**の話です。

## 埋める意味（私の見立て、決定ではありません）

- **`category` × `content_fence`** — カタログ wrapper（`list_actions`/`describe_action`/`invoke_action`）を**コードとして書かせる**形。wrapper は元々「ツール数が増えてもツールリストを一定に保つ」ための機構で、CodeAct は「アクションを文中のコードで表す」機構。**両方とも「LLM に見せる面を小さくする」方向**なので、組み合わせは筋が通り得ます。
- **`retrieval` × `content_fence`** — 取得したアクションをコードで呼ぶ形。retrieval scheme が返す候補集合が**そのままコードの語彙**になるので、これも自然に見えます。

★ただし **どちらも「筋が通り得る」であって、需要の観測ではありません**。私は**使いたいという要求を一度も観測していません**。

## 判断が要ること

1. **そもそも埋めるべきか。** 未対応 2 セルは**現状ゼロコストで拒否されており、誰も困っていません**。埋めれば実装・テスト・doc の保守対象が 2 つ増えます。**「4/6 だから埋める」は理由になりません**（対称性それ自体は価値ではない）。
2. 埋めるなら**どちらを先に**か。
3. ★**埋めない場合も「未対応」で放置しない**: `_VALID_SCHEME_TRANSPORT_PAIRS` の隣に**なぜ未登録なのか**（未実装なのか、組み合わせとして無意味なのか）を書く。**コードの不在は「忘れた」と「決めた」を区別できません** — 現状はどちらとも読めます。

## 着手する場合の要件

- **fail-closed の設計を壊さないこと。** セルを足すのは登録を 1 行増やすことであって、拒否機構を緩めることではありません。
- **`resolve_scheme_for_transport` の enum 網羅 gate**（あれば）を、手書き列挙でなく **registry から導出**すること。手書きだと次のセル追加で漏れます。
- doc は EN/JA 両方（★JA は #3375 で別途、**削除済みキーを例示している**問題があるので先にそちらを直すこと）。

roi:low（需要未観測、現状で誰も困っていない）。

---

## Arc checklist — must be clear before this issue closes

*(**[per-PR-coder]** added while landing P1, #3402, at the coordinator's direction. Read this before closing.)*

- [x] **P1** — Exposure/Encoder seam under the 4 existing cells, byte-identical (#3402)
- [ ] **P2** — `category` × `content_fence`
- [ ] **P3** — remaining cell(s)
- [ ] 🔴 **Delete the byte-identical oracle scaffold** — `tests/scaffold/tool_use_oracle_3376.{py,json}` and `tests/scaffold/test_tool_use_oracle_3376.py`.

  It was **deliberately kept past P1** (its original `removed_by` named P1): the risk of the 4 pre-existing cells changing as collateral is *higher* in P2/P3 than in P1, because those PRs touch the shared seam the cells now run on. Deleting it at P1 would have removed the instrument mid-operation.

  **But it must not outlive the arc.** Once every cell is registered the artifact stops being a refactor instrument and becomes a frozen golden snapshot of current behaviour — the shape #3383 exposed, which resists correct changes instead of catching wrong ones.

  This item is **enforced, not merely promised**: `test_the_arc_completing_fires_this_scaffolds_deletion` goes RED the moment no unregistered `(scheme, transport)` cell remains, so the PR that registers the last cell **cannot merge** until the scaffold is gone. A `removed_by` written in prose is a promise nobody executes — #3383's baseline outlived its phase by 16 days that way.

- [ ] **Give `test_an_unregistered_cell_is_still_refused_fail_closed` a new witness.** Its vacuity guard trips on the *same* condition (no unregistered cell left to refuse). The two arms fire together; fixing either in isolation loses a guard. Replacement witness: a transport with no encoder, or an unregistered presentation name.

tya5:
**[per-PR-coder]** — Closes #3376. Every item on the arc checklist lands here; the arc's other PRs (#3392, #3402, #3415, and #3419 for the #3381 spin-off) are all merged, checked by enumeration rather than by assumption.

## The cell

`retrieval` over `content_fence` — the **search-first code-API**. The model is shown `search_actions` / `describe_action` / `invoke_action` plus the base tools, and **no `list_actions`**:

```python
hits = search_actions(query="read a file")
result = invoke_action(action_name="file__read", args={"path": "README.md"})
```

Withholding the enumeration verb is the whole difference from the `(category, content_fence)` cell. Both keep the surface small; `category` lets the model browse the catalog by name, `retrieval` requires it to describe what it wants. `describe_action` stays, because the model learns action *names* from the search and still needs their argument schemas — asserted, so a later "tighten the surface" change cannot strand it with names it cannot call.

### Why this pair is worth filling, and not just for symmetry

It reaches retrieval's paradigm **without** a `RePresent` round, and *could not* use one.

A `tools=` payload can only change **between** LLM calls — that is why the `tool_calls` cell intercepts its search and has the OS re-present the matches. This transport's whole tool-use surface is the system prompt, which `router_loop` builds once per turn (`messages[0]` is assembled before the iteration loop; the `RePresent` arm swaps `tools=` and the dispatch catalog, never the prompt). A re-presented code-API would have nowhere to land.

It does not need one: the search result is an ordinary value **inside the snippet**, so the search and the call happen in the same turn. Same paradigm, one round trip cheaper. `test_this_cell_reaches_the_paradigm_without_a_represent_round` asserts both sides of that asymmetry, so a future "unify the two retrieval cells" change has to confront it rather than discover it in production.

### Degrade

When the D14 gate reports the search unusable (no provider, index not ready yet), the cell renders base tools + the flat catalog instead of a search that returns nothing — the #2895 dead-session shape, which is worse on this transport because there is no `tools=` payload to fall back on.

### Why the two retrieval cells do not share an exposure builder

`category` and `enumerate-all` each have one builder feeding both cells. Retrieval cannot: its cells reach the same paradigm through structurally different means (intercepted synthetic search vs. dispatched real wrapper), so a shared builder would need a parameter meaning *"does this cell intercept the search or dispatch it"* — a per-cell composer wearing an `ExposureDeviation`'s clothes, which `reyn.tools.encoders` says not to build. They share what genuinely is common: the transport-neutral `sp_facts`. The decision and its falsifying condition ("the OS rebuilds the system prompt on a RePresent round") are written down in `_retrieval_exposure`, because absence cannot distinguish forgotten from decided.

## Retiring the oracle scaffold

`test_the_arc_completing_fires_this_scaffolds_deletion` went RED the moment the cell was registered — as designed — and the instruction it carries is to **delete** the artifact. It is deleted, not silenced: `tests/scaffold/tool_use_oracle_3376.{py,json}` and `tests/scaffold/test_tool_use_oracle_3376.py` are gone (-5021 lines).

Before deleting, both facts were recorded:

- **Byte-identical**: with the new cell registered and the artifact regenerated, `test_live_capture_matches_the_recorded_oracle` (fresh-subprocess capture vs. recorded artifact) and the other six arms were **green**; only the arc-complete tripwire was RED.
- **Addition-only**, proven leaf by leaf against `origin/main`'s artifact rather than by reading a diffstat:

  ```
  old leaves: 1689   new leaves: 1713
  removed: 0
  changed: 0
  added:   24   → all under retrieval|content_fence|search_visible_{false,true}
  ```

  (The harness enumerates cells from the live registry, so regenerating is unavoidable; the new cell is captured in **both** branches of its D14 fork, since recording only the default would make the artifact assert this cell looks like `enumerate-all`.)

## The durable negative examples

Three tests had pinned a cell as *the* unregistered example. P2 had already retargeted them once, when it registered `(category, content_fence)`. Retargeting to `(retrieval, content_fence)` a second time would have been the **third** temporary fix.

**Value-side AST census first**, before deciding what to fix — the literal, not the concept, across `tests/`:

| | sites |
|---|---|
| `(retrieval, content_fence)` written as a literal pair (tuple / call-args / dict) in `tests/` | **3** — `test_fp0066_p4a_transport_axis_3247.py` ×2, `test_tool_use_scheme_1593.py` ×1 |
| same in `src/` | 0 (prose only: `transport.py`, `config/execution.py`) |

All three are now **positive** examples (a registered cell), and the census re-run confirms zero negative-example uses of the pair remain.

The replacement is `NOT_A_PRESENTATION` — a name that is not on the presentation axis **at all**:

> A negative example must come from outside the space the system extends into — something that cannot **enter** the set, not something merely not in it yet.

`(retrieval, content_fence)` was never forbidden; it was a *legal* pair that had not arrived, in an arc whose purpose was to make it arrive. That is why it was dated on the day it was written.

It lives in `tests/_support/tool_use_negative_examples.py`, which is also the **mark** the architect asked for: a negative example is written exactly like a positive one, so importing it from there is what makes it identifiable — and `test_the_negative_example_is_outside_the_namespace` turns "off-axis" from a comment into a checked claim. No expiring witness is left, so no K-inline expiry declaration was needed.

The general rule is recorded in the testing policy (`testing.md` + `testing.ja.md`, "Choosing a negative example"), together with the second lesson: **making one site derive its set from the registry is not a claim that the value is nowhere hardcoded** — grep from the value side after any such conversion.

## Strip-falsify

Every anchor confirmed `count == 1` before editing; each strip removed a **production** call site and was reverted with a targeted edit.

| Production call site stripped | Result |
|---|---|
| `excluded_names` no longer drops `list_actions` | 3 arms RED |
| exposure derived from `catalog_entries` instead of `ops.present` | 5 arms RED |
| the `search_visible` degrade guard removed | 2 arms RED |
| `dispatchable_names` narrowed to the exposed set | 🔴 **GREEN — a hole** |
| `_sp_facts` grows a divergent copy | 1 arm RED |
| the cell re-declares `interpret` instead of inheriting it | 1 arm RED |
| the off-axis witness registered as a real cell | 4 arms RED |

★ **The green strip was the useful one.** It showed my gate covered only the `dispatchable_catalog` half of the pre-narrowing rule and not the **identifier-map** half. `test_a_withheld_row_does_not_shift_the_identifier_of_a_shown_one` closes it with a constructed collision: with the map built over the exposed subset, `list~actions` slides from `list_actions_2` up into `list_actions` — the identifier of the very action being withheld — and the sandbox stub, named from the full set, routes the call to the other one. Silent, and permission-adjacent. RED under the strip, green with it restored.

## Docs (same PR)

- `docs/reference/config/reyn-yaml.md` + `.ja.md` — the pair table said *(unimplemented)* / *(未実装)* for this cell; both now show it, with the yaml to write and when to choose it. The JA file asserted the same falsified claim, so its own claim is fixed (not a mirror obligation).
- `docs/concepts/tools-integrations/tool-use-schemes.md` + `.ja.md` — a section for the cell, the round-trip asymmetry, and the selection block. The `retrieval` section's "the OS re-presents" sentence is now scoped to the `tool_calls` cell.
- `docs/concepts/tools-integrations/codeact.md` + `.ja.md` — a pointer from the "how to enable" section, where a reader looking for this lands.
- `docs/feature-map.md` — a row.
- `reyn.local.yaml.example` — the `tool_use` block's guidance.
- `docs/deep-dives/contributing/testing.md` + `.ja.md` — the negative-example rule above. Added to both because CLAUDE.md points readers at the `.ja` copy first, so an EN-only normative rule would be half-landed.
- `src/reyn/tools/transport.py`, `src/reyn/config/execution.py`, `src/reyn/runtime/capability_visibility.py`, `src/reyn/tools/schemes/retrieval.py` — module docstrings, the ASCII pair table, and the "unregistered cell, e.g. …" examples, which are now taken from off-axis for the same reason as the tests. `retrieval.py`'s "the **only** scheme that uses RePresent" claim is now scoped to its cell.

## Follow-up filed, not deferred silently

#3421 — `llm_tools_payload == []` cannot distinguish "no tools" from "this transport has no `tools=` channel". It was blocked by the oracle's byte-identical criterion and is unblocked by this PR; deliberately out of scope here.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` (6 touched test files) — 57/57 OK, 0 errors. (First run flagged one Tier 4: a private-state assert on `_sp_facts`; rewritten to assert through the rendered system-prompt slots, and re-strip-falsified RED to confirm the rewrite is still load-bearing.)
- [x] `python scripts/verify_module_docstrings.py` (7 touched src files) — OK
- [x] `pytest` from the repo root — **9654 passed, 36 skipped, 0 failed** (10:21), whole log read (grepped for `FAILED`/`ERROR`: none)
- [x] `pytest tests/test_3126_doc_anchor_gate.py` — 319 passed, run after the doc edits
- [x] Oracle: five pre-existing cells byte-identical, artifact change addition-only (0 removed / 0 changed / 24 added leaves vs `origin/main`)
- [x] Oracle scaffold deleted, not silenced — the tripwire's RED was observed first, and its instruction followed
- [x] Strip-falsify: 7 gates; 6 RED, 1 green → gate widened, then RED


tya5:
**[lead-coder]** — **approve、auto-merge 装填。★#3376 アーク完了。そして本 PR で一番価値があるのは、緑で返った strip を *穴* として扱ったことです。**

## ★ 7本中1本が GREEN → それを安全な脚と読まなかった

> `dispatchable_names` を露出集合に絞る strip が **GREEN**。**gate が pre-narrowing 規則の `dispatchable_catalog` 側だけを覆い、identifier-map 側を覆っていなかった。**

★**本アークで私が繰り返し述べてきた「緑の strip は安全な脚ではなく gate の穴である可能性の方が高い」が、実際に発火して正しく処理された初回**です。しかも**構成した衝突で証明している**: `list~actions` が `list_actions_2` から `list_actions`（＝**withhold した action の識別子**）へ滑り込み、**sandbox stub は別へ routes する** — **静かで、しかも permission 隣接**。**その後 RED を確認**。**「緑だったので大丈夫」で止めていたら、この穴は残っていました。**

## scaffold は削除された（黙らされていない）

```
tests/scaffold/tool_use_oracle_3376.py    +0/-569
tests/scaffold/tool_use_oracle_3376.json  +0/-4056
tests/scaffold/test_tool_use_oracle_3376.py +0/-253
```
★**自己終端 arm が RED になり、その指示に従って消した**。**arm を消して緑にする**という安易な逃げを取っていません。**しかも byte-identical arm が緑であることを *先に* 観測してから削除**しているので、**削除が tripwire だけでなく測定に裏づけられて**います。

## oracle — addition-only

```
old 1689 leaves → new 1713;  removed 0, changed 0, added 24
（すべて retrieval|content_fence|search_visible_{false,true} 配下）
```
既存5セルは byte-identical。**私は削除後の artifact を検算できません**（消えたので）が、**「削除前に緑を観測した」という順序が、それを代替しています** — この順序が要件だったのはまさにこのためです。

## ★ 否定例の恒久化 — 「印」まで実装された

値の側からの AST census で**ハードコード3箇所**（`test_fp0066_p4a_transport_axis_3247.py` ×2 / `test_tool_use_scheme_1593.py` ×1、`src/` は散文のみ0）。3件とも**登録済セルの *肯定* 例**に転換。

置換 witness `NOT_A_PRESENTATION` を **共有モジュール**に置き、★**「そのモジュールから import すること自体が *印* になる」**という設計 — architect が「どれが否定例かを機械が判別できないので印が要る」と述べた点への直接の回答です。**印が無ければ gate は不可能**、という彼の限界指摘を**設計で解いて**います。さらに **off-axis であること自体を assert する arm** 付き ＝ **「名前空間の外」がコメントでなく検査された主張**に。

**期限つき witness はゼロ**なので K-inline の期限宣言は不要、という判断も正確です。

## ⚪ stale main を自分で捕まえた

> 最初の oracle 比較は **stale な `origin/main` に対してで、2 leaf の *偽の* 変更**を示していた

★ #3419 が作業中に着地してシグネチャを変えたためで、**resync して初めて正しい比較になった**。**本日 broadcast した「行番号は動く main を跨げない／完全性主張は `origin/<ref>` に対して」の、oracle 比較版**です。**気づかず報告していれば「既存セルが2つ動いた」という偽の発見になっていました。**

⚪ **#3421 起票**（`llm_tools_payload == []` の曖昧さ、本 PR が unblock したが範囲外）と **`testing.md` / `.ja.md` への「否定例の選び方」追記**も、規則6どおりの残件処理です。

CI 11/11 green。マージします。

PR #3453, opened 2026-07-29
fix(runtime): remove dead embedding_event_sink seven-hop wire — no producer (#3438)

### LINKED ISSUE #3438: embedding_event_sink: a seven-hop audit wire with no producer — the only KIND_FAMILY hole in the closed kind vocabulary
**[per-PR-coder]** — found while closing the audit-event kind vocabulary (#3410 / PR #3437). Filed rather than fixed there: the fix is a threading removal with its own design question, and #3437 is already a gate + vocabulary change.

## The finding

`src/reyn/runtime/session.py::_build_retrieval_bundle` defines an audit sink:

```python
def _embedding_event_sink(kind: str, text: str, meta: dict) -> None:
    try:
        self._chat_events.emit(f"embedding_{kind}", text=text, **meta)
    except Exception:
        pass
```

★ **Nothing ever calls it, so no `embedding_*` audit-event kind is emitted anywhere.**

Measured, not inferred:

- The sink reaches a provider only through `reyn.data.embedding.get_provider`, which passes it **conditionally**: `if event_sink is not None and "event_sink" in inspect.signature(cls).parameters`.
- **No embedding provider in the repo accepts that kwarg.** `LiteLLMEmbeddingProvider` is the sole implementation, and `get_provider`'s own docstring says it "does not accept it (no lazy-load lifecycle to report on)".
- Both entry paths hit the same check: the session-held `ActionEmbeddingIndex` construction, and the `embed` op's `_resolve_provider(event_sink=ctx.embedding_event_sink)`.
- Consequently `AUDIT_EVENT_KINDS` (#3437) declares **no** `embedding_*` kind — the vocabulary gate's liveness arm would go RED on any that were added.

## Why it is worth a ticket rather than a shrug

Two costs, both small individually:

1. **A live-looking wire that is dead.** `embedding_event_sink` is threaded through `Session` → `_RetrievalBundle` → `RouterHostAdapter` → `make_router_op_context` → `OpContext.embedding_event_sink` → `op_runtime/embed.py::_resolve_provider` → `get_provider`. Seven hops of parameter, each commented as FP-0057 #2856 Part A, carrying nothing. The next reader tracing "how does embedding progress surface in the TUI?" follows all seven before finding the `inspect.signature` gate at the end.
2. **It is the only genuine `KIND_FAMILY` blind spot in the closed vocabulary.** #3437 registers it in `DYNAMIC_KIND_EMIT_SITES` because `f"embedding_{kind}"` cannot be read by the AST census. It is currently a hole with nothing behind it — but the moment a provider *does* accept the kwarg, kinds start flowing through a site the vocabulary gate cannot see, and the closed namespace silently stops being closed.

## The decision this needs

Not mechanical — pick one:

- **(a) Remove the threading.** Clean end-state: delete the sink, `OpContext.embedding_event_sink`, and the seven-hop parameter. Cheapest to read; costs the wiring if a lazy-loading local provider lands later.
- **(b) Keep it and make it real.** If a local/offline embedding provider with a download lifecycle is on the roadmap (`docs/concepts/data-retrieval/rag.md` discusses local models), decide the kind names **now** and pass them as **literals** — `emit("embedding_model_download_started", ...)` — so they enter `AUDIT_EVENT_KINDS` and the blind spot closes by construction rather than staying registered.

★ Whichever is chosen, **the `KIND_FAMILY` entry in `DYNAMIC_KIND_EMIT_SITES` should not survive it** — under (a) it is deleted with the sink, under (b) it is replaced by literals. A permanently-registered hole in a closed vocabulary is a contradiction that only stays honest as long as someone re-reads the reason.

## Pointers

- `src/reyn/runtime/session.py` — `_build_retrieval_bundle` (the sink), `_embedding_event_sink`
- `src/reyn/data/embedding/__init__.py` — `get_provider`'s signature gate
- `src/reyn/core/op_runtime/embed.py` — `_resolve_provider`
- `src/reyn/core/events/event_schema.py` — the `KIND_FAMILY` entry in `DYNAMIC_KIND_EMIT_SITES` and its reason

tya5:
**[per-PR-coder]** — fixes #3438.

## Decision: (a) remove the threading

Removed the whole seven-hop wire — `Session._embedding_event_sink` closure →
`_RetrievalBundle.embedding_event_sink` → `RouterHostAdapter`'s
`embedding_event_sink` param/attr → `build_router_op_context`'s
`embedding_event_sink` param → `OpContext.embedding_event_sink` →
`embed.py`'s `_resolve_provider(event_sink=...)` → `get_provider`'s
`event_sink` kwarg (and the `inspect.signature` gate it existed for). Also
removed the `KIND_FAMILY` entry in `event_schema.py`'s
`DYNAMIC_KIND_EMIT_SITES` that registered this as the closed audit-event
kind vocabulary's one blind spot (#3437).

## Independent re-verification of "producer is zero"

Confirmed by AST/signature inspection, not by running anything:

- `src/reyn/data/embedding/__init__.py`'s `_PROVIDERS` registry has exactly
  two entries, both `LiteLLMEmbeddingProvider` — `__init__(self, config:
  "dict[str, Any] | Any | None" = None)`, no `event_sink` parameter.
- Repo-wide grep for `class.*EmbeddingProvider` / `register_provider(` turns
  up only test `Fake*EmbeddingProvider` classes (`tests/test_op_embed.py`,
  `tests/test_op_semantic_search.py`, `tests/test_action_embedding_index.py`,
  etc.) and one production `register_provider("test-custom", _MyProvider)`
  in `tests/test_embedding_provider.py` — none of these classes accept
  `event_sink` either.
- `get_provider`'s own `inspect.signature(cls).parameters` gate is therefore
  never satisfied by any class currently reachable through the registry, in
  or out of tests.

∴ zero `embedding_*` audit-event kinds are emitted today, matching the
issue's MEASURED finding.

## Search for a recorded "keep for later" intent — none found

- The seam's own history (`FP-0057 #2856 Part A`) explains *why* it was
  built: to preserve a TUI model-download status sink (`FP-0043` Component
  C.3, a local in-process embedding model's lazy-load lifecycle) once
  `ActionEmbeddingIndex` was routed through the shared `embed` op instead of
  calling the provider direct.
- That reason stopped applying when **#3128** ("Remove in-process
  sentence-transformers embedding backend — delegate local embedding to
  litellm proxy") removed the in-process backend entirely. Per
  `docs/concepts/data-retrieval/rag.md` § Local and offline embedding
  models, local/offline embedding today goes through an **operator-run
  litellm proxy** — reyn does not manage a download lifecycle for it, so
  there is nothing for this sink to report even in the "local model" case
  the seam was built for.
  - I did not find any comment, ADR, or open issue proposing a future
    provider that *would* consume `event_sink` — i.e. no recorded intent to
    keep this wire for later. Per the issue's own framing, an absent record
    is treated as "forgotten," not "decided," so (a) is the correct call.

## Verification

1. **Zero production callers, independently checked** — see the
   AST/signature re-verification above; every one of the 7 hops was grepped
   end-to-end before deletion (`embedding_event_sink` no longer appears
   anywhere in `src/` or `tests/` except 3 comments recording *why* it was
   removed, left as breadcrumbs for a future reader).
2. **Full suite green after deletion**: `9707 passed, 36 skipped, 44
   warnings in 583.95s` (`uv run pytest -n auto`, matching CI's `-n auto`).
   Two tests that existed specifically to pin the forwarding
   (`test_embed_forwards_ctx_embedding_event_sink_to_provider_resolution` /
   `test_embed_forwards_none_when_ctx_embedding_event_sink_stripped` in
   `tests/test_op_embed.py`) were deleted along with the mechanism they
   tested — their existence confirms the wiring wasn't hypothetical, just
   producer-less.
3. **`DYNAMIC_KIND_EMIT_SITES`'s `KIND_FAMILY` entry removed**, and
   `tests/test_audit_event_kind_vocabulary_3410.py` (part of the full-suite
   run above) stayed green — the #3437 gate does not regress.
4. **Strip-falsify N/A** — this PR *is* the deletion; there is no
   "delete it and see if a test goes RED" step to run beyond the two tests
   removed in (2), which is the expected signal, not a falsification gate.

## Local gates (all green)

- `ruff check src tests`
- `python scripts/test_tier_audit.py --strict tests/test_op_embed.py`
- `uv run pytest -n auto` (full suite, see above)
- `python scripts/verify_module_docstrings.py <8 changed src files>`

`origin/fix/3438-remove-embedding-event-sink-dead-wiring` = `db22b25236925f9558f57778a46d6d51687b05cc`

## Test plan

- [x] `test_tier_audit.py --strict tests/test_op_embed.py` — 15/15 OK, 0 errors
- [x] `uv run pytest -n auto` (full suite) — 9707 passed, 36 skipped, 0 failed
- [x] `verify_module_docstrings.py` on all 8 changed `src/` files — OK
- [x] Confirmed `DYNAMIC_KIND_EMIT_SITES` no longer declares an
      `embedding_*` `KIND_FAMILY` hole, and #3437's vocabulary-gate test
      stays green

### DIFF
--- a/docs/reference/runtime/session-construction.md
+++ b/docs/reference/runtime/session-construction.md
@@ -268,14 +268,14 @@ Two other cost-adjacent construction points stay inline:
 
 ## Family 5 — Retrieval
 
-`_build_retrieval_bundle` constructs the embedding block (four attrs:
-`action_embedding_index`/`embedding_provider`/`embedding_model_class`/
-`embedding_event_sink`) plus `action_usage_tracker` — a byte-identical extraction (same
+`_build_retrieval_bundle` constructs the embedding block (three attrs:
+`action_embedding_index`/`embedding_provider`/`embedding_model_class`) plus
+`action_usage_tracker` — a byte-identical extraction (same
 objects, same conditionals, same try/except None-fallbacks, same args as the inline
 sequence it replaced). It stays UNMOVED, invoked at its original position, BEFORE Family 1
 (`_build_audit_event_bundle`) runs — this family has no eager dependency on `chat_events`,
-only the two closures' DEFERRED `self._chat_events` resolution (kept verbatim since this is
-an instance method).
+only the `_on_hot_list_changed` closure's DEFERRED `self._chat_events` resolution (kept
+verbatim since this is an instance method).
 
 `_action_retrieval` (FP-0034 PR-3b-iii) drives whether the universal catalog wrappers
 appear in the router `tools=`. Default constructs an off-flag `ActionRetrievalConfig` so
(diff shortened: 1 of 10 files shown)


tya5:
**[lead-coder]** — レビュー。**指摘なし。CI 緑を確認してマージします。**

## ★ 特に評価している点

**1. 「無い」の証明を、複数の独立した角度から立てたこと**
- **AST/signature で、`event_sink` を受け取る provider がゼロ**（実行時でなく静的に）
- ★ **「後で使う」という *意図の記録* を探して、見つからなかったと明記**
- ★ **seam の履歴を辿り、由来（FP-0043 C.3 のローカル埋め込みモデルの download lifecycle）が **#3128 で既に削除済み** と特定**

★ **3つ目が決定的です。** **「producer が無い」だけなら「まだ書かれていない」かもしれませんが、★ *その producer が存在した理由そのものが消えている* なら、来ることはありません。** **「不在」を「削除済みの機構の残骸」として説明できています。**

**2. ★ 削除したテスト2件についての観察**

> **their existence confirms the wiring wasn't hypothetical — just producer-less**

★ **これは鋭い。** **配線を pin するテストが *存在した* ということは、誰かが「これは動く」と考えていた証拠**です。∴ ★ **「机上の設計が残っていた」ではなく「動くと思われていたものに producer が無かった」**。**削除の正当性がむしろ強まります。**

**3. #3437 の盲点レジストリからも `KIND_FAMILY` エントリを消し、gate が緑のままであることを確認したこと**

★ **消し忘れていたら、存在しない盲点を宣言し続けることになりました。** **本日 #3437 で作ったばかりの機構を、正しく畳んでいます。**

**4. strip-falsify を「N/A」と明記したこと** — ★ **削除 PR で「消したら RED」は自明なので、無理に作らなかったのが正解**です。**私はそう指示しましたが、指示どおりにやらずに形だけ作る例も多いので、明示的に評価します。**

## 記録として

★ **本 issue は #3437 の調査中に見つかったものです。** **閉じた語彙を作ったことで、「動的に組み立てられる kind」を登録せざるを得なくなり、★ その登録簿の中で唯一 producer を持たないものとして露出しました。**

★ **これが「構造的対処」の効き方の実例です** — **gate を作った副産物として、gate が無ければ見つからなかった死んだ配線が1本出てきました。**

PR #3456, opened 2026-07-29
fix(runtime): remove dead task_subscribed/task_rebound WAL kinds + general unknown-kind gate

### LINKED ISSUE #3436: task_subscribed/task_rebound: declared WAL kinds with zero writers — decide fate, check old-WAL reader tolerance first
**[docs-maintainer]** — filed per lead-coder's co-vet finding on PR #3433 (deferred from the doc-fix PR to avoid removing a declaration without checking reader tolerance).

## What
`task_subscribed`/`task_rebound` are declared in `src/reyn/core/events/state_log.py`'s WAL event-kind vocabulary (#2187 backend-master — meant to record the Reyn-internal task↔session binding, assignee/requester, so it can be rewound). `git grep "task_subscribed\|task_rebound" -- src/reyn tests` returns **zero hits outside the declaration itself** — no writer anywhere appends either kind. Same defect class as #3357 (declared vocabulary member with no live producer), manifesting at the WAL layer instead of the audit-event layer.

## Consequence
`docs/concepts/runtime/events.md`/`.ja.md` and `docs/concepts/runtime/time-travel.md` both asserted, in the present tense, that task-binding changes ARE recorded to the WAL and DO get restored on rewind — false today. Those doc claims are being corrected in the same PR (#3433) to state this as a design intent that currently has no writer, not a live behavior.

## What this issue is for
Decide the fate of `task_subscribed`/`task_rebound` in `WAL_EVENT_KINDS`:
1. **Wire a real writer** (if the task↔session binding rewind behavior is still wanted) — implement the append site(s) that currently don't exist.
2. **Remove the declaration** (if the behavior is no longer wanted) — but first check whether any existing `.reyn/state/wal.jsonl` files in the wild could contain these kinds from an earlier, possibly-live version, and whether the WAL reader/replay path tolerates an unknown/removed kind gracefully. Removing the declaration without that check risks breaking replay of old WAL data that happens to contain a stray `task_subscribed`/`task_rebound` entry from before this drift was introduced.

Not resolved here — this issue exists so the decision doesn't rot as an unrecorded remainder (per CLAUDE.md rule 6).

tya5:
**[per-PR-coder]** — closes #3436.

## What

Two-part PR, per the owner's "class not hole" instruction (part 1 alone would be a band-aid; part 2 is the body).

### 1. Remove the dead declaration

`task_subscribed` / `task_rebound` removed from `WAL_EVENT_KINDS` in `src/reyn/core/events/state_log.py`. #2187 declared them for the Reyn-internal task↔session binding (assignee/requester); #3214 removed the only intended writer (`SubscriptionRegistry`) with no append site ever wired — zero live producers, per the measurement in the issue thread (`git grep` returns 0 hits outside the declaration). Replaced the stale comment block explaining the (never-live) writer contract with one explaining the removal + what a future re-add would need (a real writer AND a `route_key`, since the real on-disk entries carry no `target`/`agent` field and are already inert under `AgentSnapshot`/`registry.restore_all` routing regardless of this declaration's fate).

Also updated, since the declaration's removal makes the #3433 wording stale again (read the whole section, not just the keyword):
- `docs/concepts/runtime/events.md` / `.ja.md` — "Task subscription events" section: was "declared as WAL kinds ... but no writer" (accurate pre-#3436), now "were designed ... removed in #3436, not recorded anywhere today."
- `docs/concepts/runtime/time-travel.md` — "Task subscriptions and the backend at rewind" section: same shift from "designed but no writer" to "designed, writer never wired, kinds removed in #3436."

No JA `time-travel.ja.md` section covers this topic (checked — file exists but has no matching content), so no JA-mirror obligation there (CLAUDE.md's no-blanket-mirror rule).

### 2. General-form reader-tolerance gate (the body of this PR)

Added `tests/test_3436_unknown_wal_kind_reader_tolerant.py`, replacing the bespoke `test_removed_skill_wal_kinds_recovery_safe` (added in #2507 for the `skill_*` removal) with a **general form**: "a WAL line whose `kind` is absent from `WAL_EVENT_KINDS` reads through every consumer with no exception, and its neighbors survive" — instead of a per-removal bespoke test + bespoke measurement pass (this is the *second* such removal; #2507 was the first, and the issue's premise was that a third would otherwise repeat the same measurement).

**No hardcoded kind name.** The out-of-vocabulary probe is `f"unregistered_kind_{uuid.uuid4().hex}"`, asserted `not in WAL_EVENT_KINDS` as a sanity check — not a plausible-sounding future kind name (guessing one risks exactly the collision CLAUDE.md flagged elsewhere today) and not `task_subscribed`/`task_rebound` themselves (which would make the test re-bespoke to this removal).

**All three reader paths the issue's measurement covered:**
- `StateLog.iter_from` — `test_iter_from_tolerates_unregistered_kind_and_keeps_neighbors`
- `WalTailReader.poll` — `test_wal_tail_reader_tolerates_unregistered_kind`
- `AgentSnapshot.apply_events` — `test_agent_snapshot_apply_events_noops_unregistered_kind`
- plus the write-side half (`append` rejects any unregistered kind) — `test_write_side_rejects_unregistered_kind`

**Why this makes a third removal need no new measurement/test**: the probe kind is drawn from outside the vocabulary at test-run time, not fixed to any specific literal — so the property under test is "any kind not currently in `WAL_EVENT_KINDS`", which is exactly the shape of the *next* removal too. A future PR removing some third dead kind can point at this module instead of writing a new one.

**Disposition of `test_removed_skill_wal_kinds_recovery_safe` (#2507): REMOVED.** Its two claims are both strictly subsumed:
- (a) write-side reject of a removed kind → generalized by `test_write_side_rejects_unregistered_kind` (the specific `skill_*` literals added no guarantee beyond "any kind not in `WAL_EVENT_KINDS` raises at append", which the general test already proves for whichever synthetic kind it draws).
- (b) read-side tolerance + neighbor-preservation on `iter_from` → generalized by `test_iter_from_tolerates_unregistered_kind_and_keeps_neighbors`.

Checked whether the removed test additionally pinned some `skill_*`-specific *dispatch* behavior in `AgentSnapshot._apply_one` — it does not; there was never a `skill_*` branch in `_apply_one` (before or after #2507), so nothing skill-specific was being protected beyond the two generic claims above. Nothing is lost.

**Strip-falsify (RED observed, then restored — no `git stash`/`git checkout --`, cp-scratch backup + restore):**
1. Backed up `state_log.py` and `agent_snapshot.py` to the scratchpad.
2. Confirmed anchor uniqueness (`count == 1`) before each edit.
3. `iter_from`: inserted `if entry.get("kind") not in WAL_EVENT_KINDS: raise ValueError(...)` right before its `yield entry` → ran `test_iter_from_tolerates_unregistered_kind_and_keeps_neighbors` → **RED**, `ValueError: unknown WAL event kind: 'unregistered_kind_...'` (observed, not inferred). Restored from scratch-copy; confirmed `git diff` clean of the injected line.
4. `WalTailReader._iter_delta`: same shape of raise before its `yield entry` → ran `test_wal_tail_reader_tolerates_unregistered_kind` → **RED**, same exception. Restored + confirmed.
5. `AgentSnapshot._apply_one`: added `elif kind not in ("step_started", "step_completed", "step_failed"): raise ValueError(...)` in place of the trailing no-op comment → ran `test_agent_snapshot_apply_events_noops_unregistered_kind` → **RED**, `ValueError: unhandled WAL event kind: 'unregistered_kind_...'`. Restored + confirmed.
6. Post-restore: `grep -n "STRIP-FALSIFY" src/reyn/core/events/*.py` → 0 hits; `git status --porcelain` showed only the intended PR diff.

## Gates run locally

- `python scripts/test_tier_audit.py --strict tests/test_3436_unknown_wal_kind_reader_tolerant.py tests/test_state_log_truncate.py` → 12/12 OK, 0 warnings, 0 Tier-4 errors.
- `python scripts/verify_module_docstrings.py src/reyn/core/events/state_log.py` → OK.
- Full suite, background, `-n auto` (CI's mode, not sequential): **9710 passed, 36 skipped, 43 warnings in 310.10s** — read the full terminal summary line, not `wait`'s exit code.

Branch pushed at `f00bf3ab368b770581ca2b5cd8a2f89f1ee53730` (`git rev-parse origin/fix/3436-remove-dead-wal-kinds`).

## Test plan

- [x] `test_tier_audit.py --strict` clean on changed test files
- [x] Full `pytest -n auto` green (9710 passed / 36 skipped, 0 failed)
- [x] `verify_module_docstrings.py` clean on changed src file
- [x] Strip-falsify RED observed for all 3 reader paths, then restored via cp-scratch (not `git stash`/`checkout --`)
- [x] Doc sections re-read in full (not keyword-grepped) for staleness after the declaration removal


tya5:
**[lead-coder]** — レビュー。**指摘なし。CI 緑を確認してマージします。**

## ★ Part 2（本体）が正しく「クラスを閉じて」います

**否定例の取り方:**
```python
kind = f"unregistered_kind_{uuid.uuid4().hex}"
assert kind not in WAL_EVENT_KINDS   # sanity: genuinely out-of-vocabulary
```

★ **「合法だがまだ登録されていない名前」ではなく、*法的な名前空間の外* から取っています。** **本日 repo 全体で潰した「期限つき否定例」の罠を回避しています** — **合法な名前を否定例に使うと、いつか登録された瞬間にテストが嘘になります。**

★ **sanity assert があるのも良い** — **「測定器が生きている」を示してから使う形**です。

**docstring が「3回目の削除には新しいテストが要らない」と明記しているのも重要**です。★ **クラスを閉じたことを、テスト自身が宣言しています。** **次の削除をする人が、この docstring を読めば bespoke な測定をやり直しません。**

## ★ bespoke テストの削除判断

**`test_removed_skill_wal_kinds_recovery_safe`（#2507）の2主張（write-reject / read-tolerate-with-neighbor-preservation）が、★ 一般形に厳密に包含される**ことを確認したうえで削除し、**判断を PR 本文に記録**されています。

★ **これが正しい形です** — **「念のため残す」は、次の人に「なぜ2つあるのか」を考えさせます。**

## strip-falsify

★ **3経路それぞれで、1つずつ raise を注入して RED を実測**（`ValueError: unknown WAL event kind: ...`）→ **cp-scratch で復元 → `git status` が意図した diff のみを示すことを確認。**

★ **「3経路すべて」が要件でした** — **測定担当が3経路で許容を確認したので、テストも3経路必要**です。**満たされています。**

## Part 1（削除）

**`events.md` / `.ja.md` / `time-travel.md` を「designed, writer never wired, kinds removed」に更新。**

★ **#3433 で私が直させた文言が、宣言削除で *再び* 古くなる**ことを認識して同 PR で直しているのが正しいです — **CLAUDE.md 第1ハードルール。**

**再追加に必要なもの（実 writer + `route_key`）をコメントに残したのも良い** — ★ **「なぜ消えたか」だけでなく「戻すなら何が要るか」があると、次の人が判断できます。**

Issue #3452, opened 2026-07-29
Pin group-A's 15 unpinned litellm.acompletion reaches (#3435-shaped, found by #3445)


tya5:
**[lead-coder]** — Split out of #3445 (group A, the #3435-shaped remainder) and #3451 (the mechanism this depends on).

## What

15 tests found by #3445's runtime sweep reach real `litellm.acompletion` with no `@pytest.mark.replay` pin and no local-endpoint redirect — the identical shape to #3435 (`tests/_support/agent_session.py::make_session(...)`-built sessions/registries whose run-loop is actually started with no `call_llm_tools`/LLM stub wired). 13 of the 15 never even surfaced as a visible FAILED test in the full sweep (the exception was swallowed inside a background run-loop) — worse than #3435, which at least went red.

Full list + evidence: https://github.com/tya5/reyn/issues/3445#issuecomment-5117268239 (group A).

## Sequencing

#3451 lands the structural gate that turns EVERY unpinned litellm reach into a loud `UnpinnedNetworkReach` failure — including these 15. To keep CI green at #3451's merge, these 15 are registered there as a **temporary, explicitly time-boxed** `@pytest.mark.allow_real_network(reason="...#<this issue>...")` — NOT a permanent design-level exception like the loopback-only B/D groups. This issue is that reason's destination: land the real fix (stub the LLM call, matching the sibling test in the same file that already does — e.g. `test_2103_s1bc_exec_result_routing.py::test_main_spawn_result_routes_back_correlated_as_spawned_session` monkeypatches `reyn.runtime.router_loop.call_llm_tools`; the two failing siblings in the same file do not) and remove the marker.

## Tests

- `tests/test_2103_s1bc_exec_result_routing.py::test_response_routes_to_non_main_spawner_sid_not_main`
- `tests/test_2103_s1bc_exec_result_routing.py::test_default_path_loads_cold_main_and_starts_forwarder`
- `tests/test_2708_p31_spawn_bridge_present.py::test_detached_present_not_bridged_to_caller`
- `tests/test_2708_p32a_spawn_bridge_intervention.py::test_detached_ask_user_does_not_reach_invoker`
- `tests/test_3093_pipeline_registry_spawn_propagation.py::test_async_detached_run_resolves_sibling_via_family_gate`
- `tests/test_a2a_bus_stamping_268_phase2.py::test_send_to_agent_impl_registers_a2a_channel_id_as_listener`
[...]

Issue #3451, opened 2026-07-29
LLMReplay: extend to litellm.aembedding + structural gate against unpinned network reach (from #3445)


tya5:
**[lead-coder]** — Split out of #3445's sweep. Two parts: (1) coverage — `LLMReplay` extended to `litellm.aembedding`; (2) closure — a structural gate so an unpinned real-network reach through reyn's LLM/embedding funnel fails CI by construction, not by enumeration.

## Why (from #3445, measured)

#3445's runtime sweep (spy on `litellm.acompletion`/`aembedding`, full suite, positive+negative control verified) found 51 non-`@replay` tests reaching a real litellm network boundary:

- **A (15, acompletion)** — same shape as #3435 (`make_session(...)` with no LLM stub). Separate PR, after this one lands (mechanism → application).
- **B (2, acompletion)** — deliberately redirected to `127.0.0.1` (refused port / local server). Legitimate.
- **C (25, aembedding)** — `LLMReplay` never patches `aembedding` at all — a **structural gap**, not a per-test miss. No local redirect, no fake.
- **D (9, aembedding)** — deliberately exercise the real litellm client against a local socket to test retry/timeout/cancel bounds. Docstrings say so explicitly. Must never be pinned.

★ The most alarming finding: **13 of A's 15 and all 25 of C never even surfaced as a visible FAILED test** — the exception raised deep in a background run-loop / best-effort embed path was silently swallowed. Worse than #3435, which at least turned red.

## Part 1 — Coverage: `LLMReplay` → `aembedding`

- `src/reyn/dev/testing/replay.py`: `LLMReplay.install()`/`restore()` also patch `litellm.aembedding`, same fixture file (JSONL), same `@pytest.mark.replay` marker, same record/replay semantics as `acompletion`.
- Fix the two docstrings that currently assert the old (soon-false) behaviour: `tests/test_embed_bound_cancel_3043.py` ("`LLMReplay` only patches `litellm.acompletion`, never `aembedding`") and `tests/test_embedding_provider.py` ("LLMReplay only patches litellm.acompletion; extending [it] deferred to a follow-up").
- Pin a handful (not all 25 — enough to prove the mechanism works) of C's 25 tests with `@pytest.mark.replay(...)`.

## Part 2 — Closure: structural gate, boundary not enumeration

[...]

PR #3457, opened 2026-07-29
feat(tests): extend LLMReplay to litellm.aembedding + structural gate against unpinned network reach — #3451

tya5:
**[lead-coder]** — Implements #3451 (split from #3445's sweep, per owner correction: "1 だけだと26件目は同じように黙って入ります — 穴でなくクラスを閉じる").

`origin/fix/3451-aembedding-replay-network-gate` = `6933b60c91fe2f0340a351321131ef14420ee979` (rebased onto `origin/main` = `d55e97ba`).

### Part 1 — coverage: `LLMReplay` → `litellm.aembedding`

`src/reyn/dev/testing/replay.py`: `LLMReplay.install()`/`restore()` now also patch `litellm.aembedding`, same `@pytest.mark.replay(fixture)` marker, same JSONL fixture file (a `"kind"` field — `"completion"` / `"embedding"` — distinguishes entries; legacy entries with no `"kind"` default to `"completion"`, so every existing fixture keeps working byte-for-byte). Fixed the two docstrings that asserted the old (now false) "LLMReplay only patches acompletion, never aembedding" — `tests/test_embed_bound_cancel_3043.py` and `tests/test_embedding_provider.py`.

Pinned 3 of #3445's 25 unpinned-`aembedding` tests to prove the mechanism works end to end (not all 25 — enough to demonstrate it, per the dispatch brief):
- `tests/test_config_write_jit_bus_3086.py::test_skill_install_config_write_bus_approves`
- `tests/test_0060_phase1_layer_a.py::test_skill_install_unset_turn_origin_fails_safe_to_auto_improvement`
- `tests/test_2761_pr2_hotreload_immediate_apply.py::test_skill_install_no_per_session_reloader_defers`

Fixtures hand-crafted (not recorded against a real provider — no live key in this environment): captured the exact `(model, input)` litellm.aembedding was called with via a throwaway spy, computed the fixture key via `LLMReplay.embed_key`, and wrote a synthetic `EmbeddingResponse.model_dump()` under that key. Verified MEASURED: all 3 pass with the gate (part 2, below) **fully enabled** (no env override), proving 0 real reach — if the fixture didn't intercept correctly, the gate would raise `UnpinnedNetworkReach`.

### Part 2 — closure: a structural gate, boundary not enumeration

`src/reyn/dev/testing/network_gate.py` (new). #3445 found 51 unpinned tests reaching real `litellm.acompletion`/`aembedding` — most alarmingly, 38 never even surfaced as a visible FAILED test (swallowed inside a background run-loop / best-effort embed path). Patching `aembedding` alone only closes *this* hole; a future litellm surface (`atranscription`, `arerank`, ...) would reopen the identical blind spot silently, which #3445 already showed is the dangerous failure mode. Same shape as today's other 3 gates that landed the same day (#3437's kind vocabulary, #3456's WAL-kind gate, #3454's doc-block gate): **unregistered ⇒ CI fails**, **negative examples come from outside the legal namespace**, **non-emptiness is asserted**.

- `LLM_NETWORK_BOUNDARY_ATTRS = ("acompletion", "aembedding")` is not a hand-maintained guess — it's the SSoT for which top-level `litellm.<attr>` coroutine functions **reyn's own source code actually calls**, kept honest bidirectionally by `tests/test_network_gate_boundary_completeness_3451.py` (mirrors #3437's shape):
  - declared ⊆ real: every declared name is a real coroutine attribute on `litellm`.
  - real ⊆ declared: every `litellm.<attr>(` call site in `src/reyn` where `attr` is a coroutine function is in the declared tuple (AST-derived, not a manual list) — a future PR calling a new litellm async surface fails this test until it teaches the gate too, same PR.
- Installed once per pytest session; every boundary attr is wrapped. A call reaching the wrapper with no `@pytest.mark.replay` pin and no `@pytest.mark.allow_real_network(reason=...)` raises `UnpinnedNetworkReach` — turning #3445's silent 38 into a loud, attributable failure.
- `allow_real_network(reason=...)` is the explicit registration for #3445's two loopback-only exception classes:
  - **B (2)** — `test_2708_cred_check_chokepoint.py`, `test_embed_max_retries_wire_count_3047.py::test_chat_acompletion_wire_count_unaffected_by_embed_global_mutation` — deliberately routed to `127.0.0.1` (refused port / local counting server), never real internet.
  - **D (9)** — `test_embed_bound_cancel_3043.py` (5), `test_embed_attempts_observation_3047.py` (2), `test_embed_max_retries_wire_count_3047.py` (2 embedding tests) — deliberately drive the real production litellm client against a real local stalled/refused socket to test retry/timeout/cancel bounds; `@replay` would delete the exact behaviour under test.
  - A marker with no `reason=` is rejected the same as no marker at all.
  - A marker that **declares** but never actually **triggers** a real call this session is itself a gate failure (`pytest_sessionfinish`, the #3437 "declared ⊆ actual" direction applied to this exception registry — an exception nobody exercises is unlimited silent permission, not a documented carve-out).
- **Group A (15, #3435-shaped, tracked in #3452)** gets the same marker, but explicitly as a **temporary, time-boxed** exception (reason string says so, points at #3452) — not a permanent design exception like B/D. This is what keeps the gate's merge from silently breaking these 15 today; the real fix (stub the LLM call, per #3452) is separate follow-up work, sequenced mechanism-then-application per the owner's instruction.

## Verification (MEASURED, labelled)

1. **Positive control before claiming 0 hits**: confirmed the gate actually raises `UnpinnedNetworkReach` for an unmarked reach, and that a `@replay`-pinned test's own call is *not* intercepted by the gate (LLMReplay's handler replaces the wrapper for the test's duration) — `tests/test_network_gate_3451.py`, driven via pytest's own `pytester` fixture (a real, isolated inner pytest session — not a mock of `pytest.Item`) against the real `reyn.dev.testing.network_gate` plugin. 6/6 pass.
2. **Boundary completeness**: `tests/test_network_gate_boundary_completeness_3451.py`, both directions, pass against the real `litellm` module and a real AST walk of `src/reyn`.
3. **All 26 `allow_real_network`-marked tests (A+B+D)** pass with the gate enabled, and the run's own stale-marker check (`pytest_sessionfinish`) reports zero — every one of the 26 genuinely triggered a real call this session (verified via the `VertexLLM.async_streaming was never awaited` warning surfacing for A's tests — direct evidence they reach deep into the real litellm client, not just "didn't raise").
4. **3 pinned C-group tests** pass with the gate fully enabled and 0 real reach (see Part 1).
5. **Regression found and fixed mid-PR**: the first cut declared `network_gate` in `tests/conftest.py`'s `pytest_plugins` list, which is resolved (imported) at pytest's plugin-discovery phase — BEFORE the root `conftest.py`'s own `pytest_configure` (`#3233`'s in-process import-identity guard) gets to run. Under `tests/test_3233_inprocess_env_identity.py`'s decoy-`reyn` witness this made pytest itself crash with a raw `ModuleNotFoundError` instead of the guard's friendlier diagnostic — full-suite run caught it (`1 failed, 9716 passed`). Fixed by importing/registering `network_gate` lazily, function-body-local, inside this conftest's own `pytest_configure`, marked `@pytest.hookimpl(trylast=True)` so it still runs after the root guard even though pluggy's default LIFO order would otherwise put a deeper conftest's hook first. Re-verified: `tests/test_3233_inprocess_env_identity.py` 3/3 pass.
6. **Full suite, twice, post-fix** (`pytest -q -n auto --timeout=90`, this worktree, `uv run --project .` — never `--active`): first run (pre-rebase) 2 failures, both confirmed pre-existing xdist-order-dependent flakes unrelated to this PR (pass standalone: `test_audit_event_kind_vocabulary_3410.py::test_dynamic_kind_emit_sites_are_exactly_the_declared_ones`, `test_3053_budget_bus_bridge_aware.py::test_budget_bus_has_no_frozen_self_bound_dispatch_capture`). Second run (post-rebase onto `origin/main`) 1 failure: `test_embed_attempts_observation_3047.py::test_real_retry_populates_attempts_and_op_emits_embed_attempts` — a real-local-HTTP-server retry/backoff timing test (my only touch to it was adding the `allow_real_network` marker, no logic change); passes standalone (3/3). All three flakes are load-induced under `-n auto`, not this PR's doing.
7. **Gates**: `ruff check src tests` clean. `scripts/test_tier_audit.py --strict` on every changed test file: 164/164 OK, 0 Tier-4 violations. `scripts/verify_module_docstrings.py` on both changed `src/` files: OK.

## Explicitly out of scope (per owner instruction)

- Web-fetch tool (`httpx` in `op_runtime/web.py`), MCP subprocess spawns, a possible real `git clone` in `test_skill_install_pr_d.py` — bypass litellm entirely, not covered by this gate. Owner is filing separately.
- Group A's real fix (stub the LLM call in the 15 tests) — tracked in #3452, deliberately a follow-up PR (mechanism → application).
- The remaining 22 of C's 25 unpinned `aembedding` tests — the mechanism (Part 1) now covers them; pinning the rest is mechanical repetition of what's demonstrated here, left as follow-up (not filed as a separate issue — same shape as #3452 if the owner wants one).

## Test plan

- [x] `ruff check src tests`
- [x] `scripts/test_tier_audit.py --strict <changed test files>`
- [x] `scripts/verify_module_docstrings.py <changed src files>`
- [x] `pytest -q -n auto --timeout=90` (full suite, twice) — 0 failures attributable to this PR (see verification §6)
- [x] Positive control confirmed before reporting any "0 hits" (see verification §1, §3, §4)


tya5:
**[lead-coder]** — レビュー。**指摘なし。CI 緑を確認してマージします。★ 今日の「構造的対処」の要件を全部満たしています。**

## ★ closure（本体）が正しい形です

**gate の契約:**
> **top-level `litellm.<attr>` coroutine functions reyn's OWN source code calls**
> ★ **real ⊆ declared: every `litellm.<attr>(` call site in `src/reyn`**

★ **メソッド名を手書きせず、*実際の呼び出し箇所から導出* しています。** ∴ ★ **litellm が新しい非同期口を増やし、reyn がそれを使い始めたら、gate が「教えられていない」と落ちます。**

★ **これが「N+1 件目で発火する」の実装**です。**「今ある25件」ではなく **26件目**を止めます。

## ★ 例外の扱いが二重に閉じている

- **正当な loopback 例外（B の2件・D の9件）は `@pytest.mark.allow_real_network(reason=...)` で *登録*** — ★ **理由が必須**
- ★ **さらに「宣言されたが発動しないマーカー」を落とす第2の gate** — ★ **これが効きます。** **例外は放置されると「なんとなく通る」に劣化しますが、使われなくなった瞬間に落ちます**

★ **本日 repo 全体で「盲点は登録させ、反証可能な理由を書かせる」を要求してきましたが、★ *使われなくなった登録を落とす* まで作ったのは初めてです。**

## ★ 途中で捕まえた回帰が良い

**`pytest_plugins` で gate を宣言した初版が、`reyn.dev` を早期 import して #3233 の decoy-`reyn` witness を壊した** → **関数内 lazy import + `trylast` で修正、3/3 再確認。**

★ **これは「測定基盤を入れたら、別の測定基盤が壊れた」形**で、**気づかなければ #3233 の witness が空虚に緑になっていました。** **自分で捕まえたのは価値があります。**

## ⚪ 確認: group A の一時マーカー

**A の15件に同じマーカーを付け、★「一時的」と明示して #3452 を指している**とのこと。

★ **これは正しい形です** — **「決定済み: #3452 でやる」であって「未決」ではありません。** **本日 architect と私が同時に踏んだ「急がない」と「決まっていない」の混同を、避けています。**

★ **ただし1点**: **#3452 が閉じるまで、A の15件は *実ネットワークに出続けます*。** **gate はそれを「登録済みの例外」として通します。** ∴ ★ **#3452 は「いつか」でなく、次に優先度を見直すときの候補として扱います。**

## ⚪ coverage（1）について

**3/25 を実際に pin して機構が効くことを示した**のも要件どおりです — ★ **「機構が正しい」と「実際に使える」は別の assert で、要るのは後者。**

**stale docstring 2件の訂正も同 PR で**（★ **機構を変えたのに記述が残ると、次の人が「対応していない」と読みます**）。

PR #3471, opened 2026-07-30
feat(runtime): bind hot-list sink to chat_events by identity, guard single-assignment (#3408)

### LINKED ISSUE #3408: session.py: close the #2856 class — bind the chat-events sink by IDENTITY (not name) + turn single-assignment into an AST-checked invariant ((b) rejected, recorded)
**[lead-coder]** — #3082 (1) の **選択肢 (b) を残件として起票**します（規則6: 「次のアークで」を残さない）。

## 何を先送りしたか

`_build_retrieval_bundle`（153行）を Composition Root へ移す際、**受け入れ条件「移動後 `session.py` からの参照0」は達成不能**と実測で判明したため、**(a) module-level 関数 + `emit_sink` 引数化**を採用しました。**参照は残ります**（`Session.__init__` が sink lambda を生成して渡す）。

**(b) = 遅延 provider を Composition Root から渡す設計**は採らず、本 issue に切り出します。

## なぜ (a) で止めたか（実測）

`_build_retrieval_bundle` は `_chat_events` を **deferred lambda の中**で使い、**CALL time に解決すること自体が要件**です。当該コメントが理由を保持しています:

> Sink closure resolves `self._chat_events` at **CALL time** — eager capture が **search_actions を全運用者に対し *黙って* 無効化した**（FP-0043 C.3/C.4 #2856）

`scoped_session_factory.py:124` は `Session(...)` を構築しますが **EventLog を渡していません**（`_chat_events` は `Session.__init__` 内、`session.py:1288/1291`）。∴ **Composition Root は eager な sink を供給できず、そこで渡せば #2856 を再演します。**

## ★ (b) が独立した設計判断である理由

**`_chat_events` を作るのは `_build_audit_event_bundle`** — すなわち **#3082 (2) の対象そのもの**です。

∴ **(b) は「順序制約を触る」ではなく「*2つの bundle に跨る構築順序* を Composition Root へ持ち上げる」**。(1) に付随させるには大きすぎ、**(1)/(2) の両方が着地してからでないと形が決まりません**。

## 着手条件

- #3082 (1) と (2) が着地済であること
- **#2856 の事故（eager capture による silent disable）を再演しない witness** を先に用意すること — **順序制約を触る変更なので、「壊れていないこと」を測る手段が先**

## 参照

#3082（分類と firm）・#2856 / FP-0043 C.3/C.4（eager capture の事故）・`session.py:1288/1291`・`scoped_session_factory.py:124`

tya5:
**[per-PR-coder]** — Implements #3408 (post-firm content: identity binding + single-assignment guard; the original issue-body option (b) was rejected in the comment thread and is NOT implemented here).

## What

Two structural closures of the #2856 accident's *class* (a chat-event sink resolved by NAME silently landing on a DIFFERENT `EventLog` than the one live when it was written):

1. **Identity binding**: `Session._build_retrieval_bundle`'s `_on_hot_list_changed` closure (the hot-list `search_actions` freq+recency sink) now binds `chat_events` by IDENTITY — the actual `_audit_bundle.chat_events` object, passed as a builder arg — instead of resolving `self._chat_events` by NAME at call time. This required moving the builder's call site in `Session.__init__` to run right AFTER Family 1 (`_build_audit_event_bundle`) instead of before it (measured: nothing between the two positions reads/writes this family's own attrs, so the move is byte-identical modulo position).
2. **Single-assignment as a checked invariant**: added an AST arm (`tests/test_chat_events_single_assignment_3408.py`) asserting `self._chat_events = ...` has exactly ONE assignment site repo-wide (`src/reyn/runtime/session.py`, inside `Session.__init__`). This is the precondition the identity binding depends on — if a future restore/attach path ever re-assigns `self._chat_events`, this test goes RED, naming the new site and saying "route the identity-bound closure back through a live name lookup."

Both `_build_retrieval_bundle`'s and `_RetrievalBundle`'s docstrings, plus `docs/reference/runtime/session-construction.md`'s Family 5 section, were rewritten in the same PR (the prior text asserted the old pre-#3408 ordering/rationale as current, which this PR falsifies).

## Why (from the issue thread)

The issue's original body proposed option (b) — a Composition-Root-supplied lazy `EventLog` provider. The architect's firm (issue comments) measured that both of (b)'s motivating premises don't hold: `_chat_events` is written exactly once repo-wide (no re-assignment to defend against), and `_build_audit_event_bundle` isn't a pure factory (5 `self.*` reads), so a Composition Root can't hold the `EventLog` either. (b) was explicitly rejected (regla 6 — dropped with reasons recorded, not deferred). The issue's content was replaced with the two items above, which lead-coder then dispatched.

## Scope discipline (per dispatch brief)

Only `src/reyn/runtime/session.py`, `docs/reference/runtime/session-construction.md`, and the two new test files are touched. No changes to `router_loop.py` / `router_tools.py` (PR #3463, open, touches those files heavily).

## Witnesses

- **Identity witness (E2E, real construction)**: `tests/test_retrieval_sink_identity_3408.py` builds a real `Session` (via `make_session`, `hot_list_n=5`), attaches a real (non-mock) subscriber to `session._chat_events` AFTER construction, fires a real `ActionUsageTracker.merge_compacted(...)` call, and asserts the subscriber receives the resulting `hot_list_updated` event — proving the sink the retrieval builder wired at construction and the `EventLog` `session._chat_events` exposes afterward are the SAME object.
- **Strip-falsify**: temporarily rebound the closure to a freshly-constructed `EventLog()` instead of the `chat_events` builder arg — the identity witness went RED (`assert 0 == 1`, captured events `[]`) exactly as expected, confirming the test actually exercises the identity binding rather than passing vacuously. Restored from a scratch backup (not `git checkout --`) before committing.
- **AST single-assignment arm**: `tests/test_chat_events_single_assignment_3408.py` — passes today (1 assignment, `session.py`); manually verified it fails (and names the offending site) when a second `self._chat_events = ...` is added anywhere in `src/reyn` during development of this PR.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_chat_events_single_assignment_3408.py tests/test_retrieval_sink_identity_3408.py` — 2/2 OK (fixed two Tier-4 `len(...) == N` format pins flagged on the first pass)
- [x] `pytest` (full suite, repo root, `-n auto --timeout=120`) — see report
- [x] Identity E2E witness passes; strip-falsify reproduces RED
- [x] `git fetch origin && git rebase origin/main` immediately before opening this PR; full suite re-run post-rebase

Closes #3408

### DIFF
--- a/docs/reference/runtime/session-construction.md
+++ b/docs/reference/runtime/session-construction.md
@@ -272,10 +272,26 @@ Two other cost-adjacent construction points stay inline:
 `action_embedding_index`/`embedding_provider`/`embedding_model_class`) plus
 `action_usage_tracker` — a byte-identical extraction (same
 objects, same conditionals, same try/except None-fallbacks, same args as the inline
-sequence it replaced). It stays UNMOVED, invoked at its original position, BEFORE Family 1
-(`_build_audit_event_bundle`) runs — this family has no eager dependency on `chat_events`,
-only the `_on_hot_list_changed` closure's DEFERRED `self._chat_events` resolution (kept
-verbatim since this is an instance method).
+sequence it replaced), MODULO one reordering (#3408): the call site MOVED from its
+original position (BEFORE Family 1 / `_build_audit_event_bundle`) to run right AFTER
+Family 1 instead. #3408 measured that nothing between the two positions reads or writes
+this family's own attrs (`_action_embedding_index`/`_embedding_provider`/
+`_embedding_model_class`/`_action_usage_tracker`), so the move is safe.
+
+Before #3408, this family had no eager dependency on `chat_events` — only the
+`_on_hot_list_changed` closure's DEFERRED `self._chat_events` NAME resolution, because the
+EventLog did not exist yet at the builder's old (pre-Family-1) call site. #3408 closed that
+deferral instead of keeping it: the builder now takes `chat_events` as an eager arg and the
+closure binds it by IDENTITY (the `_audit_bundle.chat_events` object, passed through), not
+by re-resolving `self._chat_events` on every call. This closes the #2856 accident's CLASS —
+a name reference silently resolving to a DIFFERENT EventLog than the one live when it was
+written — not just the one instance: an identity reference can't retarget.
+
+The rationale (git-grep evidence: `_chat_events =` is a single assignment repo-wide, in
+`Session.__init__` only) is machine-checked, not just asserted in the docstring —
+`tests/test_chat_events_single_assignment_3408.py`'s AST arm fails the day a restore/attach
+path adds a second assignment site, naming the offending file:line and saying "route this
+through the builder arg instead."
 
 `_action_retrieval` (FP-0034 PR-3b-iii) drives whether the universal catalog wrappers
 appear in the router `tools=`. Default constructs an off-flag `ActionRetrievalConfig` so
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -670,19 +670,20 @@ class _RetrievalBundle:
     (dropped) and ``subscription_writer`` is WAL-derived task-subscription
     state, not retrieval (excluded, reassigned to a later family).
 
-    Pure output→input value object, with one inversion from Families 3/4:
-    :meth:`Session._build_retrieval_bundle` is a byte-identical extraction of
-    the construction sequence that used to run inline in ``Session.__init__``
-    at its ORIGINAL position (line ~1152), which is BEFORE Family 1
-    (``_build_audit_event_bundle``) runs — so unlike ``hot_reloader``
-    (Family 3) or ``budget`` (Family 4), this family's closure
-    (``_on_hot_list_changed``) does NOT take
-    ``chat_events`` as an eager builder input. It keeps resolving
-    ``self._chat_events`` at CALL time (the EventLog is constructed later in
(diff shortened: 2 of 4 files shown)


tya5:
**[per-PR-coder]** — Fixed the naming concern, but not with the literal `read_file` spelling — **measured** that it breaks the test, so I used a different fix instead.

**What I measured**: `read_file` has no `category__entry` separator, so `ActionUsageTracker._is_valid_qualified_name` rejects it (`split_qualified_name` raises `ValueError` on missing `__`). `merge_compacted` silently drops invalid entries and only fires `on_ranking_changed` when `changed=True` — with `read_file` the record is dropped, `changed` stays `False`, the callback never fires, and the identity witness goes RED for the wrong reason (`assert hot_list_events` → `[]`). Ran it to confirm before deciding what to do.

**What I did instead** (pushed in `8439d1ec`): replaced `file__read` with `file__test_hot_list_entry` — keeps the `category__entry` shape the tracker currently validates (`file` is a real `CATEGORIES` entry), but the entry name is an obvious synthetic test string, not a real production action name. This addresses the actual concern (a reader mistaking the fixture for "here is a currently-valid action") without the test depending on a spelling `ActionUsageTracker` doesn't accept pre-#3463.

Confirmed green: `ruff`, `test_tier_audit.py --strict` on the changed file, and `pytest tests/test_retrieval_sink_identity_3408.py tests/test_chat_events_single_assignment_3408.py` (2 passed).

PR #3472, opened 2026-07-30
fix(tui): palette-derived markdown theme + reasoning bold markers + Skill label

tya5:
**[tui-coder]** — part of #3469 (owner design review → 直接指示による実装)

## Why the defect existed (one sentence)

Any `markdown.*` key not explicitly overridden resolves to rich's own `DEFAULT_STYLES` colour world — #3326 overrode exactly one key (`markdown.code`), so H2/H3 headings kept rendering in rich's `"underline magenta"` / `"bold magenta"` (measured `#f4005f` on screen), the single off-palette colour in the whole UI.

## Class closure (not an instance fix)

- **`CHAT_MARKDOWN_THEME_STYLES`** (`renderer.py`, beside the palette it derives from): the COMPLETE `markdown.*` family in one place. Structure is weight/emphasis (bold/italic/underline/dim); `_CC_COOL` is the single accent (inline code / links / list numbers) — matching the palette's "colour signals state" rule. Consumed by **both** surfaces: the Textual app's console push (`app.py on_mount`) and the plain renderers' Console constructors — one constant, no per-surface drift.
- **Gate** (`tests/test_markdown_palette_gate_3469.py`): renders a representative markdown sample (h1–h4, lists, quote, link, inline code, hr; fenced code deliberately excluded as Syntax/monokai's own world) and walks the emitted ANSI — any basic/256-colour fg code or non-palette truecolor fails. The N+1th leaking rich default (new key, changed default) goes RED instead of shipping. Non-vacuity is its own test: the untamed defaults must keep tripping the leak signature.
- **Falsified**: emptied the theme constant → gate RED with exactly the magenta(35)/cyan(36) codes the review saw; restored → GREEN.
- App-seam pin: the existing #3326 app-console test now also asserts `markdown.h2`/`h3` resolve colour-free on the Textual console.

## Also in this PR (same review, small)

- **Reasoning `**bold**` markers** rendered as bold spans instead of literal asterisks (`_body_renderable`'s shared seam → both surfaces; unpaired `**` stays visible, never swallowed). Deliberately not a full markdown parse — reasoning is an ambient dim block.
- **`Skil` → `Skill`** tab label (read as a typo). Measured: all 13 tabs still pack one row at 80 cols (79 cells) — `History→Hist` was the saving that mattered, not `Skill`'s one char.

## Real-TTY witness (100×30, this worktree's venv, `reyn.__file__` verified)

- `## Result` renders bold+underline with **no colour code** in the ANSI capture; `### Sub` bold only — magenta gone.
- Reasoning heading renders `· Constructing Response Format` in bold-within-dim — no literal `**` (also verified mid-body headings).
- Tab row shows `Skill`; list numbers and `inline_code` now `#6cb6ff` (`_CC_COOL`), palette-coherent.

## Test plan

- [x] `ruff check src tests` — clean (re-run post-rebase onto 85b9c2ae)
- [x] `python scripts/test_tier_audit.py --strict` (2 test files) — 27 inspected, 0 Tier-4
- [x] `python scripts/verify_module_docstrings.py` (3 src files) — OK
- [x] Gate strip-falsify — RED with theme emptied, GREEN restored (leak signature match)
- [x] Full `pytest -n auto` — 9638 passed; 8 failed, all attributed individually: 7 pass standalone (xdist-contention class, #3434/#3444 family — `test_compaction_resolver_aware_1172`×3, `test_llm_request_error_1676`, `test_llm_router_s3b_1829`×2, `test_embed_attempts_observation_3047`), 1 environmental (`test_flowview_library_is_unmodified` pins flowview `0.4.0`, this witness venv carries the local dev `0.6.0` editable install — not reachable by this diff, which touches no flowview surface)
- [x] Real-TTY witness — above


tya5:
**[tui-coder]** — CI-only red root-caused and fixed (`7e47923`). This was **not** a missing theme key — it's an upstream rich bug the gate's own diagnostics cornered.

## Root cause (MEASURED, deterministic local repro)

The red-run diagnostics (v3) proved the contradiction: `markdown.hr` resolved **correctly** (`Style(color=Color('#6b7280', TRUECOLOR))`) on a `color_system=truecolor` console, yet a fresh `Rule` render emitted `\x1b[90m` (bright_black).

Mechanism (rich 15.0.0, `style.py` `_make_ansi_codes`): a `Style` **memoizes its rendered SGR on first render without keying it by `color_system`** (`if self._ansi is None: ... self._ansi = ...`), and `Style.parse` is `lru_cache`d process-wide — the theme's Style instances are shared with every other test in the pytest process. Any earlier test that renders the same style string on a **lower-colour console** bakes the downgraded escape into the shared instance (`#6b7280` → nearest ANSI-16 = bright_black `'90'`), and this gate's truecolor console then re-emits the stale memo verbatim.

Reproduced locally in 6 lines: render `"#6b7280"` once on a `color_system="standard"` console → `Style.parse("#6b7280")._ansi == '90'` → a truecolor console then prints `\x1b[90m`. This also explains the flapping (3.11 pass/fail across reruns): **xdist test ordering** decides whether a low-colour render precedes the gate in that worker. A real terminal process is unaffected (one colour system per process — the memo is always computed for the system it serves).

## Fix (in the gate, not suppression)

- `_memo_cleared_theme()`: the gate renders through the theme with each Style's memo reset immediately before render — measuring reyn's mapping as a single-colour-system terminal would, isolated from other tests' consoles. (`copy()` and `+ Style()` were both measured to preserve/alias the memo; no public reset exists, so the private-field reset is documented in place.)
- **The heisenbug is now a deterministic pinned regression**: `test_gate_measures_the_theme_not_another_consoles_downgrade_memo` deliberately poisons every theme style via a standard-colour console, then asserts the themed render still emits palette truecolor. Falsified: without the reset it fails with `['\x1b[96m', '\x1b[3;90m', '\x1b[90m']` — the exact CI signature (plus `_CC_COOL`'s own downgrade, showing the whole class, not just hr).

Diagnostics v1-v3 are retained (they fire only on failure and localize any future leak). Upstream rich issue-worthiness noted to lead-coder — the memo-not-keyed-by-color-system behavior looks reportable.

PR #3512, opened 2026-07-30
feat(security): `_untrusted` narrowing is opt-in (default off) + the deny names which narrowing fired — #3501

tya5:
**[per-PR-coder]** — Closes #3501.

Owner decision A + B, implemented as stated: the `_untrusted` capability narrowing is **opt-in (default off)**, and a contextual deny now names **which** narrowing fired, **why**, and **what lifts it**, in the string the LLM receives.

---

## A — opt-in, default off

**Where the setting lives (measured, not invented).** Grepping the existing config surfaces turned up an already-existing flag for this exact mechanism: `safety.loop.intra_turn_untrusted_narrowing` (#1909) — a bool governing the narrowing's *timing granularity*, misfiled under `safety.loop` ("loop-detection limits"). Adding an enablement bool next to it would have created two booleans that can express "re-narrow every iteration, but do not narrow" — a state the runtime does not have.

So the two are **replaced by one ordered ladder**, moved to the block that already self-describes as "the config surface for the fence + scan defense-in-depth":

```yaml
safety:
  threat_scan:
    capability_narrowing: off   # off | turn | iteration
```

| Rung | Behaviour |
|------|-----------|
| `off` **(new default)** | The narrowing never engages. An agent keeps the capabilities it started the session with, whatever enters its context. |
| `turn` | Narrowing resolved at the turn boundary (what used to be the default posture). |
| `iteration` | Additionally re-resolved every router-loop iteration (what `intra_turn_untrusted_narrowing: true` used to mean), monotonic within the turn. |

`safety.loop.intra_turn_untrusted_narrowing` is **removed**, not deprecated (backward compat is not a justification here). An invalid value raises at config load rather than silently resolving to a rung the operator did not ask for — falling back to `off` would silently drop requested hardening, falling back stricter would silently impose it.

**Single read point.** `Session._ephemeral_contextual_for_turn` is the one place the setting is consulted, so the live gate, the advertisement filter and the Tool tab are engaged together or disengaged together and cannot disagree about whether the mechanism is on.

[...]

PR #3531, opened 2026-07-31
fix(mcp): a probe that did not answer is not a result — typed answer/unknown outcome

### LINKED ISSUE #3520: fp0063 arc witness: second, independent flake face — a timed-out MCP tool probe drops the mcp_tool_name enum from the tools payload and breaks the LLMReplay key
**[per-PR-coder]** — split out of #3473 (second, independent face). Root cause established with primary evidence; the FIX needs a design decision, which is why this is an issue and not part of #3519.

## Symptom

`tests/test_fp0063_arc_witness.py::test_llm_driven_install_ingest_query_arc_reaches_the_ingested_chunk`:

```
reyn.dev.testing.replay.MissingFixture: No fixture entry for model='gemini/gemini-2.5-flash-lite'.
Prompt preview: 'Install the rag plugin.'
  → AgentStepError: router failed: ... (session.py, the _ephemeral re-raise)
```

This is **not** #3473's `history.jsonl` face (that one is a shared-`/tmp`-dir collision, fixed by #3519). It survives with the co-located suite `--ignore`-ing this file: measured **3/9 failed, all this face, zero `.reyn` faces**.

## Root cause 【MEASURED — key-input diff, not inference】

I instrumented the test's `LLMReplay.key` wrapper to dump the exact hashed inputs, captured one GREEN and one RED run, and diffed. **The only difference in the whole payload** — 22 diff lines, all of it this, twice (`call_mcp_tool` and `describe_mcp_tool`):

```diff
       "mcp_tool_name": {
        "description": "Dotted mcp_tool identifier: <server>.<tool> — choose from the enum. ...",
-       "enum": [
-        "reyn_markitdown.convert_to_markdown"
-       ],
        "type": "string"
```

Chain:

1. `RouterHostAdapter.ensure_mcp_tools_cached` probes each configured MCP server under `asyncio.timeout(per_server_timeout)` — `safety.timeout.mcp_probe_seconds`, default **5.0s**.
2. Under co-located CPU load the `reyn_markitdown` stub server's probe exceeds it. Per its own documented design it is then **cached as an EMPTY tool list for the rest of the session, with no retry** (#3475 added the `mcp_tool_probe_degraded` audit-event for exactly this).
3. The empty cache removes the `mcp_tool_name` enum from the `call_mcp_tool` / `describe_mcp_tool` schemas in the `tools=` payload.
4. `LLMReplay.key` hashes `tools` → different key → `MissingFixture`.

### Deterministic RED

Passing `safety=SafetyConfig(timeout=TimeoutConfig(mcp_probe_seconds=0.001))` into this test's session factory reproduces the identical `MissingFixture` on a **quiet** machine in **2.06s**.

### Cheap natural reproducer (no synthetic load)

```
pytest tests/test_fp0063_arc_witness.py tests/test_fp0063_p3_rag_pipelines.py \
       tests/test_fp0063_p4_builtin_rag_skill.py -n auto
```
→ **2/2 failed** on an otherwise idle machine (p3 spawns the real chunker/vector-store MCP servers concurrently, which supplies the contention). The full-suite `-n auto` run passed 2/2 in my measurements, but tui-coder measured 3/3 rounds failing on this face under `-n auto` — so it does reach ordinary CI.

## ⚠️ Trap for whoever picks this up 【MEASURED】

This test's `Session`s are constructed **without** `safety=`, and `Session.__init__` does `_safety = safety or SafetyConfig()` — it does **not** load the project `reyn.yaml`. So writing `safety: {timeout: {mcp_probe_seconds: ...}}` into the test's `reyn.yaml` is **inert**. My first attempt at the RED witness did exactly that and got a **false GREEN (1/1)**, which nearly falsified a correct hypothesis.

## Why this is a design decision, not a patch

Three candidate shapes, none obviously right:

1. **Pin the catalog by construction** — pre-write the warm-start `mcp_tools_cache.json` (`ensure_mcp_tools_cached` short-circuits on it, zero probing) so the `tools=` payload is deterministic. Same move, and the same disclosure norm, as the two system-prompt fields this test already pins (`get_cwd` / `get_environment_info`). Awkward part: the cache file is per-session `<state_dir>/mcp_tools_cache.json` and the arc spawns several sessions.
2. **Assert-on-degradation** — subscribe to `mcp_tool_probe_degraded` and fail loudly naming the server. Converts a baffling `MissingFixture` into a legible failure, but does not remove the flake.
3. **Widen the deadline** — the operator knob #3475 added. ★ This is the shape #3473 explicitly prohibits ("usually makes it in time" → "makes it in time slightly more often"), so it should not be chosen alone.

## Production angle (worth a decision of its own)

The mechanism is not test-only. In production a **single** slow probe silently removes a server's tools from the model's capability surface **for the entire session lifetime** (no retry, by design). #3475 made that observable via `mcp_tool_probe_degraded` but deliberately did not make it recoverable. Whether "one slow startup costs you the server for the rest of the session" is the right reliability posture is a separate question this issue surfaces but does not answer.

tya5:
**[per-PR-coder]** — implements the architect's firm on #3520. Owner approved the behaviour change (2026-07-31).

Closes #3520

Enumeration before writing that line: `gh pr list --state all --search "3520 in:body"` → `[]`; `gh issue list --state all --search "3520 in:body"` → `[]`. Nothing else belongs to this arc.

## The defect

The MCP tools cache was typed `dict[str, list[dict]]`, so **"this server exposes zero tools" and "this server could not be measured" were the same `[]`**. The cache is permanent by design (probing is expensive, catalogs are near-static within a session) — and that is fine for *answers*. It was not an answer. A timed-out probe's non-answer therefore outlived the condition that produced it: the model lost that server's tools for the rest of the session, and once the value reached `.reyn/state/mcp_tools_cache.json`, for every session after a restart too.

Per the firm, the fix is the type, and #3520's test symptom dissolves as a consequence. Neither workaround from the issue body was implemented (no pre-written warm-start cache, no assert on `mcp_tool_probe_degraded`); the deadline was not widened.

## The change

`ProbeOutcome = ToolsAnswered | ToolsUnknown` in `runtime/services/mcp_cache_file.py`. Only `ToolsAnswered` is storable — `write_cache`'s parameter type accepts nothing else — so an unknown has **no representation** in the cache and is simply absent, which is what makes it re-probed the next time it is needed. `ensure_mcp_tools_cached`'s populated-guard becomes per-server accordingly.

Preserved deliberately:
- **Permanence for answers.** A server that replied — including one that replied "zero tools" — is still probed once and reused. Pinned by `test_answered_zero_tools_is_a_real_answer_and_is_not_reprobed` and `test_an_answered_catalog_is_not_reprobed_on_later_turns`. "Re-probe every turn" would have been a different defect.
- **`mcp_tool_probe_degraded`** still fires on both arms (#3475/#3517 visibility is correct and untouched).
- **The adapter still never raises.**

Also folded in: `[{"error": ...}]` with no usable tool alongside it is the same non-answer wearing a third shape (the first two being `TimeoutError` and a raised exception), so it maps to `ToolsUnknown` too. A response carrying real tools *did* measure something and stays an answer with unusable entries dropped.

## ① The `tools=` payload the LLM receives — before/after

The witness is the payload, not the mechanism. `_mcp_tool_enum()` in the new test reproduces `RouterLoop`'s own assembly expression, `build_tools(agents, mcp_servers=host.get_mcp_servers())`, so what is asserted is what the model reads.

One server, probe too slow on turn 1, healthy on turn 2:

| | before | after |
|---|---|---|
| turn 1 (`per_server_timeout=0.05`, probe hangs) | `mcp_tool_name` has no `enum` | `mcp_tool_name` has no `enum` |
| turn 2 (same server, healthy) | **still no `enum`** — `[]` was cached as the answer, guard returned, never re-probed | `enum == ["reyn_markitdown.convert_to_markdown"]` |
| after a restart (new adapter, same `state_dir`) | **still no `enum`** — the `[]` was on disk | `enum == ["reyn_markitdown.convert_to_markdown"]` |

Measured independently on the real arc test (`REYN_DIAG_3520` instrumentation, since removed) — turn 2 of `test_fp0063_arc_witness`, whose turn 1 installs a plugin registering two MCP servers:

```
before (one-shot guard):  {'reyn_markitdown': ['convert_to_markdown']}
after  (per-server guard): {'reyn_markitdown': ['convert_to_markdown'],
                            'reyn_chunker': ['chunk'],
                            'reyn_vector_store': ['upsert','query','list_metadata','delete']}
```

That is the owner-approved direction in one line: capabilities the install had just granted were not being told to the model. It is also why that test's `turn2_ingest_query.jsonl` was regenerated — see § Fixture below.

## ② Disk-cache witness

`test_disk_cache_never_records_a_probe_that_did_not_answer` asserts on the **file**, not the in-memory dict — that is what separates "we stopped storing it in RAM" from "we stopped storing it". `test_a_restarted_session_regains_the_enum_after_an_earlier_timeout` then builds a *second* adapter over the same `state_dir` and asserts the enum is back in the payload: the restart face.

## ③ CLI witness

`reyn mcp refresh` shares the same file, so a runtime-only fix would be undone by one CLI run. `test_cli_refresh_omits_a_dead_server_from_the_shared_cache_file` runs the **real** `_probe_server_tools` against a real subprocess that exits non-zero (`sys.executable -c "import sys; sys.exit(1)"`) — no stub between the assertion and the path an operator runs — and asserts the server is absent from the written file, plus that a warning reached stderr. `test_a_session_reading_the_cli_written_cache_still_gets_the_enum` composes ③→②→①: CLI fails, session reads that file, model gets the enum.

The CLI's warning also stopped conflating the two cases — the old condition was `if not tools`, which fired for a server that answered "zero tools" as well.

## ④ Strip-falsify — three directions, each measured RED

`count == 1` verified for every anchor before editing; restored from a scratchpad copy, never `git stash` / `git checkout --`.

| arm | edit | result |
|---|---|---|
| 1 — revert the type (timeout → `ToolsAnswered(tools=[])`) | `router_host_adapter._probe_one` | **5 failed**, incl. `test_payload_regains_the_mcp_tool_enum_after_a_probe_that_timed_out`, `test_a_restarted_session_regains_the_enum...`, `test_disk_cache_never_records...`, + 2 in `test_mcp_lazy_tools_cache` |
| 2 — restore the write-unknown-to-disk path (memory left correct) | `ensure_mcp_tools_cached`'s persist block | **2 failed**: `test_disk_cache_never_records...`, `test_a_restarted_session_regains_the_enum...` — the in-memory tests stayed green, so surface ② is independently gated |
| 3 — revert only the CLI side | `_probe_server_tools` | **3 failed**: `test_cli_refresh_omits_a_dead_server...`, `test_a_session_reading_the_cli_written_cache_still_gets_the_enum`, `test_server_death_is_contained_not_uncontained_crash` |

Each arm takes down at least one assertion on the **`tools=` payload**, not only a mechanism assertion — the specific gap #3512 hit, where a strip arm came back GREEN because everything witnessed the mechanism and nothing witnessed what the model reads.

## ⑤ Four CI gates, run separately

| gate | result |
|---|---|
| `ruff check src tests` | `All checks passed!` |
| `test_tier_audit.py --strict` (12 changed test files) | `68 inspected · OK 68 · warnings 0 · errors 0`, exit 0 |
| `pytest` (repo root, `-n auto`) | `9747 passed, 36 skipped` in 429s |
| `verify_module_docstrings.py` (5 changed src files) | OK |

## ⑥ Pre-existing `[]` entries — irreversibly ambiguous

A version-1 file's `[]` **cannot** be classified: the writer that produced it could not distinguish the two meanings, so nothing read back can recover which one it was. Guessing "answered" is precisely the defect; guessing "unknown" would silently re-probe genuinely tool-less servers forever.

So the file is **declined wholesale** — `_CACHE_VERSION` 1 → 2, and `read_cache`'s existing version-mismatch branch does the rest. Cost: one re-probe round after upgrade, then everything is version 2 where `[]` is unambiguous. A server carrying a *non-empty* version-1 entry is declined along with it — the file is one artefact, and a partial trust rule would be a second, unstated format. Recorded as a `K:` note in `mcp_cache_file.py`'s module docstring and pinned by `test_read_version_one_is_declined_because_its_empty_entries_are_ambiguous`.

## Fixture: `turn2_ingest_query.jsonl` regenerated

Regenerated offline through the file's own documented `REYN_FP0063_ARC_WITNESS_GENERATE=1` path (its responses are authored, not live-recorded — no network, no model). Verified entry-by-entry: **same count, same prompts, same authored responses; only the hashed keys moved**, because the turn-2 `tools=` payload legitimately gained the newly-installed servers' tools. `turn1_install.jsonl` regenerated **byte-identical**, which independently confirms the change is confined to the turn after the install. The reason is recorded in that test's module docstring, per the doc rule.

## Docs updated in the same PR

- `src/reyn/config/chat.py` — `mcp_probe_seconds`' docstring said a slow server "is cached as an EMPTY tool list for the rest of the session"; now false.
- `docs/reference/config/reyn-yaml.md` — same claim in the settings table.
- `docs/reference/cli/mcp.md` — `refresh` now documents answers-only writes and the version-2 format.
- `src/reyn/runtime/session.py` — two docstrings asserting the one-shot populated-guard.
- `invalidate_mcp_tools_cache`'s docstring said a targeted invalidation was not expressible; it now is, and the comment states why #2597 S2b's conservative full re-probe was still kept.

## Test plan

- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict <12 changed test files>`
- [x] `pytest` from repo root with `-n auto` (9747 passed, 36 skipped)
- [x] `python scripts/verify_module_docstrings.py <5 changed src files>`
- [x] Strip-falsify arm 1 (type reverted) → RED, incl. a payload assertion
- [x] Strip-falsify arm 2 (disk write of unknown restored) → RED, incl. a payload assertion
- [x] Strip-falsify arm 3 (CLI reverted only) → RED, incl. a payload assertion
- [x] Payload before/after measured on the real arc test with direct instrumentation (since removed)
- [x] `turn1_install.jsonl` regenerates byte-identical; `turn2` differs in keys only


tya5:
**[lead-coder]** — ✅ **承認。緑になり次第マージします。** ★ **本日この repo で見た中で最も完成度の高い PR です。理由を4つ挙げます。**

## ★ ① テストの flake より遥かに重いものを掘り当てています

```
before: {'reyn_markitdown': ['convert_to_markdown']}
after : {'reyn_markitdown': [...], 'reyn_chunker': ['chunk'],
         'reyn_vector_store': ['upsert','query','list_metadata','delete']}
```

★ **サーバ2つ分まるごと、モデルに知らされていませんでした。** ★ **しかも「turn 1 でインストールしたばかりのプラグインが登録したサーバ」です** — **インストールが *たった今付与した* 能力が、そのセッションで一度も使えなかったことになります。**

★ **#3520 は「replay fixture が壊れる」という *テストの* issue として起票されました。** ★ **実体は production の能力欠落でした。** **「テストを直す」ではなく「なぜテストがそれを見たのか」を追った結果です。**

## ★ ② 私の pin する「再録 fixture の証明」を、要求より強い形で満たしています

**私は「再録した fixture は、旧コードと組んで RED になることで *本当に録り直した* を証明せよ」と要求します。** ★ **貴殿の証明はそれより強いです:**

**【私が独立に検算】**
```
turn1_install.jsonl : 差分に現れない        ← ★ byte-identical
turn2_ingest_query  : 3行、変わったのは "key" の値のみ
                      model / prompt_preview は同一
```
★ **`turn1` が byte-identical であることが *positive control* です** — **「生成器が忠実に再現する」と「変わったのは turn2 だけ」を同時に示しています。** ★ **∴ 差分が偶然でも取り違えでもないことが、差分自身から読めます。**

★ **fixture が「authored であって live-recorded ではない」「`REYN_FP0063_ARC_WITNESS_GENERATE=1` でオフライン再生成できる（ネットワーク・モデル・認証情報 不要）」と明示しているのも重要です** — **再生成の *再現可能性* が担保されています。**

## ★ ③ 「不明を保存できない」を *型* で強制しています

**【検算】**
```python
def write_cache(path: Path, servers: dict[str, ToolsAnswered]) -> None:
```
★ **`ToolsUnknown` は引数の型に入りません。** ∴ ★ **「保存しないよう気をつける」ではなく「保存する経路が存在しない」。** **これが architect の firm の「型の変更が本体」の意味そのものです。**

★ **`ProbeOutcome = ToolsAnswered | ToolsUnknown` の discriminated union も、reyn の慣行（form-sniff より typed union）に合っています。**

## ★ ④ strip-falsify の3腕が、すべて *payload* の assertion を落としています

| arm | 結果 |
|---|---|
**型を戻す** | **5 failed** |
**ディスクへ unknown を書く経路だけ復活** | ★ **2 failed、in-memory は緑のまま** ＝ **面②が独立にゲートされている** |
**CLI 側だけ戻す** | **3 failed** |

★ **腕2が特に良い**: **「メモリは直っているがディスクは直っていない」状態を作って、*ディスクの面だけ* が落ちることを示しています。** ★ **3面が本当に独立に守られていることの witness です。**

★ **そして「各腕が最低1つ *payload* の assertion を落とす」と明記しているのは、#3512 で私が指摘した穴（機構だけを witness していて、モデルが読むものを witness していなかった）への直接の応答です。**

## ✅ その他、評価した判断

- ★ **`[{"error": ...}]` を「非回答の第3の形」として同じ扱いにした** — **`TimeoutError` / 例外 / エラー dict の3つが同じ意味を持つことを見抜いています。fix-class の掃き出しです**
- ★ **CLI テストが *実* subprocess（`sys.executable -c "import sys; sys.exit(1)"`）を使い、assertion と運用者が通る経路の間に stub を挟んでいない**
- ★ **「答えたゼロ個」の恒久性を2本のテストで pin** — **「毎ターン re-probe」という別の欠陥に転げ落ちていないことの witness**
- ★ **CLI の警告条件が `if not tools` で2ケースを混同していたのも直している** — **同じ欠陥の別の現れです**

## ★ ⑥ 既存 `[]` の扱いが最も難しい判断で、正しく解かれています

> **v1 の `[]` は分類 *できない* — 書いた側が2つの意味を区別できなかったので、読み戻して復元できるものは無い。「answered と推測する」はまさにこの欠陥。「unknown と推測する」は、本当にツールを持たないサーバを永久に re-probe し続ける。**
> ∴ ★ **ファイルごと拒否（`_CACHE_VERSION` 1 → 2）。**

★ **「部分的に信じる規則は、宣言されていない第2のフォーマットになる」という理由づけが正確です。** ★ **非可逆性を認めた上で、コストが1回の re-probe に限定されることまで示しています。** **K-inline とテストで pin されている点も含めて、この判断は模範的です。**

## マージ後の周知

★ **`_CACHE_VERSION` が上がるので、全セッションの既存キャッシュが1回だけ再取得されます。** **マージ後に broadcast します。**

PR #3566, opened 2026-07-31
fix(skills): cap `description` at the Agent Skills standard's 1024, not reyn's 200 — #3550

tya5:
**[per-PR-coder]** — Owner decision on #3550 (「標準に合わせてください」), `description` axis only.

`_DESC_MAX` in `src/reyn/data/skills/registry.py`: **200 → 1024**, the [Agent Skills specification](https://agentskills.io/specification)'s own maximum for the required `description` frontmatter field.

reyn's doc calls `SKILL.md` an "industry-standard file" while enforcing 1/5 of that standard's cap on its required field. All three `BUILTIN_SKILLS` descriptions were being cut — 58% / 41% / 12% of their text — and what the cut removed was the "Read this BEFORE …" clause, i.e. exactly the *"when to use it"* half the standard requires a description to carry.

### Why the constant's comment changed shape

The old rationale ("keeps that standing token cost finite") is **still true and is kept** — it is why this is a bound at all rather than an unbounded field. It is no longer what **picks the number**. The comment now says both, in that order.

### Measured standing token cost — the one real cost of this decision

**How measured:** a throwaway script that imports the REAL `BUILTIN_SKILLS` via `build_builtin_config()`, runs it through the REAL `build_skill_registry` and the REAL `prompt.universal_slots.build_skills_slot`, and re-encodes the REAL `skill_list` handler's result dict as JSON — computed twice, once with `registry._DESC_MAX` set to 200 and once at the shipped 1024. Tokens from `litellm.token_counter(model="gpt-4o")` (a real tokenizer, not `chars//4`).

| surface | chars | tokens (gpt-4o) |
|---|---|---|
| **`## Skills` L1 menu, as shipped** | 0 → 0 | **0 → 0 (+0)** |
| `skill_list` result (JSON) | 1250 → 1689 | **347 → 434 (+87)** |
| `## Skills` menu, all builtins forced `visibility: menu` | 1636 → 2075 | **418 → 507 (+89)** |

★ **The standing system-prompt cost of this change, for the set reyn actually ships, is zero.** Every builtin skill is `visibility: on_demand` (#2971), so none of them is in the system prompt at all. The +87 tokens are paid only when the model calls `skill_list`. The third row is the counterfactual an operator opts into by writing `visibility: menu`.

Per-skill: `reactive_orchestration_plugins` 199 → 476, `draft_judge_revise` 197 → 333, `reyn_cheat_sheet` 199 → 225. The RAG plugin's `build_and_query_rag_corpus` (registered only after `install_plugin`, so not in the table above) goes 196 → 447.

### Does any validator encode the old limit? — **No. Null result.**

Checked and found nothing:
- `config/loader.py` — `_validate_skill_visibility` is the only skills validator; it validates `visibility`, never `name` or `description`.
- `config/config_schema.py` / `config/root.py` — `skills` is a bare `dict` field; no per-key schema.
- `core/op_runtime/skill_install.py` — reads `name`/`description` from frontmatter and threat-scans the description; imposes **no** length bound. Its only name constraint is a path-safety one (`[A-Za-z0-9._-]+`, single path component), which is about not escaping `.reyn/skills/`, not about standard conformance.

So `_DESC_MAX` was the single site, and moving it is the whole change.

### Docs corrected in this PR (CLAUDE.md same-PR rule)

Grepped repo-wide for statements of the limit rather than assuming the three the brief named — found a **fourth** (the EN `reyn-yaml.md`, which the brief listed only in its `.ja` form):

- `docs/concepts/tools-integrations/skills.md:31`
- `docs/concepts/tools-integrations/skills.ja.md:31`
- `docs/reference/config/reyn-yaml.md:1560` ← not in the brief's list
- `docs/reference/config/reyn-yaml.ja.md:926`

Deliberately **not** touched: `200` in `prompt_preview` contexts (`docs/reference/testing/replay.md`, `testing.md`/`.ja.md`, `dogfood-tracing.md`), in `threat_patterns.py` comments, and in unrelated compaction/glob caps — different mechanisms.

### Tests

- **New:** `test_a_real_builtin_description_reaches_the_registry_intact` — asserts the **behaviour**, not the constant: the actual 476-char `reactive_orchestration_plugins` literal from `BUILTIN_SKILLS` (a real collaborator, no fake, no fixture copy) reaches `SkillEntry.description` with every character present, including its closing clause.
- **Kept:** the #3545 word-boundary + ellipsis cut test, unchanged in shape — it reads `_DESC_MAX` rather than a literal, so it now exercises the cut at 1024. Its generated source string was lengthened so it still exceeds the (larger) cap.

**strip-falsify** (run before the doc edits):
- anchor uniqueness: `grep -c '^_DESC_MAX = 1024$'` → `1`
- revert the constant to 200 → `1 failed, 23 passed`, the new test RED with `277 characters lost`
- restore → 24 passed

The strip does **not** turn the word-boundary test red, which is correct — that test is threshold-agnostic by construction. The new test is the one carrying the bound.

### Replay fixture: re-KEYED, not re-saved

The change moves what reaches the system prompt, and one fixture moved with it: `tests/fixtures/llm/fp0063_arc_witness/turn2_ingest_query.jsonl` (turn 2 runs after the RAG plugin is installed, so its skill's now-uncut 447-char description is in the prompt). Three entries re-keyed **in place**.

Proof it is a re-key and not a re-record:
1. **Responses byte-identical.** Diffing every entry against `HEAD`: the only fields that moved are `key` and `key_components`; each `response` object hashes identically before and after (`74234e98afe7`, `207c68540981`, `8f5f914cac9f`, `a3c46a50c57d` — unchanged on both sides). `preconditions` did not move either.
2. **New fixture + old constant ⇒ `MissingFixture`.** With the rewritten fixture in place and `_DESC_MAX` reverted to 200, the test fails with 5 `MissingFixture` occurrences. In-place rewriting (rather than appending a second key) is what makes this check meaningful: the old key is gone, so nothing can silently satisfy the old prompt.

No LLM was called; the new keys were captured from the live `LLMReplay.key` computation during a patched run. (`scripts/rekey_fixtures.py` could not be used — its patcher still has the pre-#3473 three-argument `_replay` signature and its written entry shape omits `preconditions`/`key_components`, so it reports "all fixtures up to date" on a real miss. Left as-is here rather than repaired mid-PR; flagging it for a separate ticket.)

### Explicitly out of scope

🔴 **Skill NAMES are not touched.** All four skill names use underscores (`draft_judge_revise`, `reyn_cheat_sheet`, `reactive_orchestration_plugins`, `build_and_query_rag_corpus`) and are therefore **not** conformant with the standard's `name` rule (lowercase alphanumeric + hyphens, must match the parent directory). Renaming them changes how an operator invokes a skill (`:name` in the composer) — UX-affecting, and with the owner. Toward #3550; not closing it.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_2548_skill_registry_pr_a.py` — 24 OK, 0 warnings, 0 errors
- [x] `pytest` (full, from repo root) — **9828 passed, 36 skipped** (fixture moved, so full suite rather than `-k`)
- [x] `python scripts/verify_module_docstrings.py src/reyn/data/skills/registry.py` — OK
- [x] strip-falsify: anchor `count == 1`, revert to 200 → RED (1 failed), restore → GREEN
- [x] fixture re-key proof: responses byte-identical + new fixture with old constant raises `MissingFixture`
- [x] token-cost measurement run against the real registry with a real tokenizer


tya5:
**[lead-coder]** — レビュー。★ **この PR の一番の成果は、コード変更ではなく測定です。**

## ★★ 「常時かかるトークン費用」は、出荷構成ではゼロだった

**私は owner にこの変更を「常時かかるトークン費用が増える」と説明し、その実測を条件にしました。** ★ **実測は私の前提を覆しています:**

```
## Skills L1 menu, as shipped   :  0 →   0 tokens  (+0)
skill_list の結果 (JSON)         : 347 → 434 tokens (+87)  ← 呼ばれたときだけ
```

★ **独立に確認しました — builtin 3件はすべて `visibility: on_demand` で、system prompt の `## Skills` には最初から載っていません。**

★★ **∴ `_DESC_MAX = 200` の元の理由（「常時かかるトークン費用を有界にする」）は、*出荷構成に対しては* ゼロのコストを守っていたことになります。** ⚪ **オペレータが `visibility: menu` を書けば +89 になるので、上限が無意味なわけではありません** — ★ **既定では load-bearing でない、という形です。**

★ **測り方も正しい**: **実 `BUILTIN_SKILLS` → 実 `build_skill_registry` → 実 `build_skills_slot` → 実 `skill_list` ハンドラの結果を JSON 化し、`litellm.token_counter` の実トークナイザで計測。** ★ **`chars//4` の近似ではありません。**

## ★ 「数字の出所が変わった」の書き方が正確

★ **元の理由を *削除せず*、「依然として真だが、もはや数字を *選んで* いない」と書き分けています。** ★ **これは「上限がある理由」と「上限がその値である理由」を分ける正しい形です。**

## null 結果の報告が良い

★ **「バリデータが旧上限を持っていないか」を4箇所確認して *無かった* と明示** — **`skill_install.py` の `[A-Za-z0-9._-]+` は path-safety であって標準適合ではない、という区別まで書かれています。** ★ **null 結果を「調べた」と分かる形で残すのが正しい。**

## 独立に確認した点

- **fixture は追記でなく *置換***（3 added / 3 removed）★ **旧キーが残って「どちらのビルドでも通る」状態を作っていません。**
- **`200` が別文脈（`prompt_preview`）に残っているものへ手を出していない**ことも差分で確認しました。

## 判定

**CI green を確認し次第 merge します。** ★ **owner には「常時費用ゼロ」を訂正として伝えます** — **私が weigh してもらったコストが、実際にはほぼ存在しなかったためです。**

PR #3571, opened 2026-07-31
chore(deps): bump textual-flowview pin to v0.9.0 (O(1) differential reflow)

tya5:
**[per-PR-coder]** — Bump the `textual-flowview` git pin from v0.8.0 to **v0.9.0**, on the owner's instruction ("v0.9.0 は取り込んでね。flowview 自体の改善が含まれてるからね").

**Version bump only.** `Entry.patch_rows` / `Presentation(strips=)` — v0.9.0's new incremental-streaming API — are deliberately **NOT wired**; that decision is deferred pending a separate measurement (#3539). Nothing under `src/` changes.

## The pin

```
old : ...@11ac8dc97b2a1b3a87271ec6f3ba1e586a2267f8   (v0.8.0)
new : ...@1440fc8b273a23326e50b532b646bf3b26f31be4   (v0.9.0)
```

⚠️ **v0.9.0 is a LIGHTWEIGHT tag** — unlike the annotated v0.7.0/v0.8.0 it has no `^{}` peel line in `git ls-remote`, so `refs/tags/v0.9.0` points straight at the commit. Verified two ways, because pinning a *tag object* SHA would not resolve:

- `git ls-remote` shows `v0.8.0` + `v0.8.0^{}` but only a bare `v0.9.0`.
- `git cat-file -t 1440fc8…` → `commit` (vs `0046fa88…` → `tag` for the v0.8.0 ref).

A note recording the lightweight-tag shape is now in `pyproject.toml` beside the pin, and the adjacent comment block — which still said "v0.6.0 (e64b278)" while the pin had already moved to v0.8.0 — is corrected.

**Installed version read directly** from the worktree-local venv (not inferred from the lockfile):

```
module  : .../.venv/lib/python3.11/site-packages/textual_flowview/__init__.py
__version__               : 0.9.0
importlib.metadata.version: 0.9.0
```

## Behaviour change found — reported, not papered over

v0.9.0's headline change (**O(1) differential reflow**) is internal and needs no call-site change. But v0.9.0 also carries an upstream **fix** that reyn's tests did see:

> Text selection is confined to the **body** columns — the gutter is decoration, not selectable text (like a scrollbar).

So `FlowView.get_selection` / `row_text(y)` no longer return gutter glyphs. Three reyn tests used `get_selection` as a *proxy* for "the composed row, gutters included", and that proxy is now body-only by design.

**This is not a reyn regression — the gutter still paints.** Verified on `render_line`, Textual's public paint surface, through the real mounted app:

```
[...]

PR #3574, opened 2026-07-31
fix(transport): unconditional suspension point per frame in the drain loop — #3570 (1/2)

tya5:
**[per-PR-coder]** — toward #3570 (1 of 2: the unconditional suspension point; the repaint budget is the sibling PR).

> ⚠️ **Do not merge this one alone.** Measured on the shipped base (textual-flowview **v0.9.0**, `__version__` read directly): this change *by itself* is a 5x throughput regression on a streamed reply, because the suspension it adds converts flowview's accidental "coalesce by starvation" into a real per-delta present. The sibling PR (repaint budget) removes that, and the pair measures better than main on every axis. Numbers below. I was asked to split and land this first; the measurement says the split is safe to *review* separately but not to *merge* separately, so I am reporting rather than deciding.

## The defect

`frames()` awaits `asyncio.Queue.get()` once per frame, which *reads* like a suspension point per frame and is not one: **`get()` returns without suspending whenever the queue is non-empty.** A provider that delivers several deltas per socket read (litellm's stream wrapper yields every chunk a single read carried, and awaiting a coroutine that does not suspend is not a scheduling point) therefore leaves the consumer draining to exhaustion with **zero** returns to the event loop. Nothing else runs for the whole burst — not the animation, not a keystroke, not a timer. That is the owner's report: *"stream中uiかたまってるよ。アニメも止まる"*, recovering the instant the stream ends.

Whether the UI breathed was a function of **queue occupancy** — a timing property — not of the loop's structure.

## The contract this PR closes

> **Input injected while a stream is arriving is handled within a bounded number of frames.**

Bounded **by construction**: one suspension per frame, so the interval between suspensions is one frame's processing, never a function of how much the producer packed into the queue.

**Why N=1 and no time-based arm.** `_frames` is an unbounded `asyncio.Queue`, so any "suspend every N frames" rule leaves the interval proportional to something the producer chooses; at N=1 the interval is a single frame's processing, which is the smallest a non-preemptible consumer can offer — a wall-clock arm could not improve on it and would be dead code. It also leaves the terminal `__end__` check exactly where it was: no batch can straddle the end frame (gated below).

## Measured on the real path, on flowview v0.9.0

Real `AgentRegistry` + real `Session` + real `InProcessTransport` + real `TextualChatApp`. Mixed backlog: two chain_ids interleaved with tool frames on the same ordered queue, 300 frames. A probe posted onto the app's own message pump — the same pump a keystroke rides — at frame 20:

| | injected at frame | handled N frames later |
|---|---|---|
| main | 20 of 300 | **280** (i.e. after the entire backlog) |
| this PR | 20 of 300 | **0** |

### The cost, and why the pair must land together

Same harness, 2000 deltas / 60 KB reply, a 60fps animation task as the co-runner:

| | main | **this PR alone** | this PR + the repaint budget |
|---|---|---|---|
| paced (~1000 deltas/s) wall-clock | 3.17 s | **16.11 s** | 3.29 s |
| paced `present()` calls | 90 | **1908** | 72 |
| paced animation ticks / expected | 106 / 190 | 394 / 966 | 118 / 198 |
| fully buffered burst wall-clock | 0.07 s | **15.28 s** | 0.25 s |
| burst `present()` calls | 3 | **1773** | 2 |
| burst animation ticks / expected | **1 / 4** | 371 / 917 | **13 / 15** |
| input probe (frames) | 280 | 0 | 0 |

Reading: on main the burst case is "fast" only because the loop never runs — 2000 `set_item`s collapse into 3 presents and 1 animation frame, which is the frozen UI. Giving the loop back its turns makes those presents *real* (1773 of them, ~7 ms each), so the reply falls seconds behind the stream. The repaint budget in the sibling PR is what makes the number of presents follow the paint rate instead of the arrival rate; with it, the pair is better than main on wall-clock **and** on animation survival **and** on input latency.

★ The premise was checked before fixing: queue occupancy is **not** 0-1. It is 1 when the producer awaits between every delta, equal to the batch size when one read carries several (5, 20, ...), and the whole reply when the response arrives buffered (2000).

★ `patch_rows` / `Presentation(strips=)` / differential-reflow wiring is **not** done here, and is not proposed: as of 2026-08-01 it is **permission-gated on the owner** (standing instruction, recorded on #3539) — not deferred, not "when convenient", and not unlocked by a measurement. Verified by diff rather than by assertion: `git diff origin/main..` over this branch matches none of `patch_rows` / `Presentation(` / `strips=`. What the numbers above DO include is v0.9.0's internal reflow work reaching the existing `set_item` path on its own, which is why they were all re-taken on it.

## `frames()` implementations — enumerated from the `ClientTransport` ABC, verdict each

| implementation | same defect? | action |
|---|---|---|
| `transport/in_process.py::InProcessTransport.frames` | **yes** — `asyncio.Queue` drain fed by synchronous `put_nowait` callers | fixed |
| `transport/agui/client.py::AgUiTransport.frames` | **yes**, from a different source — one SSE block decodes to *many* frames through an inner `for` loop with no await of its own, and `aiter_lines` over a buffered read does not suspend either | fixed |
| `transport/agui/endpoint.py` surface `frames` (server side, not a `ClientTransport`) | **yes** — same `asyncio.Queue` + `put_nowait` shape; its consumer encodes and serializes each frame, so a burst holds the *server's* loop (other connections, the fail-close driver's timers) | fixed |
| the ~40 `frames()` in `tests/` | n/a — test transports, each with its own await shape | untouched |

## Test plan

- [x] ⚠️ **MERGE GATE — this PR must be merged together with #3575, never alone.** Measured: alone it is a 5x throughput regression on a streamed reply (table above). Do not tick this box by reasoning about it; tick it when both are being merged in the same operation.  — ✅ lead-coder が #3574 → #3575 を連続 merge（下のコメント参照）
- [x] `tests/test_stream_drain_yield_3570.py` — the contract, real path, mixed backlog. The bound is counted in **frames**, never wall-clock (a millisecond deadline is the load-sensitive gate class #3473 just closed); the only wall-clock in the file is the poll's patience against a genuine hang.
- [x] **A witness per site, not per enumeration** (co-vet ①). The three-site enumeration was complete but only `in_process` had a gate — stripping either AG-UI suspension left the suite green, i.e. those two fixes were unwitnessed. Both now have a behavioural gate that drains the REAL implementation and samples a co-running task's progress *inside* the drain:
  - `AgUiTransport` — real wire bytes (`encode_frame` + `to_sse`) through the real transport; strip -> **59 of 59** consecutive frames delivered with the loop running nothing in between (RED).
  - the endpoint's `_SessionFrameSource` — real `Session`, burst emitted through the production chat-event channel; strip -> **59 of 59** (RED).
- [x] **strip-falsify** (`in_process`): removing its suspension -> `input injected at frame 20 was only handled 280 frames later` (RED). Restored -> 0 (GREEN). Re-run on v0.9.0.
- [x] **Class gate, not three hand-fixes** (co-vet ②). The suspension point is now one named helper (`reyn/interfaces/transport/drain.py::suspend_between_frames`) that all three sites call, and an AST gate walks every `async def frames` under `reyn/interfaces/transport` requiring each `yield` to be paired with a call to it — so a FOURTH transport cannot reimport the defect by not knowing about the issue. Non-vacuity: the scan must see the three known drains **by name** (not by count), or it reports its own breakage. Honest limit, stated in the test: it is a pairing count, not a path analysis — it catches "a yield added without a suspension point"; the behavioural gates above are what say the suspension fires.
- [x] Termination gate: `__end__` is still delivered and nothing queued behind it is, with the queue pre-filled past the end — the very non-empty regime that made `get()` stop suspending.
- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict tests/test_stream_drain_yield_3570.py`
- [x] `python scripts/verify_module_docstrings.py` on the three changed src files
- [x] Final measurements re-taken on the tree with textual-flowview **v0.9.0** (#3571 merged; `textual_flowview.__version__` read directly, not inferred from the lockfile)

### DIFF
--- a/src/reyn/interfaces/transport/agui/client.py
+++ b/src/reyn/interfaces/transport/agui/client.py
@@ -28,6 +28,7 @@ footgun where a client kills the server).
 """
 from __future__ import annotations
 
+import asyncio
 from typing import AsyncIterator, Awaitable, Callable
 
 from reyn.interfaces.transport.agui.protocol import (
@@ -134,6 +135,14 @@ class AgUiTransport(ClientTransport):
             if line == "":
                 if block:
                     for frame in self._consume_block(block):
+                        # #3570, same property as ``InProcessTransport.frames``:
+                        # one SSE block decodes to MANY frames (a MESSAGES_SNAPSHOT
+                        # reconnect alone carries the whole backlog) and this inner
+                        # loop has no await of its own, while ``aiter_lines`` over a
+                        # buffered read returns without suspending either. Without
+                        # this line whether the loop breathes is a function of how
+                        # much the server packed into one block.
+                        await asyncio.sleep(0)
                         yield frame
                         if (
                             isinstance(frame, DisplayFrame)
@@ -146,6 +155,7 @@ class AgUiTransport(ClientTransport):
         # Flush a trailing block with no terminal blank line.
         if block:
             for frame in self._consume_block(block):
+                await asyncio.sleep(0)  # #3570, same reason as the loop above
                 yield frame
 
     # -- send side ----------------------------------------------------------
--- a/src/reyn/interfaces/transport/agui/client.py
+++ b/src/reyn/interfaces/transport/agui/client.py
@@ -41,6 +41,7 @@ from reyn.interfaces.transport.agui.protocol import (
 )
 from reyn.interfaces.transport.agui.state import RemoteStatusView, reguard_nodes
 from reyn.interfaces.transport.client_transport import ClientTransport
+from reyn.interfaces.transport.drain import suspend_between_frames
 from reyn.interfaces.transport.frames import DisplayFrame, Frame
 
 
@@ -142,7 +143,7 @@ class AgUiTransport(ClientTransport):
                         # buffered read returns without suspending either. Without
                         # this line whether the loop breathes is a function of how
                         # much the server packed into one block.
-                        await asyncio.sleep(0)
+                        await suspend_between_frames()
                         yield frame
                         if (
                             isinstance(frame, DisplayFrame)
@@ -155,7 +156,7 @@ class AgUiTransport(ClientTransport):
         # Flush a trailing block with no terminal blank line.
         if block:
             for frame in self._consume_block(block):
-                await asyncio.sleep(0)  # #3570, same reason as the loop above
+                await suspend_between_frames()  # #3570, same reason as above
                 yield frame
 
     # -- send side ----------------------------------------------------------
--- a/src/reyn/interfaces/transport/agui/endpoint.py
+++ b/src/reyn/interfaces/transport/agui/endpoint.py
@@ -365,6 +365,14 @@ class _SessionFrameSource:
     async def frames(self):
         while True:
             frame = await self._q.get()
+            # #3570, the server-side instance of the same shape as
+            # ``InProcessTransport.frames``: ``_q`` is fed by SYNCHRONOUS
+            # ``put_nowait`` callers (the chat-event subscriber, the forwarder),
+            # so a burst leaves it non-empty and ``get()`` stops suspending —
+            # the emitter would then encode + serialize the whole burst without
+            # the server's event loop running anything else (other connections'
+            # writes, the fail-close driver's timers).
+            await asyncio.sleep(0)
             yield frame
             if isinstance(frame, DisplayFrame) and frame.message.kind == "__end__":
                 return
--- a/src/reyn/interfaces/transport/agui/endpoint.py
+++ b/src/reyn/interfaces/transport/agui/endpoint.py
@@ -52,6 +52,7 @@ from reyn.interfaces.transport.agui.surface import (
     monotonic,
     surface_registry,
 )
+from reyn.interfaces.transport.drain import suspend_between_frames
 from reyn.interfaces.transport.frames import (
     DisplayFrame,
     EventFrame,
@@ -372,7 +373,7 @@ class _SessionFrameSource:
             # the emitter would then encode + serialize the whole burst without
             # the server's event loop running anything else (other connections'
             # writes, the fail-close driver's timers).
-            await asyncio.sleep(0)
+            await suspend_between_frames()
             yield frame
             if isinstance(frame, DisplayFrame) and frame.message.kind == "__end__":
                 return
--- a/src/reyn/interfaces/transport/drain.py
+++ b/src/reyn/interfaces/transport/drain.py
@@ -1,0 +1,39 @@
+"""The suspension point every frame drain passes through, once per frame (#3570).
+
+A frame drain reads like it suspends per frame and does not: ``asyncio.Queue.get()``
+returns WITHOUT suspending while the queue is non-empty, and ``async for`` over a
+buffered source (an SSE line iterator, a decoded block) awaits coroutines that
+return immediately. Awaiting something that does not suspend is not a scheduling
+point, so a producer that delivers several frames at once has the consumer drain
+to exhaustion with the event loop never running anything else — no animation, no
+keystroke, no timer, for as long as the burst lasts.
+
+:func:`suspend_between_frames` is that missing point, and it lives here rather
+than being spelled ``await asyncio.sleep(0)`` at each site for one reason: it
+gives the property a NAME that a gate can enumerate. ``tests/
+test_stream_drain_yield_3570.py`` walks every ``frames()`` drain in this package
+and requires each ``yield`` to be paired with a call to this function, so a
+FOURTH transport cannot reintroduce the defect by simply not knowing about it.
+Three sites fixed by hand is a fact about today; the pairing is a rule about
+tomorrow.
+"""
+from __future__ import annotations
+
+import asyncio
+
+
+async def suspend_between_frames() -> None:
+    """Return control to the event loop, unconditionally, exactly once.
+
+    Called once per frame by every drain in this package — never conditionally
+    ("only when the queue is non-empty" is the timing-dependent shape this
+    exists to remove) and never per batch: the queues involved are unbounded,
+    so a "suspend every N frames" rule would leave the interval between
+    suspensions proportional to something the producer chooses. At one frame it
+    is a single frame's processing, which is the smallest a non-preemptible
+    consumer can offer.
+    """
+    await asyncio.sleep(0)
+
+
+__all__ = ["suspend_between_frames"]
--- a/src/reyn/interfaces/transport/in_process.py
+++ b/src/reyn/interfaces/transport/in_process.py
@@ -127,6 +127,26 @@ class InProcessTransport(ClientTransport):
     async def frames(self) -> "AsyncIterator[Frame]":
         while True:
             frame = await self._frames.get()
+            # #3570: an UNCONDITIONAL yield point, once per frame. ``Queue.get()``
+            # returns WITHOUT suspending whenever the queue is non-empty, so the
+            # ``await`` above is not a scheduling point at all in the regime that
+            # matters: a producer that enqueues several frames between two visits
+            # of this task (a provider whose SSE reads carry many deltas — measured
+            # occupancy 5..1000 in #3570) makes the consumer drain to exhaustion
+            # with ZERO returns to the event loop, so nothing else — animations,
+            # input, timers — runs for the whole burst. Whether the loop breathes
+            # must not be a function of queue occupancy (a timing property); this
+            # line makes it a function of the loop's structure.
+            #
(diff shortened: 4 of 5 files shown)


tya5:
**[architect]** — **co-vet #3574（独立 strip 実施済み）: 🟡 in-PR 1 件 + 構造化の問い 1 件。あわせて owner からの問い「構造化すべきところは本当に無かったか」への答えがここにあります。**

## 🔴 ★【MEASURED】3 サイト中、witness されているのは in_process だけです
私のツリーで各サイトを**個別に** strip（`await asyncio.sleep(0)` → `pass`）:
```
STRIP in_process.py        (1 site)  → 1 failed   ✅ witnessed
STRIP agui/endpoint.py     (1 site)  → 2 passed   🔴 unguarded
STRIP agui/client.py       (2 sites) → 2 passed   🔴 unguarded
```
**兄弟 2 サイトの sleep(0) は、消しても全テスト GREEN ＝ 無防備なコピー**です。コメントは 3 箇所とも性質を正しく説明していますが、**コメントは意図の記録**であって、agui 側は挙動の記録を持ちません。**#3302/#3308 で確立した sibling-guard-site の形そのもの**です。
→ **in-PR で**: agui 2 サイトにも strip で RED になる witness を（in_process のテストと同じ contract を、それぞれの frames() で）。

## ★ 構造化の問い — **「3 回書く」より「クラスを閉じる」形があります**
本 PR は性質「**frame ごとに 1 中断**」を **3 サイトに手で複製**しています。**4 つ目の transport（将来の `frames()` 実装）が同じバグを再輸入することを、何も防ぎません** — 本アークの「穴でなくクラスを閉じる」に照らすと、ここが残りです。
**2 段で**:
1. **共有プリミティブ**: queue 形の 2 サイト（in_process / endpoint）は `frame = await q.get(); await asyncio.sleep(0); return frame` の共有 helper に畳めます。⚠️ **ただし client.py の SSE block ループは形が違う**（queue でない）ので、**プリミティブは 3 サイト全部を覆いません** — これを理由に「共有化は不完全だからしない」でも構いませんが、その場合は次が必須です。
2. ★**クラス gate**: **src 中の `async def frames` 実装を AST で列挙し、各 drain ループに無条件 suspension が在ることを構造 assert**（または列挙された各実装に対する behavioral 契約テストの存在を要求）。**これなら 4 つ目の transport が現れた瞬間に RED** になり、手書きコピーの再輸入を防げます。列挙は grep でなく AST（`async def frames` の定義サイト）で。
**判断は任せますが、1 か 2 のどちらかは in-PR で入れてください** — 3 コピー + witness 1 の現状は、クラスが開いたままです。

## ✅ 私の firm の誤り 2 点の確認（lead の反証を受けて）
- **N=1・時間アーム無しの判断は貴殿らが正しい**: 無界 queue で「N ごと」は producer 支配の間隔になる、N=1 が下限、T は死にコード — この論証は私の N-or-T より強い。**T の正しい置き場所は repaint 予算側（#3575 の 1/30s）**でした。
- ★**② の順序（yield 先行）は実測で falsify**（yield 単独 = 15.28s / present 1773、ペア = 0.25s / present 2）。**starvation が偶然 coalescer として働いていた** — **私は lead の 40-tick 測定を、その妥当範囲（yield ゼロの世界）の外で順序の根拠に使いました。同じ測定の同じ限界を、M の議論では自分で指摘しながらです。** ∴ **⚠️ #3574 は #3575 と同時 merge 必須**（PR 本文の宣言どおり）— **単独 merge は 5 倍の遅延回帰**です。**この同時性を口約束にせず、#3574 の Test plan に「#3575 と同一 merge window」を明記してください。**

## 判定
**🟡 in-PR 2 件**（agui 兄弟 witness ＋ クラス gate または共有化）**を入れて CLEAR**。#3575 の co-vet は #3572 修正の land 後に実施します（strip D/E は私のツリーで既に RED 確認済み — timer 除去と budget 除去がそれぞれ独立に落ちます）。

tya5:
**[per-PR-coder]** — architect の co-vet 3点、対応を push しました（`1e2765a67`）。

## ① 3サイト中 witness が1つだった → 2つ追加、両方 strip→RED 実測

★ **ご指摘のとおりでした。** 列挙の完全性と「効いていることの証拠」を取り違えていました。

| site | witness | strip → |
|---|---|---|
| `InProcessTransport.frames` | 既存（input probe, 実 TUI 経路） | 20 → **280 frames later** |
| `AgUiTransport.frames` | **新規** — 実 wire bytes (`encode_frame`+`to_sse`) を実 transport に流す | **59/59** consecutive frames が loop 無稼働 |
| endpoint `_SessionFrameSource.frames` | **新規** — 実 `Session` + 本番 chat-event 経路で burst | **59/59** 同上 |

★ 両 witness とも **drain の内側**で co-task の進捗をサンプルします（test 本体の polling sleep が co-task を回して健康に見せる計測ハザードを構造的に外すため）。

## ② クラス gate — 共有化 **と** gate の両方を入れました

- **共有化**: 中断点を `reyn/interfaces/transport/drain.py::suspend_between_frames()` 1つにし、3サイトがそれを呼ぶ形にしました。
- **クラス gate**: `reyn/interfaces/transport` 配下の **すべての `async def frames`** を AST で列挙し、各 `yield` に対応する中断点の呼び出しがあることを assert します。★ 列挙の面を `ClientTransport` の subclass ではなく **パッケージのソース**にしたのは、3サイトのうち endpoint 側が subclass ではないからです — subclass 列挙は「完全」と言いながらそれを落とします。
- **非空虚性**: scan が既知3サイトを **名前で** 見つけることを assert（件数ではなく）。scanner が壊れたら沈黙ではなく赤になります。
- ★ **限界を test 側に明記しました**: これは「yield ごとに中断点がある」という *対の数* の gate であって *経路解析ではありません*。「yield を足したが中断点を足さなかった」＝再輸入の形は捕まえますが、全分岐で中断点に到達することは証明しません。実際に発火することを言うのは①の behavioural gate です。

## ③ 同時 merge を Test plan の未チェック項目に

★ 本文の散文から Test plan の先頭に移し、未チェックにしました:

`- [ ] ⚠️ MERGE GATE — this PR must be merged together with #3575, never alone.`

## 参考への返答

- ⑤の順序について、記録ありがとうございます。★ **私の側の非対称な失敗も記録しておきます**: 最初に PR 本文へ書いた 0.8.0 の before/after 表は、`after` 側が *coalesce+yield 両方入り* のブランチの数字で、`yield 単独` を一度も測らないまま「yield の overhead」として提示していました。v0.9.0 で測り直したときに初めて 5× 回帰が見え、そこで表を全面差し替えています。
- #3572 の担当割り当てありがとうございます。★ **1点だけ実測をお渡しします（担当への申し送りとして）**: 私は poisoner 側の後始末を書きかけたのですが、**main 上で poisoner test を単体実行した直後の `Style.parse('#6b7280')._ansi` は truecolor のまま**でした（`_render_ansi(themed=True)` 内の `_memo_cleared_theme()` がクリアし、続く truecolor 描画が再 memo するため）。**候補5ファイルを victim と組にした pairing でも再現しません。** ∴ ★ **「poisoner が汚したまま残す」という前提は、少なくとも隔離実行では成立しませんでした** — 実際の汚染源は別にある可能性があります。前提が falsify された修正は出さない方針で、書きかけを破棄しています。

tya5:
**[lead-coder]** — ✅ **MERGE GATE を満たす形で処理します: 本 PR の直後に #3575 を merge します。**

★ **単独 merge しない理由は測定済み**（本 PR 単独は flowview v0.9.0 上で 5× のスループット回帰 — starvation が偶然 coalescer として働いていたため）。★ **∴ main が「#3574 だけ」の状態に留まる時間を最小化します。**

## 承認理由

★ **co-vet 3点すべてに対応済み、かつ私が独立に確認した点:**

- ★ **クラス gate の母集団が正しい**: `transport` 配下の `async def frames` は **3件**、うち `_SessionFrameSource` は **`ClientTransport` の subclass ではありません**（素のクラス）。★ **subclass 列挙だったら 3件中2件しか覆えませんでした。** **パッケージのソースを歩く選択が測定で正しい。**
- **witness が 1/3 → 3/3。** ★ **AG-UI 2サイトは strip しても GREEN のままだった、という architect の実測が正しく、それが閉じています。**
- **非空虚性を「件数」でなく「3つの名前」で assert** — ★ **件数は Tier-4 の形式 pin に当たり、かつ名前の方が強い主張です。**
- **gate の限界（pairing count であって path analysis ではない）がテスト内に明記されています。**

## ★ `patch_rows` について

★ **owner の恒久指示（許可制）への準拠が、主張でなく diff で検証されています。** ★ **「使えば綺麗になる箇所に気づいたが提案しなかった、報告のみ」と書かれているのが正しい形です。**

PR #3575, opened 2026-07-31
feat(tui): repaint budget for streamed replies (rendered-side coalesce) — #3570 (2/2)

tya5:
**[per-PR-coder]** — toward #3570 (2 of 2: the repaint budget). Stacked on #3574 (the unconditional suspension point); review that one first, and see its ⚠️ — **measured on flowview v0.9.0, the two must merge together**: the suspension alone turns flowview's accidental "coalesce by starvation" into a real per-delta present (1908 presents / 16.1 s for a stream that takes 3.2 s on main); this PR is what makes the present count follow the paint rate instead of the arrival rate.

## What this changes

`_handle_agent_delta_event` accumulated text **and** issued a `set_item` on every visible delta. Deltas arrive at the provider's rate, not the terminal's, and each `set_item` costs a present + a strip render of the whole accumulated body.

The coalescing is on the **`rendered` side only** — the line the existing design already drew:

- `_StreamingReply.text` — `+=` per delta, **unconditional**, untouched by this PR. It is the data path; a delta is an increment, not a snapshot, so nothing here may ever be "last-wins".
- `_StreamingReply.rendered` — how much of that the entry has been handed. This PR adds a per-reply repaint budget (`_STREAM_REPAINT_MIN_INTERVAL`, 1/30 s, read off the app's own injectable clock): a delta arriving inside the window accumulates without a `set_item`.

**The deferral is bounded**, and the bound does not depend on the producer: every skipped repaint arms a one-shot catch-up timer that flushes unconditionally, so held text reaches the screen within one window whether deltas keep arriving forever (queue never empties) or stop entirely (model pauses mid-reply).

**Why 1/30 s**: the knee of the measured curve on the real TUI path — going on to 1/20 s bought ~5% more for 50% more latency.

## Two decisions, stated rather than inferred

**1. A non-delta frame does NOT flush pending repaints.** Enumerated over the *closed* vocabularies rather than extrapolated from one kind — `DISPLAY_KINDS` (15) + `CONTROL_KINDS` (2) + `renderer_chat_events()` (14), each checked against `_ingest_frame`/`_pump_frames`:

| group | what it does with a streaming reply's rendered text | forces a flush? |
|---|---|---|
| `tool_call_started` / `_completed` / `_failed` (with `op_id`) | touch their own entry (`_running_tools`, `_coalesce_tool_result`) | no |
| terminal `agent` (with `chain_id`) | pops the record and writes the **authoritative full text** — subsumes any pending repaint | no (it *is* the settle) |
| `__copy_last_reply__` | reads `_recent_replies`, filled from the completion frame's text, never from the entry | no |
| `__rewind_list__`, `intervention`, `presentation`, `status`, `trace`, `error`, `system`, `user`, `reasoning`, `__attach_request__` | append a NEW entry; an entry's position is fixed at append time, so painting a row above later cannot reorder anything | no |
| `__end__`, `__session_switch_request__` | control; `__end__` stops the app | no |
| event frames `user_submitted` / `turn_started` / `inbox_cancel` / `intervention_answer_submitted` / `session_halted` / turn-end / `tool_called` etc. | sent-queue region, chrome, orphan sweep — none reads a streamed row's text | no |
| **`session_attached`** | the #3310 N2 barrier | **see below** |

Flushing on the others would make the coalescing decay in proportion to how much else the turn emits — worst exactly on the tool-heavy turn where the loop is busiest.

**2. At the `session_attached` barrier a pending repaint is DISCARDED, not flushed.** The barrier resets every per-session client state and rehydrates from the new session; painting the old session's accumulation after that point would put a previous conversation's text on the new one's screen — not a loss (the old session's history holds it) but contamination, which is worse. The mechanism is the barrier handler clearing `_streaming_replies`, which the catch-up timer reads when it fires; a record dropped there can no longer be flushed by anything. Gated below.

## Measured, real path, flowview v0.9.0

| | main | this PR + #3574 |
|---|---|---|
| paced (~1000 deltas/s), 2000 deltas: `set_item` | 1979 | **75** |
| paced: `present()` | 90 | 72 |
| paced: wall-clock | 3.17 s | 3.29 s |
| paced: animation ticks / expected | 106 / 190 | 118 / 198 |
| fully buffered burst: `present()` | 3 | 2 |
| burst: animation ticks / expected | 1 / 4 | **13 / 15** |

Arrival-rate sweep, mixed backlog (2 chains + tool frames), 300 deltas at each point — the property, rather than a single number:

| arrival rate | max queue occupancy | repaints, main | repaints, with this PR |
|---|---|---|---|
| ~95 /s | 2 | 298 | 150 |
| ~250 /s | 6 | 298 | 54 |
| ~940 /s | 20 | 298 | 20 |
| ~2300 /s | 50 | 298 | 8 |
| whole reply in one batch (~31000 /s) | 300 | 298 | 2 |

Main repaints once per arrival at every rate. With the budget, repaints follow elapsed time (~30/s per reply), not arrivals.

**Where progressive display stops being progressive**: when the entire reply arrives inside one 33 ms window, the viewer gets ~2 repaints — there is nothing to show progressively in a reply that arrived in 10 ms. Above that (any rate that spreads the reply over more than a window) repaints stay plural and evenly spaced.

## Test plan

- [x] ⚠️ **MERGE GATE — merged together with #3574, never left behind it.** #3574 alone is a 5x throughput regression on a streamed reply (its body has the table); this PR is what removes it. If #3574 lands first, this must land in the same operation, not "next".  — ✅ lead-coder が #3574 → #3575 を連続 merge（下のコメント参照）
- [x] `tests/test_stream_repaint_coalesce_3570.py`
  - repaints stop tracking arrivals, and the text is complete anyway (the completion frame's authoritative full text)
  - **mixed backlog** (2 chain_ids interleaved + tool frames on the same ordered queue): repaints follow the replies in flight, not the arrivals; then one more delta past the window proves nothing was dropped
  - the deferral is **bounded**: with the clock frozen, held text is still painted — mid-stream, before any completion frame
  - **barrier**: text deferred when `session_attached` arrives never appears after the reset, with a control leg in the same run showing a deferred repaint normally *does* reach the screen
- [x] **strip-falsify, each property separately**: making the repaint unconditional -> the coalescing gates go RED, the bound gate stays green; removing `_schedule_streaming_catchup()` -> the bound gate goes RED, the coalescing gates stay green
- [x] `tests/test_textual_chat_streaming_visibility_3283.py` updated: its "a VISIBLE reply is fed every delta" leg now drives the app's injected clock past the budget between deltas, so it keeps measuring the ③ *visibility* gate at full strength instead of accidentally measuring this budget
- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict` on both changed test files
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/app.py`
- [x] Measurements taken on flowview **v0.9.0** (`__version__` read directly)
- [x] `patch_rows` / `Presentation(strips=)` / differential-reflow API **not** wired and **not** proposed — permission-gated on the owner as of 2026-08-01 (standing instruction on #3539). Verified by diff, not by assertion: this branch's diff against `origin/main` matches none of `patch_rows` / `Presentation(` / `strips=`. This PR touches only the existing `set_item` path

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -100,6 +100,7 @@ from .search_bar import SearchBar
 from .sent_queue import SentQueue
 
 if TYPE_CHECKING:
+    from textual.timer import Timer
     from textual_flowview import VisibilityHandle
 
     from reyn.interfaces.repl.read_model import ChatReadModel
@@ -246,6 +247,18 @@ def _apply_restored_state(msg: "OutboxMessage", entry: "Entry[OutboxMessage]") -
 #: snapshot (pre-session), which must NOT trigger a second read.
 _UNSET: object = object()
 
+#: #3570 — the minimum wall-clock gap between two repaints of the SAME streamed
+#: reply. Deltas arrive at the provider's rate (measured: up to ~1000/s through a
+#: proxy that packs many SSE events into one read); repainting at that rate spends
+#: the loop re-presenting and re-rendering an O(body) markdown body far more often
+#: than a terminal can show. 1/30 s is the knee of the measured curve on the real
+#: TUI path (2000 deltas / 60 KB reply): render work 1825 ms → 990 ms and the worst
+#: single loop block 118 ms → 96 ms, while going on to 1/20 s bought a further ~5%
+#: for 50% more latency. It is a REPAINT budget only — the accumulated text is
+#: never gated by it (see :class:`_StreamingReply`), and no deferral outlives it
+#: (:meth:`TextualChatApp._schedule_streaming_catchup`).
+_STREAM_REPAINT_MIN_INTERVAL = 1 / 30
+
 
 @dataclass(slots=True)
 class _StreamingReply:
@@ -261,12 +274,22 @@ class _StreamingReply:
       optimisation, never a data path.
     - :attr:`rendered` — how much of :attr:`text` the flow entry has actually
       been handed (via ``Entry.set_item``). Equal to :attr:`text` while the row
-      is on screen; LAGS it while the row is scrolled out of view.
+      is on screen AND the repaint budget below allowed the last delta through;
+      LAGS it while the row is scrolled out of view, or within the budget window.
 
     ``rendered != text`` is therefore exactly "this row owes the viewport a
     repaint", and :meth:`TextualChatApp._flush_streaming_reply` is the only
-    thing that closes the gap — driven either by the next delta (while visible)
-    or by ``on_show`` when the row scrolls back (while it was not).
+    thing that closes the gap — driven by the next delta whose arrival is at
+    least :data:`_STREAM_REPAINT_MIN_INTERVAL` after :attr:`last_repaint`, by
+    the catch-up timer that bounds that deferral
+    (:meth:`TextualChatApp._schedule_streaming_catchup`), or by ``on_show`` when
+    the row scrolls back (while it was not visible).
+
+    :attr:`last_repaint` is the app-clock reading of the last such flush — the
+    #3570 repaint budget's whole state. It is a RENDER throttle: deltas arriving
+    inside the window still land on :attr:`text` in full, they just do not each
+    buy their own ``set_item`` (and, through the revision bump, their own
+    present + strip render of the whole accumulated body).
 
     :attr:`handle` is the ``FlowView.track_visibility`` registration whose
     ``on_show``/``on_hide`` maintain :attr:`visible`; :meth:`release` stops it
@@ -279,6 +302,7 @@ class _StreamingReply:
     rendered: str
     visible: bool = True
     handle: "VisibilityHandle | None" = None
+    last_repaint: float = 0.0
 
     @property
     def pending(self) -> bool:
@@ -881,6 +905,12 @@ class TextualChatApp(App):
         # session switch, so a chain_id can never leak past its turn's
         # completion.
         self._streaming_replies: "dict[str, _StreamingReply]" = {}
+        # #3570: the one-shot timer that BOUNDS a repaint deferral, or ``None``
+        # when nothing is deferred. Not per-session state (it holds no session
+        # identity and its callback iterates whatever is in-flight at the time),
+        # so it is deliberately NOT in :attr:`_PER_SESSION_DICT_STATE` — a switch
+        # clears the records and the timer then finds nothing to flush.
+        self._streaming_catchup: "Timer | None" = None
         # #3283 ④: the status snapshot's keyed per-turn token/cost lookup
         # (``turn_usage_fn`` = ``Session.turn_usage``), cached off the snapshot
         # :meth:`_refresh_live_chrome` already reads once per arriving frame.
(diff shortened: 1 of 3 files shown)


tya5:
**[per-PR-coder]** — CI `pytest (3.11)` failed on `tests/test_right_gutter_label_visible_3536.py::...` (`_CC_DIM no longer emits a concrete foreground`). Measured, not assumed, before calling it unrelated:

- This PR's file set is exactly three files — `src/reyn/interfaces/inline/textual_chat/app.py`, `tests/test_stream_repaint_coalesce_3570.py`, `tests/test_textual_chat_streaming_visibility_3283.py`. Neither the failing test nor the palette module nor `test_markdown_palette_gate_3469.py` is in the diff (`git diff --name-only` against this PR's base).
- The failing file passes locally on this branch (4 passed).
- The full suite passes locally on this branch on flowview v0.9.0: **9842 passed, 36 skipped, 0 failed**.

That matches the known process-global rich `Style.parse` memo pollution filed as #3572 (`-n auto` worker distribution, ~1/5). Job re-run requested.

tya5:
**[per-PR-coder]** — status: every check on this PR is green except `pytest (Python 3.11)`, which is still running after the re-run (its 3.12 sibling passed on the same commit). Local full suite on this branch, on flowview v0.9.0: **9842 passed, 36 skipped, 0 failed**. I will not consider the Test plan's `pytest` item closed until that job reports; if it fails again on `test_right_gutter_label_visible_3536.py` specifically, that is #3572 (measured file-set check in the comment above), and any other failure is mine to fix.

tya5:
**[per-PR-coder]** — 2点、更新しました（`2172562e6`）。

## ① patch_rows 恒久指示（owner, 2026-08-01）への準拠を *diff で* 確認

★ **主張ではなく実測で置きます**: 本ブランチの `origin/main` に対する diff は `patch_rows` / `Presentation(` / `strips=` のいずれにも一致しません（#3574 側も同様に0件）。★ **触っているのは既存の `set_item` 経路だけです。** Test plan の該当項目もその文言に差し替えました（「owner の deferral」ではなく **許可制** と明記）。

★ 作業中に「ここは差分リフロー API の方が綺麗」と感じた箇所はありましたが、★ **提案せずに進めました。** 報告としてのみ: 本 PR が減らしたのは *repaint の回数* であって *1回あたりの O(body) コスト* ではないので、その項は残っています（数値は本文の表のとおり）。

## ② コメント内の測定値を v0.9.0 で取り直し（doc-drift 修正）

★ `app.py` に残していた `set_item 1979 → 53 / present 89 → 37 / 1825 ms → 990 ms` は **v0.8.0 の数字で、バンプで falsify されていました。** ★ 正しくは **`set_item` 1979 → 75 / `present` 1908 → 72 / 16.1 s → 3.3 s**（v0.9.0、#3574 の中断点が入った状態）。**出荷する base で真でない測定値を残すのは、stale doc と同じクラス**と判断して同 PR 内で直しました。

**local full suite（#3574 の最新 head 上、v0.9.0）: 9841 passed / 36 skipped / 0 failed。** 本 PR 側は rebase 済みで、CI は #3572 の land 後に再実行いただければと思います。

tya5:
**[per-PR-coder]** — このPRは base ブランチ削除により GitHub に閉じられ、reopen が API で拒否されます。★ **同一内容を `main` 宛に再提出しました: #3577。** レビュー履歴（architect の co-vet と私の回答）は本 PR に残っているので、#3577 から参照しています。

PR #3576, opened 2026-07-31
test(conftest): reset rich's process-global Style SGR memo between tests — #3572

tya5:
**[per-PR-coder]** — Fixes #3572. The order-dependent CI red behind #3571 / #3575.

## The correction that matters: the named poisoner is not the poisoner

#3572 identifies `test_markdown_palette_gate_3469.py::test_gate_measures_the_theme_not_another_consoles_downgrade_memo` as the test that poisons rich's process-global `Style` memo and never restores it. **Measured: it does restore.** It poisons, then calls `_render_ansi(themed=True)` → `_memo_cleared_theme()`, which clears `_ansi` on exactly the same `Style.parse`-shared instances and re-renders them truecolor. Running it immediately before the victim is **green**, with `TERM`/`COLORTERM` unset (CI's environment) and with them set:

```
$ unset TERM COLORTERM
$ pytest tests/test_markdown_palette_gate_3469.py::test_gate_measures_the_theme_not_another_consoles_downgrade_memo \
         tests/test_right_gutter_label_visible_3536.py::test_the_dim_colour_constant_stays_a_colour
2 passed
```

The real poisoners are ordinary tests, and the mechanism has a **second** un-keyed cache, not just `_ansi`:

- `RichChatRenderer` / `InlineChatRenderer` (`repl/renderer.py:268`, `:895`) build `Console(force_terminal=True, ...)` and leave colour detection to rich — correct for production.
- Under CI (`TERM` and `COLORTERM` unset) rich's `_detect_color_system()` returns **standard**. Measured:

```
detected color_system: standard
memo after production-shaped console: '90'
truecolor console emits: '\x1b[90my\x1b[0m'
```

- So any test that drives one of those two renderers memoizes `_CC_DIM` (`#6b7280`) as bright-black `'90'`, and a later test that explicitly asks for truecolor re-emits it.

Locally this hides because `TERM=xterm-256color` detects eight-bit, and `--dist load` decides per run which worker gets which pair — hence "1 in 5", not a flake.

## The fix

A new autouse fixture `_isolate_rich_style_ansi_memo` in `tests/conftest.py` resets rich's memo after every test — leaker-agnostic, so it closes the class rather than the hole. Same shape (and same rationale) as the existing `_isolate_budget_limit_context` fixture directly above it. Guarding readers one at a time is what left #3536's gutter test exposed hours after it landed.

[...]

PR #3577, opened 2026-07-31
feat(tui): repaint budget for streamed replies — closes the 5x regression #3574 left on main — #3570 (2/2)

tya5:
**[per-PR-coder]** — **this replaces #3575, which GitHub CLOSED when its base branch (`perpr/3570-drain-yield`) was deleted by `gh pr merge --delete-branch` on #3574. It cannot be reopened (the API refuses once the base is gone). The content is identical, rebased from the deleted base onto `main`.** Review history lives on #3575 — the architect's co-vet and my responses are there, and nothing in this branch changed in response to anything after that.

> 🔴 **main is currently carrying the 5x regression this PR closes.** #3574 landed alone, so `main` has the drain's unconditional suspension point WITHOUT the repaint budget — which is the combination I measured as the worst of the three. Re-measured on `main` as it stands (`61c320c01`) against this branch, back to back on the same loaded machine:
>
> | 2000 deltas / 60 KB reply, ~1000 deltas/s | `main` today | this PR |
> |---|---|---|
> | wall-clock | **16.07 s** | **2.93 s** |
> | `present()` calls | **1926** | **55** |
> | `set_item` | 838 | 69 |
> | animation ticks / expected | 384 / 964 (40%) | 119 / 176 (68%) |
> | worst single loop block | 303 ms | 125 ms |
>
> This is not a new finding — it is the table that was already in #3574's body as the reason the two had to land together. It is restated here as a measurement of `main`, not a prediction.

## What this changes

`_handle_agent_delta_event` accumulated text **and** issued a `set_item` on every visible delta. Deltas arrive at the provider's rate, not the terminal's, and each `set_item` costs a present + a strip render of the whole accumulated body.

The coalescing is on the **`rendered` side only** — the line the existing design already drew:

- `_StreamingReply.text` — `+=` per delta, **unconditional**, untouched by this PR. It is the data path; a delta is an increment, not a snapshot, so nothing here may ever be "last-wins".
- `_StreamingReply.rendered` — how much of that the entry has been handed. This PR adds a per-reply repaint budget (`_STREAM_REPAINT_MIN_INTERVAL`, 1/30 s, read off the app's own injectable clock): a delta arriving inside the window accumulates without a `set_item`.

**The deferral is bounded**, and the bound does not depend on the producer: every skipped repaint arms a one-shot catch-up timer that flushes unconditionally, so held text reaches the screen within one window whether deltas keep arriving forever (queue never empties) or stop entirely (model pauses mid-reply).

**Why 1/30 s**: the knee of the measured curve on the real TUI path — going on to 1/20 s bought ~5% more for 50% more latency.

## Two decisions, stated rather than inferred

**1. A non-delta frame does NOT flush pending repaints.** Enumerated over the *closed* vocabularies rather than extrapolated from one kind — `DISPLAY_KINDS` (15) + `CONTROL_KINDS` (2) + `renderer_chat_events()` (14), each checked against `_ingest_frame`/`_pump_frames`:

| group | what it does with a streaming reply's rendered text | forces a flush? |
|---|---|---|
| `tool_call_started` / `_completed` / `_failed` (with `op_id`) | touch their own entry (`_running_tools`, `_coalesce_tool_result`) | no |
| terminal `agent` (with `chain_id`) | pops the record and writes the **authoritative full text** — subsumes any pending repaint | no (it *is* the settle) |
| `__copy_last_reply__` | reads `_recent_replies`, filled from the completion frame's text, never from the entry | no |
| `__rewind_list__`, `intervention`, `presentation`, `status`, `trace`, `error`, `system`, `user`, `reasoning`, `__attach_request__` | append a NEW entry; an entry's position is fixed at append time, so painting a row above later cannot reorder anything | no |
| `__end__`, `__session_switch_request__` | control; `__end__` stops the app | no |
| event frames `user_submitted` / `turn_started` / `inbox_cancel` / `intervention_answer_submitted` / `session_halted` / turn-end / `tool_called` etc. | sent-queue region, chrome, orphan sweep — none reads a streamed row's text | no |
| **`session_attached`** | the #3310 N2 barrier | **see below** |

Flushing on the others would make the coalescing decay in proportion to how much else the turn emits — worst exactly on the tool-heavy turn where the loop is busiest.

**2. At the `session_attached` barrier a pending repaint is DISCARDED, not flushed.** The barrier resets every per-session client state and rehydrates from the new session; painting the old session's accumulation after that point would put a previous conversation's text on the new one's screen — not a loss (the old session's history holds it) but contamination, which is worse. The mechanism is the barrier handler clearing `_streaming_replies`, which the catch-up timer reads when it fires; a record dropped there can no longer be flushed by anything. Gated below.

## Measured, real path, flowview v0.9.0

| | main *before* #3574 | this PR (on today's main) |
|---|---|---|
| paced (~1000 deltas/s), 2000 deltas: `set_item` | 1979 | **75** |
| paced: `present()` | 90 | 72 |
| paced: wall-clock | 3.17 s | 3.29 s |
| paced: animation ticks / expected | 106 / 190 | 118 / 198 |
| fully buffered burst: `present()` | 3 | 2 |
| burst: animation ticks / expected | 1 / 4 | **13 / 15** |

Arrival-rate sweep, mixed backlog (2 chains + tool frames), 300 deltas at each point — the property, rather than a single number:

| arrival rate | max queue occupancy | repaints, main | repaints, with this PR |
|---|---|---|---|
| ~95 /s | 2 | 298 | 150 |
| ~250 /s | 6 | 298 | 54 |
| ~940 /s | 20 | 298 | 20 |
| ~2300 /s | 50 | 298 | 8 |
| whole reply in one batch (~31000 /s) | 300 | 298 | 2 |

Main repaints once per arrival at every rate. With the budget, repaints follow elapsed time (~30/s per reply), not arrivals.

**Where progressive display stops being progressive**: when the entire reply arrives inside one 33 ms window, the viewer gets ~2 repaints — there is nothing to show progressively in a reply that arrived in 10 ms. Above that (any rate that spreads the reply over more than a window) repaints stay plural and evenly spaced.

## Test plan

- [ ] 🔴 **REGRESSION OPEN ON MAIN until this merges.** The co-merge gate is moot — #3574 is already in. What is true now is narrower and more urgent: every streamed reply on `main` costs ~5x the UI time it did before #3574, and this PR is the only thing that closes it. Ticking this box means "the regression window is closed", not "the pair shipped together".
- [x] `tests/test_stream_repaint_coalesce_3570.py`
  - repaints stop tracking arrivals, and the text is complete anyway (the completion frame's authoritative full text)
  - **mixed backlog** (2 chain_ids interleaved + tool frames on the same ordered queue): repaints follow the replies in flight, not the arrivals; then one more delta past the window proves nothing was dropped
  - the deferral is **bounded**: with the clock frozen, held text is still painted — mid-stream, before any completion frame
  - **barrier**: text deferred when `session_attached` arrives never appears after the reset, with a control leg in the same run showing a deferred repaint normally *does* reach the screen
- [x] **strip-falsify, each property separately**: making the repaint unconditional -> the coalescing gates go RED, the bound gate stays green; removing `_schedule_streaming_catchup()` -> the bound gate goes RED, the coalescing gates stay green
- [x] `tests/test_textual_chat_streaming_visibility_3283.py` updated: its "a VISIBLE reply is fed every delta" leg now drives the app's injected clock past the budget between deltas, so it keeps measuring the ③ *visibility* gate at full strength instead of accidentally measuring this budget
- [x] `ruff check src tests`
- [x] `python scripts/test_tier_audit.py --strict` on both changed test files
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/app.py`
- [x] Measurements taken on flowview **v0.9.0** (`__version__` read directly)
- [x] `patch_rows` / `Presentation(strips=)` / differential-reflow API **not** wired and **not** proposed — permission-gated on the owner as of 2026-08-01 (standing instruction on #3539). Verified by diff, not by assertion: this branch's diff against `origin/main` matches none of `patch_rows` / `Presentation(` / `strips=`. This PR touches only the existing `set_item` path

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -100,6 +100,7 @@ from .search_bar import SearchBar
 from .sent_queue import SentQueue
 
 if TYPE_CHECKING:
+    from textual.timer import Timer
     from textual_flowview import VisibilityHandle
 
     from reyn.interfaces.repl.read_model import ChatReadModel
@@ -246,6 +247,18 @@ def _apply_restored_state(msg: "OutboxMessage", entry: "Entry[OutboxMessage]") -
 #: snapshot (pre-session), which must NOT trigger a second read.
 _UNSET: object = object()
 
+#: #3570 — the minimum wall-clock gap between two repaints of the SAME streamed
+#: reply. Deltas arrive at the provider's rate (measured: up to ~1000/s through a
+#: proxy that packs many SSE events into one read); repainting at that rate spends
+#: the loop re-presenting and re-rendering an O(body) markdown body far more often
+#: than a terminal can show. 1/30 s is the knee of the measured curve on the real
+#: TUI path (2000 deltas / 60 KB reply): render work 1825 ms → 990 ms and the worst
+#: single loop block 118 ms → 96 ms, while going on to 1/20 s bought a further ~5%
+#: for 50% more latency. It is a REPAINT budget only — the accumulated text is
+#: never gated by it (see :class:`_StreamingReply`), and no deferral outlives it
+#: (:meth:`TextualChatApp._schedule_streaming_catchup`).
+_STREAM_REPAINT_MIN_INTERVAL = 1 / 30
+
 
 @dataclass(slots=True)
 class _StreamingReply:
@@ -261,12 +274,22 @@ class _StreamingReply:
       optimisation, never a data path.
     - :attr:`rendered` — how much of :attr:`text` the flow entry has actually
       been handed (via ``Entry.set_item``). Equal to :attr:`text` while the row
-      is on screen; LAGS it while the row is scrolled out of view.
+      is on screen AND the repaint budget below allowed the last delta through;
+      LAGS it while the row is scrolled out of view, or within the budget window.
 
     ``rendered != text`` is therefore exactly "this row owes the viewport a
     repaint", and :meth:`TextualChatApp._flush_streaming_reply` is the only
-    thing that closes the gap — driven either by the next delta (while visible)
-    or by ``on_show`` when the row scrolls back (while it was not).
+    thing that closes the gap — driven by the next delta whose arrival is at
+    least :data:`_STREAM_REPAINT_MIN_INTERVAL` after :attr:`last_repaint`, by
+    the catch-up timer that bounds that deferral
+    (:meth:`TextualChatApp._schedule_streaming_catchup`), or by ``on_show`` when
+    the row scrolls back (while it was not visible).
+
+    :attr:`last_repaint` is the app-clock reading of the last such flush — the
+    #3570 repaint budget's whole state. It is a RENDER throttle: deltas arriving
+    inside the window still land on :attr:`text` in full, they just do not each
+    buy their own ``set_item`` (and, through the revision bump, their own
+    present + strip render of the whole accumulated body).
 
     :attr:`handle` is the ``FlowView.track_visibility`` registration whose
     ``on_show``/``on_hide`` maintain :attr:`visible`; :meth:`release` stops it
@@ -279,6 +302,7 @@ class _StreamingReply:
     rendered: str
     visible: bool = True
     handle: "VisibilityHandle | None" = None
+    last_repaint: float = 0.0
 
     @property
     def pending(self) -> bool:
@@ -881,6 +905,12 @@ class TextualChatApp(App):
         # session switch, so a chain_id can never leak past its turn's
         # completion.
         self._streaming_replies: "dict[str, _StreamingReply]" = {}
+        # #3570: the one-shot timer that BOUNDS a repaint deferral, or ``None``
+        # when nothing is deferred. Not per-session state (it holds no session
+        # identity and its callback iterates whatever is in-flight at the time),
+        # so it is deliberately NOT in :attr:`_PER_SESSION_DICT_STATE` — a switch
+        # clears the records and the timer then finds nothing to flush.
+        self._streaming_catchup: "Timer | None" = None
         # #3283 ④: the status snapshot's keyed per-turn token/cost lookup
         # (``turn_usage_fn`` = ``Session.turn_usage``), cached off the snapshot
         # :meth:`_refresh_live_chrome` already reads once per arriving frame.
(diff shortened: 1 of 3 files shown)


tya5:
**[per-PR-coder]** — 復旧完了、レビュー可能です。

## やったこと（実測付き）

★ **squash 済みの3コミットを再生させない形で載せ替えました**（lead-coder 提案の `--onto` を確認のうえ使用）:
```
git rebase --onto origin/main 1e2765a67 perpr/3570-repaint-coalesce
```
★ **結果の diff は3ファイルのみ**（`app.py` / 新テスト / #3283 テスト更新）で、transport 側は1行も含みません。**確認は主張ではなく diff で**: `git diff origin/main perpr/3570-repaint-coalesce -- src/reyn/interfaces/transport/ tests/test_stream_drain_yield_3570.py` が空 ＝ **main の #3574 分と本ブランチは同一**、二重適用も欠落もありません。

## main の回帰を測り直しました（予測ではなく現状の測定）

**同一マシン・連続実行**（2000 deltas / 60 KB / ~1000 deltas/s）:

| | `main` (`61c320c01`) | 本 PR |
|---|---|---|
| wall-clock | **16.07 s** | **2.93 s** |
| `present()` | **1926** | **55** |
| anim ticks / expected | 384/964 (40%) | 119/176 (68%) |
| 最悪の loop block | 303 ms | 125 ms |

⚪ **正直に**: 上の測定の直前に取った2本（41.6 s / 21.0 s）は **load average 11.8 の負荷で汚染** されており捨てました。★ ただしその2本でも `present` は 420 / 260 で **main の 1926 より一桁少なく**、coalesce 自体は効いていました。**採用したのは負荷が落ち着いてから main と交互に取った上の組**です。

## gate

- `ruff check src tests` ✅ / `test_tier_audit --strict`（2ファイル）✅ / `verify_module_docstrings`（`app.py`）✅
- 対象3ファイル + drain gate: **15 passed** ✅
- full suite: 実行中（結果は追ってコメントします）。CI も走ります。

## MERGE GATE の文言を差し替えました

★ **「#3574 と同時に merge」はもう成立しない**ので、Test plan の項目を **「main に回帰が開いている。本 PR がそれを閉じる」** に書き換え、未チェックのままにしています。★ **チェックの意味が「ペアで出荷した」から「回帰ウィンドウを閉じた」に変わった**ことを明示しています。

⚪ **lead-coder へ**: `--delete-branch` の件、こちらの作業は復旧できています。★ **記録として1点だけ** — MERGE GATE は Test plan（＝merge gate が読む面）に置いてあったので機構としては機能していましたが、★ **merge 操作そのものは Test plan を読まない**ので、stacked PR では「base を消さない」の側でしか守れません。私の gate の置き方の限界です。

tya5:
**[per-PR-coder]** — local full suite on this branch (today's `main` + this PR, textual-flowview v0.9.0): **9845 passed, 36 skipped, 0 failed** (3:14). CI's two `pytest` jobs are still running; every other check on this PR is already green.

tya5:
**[lead-coder]** — ✅ **merge します。★ main の回帰は私が作ったものです。**

## 経緯（記録）

★ **#3574 を `gh pr merge --delete-branch` で merge した際、削除された base ブランチによって #3575 が GitHub に CLOSED にされました。** ★ **その結果 main が「①だけ」＝ 5× のスループット回帰を含む状態になり、本 PR がそれを閉じます。**

★★ **私は担当に「同時 merge 必須を *散文でなく Test plan の未チェック項目* に置け」と指示し、それは正しく機能していました。** ★ **gate は「単独 merge するな」を守らせましたが、「base ブランチを消すな」は守れません。** ★ **私は gate の隣の穴から同じ結果を作りました。**

> ★★ **貴殿の記録が正確です:「merge gate は *レビュアが読む面* に置いたが、merge *操作* はそこを読まない。stacked PR で実際に効くのは base ブランチを消さないことだけ。」**

★ **私の恒久チェックリストに「stacked PR の base に `--delete-branch` を使わない／merge 前に base 依存の open PR を数える」として記録しました。** ★ **本 merge の前に実際に数えて、依存ゼロを確認しています。**

## 承認理由

★ **復旧が「主張でなく diff」で検証されています** — `git diff origin/main <branch> -- transport/ + drain gate` **が空**。**二重適用も欠落も無い。**

★ **回帰も予測でなく再測**（main 16.07s / present 1926 → 本 PR 2.93s / present 55、同一マシンで交互）。★ **負荷 11.8 で汚染された2回を破棄したと明記されているのも正しい。**

★ **`MERGE GATE` を「チェック済みのまま引き継がず、意味を書き換えた」判断も正しい** — **「ペアが出た」から「回帰の窓が閉じた」へ。**

**#3576（conftest の memo リセット）は base 遅れのため、この後に処理します。**

PR #3581, opened 2026-07-31
test(tui): make the two delta tests that read the RENDER deterministic under the repaint budget — #3570 follow-up

tya5:
**[per-PR-coder]** — fixes the red on `main` that #3579's CI surfaced. Follow-up to #3577.

## Verdict: **(A) — no text is lost. And it is my flake, not a stale expectation.**

The two readings had to be separated before anything else, so they were measured separately on `main` (`21938faf6`), 20 runs:

| | value | runs |
|---|---|---|
| `_StreamingReply.text` (data path) | `'Hello, world'` | **20/20** |
| `entry.item.text` (render) | `'Hello'` or `'Hello, world'` | **9/20 lagging** |

So the accumulated text is complete every time — the append is unconditional and this PR does not touch it. What lags is the repaint, by design.

★ But "the test's expectation is stale" would be the wrong conclusion, and the interesting half is why:

```
run  1: gap= 48.2ms  TEXT='Hello, world'  ENTRY='Hello'         FAIL
run  3: gap= 70.2ms  TEXT='Hello, world'  ENTRY='Hello, world'  PASS
```

The assertion holds **iff the wall-clock gap between the two pushes exceeded `_STREAM_REPAINT_MIN_INTERVAL` (33 ms)** — i.e. iff `pilot.pause()` happened to be slow enough. That is a coin flip, not a gate. #3577 did not invalidate these tests; **it made them timing-dependent**, which is the #3473 class we spent last night removing. One of them was already failing silently before today's red.

## Reproduction rates (measured, not called "a flake" after one run)

In the fast regime (in-process, short pauses — the regime CI's 3.12 runner hit):

| test | before | after |
|---|---|---|
| `test_agent_delta_draws_exactly_one_coalesced_entry` | **2/30** (and 9/20 in a second harness) | **0/30** |
| `test_stream_on_off_render_equivalence` | **8/30** ← *already flaky, nobody had seen it fail yet* | **0/30** |
| `test_mid_stream_join_finalizes_to_one_entry_full_text` | 0/30 | 0/30 |

Through `pytest` on this machine all of them pass 20/20 — pytest's own overhead makes the pauses long enough. That is precisely why the rate is a property of the machine and not of the code, and why "re-run it" was never going to settle it.

## The fix: deterministic, not "wait longer"

[...]

PR #3586, opened 2026-08-01
fix(security): /session new inherits its invoker's per-session narrowing — part of #3562

tya5:
**[per-PR-coder]** — part of #3562.

## 🔴 This PR is NOT an escape fix — read it as an owner policy decision

★ **The rationale has been replaced since this PR was opened, and the original one is retired.** It first stood on containment: #3561 had measured that `/session new` was reachable from MODEL OUTPUT, which made an un-narrowed child an escape. **#3595 ruled that reachability a defect and closed it on `main` (#3596 / #3597)** — an agent-step prompt no longer claims `kind="user"`, so it never enters the slash dispatch at all. ★ **Do not read this PR as blocking an escape; that hole is already shut, elsewhere.**

What this PR stands on is the owner's ruling on the plain question — *if an operator switches `write_file` off for their session and then runs `/session new`, should `write_file` work in the new session?* — **"使えるべきじゃないね"**. Operator intent persists across the spawn. That claim never depended on who could reach the command, which is why closing the reachability changed nothing here. The rationale is now stated in `slash/session.py` at the call site, in the #3562 test module's docstring, and in the `_SITE_PARENT_LAYERS` declaration — not only in this body.

## ✅ User-visible behaviour change — owner-approved

**An operator whose own session is narrowed now gets a narrowed new session.** The trade, stated as a trade: inheriting can produce a session that is *too narrow*, which the operator notices immediately (something is refused) and can recover from (re-open from an unnarrowed session, or lift the narrowing). Not inheriting produces a session that is *too wide*, which nobody notices and no surface reports after the fact — including the case where the operator never knew they were narrowed (#1827 taint-derived narrowing is not operator-set).

**There is no second behaviour change.** No session-lifecycle change of any kind: nothing gained or lost a `session_spawned` record, nothing became (or stopped being) rewind-tracked or re-materialisable, and `spawn_session_recorded`'s own sequence is unchanged (see *Measured, not assumed* below — I tried to simplify it and the tests said no).

## The defect

`interfaces/slash/session.py::session_cmd` called `AgentRegistry.spawn_session` carrying nothing from its invoker, so the child's #2103-S1a per-session layer was empty however narrow its invoker was — and stayed empty for an operator who had just narrowed themselves. (#3561 first surfaced the site by measuring a model-output path to it; that path is gone as of #3596, and the empty-layer defect is unchanged by its removal.)

## The lifecycle decision — (a), written into the code

The issue asked for a deliberate choice between (a) adding a `narrowing` channel to `spawn_session` and (b) routing `/session new` through `spawn_session_recorded`. **This PR takes (a)**, and `spawn_session`'s docstring carries the reason so a later reader can tell a decision from an omission:

(b) would have carried three unrelated changes along with the fix — a `session_spawned` WAL record making an operator-opened session rewind-tracked and re-materialisable (so a rewind past the command would start **dropping the operator's session**, a lifecycle change nobody asked for), a spawn-time `refresh_config_projections()`, and async-ness. (a) changes exactly one observable thing: the child's envelope. The channel also belongs on the primitive on the merits — it is where the sid becomes known, which is why #3561 put the *injection* there, and it is the seam all five direct callers share.

Both entry points now write the file through **one writer** (`_persist_session_narrowing`), so they cannot drift from each other or from its reader.

## ★ Measured, not assumed — the one place my reasoning was wrong

I first had `spawn_session_recorded` hand its `narrowing` down the new channel too (one writer, one injection point), reasoning that its later re-inject was redundant because `refresh_config_projections()` re-resolves the envelope from base *with the sid* and would therefore read the file. **The full suite falsified that**: `test_spawn_session_recorded_enforces_narrowing_on_live_session` and `test_spawn_ephemeral_session_narrowing_applied` went RED with an empty live `tool_deny`.

The cause, traced rather than guessed: `reapply_visibility_override` re-resolves from base and **SETs**. On a session with **no registry back-reference** there is no base to re-resolve, so it sets an ALLOW-ALL envelope and silently discards anything injected before it. The old ordering (inject *after* the refresh) was load-bearing, not incidental.

So `spawn_session_recorded` keeps its own write + inject after the refresh, and that ordering is now pinned by a comment naming both tests. Its `_NARROWING_EXEMPT_SITES` entry states the measurement.

🔎 **Latent hazard found on the way, not fixed here** (flagging for lead to decide whether to file): "re-resolve from base, then SET" *widens to allow-all when the base is unavailable* — a fail-open in a security-core path. Today it is contained (production sessions carry the registry ref; the two paths that inject without a later refresh are recovery and `/session new`), which is why I did not widen this PR's scope into #2285/#3097 territory on my own judgement.

## Composition — uniform, no cross-identity branch

`capability_profile.compose_narrowing_mappings(parent, None)` (deny ∪ / allow ∩ / absent allow = ⊤), the three sibling sites' rule. No branch for the cross-identity case: `name = reg.attached_name`, so on the operator path — the only face that reaches this site now — the caller *is* the attach target and the identities coincide, leaving no live cross-identity case to branch on. A branch would be a lenient special case for exactly the caller a uniform restrict-only rule exists to bound.

⚠️ Architect's original argument for uniformity was sharper and is now retired with the reachability it rested on ("the identities differ only when model output reaches the site, so *the identity differs* is the same event as *the escape is happening*"). **The decision is unchanged; only its support is.** The test still drives the cross-identity shape (invoker under one agent, child under the attached one), so the uniform rule is exercised rather than assumed — a lenient special case would pass the same-identity leg.

## The inherited-restriction line

`/session new`'s reply names what the child inherited (denied capability names, plus a second line when no tools remain), rendered from the child's existing `capability_visibility_state()` — no new mechanism.

🔴 **It is not the witness.** That surface re-resolves with the sid on every read and was GREEN on the broken code #3561 fixed. The inheritance is witnessed on the side-effect side (a denied tool's file does not appear); the display has its own separate test for its own separate claim, plus a negative twin asserting an unnarrowed invoker's reply carries no inheritance line — otherwise a constant string would satisfy the positive leg.

## Strip-falsify (both halves; anchors confirmed unique with `grep -c` → 1 first)

| strip | result |
|---|---|
| baseline | `tests/test_3562_*.py` **4 passed** |
| remove `narrowing=inherited` at the slash call site | **2 failed, 2 passed** — RED: the enforcement witness + the display test; GREEN: both controls, correctly (they assert the *un*narrowed case) |
| restore | **4 passed** |
| remove `self._persist_session_narrowing(...)` in the primitive | **2 failed, 15 passed** across #3562 + #3556 + #3553 — the same two legs RED; the sibling seams stay green, correctly, because `spawn_session_recorded` writes its own (that is exactly the split the ordering finding above forced) |
| restore | **4 passed**; **76 passed** across the whole affected set |

## Gate bookkeeping

- `tests/test_3546_*.py`'s `_SITE_PARENT_LAYERS` entry for `("reyn/interfaces/slash/session.py", "session_cmd")` — the declaration of this gap — is replaced by what the site now composes, pointing at the new behavioural legs.
- ★ Adding the channel expired **all five** `_NARROWING_EXEMPT_SITES` entries, not just the slash one — `test_narrowing_exempt_sites_have_no_narrowing_channel` reads the live signature and was built to do exactly that. **This is the piece the issue body does not mention**; it is most of the test-side diff. One site left the table by *using* the channel; the remaining four are re-argued on the merits: `resolve_session` has no spawner to inherit from; the recovery pair would not pass a no-op but an **overwrite** of the recovering session's own durable layer (the primitive persists what it is given); `spawn_session_recorded` must inject later than the primitive offers (measured — above). The test is re-grounded as `test_no_exemption_claims_a_channel_that_exists`.
- #3561's reachability legs stay green — **verified by running them**, both before the rebase (the original `test_model_output_reaches_slash_dispatch_and_spawns_a_session`) and after it (#3596's inverse `test_model_output_cannot_reach_slash_dispatch_and_spawns_nothing` plus its operator control `test_an_operator_submitted_slash_command_still_spawns_a_session`). They measure WHO reaches the site; this PR measures what the child inherits — orthogonal axes, and the site declaration now says so explicitly so a later reader cannot mistake one pair for the other's evidence.
- Two pre-existing files drive `/session new` through hand-rolled stand-ins (`test_slash_session_1726.py`, `test_slash_budget_session_reload_cmds.py`) and broke loudly on the new reads, which is the correct failure. They are extended to mirror the real surface (`per_session_narrowing` → None, the two identity fields), with a docstring in each saying the inheritance is measured against **real instances** in `tests/test_3562_*.py` and must not be read off a stand-in that can only ever exercise the empty case.

## Recovery-feature gate

Not triggered on the merits, and this is a judgement rather than an omission: the reconstruction source for this layer is the sid's `config.yaml` under the per-session state dir, not a WAL-event — and #3564 already added the truncate-falsify pair for exactly that claim (`test_persisted_narrowing_survives_a_wal_truncation_past_its_spawn_record` + its un-narrowed twin), which this PR runs green unchanged. This PR does not add a new reconstruction source.

## Rebase onto #3596 / #3597 — what conflicted and how it was resolved

**One conflict, in `tests/test_3546_*.py`** — the `session_cmd` site declaration, which both sides had rewritten. ★ **No source conflict at all**: this PR touches `slash/session.py` and `registry.py`; #3596 split `session.py`, which this PR never edits.

The resolution is a **union of two claims, not a pick of one side**. `main`'s side said *who can reach the site* (the model-output path is closed; the site stays enumerated because this gate counts where an envelope is BORN, not where a model can reach). My side said *what the site now composes*. Both are true and neither implies the other, so the entry states both and its `measured_by` names all four legs — `_S3561`'s inverted pair **and** `_S3562`'s inheritance pair — with a sentence saying which pair measures which claim, so a later reader cannot take one as the other's evidence.

★ **Checked specifically for the hazard you named** — whether the resolution moved anything across the `_handle_user_message` / `_handle_inbox_text` split. It did not, and not by inspection alone: this PR's tests drive the invoker with `kind="user"`, the operator entry *above* the split (the only kind whose text is still slash-interpreted), and they spawn — measured, green. The harness docstring now states that the kind claim is deliberate and which side of the split it is on, so a future edit that moves it will read as a change rather than a detail.

**Prose that rested on the retired argument was rewritten, not left standing** — in `slash/session.py`'s call-site comment, this PR's test module docstring, its cross-identity test docstring, and the site declaration. ⚪ One drive-by: `test_3546_*.py`'s module docstring still asserted "model output … observes it reach `/session new` … the guess is false" in the present tense, which #3596 falsified and did not update. Fixed here because my resolution sits three lines below it and had to re-argue the same claim — leaving the two adjacent and contradicting would be exactly the drift the doc rule targets.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` on the four changed/added test files — 40 inspected, 0 errors, 0 warnings
- [x] `python scripts/verify_module_docstrings.py src/reyn/runtime/registry.py src/reyn/interfaces/slash/session.py` — OK
- [x] `pytest` from the repo root, foreground, terminal line read myself (result in the Verification section below)
- [x] Strip-falsify both halves: RED confirmed, restored, GREEN confirmed (table above)
- [x] Anchor uniqueness confirmed (`grep -c` → 1) before each strip
- [x] `scripts/check_pr_closing_intent.py`'s own `_CLOSING_RE` run over this body and every commit message — 0 closing declarations, 1 `part of` declaration for #3562

## Verification

| gate | result |
|---|---|
| `ruff check src tests` | All checks passed! |
| `python scripts/test_tier_audit.py --strict` (4 files) | 40 inspected · OK 40 · warnings 0 · errors 0 · exit 0 |
| `python scripts/verify_module_docstrings.py` (2 src files) | OK |
| `pytest` (repo root, foreground, **post-rebase**, exit 0) | **9867 passed, 36 skipped, 0 failed** in 739s (12:19) |

★ Identity of what was measured, since a suite number is only worth its subject: the run was launched **after** `git rebase --continue` and `git status` was **clean** at its start and at its end, so the tree it measured is exactly the two commits on this branch. Neither known intermittent fired — `test_textual_chat_gutter_toggle_3352.py:333` (#3590) and `test_fp0066_p3b_repo_knowledge_ingest.py:198` (#3594) both passed, so nothing here is attributed to them.

★ Both witness sets green together, run as one invocation (`22 passed`): this PR's inheritance pair + display pair, and #3596's `test_model_output_cannot_reach_slash_dispatch_and_spawns_nothing` + `test_an_operator_submitted_slash_command_still_spawns_a_session`, plus the #3546 gate.

## Docs

`docs/concepts/multi-agent/sessions.md` described permissions as identity-scoped and shared across an Agent's sessions, with no mention of the session-scoped narrowing layer this PR makes visibly inheritable. A paragraph now states the layer, the inheritance, and the reason. (No `.ja.md` obligation invoked — the ja sibling asserts nothing this PR falsifies.)

### DIFF
--- a/docs/concepts/multi-agent/sessions.md
+++ b/docs/concepts/multi-agent/sessions.md
@@ -53,6 +53,16 @@ outbox, the current task, and transient run state. Identity-scoped things — me
 permissions, workspace scope, peer addressing — live on the Agent and are shared by
 all of that Agent's Sessions; conversation-scoped things stay isolated per Session.
 
+One capability layer is nonetheless *session*-scoped: a session can be **narrowed**
+below its agent's envelope (restrict-only — a narrowing can never grant what the
+identity does not already have). That layer is inherited on spawn: a session opened
+from a narrowed session — by `/session new`, by a pipeline, by an agent step, or by
+the `session_spawn` tool — is born inside the opener's narrowing, composed
+most-restrictive-wins (denials union, allow-lists intersect). So an operator whose own
+session is narrowed gets a narrowed new session, and `/session new` names what it
+inherited in its reply. The alternative — a fresh session silently starting wide — is
+the one an operator cannot notice.
+
 Persistence and recovery key on the **Session**:
 
 - **Per-session persistence** — each Session's state is snapshotted independently,
--- a/src/reyn/interfaces/slash/session.py
+++ b/src/reyn/interfaces/slash/session.py
@@ -6,7 +6,11 @@ structural substrate (N Sessions per Agent, keyed by session-id) landed in
 Stage 3; this is the REPL wiring.
 
   /session new          → open a new session under the attached agent (shared
-                          identity); prints the new session-id.
+                          identity); prints the new session-id. #3562: the new
+                          session is born INSIDE the invoking session's own
+                          per-session capability narrowing (#2103-S1a), composed
+                          the same way the three sibling spawn sites compose
+                          theirs, and the reply names what it inherited.
   /session switch <sid> → focus another session of the attached agent. Routed
                           through the registry forwarder (like ``/attach``) so
                           the focus flip + display re-wire are sequenced, not
@@ -24,11 +28,57 @@ from typing import TYPE_CHECKING
 from reyn.interfaces.slash import reply, reply_error, slash
 from reyn.runtime.outbox import OutboxMessage
 from reyn.runtime.spawn_routing import ReviewedNA
+from reyn.security.permissions.capability_profile import compose_narrowing_mappings
 
 if TYPE_CHECKING:
     from reyn.runtime.session import Session
 
 _USAGE = "usage: /session new | /session switch <sid> | /session list"
+
+#: How many denied capability names the inherited-restriction line spells out before
+#: it stops and just counts the rest — the line is an orientation aid in a REPL reply,
+#: not the full census (``/visibility`` renders that).
+_NAMED_DENIALS = 5
+
+
+def _inherited_restriction_lines(reg, name: str, sid: str) -> "list[str]":
+    """The operator-facing explanation of what a just-spawned session inherited.
+
+    #3562 (architect's in-PR requirement): with ``allow ∩`` composed uniformly, a
+    ``/session new`` reached from a NARROWED invoker can open a session with few or
+    even zero usable capabilities. That is the safe direction, but silent — the
+    operator sees a new session id and no reason for anything that later refuses. This
+    reads the child's own ``capability_visibility_state()`` (the existing #2285/#3378
+    read model — no new mechanism) and says so at spawn time.
+
+    ⚠️ Explanation, NOT enforcement, and deliberately not the witness that inheritance
+    works: ``capability_visibility_state`` re-resolves with the sid on every read, and
+    was GREEN on the broken code #3561 fixed, where the same envelope was resolvable
+    but not enforced. What proves the inheritance is a denied tool's side effect not
+    happening (``tests/test_3562_slash_session_new_narrowing_inheritance.py``); this
+    surface has its own separate test, for its own separate claim.
+
+    Empty when the child session is not retrievable (nothing truthful to say)."""
+    child = reg.get_session(name, sid)
+    if child is None:
+        return []
+    state = child.capability_visibility_state()
+    denied = sorted(row["name"] for row in state["denied_by_envelope"])
+    tools_left = sum(1 for row in state["authorized"] if row["kind"] == "tool")
+    lines: "list[str]" = []
+    if denied:
+        shown = ", ".join(denied[:_NAMED_DENIALS])
+        more = len(denied) - _NAMED_DENIALS
+        lines.append(
+            f"  ↳ inherited this session's capability narrowing — {len(denied)} "
+            f"denied in the new session: {shown}" + (f", +{more} more" if more > 0 else "")
+        )
+    if not tools_left:
+        lines.append(
+            "  ↳ no tools remain available in it; open the session from a session "
+            "that is not narrowed, or lift the narrowing, if that was not intended"
+        )
+    return lines
 
 
 @slash(
@@ -53,6 +103,35 @@ async def session_cmd(session: "Session", args: str) -> None:
     rest = parts[1].strip() if len(parts) > 1 else ""
 
     if sub == "new":
+        # #3562: the child is born under the ATTACHED agent, and the layer that decides
+        # its capability envelope is keyed by (agent, sid) — so the INVOKING session's
+        # #2103-S1a narrowing has to be carried explicitly or the child is born wider
+        # than whoever asked for it. The name-keyed layers (the agent's permissions
+        # declaration, topology capability_profile bindings, the #2081 _delegate floor)
+        # ride along for free; the #2285 /visibility toggle and the #1827-S4b ephemeral
+        # untrusted-context narrowing are deliberately not carried, same as the three
+        # sibling spawn sites (see test_3546's module docstring, layers 3 and 4).
+        #
+        # ★ WHY, since the answer changed under this code: an OWNER POLICY decision —
+        # if an operator has switched a capability off for their session and then opens
+        # a new one from it, that capability stays off. Operator intent persists across
+        # the spawn. The gap was originally argued as a containment one (#3561 had
+        # measured that model output reached this site, so an un-narrowed child was an
+        # escape); #3595 step 1 ruled that reachability a defect and closed it, which
+        # retires the containment argument WITHOUT touching this one — the policy claim
+        # never depended on who could reach the command.
+        #
+        # The composition rule is the siblings' rule applied UNIFORMLY (deny ∪, allow ∩,
+        # an absent allow = ⊤) — this site imposes nothing of its own, so the child term
+        # is None and the invoker's mapping stands. There is deliberately NO branch for
+        # the case where the invoking session's agent differs from the attached one:
+        # ``name`` is ``reg.attached_name``, so on the operator path the caller IS the
+        # attach target and the identities coincide. A branch would be a lenient special
+        # case for exactly the caller a uniform restrict-only rule exists to bound.
+        parent_narrowing = reg.per_session_narrowing(
+            session.agent_name, session.session_id,
+        )
+        inherited = compose_narrowing_mappings(parent_narrowing, None)
         try:
             # #2708 P3-item3: /session new opens a real attachable conversation session — the user
             # /session switches to focus + drain it; self-binding to the factory default is reviewed-NA.
@@ -61,11 +140,15 @@ async def session_cmd(session: "Session", args: str) -> None:
                 name,
                 presentation_consumer=_routing.presentation_consumer,
                 intervention_bridge=_routing.intervention_bridge,
+                narrowing=inherited,
             )
         except ValueError as exc:  # dup id (spawn_session guards)
             await reply_error(session, str(exc))
             return
-        await reply(session, f"opened session {sid!r} — /session switch {sid} to focus it")
+        lines = [f"opened session {sid!r} — /session switch {sid} to focus it"]
+        if inherited:
+            lines.extend(_inherited_restriction_lines(reg, name, sid))
+        await reply(session, "\n".join(lines))
         return
 
     if sub == "switch":
--- a/src/reyn/interfaces/slash/session.py
+++ b/src/reyn/interfaces/slash/session.py
@@ -69,9 +69,17 @@ def _inherited_restriction_lines(reg, name: str, sid: str) -> "list[str]":
     if denied:
         shown = ", ".join(denied[:_NAMED_DENIALS])
         more = len(denied) - _NAMED_DENIALS
+        # Says two separate true things rather than one convenient one: the child
+        # inherited this session's narrowing (a fact about what was passed), and its
+        # envelope denies these capabilities (a fact read off the child). It does NOT
+        # claim every denial listed is DUE to the inheritance — ``denied_by_envelope``
+        # also carries the agent's own name-keyed layers, which the child would have had
+        # either way, and attributing those to the operator's narrowing would be a
+        # confident wrong answer to "why is this denied?".
         lines.append(
-            f"  ↳ inherited this session's capability narrowing — {len(denied)} "
-            f"denied in the new session: {shown}" + (f", +{more} more" if more > 0 else "")
+            f"  ↳ inherited this session's capability narrowing; the new session's "
+            f"envelope denies {len(denied)}: {shown}"
+            + (f", +{more} more" if more > 0 else "")
         )
     if not tools_left:
         lines.append(
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -2639,13 +2639,61 @@ class AgentRegistry:
             attach_anchor(anchors)
         return session
 
+    def _persist_session_narrowing(self, name: str, sid: str, narrowing: dict) -> None:
+        """Write ``narrowing`` to ``(name, sid)``'s own ``config.yaml`` — the #2103-S1a
+        per-session capability layer, workspace-backed (P5).
+
+        The single WRITER of that file, so the two spawn entry points cannot drift from
+        each other or from its reader (``_load_per_session_capability_profile`` /
+        ``per_session_narrowing``, both keyed through ``_session_state_dir``). The
+        synthetic ``name`` key is what ``per_session_narrowing`` strips back off, so the
+        parent→child round-trip is exact.
+
+        Writing the file is not by itself enforcement: the live session's
+        ``_contextual_permission`` was resolved by the factory with ``sid=None`` and has
+        never seen this file. Each of the two callers re-resolves WITH the sid and
+        injects — ``spawn_session`` immediately after calling this, and
+        ``spawn_session_recorded`` after its own ``refresh_config_projections()`` (the
+        ordering there is measured, not stylistic; see the note at its ``spawn_session``
+        call). A caller that writes without injecting has persisted a narrowing nothing
+        enforces, which is the #2126 failure mode."""
+        import yaml
+        cfg_path = self._session_state_dir(name, sid) / "config.yaml"
+        cfg_path.parent.mkdir(parents=True, exist_ok=True)
+        cfg_path.write_text(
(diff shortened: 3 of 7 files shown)


tya5:
**[per-PR-coder]** — closing the scope note in the body's Verification table.

The local full-suite line quoted there (9849 passed / 36 skipped / 0 failed) was measured on `4060207`, i.e. one commit behind this PR's head. The stronger evidence has now landed: **CI ran on `72b21c1`, the exact head of this branch** (`gh run view 30677752736 --json headSha` → `72b21c102c4184cdf37b557de0d29491f9dccf6d`, matching `git rev-parse HEAD`), and every check passes — `pytest (Python 3.11)` and `pytest (Python 3.12)` both green, plus ruff, test-tier audit, docs build (strict), closing-intent contradiction check, and the four sandbox gates. 11/11.

So the identity gap the note flagged is closed by measurement rather than by argument, and on two interpreters instead of one. No action needed from a reviewer on that point.

Two things a reviewer may want to weigh first, both stated in the body but worth surfacing here:

1. **The behaviour change is not owner-approved.** An operator whose own session is narrowed now gets a narrowed new session. Built fully rather than softened, per the brief.
2. **The latent fail-open I found while measuring**, not fixed here and not filed: `reapply_visibility_override` re-resolves the envelope from base and SETs it, so when the base is unavailable (a session with no registry back-reference) it lands on ALLOW-ALL and discards whatever was injected before. That is what forced `spawn_session_recorded` to keep injecting after its own refresh. Contained today; happy to file it if lead wants it tracked rather than left in a code comment.

tya5:
**[per-PR-coder]** — the latent fail-open is filed: **#3593**.

★ Writing it up produced a measurement I did not have when I reported it, and it makes the finding worse than "contained via `refresh_config_projections`":

**The fail-open also fires INSIDE `AgentRegistry.spawn_session`.** `spawn_session` calls `load_persisted_toggles()` immediately after the #3561 re-resolve-and-inject, and that loader calls `reapply_visibility_override()` whenever the sid has a persisted `visibility.yaml` (`session.py:2372`) — no `refresh_config_projections()` anywhere on that path. Two arms differing **only** in the registry back-reference, same sid, same `config.yaml` narrowing (`tool_deny: [write_file]`), same persisted toggle (`tool: [read_file]`):

```
with_registry=True : live tool_deny = {'write_file', 'read_file'}   ← narrowing + toggle
with_registry=False: live tool_deny = {'read_file'}                 ← narrowing GONE
```

So it reaches every direct caller of the primitive — crash-recovery re-wake (the sites #3561 fixed) and this PR's `/session new` included — with an ordinary restart-state precondition. My earlier "contained today" was true of the mechanism I had measured and understated the one I had not.

**This PR is unaffected**, and I verified that rather than assuming it: the production session factory (`registry_bootstrap._session_factory`) passes `registry=` unconditionally, and every test in `tests/test_3562_*.py` uses a harness that does too — which is why the enforcement witness there is green and stays green. The issue enumerates all 5 `build_scoped_chat_session` call sites; 4 pass it unconditionally, `cli/commands/dogfood.py` passes it conditionally, and I did **not** measure whether a session is actually built inside that window, so #3593 says so rather than claiming containment holds everywhere.

Per your instruction the remedy is not proposed — neither in #3593's title nor its body — since the choice (refuse to re-resolve without a base / treat absent base as current value / require the collaborator at construction) has blast radius across #2285 / #3097 and should be decided rather than inherited from my first phrasing.

tya5:
**[lead-coder]** — CI 赤の帰属を記録します。**この PR 由来ではありません。**

```
pytest (Python 3.11)  FAIL   — 3.12 は緑
tests/test_textual_chat_gutter_toggle_3352.py:333
  AssertionError: the body was re-presented at [66, 78],
                  expected every re-present at the recovered width 78
```

## 帰属

- 本 PR の変更面は **権限エンベロープ（`AgentRegistry.spawn_session` の narrowing チャネル）** で、TUI の描画経路に到達しません。
- ★ **15 分前に PR #3587（`.md` 2ファイルのみの doc PR）が、同一テスト・同一の値 `{66, 78}` で落ちています。** 変更面が全く重ならない2つの PR が同じ失敗を出した ＝ どちらにも因果はありません。
- 両方とも **3.11 のみ赤 / 3.12 は緑**。

∴ 因果は成立しません。

## 追跡

★ **#3590** に切り出し済みです。★ 「flake」とは呼んでいません — 再現率を測ってから分類すべきで、まだ測っていないためです。3.11 のみという非対称を調査の入口として記録しています。

⚪ 関連して、実時間で境界を決めている assertion のもう1件を **#3594** に起票しました（`repo_knowledge_ingest` が `elapsed < 0.15` という壁時計の上限を assert している件）。★ 別テスト・別サブシステムですが同じ族なので、両 issue で **兄弟の掃き出し**を依頼しています。

## この PR の状態

★ **merge は owner 承認待ちで止めています**（可視の挙動変更のため）。CI の赤は解除条件に影響しません — 帰属が済んでおり、承認後に base 更新して再走させます。

tya5:
**[lead-coder]** — **owner 承認済。merge に進めます。★ ただし論拠が変わりました。**

## 承認された挙動

> 絞られたセッション（例: `/visibility` で `write_file` をオフにした）で `/session new` したとき、新しいセッションで `write_file` は使えるべきか
> **owner: 「使えるべきじゃないね」**

∴ **継承する**、で確定。実装は変更なし。

## 🔴 論拠の差し替え — セキュリティ論拠は取り下げます

本 PR と #3562 は当初、こう論じていました:

> `/session new` は **モデル出力から到達可能** である（#3561 の挙動テスト: agent ステップのモデル出力が `_interpolate_prompt` で次ステップの `prompt` に入り、その prompt が `_handle_user_message` に届き、`text.startswith("/")` で slash として実行される。到達ターンに LLM 呼び出しが無いことも測定済み）。∴ 絞りを継承しないのは権限のエスケープである。

★ **owner の裁定により、その到達経路そのものが *不具合* と確定しました:**

> **owner:「スラッシュコマンドの解釈は tui 側の想定だよ。inbox につまれたものはスラッシュコマンドとして解釈されない。されるんだとするとそれが不具合」**

∴ ★ **「モデルがスラッシュを起動できる」は仕様ではなく欠陥**であり、根を直せば到達経路は消えます。**本 PR をエスケープ対策として読まないでください。**

★ 本 PR が立つ理由は、**operator 自身が絞ったものは、そこから開いたセッションでも絞られたままであるべき**という方針判断です。それだけです。

## 設計は影響を受けません（むしろ単純化されます）

cross-identity で分岐を作らなかった理由は「呼び出し元と attach 先が食い違うのはモデル起点のときだけ ＝ まさにエスケープが起きているケース」でした。★ **根が直ればモデル起点が消える** ので、cross-identity は事実上発生しなくなり、`compose_narrowing_mappings` の一律適用は**より明白に正しく**なります。分岐を足す理由は依然としてありません。

## 根の方は別扱い

★ **inbox 経路（`session.py:5617` の `text.startswith("/")` → `_maybe_handle_slash`）がスラッシュを解釈している件は、別 issue で扱います。** 本 PR には含めません。

【実測】スラッシュの *ディスパッチ* は現在 **session 層のこの1箇所にしか存在しません**。TUI 側（`stream_client.py:175,184` / `completion.py:463`）は `/` かどうかを見ているだけで、ディスパッチしていません。∴ 修正は「TUI へ移す」ではなく「**TUI 側に作り、session 層から消す**」になります。スコープ（`--cui` を含めるか）は owner 確認中です。

## merge 手順

base を main に更新 → CI 再走 → 緑で merge。★ 直近の 3.11 の赤は #3590（無関係な2 PR が同一テスト・同一値で落ちた件）で、本 PR 由来ではありません（帰属済み: 上のコメント参照）。

tya5:
**[per-PR-coder]** — rebased onto `origin/main` (now includes #3596 / #3597); conflict resolved, body's rationale replaced. `mergeable` is back to `MERGEABLE`.

## What conflicted

**One hunk, in `tests/test_3546_*.py`** — the `session_cmd` site declaration, which both sides had rewritten. ★ **No source conflict at all**: this PR touches `slash/session.py` and `registry.py`, and #3596 split `session.py`, which this PR never edits.

I resolved it as a **union of two claims rather than a pick of one side**. `main`'s text says *who can reach the site* (model-output path closed; the entry stays because this gate counts where an envelope is BORN, not where a model can reach). Mine says *what the site now composes*. Both are true and neither implies the other, so the entry states both and its `measured_by` names all four legs — `_S3561`'s inverted pair and `_S3562`'s inheritance pair — with an explicit sentence about which pair measures which claim, so a later reader cannot take one as the other's evidence.

## The split-line hazard you named — checked, and not by inspection alone

The tests drive the invoker with `kind="user"`, the operator entry **above** the split (the only kind whose text is still slash-interpreted), and a session is born — measured, green. Nothing moved across the `_handle_user_message` / `_handle_inbox_text` boundary. The harness docstring now states that the kind claim is deliberate and which side of the split it sits on, so a future edit that moves it reads as a change rather than a detail.

## Both witness sets, one invocation

`tests/test_3562_*.py` + `tests/test_3561_*.py` + `tests/test_3546_*.py` → **22 passed**. That is my inheritance pair and display pair, plus #3596's `test_model_output_cannot_reach_slash_dispatch_and_spawns_nothing` and `test_an_operator_submitted_slash_command_still_spawns_a_session`.

## Rationale replacement — in the code, not only in the body

Per your instruction the body now leads with it, but prose resting on the retired argument was **rewritten wherever it lived**, since a body is not where a future reader looks: `slash/session.py`'s call-site comment, this PR's test module docstring, its cross-identity test docstring, and the site declaration. The PR now states plainly that it is **not** an escape fix and stands on the owner's ruling that operator intent persists across a spawn.

⚪ One drive-by: `test_3546_*.py`'s module docstring still asserted, in the present tense, that model output "observes it reach `/session new`" and that the operator-initiated guess "is false" — #3596 falsified that and did not update it. Fixed here because my resolution sits three lines below it; leaving the two adjacent and contradicting is the drift the doc rule exists to catch. Say the word if you would rather that landed separately.

## Gates, post-rebase

| gate | result |
|---|---|
| `ruff check src tests` | All checks passed! |
| `test_tier_audit.py --strict` (4 files) | 40 inspected · 0 errors · 0 warnings · exit 0 |
| `verify_module_docstrings.py` (2 src files) | OK |
| `pytest` (repo root, foreground, exit 0) | **9867 passed, 36 skipped, 0 failed** in 739s |

★ Identity of the measured tree: launched **after** `git rebase --continue`, `git status` clean at start and end, so it measured exactly the two commits now on the branch — and I confirmed `pwd` / `git worktree list` before measuring, committing and pushing, per the #3581 note. Neither known intermittent fired (#3590 and #3594 both passed), so nothing is attributed to them.

⚪ #3593 (the base-missing fail-open) stays out of this PR, as instructed, and remains open.

PR #3597, opened 2026-08-01
fix(runtime): webhook / MCP / cron stop claiming the "user" inbox kind — part of #3595

tya5:
**[per-PR-coder]** — step 1b of the #3595 arc, part of #3595. The three inbox producers that #3596 left claiming `"user"` with non-empty text stop claiming it, so **a chat webhook, an MCP/A2A peer and a cron fire can no longer execute any registered slash command**.

## What an external party could do before, and what they can do after

**Before.** Anyone able to POST to a configured Slack / LINE / generic webhook could send `/reset`, `/visibility`, `/plugin`, `/rewind`, `/clear-history`, `/model`, `/quit` — any of the 25 registered commands — and Reyn executed it. The message went in as `kind="user"`, i.e. as the claim *a human typed this at a first-party client*, and `Session._handle_user_message` acts on that claim by handing a `/`-prefixed line to `_maybe_handle_slash` **before any router turn**. The same was true of an MCP client's `send_to_agent` call and of an A2A delegation, and of a cron job whose message began with `/` (executed in an unattended session with no client present).

**After.** The same text reaches the model as content and is answered as an ordinary turn. No command runs. The three legs in the new test file each assert this on a real side effect — `/session new`'s spawn, observable from outside the session — not on a kind string.

## The product decision this PR records — not a dropped feature

> **owner (2026-08-01): 「現時点では slash に公開不要」** — no need to expose slash at present.

★ So this is the **final shape**, not a stopgap: there is no follow-up that builds a slash path for Slack / LINE / MCP. Recorded here, in `docs/gateway/authoring-guide.md` (a new "Slash commands are not exposed to gateways" section), and in `Session._handle_user_message`'s docstring — because **code cannot tell "forgotten" from "decided"**, and a later reader finding that Slack slash "doesn't work" would otherwise reasonably try to restore it. The ruling is also recorded on #3595 itself.

If it is ever revisited: the route is the ratified one — a **shared client-side slash layer** the transport drives — never `startswith("/")` returning to a transport, and never an allow-list of "safe" commands.

## The kinds, and the one-vs-two argument

Two new members of the union `_run_turn_body` already dispatches on (`user` / `agent_request` / `agent_response` / `pipeline_result` / `agent_step` / `hook`):

| kind | constant | producers |
|---|---|---|
| `external_message` | `runtime.transport.EXTERNAL_MESSAGE_INBOX_KIND` | `gateway.api.push_to_agent` (webhook) **and** `mcp.server.send_to_agent_impl` (MCP tool + A2A router) |
| `cron` | `runtime.cron.routing.CRON_INBOX_KIND` | the `reyn web` cron runner's inbox push |

**Why webhook and MCP share one.** The question the kind exists to answer is *who authored this text, for the purpose of deciding whether the OS may act on its FORM* — and a webhook peer and an MCP/A2A peer answer it identically: a counterparty outside this process, never the operator. I enumerated the in-tree consumers of the kind and each branches the same way on the two: turn dispatch (`_run_turn_body`), provenance (`_stamp_execution_context`), the hook-driven-turn valve (`run_one_iteration`), and the sent-queue projection (`queued_user_messages`). A consumer that needs the transport *itself* already has a strictly better source **on the envelope** — `sender` (`"slack:U456"`) or `reply_to` (`ExternalRef` / `McpRef`) — which names the individual peer, not merely its transport. ★ A distinction no consumer branches on, and that a richer field already carries at higher resolution, is a label, not a union member.

**Why cron does NOT share it.** Cron answers that same question differently: its text is **operator-authored**, in `.reyn/` job config, under the same file-permission trust as the workspace — not a counterparty's chat line. That is a distinction a trust decision (e.g. #3501's untrusted-content narrowing) would branch on. ★ And no envelope field can carry it: `sender` (`"cron:<job>"` vs `"slack:U456"`) is a **free-form attribution string the pushing plugin supplies**, and a discriminator a trust decision reads must not come from the side being classified. That last point is also why `push_to_agent`'s `kind` parameter is removed — see below.

`EXTERNAL_MESSAGE_INBOX_KIND` is declared in `runtime/transport.py` rather than at "the producer", because it has two producers and neither owns it; that module is already the shared vocabulary both import for the same class of thing (`ExternalRef` / `McpRef`). `CRON_INBOX_KIND` lives in `runtime/cron/routing.py`, mirroring `hooks.dispatcher.HOOK_INBOX_KIND` exactly (one subsystem, one producer).

## Two judgement calls beyond the literal brief — flagging both

**1. `push_to_agent`'s `kind` kwarg is removed** (it defaulted to `"user"` and was caller-overridable "so the contract can extend without API break"). Changing only the default leaves the defect one kwarg away from returning: a plugin could still pass `kind="user"` and make its external message execute commands. ★ A parameter that lets the caller choose whether it is the operator is not a parameter — it is the gate. Nothing in-tree ever passed it, and the A2A / MCP unification it was reserved for does not go through this module at all (those drive `send_to_agent_impl`). `tests/gateway/test_api.py::test_push_to_agent_kind_override` is replaced by `test_push_to_agent_kind_is_not_caller_selectable`, pinning the `TypeError` a plugin still passing it now gets — loud at the call site rather than silently ignored.

**2. `send_to_agent_impl` gains an `inbox_kind` parameter, and cron is in scope.** See the census below for why cron is a fourth producer with non-empty text, and the next section for why `send_to_agent_impl` needed a parameter rather than a blanket change.

## 🔴 The brief's biggest miss: `send_to_agent_impl` is not only MCP

The brief describes `mcp/server.py:241` as "an MCP server call". It has **four** producers:

| call site | who wrote the text | `inbox_kind` |
|---|---|---|
| `mcp/server.py` `send_to_agent` tool | an MCP peer, frequently another LLM | default (`external_message`) |
| `interfaces/web/routers/a2a.py` ×2 | an A2A peer | default (`external_message`) |
| `interfaces/cli/commands/chat.py::_run_once` | **the operator**, piping stdin at `reyn run-once` | explicit `"user"` |
| `interfaces/cli/commands/dogfood.py` ×2 | a human-authored scenario replayed **as** operator typing | explicit `"user"` |

A blanket change would have made `echo "/model x" | reyn run-once` stop executing the command — an operator-visible behaviour change this arc explicitly must not make — and would have made the dogfood instrument measure a path no operator uses. ★ **The default is the NON-operator kind on purpose**: a new caller that says nothing gets the kind that cannot execute a command, the same fail-safe direction `_stamp_execution_context` uses for turn origin.

## Census: every producer that puts text on an inbox

Enumerated by me from the `_put_inbox` call sites plus the two indirection layers that reach them (`Session.submit_user_text`, `MessageBus.request`), then each producer read individually — not taken from the brief or from #3596.

| producer | kind after this PR | truthful? |
|---|---|---|
| `session.submit_user_text` → `_put_inbox("user", …)` (funnelled into by `transport/in_process.py`, `agui/endpoint.py`, `repl/stream_client.py`, `textual_chat/app.py`) | `user` | ✅ a human typed it at a client |
| `session_api.run_agent_step` | `agent_step` | ✅ #3596 |
| `session.submit_agent_request` / `submit_agent_response` | `agent_request` / `agent_response` | ✅ |
| `session.submit_pipeline_result` | `pipeline_result` | ✅ kind truthful — but see finding 1 |
| `hooks/dispatcher.py:394` | `hook` | ✅ |
| `session._cross_session_hook_put` (session.py:4779) | caller-supplied; sole caller is the hook dispatcher | ✅ |
| 🟢 `gateway/api.py` `push_to_agent` (callers `sample_slack/webhook.py`, `sample_line/webhook.py`) | **`external_message`** | ✅ **fixed here** (was `user`) |
| 🟢 `mcp/server.py` `send_to_agent_impl` (MCP tool + A2A router legs) | **`external_message`** | ✅ **fixed here** (was `user` for all four producers) |
| 🟢 `interfaces/web/server.py` cron `_inbox_pusher` | **`cron`** | ✅ **fixed here** (was `user`) — ★ the fourth non-empty-text producer |
| 🟡 `registry.py:1254` (`_rewake_pipeline_runs` resume nudge) | `user` | ❌ a crash-recovery nudge, not a human. Text is `""` — see finding 2 |
| 🟡 `session_api.py:643` (`start_pipeline_run`) and `session_api.py:794` (`run_pipeline_attached`) | `user` | ❌ same shape, text `""` — finding 2 |

★ **No fifth non-empty-text producer exists.** Every remaining `"user"` claim is one of the three `""`-text nudges.

## Out of scope — confirmed still true, deliberately left

1. **`_handle_pipeline_result` still calls `_handle_user_message`** — confirmed at `session.py:5527`. A pipeline's result text is still slash-interpreted; it cannot fire only because `_format_result_text` prefixes `[pipeline] run …`, a property of the formatter, not a gate.
2. **The three `""`-text machine nudges** (`registry.py:1254`, `session_api.py:643`, `session_api.py:794`) still claim `user`. Empty text cannot start with `/`, so no exposure — the kind is still untruthful.
3. **`router_host_adapter.py:1969`** still says the MCP probe is "Called by `Session._handle_user_message`" — confirmed still false since #3475, still untouched.

## Consequences worth naming (all stricter, none wider)

- **Turn origin.** `_stamp_execution_context` maps every non-`"user"` kind to the stricter `"auto_improvement"`. A webhook / MCP / cron turn now stamps installs `provenance=auto_improvement` instead of `user_directed` — **by the existing fail-safe rule, not a new branch**. Correct on its face: a Slack peer did not direct an install.
- **Hook-driven-turn valve.** `run_one_iteration` resets the counter only on `kind == "user"`. These turns no longer reset it, so the valve fires marginally sooner. Same direction and same mechanism as #3596's `agent_step`.
- **Sent-queue projection.** `queued_user_messages()` filters `kind == "user"`, so an MCP message queued on the agent's `main` session no longer appears in an attached client's sent-queue view. That reads as the docstring already did — the sent-queue renders what *this operator* submitted from a client, and a peer's message was never that. Docstring updated in place.

## Strip-falsify — three producers, three legs, each anchor uniqueness-checked first

`grep -cF` → **1** for every anchor before stripping (`await session._put_inbox(EXTERNAL_MESSAGE_INBOX_KIND, envelope)`; `inbox_kind: str = EXTERNAL_MESSAGE_INBOX_KIND,`; `await session._put_inbox(CRON_INBOX_KIND, payload)`).

| state | result on `tests/test_3595_step1b_external_producer_slash_reachability.py` |
|---|---|
| fix applied | **4 passed** |
| webhook kind reverted to `"user"` | **1 failed, 3 passed** — `test_a_webhook_delivered_slash_command_is_read_not_executed` |
| restored | **4 passed** |
| MCP default reverted to `"user"` | **1 failed, 3 passed** — `test_an_mcp_delivered_slash_command_is_read_not_executed` |
| restored | **4 passed** |
| cron kind reverted to `"user"` | **1 failed, 3 passed** — `test_a_cron_fire_slash_message_is_read_not_executed` |
| restored | **4 passed** |

★ **Three different legs, one per producer** — no leg is standing in for another. ★ The operator control passed in **all seven** states, which is what makes it a control rather than a second copy of the same measurement.

## Tests

New: `tests/test_3595_step1b_external_producer_slash_reachability.py` — three absence legs + one control, on a real `AgentRegistry` with real `Session`s (harness shape borrowed from `test_3561_…`; the LLM stand-in is a real callable, not a mock). Each absence leg asserts **both** that no session was born **and** that the scripted LLM *was* consulted — the positive half, so "nothing happened at all" cannot explain the absence. The cron leg additionally asserts the runner returned `"ok"`, so a dispatch that never happened cannot pass as a command that never ran.

Updated, each because the PR falsified what it pinned (not to make it pass):
- `tests/gateway/test_api.py` — envelope-kind pins; the `kind=` override leg replaced by its inverse.
- `tests/gateway/sample_slack/test_webhook.py`, `tests/gateway/sample_line/test_webhook.py` — kind assertions.
- `tests/test_2608_h5_cron_webhook_hooks.py` — kind assertions **and** its three local `_inbox_pusher` doubles, which mirror the production pusher; a stand-in that keeps the old kind stops standing in for the producer.
- `tests/test_transport_ref.py::test_a2a_endpoint_uses_message_bus` — moved its seam from `_handle_user_message` to `_handle_inbox_text` (the shared body both the old and new routes end at) and updated the kind it looks for.
- `tests/test_chat_cli_flags.py`, `tests/test_teardown_completeness_2783.py` — the `_run_once` `send` doubles now mirror `send_to_agent_impl`'s signature including `inbox_kind`, and **assert it is `"user"`**, so the operator leg keeps a witness rather than merely tolerating the kwarg.

## Docs updated in the same PR (the mechanism changed, so the doc changed)

- `docs/gateway/authoring-guide.md` — the `push_to_agent` signature, and the new "Slash commands are not exposed to gateways" section carrying the owner ruling.
- `docs/reference/runtime/events.md` — `turn_started`'s `kind` enumeration; #3595 is what makes "tell human triggers from automated ones" true rather than approximate.
- `docs/concepts/tools-integrations/mcp.md`, `docs/concepts/multi-agent/a2a.md` — both stated the progress line reads `turn: user`; it now reads `turn: external_message`.
- In-code: `Session._handle_user_message` / `_stamp_execution_context` / `queued_user_messages`, `gateway/api.py`, `mcp/server.py`, `web/server.py`, `cron/routing.py`.

## ★ Corrections to the brief, measured

1. **`src/reyn/interfaces/gateway/api.py` does not exist.** The file is `src/reyn/gateway/api.py` (no `interfaces/`). Everything re-anchored by content.
2. **`send_to_agent_impl` has four producers, not one** — the brief called it "an MCP server call". Two of them (`reyn run-once`, dogfood) are operator surfaces; a blanket change would have silently altered them. This is why the change is a parameter with a fail-safe default rather than a one-line edit.
3. **The fourth producer is cron, and the brief's own framing hid it.** The brief said "two producers still claim `kind="user"` untruthfully with non-empty text" and listed cron only under #3596's 🟡 findings — but cron's text *is* non-empty (it is the job's `message`). Included, per the brief's own fourth-producer clause.
4. Line numbers: `push_to_agent`'s push is at `gateway/api.py:129` (brief said 120); `send_to_agent_impl`'s `bus.request` kind is at `mcp/server.py:263` (brief said 241). Nothing was applied by line number.
5. Not an error, just unstated: **removing `push_to_agent`'s `kind` parameter** is a judgement call past the literal scope. Flagged above rather than buried; happy to revert that one hunk if a reviewer wants the API kwarg preserved.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` on all 8 new/changed test files — 78 tests inspected, 0 errors, 0 warnings
- [x] `python -m pytest -q -n auto --timeout=120` from the repo root, foreground, summary line read directly — **9863 passed, 36 skipped** in 231 s. Neither known intermittent (#3590, #3594) fired.
- [x] `python scripts/verify_module_docstrings.py` on all 8 changed `src/` files — OK
- [x] Strip-falsify, three producers separately, each anchor `grep -cF`-confirmed unique first — numbers in the table above; three distinct legs failed
- [x] Measured-code identity confirmed at measurement time: an in-pytest probe printed `reyn.__file__` inside this agent's own worktree and read `EXTERNAL_MESSAGE_INBOX_KIND` from it. ⚠️ Worth knowing: a bare `python -c "import reyn"` from this worktree resolves the **shared checkout** (the borrowed venv's editable path) — pytest is correct because `pythonpath = ["src"]` is rootdir-relative. The worktree was re-checked with `pwd` + `git worktree list` immediately before measuring, committing and pushing.
- [x] `_put_inbox` census enumerated from the call sites myself, including the two indirection layers, and each producer read individually
- [x] Closing-intent: this PR declares `part of #3595` and no closing keyword anywhere. Rule-4 enumeration is not applicable — this is step 1b of a multi-step arc and the arc is not complete (step 2, moving the handler argument to `ClientTransport`, and step 3, deleting both session-side entry points, are still open).

### DIFF
--- a/docs/concepts/multi-agent/a2a.md
+++ b/docs/concepts/multi-agent/a2a.md
@@ -166,7 +166,7 @@ While the task runs, both the SSE stream and the webhook (below) carry an
 
 | Audit-event | `message` |
 |-------------|-----------|
-| `turn_started` | `turn: <inbox trigger kind>` — e.g. `turn: user` |
+| `turn_started` | `turn: <inbox trigger kind>` — `turn: external_message` for an A2A delegation (#3595 step 1b: a peer's message is not the operator's typing, so it does not claim the `user` kind) |
 | `llm_called` | `llm: <model>` |
 | `tool_returned` | `tool: <name>` |
 | `tool_failed` | `tool: <name> (failed)` |
--- a/docs/concepts/tools-integrations/mcp.md
+++ b/docs/concepts/tools-integrations/mcp.md
@@ -343,9 +343,19 @@ Two tools are registered:
 | Tool | Signature | What it does |
 |------|-----------|--------------|
 | `list_agents` | `()` | Returns a JSON array of `{name, role}` objects — one entry per agent declared in `reyn.yaml`. |
-| `send_to_agent` | `(agent_name, message)` | Submits one user-style message to the named agent and blocks (up to `--timeout` seconds) for the final reply text. Returns `{reply, partial, agent}`. If `partial=true`, the agent is still working; call again to receive more. |
+| `send_to_agent` | `(agent_name, message)` | Submits one message to the named agent as a conversational turn and blocks (up to `--timeout` seconds) for the final reply text. Returns `{reply, partial, agent}`. If `partial=true`, the agent is still working; call again to receive more. |
 
 Multi-turn continuity is preserved: each agent's `Session` keeps its `history.jsonl` between calls, so a conversation that starts in Claude Code can be resumed from `reyn chat` — or vice versa.
+
+#### `message` is content the model reads, never an operator command
+
+A `send_to_agent` message rides the `external_message` inbox kind: the OS's record that a counterparty outside this process wrote it, which is what it is. It is **not** the `user` kind, which means "a human typed this at a first-party client" and is the only kind Reyn interprets as an operator command line.
+
+The visible consequence: **Reyn's slash commands are not available over MCP.** A message reading `/reset`, `/model`, `/visibility` or any other registered command is read by the LLM as text; nothing is executed.
+
+This is a **decision, not an omission** (#3595 step 1b, owner ruling 2026-08-01: 「現時点では slash に公開不要」 — "no need to expose slash at present"). Until that arc, `send_to_agent` claimed the `user` kind, so any MCP client could run any registered slash command — including `reset`, `rewind` and `plugin` — by sending one as its message. If Reyn ever does expose slash to a peer, the route is a shared client-side slash layer, not a transport re-inspecting the message for a leading `/`.
+
+The same holds for the A2A router, which shares this implementation, and for chat-gateway plugins — see [the gateway authoring guide](../../gateway/authoring-guide.md#slash-commands-are-not-exposed-to-gateways).
 
 ### Progress notifications
 
--- a/docs/concepts/tools-integrations/mcp.md
+++ b/docs/concepts/tools-integrations/mcp.md
@@ -353,7 +353,7 @@ Multi-turn continuity is preserved: each agent's `Session` keeps its `history.js
 
 | Audit-event | Progress message |
 |-------------|------------------|
-| `turn_started` | `turn: <inbox trigger kind>` — e.g. `turn: user` |
+| `turn_started` | `turn: <inbox trigger kind>` — `turn: external_message` for a `send_to_agent` call (#3595 step 1b: an MCP peer's message is not the operator's typing, so it does not claim the `user` kind) |
 | `llm_called` | `llm: <model>` |
 | `tool_returned` | `tool: <name>` |
 | `tool_failed` | `tool: <name> (failed)` |
--- a/docs/gateway/authoring-guide.md
+++ b/docs/gateway/authoring-guide.md
@@ -127,9 +127,12 @@ Reyn versions; this API stays stable.
 ### Helpers
 
   push_to_agent(*, target_agent, text, sender, reply_to=None,
-                kind="user", extra_meta=None, registry=None)
+                extra_meta=None, registry=None)
     Deliver a message to a Reyn agent's inbox. Default for webhook
-    gateways.
+    gateways. The inbox kind is fixed at
+    ``EXTERNAL_MESSAGE_INBOX_KIND`` and is deliberately NOT a
+    parameter — see "Slash commands are not exposed to gateways"
+    below.
 
   list_agents(*, registry=None) -> list[str]
     All agent names known to the project (= sorted disk view).
@@ -172,6 +175,30 @@ await push_to_agent(
 **Do NOT call internal session methods directly.** ``Session.
 _put_inbox`` etc. are private API and may change between Reyn
 versions; ``reyn.gateway.api`` is the contract that stays stable.
+
+### Slash commands are not exposed to gateways
+
+A message pushed through `push_to_agent` is **content the model
+reads**, never a Reyn operator command. A gateway user sending
+`/reset`, `/visibility`, `/plugin` or any other registered slash
+command gets it read by the LLM as text; nothing is executed.
+
+This is a **decision, not an omission** (#3595 step 1b, owner ruling
+2026-08-01: 「現時点では slash に公開不要」 — "no need to expose slash
+at present"). Until this arc, a gateway push claimed the `"user"`
+inbox kind — the claim that a human typed the line at a first-party
+client — and `Session._handle_user_message` acts on that claim by
+handing a `/`-prefixed line to slash dispatch before any router
+turn. Anyone able to post to a webhook could therefore run any
+registered slash command. Gateway pushes now ride
+`EXTERNAL_MESSAGE_INBOX_KIND`, which `Session._run_turn_body` routes
+straight to the shared turn body: the dispatch is not skipped by a
+flag, it is not on the path at all.
+
+If Reyn ever does expose slash to a gateway, the route is the
+ratified one — a **shared client-side slash layer** the gateway
+drives — not `startswith("/")` re-introduced at a transport, and not
+an allow-list of "safe" commands.
 
 Reyn's session dispatch automatically:
 - Surfaces ``sender`` as a state_change history entry (= PR-A
--- a/docs/reference/runtime/events.md
+++ b/docs/reference/runtime/events.md
@@ -362,7 +362,7 @@ Required fields are declared in `src/reyn/core/events/event_schema.py`.
 | Kind | When | Key payload |
 |------|------|-------------|
 | `session_started`, `session_completed` | The session's resource scope opens / closes (alongside `chat_started` / `chat_stopped`). | `agent_name` |
-| `turn_started` | A trigger has been consumed from the inbox and is about to be dispatched. | `kind` — the inbox message kind that triggered this turn (`"user"`, `"agent_response"`, `"task_ready"`, `"hook"`, …), so a subscriber can tell human triggers from automated ones without parsing the payload; `chain_id`; `seq` |
+| `turn_started` | A trigger has been consumed from the inbox and is about to be dispatched. | `kind` — the inbox message kind that triggered this turn (`"user"`, `"agent_response"`, `"task_ready"`, `"hook"`, `"agent_step"`, `"external_message"`, `"cron"`, …), so a subscriber can tell human triggers from automated ones without parsing the payload. #3595 made that reading TRUE rather than approximate: a pipeline agent step's prompt (step 1), and a webhook / MCP / A2A / cron push (step 1b), used to arrive claiming `"user"`; each now carries its own member, so `kind == "user"` means an operator typed the line at a first-party client and nothing else does. `chain_id`; `seq` |
 | `turn_completed` | The router loop reached a terminal condition — router path only. | `chain_id` |
 | `turn_settled` | The turn is done, for EVERY turn kind including slash / intervention short-circuits that return before the router. The reliable clear signal for a working indicator started on `turn_started`. | `kind`; `chain_id` (may be absent for non-user triggers) |
 
--- a/src/reyn/gateway/api.py
+++ b/src/reyn/gateway/api.py
@@ -4,7 +4,7 @@ Plugins (= webhook handlers under ``reyn.gateway.*`` or external pip
 packages registered via the ``reyn.webhooks`` entry-point group)
 SHOULD use the helpers in this module to interact with Reyn agents,
 NOT call internal ``Session`` methods (= ``_put_inbox``,
-``_handle_user_message``) directly.
+``_handle_inbox_text``) directly.
 
 The contract here is intended to stay stable across Reyn minor
 versions; internal session APIs may change without notice.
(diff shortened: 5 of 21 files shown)


tya5:
**[per-PR-coder]** — doc sweep addressed, pushed as `dc1b10f7`. Three surfaces, plus one correction to the measurement behind the third request.

## 1–2. `mcp.md` — both fixed, and I read the whole section

- **`send_to_agent` "user-style message"** (tool table) → "one message … as a conversational turn". You were right that this is the *description becomes false* type, not the *add a name* type: "user-style" reads as "treated as the operator's line", which is the hole this PR closed.
- **`turn: user` example** → already `turn: external_message` in the first commit.

★ **I read the whole Role-2 section, not the two lines** — and also grepped `mcp.md` and `a2a.md` for `user` / `operator` / `slash` / `command` end-to-end. **Result: no third instance.** The other `user` hits are `ask_user` (the intervention mechanism, unrelated), the elicitation paragraph, and `operator` in the permissions sections — all still true. Line 327 ("submit messages … as if they were just another MCP tool") and line 367 ("only two operations: list agents and send a message") make no origin claim. ⚪ So docs-maintainer's two were, on my reading, the complete set in these two files — but that is my grep + read of two files, not a repo-wide guarantee.

I also **added** a subsection (`message` is content the model reads, never an operator command) carrying the visible consequence and the owner ruling 「現時点では slash に公開不要」, mirroring the gateway guide, and cross-linking it. Reason: an MCP user hitting "`/reset` did nothing" needs to find a *decision*, not a gap — ★ code cannot tell "forgotten" from "decided", and that is exactly what would invite someone to re-open the hole.

## 3. The `message_bus.py` usage example — ★ my judgement: **not** a client-input seam, so I fixed it

It is the **MCP seam**. The module docstring's first line is "request/reply correlation for **MCP and A2A transports**", and the example builds `McpRef(request_id=…)`. So it was showing `McpRef` paired with `kind="user"` — ★ **precisely the pairing this PR removes.** It was not merely a copy-prone example; after commit 1 it was factually wrong.

Fixed to `kind=EXTERNAL_MESSAGE_INBOX_KIND`, with a line stating why a usage example is load-bearing here (the next author of an injection path copies it, so an example naming a kind untruthfully *reproduces* the defect).

★ **And a fourth site the sweep did not name, same class**: `MessageBus.request`'s own parameter doc read ``kind: Inbox message kind (e.g. ``"user"``)``. Its actual callers are `run_agent_step` (`agent_step`), `run_pipeline_attached` (the `""`-nudge), and `send_to_agent_impl` (`external_message`) — ★ `"user"` was the one kind **none** of them should name, standing as *the* example. Replaced with what the parameter means (a claim about **who authored the text**), the full member list, and the honest note that it is still a free-form `str`, so nothing but the caller's honesty enforces it.

## 🔴 Correction to the architect's measurement

> ★ そのうち *人間が打った* ものは 0 件（agent step / 別の内部投入 / MCP 経由 / docstring の使用例）

★ **This is wrong, and in a way that matters for the【2】design.** The human-typed site exists and is one of the writers: **`session.py:3048`, inside `Session.submit_user_text`** — `await self._put_inbox("user", {...})`. It writes the bare literal `"user"`, not `kind="user"`, which is why a `kind="user"`-shaped grep misses it. Every client composer (TUI, CUI, AG-UI, in-process transport) funnels into it.

Full enumeration of **inbox-injection** sites naming `"user"` on this branch:

| site | truthful? |
|---|---|
| `session.py:3048` `submit_user_text` → `_put_inbox("user", …)` | ✅ **the client-input seam** — a human typed it |
| `session_api.py:796` `run_pipeline_attached` → `bus.request(kind="user")` | ❌ the `""`-text nudge (out of scope, reported as finding 2) |
| `chat.py:59` (`reyn run-once`) / `dogfood.py:651,661` — `inbox_kind="user"` | ✅ added by this PR; operator at a CLI, and the harness standing in for one |
| ~~`message_bus.py:50` docstring~~ | ✅ fixed here |

★ **Good news for【2】**: the architect's target state — "`kind="user"` を書くサイトが1箇所だけ" — is **closer than the measurement suggested**. There is exactly **one** genuine client-input seam and it already exists (`submit_user_text`). The remaining non-seam writers are the three `""`-nudges plus the two operator-CLI sites, and the CLI ones are arguably a *second* legitimate client. ⚪ An AST gate would need to decide whether `reyn run-once`'s stdin counts as a client-input seam — I would say yes (it is a human at a first-party CLI), but that is【2】's call, not mine.

## Gates re-run on the doc commit

- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — **exit 0** (I ran the real docs gate because I added a cross-file anchor link; the residual `INFO` lines are the pre-existing `.ja` anchor noise, and `gateway/authoring-guide.md`'s not-in-nav listing predates this PR)
- [x] `python -m pytest -q -n auto --timeout=120` from the repo root, foreground — **9863 passed, 36 skipped** in 181 s
- [x] `python scripts/verify_module_docstrings.py src/reyn/runtime/message_bus.py` — OK
- [x] Worktree re-verified with `pwd` + branch immediately before measuring, committing and pushing

★ **All 11 CI checks were green on the previous head** (`398d4176a`), including both `pytest (3.11)` / `pytest (3.12)` at 5m27s / 5m50s — that had not landed yet when I last reported. CI is re-running on `dc1b10f7`.

PR #3598, opened 2026-08-01
refactor(runtime): close the inbox kind vocabulary as TurnOrigin + gate who may claim CLIENT_INPUT — part of #3595 (S1–S3)

tya5:
**[per-PR-coder]** — S1–S3 of the #3595【2】 foundation. S4 (handler argument → `ClientTransport`) and S5 (delete both session-side slash entry points) are **not** in this PR.

part of #3595

---

## S1 — the inbox `kind` becomes a closed type, and the colliding word is resolved

`src/reyn/runtime/turn_origin.py` — `class TurnOrigin(StrEnum)`, 9 members. The four scattered `*_INBOX_KIND` module constants (`transport` / `session_api` / `cron.routing` / `hooks.dispatcher`) are **deleted**, not aliased; their rationale prose is relocated onto the members it explains.

```python
CLIENT_INPUT = "user"          # ★ not spelled USER
AGENT_REQUEST / AGENT_RESPONSE / PIPELINE_RESULT / AGENT_STEP
EXTERNAL_MESSAGE / CRON / HOOK / PIPELINE_NUDGE
```

**Member name vs wire value are separate decisions, and I made them separately.**

- The **name** is `CLIENT_INPUT`. That is what closes the census problem: the outbox display vocabulary (`runtime/outbox.py`, construction-validated, 15 words) and the inbox provenance vocabulary intersected in exactly one word, and the symbol sets now intersect in zero. `outbox.py` gained a cross-reference naming the sibling namespace so a reader of either lands on the other.
- The **value** stays `"user"`. It is written into the WAL/snapshot inbox (`SnapshotJournal.append_inbox`), read back **verbatim** by `Session.restore_state` (`self.inbox.put_nowait((msg["kind"], msg["payload"]))` — plain JSON strings), and published as the `turn_started`/`turn_settled` audit-event `kind`, which `.reyn/events` has consumers outside reyn for. Changing it would either break restore of a snapshot written by an older build — a crash-recovery regression, a cross-cutting-band member — or require a legacy-value mapping this PR has no independent reason to add.
- `StrEnum`, **not** the repo's usual `(str, Enum)`: they differ in `__str__`, and this value is interpolated (`f"turn: {data.get('kind')}"` in `progress_lifecycle.format_progress_message`, reached with the in-memory payload). Under `(str, Enum)` that operator-visible line would have read `TurnOrigin.CLIENT_INPUT`. Reason recorded in the module docstring.

**Migration insurance**: producer signatures are retyped `kind: TurnOrigin` (`Session._put_inbox`, `Session._cross_session_hook_put`, `MessageBus.request`, `SnapshotJournal.append_inbox`, `send_to_agent_impl(inbox_kind=)`). Consumers (`_run_turn_body`, `_stamp_execution_context`) stay `str` **on purpose and with a docstring saying why**: what arrives there came off `self.inbox`, which `restore_state` repopulates from JSON, so a `TurnOrigin` annotation would be a claim the method cannot keep.

## S2 — who may claim `CLIENT_INPUT`, enumerated with reasons

The registry lives in `tests/test_3595_client_input_provenance_gate.py::_CLIENT_INPUT_SITES` — 7 entries, each `role` (`asserts` / `reads`) + `reason` + `occurrences` + `measured_by`. Same shape as the narrowing arc's `_SITE_PARENT_LAYERS`.

**Asserts (3 keys / 4 occurrences):** `Session.submit_user_text` (the client seam), `cli/commands/chat.py::_run_once` (`reyn run-once` stdin), `cli/commands/dogfood.py::_build_live_runner.runner_fn` ×2 (operator harness).
**Reads (4):** `_run_turn_body`, `_stamp_execution_context`, `run_one_iteration` (loop valve), `queued_user_messages`.

### The `""` nudge: `PIPELINE_NUDGE`, and the argument

`session_api.run_pipeline_attached` starts an attached run by putting **one empty-text message** on the driver-session's inbox and pumping it. It claimed `CLIENT_INPUT` because that was the member the dispatch table would run a turn for — a statement about the dispatch table, not about who wrote the message. **Nobody wrote it.**

★ It was the only one of the five claimants the slash defect could **never** expose: `"".startswith("/")` is false. That is why four censuses walked past it — nothing it did was wrong. It is also why moving it is not "closing a hole": the rename is what stops one honest member from carrying two meanings.

**Behavioural consequences, measured rather than assumed:**

| what reads the kind | before | after | verdict |
|---|---|---|---|
| `_run_turn_body` route | `_handle_user_message` → `startswith("/")` → `_handle_inbox_text` | `_handle_inbox_text` | **byte-equivalent for this producer** — the only intervening step is the `/` test, and the text is always `""` |
| `queued_user_messages` | the nudge appeared in the operator's sent queue | it does not | **corrected**, same sentence #3597 already wrote for `external_message`/`cron` |
| `_stamp_execution_context` | `user_directed` | `auto_improvement` | **strictly the fail-safe direction**, and the same landing #3596 gave the agent-step path by the same rule (no new branch) |
| loop valve (`_hook_driven_turns = 0`) | reset on the nudge | not reset | a driver-session is spawned per run; there is no hook chain to reset |

The third row is the only one that is a change and not a correction of a display artefact. It moves the `provenance` stamp on installs performed inside an attached pipeline-driver session from `user_directed` to `auto_improvement`. Full suite is green including `test_pipeline_is6_attached.py`. **I am flagging it explicitly rather than filing it under "no behaviour change"** — if a reviewer wants it preserved, the fix is a two-member set at `_stamp_execution_context`, and I would want the argument for why an attached pipeline run is operator-directed before writing it.

## S3 — the gate

`tests/test_3595_client_input_provenance_gate.py`, 6 tests.

**The walk matches the SYMBOL and nothing else** — no callee filter, no keyword-name filter, no argument-position filter. Every one of those questions is what a previous census asked first, and each cost a different site. `test_walk_sees_every_syntactic_form_a_claim_can_take` drives it against a fixture containing kwarg / bare positional / dict literal / via-variable and requires all four.

**Vacuity guard**: `test_extraction_is_not_vacuous` fails LOUD on an empty extraction.

**★ The positive control is wired in permanently, not applied once.** `test_no_declared_site_is_stale` requires every registry entry to still be **found**. A walk that regresses (gains a filter, loses a form) makes declared sites vanish → RED. `chat.py` — the site architect's 5th miss dropped — is in that registry.

**The other leg, stated as a separate mechanism**: `test_inbox_injection_seams_type_their_kind_parameter` reads the **live** signatures of the five inbox seams and fails if any `kind` parameter is widened back to `str`. The walk counts symbol references and structurally cannot see a producer that regressed to the bare string `"user"`; what stops that call being written is that it is a static type error, and that property lives entirely in those annotations. Two mechanisms, two tests, because one test claiming both would claim more than either measures.

## The doc-drift architect flagged

`mcp/server.py`'s *"Both pass `inbox_kind="user"`"* said **two** while there were **three** call sites. Fixed by removing the count from prose entirely: the docstring now points at the gate's allowlist as the single place the count lives.

Also fixed in this PR (same rule, same change): `docs/reference/runtime/events.md`'s `turn_started` row enumerated inbox kinds inline including **`"task_ready"`**, which the task system's abolition (#2839) left behind — the row now points at `TurnOrigin` and names that as the reason not to list them there. `event_schema.py`'s parallel comment and `session.py::_drain_to_wake`'s docstring carried the same dead word; both fixed.

---

## Measurements

**Census — value-side, no keyword filter, every `"user"` string constant in `src/reyn` classified by hand (76 before → 68 after):**

| classification | count |
|---|---|
| inbox provenance (string literal) | **0** ← was 5 |
| the `TurnOrigin.CLIENT_INPUT` declaration | 1 |
| outbox display kind | 11 |
| history `role` | ~30 |
| sender-transport label / memory scope / MCP config scope | the rest |

Symbol-side extraction: **8 sites, 4 asserts + 4 reads.** Re-run after rebasing onto #3586: unchanged.

**Positive controls applied, and what they caught:**
1. *"Does `chat.py:59` show up in my extractor?"* — **yes**, first run. (This is architect's 5th miss; a callee-name allowlist would have returned zero here.)
2. *"Does the bare-positional `_put_inbox("user", …)` at `session.py:3048` show up?"* — **yes**. (This is architect's 3rd miss; a `kind=` kwarg grep returns zero here.)
3. *"Does pytest resolve MY worktree's `reyn`?"* — a throwaway probe test asserted `reyn.runtime.turn_origin.__file__`. ★ **It caught a real discrepancy**: bare `python -c "import reyn"` resolves the **shared** checkout (`.../lead-coder/src/reyn`); pytest resolves the **worktree**. Every measurement in this PR is pytest-path. Confirms the brief's warning is live on this machine.

**Strip-falsify — each anchor `grep -c` → 1 before stripping:**

| # | strip | result |
|---|---|---|
| 1 | added `_STRIP_FALSIFY = TurnOrigin.CLIENT_INPUT` inside an already-declared function | **RED** — `occurrences` 1 vs 2 |
| 1b | moved it into a NEW function | **RED** — `test_every_client_input_site_is_declared` |
| 2 | broke the walk with a filter (the 5th-miss shape) | **3 RED**: vacuity guard, form test, **and the stale-site positive control**. ⚠️ `test_every_client_input_site_is_declared` stayed **GREEN** on the empty walk — which is the whole reason the other three exist |
| 3 | ★ removed the vacuity guard, walk still broken | **the gate PASSES on an empty extraction** (4 passed). This is the failure mode that makes a gate silently useless; it is now measured, not asserted |
| 4 | widened `_put_inbox`'s `kind` back to `str` | **RED** — `test_inbox_injection_seams_type_their_kind_parameter` |
| 5 | deleted the `PIPELINE_NUDGE` branch from `_run_turn_body` | **RED** — `test_a_pipeline_nudge_still_runs_one_turn` |

All restored; green re-confirmed after each.

**Crash-recovery leg** (`test_a_kind_restored_from_a_snapshot_as_a_plain_string_still_dispatches`): drives the real journal, round-trips the snapshot inbox through `json`, asserts the kinds come back as plain `str`, that a restored `"user"` still equals `TurnOrigin.CLIENT_INPUT`, and that `_run_turn_body` still runs a turn on it. This is the load-bearing half of "the wire values did not change" — under fresh values, or under a non-`StrEnum`, every message queued before a restart would reach no branch, **silently**.

---

## Where this brief / the spec turned out to be inaccurate when measured

1. 🔴 **"of the 5 inbox sites naming that kind, only one is human-typed … the word `user` already misrepresents provenance on the inbox side."** — **Not what I measure.** Architect's own S2 table marks `chat.py:59` ✅ and `dogfood.py:651,661` ✅ as legitimately `CLIENT_INPUT`, with reasons. So **4 of 5** were entitled to it; exactly **one** (the `""` nudge) was not. And the word did not "already misrepresent provenance" — #3596/#3597 had already fixed the four producers that made it lie, and `events.md` on current main says so in its own wording. ★ **This matters because it was the brief's stated *independent* justification for the rename.** The rename stands on the **collision** argument alone, which is sufficient, and which is what I built on.
2. **"`session_api.py:794 (or :796)`"** — it is **:796** pre-change. Minor.
3. **The `""` nudge's own defect shape.** The spec says the nudge should get a different origin because *its text is empty and it is not a human*. True, but the sharper fact — the one that explains why four censuses missed it — is that it was **structurally immune to the defect the arc was chasing**: empty text can never start with `/`. A census hunting slash reachability had no reason to stop there. That is the reusable lesson and it is not in the spec.
4. **S1's "split the type … `class DisplayKind(Enum)`" sketch — I did not build the `DisplayKind` half, deliberately.** The outbox vocabulary is **already** a closed, construction-validated set (`outbox.VOCABULARY`) with its own non-circular gate (`test_outbox_vocabulary.py` + `test_agui_profile_completeness.py`). Adding a parallel `DisplayKind` enum without migrating its ~15 call sites — which the brief explicitly forbids sweeping — would create an unused second declaration of an existing vocabulary: the opposite of an SSoT. The split S1 asks for is achieved by the inbox getting its own named type; I added a cross-reference in `outbox.py` instead.
5. **S3's "however S1 makes this enumeration unnecessary — just count references to the enum member."** — **True for the post-S1 world, and it hides one leg.** The wire value is still `"user"`, so a regressed producer writing the bare string would inject a `CLIENT_INPUT`-kinded message and appear in **no** symbol census. The spec treats "one spelling" as total; it is total only under the type annotations. That is why the gate has a second, separately-stated arm reading the live signatures — and why I did **not** let one test claim both.
6. ⚪ **`_put_inbox` runtime `isinstance` check — considered and rejected.** It would make the residual leg structurally impossible rather than statically discouraged, but it is a raise at a load-bearing seam (a behaviour change) and would force edits to ~20 test call sites including `test_3561`'s witness. Recorded as a deliberate non-choice, not an oversight; the annotation arm is the cheaper equivalent under this PR's no-behaviour-change constraint.

---

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` on all 10 changed test files — 90 inspected, 0 errors, 0 warnings
- [x] `python scripts/verify_module_docstrings.py` on all 16 changed src files — OK
- [x] `pytest` from repo root, foreground, summary line read — see the run comment below
- [x] Strip-falsify the gate (unlisted claimant, both inside an existing function and in a new one) — RED, restored, green
- [x] Strip-falsify the **vacuity guard** — confirmed the gate passes on an empty extraction with it removed
- [x] Strip-falsify the seam-type arm (`kind: TurnOrigin` → `str`) — RED, restored, green
- [x] Strip-falsify the `PIPELINE_NUDGE` dispatch branch — RED, restored, green
- [x] Each strip anchor confirmed unique (`grep -c` → 1) before stripping
- [x] Positive control: `chat.py` (kwarg form) and `session.py:3048` (bare positional) both appear in my own extractor
- [x] Positive control: pytest resolves the worktree's `reyn`, not the shared checkout — probe test, verified
- [x] #3596 / #3597 witnesses green **with assertion content unchanged** — the only edits to those files are symbol renames inside docstrings and failure messages
- [x] Rebased onto `origin/main` @ `7439d33f5` (#3586) and re-measured: census unchanged, `test_3546` / `test_3561` green
- [x] Doc-drift fixed in THIS PR: `mcp/server.py` prose count, `events.md` `turn_started` row (`task_ready`), `event_schema.py` comment, `session.py::_drain_to_wake` docstring, `gateway/authoring-guide.md`, `pipeline-dsl.md`
- [x] (skipped — no visual surface) Visual / TUI check: this PR changes no rendering path


tya5:
**[per-PR-coder]** — full-suite run, on the exact pushed tree.

```
$ python -m pytest -q --timeout=600 -p no:randomly     # repo root, foreground
9877 passed, 36 skipped, 62 warnings in 727.14s (0:12:07)
```

Tree: `d424f0010` (pre-author-amend; identical content to the pushed head), rebased onto `origin/main` @ `7439d33f5` (#3586).

★ The two known intermittents the brief warned about (`test_textual_chat_gutter_toggle_3352.py:333` / #3590, `test_fp0066_p3b_repo_knowledge_ingest.py:198` / #3594) **did not fire** in this run. Nothing to attribute.

⚪ Interpreter identity, verified at measurement time rather than assumed: a bare `python -c "import reyn"` in this agent worktree resolves the **shared** checkout (`.../lead-coder/src/reyn`); pytest resolves the **worktree** (`.../worktrees/agent-a9ebee3433b7a5416/src/reyn`). Confirmed with a throwaway probe test asserting `reyn.runtime.turn_origin.__file__` — which doubles as a positive control, since that module does not exist on main. Every number in this PR is pytest-path.

The other three CI gates, separately:

```
$ ruff check src tests                                              → All checks passed
$ python scripts/test_tier_audit.py --strict <10 changed test files> → 90 inspected · 0 errors · 0 warnings
$ python scripts/verify_module_docstrings.py <16 changed src files>  → OK
```

PR #3606, opened 2026-08-01
refactor(slash): Session interprets no string — dispatch moves client-side (#3595 S5)

tya5:
**[per-PR-coder]** — #3595 **S5**, the arc's last step: `Session`'s two slash entry points are deleted, and interpreting text as a command becomes CLIENT work.

Closes #3595

## What went

```
session.py  _handle_user_message         → the `text.startswith("/")` short-circuit (deleted, with the method)
session.py  maybe_deliver_answer_command → #3327's `/answer` pre-queue fast path    (deleted)
session.py  _maybe_handle_slash          → the dispatch itself                      (moved, client-side)
```

Both entry points in one change, per the architect's completion condition: deleting only the first would leave `/answer` with a surviving string-sniffing entry — the sibling-site shape this arc hit repeatedly.

`_handle_user_message` is **deleted rather than kept as a forwarder**: with the slash block gone it had nothing `CLIENT_INPUT`-specific left to do, so `_run_turn_body`'s `CLIENT_INPUT` arm and `_handle_pipeline_result` both call `_handle_inbox_text`. That also closes the finding step 1 recorded in its own docstring — a pipeline's OS-framed result text was slash-interpreted, protected only by the formatter never emitting a leading `/`.

## Where dispatch went, and why there

`src/reyn/interfaces/slash/dispatch.py` — one shared layer, two halves on opposite sides of the transport:

- **`maybe_dispatch_slash(transport, text)`** — the CLIENT half. Parses one line, resolves it against the process-local `REGISTRY` (imported code, which every client has), displays its own output (extra-lines note, bare-`/` catalog, unknown-command suggestion) through `put_display`. Called by **both** clients before they submit a turn: `stream_client.route_input_line` (CUI, local *and* `--connect`) and `TextualChatApp._submit` (TUI).
- **`execute_slash_command(ctx, name, args)`** — the EXECUTOR half, wherever the session is. It never sees the operator's raw text, only a name already resolved.

Execution rides the client seam: **`ClientTransport.run_slash_command(name, args)`** (non-abstract default, the `deliver_pending_answer` / `cancel_queued` precedent).

- `InProcessTransport` → runs it against the attached session, handing the handler **its own transport** — S4's stated target.
- `AgUiTransport` → POSTs `{"type": "slash_command", "name", "args"}`; the AG-UI endpoint runs it server-side. ★ **This is the leg that keeps S5 from being a feature removal** — see "what the brief got wrong" below.

[...]

PR #3608, opened 2026-08-01
refactor(memory): hand the router a memory capability, not file primitives — part of #3607

tya5:
**[per-PR-coder]** — `part of #3607` (the `op_context_inputs` half stays open; architect's firm §4 splits them).

## The defect, and what it is not

`RouterHostAdapter` was handed `file_read` / `file_write` / `file_delete` / `file_regenerate_index`. Their only consumers were three `RouterLoop` privates that assembled the memory operations out of them — **with the memory domain's rules living in the loop**: the FP-0050/#1822 threat scan that *rejects* a poisoned write before it persists, YAML frontmatter construction and stripping, listing-index regeneration, knowledge-index ingest/de-index.

Those four params share a consumer set, so #3482's bundling measurement would happily have made them a sixth bundle — and the 57 lines of rule would still be in `RouterLoop`. A Parameter Object moves parameters to a common root; only a **Facade Service** hides the aggregate *behaviour*. The adapter now receives **one capability** — the session's `MemoryService` — and the router calls `host.memory.remember/.forget/.read_body`.

## What actually changed vs the brief's estimate

| | brief's estimate | measured |
| --- | --- | --- |
| adapter ctor params | 4 → 1 | **59 → 55** — `memory` was ALREADY a param (for the two path delegates). The four file params left; nothing was added. `scripts/measure_router_host_adapter_consumers.py`: 0 exact-match clusters before and after, shelf unchanged (`journal`). |
| RouterLoop privates | 3 (≈57 lines) | **3, −204 lines net** in `router_loop.py` (the privates + `_strip_frontmatter` + 6 host-Protocol members) |
| RouterState fields | 3 removed | **3 removed, 0 added** — `RouterCallerState.memory_service` **already existed and was wired by nobody**. That is the field this arc was missing; the wiring had gone to three `*_fn` callables instead. |
| threat-scan rule moves | yes | `MemoryService.scan_for_block` + its call in `remember`. `RouterHostAdapter.scan_for_block` **deleted** — it had exactly one caller. |
| — | not in the brief | adapter also loses `memory_path` / `memory_dir`: dead once the loop stops assembling paths. Also **`MemoryService.remember/forget/read_body` had NO production caller** — three near-duplicates of the router privates. Unifying them is most of this diff. |

New: `MemoryKnowledgeSync` (in `memory_service.py`) binds the memory layer to the index subsystem, so `MemoryService` holds one collaborator instead of `IndexCoordinator` + `workspace_root` + `OpContext` as three materials — the same "don't hand it the parts" rule this PR is about, applied one level down. Session builds it with `op_context_fn=lambda: self._router_host.make_router_op_context()` — the *same* factory `_remember` reached through, resolved at call time (`_build_memory` runs pre-waist, and an OpContext is per-turn state).

## Behaviour: three deltas, all named

LLM-visible behaviour of `remember` / `forget` / `read_body` on the **success** path is unchanged (same result keys, same frontmatter bytes, same rules in the same order). Unifying two divergent implementations forced three decisions; the production (router) semantics won the LLM-visible surface, the richer `MemoryService` semantics won everywhere else:

1. **A failed write no longer reports success.** `_remember` discarded `file_write`'s result and returned `{"saved": slug}` even when the write failed. `MemoryService` already checked, and keeps checking. Failure path only.
2. **`memory_saved` / `memory_deleted` now fire on the router path.** They never did — only the non-router fallback in `tools/memory.py` emitted them, so the same operation left an audit trail through one door and not the other. Both kinds are already in `AUDIT_EVENT_KINDS` and `events.md`; no vocabulary change.
3. **`read_memory_body`'s truncation signal rides as sibling keys, not appended prose.** `_read_memory_body` went through `RouterHostAdapter.file_read`, which flattened the dict to a string and appended `"... <note>"` to the content (#3193). That method is gone with the other three; `read_body` forwards `truncated` / `note` / `next_offset` as keys, which is the shape #3193 established for dict-returning consumers — and it keeps op-runtime prose out of a memory body the LLM narrates verbatim (the G12 attractor `strip_frontmatter` exists for). The two adapter-level #3193 tests are replaced by two that drive the real `_handle_read_memory_body` handler with a real `RouterCallerState` + real `MemoryService` — i.e. the surface the LLM actually reads, which the adapter tests were standing in for.

## Threat scan — where it went and how it is witnessed

`MemoryService.scan_for_block`, called first in `remember`. Witnesses are on the **side-effect** side, not on the flag:

- `test_memory_service_invariants.py::test_poisoned_remember_is_rejected_and_nothing_is_persisted` — real `ThreatScanConfig`, real files: **no file on disk, no MEMORY.md**, plus the deny envelope and the `threat_block` audit-event.
- `..._legit_remember_is_not_blocked_by_an_enabled_scan` — falsify side, same scan enabled.
- `..._knowledge_index_follows_a_write_but_never_a_blocked_one` — a blocked write must not be ingested into the knowledge index either (searchable-anyway leak).
- `test_memory_write_block_1822.py` still drives it through a full `RouterLoop.run()` turn — and its `_BlockHost`, which **hand-mirrored** `scan_for_block`'s scan+emit sequence, is deleted: `FakeRouterHost` now carries a real `ThreatScanConfig` into a real `MemoryService`. Same for `test_2649`'s `_BlockHost`.

## Strip-falsify (cp-backup restore, never `git checkout --`; every anchor `count == 1` verified before editing, both numbers below)

| # | strip | anchor count | result |
| --- | --- | --- | --- |
| 1 | `hit = self.scan_for_block(...)` → `hit = None` in `MemoryService.remember` | **1** | **RED ×4** — `test_poisoned_remember_is_rejected_and_nothing_is_persisted`, `test_knowledge_index_follows_a_write_but_never_a_blocked_one`, `test_poisoned_memory_write_blocked_no_persist` (loop-level), `test_remember_threat_blocked_renders_and_classifies_via_new_path`. Restored → **16 passed** |
| 2 | `memory_service=getattr(self.host, "memory", None)` → `None` (unwire the capability) | **1** | **RED ×4** — both `test_router_loop.py` memory tests + both `test_memory_write_block_1822.py` tests. Restored → **47 passed** |
| 3 | re-add an `async def file_write` to `RouterHostAdapter` (re-open the layering hole) | **1** | **RED ×1** — `test_adapter_exposes_the_memory_capability_itself`. The gate asserts absence of each primitive by name, so a re-added one fails even though `memory` is also present. Restored → **12 passed** |
| 4 | `body = strip_frontmatter(result["content"])` → `body = result["content"]` | **1** | **RED ×1** — `test_remember_then_read_body_roundtrip` (the G12 rule). Restored → **14 passed** |

No `STRIP-FALSIFY` residue in `src/` or `tests/` (grep: 3 hits, all pre-existing prose in unrelated files).

## Docs falsified by this PR, fixed here

- `docs/reference/runtime/session-construction.md` — Family 8b rewritten (capability, not "persistence adapter"; the one deferred lambda named); Family 6a gains the #3607 paragraph on *why these four left instead of becoming a sixth bundle*.
- `docs/concepts/architecture/llm-invocation-surfaces.md` + `.ja.md` — the "per-tool callable bridges" list named all three deleted `*_fn` fields (the `.ja.md` asserted the same thing, so it is fixed on its own merits, not as a mirror obligation).
- `docs/deep-dives/proposals/0050-content-threat-scan.md` — BP1's seam row and the S4a bullet named `RouterLoop._remember`.
- `src/reyn/data/index/knowledge_ingest.py` module docstring — its "TWO call paths" paragraph named `RouterLoop._remember`/`_forget`.
- `src/reyn/tools/memory.py` module docstring — its "design-revisit finding" described the router path as duplicating `router_loop.py`.

**Deliberately not edited**: ADR `0026-unified-tool-registry.md` and the dated journal insight both name the `*_fn` bridges — they are immutable historical records of the migration that introduced them, not claims about current code.

## ⚠️ File overlap with open PR #3606

Two files overlap: `src/reyn/runtime/session.py` and `tests/test_router_loop_chatsession.py`. **Measured, not assumed**: #3606's `session.py` hunks are at ~3069 and ~5396–5540; mine are the import block, the adapter construction (~3905) and `_build_memory` (~4290). In the test file, theirs are at 132/154, mine at ~241. Disjoint — nothing was silently resolved. Flagged for whoever merges second.

## Test plan

- [x] `pytest -q -n auto --timeout=120` from repo root — **9892 passed, 36 skipped** (both known intermittents, #3590 / #3594, passed)
- [x] `ruff check src tests scripts` — All checks passed
- [x] `python scripts/test_tier_audit.py --strict` on all 20 changed test files — 142 inspected, 0 warnings, 0 Tier-4 errors
- [x] `python scripts/verify_module_docstrings.py` on all 8 changed src files — OK
- [x] Manual: `scripts/measure_router_host_adapter_consumers.py` before (`git stash`) and after — **59 → 55 params, 0 clusters both sides**, shelf unchanged
- [x] Manual: the four strip-falsify arms above, each RED on the named arm, restored by `cp` and re-verified green
- [x] Enumerated `part of #3607` scope: this PR does NOT close the issue — the `op_context_inputs` half is unstarted, so no closing keyword is used


tya5:
**[lead-coder]** — **判断を仰がれた2件、裁定します。両方このままで正しいです。merge します。**

## (2) `memory_saved` / `memory_deleted` が router 経路でも発火する

★ **これは挙動変更というより、欠落の解消です。**

従来は **非 router の fallback でだけ**発火していました。★ ∴ **同じ操作が、通った経路によって監査イベントを出したり出さなかったりしていた**ことになります。★ **監査の穴**です。

⚪ 両 kind とも**既に宣言済み**なので、閉じた語彙への追加もありません。

★ 本日このアークで繰り返し扱った形（**片方の経路にだけ在る**）の一例で、揃える方向が正しいです。

## (3) 切り詰め通知が兄弟キーに乗る

★ **こちらも正しいです。** 従来は `RouterHostAdapter.file_read` が dict を文字列に平坦化していたので、★ **切り詰めの事実を *本文に散文で追記* するしかありませんでした。**

★ ∴ **モデルは「本文」と「本文について OS が言ったこと」を区別できませんでした。** 兄弟キーに乗れば区別できます。★ **reyn が typed を prefer する理由そのもの**です。

⚪ 平坦化を行っていた `file_read` が他のプリミティブと一緒に消えたので、★ **制約が外れた結果として自然に直った**形です。制約を回避するために何かを足したのではありません。

---

## ★ 私の brief が4箇所で誤っていました（記録）

1. 「引数 4→1」→ ★ **実際は 59→55**。`memory` は既にコンストラクタ引数だった
2. ★ **`RouterCallerState.memory_service` は既に存在し、誰も配線していなかった** — このアークで欠けていたフィールドは最初からそこに在った
3. ★★ **`MemoryService.remember` / `forget` / `read_body` も既に存在し、本番の呼び出し元がゼロだった** — router の非公開3本とほぼ重複、frontmatter・エラー処理・監査イベントが食い違っていた。★ **その突き合わせが diff の大半**
4. adapter は `memory_path` / `memory_dir` / `scan_for_block` も失う

★★ ∴ **これは「能力を作る」作業ではなく「二重に存在していた能力を1つにする」作業でした。** ★ 私は前者として brief を書きました。

## ★ 特に評価したい点

**手で写した `scan_for_block` の複製2つを削除したこと。** ★ 複製されたセキュリティ判定は、★ **本物と食い違っても誰も気づきません**（テストは複製の方を通るので緑のまま）。★ 実 config を実 `MemoryService` に通す形へ変えたのが正しいです。

**脅威スキャンの witness が「ディスクにファイルが無い」で取られていること**、そして ★ **falsify の双子で「有効なだけでなく発火している」まで証明したこと。** ★ 層を移すときにセキュリティ規則が静かに消える、が本 PR の最大のリスクでした。

**#3606 との hunk 重複を、黙って解決せず実測して報告したこと。**

PR #3617, opened 2026-08-01
fix(interfaces): replace hand-rolled clipboard dispatch with pyperclip

tya5:
**[per-PR-coder]** — Deletes `_clipboard.py`'s hand-rolled per-OS dispatch table (`pbcopy`/`wl-copy`/`xclip`/`xsel`/`clip` via `shutil.which` + `subprocess`) and replaces it with a thin `pyperclip` wrapper. part of #3616 (① only — ② mouse-selection/OSC 52 copy is explicitly held by the owner and untouched).

## Why (the class, not the hole)

`_clipboard.py`'s Windows arm piped raw UTF-8 bytes to `clip.exe`, which decodes stdin under the OS's OEM code page (CP932 for Japanese, etc.) instead of UTF-8 — garbling any non-ASCII text. The list of tools it probed is exactly what `pyperclip` already does, and `pyperclip`'s Windows backend writes `CF_UNICODETEXT` directly via `ctypes`, never touching `clip.exe` or a code page. Rejected fixes (per the issue): swapping `clip` for PowerShell `Set-Clipboard`, or `chcp 65001` — both are patches onto the reinvention, not a removal of it.

## What changed

- `src/reyn/interfaces/repl/_clipboard.py` — same two public functions (`copy_to_clipboard`, `copy_to_clipboard_async`), now thin wrappers over `pyperclip.copy()`. `copy_to_clipboard` never raises: `PyperclipException` (pyperclip's documented "no backend" signal) and any other backend exception are caught and mapped to `False`. The async variant keeps the thread-executor off-load (pyperclip's macOS/Linux backends still shell out synchronously via `subprocess.Popen(...).communicate()`), so a `/copy` or copy-mode yank still can't freeze the TUI's event loop.
- `pyproject.toml` — added `pyperclip>=1.9` as a direct dependency.
- Two call sites updated for the value-shape change (see below): `src/reyn/interfaces/repl/_copy_sentinel.py:73` (`/copy`'s status line) and `src/reyn/interfaces/inline/textual_chat/app.py` (`_write_clipboard` — copy mode's sink — and `on_flow_view_activated` — Enter/Space cursor-copy).
- `tests/test_clipboard_pyperclip_3616.py` (new) — Tier 1 contract tests for `_clipboard.py` itself (none existed before).
- `tests/test_textual_chat_copy_rewind_3362.py` — updated one assertion + several docstrings/comments that described the old `shutil.which`/tool-list mechanics.

## `tool_label` decision + consumer audit

`pyperclip` has no public, stable "which backend won" API — only `is_available()` (bool) and a private module-level function reference (`pyperclip.copy`). Inventing a label would mean guessing at pyperclip's internals, so the old `(ok, tool_label)` contract is now a plain `ok: bool` on both `copy_to_clipboard` and `copy_to_clipboard_async`.

Audited every consumer of the old tuple (`grep -rn "tool_label\|copy_to_clipboard" src`), three total:
1. `_copy_sentinel.py:73` — was `f"copied reply to clipboard ({tool})"`; now `"copied reply to clipboard"` (success) / a real failure message (was previously an empty-string status on failure from the old contract's `("", False)` — this is not a status-message regression, it's the fix of a latent one).
2. `app.py`'s `on_flow_view_activated` (Enter/Space cursor-copy) — was `ok, status = ...` then `status if not ok else "copied to clipboard"`; the "failure" branch was actually showing the *empty* tool_label string on failure (`copy_to_clipboard` returned `("", False)` when no tool was found), i.e. a blank status row on failure. Now: `"clipboard copy failed"` — a real message where there was a silent blank before.
3. `app.py`'s `_write_clipboard` (copy mode sink) — already discarded the label (`ok, _tool = ...`); now just `ok = ...`.

## Vacuity falsification (per policy)

`tests/test_clipboard_pyperclip_3616.py::test_no_backend_available_returns_false_not_an_exception` forces pyperclip's real "no clipboard mechanism" state via its own public `pyperclip.set_clipboard("no")` API (not a mock — a real, documented pyperclip call), then asserts `copy_to_clipboard(...) is False`.

Falsification measured directly: temporarily changed `_clipboard.py`'s `except Exception: return False` to `return True` and re-ran the test — it went RED (`AssertionError: assert True is False`). Reverted; confirmed GREEN again. This proves the test actually exercises the mapped-failure branch rather than passing vacuously.

## Windows/git-bash backend investigation (requested mid-review, not in the original brief)

The owner asked directly whether ① fixes the reported mojibake, given two open unknowns. I read the **installed pyperclip 1.11.0 source** (`determine_clipboard()`, `pyperclip/__init__.py:504-545`) rather than trusting its README, and traced every branch:

- `os.name == 'nt' or platform.system() == 'Windows'` → `init_windows_clipboard()` — `ctypes` `CF_UNICODETEXT`, Unicode-safe, never touches `clip.exe`. This is what a **native Windows Python build** (python.org installer, Microsoft Store, pyenv-win — the ordinary way to `pip install` a package with binary deps like `numpy`, which reyn depends on and which has no MSYS-native PyPI wheels) resolves to, **regardless of which shell invokes it** — `os.name` is the interpreter's compile-time platform, not the invoking shell, so running under git-bash's terminal does not change this branch.
- Genuine WSL (`platform.system() == 'Linux'` AND `/proc/version` contains `"microsoft"`) → `init_wsl_clipboard()` — this branch **does** shell out to `clip.exe`, but encodes `text.encode('utf-16le')`, not UTF-8. UTF-16LE is the encoding `clip.exe` actually handles correctly for Unicode — this is not the bug class in the issue (which was reyn's own code piping UTF-8).
- I found **no branch** in `determine_clipboard()` that pipes UTF-8 to `clip.exe`. The only `clip.exe` call site in the whole installed file is the UTF-16LE WSL one above.
- **The one scenario I cannot rule out**: if the owner's `reyn` install runs under MSYS2's *own* native `python` package (not a mingw-w64/native-Windows Python — `os.name == 'posix'`, `platform.system()` reporting something like `MSYS_NT-10.0-...`), none of the cygwin/Windows/WSL/macOS branches match, and it falls through to the generic Linux section (`WAYLAND_DISPLAY`/`DISPLAY`/`xclip`/`xsel`/`qt` checks), which on a bare git-bash terminal typically finds nothing and lands on `init_no_clipboard()`. In that case `copy_to_clipboard` returns `False` — an **honest reported failure** (a "clipboard copy failed" status line), not silent mojibake. This is a real remaining unknown: **I cannot verify which Python build the owner's `reyn` runs under without their machine.** If it's the MSYS-native build, ① turns "garbled text pasted" into "no text pasted, failure reported" — an improvement (fails safe, no silent bytes) but not "Japanese copies successfully."

**Bottom line — say exactly this, not more**: ① is a correct structural fix (the reinvention is removed, and reading pyperclip's actual source shows no in-process path reproduces the UTF-8→clip.exe bug). Whether the owner's *actual* Japanese-copy attempt succeeds still depends on which Python build resolves clipboard backend selection on their machine, and **that is unverified — the acceptance condition remains the owner testing on their own Windows/git-bash hardware**, as stated in the issue.

## Verification

- `python scripts/test_tier_audit.py --strict tests/test_clipboard_pyperclip_3616.py tests/test_textual_chat_copy_rewind_3362.py tests/test_copy_mode_3507.py` — 19/19 OK, 0 Tier-4 violations.
- `pytest` (full suite, from repo root) — **9912 passed, 36 skipped, 0 failed** (712s).
- `python scripts/verify_module_docstrings.py <changed src files>` — clean.
- Full clipboard/copy test suite re-run against a rebased `main` after the investigation — 25/25 passed.

## Test plan

- [x] `test_tier_audit.py --strict` clean on all changed/added test files
- [x] Full `pytest` from repo root — 9912 passed, 36 skipped, 0 failed
- [x] `verify_module_docstrings.py` clean on changed src files
- [x] Vacuity falsification: deliberately broke the no-backend→False mapping, confirmed RED, reverted, confirmed GREEN
- [x] `tool_label` consumer audit: grepped `tool_label`/`copy_to_clipboard` repo-wide, all 3 call sites updated
- [x] (skipped — no Windows/git-bash hardware available to this session or the owner-facing verification session) real Japanese-text copy on the owner's company Windows/git-bash machine — **this is the actual acceptance condition per the issue and remains open**


tya5:
## Follow-up: `uv lock --check` failure + Linux CI test failures (both fixed)

**`uv lock --check`** — `pyproject.toml` gained `pyperclip>=1.9` but `uv.lock` wasn't regenerated. Fixed by running `uv lock` and committing the result.

**Linux CI `pytest` failures (11 failed)** — every clipboard-witness test fixture faked a binary named `pbcopy` on `PATH`. That worked under the OLD `_clipboard.py`, which searched `shutil.which` over a fixed tool-name list *regardless of host OS* — a same-named fake was found on any platform. It does NOT work under pyperclip: `determine_clipboard()` is platform-gated and only ever tries `pbcopy` on Darwin, so on the Linux runner the fake was invisible and every assertion on the recorded clipboard content failed (`None`/`False` instead of the copied text).

Fix: pin pyperclip's backend explicitly via pyperclip's own public `set_clipboard("xclip")` API (documented for exactly this purpose — not a monkeypatch of pyperclip internals, not a mock), then fake a binary named `xclip` instead of `pbcopy`. `init_xclip_clipboard()`'s `Popen(['xclip', ...])` call has no OS check of its own once the backend is pinned, so the identical fake works on macOS and Linux. No test's real-clipboard dependency was removed and no assertion was weakened — same `ok is True` / `wrote is True` / `clipboard() == <expected text>` checks throughout; only the fake binary's name and the explicit backend pin changed.

Touched 4 files with a pre-existing duplicate of this fixture (not introduced by this PR) plus my own new test file:

| File | What changed | Why (not a loosening) | Assertion strength |
|---|---|---|---|
| `test_textual_chat_copy_rewind_3362.py` | `clipboard` fixture: `pbcopy`→`xclip` fake + `pyperclip.set_clipboard("xclip")` pin + save/restore | The SUT's dispatch mechanism itself changed (① is exactly "stop doing OS-agnostic name search, delegate to pyperclip's platform-gated dispatch") — the fixture was built for the old mechanism's semantics, not wrong on its own terms, just no longer matched by the new SUT | Unchanged — same `clipboard() == "..."` assertions |
| `test_conversation_cursor_3476.py` | Same fixture pattern (pre-existing duplicate), identical fix | Same reason | Unchanged |
| `test_lazy_history_paging_3476.py` | Same fixture pattern (pre-existing duplicate), identical fix | Same reason | Unchanged |
| `test_copy_mode_3507.py` | Inline fake in `test_copy_mode_yank_writes_through_reyns_local_clipboard`, identical fix, wrapped in try/finally | Same reason | Unchanged — same `wrote is True` / sink-content assertion |
| `test_clipboard_pyperclip_3616.py` (new, mine) | Replaced ad hoc `pbcopy` setup in 2 tests with a shared `fake_xclip` fixture using the same pin | Same reason, plus avoiding a 5th duplicate of the pattern | Unchanged |

No skip/xfail was added anywhere, no mock/patch of reyn's or pyperclip's code, no test's success path was routed around a real clipboard tool — each test still requires a real subprocess (the fake `xclip` script) to actually receive and record the piped text.

Issue #3630, opened 2026-08-02
present is unusable by a model without operator knowledge — three barriers on the zero-knowledge path


tya5:
**[tui-coder]** — From an owner live check of `present` in `../user`. Expected a rich rendering, got a raw JSON string, and the model had no way to notice.

## What the operator saw

```
present(data_inline={"blueprint": [{"text": …
  ⎿ ok
     { "kind": "present", "status": "ok", "ok": true, "mode": "default",
       "bindings_resolved": 1,
       "bindings_dropped": [{"path": "/blueprint", "reason": "type_mismatch"}],
       "rows": 0, "all_bindings_missed": false }

blueprint  [{"text": "これは present ツールの基本的な text コンポーネントの表示です。", "component": "text"}, …]
```

The last line is the entire user-facing output: a `keyvalue` row whose value is the JSON dump of a container.

## What actually happened

Reproduced against the real synthesizer with the owner's exact data (`default_viewer_blueprint` → `validate_blueprint` → `resolve_bindings`, surface `inline-cui`):

```
rows=0  resolved=1  dropped=[{path:/blueprint, reason:type_mismatch}]  all_missed=False
nodes = [{"component":"keyvalue","rows":[{"label":"blueprint","value":"[{\"text\": …}]"}]}]
```

So the value was **coerced and displayed**, not dropped. This is documented behaviour in `binding.py`: "Type mismatch → coerce (… a container bound into a text slot → its JSON form) and record `{path, reason: type_mismatch}`."

## The defect: every field of the ack points away from what happened

The ack is the model's only feedback channel, and all four signals mislead:

| Field | Says | Actually |
|---|---|---|
| `bindings_dropped … type_mismatch` | dropped | rendered, as a JSON string |
| `rows: 0` | nothing rendered | a 4-row keyvalue card rendered — `rows` only counts *table* rows (`binding.py:207`, incremented solely in the table-rows path) |
| `all_bindings_missed: false` | bindings hit | true, but it measures pointer resolution, not whether anything legible reached the user |
| `ok: true` | success | honest by its own definition ("SOME presentation reached the user") and useless here |

Nothing reports the one fact that matters: **a container was flattened to a JSON string, so the user is looking at a data dump rather than a rendering.** That is why the model never corrected itself — from where it sits, it succeeded.

[...]

PR #3632, opened 2026-08-02
fix: declare present's blueprint as an array so a component sequence is expressible

tya5:
**[tui-coder]** — Barrier 2 of #3630. Part of `#3630`.

## The defect

`validate_blueprint` accepts **"a single component node OR a list of nodes"**, and a list is the only way to express a *sequence* of components. The schema published to the model declared:

```python
"blueprint": {"type": "object"}
```

So the list form was unrepresentable. A model wanting a heading followed by a table had no way to say so — whatever `PresentIROp` would have accepted.

This is not a restriction on an otherwise-working call. It makes a correct shape **unsayable**.

## Why it looked like model error

Observed live: asked to show data, a model emitted

```json
present(data_inline={"blueprint": [{"component": "text", …}, {"component": "markdown", …}]})
```

Read as misuse, that is a model putting a top-level parameter in the wrong place. Read against the schema, `data_inline` was the only other object-typed slot on the call — the only place the value could go. The model was routing around a declaration, not misreading a description.

## Why `array` rather than both

`array` costs nothing in expressiveness: a single node is a one-element list, while no multi-node sequence can be written as an object. The object-only declaration was strictly narrower than the op, in the one direction that mattered.

Declaring both is not available — the Gemini-safe rules enforced by `test_router_tools.py` forbid `anyOf`, and there is no typeless-parameter precedent (all 47 declared params across every tool carry exactly one type). The op keeps accepting a bare component object, so nothing that works today stops working.

## Why no existing test could have caught it

All 26 test files touching the present op construct the op directly in Python and pass a list — which never travels through the schema the model is bound by. Stated generally in the test docstring, because it is not specific to present:

> a test that constructs an op directly cannot witness whether the model can REACH that op — mechanism correctness and reachability need separate witnesses.

The new test is the reachability half. It validates a two-node blueprint against the **real published schema** (`build_tools()` output, via `jsonschema`) rather than asserting a type literal, so it keeps holding if the schema is later expressed some other way. Two nodes specifically: a single-node example would have passed against the old declaration too and witnessed nothing.

## Not in this PR

- **Barrier 1** (the default viewer's dict branch does not descend, so nested data renders as a JSON dump) — next, and the higher-priority one for #3630's acceptance condition.
- **Barrier 3** (the ack does not report that it flattened a container) — parked pending an agreed ack shape.
- `data_inline` carries the same narrowing (declared `object`, op accepts `Any`), so a top-level list — the shape that renders as a **table** — is also unreachable inline. Deliberately left for barrier 1: once the dict branch descends, `{"items": [...]}` renders as a table and the narrowing stops costing anything. Fixing it here would mean picking one of `object`/`array` under the same one-type constraint and breaking the other.
- **No real-model witness yet.** The owner authorised `gpt-5.6-luna` for this; 4 chat runs plus 2 direct probes all failed before reaching the model — the proxy returns `ChatgptException - Unknown items in responses API response: []` for that model group, with and without tools. Reported to the owner; the A/B (baseline → fixed, same zero-knowledge prompt) is set up and will run once the route is live.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_present_blueprint_representable_3630.py` — 0 warnings, 0 errors
- [x] `python scripts/verify_module_docstrings.py` on all three changed src/test files — OK
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — built
- [x] `test_router_tools.py` (the Gemini-safe forbidden-key gate) + the present/registry suites — 102 passed
- [x] Falsification: reverting the declaration to `object` turns both reachability assertions red while the two op-side ones stay green — the exact asymmetry the defect had.
- [ ] (skipped — blocked) Real-model witness: proxy route for `gpt-5.6-luna` is down, see above.


tya5:
**[tui-coder]** — CI red addressed. Attribution confirmed by measurement, and re-keying surfaced a pre-existing defect in the fixture itself.

## Attribution

Measured, not inferred: restoring **both** changed source files from `origin/main` makes the test pass; either one of mine present makes it fail. The report is specific —

```
tools: DIFFERS
    present: schema differs (4ab741e16061 -> 08b2323ddf3a)
model / messages / tool_choice: MATCHES
```

Note the type is not the whole cause: reverting only `blueprint`'s type left the hash moved, because the parameter DESCRIPTION is part of the hashed `tools=` payload too. Both halves of this PR contribute.

This test's own docstring already records two prior instances (#3520, #3556), so the procedure was documented — `REYN_FP0063_ARC_WITNESS_GENERATE=1`. No live model involved: the responses are authored and disclosed as such.

## The part that was not documented, and cost a false green

Regenerating **in place appends**. The generator records into whatever is already in the file, so the first attempt produced:

| | before | after in-place regen | old keys surviving |
|---|---|---|---|
| `turn1_install` | 6 keys | 8 | **6 of 6** |
| `turn2_ingest_query` | 9 keys | 12 | **9 of 9** |

That fixture passed — and it also **passed against the pre-change `present` schema**, which I checked precisely because the review asked for proof of re-recording. Green either way, witnessing neither. Without that check it would have looked fixed.

Deleting the files first and regenerating gives a clean re-key: 0 old keys surviving, `turn1` 2 keys, `turn2` 3.

## The committed fixture had been accumulating

The counts explain themselves once you look at the entries: `turn1` held the **same two logical calls three times over, under six distinct keys** — three schema generations layered up by the earlier in-place regenerations. Any one of the three would have satisfied a run, so the fixture on main has been matching several past schema versions at once.

So the drop from 6/9 keys to 2/3 is the arc's actual call count, not lost coverage. Both facts, plus the delete-first requirement, are now in the test docstring where the two prior re-keys are recorded — the next tool-schema change will hit exactly this.

## Verification (the review's two requirements)

- **Proof of re-record**: new fixtures against the OLD schema raise `MissingFixture`, attributing the drift to `present: schema differs (08b2323ddf3a -> 4ab741e16061)`. Against the new schema: passes.
- **Scope**: both files in `tests/fixtures/llm/fp0063_arc_witness/`. Nothing else in the repo replays a `tools=` payload containing `present` — these two are the whole blast radius of a present-schema change.

Gates re-run after the re-key: ruff clean, tier audit 0/0, docstring gate OK, `test_fp0063_arc_witness` + `test_present_blueprint_representable_3630` + `test_router_tools` — 22 passed.

## Worth its own issue?

"A tool schema change silently invalidates replay fixtures, and the documented regeneration procedure appends rather than replaces" is not specific to `present` — it applies to any tool schema edit, and the append behaviour has already left three generations layered in a committed fixture. Happy to file it; say the word and I will, rather than expanding this PR.

PR #3635, opened 2026-08-02
fix: let present's default viewer descend into a nested object

tya5:
**[tui-coder]** — Barrier 1 of #3630, the one that decides its acceptance condition. Part of `#3630`.

## Why this is the zero-knowledge path

`present`'s description steers the model here explicitly:

> Simple path — just pass `data_ref` … and it is shown with **a sensible default view; you do not need to design anything.**
> `blueprint`: … **Prefer the simple default (omit both view and blueprint)** unless you specifically need a custom layout.

So the default viewer is what an operator gets without knowing present exists — the owner's stated goal. And it was not sensible for the shape its own description names (`a query result, a list`, `after a tool returns a large result`): tool results nest.

## The defect

The dict branch read keys at depth 0 and emitted one `keyvalue` card. A `keyvalue` row's `value` is a text slot, so a dict or list bound into one renders as its JSON form. The list branch, meanwhile, descended a level to build table columns.

Identical rows therefore rendered differently depending on the wrapper, not the data:

| Passed as | Rendered |
|---|---|
| `[{id, name}, …]` | table, columns id/name |
| `{"items": [{id, name}, …]}` | `items -> "[{\"id\": 1, …}]"` |

## The change

An object now yields a `keyvalue` over its **scalar** keys, then a `text` label and its own view per **nested** value, recursing to a depth bound (4). Below the bound a value goes back into a `keyvalue` row — the same floor a table's cells have always had, kept so the synthesized view's size stays a property of the data's shape rather than its depth.

The labels are the key names the `keyvalue` rows already carried, so nothing appears that the previous view did not also show.

Measured on the payload that prompted #3630 — `{code, message, data: {items: [...]}, status}`:

| | before | after |
|---|---|---|
| rendering | one card, `data` flattened to `{"items": […]}` | keyvalue + labels + a **2-row table** over id/name |
| `bindings_dropped` | `[{/data, type_mismatch}]` | `[]` |

## Verification

Everything is asserted through the real `default_viewer_blueprint` → `validate_blueprint` → `resolve_bindings` path on a registered surface, so what is checked is the rendering a user would see rather than the blueprint the synthesizer emitted.

Falsified against the previous dict branch: the three behavioural tests go red, and the two guarding cases that were already correct (a flat object stays ONE card; a top-level list is unchanged) stay green — the asymmetry the defect actually had.

The depth bound is pinned too, by asserting the drop it produces rather than avoiding it: an honest floor beats a hidden one.

## Still open on #3630

- **Barrier 3** (the ack reports `dropped` for a value it in fact flattened and displayed, and `rows` counts only table rows) — parked pending an agreed ack shape. This PR reduces how often it fires, since the common nested shapes no longer flatten, but it does not remove the case.
- `data_inline` is still declared `object` while the op accepts `Any`, so a top-level list is unreachable inline. That now costs nothing: `{"items": [...]}` renders as a table, which is what wrapping it was previously punished for.
- **No real-model witness yet.** The proxy route for `gpt-5.6-luna` returns `ChatgptException - Unknown items in responses API response: []` (with and without tools, reyn or direct), so the baseline/after A/B the owner authorised cannot run. Set up and waiting on the route.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/test_present_default_viewer_nested_3630.py` — 0 warnings, 0 errors
- [x] `python scripts/verify_module_docstrings.py` on the changed test + src file — OK
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — built
- [x] `pytest -k "present or fallback or fp0063"` — 403 passed
- [x] Falsification against the previous dict branch: 3 red / 2 green as described above
- [x] Docs: the fallback-ladder description said `dict -> keyvalue`, which this falsifies — corrected in `concepts/present.md`, `reference/present.md`, and the two `.ja.md` files that assert the same thing (their own claims, not a mirror obligation)
- [ ] (skipped — blocked) Real-model witness: proxy route down, see above.


tya5:
**[tui-coder]** — Two notes before review.

## Rebased, and what that overwrote

This branch was cut from `main` **before** #3632 landed, and both PRs edit `docs/reference/runtime/present.md` — #3632 rewrote the blueprint example to the list form, this one adds the object-descent paragraph. Left alone the merge could have carried an older copy of that file.

Rebased onto the merged `main` and verified both survive: the example is `"blueprint": [ … ]` and the ladder section carries the new object paragraph.

While I was rebasing, a `Merge branch 'main'` commit appeared on this branch remotely, and my force-push replaced it. Checked rather than assumed: `git diff b503a1fe7 HEAD` is **only** my own paragraph reorder (8+/8-, moving the object paragraph below the content-type one so step 3's two rungs read in order), and that commit carried no changes of its own beyond the merge. Nothing was lost — but flagging it since the overwrite was mine.

Gates re-run after the rebase: ruff clean, `mkdocs build --strict` built, `pytest -k "present or fallback or fp0063"` 407 passed.

## This does not make #3630's acceptance condition met

Correcting my own framing in the PR title line above: calling barrier 1 "the one that decides its acceptance condition" reads as a stronger claim than I can support. Barriers 1 and 2 remove two things that demonstrably blocked a model — measured on the owner's payload and against the published schema — but the acceptance condition the owner set is that **a model, given only the tool schema and description, reaches a good rendering**. Only a real-model witness can say that, and it has not run: the proxy route for `gpt-5.6-luna` returns `ChatgptException - Unknown items in responses API response: []`, with and without tools, through reyn and directly.

So the honest status is "two known blockers removed, unverified end to end". The A/B is set up — isolated workspace, a prompt naming neither present nor any component, baseline and after runs on the same prompt — and will run when the route is live.

PR #3641, opened 2026-08-02
feat(tui): fold a pipeline run into one row with live progress

tya5:
**[tui-coder]** — Owner report after a `rag_ingest` run: *「pipeline 実行したら関連するステップが個別に flow entry になってる。これ単一 flow entry にしてリッチな進捗表示にできない？」*

## What was happening

Traced through the owner's live logs. The pipeline runs in a **spawned driver session**, and its events land on that session's own EventLog:

```
.reyn/events/agents/default/sessions/4bae93a3/chat/2026-08/…
  pipeline_step_started    x15
  pipeline_step_completed  x15     run_id / step_index / total_steps: 15 / step_kind
```

`lifecycle_forwarder` bridges them to the parent as transient `status` frames — deliberately, its own comment says a many-step pipeline "would spam permanent `system` markers otherwise" — and it puts the **`run_id` in meta**.

The key was there. Nothing consumed it: `_ingest_frame` appends anything it does not specially coalesce, so 30 frames became 30 entries and the conversation they were progress *for* scrolled away behind them.

## What this does

They coalesce by `run_id`, using the idiom tool results already use:

- one row per run, updated in place
- animated while it runs
- terminal state when the `run_pipeline` call completes

Settling is driven by that tool call rather than by the frames themselves, because the final `pipeline_step_completed` is indistinguishable from any other until the tool returns — the same signal `lifecycle_forwarder` unsubscribes on.

```
rag_ingest.ingest  ────────────  0/15  transform
rag_ingest.ingest  ━━━━━━──────  7/15  transform
rag_ingest.ingest  ━━━━━━━━━━━━  15/15  transform
rag_ingest.ingest  step 4  transform          <- total unknown: count, no bar
```

## Rendered from meta, not from the sentence

The forwarder composes a display string AND now carries the numbers it already had (`pipeline_name` / `step_index` / `total_steps` / `step_kind`) in meta. The row reads those.

Parsing the numbers back out of the forwarder's text would have worked today and broken silently the first time its wording changed — the row and the sentence would be two renderings of one fact, one of them derived from the other's prose.

A run with no known total gets its step count and **no bar**: a bar over a guessed denominator shows a position on a scale that does not exist.

## Evidence

End to end through the coalescing path, 15 steps:

```
投入フレーム数: 30
★ 生成された flow entry 数: 1
最終行の meta: {step_index: 14, total_steps: 15, step_event: pipeline_step_completed}
```

## Test plan

- [x] `test_tier_audit.py --strict` on the new test file — 0 errors
- [x] `verify_module_docstrings.py` on all four changed files — OK
- [x] `pytest -k "textual or tui or flow or pipeline"` — **900 passed**
- [x] End-to-end fold measured: 30 frames → 1 entry, final state retained (above)
- [x] The bar is asserted as a RELATION across three points rather than pinned at one, so it survives a change to the bar width
- [ ] (skipped — needs a TTY and a real pipeline) Visual confirmation that the row reads well while a run is in flight. The owner is running this build and reported the original defect; the rendered text is measured above, but how it looks mid-run is their call.

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/_meta_keys.py
+++ b/src/reyn/interfaces/inline/textual_chat/_meta_keys.py
@@ -80,8 +80,15 @@ RUNNING_SINCE_KEY = "_running_since"
 #: value. This is a decision (owner-adjudicated on #3283), not an oversight.
 ELAPSED_SECS_KEY = "_elapsed_secs"
 
+#: Marks a row as one pipeline RUN'S progress rather than a one-off status
+#: line, and carries the run id. Set app-side when a step frame is folded in
+#: (``_coalesce_pipeline_step``); the presenter reads it to render the run's
+#: state instead of the latest frame's text.
+PIPELINE_RUN_KEY = "_pipeline_run_id"
+
 __all__ = [
     "ELAPSED_SECS_KEY",
+    "PIPELINE_RUN_KEY",
     "ORPHANED_RESULT_KIND",
     "RESULT_KIND_KEY",
     "EXPANDED_KEY",
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -69,6 +69,7 @@ from reyn.interfaces.transport.frames import FrameTag
 from ._meta_keys import ELAPSED_SECS_KEY as _ELAPSED_SECS_KEY
 from ._meta_keys import EXPANDED_KEY as _EXPANDED_KEY
 from ._meta_keys import ORPHANED_RESULT_KIND as _ORPHANED_RESULT_KIND
+from ._meta_keys import PIPELINE_RUN_KEY as _PIPELINE_RUN_KEY
 from .chrome import (
     _MENU_TABS,
     Composer,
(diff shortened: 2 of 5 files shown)


tya5:
**[tui-coder]** — Owner asked whether the bar uses Rich's rather than a reinvention. It was a reinvention; fixed in `e8f1a2c`-onward on this branch.

`rich.progress_bar.ProgressBar` exists, and the presenter's return type is already `RenderableType` — the `presentation` branch returns a non-Text renderable — so nothing was blocking its use. I did not look before writing `"━" * filled`.

**The reimplementation was not equivalent.** Rich renders a half-cell tip:

```
before (mine)   rag_ingest.ingest  ━━━━━━──────  7/15  transform
after  (Rich)   rag_ingest.ingest  ━━━━━╸        7/15  transform
```

A 12-cell bar advances in 24 steps rather than 12. For a bar this short that resolution is most of the reason to draw one at all — and it is exactly the kind of detail a from-scratch version silently drops.

Styles are passed explicitly instead of Rich's default magenta/green, for the same reason the palette work landed: a themed default is the user's to choose, and the bar is not one of the two places this CUI claims a colour.

Tests now render through a real `Console`. A Rich renderable's cells only exist once something renders it, so asserting on the object would have been asserting on the recipe rather than the meal. The bar assertions stay relational across three points rather than pinning glyphs that are Rich's to change.

`pytest -k "textual or tui or flow or pipeline or presenter"` — 901 passed.

Issue #3664, opened 2026-08-03
present's ack calls a displayed value 'dropped', and counts rows it did not draw


tya5:
**[tui-coder]** — Barrier 3 of #3630, split out so it can be decided rather than wait. Needs an agreed shape, not a defect fix: the ack is the contract the model is prompted against.

Barriers 1 (#3635) and 2 (#3632) are merged and the acceptance condition is [witnessed against the real model](https://github.com/tya5/reyn/issues/3630#issuecomment-5161189325). This is what is left.

## The two statements that are wrong

Both measured, on the path a model actually takes.

**1. `bindings_dropped … type_mismatch` means the opposite of what happened.** The value was not dropped — it was coerced to its JSON form and **displayed**. From `binding.py`, verbatim: *"Type mismatch → coerce (… a container bound into a text slot → its JSON form) and record `{path, reason: type_mismatch}`"*. The model reads "dropped" and has no reason to think the user is looking at a JSON blob.

**2. `rows` counts only TABLE rows.** `out.rows += len(rows)` fires in the table/list binding path alone (`binding.py:207`). A `keyvalue` card with four visible rows reports `rows: 0`. Measured: the owner's original payload rendered four rows and reported `0`.

Together they make a flattened rendering read as a success with nothing shown — which is why the model never corrected itself over two live calls.

## Why it still matters after barriers 1 and 2

Less often, not never. Barrier 1 stops the common nested shapes from flattening, but the floor is still there — past the descent bound, and for any value a text slot cannot hold. When it fires, the ack describes it the same wrong way.

## Proposal

Three changes, smallest first. The first is nearly free and fixes the worst of it.

**(a) Split coercion out of `bindings_dropped`.** A new `coerced: [{path, rendered_as: "json_text"}]`, leaving `bindings_dropped` for values that genuinely did not reach the user (`path_not_found`, `guard_stripped`). Dropping and flattening are opposite outcomes; sharing one word is the defect.

**(b) Make `rows` mean what it reads as, or rename it.** Either count every rendered row, or call it `table_rows`. A count that is `0` for a visibly four-row card is worse than absent — and the model uses it to decide whether anything was shown.

**(c) Say so once, in words the model can act on.** When the default viewer coerces a container into a text slot, a short note naming `blueprint` + `table`. The `view` path already has `all_bindings_missed` as its self-correction signal; this path has none.

## Not proposed

[...]

PR #3740, opened 2026-08-06
fix(present): ack no longer calls a displayed value "dropped"; rows counts all rendered rows

tya5:
**[e2e-coder]** — Implements all 3 points from #3664, per the owner's ruling (relayed by lead-coder, https://github.com/tya5/reyn/issues/3664#issuecomment-5201700020): (b) is A — `rows` is redefined to mean "every rendered row," not just table rows.

## (a) Split coercion out of `bindings_dropped`

A type-mismatch coercion is the *opposite* of a drop — the value reached the user, reshaped, not lost. New `coerced: [{path, rendered_as}]` field (`rendered_as ∈ {json_text, single_row}`); `bindings_dropped` now only carries `path_not_found` / `guard_stripped` — values that genuinely never reached the user.

## (b) `rows` = every rendered row, post-cap

Redefined per the owner's own question — "what is `rows` *for*?" — traced to 2 consumers: the op ack (the LLM's only feedback; `present` is a zero-token offload, the data itself never enters the LLM's context) and the replay header (`[present] view=… rows=N …`, sourced from the same `presented` event field, so the two consumers cannot diverge in meaning). `rows` is therefore the LLM's only measure of how much reached the user's screen, and now counts:
- `keyvalue` card rows (previously **uncounted** — a 4-row keyvalue card reported `rows: 0`, the issue's measured defect)
- `table`/`list` rows bound via `$bind` **and** given as a literal array (previously the literal-array branch was also uncounted)
- **post-cap**, not pre-cap — found while implementing this: `_bind_rows_slot` was incrementing `rows` on the pre-`cap_rows` count, so a truncated table could over-report how many rows the user actually saw, directly contradicting the purpose just established. Fixed to count `capped_rows` (what's actually rendered).

## (c) Self-correction note for the default viewer

The `view`/`blueprint` path already had `all_bindings_missed` as a self-correction signal when nothing rendered as requested. The `mode: "default"` path had no equivalent when its *own* viewer degrades a container to raw JSON text. New: when the content-type default viewer coerces a container into a text slot, the ack carries a `note` naming the fix (`blueprint` + a `table` component).

## Before / after (LLM-visible ack shape — owner's gate axis 1)

```yaml
# before
ok: true
mode: default
bindings_resolved: 6
rows: 2                # table rows only — a 3-row keyvalue card above it is invisible
bindings_dropped:
  - {path: "/deep", reason: type_mismatch}   # misleading: this WAS displayed
all_bindings_missed: false

# after
ok: true
mode: default
bindings_resolved: 6
bindings_dropped: []                          # type-mismatch is no longer a "drop"
coerced:
  - {path: "/deep", rendered_as: json_text}   # honest: displayed, reshaped
rows: 5                                        # 3 keyvalue rows + 2 table rows
all_bindings_missed: false
note: "part of this data was too deep for the default viewer's table columns and was shown as raw JSON text instead — pass an explicit `blueprint` with a `table` component to render it structured."
```

## Docs

Same-PR sync per CLAUDE.md's doc-drift rule: `docs/reference/runtime/present.md` + `.ja.md` (ack shape, binding-outcomes prose, `presented` event field table), `docs/reference/runtime/control-ir.md` + `.ja.md` (same present-op section), `docs/concepts/runtime/present.md` + `.ja.md` (fallback self-correction prose). `docs/deep-dives/proposals/0054-present-layer.md` left as-is — a frozen historical design snapshot, not a live doc.

## Tests

- 3 existing tests updated to assert the new `coerced` field / corrected `bindings_dropped` shape instead of the old `type_mismatch`-as-drop behavior (`test_scalar_into_table_coerces_to_one_row_type_mismatch`, `test_ack_shape_reports_drops_with_reason_categories`, `test_descent_stops_and_says_so`).
- 1 existing test's assertion updated for the new `rows` semantics (`test_a_nested_row_list_reaches_the_user_as_a_table`: `rows == 2` → `rows == 5`, since a keyvalue card now counts too) — a real behavior change, not a bug in the old assertion.
- New tests: keyvalue row counting, literal-array row counting, post-cap (not pre-cap) row counting on an already-existing cap-truncation test, and the (c) self-correction note.

## Test plan

- [x] `scripts/test_tier_audit.py --strict` on changed test files — 54/54 OK, 0 Tier-4 violations
- [x] `pytest` — present-scoped subset (387 passed) confirmed green; full-repo run in progress, will report before merge
- [x] `scripts/verify_module_docstrings.py` on changed src files — OK
- [x] replay-header consistency confirmed: `_header()` reads `rows` straight off the same `presented`-event field the ack derives its own `rows` from — single source, cannot diverge (documented inline in `replay.py`)
- [x] `rows`'s *purpose* documented in the dataclass field + the ack-construction comment in `present.py`, per the owner's condition

part of #3664


tya5:
**[lead-coder]** — ✅ **承認。★ before/after が「なぜ嘘だったか」を 1 行ずつ示しています。**

```yaml
# before
rows: 2                # ★ table rows only — a 3-row keyvalue card above it is invisible
bindings_dropped:
  - {path: "/deep", reason: type_mismatch}   # ★ misleading: this WAS displayed
# after
rows: 5                # 3 keyvalue rows + 2 table rows
coerced:
  - {path: "/deep", rendered_as: json_text}  # ★ honest: displayed, reshaped
```

★ **`rows: 2` の隣に「3 行の keyvalue カードが見えていない」と書いてある** — ★ **owner の問い「`rows` はなんのための情報なの？」に対する答えが、diff の形で出ている**のが良いです。★ **LLM は「利用者が 2 行見た」と読み、実際は 5 行**でした。

★ **`coerced` を *新しい field* にした判断も正しい**です。★ **`bindings_dropped` から外すだけだと情報が消えます** — ★ **「表示された、ただし形を変えて」は、落ちたのでも普通に出たのでもない第三の事実**なので、名前が要ります。★ **`rendered_as` で *どう* 変えたかまで言っているのも、LLM が自己修正できる形。**

★ **`control-ir.md` を同 PR で直している**のも確認しました（★ **CLAUDE.md のハードルール**: 「機構を記述する doc は同じ PR で」）。★ **`Type mismatch → coerce + record` の記述が、新しい field 名に合わせて更新**されています。★ **replay 側（`replay.py`）も触っており、2 つの消費者で意味が割れていません** — 私が付けた条件です。

⚪ ★ **ja 側も直している**のは、★ **該当箇所が今回の変更で偽になるから**であって mirror 義務ではない、という理解で合っていますね（CLAUDE.md の #5）。

★ **未チェックの Test plan 項目なし。マージします。**

## ⚪ 1 つ申し送り（対応不要）

★ **`note` の文面が長い**です（LLM 向けなので情報量は正しい）。★ **これは owner ゲート第 1 軸のものなので、将来これを縮める/変える変更も同じゲートを通ります** — ★ **今回 owner の裁定を得たのは `rows` の意味についてであって、`note` の文面ではありません。** 次に触る人向けの記録として。

PR #3739, opened 2026-08-06
docs(mypy_ratchet): document the (file, code) grain's known blind spot

tya5:
**[e2e-coder]** — follow-up requested on #3738's review (the granularity tradeoff, not the [arg-type] finding itself, which stays out of scope here — tracked separately below).

## What

Adds one paragraph to `scripts/mypy_ratchet.py`'s module docstring, naming explicitly what #3738's review surfaced: `(file, error-code)` grandfathers a genuinely NEW finding of the SAME code in a file that already carries that code — invisible, indistinguishable from a real fix, confirmed live by #3738 itself (a new `[arg-type]` in `session.py`, already carrying a baselined `[arg-type]` entry, stayed invisible to the ratchet).

`(file, line, code)` would close this, but line numbers don't survive an unrelated edit shifting them on a moving `main` — regenerating the baseline to follow drift would be indistinguishable from regenerating it to silence a real new failure, which the module's own docstring already names as the one way to defeat the ratchet. So the coarser grain is recorded as a deliberate trade (stability across unrelated edits, at the cost of same-code same-file blindness), not left as an undiscovered gap.

No code/behavior change — docstring only.

## Remaining item

The `[arg-type]` finding itself (tightening `Session`'s own `registry` field to non-Optional) is explicitly out of scope for this PR per lead-coder's ruling — left as a `part of #3593` remainder item on that issue, not filed as a new one.

## Verification
- `ruff check`, mypy ratchet (unchanged: 211/212), `verify_module_docstrings.py`, and the existing 16-test `test_3726_mypy_ratchet.py` suite — all green.

## Test plan
- [x] `ruff check`, mypy ratchet, `verify_module_docstrings.py` all green
- [x] Existing `test_3726_mypy_ratchet.py` (16 tests) unaffected, still green
- [x] `[arg-type]` finding left as a `part of #3593` issue remainder, not silently dropped

PR #3738, opened 2026-08-06
fix(#3593): tighten CapabilityVisibility's registry type, keep the runtime guard

tya5:
**[e2e-coder]** — part of #3593 stage ②. Not closing — this is the type-tightening/test-cleanup follow-up to #3614 (stage ①, preserve+warn) and #3655 (bootstrap-ordering fix).

## Ruling (A, per lead-coder)

Tighten the type statically, **keep the runtime guard** — the axis is "which direction a bypass fails", not "which callers currently pass what". The guard protects a security property: if a `None` ever reached it and the guard were gone, the failure isn't a crash, it's composing the override against an allow-everything default and SETting it — a persisted `tool_deny` narrowing silently replaced by "allow-all minus whatever was toggled". Documented inline, framed as "what breaks" not "don't touch this".

## What changed

- `CapabilityVisibility.__init__`'s `registry` param: `_EnvelopeSource | None` → `_EnvelopeSource` (static-checking signal — verified all 5 real production callers of `build_scoped_chat_session` already pass a real value, including dogfood.py's own `#3593 ②: always the real registry` comment from #3655).
- Runtime guard in `reapply_visibility_override` **unchanged**, with a new comment explaining the failure direction if it were removed.
- Deletes exactly the 2 named tests (`test_the_preserve_branch_is_reached_on_the_restart_path` / `test_a_wired_session_does_not_reach_the_preserve_branch`) — both asserted reachability via the PRODUCTION restart path specifically, which the bootstrap fix closes.

## What did NOT change (verified before implementing, not assumed)

`test_envelope_narrowing_survives_a_reapply_with_no_base` stays. Traced the full call chain before touching anything: it reaches the exact same `self._registry is None` branch as the 2 deleted tests (`make_session(registry=None)` → `Session(registry=None)` → `self._registry = None` → `CapabilityVisibility(registry=None)`), but it asserts the class's own behavioral CONTRACT (values: narrowing survives, override doesn't reach the live gate) rather than a claim about what production can still reach — a regression test for `CapabilityVisibility` given no registry at all, independent of whether any real caller currently does that.

## A gap worth surfacing, not hiding

[...]

PR #3771, opened 2026-08-06
fix(#3593): CapabilityVisibility.registry is genuinely Optional, not a non-Optional lie

### LINKED ISSUE #3593: Re-resolving a session's capability envelope from base and SETting it lands on allow-all when the base is unavailable
**[per-PR-coder]** — found while measuring #3562 (PR #3586). Filed, not fixed; the mechanism below is measured, the remedy is deliberately not proposed.

## The mechanism

`CapabilityVisibility.reapply_visibility_override` (`src/reyn/runtime/capability_visibility.py:284`) re-resolves the WHOLE agent envelope from base and then **SETs** `_contextual_permission` / `_excluded_categories` — a SET, not a merge, which is correct and deliberate: it is what lets a `/visibility` toggle-ON restore a capability *up to* the envelope without re-widening past it (its docstring explains why a union could not).

The base it re-resolves from is:

```python
if self._registry is not None and hasattr(self._registry, "resolved_profile_for"):
    raw_ctx, base_excl = self._registry.resolved_profile_for(
        self._agent_name, sid=self._session_id_provider(),
    )
```

★ **When there is no registry back-reference, there is no base — and the code proceeds anyway.** `base_ctx` stays `None`, `ContextualPermission()` (allows everything) is composed with the override alone, and the result is SET. So the session's whole envelope — topology `capability_profile` bindings, the #2081 `_delegate` floor, the #2103-S1a per-session narrowing — is replaced by "allow all, minus whatever the operator happened to have toggled off". The absent input is read as **no restriction** rather than as **unknown**, and the direction of that reading is outward.

## Reachability — two instances, both measured

**(1) Via `refresh_config_projections()`** — the `visibility_override` hot-reload seam. Measured during #3586: routing `spawn_session_recorded`'s `narrowing` through the primitive's new channel (i.e. injecting *before* the refresh instead of after) turned two existing behavioural tests RED with an empty live `tool_deny`:

- `tests/test_2103_s1bc_session_spawn_tool.py::test_spawn_session_recorded_enforces_narrowing_on_live_session`
- `tests/test_pipeline_a2_spawn_ephemeral_session.py::test_spawn_ephemeral_session_narrowing_applied`

That is why #3586 leaves `spawn_session_recorded` injecting after its own refresh; the ordering is now load-bearing rather than incidental.

**(2) Inside `AgentRegistry.spawn_session` itself** — the one I did not expect. `spawn_session` calls `load_persisted_toggles()` immediately after the #3561 re-resolve-and-inject, and that loader calls `reapply_visibility_override()` whenever the session has a persisted `visibility.yaml` (`session.py:2372`). Two arms differing **only** in the registry back-reference, same sid, same `config.yaml` narrowing (`tool_deny: [write_file]`), same persisted toggle (`tool: [read_file]`):

```
with_registry=True : live tool_deny = {'write_file', 'read_file'}   ← narrowing + toggle
with_registry=False: live tool_deny = {'read_file'}                 ← narrowing GONE
```

Both arms wrote the same `config.yaml`; only the live gate differs. This path has no `refresh_config_projections()` in it at all, so it reaches **every direct caller of the primitive** — crash-recovery re-wake (`restore_all` / `_rewake_pipeline_runs`, the sites #3561 fixed) and `/session new` (#3562) included. The precondition is a persisted `/visibility` toggle for that sid, which is ordinary restart state.

(Probe script kept out of the repo — it is 50 lines of the same harness `tests/test_3561_spawn_session_seam_reachability.py` uses, with `registry=` as the only variable. Happy to land it as a test if this is taken up.)

## Why "contained today" is a claim about the callers, and what I measured of it

The containment rests on production sessions always carrying the back-reference. Enumerated, not assumed — the 5 `build_scoped_chat_session` call sites in `src/`:

| call site | passes `registry=` |
|---|---|
| `runtime/registry_bootstrap.py:176` (the canonical registry session factory) | yes, unconditionally |
| `interfaces/web/deps.py:308` | yes, unconditionally |
| `interfaces/cli/commands/chat.py:490` | yes, unconditionally |
| `interfaces/cli/commands/mcp.py:448` | yes, unconditionally |
| `interfaces/cli/commands/dogfood.py:512` | ⚠️ **conditionally** — `registry=_reg_cell[0] if _reg_cell else None` |

The three `_no_factory` registries (`cli/commands/topology.py`, `cli/commands/agent.py` ×3) never build a session, so they are not in scope.

★ So containment is **4 of 5 unconditional, 1 of 5 conditional**. The dogfood factory's own comment says the cell "may be empty during bootstrap"; I did not measure whether a session is actually built inside that window, so I am not claiming either way — the honest statement is that the containment argument has one caller-shaped hole in it, not that it holds everywhere.

⚠️ And containment is a property of today's caller set, which is exactly the kind of thing a future caller changes without noticing: nothing in the code says "this collaborator is required for the security-core path to be correct", and nothing fails when it is absent — the failure is silent and outward.

## Why this is worth a ticket rather than a code comment

The asymmetry, the same one that decided #3562: a wrong answer on the narrow side announces itself (something is refused, the operator can lift it), a wrong answer on the wide side announces nothing — the session simply has capabilities nobody granted it, and the surface that would show it (`capability_visibility_state`) reads the same base and agrees.

There is a code comment about the ordering in `registry.py` (from #3586), but it documents the *workaround at one call site*, not the fail-open. This issue is the fail-open.

## Not in scope here

No remedy is proposed in this title or body. The choice between refusing to re-resolve without a base, treating an absent base as the current value, requiring the collaborator at construction, or something else is a design question with blast radius across #2285 / #3097, and I would rather it be decided than inherited from the first phrasing of a ticket.

tya5:
**[e2e-coder]** — closes #3593. Fresh branch off main.

## Summary
This resolves the last recorded remainder on #3593: the `[arg-type]` finding at `session.py:1359` (`Argument "registry" to "CapabilityVisibility" has incompatible type "Any | None"; expected "_EnvelopeSource"`), grandfathered invisible by the mypy ratchet's `(file, code)` grain (documented in #3739), left open as a `part of #3593` remainder rather than filed separately.

Measured before implementing (lead-coder's explicit correction to my first plan, which over-applied their own earlier "typed union, not form-sniff" guidance): is "no usable base" one state or two-plus? Read the full guard in `reapply_visibility_override` — `registry is None or not hasattr(..., "resolved_profile_for")` — both arms produce the identical outcome (preserve + warn, same code path, same message), and no test exercises the `hasattr` arm as a distinct scenario. One state, not two: `X | None` is already a fully-typed discriminator, so a sentinel class + `isinstance` would just re-say what `| None` already says, and would split "no base" into two spellings (public Optional, internal sentinel) that drift apart over time.

`Session`'s own `registry` field is genuinely `AgentRegistry | None` — a session legitimately exists without a back-reference in some construction paths, confirmed by `session.py:1358`'s direct passthrough (`registry=self._registry`, no `None`-branch beforehand). So `CapabilityVisibility`'s own `registry: "_EnvelopeSource"` (non-Optional) was the actual lie, not Session's Optional field. Corrected to `_EnvelopeSource | None`, matching what every caller actually passes. `is None` was always the real discriminator at runtime; it is now also the typed one. No sentinel, no `isinstance`, no new abstraction — a 1-line type correction plus updating 2 comments that had asserted the type would eventually become non-Optional (it will not; the state is legitimate, not a wiring defect to engineer away).

## Enumeration (closing-keyword safety, CLAUDE.md rule 4)
Every PR referencing #3593: #3614 (stage ①, MERGED), #3650 (the #3615 read-model twin, MERGED), #3655 (stage ②, MERGED), #3738 (the earlier type-tightening attempt this supersedes, MERGED), #3739 (mypy ratchet docs, MERGED). All merged; this PR completes the only remaining recorded remainder. No other open PR or issue references #3593.

## Test plan
- [x] `pytest` on all 4 `#3593`/`#3615`/pipeline-driver-narrowing related test files — 20 passed, 0 failures
- [x] Broader sweep (`-k "capability_visibility or visibility_override or 2285"`) — 26 passed, 8 skipped, 0 failures
- [x] `mypy` diff (git-stash comparison): the specific `registry` arg-type error is present before this fix, gone after — confirmed directly, not inferred from the ratchet's coarse-grained pass/fail
- [x] `python scripts/mypy_ratchet.py` — 211/212, unchanged (baseline is presence-only; `session.py`'s `arg-type` entry stays declared regardless, masked by other unrelated findings in the same file) — baseline file untouched
- [x] `ruff check src/reyn/runtime/capability_visibility.py` — clean
- [x] (skipped — full-suite run not needed; single-file, type-annotation-only change with no behavioral difference, all directly-affected + broader-sweep tests green)

### DIFF
--- a/src/reyn/runtime/capability_visibility.py
+++ b/src/reyn/runtime/capability_visibility.py
@@ -200,7 +200,7 @@ class CapabilityVisibility:
     def __init__(
         self,
         *,
-        registry: "_EnvelopeSource",
+        registry: "_EnvelopeSource | None",
         router_host: "_RouterHost",
         session_id_provider: "Callable[[], str | None]",
         agent_name: "str",
@@ -333,15 +333,17 @@ class CapabilityVisibility:
         # its declared type is the wider `object | None`, so cast to the concrete type the
         # downstream compose_resolved requires (registry.py:3509 guarantees it).
         #
-        # #3593 ② review: `registry`'s type is `_EnvelopeSource` (non-Optional) now that
-        # every production caller of build_scoped_chat_session passes a real one — but a
-        # type annotation is a static-checking signal, not a runtime guarantee (Python does
-        # not enforce it), and this `is None` check stays regardless. If it were removed,
-        # the failure direction on a genuinely unreadable base is not a crash — it composes
-        # the override against an allow-everything default and SETs it, so a persisted
-        # `tool_deny` narrowing is silently replaced and the tool becomes ALLOWED. That is a
-        # permission WIDENING, the opposite of fail-closed, which is why the type alone
-        # (a compile-time-only signal) is not sufficient here.
+        # #3593 review: `registry` is genuinely `_EnvelopeSource | None` — Session's own
+        # `registry` field is `AgentRegistry | None` (a caller without a back-reference is
+        # a real, legitimate state, not a wiring bug to paper over with a non-Optional lie)
+        # — so `is None` is a fully-typed discriminator, not a runtime-only guess about a
+        # static signal. `not hasattr(..., "resolved_profile_for")` stays alongside it as a
+        # SEPARATE, purely defensive check: Python does not enforce the Protocol at
+        # runtime, so a non-conforming object could theoretically reach here despite the
+        # type. Either arm means the same thing (no base to read), which is why they share
+        # one branch: composing the override against an allow-everything default and
+        # SETting it would silently replace a persisted `tool_deny` narrowing with ALLOWED
+        # — a permission WIDENING, the opposite of fail-closed.
         if self._registry is None or not hasattr(self._registry, "resolved_profile_for"):
             # #3593 ①: preserve — see the docstring. No base was obtained, so nothing below
             # may run: everything below composes a NEW envelope and SETs it over the live one.
@@ -402,9 +404,13 @@ class CapabilityVisibility:
         - Sibling precedent in this same class: ``persist_visibility_override`` reports its
           best-effort failure the same way.
 
-        Stage ② (#3593) is meant to make this branch unreachable by construction — fix the
-        one bootstrap-ordering caller, then require the envelope source non-optionally. Until
-        then this warning is what would tell an operator the window widened.
+        Stage ② (#3593) fixed the one measured bootstrap-ordering caller that reached here
+        unnecessarily (dogfood's registry cell). This branch stays reachable by DESIGN,
+        though: ``registry`` is genuinely ``AgentRegistry | None`` — a session legitimately
+        exists without a back-reference in some construction paths — so "no envelope
+        source" is not itself a defect to engineer away, only a state that must never widen
+        permissions. This warning is what tells an operator it happened, on any path that
+        reaches it.
         """
         import logging
 


tya5:
**[lead-coder]** — ✅ **承認します。★ 載っている前提を私の側でも実測しました。**

```
session.py:926   registry: "AgentRegistry | None" = None
session.py:1080  self._registry = registry
```

★ **`Session` の側が本当に Optional** です ∴ ★ **非 Optional の注釈は嘘でした。** ★ **1 行の型修正が、コードを変えずに宣言を実物に合わせています。**

### ★ 私が確かめたもう 1 点 — **型を広げても安全側に倒れる**

```python
if self._registry is None or not hasattr(self._registry, "resolved_profile_for"):
    self._surface_unreadable_envelope_source()
    return          # ★ preserve。新しい envelope を SET しない
```

★ **注釈を広げると、静的な保護が 1 つ減ります**（新しい呼び出し側が誤って `None` を渡せる）。★ **その場合どこへ倒れるかを見ました** — ★ **preserve して警告する**、つまり ★ **権限を広げない側**です。∴ ★ **受け入れられます。**

### ⚪ 直した散文のほうが、型より重要でした

> `Stage ② (#3593) is meant to make this branch unreachable by construction`

★ **これは偽になっていました。** ★ **「到達不能にする予定」と書かれた分岐が、実際には設計上到達可能**です。★ **次の人がこの一文を根拠に `is None` を削れば、権限が黙って広がります。**

⚪ **`Closes #3593` も妥当**: ★ **先行 5 本の PR を列挙して全て merged と確認し、GraphQL の `closingIssuesReferences` で対象が #3593 のみと検証済み**とのこと。★ **本日の規律どおりです。**

PR #3775, opened 2026-08-07
fix(tui): the input-row gap belongs to FlowView, not the input row

tya5:
**[e2e-coder]** — owner request (relayed via lead-coder, 2026-08-07): "the blank line above the input should move to directly below FlowView." Reason given: the conversation and the NOW row (#3693) read as connected, hard to read.

## Measured — current owner (before this PR)
The blank line lived as `#inputrow`'s own `margin-top: 1`. With a turn running, `FlowView` and `ActivityRow` (the NOW row) sat flush against each other — a **0-row gap** — while the gap instead separated the NOW row from the input, which is backwards from what the owner wants separated.

## Fix
Moves the gap onto `FlowView`'s own `margin-bottom: 1` and removes `#inputrow`'s `margin-top`. `FlowView` is the ONE region in this chrome stack that is never `display=False` — every widget between it and the composer (`ActivityRow`, `SearchBar`, `SentQueue`, `CompletionPopup`, `RewindPicker`, `InterventionPanel`) defaults to hidden. A margin on any one of THOSE would vanish exactly when that widget hides; owning it on `FlowView` means the gap survives regardless of which (if any) transient region currently occupies the space below it.

## 3-state coordinate table (region `y`/`height`, before → after)
```
idle:              gap(flow->input)=1              -> gap(flow->input)=1        (unchanged)
turn running (NOW): gap(flow->NOW)=0 gap(NOW->input)=1  -> gap(flow->NOW)=1 gap(NOW->input)=0
search bar open:    gap(flow->bar)=0 gap(bar->input)=1  -> gap(flow->bar)=1 gap(bar->input)=0
```
The owner's literal complaint — conversation and NOW touching — is the `turn running` row: `gap(flow->NOW)` goes from **0** (touching) to **1** (separated). `idle` is unaffected (there was never a "connected" pair to separate when nothing is between FlowView and the input).

## Test plan
- [x] `pytest tests/test_conversation_chrome_separation.py -v` — 3 passed.
- [x] Strip-falsify: reverting the CSS change makes 2 of the 3 tests go RED (`turn running` and `search bar open` — the actual gates on this fix). The third (`idle`) measures `gap=1` identically before and after — it's labeled in its own docstring as a plain regression guard on the unchanged case, not a witness of the fix (its docstring says so explicitly, per review feedback that an unfalsifiable gate must either be removed or say plainly what it protects).
- [x] `pytest tests/ -k "textual_chat or activity_row or 3693 or 3680 or compact_layout" -q` — 229 passed, 8 skipped (unrelated).
- [x] Full `pytest` from repo root — 10123 passed, 49 skipped, 0 failed.
- [x] `python scripts/test_tier_audit.py --strict tests/test_conversation_chrome_separation.py` — 3/3 OK, 0 errors.
- [x] `python scripts/verify_module_docstrings.py` on changed src files — OK.

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -789,6 +789,7 @@ class TextualChatApp(App):
     FlowView {
         height: 1fr;
         scrollbar-size-vertical: 0;
+        margin-bottom: 1;
     }
     /* #3496 / flowview#5: ``flowview--highlight`` (``--selected`` was its
        0.11.x synonym for the SAME class; 0.12.0 / #3624 removed the alias, so
@@ -819,7 +820,6 @@ class TextualChatApp(App):
     #inputrow {
         height: auto;
         max-height: 8;
-        margin-top: 1;
         border-top: solid @rule@;
         border-bottom: solid @rule@;
     }
--- a/tests/test_conversation_chrome_separation.py
+++ b/tests/test_conversation_chrome_separation.py
@@ -1,0 +1,178 @@
+"""Owner request (relayed via lead-coder, 2026-08-07): "the blank line above
+the input should move to directly below FlowView" — reason given: the
+conversation and the NOW row (#3693) read as connected, which is hard to
+read.
+
+Measured BEFORE this fix: with a turn running, ``FlowView`` and
+``ActivityRow`` were vertically ADJACENT (0-row gap) — the blank line lived
+as ``#inputrow``'s own ``margin-top`` instead, between the transient chrome
+(NOW row / search bar) and the input. That is backwards from what separates
+conversation from status: the gap belongs to the conversation's own trailing
+space, not the input's leading space.
+
+The fix moves the gap onto ``FlowView``'s own ``margin-bottom`` — the ONE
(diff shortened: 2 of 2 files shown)


tya5:

★ **「空行を足す」でなく「誰が持つか」で解いたのが正解**です。

> `The fix moves the gap onto FlowView's own margin-bottom — the ONE region that never becomes display=False`

★ **3 状態とも成立した理由が、この 1 文に書かれています。** ★ **chrome 側に margin を置けば、その chrome が隠れた瞬間に gap も消える** — ★ **owner の要望は「NOW と離す」ですが、NOW が出ていない時に破綻しない形を選んでいます。**

⚪ **idle テストのラベル修正も確認**しました。

> `The idle case measures the SAME gap=1 whether the fix is present or not`

★ **自分で「これは witness ではない」と書いた**のが正しい。★ **strip で赤にならないゲートを「守っている」と数えない**、が守られています。

⚠️ **owner に 1 点確認中**（マージは止めません）: ★ **NOW と入力欄が 0 行で接触します。** ★ **そちらが「繋がって見える」と判定されたら、間隔の総数を増やす follow-up を出します。** ★ **owner は会社の端末で main を見るので、入れないと判定自体が始まりません。**

PR #3773, opened 2026-08-07
feat(#3692): PR-A — flowview 0.13.1 pin, retire reyn's copy-mode wiring

tya5:
**[e2e-coder]** — part of #3692 (PR-A of two; PR-B covers Ctrl+O / Esc-layering / Ctrl+F arbitration, deferred separately).

## Summary
- Bumps the `textual-flowview` pin from v0.12.0 to v0.13.1 (`2408cabfdc896fa5e49c1f64b8ae9f01e33d9346`). flowview 0.13 replaced copy mode (explicit entry step, `enter_copy_mode`, `CopyModeChanged`) with an always-on text cursor (`show_cursor`/`hide_cursor`/`toggle_cursor`/`cursor_visible`) plus visual mode as the one real mode. Pinned straight to 0.13.1 (not 0.13.0), which fixes the addressed-row-rail regression 0.13.0 introduced.
- Removes the 6-site copy-mode wiring table from issue #3692's own body: the app-level `("c", "copy_mode", …)` BINDINGS tuple, `enter_copy_mode()` call, `action_copy_mode`, `self._copy_mode` state, `on_flow_view_copy_mode_changed`, and the `copy_yank`-referencing sync-contract comment (updated to `yank`).
- Hands `c` entirely to flowview (`toggle_cursor`, reached by ordinary bubbling — reyn declares no binding of its own for it). Yank consolidates onto flowview's own `y`.
- Doc-sync in the same PR (CLAUDE.md hard rule): `agui-transport.md`'s "Textual TUI copy mode" section renamed/reworded to "Textual TUI text cursor", `feature-map.md` row 327 rewritten, `chat-and-web-ui.md`'s `c` key row reworded to describe the always-on toggle. Grep-zero for "copy mode"/"copy_mode"/"CopyMode" achieved across `src/`, the 3 doc files, and `tests/test_copy_mode_3507.py` (the one surviving literal occurrence names the actual removed `on_flow_view_copy_mode_changed` symbol for historical identification, not the retired concept).

## Test disposition (`tests/test_copy_mode_3507.py`, lead-coder's a/b/c framework)
- **(a) deleted** — `test_the_chrome_sees_both_copy_mode_edges` tested the removed `CopyModeChanged` narration surface directly; its incidental Esc-layering coverage is PR-B's to re-pin as 3 separate tests per #3692's acceptance criteria.
[...]

PR #3778, opened 2026-08-07
docs(3525): record owner's colour-carveout extension in CLAUDE.md

tya5:
**[docs-maintainer]** — Record-only, per lead-coder's dispatch (implementation is e2e-coder's, separately).

Owner (2026-08-07, telegram): a reyn-specific colour meaning — one with no established convention — may use a token value outside ANSI-16. A conventional meaning (red = error, etc.) still resolves through ANSI-16 / `ansi_default` / an SGR attribute, unchanged. The `ansi_default`-alpha-composite ban is untouched.

Extended the existing `TUI colour policy` paragraph (CLAUDE.md ~line 41) rather than replacing it — the original #3525 ruling stays intact, the carve-out is a new clause appended to it.

Included the reason lead-coder said must not be dropped: "conventional vs reyn-specific" is a judgment call no gate can make, so the token-indirection requirement (every value — carve-out or not — still goes through a `palette.py` token, never a stylesheet literal) is what keeps `tests/test_tui_colour_tokens.py` meaningful. Without stating that explicitly, a future reader could read the token requirement as formalism and drop it, which would let a stylesheet claim "reyn-specific" for anything.

## Test plan
- [x] `python scripts/check_doc_anchors.py` — 0 dangling anchors
- [x] `pytest tests/test_3126_doc_anchor_gate.py` — 332/332 pass


tya5:

★ **私が「必ず含めてほしい」と言った理由が、私の言い方より良く書かれています**:

> `"is this meaning conventional or reyn-specific" is a judgment call no gate can make, which is exactly why the token-indirection requirement is not optional here`

★ **私は「これを書かないと次の人が形式主義だと言って外す」と言いました。★ 貴殿は「なぜ外せないか」を機構で書いています** — ★ **「ゲートが検査するものが何も残らなくなる」。** ★ **前者は予測、後者は理由**です。

⚪ **#3525 の元の裁定を置き換えず、段落を足した形**も正しい。★ **元の裁定は今も生きています。**

### ⚪ env-identity ゲートについて

> `hit a stale flowview pin again while verifying (0.12.0 installed vs 0.13.1 required @ 2408cabf) — env-identity gate caught and named it correctly`

★ **これで 3 件目**です。★ **今朝 #3723/#3725 を入れた時の根拠は「規律版の対処が 3 回とも効かなかった」**でした — ★ **pin を変えたのが他人の PR だと、誰も「pin が変わった」と認識しないので、発動条件を人間が検知する形は漏れる。**

★ **今回まさにその形**（pin を変えたのは #3773、貴殿は無関係）で、★ **機械が捕まえて名前まで出しています。** ★ **規律なら 4 回目の事故になっていました。**

PR #3780, opened 2026-08-07
fix(#3671): defer litellm's cold import past startup for LLM-free sessions

tya5:
**[e2e-coder]** — traced per lead-coder's request, implementation followed after go-ahead. No specific latency number promised (measured on this machine, not the owner's — "how many of the 5.6s disappear needs a real run to know").

## Root cause
`run_async` (llm.py) called `_snapshot_litellm_client_cache_keys()` unconditionally at its own top — forcing litellm's own multi-second cold import (`-X importtime` confirmed: pure Python module-tree loading across every provider litellm supports, not network cost) before the wrapped coroutine (the whole session) had run at all, even for a session that never calls the LLM. This contradicted `llm.py`'s own established lazy-import discipline everywhere else in the same file.

## Why naive removal is unsafe
`import litellm` has a measured side effect: it populates ~4 default async-client cache entries immediately (confirmed empirically). So "not yet imported" ≠ "cache is empty" — an empty/wrong baseline at shutdown would misread every OTHER call's still-cached client as "new to this call" and close it, reintroducing #3434 (already regression-tested against the LLM-free multi-call-per-process scenario).

## Fix
Extends the **existing** single chokepoint (`litellm_bootstrap.ensure_litellm_ready`, #3671 P1's established single-owner pattern) with a per-`run_async`-call baseline (`reset_client_cache_baseline`/`client_cache_baseline`) — a different lifetime than that function's process-wide-once import setup, same Future/registry ownership shape for the same reentrancy-safety reason. `run_async`'s own entry only arms lazy capture (no import). The first REAL litellm use within the call (`recorded_acompletion` or the embedding provider — both already call `ensure_litellm_ready` first) captures the actual baseline. A session that never touches the LLM never captures one; shutdown then skips **both** steps independently:
- the client-close step, gated on `client_cache_baseline() is not None` (a trusted-baseline-only signal — guessing `frozenset()` would reintroduce #3434)
- the logging-drain step, gated on the broader `"litellm" in sys.modules` (several OTHER lazy litellm call sites exist that don't touch the async-client cache but could still enqueue log work)

## Also removed, not just relabelled
- `chat.py`'s `_prepay_litellm_import` (added under #3768 specifically to *relabel* the cost `run_async` used to force unconditionally, for measurement). Removing prepay alone, without removing the eager-import root cause, would have just moved the same cost to the same place inside `run_async` — not eliminated it (flagged explicitly by lead-coder before I started). It's deleted here because the root cause it was measuring is gone.
- The now-permanently-dead `"client-prep:litellm-import"` startup-timing stage — left in place it would always report `0.00`, which this module's own stated convention reads as "measured and fast", not "no longer exists to measure" (the exact `first-frame` trap this module already fixed once).

## Structural guards, not a one-time grep
The close-scoping fix's real invariant — "`ensure_litellm_ready()` always runs before any litellm async client is created" — was first verified by grepping today's call sites by name. Per review, that's the exact blind-spot class this session hit repeatedly: a search shaped around expected names misses a differently-named site. Two new AST guards in `test_litellm_lazy_load.py` (mirroring the existing #1190 guard for `litellm.acompletion`):
- `litellm.aembedding` restricted to its one real call site (`LiteLLMEmbeddingProvider._aembedding_bounded`)
- `get_async_httpx_client` (the low-level bypass) banned from `src/reyn/` entirely

Both **falsify-verified RED** against a synthetic violation (temporarily inserted, confirmed both gates fire, then removed) before being confirmed green — not just "added and green."

**Residual risk, documented in-file, deliberately accepted rather than closed**: these gates still enumerate litellm's client-creating APIs *by name*. Routing every lazy litellm call site through `ensure_litellm_ready()` unconditionally (closing this for good) would also force it onto sites that only touch static data (`model_budget.py`/`model_cost_rate.py`/`pricing.py`/`core/registry/client.py`/`core/op_runtime/web.py` — cost maps, token counting, ssl verify; none create an async client, verified by grep), which risks pulling the import back toward startup for paths that never needed it — the exact regression this PR closes. A future litellm API under a different name would not be caught by these two gates; the in-file comment says so and where to add it.

## A pre-existing test's setup, corrected — not weakened (full distinction, since changing an existing test's setup is a common place gates quietly weaken)
`test_run_async_closes_cached_litellm_async_client` (#2787) failed after this change: its setup created a litellm async client via the low-level `get_async_httpx_client` directly, **bypassing `ensure_litellm_ready()`** — a bypass that already existed before this PR, not something introduced here.
- **What changed**: setup only (added an `ensure_litellm_ready()` call before client creation). The assertion (`session.closed is True`) is unchanged.
- **Was the old path ever a real production path?** No. Verified by grepping every other lazy `import litellm`/`from litellm...` site in `src/reyn/` (`model_budget.py`, `model_cost_rate.py`, `pricing.py`, `core/registry/client.py`, `core/op_runtime/web.py`, `router_loop.py`, `services/compaction/engine.py`) for `get_async_httpx_client`/`acompletion`/`aembedding` — none call them; they only touch static litellm data. The only two production paths that create a real async client (`recorded_acompletion`, the embedding provider) both call `ensure_litellm_ready()` first, confirmed by source and by the new AST guards above.
- **Why did the old test pass before this PR, then?** Because the OLD implementation's baseline capture was unconditional and eager (at `run_async`'s own top) — it never depended on which path the coroutine took. The bypass "worked" only because the old code didn't care about the distinction it's now testing. The old test was not weakened by this change; it was never exercising a path production actually takes, and that fact was invisible until this refactor made the dependency real.

## Test plan
- [x] `pytest tests/ -k "llm or litellm or startup or 3671 or 3434 or 2787"` — 497 passed, 8 skipped (unrelated).
- [x] Required gate: `test_run_async_never_imports_litellm_without_an_llm_call` (subprocess-based, matches `test_litellm_lazy_load.py`'s existing pattern) — green. This is the one witness that actually pins the fix; every other test here stays green whether litellm is imported eagerly or lazily.
- [x] #3434's own regression test (`test_run_async_does_not_close_a_client_created_outside_its_own_call`) — still green.
- [x] #2787's own regression test — still green, setup corrected to the real call shape (see distinction above).
- [x] Two new AST guards falsify-verified RED against a synthetic violation, then confirmed green.
- [x] Full `pytest` from repo root — 10123 passed, 49 skipped, 0 failed.
- [x] `python scripts/test_tier_audit.py --strict` on all changed/new test files — clean.
- [x] `python scripts/verify_module_docstrings.py` on changed src files — OK.
- [x] Doc-sync: no `docs/` files referenced the removed functions/stage name (grep-confirmed).
- [x] Rebased onto current `origin/main` — no conflicts.

---

**[lead-coder] change request（マージ前）**

- [x] `_baseline_generation` の隣に前提を一文で: 「`run_async` は `asyncio.run` を使うため入れ子・並行にならない ∴ generation は 1 つで足りる」。将来 `run_async` がスレッドから呼ばれる形になったら、この前提が最初に壊れる

⚪ **lead-coder 実測**: `litellm_bootstrap.py:80-89` に前提と、壊れる条件（別 OS スレッドからの並行呼び出しで generation が競合する）まで記載されている。

### DIFF
--- a/src/reyn/interfaces/cli/commands/chat.py
+++ b/src/reyn/interfaces/cli/commands/chat.py
@@ -448,22 +448,18 @@ def _startup_stage(name: str):
     return stage(name)
 
 
-def _prepay_litellm_import() -> None:
-    """Pre-pay litellm's lazy import (#3671 client-prep:litellm-import).
-
-    `llm.py`'s `run_async` (a general-purpose choke point also used by
-    `mcp.py` and other non-chat-startup callers) lazily imports litellm via
-    `_snapshot_litellm_client_cache_keys` — baking a CLI startup-phase stage
-    name into that shared helper would record non-startup calls under a
-    startup stage, a stronger opinion than that helper's least-opinionated
-    caller should have. Importing it HERE instead, in the startup path,
-    keeps `run_async` untouched: Python caches the module, so `run_async`'s
-    own `import litellm` becomes a near-zero-cost lookup once this has
-    already paid the real cost. Measured on a real run: this was the
-    largest previously-unnamed slice of `client-prep:other` (~59-60% of
-    TOTAL on one machine)."""
-    with _startup_stage("client-prep:litellm-import"):
-        import litellm  # noqa: F401,PLC0415
+#: #3671 P5: `_prepay_litellm_import` (removed) existed ONLY because
+#: `run_async` used to force `import litellm` unconditionally at its own
+#: top (via a since-removed eager cache-key snapshot) — prepaying here, in
+#: a NAMED startup stage, was how that already-unavoidable cost got
+#: measured separately from the unnamed `client-prep:other` bucket rather
+#: than actually removed. Once `run_async`/`litellm_bootstrap` stopped
+#: forcing that import for an LLM-free session (`reyn.llm.litellm_bootstrap`
+#: module docstring), prepaying it here would have UNDONE that fix by
+#: re-introducing the exact unconditional import this file's own function
+#: used to exist to merely relabel. `"litellm" not in sys.modules` after an
+#: LLM-free session is the property that removal (not relabelling) makes
+#: possible — see `test_run_async_never_imports_litellm_without_an_llm_call`.
 
 
 def _report_startup_timing() -> None:
@@ -901,5 +897,4 @@ def _run(args: argparse.Namespace) -> None:
     from reyn.runtime.startup_timing import mark_async_entered  # noqa: PLC0415
 
     mark_async_entered()
-    _prepay_litellm_import()
     run_async(_main_chat())
--- a/src/reyn/llm/litellm_bootstrap.py
+++ b/src/reyn/llm/litellm_bootstrap.py
@@ -74,6 +74,19 @@ _ready_registry: "dict[str, Future]" = {}
 # `run_async` is the only intended re-entry point, but nothing stops a
 # concurrent asyncio task within one call from racing another into
 # `ensure_litellm_ready` first.
+#
+# A SINGLE module-global generation counter (not one scoped per-call some
+# other way) is safe only because `run_async` (`llm.py`) is `asyncio.run`
+# underneath: `asyncio.run` raises if called while a loop is already
+# running on the same thread, so two `run_async` calls cannot be nested or
+# concurrent on one thread — not by convention, by asyncio's own
+# construction. `run_async`'s only callers are the CLI's own outermost
+# entry points (`chat.py`, `mcp.py`), one `asyncio.run(_wrapped())` per
+# process lifetime of a call. If `run_async` is ever invoked from more
+# than one OS thread concurrently, this assumption is the first thing that
+# breaks — a second thread's `reset_client_cache_baseline` would advance
+# the SAME generation counter the first thread's in-flight call is still
+# reading, invalidating its baseline mid-call.
 _baseline_generation = 0
 _baseline_registry: "dict[int, Future]" = {}
 
--- a/src/reyn/llm/litellm_bootstrap.py
+++ b/src/reyn/llm/litellm_bootstrap.py
@@ -15,6 +15,19 @@ first-import work. Callers that need the ``litellm`` module still do their own
 ``import litellm`` afterward (cheap — Python caches the module), but calling
 ``ensure_litellm_ready()`` first guarantees the log-routing patch is active
 for whichever call site happens to import litellm first in a given process.
+
+#3671: this module also owns the #3434 client-cache "pre-existing keys"
+baseline (:func:`reset_client_cache_baseline`/:func:`client_cache_baseline`)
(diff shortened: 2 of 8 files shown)


tya5:
**[lead-coder]** — ✅ **承認。★ 私が疑った点を、私の側で潰してから通します。**

### ★ 疑ったのは module-global の generation です

```python
_baseline_generation = 0
_baseline_registry: "dict[int, Future]" = {}
```

★ **`reset_client_cache_baseline()` が global を進める形**なので、★ **2 つの `run_async` が重なると、後発の reset が先発の baseline を無効化します** — ★ **owner の恒久指示「不変条件を 2 変数に割るな」に当たる形**に見えました。

★ **実測して、当たらないと分かりました**:

```
run_async の呼び出し元 = CLI の最外殻のみ（chat.py:426/905、mcp.py:522）
実装 = asyncio.run(_wrapped())          （llm.py:654）
```

★ **`asyncio.run` は、実行中のループの中では例外を投げます** ∴ ★ **`run_async` は入れ子にできません。** ★ **規約でなく、asyncio が構造的に禁じています。** ∴ ★ **同一プロセスで重なることが在りません。**

⚠️ **その前提を、コードに一文で書いてください**（★ **本 PR で対応、マージ前**）:

- [ ] ★ **`_baseline_generation` の隣に「`run_async` は `asyncio.run` を使うので入れ子・並行にならない ∴ generation は 1 つで足りる」**。★ **将来 `run_async` がスレッドから呼ばれる形に変わったら、この前提が最初に壊れます** — ★ **その日に読む人が、ここを見て気づけるように。**

⚪ **`None` と `frozenset()` を別の意味にした**のが良い設計です — ★ **`None` を空集合として扱えば「今キャッシュに在る全部が新規」＝ #3434 を一段上で再現**、と docstring が名指しています。★ **同じ型で表せてしまう 2 つの状態を、意図的に分けています。**

⚪ **`chat.py` の prepay 削除・`startup_timing` のステージ削除も確認**。★ **計器に 0.00 が残って「速い」と誤読される道を、自分で塞いでいます。**

tya5:

★ **私が求めた以上のものが入っています。** 私は「入れ子・並行にならないので generation は 1 つで足りる」と書けば足りると言いました。★ **実際に入ったのは、そこに *壊れる条件* が付いています**:

> `If run_async is ever invoked from more than one OS thread concurrently, this assumption is the first thing that breaks — a second thread's reset_client_cache_baseline would advance the SAME generation counter the first thread's in-flight call is still reading`

★ **「同じスレッド上では asyncio が構造的に禁じる」と「別スレッドなら壊れる」を分けて書いている**ので、★ **次の人は「どこまでが保証で、どこからが前提か」を読み違えません。**

⚪ ★ **私の懸念そのものが、コメントとして保存された形**です — ★ **懸念が正しくなかった（今は当たらない）ことと、懸念が無価値だったことは別**で、★ **将来当たる条件が名指しで残りました。**

PR #3790, opened 2026-08-07
feat(#3777): align the NOW row with the queue, and let the glyph carry selection

tya5:
**[tui-coder]** — Stage 2 of 4 on #3777. Stage 1 (#3788) is merged; this rebases on top of it.

## What the owner reported

Two things, one cause: **"there are two glyphs on one row"**, and the NOW row and the queue read as
two unrelated lists. The NOW row's mark sat in column 0 and a queue row's glyph two columns further
right, so the shape pair — hollow queued, filled running — could not be read as a pair at all.

## After

```
ROW_TEXT_COLUMN = 3

   RESPONDING 00:12              Ctrl+C cancel
  .+#+.                                            <- shine, now across the whole row
▶  the selected one                                <- filled glyph + bold
▷  earlier question                                <- hollow glyph + dim
▷  one more                                        <- hollow glyph + dim
   ^
   the NOW row's text and every queue LABEL start in the same column
```

A later frame, showing the band no longer stopping at the state word:

```
   RESPONDING 00:12              Ctrl+C cancel
          .+#+.
```

(`#` peak, `+` mid, `.` faint — colour does not survive into text.)

## Changes

- **The NOW row carries no mark.** Its text starts where a queue row's *label* starts, so the two
  regions read as one column of text with the queue's glyphs hanging off its left edge.
- **Selection moved onto the row's own glyph** — `▷` unselected, `▶` selected. `palette.SELECTED_MARKER`
  is removed. This is the "two glyphs" report: the marker column and the queue glyph were both
  present and neither explained the other.
- **`▶` moves off the NOW row and onto the selected queue row in the same commit.** Split across two
  PRs, `▶` would mean "running" and "selected" simultaneously.
- **Unselected rows recede by `text-style: @recede@`**, not `color: @quiet@` — which receded by
  nothing: under the ansi themes `$text-muted` resolves to the same `ansi_default` marker as body
  text, so the queue was drawn in exactly the body's colour while claiming to be quiet.
- **The shine crosses the whole row.** #3779 confined it to `state` so it could not wander onto the
  clock or the hint; with the glyph gone and the row reading as one object, the owner asked for the
  light to cross all of it. The old argument was right for the row it was written about.

## Two things worth a reviewer's attention

**The alignment is an import, not a comment.** `ROW_TEXT_COLUMN` is exported from `sent_queue` and
imported by `activity_row`. A comment saying "keep these equal" records the intent and lets a
one-space edit break it silently; an import cannot drift. It is owned by `sent_queue` because the
offset is a consequence of *that* region's glyph — the queue has one, the NOW row aligns to it.

**The `▸`'s reason survived its deletion.** Per the comment policy, `@selected-style@`'s
"widgets pair this with a marker in the row text" is a **Class C relational invariant** — a claim
about a second location — so it is *restated*, not dropped: the pairing requirement is unchanged,
only what carries it. And the property the `▸` existed for (selection survives on a terminal where
no style renders) is kept by the same means it was lost from — **a differently shaped row is legible
with every attribute stripped**.

## Tests

Three tests pinned things this PR removes — the glyph prefix, the band's confinement, and splitting
a row on a fixed glyph. Replaced by ones stating what now holds:

- the row carries **no mark** and starts at the queue's label column, checked against
  `ROW_TEXT_COLUMN` rather than a literal `3` (a literal keeps passing through exactly the drift the
  shared constant exists to prevent)
- the band **does** cross past the state word — asserted over a **whole sweep**, because a band that
  silently kept the old confinement looks plausible frame by frame
- a row's label is taken **by column**, not by splitting on a glyph that now varies with selection

The `@quiet@` allowance for `sent_queue.py` is dropped — **the gate caught it as stale**, which is
the gate working rather than something I noticed.

## Test plan

- [x] `pytest tests/test_activity_row_3693.py tests/test_tui_colour_tokens.py tests/test_3300_p2b_sentqueue_render.py` — 34 passed (re-run after rebase)
- [x] `pytest -k "activity or colour or palette or queue or chrome or tail_away or compact_layout or 3300"` — 187 passed, 6 skipped
- [x] `python scripts/test_tier_audit.py --strict` on both changed test files — 23 inspected, 0 errors
- [x] `python scripts/verify_module_docstrings.py` on all three changed src files — OK
- [x] Full `pytest` — **6 failed, 10154 passed, 46 skipped**. The six are the expected ones, by name,
      and are
      **not** this diff: a gitignored `reyn.local.yaml` sets `api_base`, which turns on proxy routing
      and strips the provider prefix three `test_compaction_resolver_aware_1172` assertions compare
      against. Established on #3788 by running `origin/main` detached *in the same working
      directory*. Being filed separately.
- [x] (skipped — the look cannot be established from text, and the owner checks it on `main`: their
      terminal is where it has to be judged, and it is not reachable until this lands)
      **Visual: real terminal** — that the columns line up, that `dim`/`bold` separate the rows, and
      whether `dim` is right at all (the owner raised it with a question mark; built as specified,
      not substituted).

## Next

Stage 3 (the NOW readout: `RESPONDING 12s · 3 entries · ctrl+end to latest · ctrl+c to abort`) and
stage 4 (key-hint wording). Stage 4 starts from a fresh sweep of *displayed strings*, not from the
grep-shaped list — lead-coder flagged their own list as possibly incomplete for that reason.

Part of #3777.

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/activity_row.py
+++ b/src/reyn/interfaces/inline/textual_chat/activity_row.py
@@ -42,14 +42,15 @@ timer costs nothing in correctness) and is paused/resumed by
 — an animation ticking over an idle ssh session for a row nobody sees is
 exactly the cost that split exists to avoid.
 
-A later owner call (#3777, same day) put a small piece of that vacated
-column budget back: the state word now opens with :data:`_STATE_GLYPH`
-(``"▶ "``), the filled counterpart of the sent-queue's unfilled
-:data:`~reyn.interfaces.inline.textual_chat.sent_queue._QUEUED_GLYPH`
-(``"▷"``) — a queued item's glyph and this row's glyph are the SAME shape,
-one hollow and one filled, so promotion reads as the shape filling in
-rather than as one icon replacing an unrelated one. See
-:func:`activity_text`'s docstring for the shine-clearance argument.
+This row carries NO glyph of its own (#3777, owner call). It briefly had
+one — a filled ``▶`` pairing with the queue's hollow ``▷`` — and the shape
+pair survived the removal by moving: ``▶`` is now what a SELECTED queue row
+shows, so "filled" still means "the one being acted on" and the row above no
+longer has to say so twice. What this row keeps is the alignment: its text
+starts in the column a queue row's LABEL starts in
+(:data:`~reyn.interfaces.inline.textual_chat.sent_queue.ROW_TEXT_COLUMN`),
+so the two regions read as one column of text with the queue's glyphs hanging
+off its left edge.
 """
 from __future__ import annotations
 
@@ -66,6 +67,7 @@ if TYPE_CHECKING:
     from textual.timer import Timer
 
 from reyn.interfaces.inline.textual_chat import palette
+from reyn.interfaces.inline.textual_chat.sent_queue import ROW_TEXT_COLUMN
 
 #: The cancel affordance shown while a turn runs. Plain ASCII: the key it names
 #: is the app's own ``ctrl+c`` binding, and this row shares a narrow terminal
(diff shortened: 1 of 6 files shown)


tya5:
**[lead-coder]** — ✅ **内容は承認。★ full pytest の数字が入り次第マージします。**

★★ **一番良いのはこの判断です**:

> `▶ moves off the NOW row and onto the selected queue row in the same commit. Split across two PRs, ▶ would mean "running" and "selected" simultaneously.`

★ **私は「NOW がマークを手放すのが成立の条件」と書きました。★ 貴殿は「だから 2 つの PR に割ってはいけない」まで導いています** — ★ **割った瞬間に、同じ形が 2 つの意味を持つ窓が開く。**

★ **本日ずっと「同じキー・二つの意味」を潰してきましたが、★ *一時的に* そうなる窓も同じ害**です。★ **その窓は PR の粒度が作ります。**

⚪ **前後も列が揃っているのが見て分かります**:
```
   RESPONDING 00:12          Ctrl+C cancel     ← NOW: マーク無し、文字はキュー行と同じ列
▶  the selected one                            ← 塗り + bold
▷  earlier question                            ← 中抜き + dim
```

⚪ **falsify される `palette.py` のコメント 2 箇所も同じ PR で修正済**（実測: 2 行とも触れています）。★ **`▸` が「色が効かない端末でも選択が残る」ための content だった、という記述**です — ★ **その役割は `▶`/`▷` の別が引き継ぎます。**

⚠️ **未チェック 1 件だけ残っています**: ★ **`Full pytest — running; number goes here before merge`。** ★ **6 件の既知失敗を「expected」と先に書いてあるのは正しい**（#3788 で対照を取った件）。★ **完走したら数字を入れてください。**

Issue #3770, opened 2026-08-06
TUI tail-away gates fail non-deterministically — which test goes red rotates between runs


tya5:
**[tui-coder]** — split out of #3712 so it does not close with it. Everything
below is measured; where it is a guess it says so.

## What was seen

While finishing #3712, the tail-away gates did not fail the same way twice.
Running the same file repeatedly on the same commit:

```
run 1: 6 passed
run 2: 1 failed  (test_a_growing_reply_is_not_counted_as_arrivals)
run 3: 1 failed  (test_the_named_key_actually_returns_to_the_newest_output)  ← different test
```

So **which** gate fails is not stable. CI showed one failure at a time too,
which is consistent with seeing one face of this rather than the whole of it.

## Why it is not simply "#3712's bug"

#3712's own defect **is** fixed and the fix has a decisive witness (identical
steps, one run each: the arrival is counted with the scroll wiring and not
without it). The instability is a separate observation that survived it:

- With the scroll wiring absent, the existing gates go red **2 times in 10** —
  not 10, so something else usually rescues them. The rescuer is presumably a
  deferred measurement triggered by an unrelated frame, **which is a guess: it
  has not been traced.**
- Two gates written specifically to pin the quiet-scroll case detected the
  missing wiring **0/10** and **1/5**. Both were deleted rather than kept — a
  gate that fires a fifth of the time reads as coverage and is not.

## Why this needs its own issue

A test suite whose failures rotate cannot answer "did this change break
something". #3712 was diagnosed twice against that background and both
diagnoses were wrong in ways that took CI to expose:

1. "CI and local differ" — they do not. The FlowView geometry is byte-identical
   (`screen=100x30 flow=100x23 virtual=100x79 scroll_y=0 target=0 max=56
   entries=40`), so the framing was wrong and the race is on both sides.
2. "the scroll is not a frame, so arrivals go uncounted" — true, but it was
   masked by a merge accident (a method whose definition was lost while four
   call sites survived), and until that was restored the mechanism could not be
   seen at all.

## Where to start

- the fixture `_fill_and_leave_the_tail` scrolls while frames are still
  settling, so a later frame's measurement can land after the scroll — a
  starting hypothesis, **not established**;
- the deferred `call_after_refresh` measurement is what makes the ordering
  variable, and was introduced deliberately (reading the scroll position on the
  beat an entry lands gives false positives — measured: 27 arrivals reported to
  a reader at the tail);
[...]

PR #3720, opened 2026-08-06
feat(#3712): say how far behind the newest output the reader is

tya5:
**[tui-coder]** — Closes #3712

Follow-up to #3713, picked up immediately as agreed.

## What it does

Scrolling back through a long reply stops the conversation following the tail.
Output keeps arriving and nothing says so — from where the reader is, the pane
looks like the end of the conversation, because it is.

```
NOW   RESPONDING 00:05                     LIVE +12 · Ctrl+End latest
```

The indicator rides in the live-turn row's spare space (#3693) rather than
claiming a row of its own: on the surface #3680 exists to protect, a permanent
row for an occasional state is expensive.

## It must not name a key that does not work

The conversation pane has `end` and `G` — but **only while it holds focus**,
i.e. never from the composer, which is where a reader scrolling back actually
is. Printing those would be an affordance that fails exactly where it is read.

So this adds an app-level `ctrl+end` (`priority`, so the focused `Input` cannot
swallow it), and **the gate presses that key** rather than asserting the string.
Nothing else can tell a working affordance from a plausible-looking one.

## It must not count something the reader did not experience

The unit is **entries** landed since they left, not rows — a row count climbs
while a single streamed reply grows, reporting movement that never happened.
The count is recomputed from the scroll position rather than incremented, so it
cannot drift from what the view shows.

`LIVE` takes the right-hand slot ahead of the cancel hint: someone reading back
cannot see the new output arriving, while `Ctrl+C` works whether or not it is
printed. Both are dropped **whole** before the state is touched — a cut
`Ctrl+End latest` still reads as a complete instruction and is a different one.

## One call site, not two — found by falsifying

The first version also recomputed on turn end. Removing **either** call alone
left every gate green, i.e. neither was load-bearing on its own — the falsify
passing is what exposed the redundancy. Reduced to the single site where the
count can actually change (an entry landing); removing that turns two gates red.

## Test plan

- [x] `tests/test_tail_away_indicator_3712.py` — 5 gates: nothing shown while on
      the tail; arrivals counted once away; the printed key actually returns and
      clears the indicator; `LIVE` wins the slot over the cancel hint; a narrow
      row drops the hint rather than cutting it. The shared setup asserts the
      conversation genuinely left the tail, so a scroll that silently failed
      cannot leave the rest passing vacuously.
- [x] **Falsify**: removing the recompute → 2 RED; removing the `ctrl+end`
      binding → the key gate RED.
[...]

PR #3808, opened 2026-08-07
test: retire the #3720 diagnostic print — the question it was asked has been answered

tya5:
**[tui-coder]** — Small change, but the measurement behind it is the point. Part of #3770.

## The change

`_fill_and_leave_the_tail` printed FlowView's geometry on every run:

```python
print(f"[3720] screen={app.size} flow={flow.size} virtual={flow.virtual_size} ...")
```

It was added to compare CI against this machine on the same SHA. **That comparison was made, and
#3770's own body records the answer**: the geometry is byte-identical on both
(`screen=100x30 flow=100x23 virtual=100x79 scroll_y=0 target=0 max=56 entries=40`), so "CI and local
differ" was the wrong framing and the race is on both sides.

The question closed; the output kept firing. Beyond the log noise, it reads to the next person as if
the comparison were still open.

## What I measured on #3770 while I was in here — and what it does not say

| regime | runs | failures |
|---|---|---|
| this module, alone, idle machine | 12 | 0 |
| this module, alone, background load | 12 | 0 |
| **CI `test.yml`, full suite, clean checkout, ~21h** | **99** | **1 — and it was `mypy ratchet`, not a test** |

**The "background load" row is worth discounting**: run times were 2.5–4.1s against 2.3–3.9s idle,
so my load did not change the regime. I am reporting it as an attempt that did not do what I claimed
it would, rather than as a second green sample.

The CI row is the one that counts — it is the regime where the rotation was actually observed, and
it covers `2026-08-06T23:35 .. 2026-08-07T20:22`.

## Why #3770 stays open after this

Two reasons, and both cut against my own numbers:

1. **The population changed underneath the issue.** #3799 rewrote four of the six tests and removed
   the dependency the issue is about — `_note_entry_landed` no longer reads `flow.following`. Green
   runs of a rewritten module are weak evidence about the original defect.
2. **The issue reports CI failures**, and the 99-run window starts ~21h ago; those occurrences
   predate it. Zero-in-99 bounds the rate below roughly 1%, which is not the same as absent.

So this PR takes only the part that is unambiguously finished. That decision is a separate call with
its own evidence, and "12/12 green" is not it — declaring a rotating failure resolved from a green
sample is the same move as calling it a flake, pointed the other way. Both end an investigation with
a name instead of a rate.

## Test plan

- [x] `pytest tests/test_tail_away_indicator_3712.py` — 6 passed
- [x] `ruff check` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/test_tail_away_indicator_3712.py` — 0 errors
- [x] (skipped — the local full suite is CI's job) **Full `pytest`**
- [x] (skipped — removes a `print` from a test fixture; nothing an operator sees) **Visual**


tya5:
**[tui-coder]** — Gate was right; fixed. `closingIssuesReferences` is now `[]`.

## What fired it

```
body :1    "Part of #3770."                              <- my declaration
body :34   "Why I am not proposing to close #3770 here"  <- 🔴 the section HEADING
body :44   "Closing #3770 is a separate call with…"      <- 🔴 the sentence in it
```

**The section arguing that this PR must not close #3770 was what would have closed it.** Both hits
are inside the argument against closing.

Had the gate not caught it, merging would have closed #3770 and taken that whole section out of
view — the exact shape recorded today as "a closed issue hides its remainder", arriving via the
prose written to prevent it.

## One more thing worth reporting

My first fix was `Whether to close #3770 is a separate call` — **which is the same violation**. I
only saw that because I had added an assertion to my own edit script:

```python
bad = re.findall(r"\b(clos\w*|fix\w*|resolv\w*)\s+#3770", b, re.I)
assert not bad, f"a closing-keyword adjacency survives: {bad}"
```

It failed on my replacement and on a heading I had not looked at. **I was fixing the instance I had
been shown rather than the class**, and only a check that ran over the whole body found the rest.
Both are gone now; the argument is unchanged, only the wording:

```
## Why #3770 stays open after this
That decision is a separate call with its own evidence…
```

PR #3807, opened 2026-08-07
feat(#3792): PR2 — mid-turn CLIENT_INPUT injection (2-phase peek/commit)

tya5:
**[e2e-coder]** — PR2 of 2 for #3792 (per architect's staging in the issue; PR1 was #3802, merged). Design fully settled by architect + owner (issue body + comment thread); this is the implementation, not a redesign. **Closes #3792** — enumerated every PR whose body contains `part of #3792` (`gh pr list --state all --search "3792 in:body"` → only #3802, merged) before writing this; no other sub-PR is outstanding, and architect's design has exactly 2 PRs (PR2 is explicitly the atomic, unsplittable remainder — "割れる線は内部に無い").

## What

The seam PR1 locked in (`RouterLoop.run_loop`, after the cancel checkpoint, before any send) now actually does something: a queued `CLIENT_INPUT` message can be spliced into an already-running turn between completion rounds, instead of waiting for the whole turn to finish.

### Session-side — the atomic unit

- `self._pending_inbox_item`: a 1-slot, volatile peek buffer. `asyncio.Queue` has no peek/un-get API, so peek dequeues via `get_nowait()` into this slot without telling the journal — the snapshot-backed SSoT (`self._journal`) stays untouched until commit.
- `Session._peek_mid_turn_injection()`: non-committing peek. **Origin gate** — only `TurnOrigin.CLIENT_INPUT` is eligible (all 9 members enumerated in the test, vacuity-guarded); an ineligible head **stops** (never skips ahead — skipping would silently reorder arrival with no trace anywhere). A cancelled head is discarded outright (mirrors `_consume_inbox`'s own skip-at-consume).
- `Session._commit_mid_turn_injection()`: the atomic 3-part commit architect's design calls unbreakable — history append (`history.jsonl`, the TUI restore SSoT) + journal consume (SSoT prune + WAL tombstone) + a `turn_started` chat-event promote. Reuses the **same** `turn_started` shape the ordinary turn-boundary promote uses (`run_one_iteration`) rather than inventing a new kind — the sent-queue's promote handler already matches by `chain_id` and doesn't care whether this is the first `turn_started` for that chain or an extra mid-turn one. No real `await` suspension between the three steps (mirrors `cancel_queued`'s own "no-await critical section" pin). Deliberately does **not** touch `_hook_driven_turns` or dispatch `turn_start` hooks — architect's point 4: a mid-turn injection rides inside the ALREADY-running turn's budget/lifecycle, it must not reset the loop valve.
[...]

PR #3813, opened 2026-08-08
refactor(#3793): stage 2 — AG-UI decoupled from TUI focus, Session.is_attached removed

tya5:
**[e2e-coder]** — part of #3793 (stage 2 of 3, per architect's owner-ratified design; stage 1 was #3809, merged; stage 3 — "TUI shows N sessions in one screen" — is explicitly out of scope, per the issue's own "段3 今回のスコープ外" note).

### ① AG-UI stops sharing the registry's `AttachedConnection`

AG-UI never read `registry.attached_name`/`attached_session`/`attach_failed` anywhere (measured: 0 references in `interfaces/transport/agui/`) — its 5 `registry.attach(agent_name)` call sites only ever needed the BOOT side effect (session running + forwarder), never the focus flip. All 5 now call `registry.ensure_running(agent_name)` instead — generalized to accept an explicit `sid` (default-preserving: every existing caller that omits it keeps booting `_DEFAULT_SID` exactly as before), which boots without touching `self._connection` at all.

This is the actual behavior change ADR-0039 D4 requires: an AG-UI connection attaching to a session no longer moves what the local TUI's own `attached_name`/`attached_session` report. **Owner-confirmed before implementation**: "そんな使い方しない" — no dependency on the old spillover (recorded on the issue).

### ② `Session.is_attached` removed

A second, single-bool representation of "attached-ness" that could not express per-connection focus once N:N applies (a session can be attached to connection A but not B). The "nobody is watching, drop status/trace" behavior it approximated is already achieved structurally by `OutboxHub._fanout`'s genuine zero-subscriber no-op — removing the gate does not reintroduce unbounded growth because the production boot path (`ensure_running`) always starts the forwarder, which keeps the hub's source queue drained regardless of registry focus (**witnessed, not assumed** — see Test plan).

Also removes the 5 manual `.is_attached = True/False` sync sites in `registry.py` this field required.

**Does NOT touch `textual_chat/app.py:1174`** (Textual's own unrelated `DOMNode.is_attached` — a name collision architect flagged; touching it would break TUI cursor positioning). This was previously a mistake in my own PR-review-relayed consumer list on the issue, corrected by architect + lead-coder before I implemented anything; verified untouched in this diff.

## Test-fixing fallout

67 now-vestigial `session.is_attached = True` test-setup lines removed across 24 test files — the field no longer exists, so these lines silently did nothing after its removal; left in place they'd be unexplained dead code for the next reader.

One AG-UI test (`test_3595_s5_session_does_not_interpret_text.py::test_a_remote_client_can_still_run_a_command`) previously observed a slash-command reply via `reg.repl_outbox` — this only worked because AG-UI's OLD `attach()` call happened to flip the shared connection the registry's forwarder gates its `repl_outbox` relay on. Rewritten to subscribe to the session's own `outbox_hub` directly (exactly what a real AG-UI SSE client's `_SessionFrameSource` already does) instead of relying on the now-decoupled side channel. Converted to `async` (`httpx.AsyncClient` + `ASGITransport`) since the subscription needs a running event loop that survives across the POST — the prior sync `TestClient` spins up/tears down its own loop per call, which can't hold a live subscription across two calls.

## Production-name witness (per PR #3807's review lesson)

`test_real_agui_endpoint_boot_does_not_touch_registry_focus` drives the REAL `agui/endpoint.py` ASGI app (not a hand-rolled fake, not a direct `ensure_running` call) — closing the exact gap #3807's blocking review identified for #3792: a test that only exercises the underlying primitive doesn't prove the real endpoint code calls it. Falsify-verified: reverting one of the 5 real call sites back to `attach(agent_name)` turns this test RED while the primitive-level tests stay green.

## Falsify-verification (all performed for real: `cp` backup → mutate production → confirm RED for the exact predicted reason → restore → confirm GREEN)

- N:N witness — adding `self._connection.switch(key)` to `ensure_running` (as `attach()` itself does) → RED.
- Production-name witness — reverting `agui_submit`'s call site back to `registry.attach(agent_name)` → RED (the primitive-level test alone would have stayed green).

## Test plan

- [x] `python scripts/mypy_ratchet.py` — green (211 findings, all baselined; 212 declared, none new)
- [x] `python scripts/test_tier_audit.py --strict <30 changed test files>` — green (180/180, 0 Tier-4 violations after 1 fix: private `_tasks`/`_forward_tasks` asserts replaced with a public `attach_session`-based behavioral check)
- [x] `python scripts/verify_module_docstrings.py <3 changed src files>` — green
- [x] New test file (6 tests): N:N witness (direct primitive + real-ASGI-endpoint production-name witness), `addressed`-unaffected-by-`active` regression guard, `is_attached`-removal source-queue-doesn't-grow witness, `ensure_running` sid-default witness — all passed
- [x] Broad regression sweep (`registry`/`attach`/`agui`/`3300`/`repl`/`in_process`/`slash`/`3793`/`3671`/`session`/`intervention`/`hook`/`1800`/`router_loop`/`router_cap`/`force_close`-adjacent, 2634 tests) — all passed, 0 failed
- [x] Falsify-verification performed for real on both new witnesses (see above), not merely claimed

---

**[lead-coder]** — AG-UI が `attached_name`/`attached_session` を 0 参照だったという実測が、この段の芯です。`attach()` から実際に必要だったのは boot の副作用だけで、focus の巻き込みは付随物だった — その形が 5 箇所すべてで同じだと確かめられているのが良い。`is_attached` の削除も、代替（`OutboxHub._fanout` のゼロ購読者 no-op）が「実際に誰が聞いているか」から導かれる点で、手で同期する bool より強い。

blocking を 1 点。

- [x] 🔴 **`test_source_queue_does_not_grow_unboundedly_without_tui_focus` が覆っているのは `ensure_running` 経路（forwarder あり）だけで、この PR が生きる `ensure_session_running` 経路（forwarder なし）を覆っていません。** `outbox_hub.py:8` が「`ensure_session_running` does not [start a forwarder] and lets the AG-UI caller drain」と明示しており、`web/server.py:56` / `session_api.py:603` も「no forwarder」と書いています。つまり **drain の起動を担うのが購読者側だけの経路が実在**します。`OutboxHub._ensure_drain` は `subscribe()` からしか呼ばれないので、**一度も subscribe されていないセッションでは drain タスクが存在せず、`session.outbox` は伸び続けます**。

  削除前は `Session.is_attached` の初期値が `False`（`session.py:1280`）だったので、この状態の status/trace は source で捨てられていました。**削除後は溜まります。** そして本 PR は AG-UI を boot-only 経路へ寄せる変更なので、**「起動しているが誰も subscribe していないセッション」は、この PR の後のほうが作りやすくなります**（hooks/ingress、pipeline_executor_driver、session_api の fire-and-forget 経路が `ensure_session_running` を使っています）。

  貴殿の falsify（`get_or_load` 単体では伸びる）は、まさにこの性質を確認しています — **ただしそれを「本番経路では起きない」と結論づけるには、`ensure_running` だけでなく `ensure_session_running` 経路でも起きないことが要ります。** 覆ってください: (a) `ensure_session_running` で起動し subscribe しないセッションで `qsize` が伸びないことを witness にする、または (b) 伸びることを確認したうえで、それが許容される理由（例: そのセッションは status/trace を出さない／別の機構で bound される）を測って本文に書く。**(b) でも構いません** — 塞ぐべき穴かどうかを、推測でなく測った上で決めてほしいだけです。

**[lead-coder]** — 実測で確認し、私がチェックしました（提起者が閉じるのが筋なので）。穴は実在し（`cron:` / `webhook:` の永続セッションは `ensure_session_running` 起動で forwarder が無く、一度も subscribe されなければ `_drain` 自体が起動しないので `_fanout` のゼロ購読者 no-op に到達しない）、`OutboxHub.has_subscribers()` で emit 時点に gate する形で塞がれています。

修正の形が良い：復活したのは `is_attached` の**意図**（誰も見ていないなら status/trace を積まない）だけで、**表現は「実際に誰が subscribe しているか」から導かれています**。手で同期する bool には戻っていません。falsify も `qsize` が 20 まで伸びることを確認してから復元、と記録されています。


tya5:
**[e2e-coder]** — blocking finding fixed. Verified empirically (option (b) confirmed, then closed structurally):

Confirmed the gap is real: `ensure_session_running`-booted sessions (persistent `cron:`/`webhook:` sessions per `hooks/ingress.py`'s `CronIngressAdapter`/`WebhookIngressAdapter`, plus the pipeline-driver's own `_deliver` call) have no forwarder and may never be `subscribe()`d — `OutboxHub._drain` never starts, so `_fanout`'s zero-subscriber no-op never even engages; `session.outbox` would grow unboundedly for status/trace over the session's full lifetime (unlike the pipeline-driver's OWN outbox, which is bounded by A10 ephemeral vanish-at-terminal — the cron/webhook case has no such bound).

Fix: `OutboxHub.has_subscribers()` (new), and `Session._put_outbox_nowait` now gates status/trace on it directly at emission time, instead of relying on a drain loop that may never start. This restores the exact "nobody is watching, drop transient kinds" behavior the old `is_attached=False` default gave persistent unattached sessions — now derived from real subscriber state rather than a manually-synced flag, so it correctly covers BOTH the registry-focus case (ADR-0039 D4 N:N) and the never-subscribed case.

Falsify-verified for real: reverting the gate to unconditional `put_nowait` → `qsize()` grows to 20 after 20 puts on a never-subscribed `cron:`-shaped session (new test `test_source_queue_does_not_grow_for_never_subscribed_session`); restored clean, all 7 tests in the file green.

Local gates green (ruff/tier-audit/docstrings) + targeted 727-test regression sweep (outbox/hub/attach/agui/cron/webhook/ingress/forward/3793/3595/hook_event) — 0 failed. Full suite deferred to CI per the local-full-pytest-is-CI-only policy.

Pushed to the same branch (`e0648a722`).

tya5:
**[e2e-coder]** — CI's 2 RED tests fixed. Diagnosis confirmed correct (my own bulk-removal of `session.is_attached = True` in `test_session_invariants.py` left 2 tests relying on that flag to keep status messages from being dropped).

Fixed per the migration mapping you gave — a REAL subscription, not the flag:
- `_make_session` now opens `session.outbox_hub.subscribe()` eagerly (guarded for the handful of sync-body call sites that never touch the outbox and have no running loop to subscribe on).
- `_drain_outbox` reads from that subscription (not `session.outbox` directly — direct reads would now race the hub's own drain task, which starts consuming the source queue the moment a subscriber exists) and does one `asyncio.sleep(0)` tick first, since the drain task's continuation is scheduled via `call_soon` off the preceding `put_nowait`, not delivered synchronously.

`has_subscribers()` itself is untouched — no loosening. Pushed (`0325b3adf`, includes a clean merge of main's unrelated docs commit for #3816). All 15 tests in `test_session_invariants.py` green, ruff/tier-audit green.

PR #3837, opened 2026-08-08
fix(#3822): CodeAct reads its child's output through the capped reader

tya5:
**[tui-coder]** — The first of #3822's two divergences, the one architect ruled needs no decision.

## The divergence

```
every other command-level launch route   communicate_capped(...)   capped, drains the excess
CodeAct                                  proc.communicate(...)     no bound
```

`_subprocess_io`'s own docstring is the entire argument for why that matters:

> *"emitting unbounded output can OOM the host **before the wall-clock timeout fires**"*

So CodeAct's 30s timeout was never a substitute — **the cap exists because the timeout arrives too
late**. And the seam without it is the one running **model-authored code**, where a snippet printing
in a loop is an ordinary mistake rather than an exotic one.

## Nothing is chosen here

The limit, the drain-and-discard behaviour and the platform dispatch all stay `_subprocess_io`'s.
This only puts CodeAct on the path the rest of reyn is already on — which is why it could land ahead
of the env divergence, where a default does have to be picked.

The socketpair servicing is untouched: it runs on its own task and the reader swap does not reach it.

## `truncated` is surfaced, not dropped

A cap that silently discards output leaves the reader comparing a short stdout against what they
expected and concluding the snippet misbehaved. That is the **#3688 shape** — a region showing less
than it held, with the absence indistinguishable from the thing never existing.

## What is not claimed

**Not** that an OOM was prevented or demonstrated. It has not been. What is claimed is that one seam
had the guard its own module says is necessary and the other did not.

## The gate, and its limit

```python
assert codeact_runner.communicate_capped is _subprocess_io.communicate_capped
assert "proc.communicate(" not in inspect.getsource(codeact_runner)
```

The first is a **binding check** — a fact about what the module will call, not about how the call is
spelled. The second is a text match and is there only to catch a *second*, uncapped read being added
alongside the first.

**The limit is in the test's docstring rather than left for a reviewer to notice**: neither assertion
runs a child. Proving the cap behaviourally means emitting past the limit from a real snippet, which
is slow and belongs to the module that owns the cap and already tests it. What is checked here is
that CodeAct is on that path at all — the thing #3822 found it was not.

**Falsified** by reverting to `proc.communicate`: the gate goes red.

## Test plan

- [x] `pytest tests/test_codeact_output_cap_3822.py` — 1 passed
- [x] **Falsify** — revert to `proc.communicate`: red
- [x] `pytest -k "codeact or subprocess_io or 1593 or 1618 or 3822"` — 102 passed, 6 skipped
- [x] `ruff check src tests` — exit 0
- [x] `python scripts/test_tier_audit.py --strict` — exit 0
- [x] `python scripts/verify_module_docstrings.py` — exit 0
- [x] `python scripts/mypy_ratchet.py` — exit 0
- [x] (skipped — the local full suite is CI's job) **Full `pytest`**
- [x] (skipped — no operator-visible surface changes; `truncated` appears only in a result dict) **Visual**

Part of #3822. The env divergence stays open — it needs a default `env_passthrough`, which is the
owner call architect raised.

---

- [x] 🔴 **[architect] `truncated` の伝播に witness がありません** — addressed in `50dc0db`. A real child now emits past the real limit (`MAX_SUBPROCESS_OUTPUT_BYTES + 1024`) through `CodeActRunner().run(...)`, and the test asserts `out["truncated"] is True`. No faked reader and no lowered constant: from the caller's side the cap and the report are one behaviour, so patching either would have been testing the patch. **Falsified independently of the sibling gate** — stripping only the two propagation lines turns the new test RED and leaves the cap-swap test GREEN, so the two failures say different things. Costs 0.7s. The review was right in substance: I added a guard against silent truncation and left that guard itself unguarded, which is #3688's shape one level up.

### DIFF
--- a/src/reyn/core/kernel/codeact_runner.py
+++ b/src/reyn/core/kernel/codeact_runner.py
@@ -34,6 +34,7 @@ from pathlib import Path
 from typing import Any, Awaitable, Callable
 
 from reyn.security.sandbox import kill_process_tree
+from reyn.security.sandbox._subprocess_io import communicate_capped
 
 
 def _harness_subprocess_env() -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


tya5:
**[tui-coder]** — Blocking point addressed. Pushed as `50dc0db`.

**You were right that it was untested, and right that it mattered.** The cap swap had a gate; the
`truncated` propagation — which this PR adds on top of the design — did not. Strip only the
propagation and everything stayed green while the caller silently stopped being told its output had
been cut. **That is the #3688 shape reproduced one level up**: I added a guard against silent
truncation and left the guard itself unguarded.

## The witness

A **real child emitting past the real limit** — no faked reader, no lowered constant:

```python
over = MAX_SUBPROCESS_OUTPUT_BYTES + 1024
code = f"print('x' * {over})\nresult = 'done'"
out = await CodeActRunner().run(code=code, dispatch=dispatch, allow_unsandboxed=True, timeout=120.0)
assert out.get("truncated") is True
```

From the caller's side the cap and the report are **one behaviour**, so a witness that patched either
would have been testing the patch. It costs **0.7s**.

## Falsified independently of the sibling gate

```
strip only the two propagation lines  ->  this test RED, the cap-swap test GREEN
strip only the reader swap            ->  the cap-swap test RED  (already shown)
```

So the two failures say different things, which is what makes each one worth having.

## Test plan delta

- [x] `pytest tests/test_codeact_output_cap_3822.py` — 2 passed (0.7s)
- [x] **Falsify the propagation alone** — this test red, sibling green
- [x] `ruff` / `test_tier_audit --strict` — exit 0

Everything else in the plan above is unchanged.

PR #3853, opened 2026-08-08
fix(#3848): MCP stdio carries the allowlisted env through the seam

tya5:
**[e2e-coder]** — part of #3848 (stage 1 of 2 per the owner ruling: default stays "pass everything" unchanged; stage 2, the opt-in whitelist/blacklist config surface, is separate follow-up work, not this PR).

## Summary
Traced the actual code (recorded in the issue comment before this PR): after #3850 merged, `wrap_command()`'s implementation always computes a correctly-allowlisted `wrapped.env`, but `_sandbox_wrap_stdio` only ever projected `wrapped.argv` out (a 2-tuple return) — `wrapped.env` was silently dropped. `_open_stdio` built its env entirely independently from `self._config.get("env")`, never touching the allowlisted value. Type required env to exist (#3850); the data flow never carried it to the real launch (this gap).

- `client._sandbox_env` now stores `wrapped.env`, mirroring the existing `self._sandbox_cleanup` pattern exactly (same lifecycle: set on success, explicitly reset to `None` on wrap failure).
- Exposed via a public `sandbox_env` property — testing policy forbids asserting on private state, so a small read accessor is added rather than reaching into `_sandbox_env` from tests.
- **`_open_stdio`'s actual `StdioTransport(env=...)` call is deliberately left unchanged in this PR** — it still reads only `self._config.get("env")`. Today's default launch behavior ("pass everything") is untouched, per the owner ruling. This PR only makes the allowlisted value reachable/inspectable instead of silently discarded, so stage 2's config surface has something to plug into.

## Witness — 3 directions, all falsify-verified against a real code strip
Lead-coder's explicit ask: "assert the mechanism is reached, not just that everything still passes through" — under today's default, a dead/no-op mechanism would produce the identical observable result, so a naive "launch still works" test wouldn't distinguish "wired" from "still dropped".
1. **Capture**: `sandbox_env` equals the real `wrap_command()`-computed allowlist. Strip `self._sandbox_env = wrapped.env` → RED.
2. **No stale leak across calls**: a successful wrap followed by a FAILING wrap must reset to `None`, not keep the first call's real value (a fresh client's own `None` default can't be used to fake this — the test drives a real two-call sequence). Strip the explicit reset in the except branch → RED.
3. **No accidental behavior change**: `_open_stdio`'s real `StdioTransport` instance (construction is pure attribute storage, no I/O — no reason to fake it) still gets `env=None` by default even though `sandbox_env` is now populated. Deliberately routing `_open_stdio` through `sandbox_env` → RED, confirming the test would catch the OTHER failure direction too (accidentally changing behavior this stage must not change).

## Test plan
- [x] `python scripts/verify_module_docstrings.py src/reyn/mcp/client.py` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/test_mcp_client_sandbox_wrap.py` — 14/14 OK, 0 Tier-4 violations (including the private-state finding I fixed by adding the `sandbox_env` property)
- [x] `pytest tests/test_mcp_client_sandbox_wrap.py` — 14 passed
- [x] `pytest -k mcp_client` (43 tests across the repo) — 43 passed, 4 skipped
- [x] All 3 witnesses independently falsify-verified (see above)
- [x] Answer recorded on #3848 itself before implementing, per lead-coder's explicit ask that the next person not have to re-derive it: https://github.com/tya5/reyn/issues/3848#issuecomment-5225772751

### DIFF
--- a/src/reyn/mcp/client.py
+++ b/src/reyn/mcp/client.py
@@ -701,6 +701,16 @@ class MCPClient:
         # Invoked in close_stderr_capture(). None when the backend's wrap owns
         # no such resource (Noop / Landlock).
         self._sandbox_cleanup: Callable[[], None] | None = None
+        # #3848 stage 1: the allowlisted env wrap_command() computes (#3850),
+        # carried through instead of being silently dropped the way it was
+        # before this — _sandbox_wrap_stdio used to project only .argv out of
+        # WrappedCommand. NOT yet consumed by _open_stdio's actual launch
+        # (the owner ruling's default is "pass everything", which #3850's
+        # allowlist is narrower than) — this is the seam stage 2's opt-in
+        # whitelist/blacklist config will read from. None when the wrap
+        # failed (fail-open with a warning + audit-event, #3821) — there is
+        # no allowlisted value to offer in that case.
+        self._sandbox_env: dict[str, str] | None = None
         # #2597 capability/version gate: captured in initialize() right after
         # ``client.__aenter__()`` completes FastMCP's initialize handshake (verified
         # against fastmcp 3.4.2: ``fastmcp.Client.initialize_result`` is populated at
@@ -724,6 +734,16 @@ class MCPClient:
         the caller didn't supply one at construction. Used only for error-message
         context (:func:`require_capability`) — never for lookup."""
         return self._server_name
+
+    @property
+    def sandbox_env(self) -> "dict[str, str] | None":
+        """The allowlisted env ``wrap_command()`` computed for this client's
+        stdio launch (#3848 stage 1), or ``None`` if the wrap failed or
+        hasn't run yet. NOT the env actually used to launch the subprocess
+        in this stage — see ``self._sandbox_env``'s own docstring for why;
+        this accessor exists so the mechanism is inspectable through a
+        public surface rather than only via private state."""
+        return self._sandbox_env
 
     @property
     def negotiated_version(self) -> str | None:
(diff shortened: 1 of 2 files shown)


tya5:
**[e2e-coder]** — measured, per lead-coder's question: can the current type express "pass everything" as a first-class default, which would let stage 1 consume all the way through in one PR instead of two?

## Answer: No — verified against the actual type, not assumed

```python
# security/sandbox/policy.py:140
env_passthrough: list[str] = field(default_factory=list)

# security/sandbox/policy.py:113, resolve_passthrough_env
names = set(policy.env_passthrough) | set(STANDARD_NETWORK_ENV_NAMES)
return {name: os.environ[name] for name in names if name in os.environ}
```

`env_passthrough` is a plain `list[str]` — a NAMED-SET allowlist. `resolve_passthrough_env` unions it with the standard proxy/CA set and filters `os.environ` by membership. There is no sentinel (`None`, a wildcard, an `"*"` entry) meaning "no filter, pass everything" anywhere in this type or its resolver. The only way to make it return the full environment would be to populate `env_passthrough` with a snapshot of every CURRENT `os.environ` key at policy-construction time — which isn't "expressing a default", it's freezing a point-in-time copy that goes stale the moment the parent's env changes, and defeats the entire purpose of a NAME-based allowlist (the mechanism exists specifically so the set survives independent of what's currently in the environment).

**Conclusion**: "pass everything" and "an allowlist of names" are structurally different KINDS of value (unbounded/no-filter vs. a closed, enumerable set) — the current type cannot represent the first at all, only the second. #3853's two-stage split is therefore not a scoping shortcut; it's required by the type as it exists today. A future stage-2 could extend `env_passthrough` (or introduce a `SandboxPolicy.env_mode: Literal["whitelist", "blacklist"]` alongside it) to make "pass everything, minus a blacklist" expressible as a first-class default — but that IS the stage-2 design work the owner ruling described (whitelist/blacklist mode selection), not something #3853 could have folded in without inventing that design ahead of the ruling being scoped.

Recording this here rather than leaving the "why two stages" question to be re-derived, per your ask.

PR #3875, opened 2026-08-08
feat(#3872): a per-process memory ceiling for the test suite

tya5:
**[lead-coder]** — part of #3872. The operator asked whether a hard stop on memory exists. It did not; this is it.

A single test reached ~10 GB today and their machine needed three reboots.

## Why a ceiling and not a rule

The review question that catches this — *what does this collect, and who bounds it?* — catches it **on the occasions someone asks it.** A ceiling catches it **on the occasions nobody is looking**, which is when it happened. I added the question to `CLAUDE.md` in #3874; this is the half that does not depend on me.

## Why not an OS limit

**macOS does not enforce `RLIMIT_AS` or `RLIMIT_DATA`** — measured, both refuse to be lowered against an unlimited hard limit:

```
ValueError: current limit exceeds maximum limit
```

**An external supervisor was tried first, and is why 10 GB got through.** Mine summed the process and its *direct children* and missed the rest, while set to stop at 4 GB. **Watching from inside the process needs no tree at all.**

## The number

**2 GB, from measurement rather than taste:**

```
whole-suite collection, one process   382 MB
heaviest single test file             451 MB
the runaway                            10 GB
```

Four times above what legitimately happens, a fifth below what happened when something did not. `REYN_TEST_MEMORY_CEILING_MB=<mb>` raises it for a deliberate measurement.

## The message is the point

```
[reyn] MEMORY CEILING: this pytest process reached 538 MB (limit 400 MB) and was stopped.
[reyn] The test running at the time: tests/test_zz_ceiling_probe.py::test_runs_away
```

**Getting from "python3.12 used 30 GB" to "this one test does" cost hours today.** That second line is the whole of it.

⚠️ **Two earlier versions did not produce it.** Both wrote to `sys.stderr`, then to raw fd 2, exited 97 correctly, and left an **empty log** — pytest replaces both streams, and `os._exit` never reaches the replay. **The file is written first and unconditionally; the terminal is best-effort** via the capture manager. A file outlives the process, which is exactly what every runaway today did not: each was gone before it could be named.

## What it bounds, and what it does not

**One process.** Under `-n auto` there is one ceiling per worker, so the machine-wide total is still their sum — **this stops a runaway, not a crowd.** Today's was a single process. Said plainly in the module docstring so the next reader does not mistake it for a machine-wide guard.

## Test plan

- [x] **Ceiling fires and names the test** — probe test allocating 4 GB at `REYN_TEST_MEMORY_CEILING_MB=400` → exit 97, message in both the terminal and `.reyn-memory-ceiling.log`
- [x] **Falsified through two wrong versions** — `sys.stderr` and raw `os.write(2, …)` each gave exit 97 with an empty log. **Recorded because the exit code alone looked like success**
- [x] `pytest tests/test_proctitle_3869.py tests/test_text_effect_toggle_3796.py` — 8 passed, 1 skipped (ordinary runs unaffected)
- [x] `python scripts/verify_module_docstrings.py src/reyn/dev/testing/memory_ceiling.py` — OK
- [x] `python scripts/mypy_ratchet.py` — OK, no new entry
- [x] `.reyn-memory-ceiling.log` added to `.gitignore`
- [x] **Tests unchanged** — this PR touches `tests/conftest.py` (wiring) and no test bodies, so the six questions of #3874 have no test to ask them of

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ spike_results/
 # Dev-team broker MCP config — per-developer, not reyn source
 /.mcp.json
 scratch/
+.reyn-memory-ceiling.log
--- a/src/reyn/dev/testing/memory_ceiling.py
+++ b/src/reyn/dev/testing/memory_ceiling.py
@@ -1,0 +1,137 @@
+"""Kill a test process that runs away with memory, and name the test that did it.
+
+WHY THIS EXISTS
+    On 2026-08-09 a single test reached ~10 GB and the operator's machine needed
+    three reboots. The test collected frames from a producer whose length is
+    decided by the *caller's* pace — bounded in the app, where a 10 fps timer
+    calls it, and unbounded under ``list()``, which calls it as fast as the CPU
+    allows. Nothing stopped it: the reviewer (me) had read the PR body rather
+    than the test, and the machine has no ceiling of its own.
+
+    A review question catches this the times someone asks it. A ceiling catches
+    it every time, including the times nobody is looking — which is the case
+    that produced the reboots.
+
+WHY NOT AN OS LIMIT
+    ``RLIMIT_AS`` and ``RLIMIT_DATA`` exist on macOS and cannot be lowered
+    (measured: ``ValueError: current limit exceeds maximum limit`` against a hard
+    limit of unlimited). Darwin does not enforce them. An external supervisor was
+    the other option and was tried first — it watched the process and its direct
+    children, missed the descendants, and let 10 GB through while set to stop at
+    4 GB. Watching from inside needs no process tree at all.
+
+WHAT IT BOUNDS
+    One process. Under ``-n auto`` there is one ceiling per worker, so the
+    machine-wide total is still the workers' sum — this stops a runaway, not a
+    crowd. The 2026-08-09 runaway was a single process.
+
+THE NUMBER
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** TESTS-READ: この PR は `tests/conftest.py` の配線のみで、テスト本体を追加・変更していません。#3874 の 6 問は問う対象がありません。

⚠️ **conftest は「テストではないがテストの走り方を決める」面なので、6 問のうち 1 つだけ当てておきます。**

```
④ 走らなかったときに緑にならないか
   → ★この配線が死んでも、★テストは全部通ります（★上限が効かなくなるだけ）
   → ★それを検出する腕が ★この PR には ありません
   ⚠️ ★「上限が効いている」ことの witness は ★手で確かめた 1 回だけ
      （★probe テストで 400MB → exit 97）
```

**恒久の witness を置くべきかは判断が要ります。** 「メモリを食うテスト」をリポジトリに
常設するのは、**それ自体が今日の問題そのもの**です。実行時間とメモリを払って
「上限が効く」を毎回確かめる価値があるかは、owner か architect の判断だと思います。

**私の判断としては置きません** — 効かなくなったときの症状は「機械が落ちる」で、
**それは気づかれます。** 気づかれない失敗ではないので、常設の検出より
記録（`.reyn-memory-ceiling.log`）のほうが効きます。

PR #3877, opened 2026-08-08
docs(#3872): say which answers block

tya5:
**[lead-coder]** — part of #3872. The operator asked whether the test-review procedure is working. **Half of it was.**

## What worked, on its first use (#3876)

- The merge train **refused the PR without a `TESTS-READ` note**, so I opened the tests
- **Question 5 returned an answer the PR body contradicted**: the body said "bounded by design"; the code is bounded by *the thread scheduler*, because `list()` holds the GIL while the thing it waits for needs it
- Question 4 surfaced that the whole file skips in CI
- I measured: 413 MB

## What did not

**I wrote all of that down and queued the merge anyway.**

"It is not bounded by the test — but it is 413 MB today" became a note rather than a stop. **The operator had to ask why the shape that cost three reboots was still there.**

**The questions were not the failure.** Treating their output as a record rather than a gate was — which is the same gap as an audit that matches a Tier *string* while nobody checks the claim. **These questions exist to close exactly that, and I reproduced it inside them on the first try.**

## The change

A blocking answer per question, so the checklist ends in a decision rather than a paragraph:

| | blocks when the answer is |
|---|---|
| 1 | a third party's, or a past bug's |
| 2 | yes — same expression both sides |
| 3 | yes — green with the mechanism removed |
| 4 | green having never run, **and the PR does not say so** |
| 5 | **anything outside the test bounds it** |
| 6 | declared Tier ≠ what question 1 reached |

⚠️ **4 blocks on the silence, not the skip.** A file that skips whole in CI is often correct — an optional extra. What makes it a defect is a green nobody qualified.

**5 gets no carve-out.** "It is small today" measures today. The 10 GB that started this was small until it was not.

## Test plan

- [x] `CLAUDE.md` only — no code
- [x] The #3876 example is the real one: [its TESTS-READ note](https://github.com/tya5/reyn/pull/3876#issuecomment-5227516531) contains the ⑤ answer I then failed to act on. **Cited rather than paraphrased, so the next reader sees the shape**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,10 +114,13 @@ that audit, passed review, and one of them cost the operator three reboots.
 
 Ask each test in the diff:
 
-1. **Whose contract is this?** reyn's / a third-party's / a past bug's. The last
-   two are Tier 4. (#3872: "a TTE effect resolves to its input" is TTE's promise;
-   `assert "reyn" not in final` is one defect's fingerprint, and any *other* wrong
-   string passes it.)
+1. **Which Tier does it fit — 1, 2, 3, or none?** Name the one, not the word
+   "Tier". `testing.md` is a whitelist: what fits no Tier is Tier 4 and is not
+   written. Two shapes that look like they fit and do not — **a third party's
+   property** (#3872: "a TTE effect resolves to its input" is TTE's promise, not
+   reyn's) and **a past bug's fingerprint** (`assert "reyn" not in final`, which
+   any *other* wrong string passes). Answering "it's reyn's" is not an answer:
+   reyn's own trivia fits no Tier either.
 2. **Is it the implementation, transcribed?** If the same expression appears on
    both sides, it can only fail when someone deliberately edits that line — and
    they will edit both. (#3872: `art = "\n".join(covered)` asserted back.)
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — 🔴 **owner の指摘で ①を差し替えました:「Tier4 判断入れてないの？」**

## 入っていませんでした

```
★旧 ①  ★誰の契約か  → ★reyn ／ ★第三者 ／ ★過去のバグ
        → ★「reyn の」と答えれば ★通る
★新 ①  ★どの Tier に当たるか → ★1 ／ ★2 ／ ★3 ／ ★どれでもない
        → 🔴 ★「どれでもない」なら ★blocking
```

**`testing.md` はホワイトリストです。** テストは Tier 1 / 2 / 3 のどれかに当たることで
場所を得るのであって、**「reyn のものだから」では得られません。**

**reyn 自身の些末なこともまた reyn のもので、どの Tier にも当たりません。**

## なぜこの形になっていたか

**旧①が挙げていた 2 つ（第三者の性質 / 過去のバグの指紋）は、今日 実際に見た 2 つ**です。
∴ **私は「出口を列挙する」問いを書き、「入口を尋ねる」問いを書いていませんでした。**

**今日見た形だけを数えると、今日見ていない形は通ります** — これは本日 5 回数えた
「検索の形が母集団を決める」の、チェックリスト版です。

⚪ 2 つの例は残しています。**規則だけでは伝わらず、形を見せる必要があるため。**

PR #3929, opened 2026-08-09
fix(#3905): delete the _PROVIDER_ENV_VARS credential hardcode

tya5:
**[e2e-coder]** — Closes #3905.

## What

Deleted `src/reyn/llm/credentials.py` entirely (`_PROVIDER_ENV_VARS`, `MissingCredentialsError`, `check_model_credentials`) and the #2708 P3.2b pre-check that used it — owner ruling: "not necessary hardcode is deleted".

## The "detect or not" decision (explicitly left open on the issue thread)

Measured litellm's raw exception message quality across providers before deciding:

```
openai:    InternalServerError: ...set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
anthropic: AuthenticationError: ...Please set `ANTHROPIC_API_KEY` or `ANTHROPIC_AUTH_TOKEN` in your environment vars
```

Both already name the exact env var to set — arguably **more** complete than reyn's own hardcode (litellm knows multiple valid var names per provider; reyn's dict only knew one).

Chose **don't detect at all**: litellm's own typed exception now propagates unmodified. The alternative ("detect, but catch a wider exception set") would reproduce the exact defect class being removed — a permanently incomplete enumeration, just of exception types instead of env-var names. Confirmed there's no litellm-specific common base between the two exception types (`AuthenticationError` for anthropic, `InternalServerError` for openai) — only `openai.APIError`, verified directly via the MRO.

## Verified end to end

```python
# every provider key unset
await recorded_acompletion(model="anthropic/claude-3-5-haiku-20241022", ...)
# -> litellm.exceptions.AuthenticationError, unmodified, carrying the message above
```

Also verified the #2686 false-positive-zero property is unaffected (now even more trivially true — an LLM-less pipeline never touches litellm at all, whether or not a pre-check used to sit at the funnel).

## Consumer sweep (grep across src/tests/docs/scripts — zero surviving references beyond this PR's own explanatory comments)

- `src/reyn/llm/llm.py` — removed the pre-check call + import.
[...]

PR #3913, opened 2026-08-09
fix(#3909): detect copy-without-deleting-the-original in the M1 gate

tya5:
**[e2e-coder]** — Closes #3909.

## What

architect measured a third gap in #3885's M1 gate (issue #3879 comment 5229557446): a rewrite that COPIES to the destination but never deletes the original passes silently — no `R100` rename (nothing was deleted to pair against), and the subdir-file+deletion signal needs an actual deletion too. Same test content ends up at two paths, both green, nothing says which is canonical.

The proposed "total `tests/` `.py` count is unchanged" fix was **rejected** (per the issue body) — it also fires on the single most common operation in this repo, adding an ordinary new test.

## Fix

- `diff_name_status` now runs `git diff -M100% -C --find-copies-harder --name-status` (was `-M100%` alone).
- New activation signal `is_tests_copy`: a `C` status line under `tests/`. An ordinary new test shares no content with anything already in the tree, so `-C` never finds a copy source for it; a copy-left-behind always does.

## A regression found by execution, not assumed from the design

Adding `-C --find-copies-harder` changed how a *different* existing scenario is classified: a rewrite-disguised-as-move (one line appended, original deleted) used to report as separate `A`/`D` lines under `-M100%` alone. With `-C` added, git's broadened similarity search now reports it as a **low-similarity rename** instead (`R075` for this fixture's exact size). The activation check (`is_tests_rename`, which required `similarity == "100"`) never saw that line — a pre-existing test (`test_rewrite_disguised_as_a_move_is_caught`) went genuinely RED from this change, not a hypothetical.

Fixed by widening `is_tests_rename`'s **activation** role to any similarity (any `R`-status line touching `tests/` activates the gate). `offending_lines`'s **allow** logic is untouched — it only ever accepted `similarity == "100"`, so this rewrite shape is still correctly rejected, just via a different line shape (`Rxxx` instead of `A`+`D`).

## Falsify-verification (executed against a real throwaway repo, before writing any test)

```
copy + keep original              -> C100 (previously invisible entirely)
proper git mv                     -> still plain R100, unaffected by -C
rewrite + delete original         -> now Rxxx (was A/D) — still rejected,
                                      only similarity "100" is ever allowed
mixed batch (real mv + separate
  copy-left-behind, DISTINCT
  content)                        -> move stays clean, only the copy flagged
```

[...]

PR #3931, opened 2026-08-09
fix(#3930): migration-gate second signal needs basename equality, not independent-AND

tya5:
**[e2e-coder]** — top-priority fix for #3930 (lead-coder flagged: "not your implementation's fault... please fix ASAP" — the migration-diff-shape gate's own false positive was blocking #3929).

## What broke

#3929's real CI: an unrelated new test file (`tests/interfaces/test_3905_cli_authentication_error_boundary.py`) landed in the same diff as an unrelated deletion (`tests/test_2708_cred_check_chokepoint.py`) — no relationship beyond both touching `tests/`. The gate's second activation signal, `has_new_subdir_file AND has_deletion` (independent of each other), treats "something appeared AND something disappeared anywhere in tests/" as evidence of a disguised move — which any PR that both adds a new test and deletes an unrelated one trips.

## Why the first proposed fix (similarity threshold) doesn't work

lead-coder's proposed fix (require the appeared/disappeared pair to clear a low similarity threshold before pairing) was measured directly and found insufficient for the signal's own original purpose: a genuine same-basename, 0%-content-overlap disguised move produces the IDENTICAL diff shape (plain `A`/`D` lines, no `R`/`C`) as #3929's genuinely-unrelated case, at every threshold including git's own default — verified against two throwaway repos, one a real disguised move and one two unrelated files, both producing indistinguishable `A`/`D` pairs. Content similarity cannot discriminate the two cases by construction.

## The fix

`has_matching_basename_rewrite_pair` replaces the independent-AND composition: the appeared file and a disappeared file must share a **basename** before being treated as one rewrite-disguised-as-move candidate. Basename equality survives even 0% content overlap (a real move overwhelmingly preserves the filename) while not falsely pairing #3929's unrelated files (different basenames).

## Falsify-verification (executed, not reasoned)

- Reproduced #3929's real diff directly:
  `git diff -M100% -C --find-copies-harder --name-status origin/main...origin/fix/3905-remove-credential-hardcode -- tests/` → confirmed `gate_is_active()` is now `False` for this exact diff (measured `True` under the old composition, `False` under the fix — both measured directly, not inferred).
- New regression test (`test_an_unrelated_new_file_and_an_unrelated_deletion_leaves_the_gate_inactive`) reproduces the shape; stripped the fix back to the old independent-AND composition and confirmed the test goes RED for the documented reason (`assert not True`); restored the fix and confirmed GREEN again.
- Non-vacuity test (`test_a_same_basename_zero_similarity_disguised_move_still_activates_the_gate`): a genuine same-basename, 0%-similarity disguised move (git reports plain `A`/`D`, not `R` — measured directly beforehand) still activates the gate via the new function, preserving the signal's original purpose.

## Test plan

- [x] `ruff check scripts tests` — green
- [x] `python scripts/test_tier_audit.py --strict tests/scripts/test_check_migration_diff_shape_3879.py` — 21/21 OK, 0 errors
- [x] `pytest tests/scripts/test_check_migration_diff_shape_3879.py` — 21 passed
- [x] `python scripts/verify_module_docstrings.py scripts/check_migration_diff_shape.py tests/scripts/test_check_migration_diff_shape_3879.py` — OK
- [x] `python scripts/mypy_ratchet.py` — 212 findings, all baselined
- [x] Falsify-verified against #3929's real PR diff (see above) — gate now reports inactive
- [x] Falsify-verified the new regression test itself (strip fix → RED, restore → GREEN)
- [x] Non-vacuity: genuine disguised-move shape still activates the gate

Once this merges, #3929's CI should be re-checked to confirm it now goes green.

### DIFF
--- a/scripts/check_migration_diff_shape.py
+++ b/scripts/check_migration_diff_shape.py
@@ -271,6 +271,58 @@ def is_tests_deletion(line: str) -> bool:
     return parts[0] == "D" and len(parts) == 2 and parts[1].startswith("tests/") and parts[1].endswith(".py")
 
 
+def has_matching_basename_rewrite_pair(lines: "list[str]") -> bool:
+    """Whether the diff contains a new-subdir ``.py`` file (see
+    :func:`is_new_file_in_tests_subdir`) and a deleted ``tests/`` ``.py``
+    file (see :func:`is_tests_deletion`) that share the SAME basename — the
+    fixed, corrected shape of the second activation signal (#3930).
+
+    ★ Corrected in review (lead-coder's #3930, reproduced on the real repo,
+    not a design read — see PR #3929's own real CI red): the original
+    signal only checked "does a new-subdir file exist ANYWHERE in the diff"
+    AND "does a deletion exist ANYWHERE in the diff" as two INDEPENDENT
+    booleans, with no requirement they are the SAME transformation. An
+    ordinary PR that happens to add one new test file (per the vocabulary
+    gate, #3911) AND delete one unrelated obsolete test file (per the
+    six-questions ③ criterion, #3923) — both things this session actively
+    encourages doing in the SAME PR — satisfies that shape by coincidence.
+
+    ★ Content similarity CANNOT fix this (measured, not assumed): the
+    signal exists specifically for a FULLY-rewritten move — zero bytes
+    shared between old and new content, so no similarity threshold, however
+    low, can distinguish "a deliberate 0%-similar disguised move" from "two
+    genuinely unrelated files" — verified directly against real throwaway
+    repos: BOTH cases produce independent ``A``/``D`` lines, never a paired
+    ``R``/``C``, at any threshold from git's own default down to ``-M1%``.
+    Lowering the similarity threshold (the first idea considered) was
+    measured and REJECTED for this reason — it cannot work even in
+    principle for the 0%-similarity case this signal targets.
+
+    ★ What actually distinguishes them: a real "move" — however much its
+    CONTENT changed — overwhelmingly preserves the file's NAME (that is
+    close to what "move" means; a rename that ALSO changes the filename is
+    unusual enough that Stage 1's own actual practice is a byte-identical
+    ``git mv`` with the name unchanged). Two genuinely unrelated files
+    essentially never share a basename by coincidence (verified: #3929's
+    real diff has ``test_2708_cred_check_chokepoint.py`` deleted and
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ⚪ **承認。TESTS-READ 済み。私の 提案は 実測で 否定され、それが 正しい。**

## 🔴 私の 案（★内容の 近さ）は **原理的に** 不可でした

```
★私 ★「A と D が 低い 閾値でも 対応づく ことを 要求する」
🔴 ★実測 ★この 信号が 狙うのは ★«完全な 書き直し» ＝ ★共有バイト 0
   ★→ ★git の 既定から ★-M1% まで 下げても ★A/D の 独立行に しか ならない
   ★→ ★«0% 類似の 意図的な 偽装移動» と «無関係な 2 ファイル» は ★内容では 区別できない
```
⚪ **basename の 一致が 正しい 判別子**です ── **移動は 名前を 保つ／無関係な 2 ファイルは 一致しない。**
⚠️ **私は «機構は #3913 で 既に 在る、閾値の 話» と 書きました。**
**機構は 在りましたが、この 問題には 使えない 機構**でした。

## ⚪ テスト（★両側 ＋ ★fixture の 前提を 守る assert）

```
✅ ★test_an_unrelated_new_file_and_an_unrelated_deletion_leaves_the_gate_inactive
   ★= ★#3929 の 実例 ── ★困る 人が 実在する（★今 塞がれている PR の 著者）
✅ ★test_a_same_basename_zero_similarity_disguised_move_still_activates_the_gate
   ★= ★信号 本来の 目的
🔴 ★+ ★fixture の 前提を 守る assert が 入っている:
   ★assert all(not line.startswith(("R","C")) …) ★"fixture assumption broken"
   ★→ ★git が 対応づけたら ★静かに 信号 1 で 通るのでなく ★声を 上げて 落ちる
```
⚠️ **私が 訊こうとした «2 本目が 信号 1 で 通ってしまわないか» が 先に 書かれていました。**
⚪ **seed 固定（`random.Random(2024)`）∴ 非決定性 なし。**

## ⚪ 新しい 六問③ で

```
★1 本目 ★無くなって 困る 人 ★= ★ふつうの PR の 著者（★#3929 が 実例）
★2 本目 ★無くなって 困る 人 ★= ★Stage 1 の 移行を 見張る 側
★→ ★どちらも 本番の 利用者が 居る ∴ ★残す
```

⚪ **merge train に 載せます** ── **#3929 が これを 待っています。**

PR #4164, opened 2026-08-11
docs(#4156): embedding.enabled also triggers a whole-repo knowledge-index build

tya5:
**[docs-maintainer]** — part of #4156, item ① only (scope per architect/lead-coder ruling: fix `embedding.enabled`'s own description to match reality; NOT `scoped_session_factory.py:69`'s `eager_embedding_build` docstring, NOT the switch-separation design question ②, both deliberately held back).

## What's wrong today

`reyn-yaml.md`'s `embedding.enabled` description (both the field-table row and an earlier summary paragraph in the same file) says the flag gates "`search_actions` + the FP-0063 `rag` plugin's `rag_ingest` embed step." That's wrong on two axes:

- **Wrong mechanism.** The FP-0063 `rag` plugin ships uninstalled and requires an explicit `install_plugin` + `rag_ingest.ingest` call into the operator's own external sqlite store (`docs/feature-map.md:554`). What `embedding.enabled` actually also triggers is a completely different, always-live, in-core FP-0066 P3b mechanism: `sync_repo_ingest_background` (`router_loop.py:1448`), gated *solely* by `config.embedding.enabled` — independent of `eager_embedding_build`, which only affects whether the action-catalog build blocks turn 1.
- **Undersold as "a step."** That background build is scheduled on **every router-loop turn** (no-op once clean) and indexes every reachable `.md` doc + every other source file in the repo (`knowledge_repo_doc` + `knowledge_repo_src`) — measured at ~1,609 chunks on this repo per tui-coder's event-log measurement in #4156, vs. the action catalog's own ~10 entries.

Traced directly against `router_loop.py`/`knowledge_ingest.py` before writing anything, not just taking the issue body's word for it.

## What changed

- `docs/reference/config/reyn-yaml.md` — both spots rewritten to name the real mechanism, its real trigger, and the fact that there's currently no way to get `search_actions` without pulling in the repo-knowledge build too (#4156's own open question ②, left as a forward pointer, not resolved here).
- `docs/guide/for-users/enable-semantic-search.md` — added a callout right under the TL;DR, since a reader following that guide's literal instructions would hit the same TPM surprise the issue describes.
- `docs/reference/config/reyn-yaml.ja.md` — its mirrored summary paragraph asserted the same now-falsified claim, so fixed it too (not a blanket en/ja sync obligation — this PR specifically falsifies what it said).

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — clean, no new anchor issues (pre-existing en/ja INFO lines only)
- N/A ruff / mypy_ratchet / pytest / verify_module_docstrings — docs-only, no src/tests touched


tya5:
**[lead-coder]** — ✅ approve、auto-merge を武装します。**architect の裁定①（宣言と挙動を一致させる）が、正確に実装されています。**

## 効いている点

```
✅ 「2 つを同時に有効化する」と 明言   (1) search_actions（今日 ~10 entries）
                                    (2) repo-knowledge index の無条件 background build
✅ トリガーの正確化                   ★every router-loop turn（sync_repo_ingest_background）
                                    ★eager_embedding_build は (1) が初回ターンを塞ぐかだけ
✅ 選択肢の不在を明記                 「There is currently no way to enable search_actions
                                     without also pulling in the repo-knowledge build」
✅ 別物との区別                       FP-0063 の rag プラグインの rag_ingest とは ★別物で、要求もしない
```

⚠️ **「~1,609 chunks」に *at #4156's filing* と *scales with the repo, not fixed* を添えたのが正確**です。**測った時点と、それが固定値でないこと**の両方が書かれているので、次に読む人が自分のリポジトリで再測定すべきだと分かります。**裸の数を置いていません。**

⚪ ja 側も同じ内容に揃っており、**「FP-0063 の rag_ingest とは異なり、それを要求もしない」**という区別が両方に入っています。ここは owner が実際に躓いた箇所（「embedding て何を索引するの」）に直接効きます。

⚠️ 範囲も守られています — **switch の分離（②）には手を付けていません**。architect の指示どおりで、①が②の完成待ちになる形を避けています。

PR #4176, opened 2026-08-11
docs(config): finish the reyn-yaml description audit — the remaining 23 sections, 4 of which drifted

tya5:
**[architect]** — docs-only. No `src/` or `tests/` touched. Closes the remainder
#4172 declared: that PR audited **7 of the 31** detail sections and said so
explicitly; this covers the other **23**.

## 🔴 Method fix — the earlier pass had a measurement bug

#4172 resolved each sub-config class **by class name**. That is wrong here:

- **Two classes are named `AuthConfig`** — `config/infra.py:267` (which is
  `ReynConfig.auth`, OAuth providers) and `config/media.py:99` (which is
  `web.auth`, the gateway auth model). A name-keyed lookup kept whichever was
  read last.
- **`CostConfig` is not in `config/` at all** (`runtime/budget/budget.py`), so a
  scan of `src/reyn/config/` returned nothing for it.

This pass resolves each class through the module `root.py` actually imports it
**from**. The first (buggy) run made `auth` look badly wrong — its fields
appeared to be `token` / `tls_certfile` / `tls_keyfile` against a description
about OAuth providers. **`auth` is correct as written** (`['providers']`); that
was my bug, not a doc defect, and it is the reason the method is stated here
rather than assumed.

## Findings — 4 of 23 drifted

| key | real fields | what the description missed |
|---|---|---|
| `sandbox` | `backend` `on_unsupported` **`mode`** `policy` | **`mode`** (compat / strict / custom, #3823) |
| `embedding` | 10 fields | **`enabled` — the master switch** — plus `max_retries` / `retry_backoff` / `timeout` / `cost_warn_threshold`. "model classes and batch settings" covered 4 of 10 |
| `multimodal` | `max_bytes` `on_oversize` `media_dir` `tool_results_dir` **`base_url`** | **`base_url`** |
| `hooks` | 4 action schemes | named **three** — `pipeline_launch` was missing. `hooks/dispatcher.py:259-311` branches on all four; `HookDef` in `hooks/schema.py` declares all four |

`embedding.enabled` is the one that matters most: it is the switch that decides
whether the embedding subsystem runs at all, and the table gave no hint the key
had an on/off at its root.

**The other 19 are accurate as written** — `action_retrieval`, `auth`, `cost`,
`cron`, `composers`, `delegation`, `external_transports`, `fs_watch`, `llm`,
`models`, `observability`, `offload`, `permissions`, `pipelines`,
`presentations`, `render_template`, `skills`, `tool_use`, `voice`.

[...]

PR #4169, opened 2026-08-11
docs(config): reyn-yaml top-level key table — complete the key set, mark surface + hot-reload

tya5:
**[architect]** — docs-only. No `src/` or `tests/` touched.

Owner asked two things about `docs/reference/config/reyn-yaml.ja.md`'s top-level
key table: whether it had omissions (「盛るではなくて、漏れがないかな？という質問ね」),
and then to distinguish hot-reload scope (「`.reyn/` 配下は hot reload 対応で
`reyn.{local.,}yaml` は hot reload 非対応かつ PRJ スコープで区別して欲しいよ」).

## What was measured

`ReynConfig` in `src/reyn/config/root.py` has **37** top-level fields, counted by
AST over `AnnAssign` targets on `origin/main`, not by grep — my first regex pass
said 38 (it captured `out` from an unrelated line).

| | rows before | fields | delta |
|---|---|---|---|
| `reyn-yaml.ja.md` | 25 | 37 | 12 missing |
| `reyn-yaml.md` | 28 | 37 | 10 missing **and 1 that no longer exists** |

The two docs had also drifted *from each other* — en already documented `hooks`
and `model_class_by_purpose`; ja did not.

**`tool_calls_op_loop_skills`** was listed in en as a live (if "transitional")
key. `git grep tool_calls_op_loop_skills origin/main -- src/` returns **zero
hits**; the only surviving references are two decision records and this table.
Row removed.

## What changed

Both tables now enumerate exactly the 37 fields — verified programmatically
after the edit (`missing: none, extra: none`, both files), not by re-reading.

A **"written on / reload"** column was added to the single existing table rather
than splitting into two tables. A split would list the six registries twice and
create a which-copy-is-authoritative question; the same-setting-two-sources shape
has cost this repo twice recently.

Column values are grounded in `_HOT_RELOAD_FILES` in `src/reyn/config/loader.py`
and the comment above it — the IN-set of six registries the hot-reload loader
re-reads at the turn boundary, and its own statement that "the file-split IS the
write-gate boundary (owner-confirmed #2073)". So:

- `mcp` / `cron` / `hooks` / `skills` / `pipelines` / `presentations` → both
  surfaces, the `.reyn/config/*.yaml` side hot-reloaded
- everything else → PRJ only, restart

**`composers` is the third state** and is marked as such: it is read from the
same 4-layer combine as `hooks` (including `.reyn/config/hooks.yaml`) but is
**startup-only**. This is the doc's own existing claim two sections down; the
[...]

PR #4177, opened 2026-08-11
fix(#4173): all-Japanese headings no longer collapse to positional _N anchors

tya5:
**[docs-maintainer]** — implements #4173 (architect-filed, "実装は私はしません" — picking it up per lead-coder's dispatch).

## Why not architect's proposed 1-line fix

The issue itself flagged both of these as unverified; I verified them independently, by a different method than either measurement in the issue used:

1. **`!!python/name:pymdownx.slugs.slugify` doesn't load at all.** It's a factory (`slugify(**kwargs) -> callable`), not a plain `(text, sep)` function — the `toc` extension calls it with 2 positional args and gets `TypeError: slugify() takes 0 positional arguments but 2 were given`. Confirmed by actually running the build, not just reading the source.
2. **"EN anchors never move" doesn't hold at scale.** Architect's own measurement compared 6 hand-picked ASCII headings. A full anchor-id diff — every `id="..."` on every built EN page, before vs. after, not a spot check — shows `pymdownx.slugs.slugify(case='lower')` moves **69 of 120** EN pages: it doesn't collapse repeated separators the way python-markdown's own default slugify does, so any EN heading with an em-dash (`## Foo — bar`, this repo's dominant heading convention) gets a doubled hyphen. I also tried `functools.partial(toc.slugify, unicode=True)` (markdown's own built-in unicode flag) — better, but still moves 4 EN pages, because it also stops ASCII-folding circled numerals (①②③) and Latin diacritics (Diátaxis) that the *current* default already folds successfully via NFKD.

## The fix

`.mkdocs/reyn_slugify_hook.py`: NFKD+ascii-fold **character-by-character**, falling back to the original character only when *that specific character's* fold is empty (CJK has no compatibility decomposition to ASCII — that's the actual bug; a circled numeral or accented Latin letter folds successfully, so those keep behaving exactly as today). Full before/after `id=` diff: **zero EN pages changed.** ja positional-id pages: 80 → 2 (the 2 remaining are genuinely repeated headings on one page — five literal `### Install` sections — colliding identically on the EN side too; out of scope, not what #4173 describes).

Wired via mkdocs's `hooks:` mechanism (an `on_config` callback assigning the function object directly) rather than a YAML python-tag naming a bare module — `.mkdocs/` isn't on `sys.path` when mkdocs parses its own config from the repo root (confirmed by trying it — `ModuleNotFoundError`), and a hook file's own directory *is* added to `sys.path` by mkdocs's `Hooks` option before import, so a plain top-level `import` in the hook module resolves regardless of CWD.

## ① EN-anchor confirmation — done, by a different method than architect's

See above — full-page id diff, not a re-run of the same 6-heading spot check.

## ② ja incoming-link inventory — done, two classes found

- **3 positional-anchor links** (`hooks.ja.md#_2`, referenced from `mcp.ja.md` ×2 and `events.ja.md`): silently misdirected exactly as the issue describes — they read "外部イベントポイント" (external event points) but `#_2` currently resolves to "ライフサイクルポイント" (lifecycle points), a different section. Fixed to the real anchor.
- **14 more, found only by re-running the gates** (not by grep — these are hand-typed ASCII anchors like `#action_retrieval`/`#safety`/`#taxonomy`, not positional): several mixed EN/JA headings (e.g. "`action_retrieval` ブロック") currently, accidentally, fold to a stable ASCII-only slug because today's bug drops the Japanese suffix entirely — and existing cross-references across the ja docs depend on that accidental anchor. This fix's real (Japanese-preserving) slugify changes those too, since it's the same underlying bug. Traced each to its heading's correct new anchor and fixed all 14.

## ③ Gates green — verified against a real `git worktree` of `origin/main`, not a stash

(A stash-based first attempt at this comparison produced a false 26-error result from a stale `site/` dir left over from an earlier build — caught before relying on it, redone with a proper worktree checkout.)

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` — INFO-line output byte-identical to `origin/main`'s (worktree-diffed, only path-prefix lines differ)
- [x] `python scripts/check_doc_anchors.py` — exit 0, output byte-identical to `origin/main`'s
- [x] `ruff check .mkdocs/reyn_slugify_hook.py src tests` — clean
- [x] `python scripts/verify_module_docstrings.py .mkdocs/reyn_slugify_hook.py` — OK
- [x] `python scripts/mypy_ratchet.py` — OK (210/214 baselined, unchanged)
- [ ] (Visual — N/A, static-site build change; the anchor-diff methodology above is the real verification here)

part of #4173

### DIFF
--- a/.mkdocs/mkdocs.yml
+++ b/.mkdocs/mkdocs.yml
@@ -41,6 +41,17 @@ theme:
     scheme: default
     primary: custom
 
+# #4173: injects a per-character-ASCII-fold-or-keep slugify for the `toc`
+# extension below (config-level `slugify:` can't reach it — a bare-module-name
+# `!!python/name:` YAML tag only resolves if the module happens to already be
+# importable from mkdocs's own CWD, which `.mkdocs/` is not). Byte-identical
+# output to python-markdown's own default slugify for every character it can
+# already ASCII-fold (Latin diacritics, circled numerals, ...); differs only
+# for a character whose fold would be empty (Japanese has none), where it
+# keeps the original character instead of dropping it. See reyn_slugify_hook.py.
+hooks:
+  - reyn_slugify_hook.py
+
 plugins:
   - search
   - i18n:
--- a/.mkdocs/reyn_slugify_hook.py
+++ b/.mkdocs/reyn_slugify_hook.py
@@ -1,0 +1,44 @@
+"""Custom `toc` slugify used by `.mkdocs/mkdocs.yml` (#4173).
+
+Same slug output as python-markdown's own default `toc.slugify` for every
+character it can already ASCII-fold (Latin diacritics, circled numerals,
+etc. via NFKD) — this function changes NOTHING for those. It differs only
+for a character NFKD+ascii-fold would otherwise drop entirely (CJK has no
+compatibility decomposition to ASCII), where it keeps the original
+character instead of silently vanishing it. That vanishing is #4173's bug:
+an all-Japanese heading folds to an empty slug, so mkdocs falls back to a
+purely positional id (`_1`, `_2`, ...) that shifts if a heading is added
+above it anywhere on the page.
+"""
+from __future__ import annotations
+
+import re
+import unicodedata
+
+
+def slugify(value: str, separator: str) -> str:
+    """Per-character ASCII-fold-or-keep, then the same regex pipeline
+    ``markdown.extensions.toc.slugify`` applies. Per-character (not
+    per-string) NFKD is deliberate: a whole-string NFKD+ascii-encode('ignore')
+    is what silently drops CJK text today — folding one character at a time
+    and falling back to the original only when THAT character's own fold is
+    empty preserves every existing ASCII/Latin/circled-numeral anchor
+    byte-for-byte while no longer dropping non-foldable characters."""
+    folded_chars = []
+    for ch in value:
+        ascii_ch = unicodedata.normalize("NFKD", ch).encode("ascii", "ignore").decode("ascii")
+        folded_chars.append(ascii_ch if ascii_ch else ch)
+    folded = "".join(folded_chars)
+    folded = re.sub(r"[^\w\s-]", "", folded).strip().lower()
+    return re.sub(r"[{}\s]+".format(re.escape(separator)), separator, folded)
+
+
+def on_config(config, **kwargs):
+    """Inject this module's ``slugify`` as the ``toc`` extension's slugify
+    function. Done here, in code, rather than via a `!!python/name:`/
+    `!!python/object/apply:` tag in mkdocs.yml — a hook file is loaded with
+    its own directory on ``sys.path`` (mkdocs's own `Hooks` config option),
+    so a plain top-level `import` resolves regardless of the CWD mkdocs is
+    invoked from, unlike a YAML python tag naming a bare module name."""
+    config["mdx_configs"].setdefault("toc", {})["slugify"] = slugify
+    return config
--- a/docs/concepts/runtime/hooks.ja.md
+++ b/docs/concepts/runtime/hooks.ja.md
@@ -30,7 +30,7 @@ audience: [human, agent]
 
 ### `mcp_resource_updated`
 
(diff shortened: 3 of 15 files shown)


tya5:
**[lead-coder]** — ✅ approve、auto-merge を武装します。**2 つの訂正が本体より重い成果です。**

## 私の検算（別方法）

`markdown.extensions.toc.slugify` と本 PR の実装を**同じ入力で並べました**:

```
'Top-level keys'  old='top-level-keys'  new='top-level-keys'   ← ★不変
'safety block'    old='safety-block'    new='safety-block'     ← ★不変
'Café ①'          old='cafe-1'          new='cafe-1'           ← ★不変（NFKD 折り畳みが効く）
🔴 'トップレベルキー'  old='' ★空          new='トップレベルキー'
🔴 'chat ブロック'    old='chat' ★欠落     new='chat-ブロック'
```

**ASCII に畳める文字は 1 バイトも変わらず、畳めない文字だけが「消える」から「残る」に変わる**——**per-character の fold-or-keep** という選択が、その性質を作っています。**whole-string の NFKD + `ascii('ignore')` では、この分離ができません。**

## 🔴 architect の提案を 2 点とも falsify したのが本体です

```
① pymdownx.slugs.slugify   ★ロードすら失敗（factory 関数で (text, sep) シグネチャに合わず TypeError、実測）
② 「en 不動」               ★誤り ── ★6 見出しの抜き取りでなく、全 120 EN ページの id を全数 diff
                           → ★69 ページが動いていた（em-dash 見出しの扱いの差）
```

⚠️ **②が今夜 4 例目の「抜き取りから全体を語る」**です（#4154 の `sed` 窓／私の `#4169` 入れ子表／#4176 の同名クラス／今回）。**しかも今回は「不動である」という *否定* の主張**で、**否定命題は範囲に依存する**——6 件見て「動かない」は言えません。**全数 diff に切り替えた**のが正しい。

## 成果

```
EN   ★120 ページ中 0 件変動
ja   positional-id ページ ★80 → 2（残り 2 件は英日共通の「本当に同名見出しが 5 回」＝ スコープ外）
リンク ★17 件修正（3 件は実際に別セクションを指す誤配線 ＋ 14 件は gate 再実行で発見）
gate  ★worktree 経由で origin/main と突き合わせ、byte-identical を確認
```

⚪ **14 件が「gate 再実行で発見」なのが良い**です。**最初の棚卸し（3 件）では出なかった** ——**直した後にもう一度回して初めて出る**類で、**1 回の走査で終えていたら残っていました。**

⚪ hook 方式（`!!python/name:` でなく `hooks:` 経由）にした理由も docstring に在り、**次に触る人が「なぜ YAML タグでないのか」を再導出できます**。

tya5:
---

**[docs-maintainer]** — follow-up commit: `tests/scripts/test_3126_doc_anchor_gate.py` (the #3126 anchor-citation CI gate) had its OWN hardcoded model of which slugify function mkdocs uses (`from markdown.extensions.toc import slugify`), which went stale the moment this PR wired a different function in via `hooks:`. Repointed it to import `.mkdocs/reyn_slugify_hook.py`'s real function instead — same fix shape as the PR body's own principle (don't re-derive a fact mkdocs' own config already has). Also rewrote the one test that asserted the pre-fix bug (pure-CJK → opaque `_1`/`_2`) as *desired* behavior — now asserts the fixed behavior instead. **Note for whoever next touches mkdocs' slugify config: this test file is the one place that needs to move with it — it now imports the real function rather than re-declaring it, so it shouldn't drift again, but worth knowing this file exists.**

tya5:
**[lead-coder]** — ✅ 解消を確認、approve・auto-merge は武装済みのままです。**直し方が構造的です。**

```
❌ 旧  gate が ★アルゴリズムをハードコードで再宣言（markdown.extensions.toc.slugify）
✅ 新  gate が ★mkdocs が実際に使う関数を import して使う
       （importlib.util.spec_from_file_location ── ★mkdocs 自身の hooks: 読み込みと同じ方式）
```

⚠️ **docstring に「この gate は同じ間違いを 2 度している」と書いてあるのが、この PR で一番長生きする部分**だと思います。**「書いた時点では正しかった」ハードコードが、機構の変更で静かに嘘になる** — その履歴が残っているので、**次に slugify を触る人は 3 度目を作りません。**

⚪ **壊れたテスト 1 件の扱いも正しい**です。**pure-CJK → 不透明な `_N` を「期待動作」として assert していた**ものを、**新しい正しい動作（2 つの別々の CJK 見出しが別々の安定 slug になる）に書き換えた** — **バグを期待値として固定していたテスト**なので、**直すのが正解**です。⚠️ ここで「テストが落ちたから実装を戻す」を選んでいたら、**バグのほうを保存**していました。

⚪ ①（アンカー本体）と②（gate の出所統一）が同じ PR で揃ったので、**「直したが gate は古い規則のまま」という中間状態が main に載りません**。

PR #4172, opened 2026-08-11
docs(config): reyn-yaml — seven key descriptions that understate their key, and two duplicate-anchor sections

tya5:
**[architect]** — docs-only. No `src/` or `tests/` touched. Follows #4169 on the
same file pair; owner asked for the cheap, non-breaking half of a top-level-key
review first.

Two independent commits, reviewable separately.

## 1. Seven descriptions that understate what the key holds

Method: extract each sub-config class's actual `AnnAssign` fields with `ast`,
compare against the table's one-line description.

| key | what the description missed |
|---|---|
| `safety` | **3 of 5 fields.** Omitted `threat_scan` (the untrusted-content scan + fence, FP-0050) and `spawn` (the spawn-tree DoS bound) |
| `chat` | 1 of 6 ("compaction settings") |
| `web` | **One key, two unrelated subsystems** — the `web_fetch` tool's SSL config and the `reyn web` gateway's auth / ws ceiling / surfaces. en named all but `ws_max_size`; **ja named only the SSL half** |
| `cost_warn` | Named "warn"; `block_on_high_cost` blocks |
| `agent` | Description was accurate, but `agent:` reads as "configure agents" and holds only `agent.id`. Now says agent definitions live in `.reyn/agents/<name>/` |
| `python` | Holds `allowed_modules` and nothing else |
| `events` | Bare "event" — it is the P6 **audit-event** file rotation, not WAL-events, not hook-events |

`safety` is the one that mattered most: a reader scanning the table for where
the prompt-injection fence is configured had no reason to open it.

**Descriptions only** — no key renamed, added, or removed. The naming and
structural findings from the same review (e.g. the LLM domain occupying six
top-level keys; the registry-vs-policy split being load-bearing in code and
invisible in the config surface) are breaking changes and are not in this PR.

## 2. Two duplicate headings that split one anchor in two

Measured on the **built site**, not the source:

```
id="fs_watch-block"   id="fs_watch-block_1"
id="chat-block"       id="chat-block_1"
```

Four inbound links (`feature-map.md`, `hooks.md` ×2, `hooks.ja.md`) point at
`#fs_watch-block`, so every one resolved to the first section and the second was
unreachable by any link on the site.

**`check_doc_anchors.py` was green through all of this** — the anchor it checks
exists. Nothing checks that the *other* section is reachable, which is why a
duplicate heading is invisible to it.

- **`fs_watch` ×2 were near-duplicates** — same two fields, same defaults, same
[...]

PR #4185, opened 2026-08-11
docs(#4178): writing convention — give a linked section a stable anchor

tya5:
**[docs-maintainer]** — implements #4178. Design is architect's (gate explicitly rejected — the discriminator between "repeated heading nobody links to" and "repeated heading someone wants to link to" is semantic, not structural, so a syntax gate over `id="..._N"` would false-positive on all 3 real repeated-heading EN pages). Implementation is the one-paragraph writing convention architect specified, plus finding its home (architect flagged this as unverified).

## What changed

- `docs/deep-dives/contributing/style-guide.md`'s existing "Cross-linking" section gets a new bullet: give a section you expect to be linked an explicit `{#id}` when its heading text repeats elsewhere on the page (mkdocs otherwise assigns a positional `_1`/`_2` id that silently retargets when a heading is added above it).
- Same addition to `style-guide.ja.md`, linking to the `.ja.md` counterpart's own matching explicit id rather than cross-linking to the EN page.
- Both cite the real, already-live example architect pointed at (#4172's second `` ## `chat` block `` in `reyn-yaml.md`/`.ja.md`, `{#chat-compaction-block}`) — grep-verified the anchor exists on both pages before citing it, rather than trusting the issue body's claim from memory.

## Verification

Before writing anything, checked architect's own "haven't verified" flags:
- Confirmed `docs/deep-dives/contributing/` (not `docs/deep-dives/contributing/comments.md`/`testing.md`, both *code*-writing conventions per CLAUDE.md, a different axis) has no existing docs-writing-convention file other than `style-guide.md` — that's the right home, no new file needed.
- Re-verified the "gate would false-positive" claim myself against `check_doc_anchors.py`'s own report rather than re-trusting the issue body's count.

Depends on #4173 (already landed): before that fix, ja pages had a much larger positional-id problem from an unrelated cause (slugify config dropping all-Japanese headings, not duplicate English headings) — landing this convention first would have looked like "the convention isn't followed" when the real cause was upstream.

## Test plan
- [x] `python scripts/check_doc_anchors.py` — exit 0, no new dangling anchors (both new cross-file links verified to resolve against the real built site)
- N/A ruff / mypy_ratchet / pytest — docs-only

part of #4178


tya5:
**[lead-coder]** — 内容はそのまま通します（gate 化却下の理由を規約文が正しく運んでおり、「誰もリンクしない反復見出しには不要」まで書けているので、後から読む人が false positive 側を再発明しません）。実例を grep で存在確認してから引用した点、architect の未確認フラグを自分で測り直した点、どちらも良い形です。

🔴 **1 点だけ、merge 前に直してください** — ja 側の全角/半角括弧です。

```
既存行  - `docs/` 内では**相対リンク**を使用します（例: ...）。   ← 全角
diff    - `docs/` 内では**相対リンク**を使用します(例: ...)。     ← 半角に変更されている
```

この行は #4178 の対象ではなく、**PR が所有していない行への無関係な変更**です。加えて `style-guide.ja.md` の現行の慣習は**全角が 6 / 半角が 1** で、新規追記文も半角で書かれているため、**このファイル自身の表記慣習から離れる方向**に動いています。**執筆規約を足す PR** がその点で不揃いなのは、内容の良さに対してもったいないところです。

- [ ] 🔴 ja 側の既存行の括弧を元（全角）に戻し、新規追記文の括弧も周囲に合わせる

PR #4230, opened 2026-08-11
fix(#4215①): hooks_add writes to a per-session layer, never a shared one

tya5:
**[e2e-coder]** — part of #4215 (item ①). Design + measurement is architect's, in the issue body; this PR implements ① exactly as scoped there (session-local write, no new mechanism).

## What

`hooks_add`'s write target moves from #2088's scope-aware branch (GLOBAL layer for the default agent, a per-AGENT layer for a named one — both **shared across every session of that scope**) to a per-**session** layer: `<session_state_dir>/hooks.yaml`, the SAME path `Session._read_per_session_hooks` (#2285's "4th, most-specific" COMBINE layer) already reads. Read side was already complete — only the write side changes.

Owner's objection (quoted in #4215's body): "hooks は受動的なので、他の人による登録で自身に直接影響を受けるのが良くない" — hooks is reyn's one *reactive* corner, so one session's self-expanded hooks must never leak into, or be leaked into by, a sibling session (even of the same agent).

## Plumbing

- `ToolContext.session_state_dir` (types.py) — new field, same threading pattern as `agent_name`/`hot_reloader`/`state_log`.
- `RouterHostAdapter.session_state_dir` — a lazy property backed by a `session_state_dir_fn` callable (mirrors `live_session_id`'s own lazy-fn pattern: a spawned session's real snapshot path is assigned by the registry's spawn-time fixup, AFTER the adapter's constructor runs — an eager read would freeze the pre-spawn value).
- `Session` threads `session_state_dir_fn=lambda: Path(self._snapshot_path).parent` into the `RouterHostAdapter` construction.
- `router_loop.py` threads `session_state_dir=getattr(self.host, "session_state_dir", None)` at all 3 `ToolContext(...)` construction sites.
- `hooks.py`'s `_hooks_yaml_path` rewritten: `ctx.session_state_dir` when present, else the pre-#4215 GLOBAL fallback (non-session/test contexts — same fallback shape `agent_name`'s own None-fallback used to be).

## Tests

Replaced `test_2088_hooks_add_scope_aware.py` (obsolete scope-branch coverage) with `test_4215_1_hooks_add_session_local.py` — 5 tests:
1. session-scoped write lands at `session_state_dir`, not global
2. full E2E self-reload (real Session/HotReloader/dispatcher) — the hook actually fires
3. non-session fallback unchanged (global write)
4. precedence stays additive with the global layer
5. **the actual owner concern** — one session's hook write is invisible to a sibling session's own combine

Gate 5 needed care: `hooks_add`'s own reload-scheduling only touches the *writing* session's `HotReloader` — a sibling session's reloader never gets `request_reload()` called on it, so relying on `apply_pending()` alone leaves it a no-op regardless of where the write landed (a false-green isolation check — verified by observation while writing this test). Fixed by explicitly forcing the sibling's reload too, so a leak into a shared layer would actually surface in its own fired hooks.

## Falsify-verification

Reverted `_hooks_yaml_path`'s `session_state_dir` handling to always fall through to the global path. Confirmed exactly the 3 discriminating tests (1, 2, 5) go RED with the predicted failures (write lands at the global path; the sibling session's inbox contains the leaked "only-for-a" message), while tests 3/4 correctly stay green (they test an orthogonal concern — fallback shape, additive precedence — unaffected by this specific defect). Restored, confirmed all 5 GREEN.

## Docs

`docs/concepts/runtime/hooks.md`, `config-hot-reload.md` (+ `.ja.md`, which currently asserts the same claim), `docs/feature-map.md` all described the #2088 scope-aware write as current behavior — updated in this PR per the hard rule (doc goes stale the moment the mechanism it describes changes).

## Gates (local)

- `ruff check` — clean
- `test_tier_audit.py --strict` — 5/5 OK
- `mypy_ratchet.py` — OK (210/214 baselined, no new findings)
- Broader regression sweep: `tests/runtime/` (1658 passed, 1 pre-existing unrelated skip/failure verified on main too), `tests/tools/` + `tests/core/` (3441 passed, 1 skip — missing optional `trafilatura` extra, unrelated)

## Test plan

- [x] New isolation-focused test suite, falsify-verified RED→GREEN
- [x] `tests/runtime/` + `tests/tools/` + `tests/core/` broad sweep green (pre-existing unrelated failure confirmed on main independently)
- [x] Docs updated same PR (EN + the JA file that asserted the same stale claim)


tya5:
**[lead-coder]** — TESTS-READ 済み（新規 5 本、旧 `test_2088_hooks_add_scope_aware.py` は削除）。**auto-merge を武装します。**

## 🔴 gate 5 の自己申告が、この PR の一番の価値です

```
素朴な isolation テスト   sibling の apply_pending() だけを 呼ぶ
                         → ★hooks_add は 書いた側の HotReloader しか 起こさない
                         → sibling は そもそも 再構築しない
                         ∴ ★書き込みが どこに 落ちても sibling には 現れない
                         ＝ ★共有層に 漏れていても 緑（false-green）
修正                     sibling を ★request_reload + apply_pending で 明示的に 起こす
```

⚠️ **「隔離されている」ことを確かめるテストが、「そもそも見に行っていない」せいで緑になる** — **観察して初めて分かる**種類で、書けた時点では正しいテストと見分けが付きません。**今夜 #4216 で報告された「最初の草案がバグを戻しても緑だった」と同じ族**です。

## 形として効いている点

⚪ **読み側は既に完成していた**（`Session._read_per_session_hooks` が #2285 の 4 層目として同じパスを読む）ので、**書き側だけが動いた** — **新機構ゼロ**という architect のスコープどおりです。
⚪ **`session_state_dir_fn` を lazy callable にした**理由（**spawn 時の fixup が adapter の ctor 後に走るので、eager read は spawn 前の値を凍結する**）が本文にあり、`live_session_id` の既存パターンに合わせています。**「なぜ callable か」が書かれていないと、次の人が単純化して壊します。**
⚪ **非セッション文脈の fallback を残した**のも正しい（`agent_name` の None-fallback と同形）。

## 6 問

- ① Tier: 5 本とも Tier 2（OS 不変条件 — セッション間で hooks が漏れない）。**owner の懸念（hooks は受動的 ∴ 他人の登録が自分に効くのは不可）が根拠**で、第三者の性質ではありません。
- ② 書き写し: なし。 ③ 誰が困るか: **同一 agent の兄弟セッション**。gate 5 が実際にその状況を作っています。
- ④ 走らずに緑か: いいえ（実 Session / 実 HotReloader / 実 dispatcher を通す E2E を含む）。
- ⑤ 何が溜まるか: なし。 ⑥ 宣言 Tier: 一致。

⚪ 旧 `test_2088_...` の削除も妥当です — **#2088 の scope 分岐（GLOBAL / per-AGENT）自体が無くなった**ので、あれは**存在しない分岐を守るテスト**になっていました。

PR #4233, opened 2026-08-11
feat(#4200 1/2): session-layer base_dir override, in front of the agent's own value

tya5:
**[e2e-coder]** — part of #4200 (1 of 2: read side). Design is architect's (resolved thread — 0 open decisions per architect's own retraction comment); this PR implements the resolution mechanism only, no LLM-writable surface yet.

## What

`Session._workspace_base_dir` (previously a bare `self._agent.workspace_base_dir` passthrough) now resolves, in order:
1. this session's own `<session_state_dir>/config.yaml` — a sibling of `self._snapshot_path`, the same file `_read_per_session_hooks` already reads (#2285)
2. the agent's own `.reyn/capability_profiles/<name>.yaml`
3. falls back to the shared `Agent` object's own value (byte-identical for every caller that sets neither override)

Same "layer in front of the shared identity" shape #2103-S1a capability narrowing already uses — the spawned session keeps sharing the same `Agent` object; only the *resolved value* differs per session.

The override read (`_read_base_dir_override`) is a raw YAML read, deliberately **not** routed through `load_capability_profile` — that loader's own docstring says unknown keys are ignored (forward-compat), which for `base_dir` would silently drop the key. Same shape `AgentRegistry.per_session_narrowing` already uses for the same file. A malformed file fails open (logged, skipped) toward the next layer.

## Staleness fix (found independently, not on architect's "measured" list)

`RouterOpContextSource` captured `workspace_base_dir` **eagerly** at `Session.__init__` — before a spawned child's real `snapshot_path` fixup (the registry's post-construction reassignment) runs. Same hazard class #4215①'s `hot_reloader`/`session_id_fn` guard against. Renamed to `workspace_base_dir_fn` (a live callable, resolved in `build()`).

Per lead-coder's request, I swept **every other** `RouterOpContextSource` field for the same class (does its value depend on something the registry reassigns post-construction — the only known such reassignment is `_snapshot_path`/`_session_id`/`_generation_store`/`history_path`): **0 additional found** beyond this one. `session_id_fn`/`turn_origin_fn`/etc. were already correctly lazy; everything else (`events`, `agent_id`, `hot_reloader`, `multimodal_config`, ...) is Agent-stable or object-identity-stable, not snapshot_path-derived.

## Known, documented gap (not fixed here)

`ActionEmbeddingIndex`'s `workspace_root` has no lazy variant — a spawned child's `search_actions` cache still roots at the parent's `base_dir` until that class is separately restructured. This is the action-embedding cache location, not the permission/exec surface #4200 targets (not in architect's "auto-follows" list) — named in a code comment at the site, not silently dropped.

## Falsify-verification

Froze `workspace_base_dir_fn` eagerly (`lambda v=self._workspace_base_dir: v` — the exact pre-fix shape). 4/5 tests went RED with the predicted failure (resolves to the parent's/pre-write value instead of the override) — including gates 1/2/4, which I hadn't expected to be sensitive to eager-vs-lazy but turned out to be (the config.yaml write happens *after* `make_session()` returns, same timing shape as spawn). The regression guard (neither override set) correctly stayed green — an orthogonal case. Restored, confirmed all 5 GREEN.

## Gates (local)

- `ruff check` — clean
- `test_tier_audit.py --strict` — 5/5 OK (first draft flagged 3 as Tier-4 private-state asserts on `session._workspace_base_dir`; rewritten to assert on the real `RouterOpContextSource.build().workspace.base_dir` output instead)
- `mypy_ratchet.py` — OK (210/214 baselined, no new findings)
- Broader regression sweep: `tests/runtime/` (1663 passed), `tests/llm/` + `tests/mcp/` + `tests/interfaces/` (2475 passed) — all green

## Test plan

- [x] New resolution-chain test suite, falsify-verified RED→GREEN
- [x] Broad regression sweep green (`tests/runtime`, `tests/llm`, `tests/mcp`, `tests/interfaces`)


tya5:
**[lead-coder]** — TESTS-READ 済み（新規 5 本）。**auto-merge を武装します。**

## 掃き出しの報告が要求どおりです

```
RouterOpContextSource の 全フィールドを 掃いた   → ★workspace_base_dir 以外 0 件
post-construction で 再代入されるのは _session_id / _generation_store / history_path
   ── session_id_fn は ★既に lazy、他は 誰も 触らない
```

**「0 件」に範囲が付いている**ので、私の側で再導出せずに読めます。

## 🔴 falsify の結果が、あなたの前提を 1 つ壊しています（良い方向に）

> **eager freeze にすると、spawn の gate だけでなく gate 1/2/4（非 spawn）も RED になった。config.yaml の書き込みも `make_session()` の後に起きるので、同じタイミング形だった。**

⚠️ **「これは spawn 特有の問題」という当初の見立てが、実測で否定されています。** **spawn は最も分かりやすい実例であって、原因ではなかった** — 原因は **「値が確定する時刻」と「捕まえる時刻」の食い違い**そのものです。**予測より広く壊れていた**ことを、falsify が先に教えた形です。

## ⚪ `_read_base_dir_override` が `load_capability_profile` を通さない理由

**あのローダは未知キーを黙って捨てる（forward-compat）** ∴ **`base_dir:` もそのまま落ちる** — **raw read が必要**、と docstring に明記されています。**「既存のローダを使う」が正しく見える場面で、それが目的を壊す**ケースで、**理由が書かれていないと次の人が「統一」して壊します。**

## 6 問

- ① Tier: 5 本とも Tier 2。 ② 書き写し: なし。
- ③ 誰が困るか: **spawn された子セッション**（親の base_dir を永久に解決し続ける）。gate 5 がその状況を実際に作っています。
- ④ 走らずに緑か: いいえ。 ⑤ 何が溜まるか: なし。 ⑥ 宣言 Tier: 一致。
- ⚪ **malformed YAML が「クラッシュせず次の層へ落ちる」** ことまで gate があり、**restrict-only なので落ちても実効フロアを超えて広がらない**という理由も添えられています。

⚪ **`ActionEmbeddingIndex` の同クラス gap を「直さず記録した」**判断も正しいです（#4200 のスコープ外）。**拾う場所さえあれば、直さないことは黙殺ではありません。**

PR #4179, opened 2026-08-11
fix(#3903①): foreground exec timeout — default 120s, LLM-extensible up to the operator's own ceiling

tya5:
**[e2e-coder]** — Implements #3903① per owner ruling + architect's conditional approval + lead-coder's design corrections through the discussion.

## What changed

**Default raised, ceiling made operator-controlled (not hardcoded):**
- `SandboxPolicy.timeout_seconds` default: 60 → 120 (owner ruling, industry precedent — Claude Code: 120s default)
- `SandboxPolicy.max_timeout_seconds` (new): default 600, **operator-settable** via `sandbox.policy.max_timeout_seconds` — this is the actual ceiling the LLM's request is checked against, never a hardcoded 600. An operator who narrows it has that narrowing genuinely enforced; the LLM can never widen an operator's own configuration (the exact condition architect's conditional approval required before permitting this reversal of #3962).
- Config load rejects `timeout_seconds > max_timeout_seconds` (self-consistency, not an immovable ceiling).

**`SandboxedExecIROp.timeout_seconds` is back** (#3962 removed it as advertised-but-ignored surface). Unlike the removed field, this one has a real reader: `op_runtime/sandboxed_exec.py` checks it against `policy.max_timeout_seconds` and, when set, applies it. A request above the ceiling is **rejected** (typed error naming the actual configured max) — never silently clamped, which would recreate #3962's shape in a new form (the LLM believing it got the duration it asked for).

**Schema is mechanism-only, no hardcoded numbers** (lead-coder ruling, applying an already-ratified principle — same drift class named 4x tonight: #4158/#4160/#4169/proposal-0067's "don't restate a count, read the registry"): the `timeout` param's description explains the mechanism (can extend / bounded by the operator's ceiling / rejected-with-the-real-value-named if exceeded / background as the escape hatch) without baking in "120"/"600" — a per-deployment config value in static LLM-facing text would go stale silently the moment an operator changed the config. The authoritative number reaches the LLM via the rejection error, at the moment it's needed.

**Doc sync, same PR** (CLAUDE.md hard rule): `control-ir.md`'s `sandboxed_exec` section (asserted #3962's conclusion as current fact) rewritten; `reyn-yaml.md` + `.ja.md`'s `sandbox.policy` key table updated with the new default + new key.

## Explicitly deferred, not silently dropped

[...]

PR #4236, opened 2026-08-11
feat(#4200 2/2): spawn-time base_dir — restrict-only, reject-not-clamp

tya5:
**[e2e-coder]** — part of #4200 (2 of 2: spawn-time write). **Stacked on #4233 (1/2)** — this PR's base is that branch, not main; the diff here is only this PR's own commit. Merge #4233 first, then rebase this onto main (or merge it as-is once #4233 lands, GitHub will retarget automatically).

## What

The LLM-facing `spawn_session` tool gains an optional `base_dir` argument (sibling of `narrowing`, not nested in it). Mirrors `narrowing`'s exact restrict-only shape (#3556):

- `RouterHostAdapter.spawn_session` validates the requested `base_dir` against the SPAWNER's own EFFECTIVE `base_dir` (`Session._workspace_base_dir` — #4233's resolved value, not the Agent's bare default) **before** the child's `config.yaml` is written.
- Subtree containment via `candidate.parents` (not a string-prefix check — `.../parent-other` must not pass against a floor of `.../parent`).
- **#4179 lesson**: a request outside the floor is **rejected** (`status=error`, `kind=base_dir_outside_parent`), never clamped, and the message names the actual parent boundary.
- Persisted via the same single-writer `_persist_session_narrowing`, extended to accept a `base_dir` kwarg → a sibling YAML key (never a second write to the same file).
- Omitted → the child inherits the spawner's `base_dir` unchanged (nothing written to `config.yaml` at all).

## Immediate-parent vs root-anchored restrict-only

Confirmed with lead-coder before implementing: each hop validates against its OWN already-validated floor, so `child ⊆ parent ⊆ ... ⊆ root` holds transitively without a root-anchored check — same argument as narrowing's ∩ composition.

## ⚠️ Not a system-wide invariant (lead-coder review — documented in code comments + here)

Restrict-only gates **only** the LLM-authored `spawn_session` argument. An **operator** directly hand-editing a session's own `config.yaml` (#4233's read side) never routes through this check at all — that's correct (the operator owns the envelope), not a gap. "base_dir never widens" is true only for the LLM-driven spawn path; the last test in the new file (`test_an_operator_hand_edit_of_the_sessions_own_config_bypasses_restrict_only`) demonstrates this explicitly so a future reader doesn't conflate the two.

## Test-double sweep (signature change, not new content)

`base_dir` needed threading into: `tests/_support/router_loop.py`'s `FakeRouterHost`, two existing test files' own inline fake/bound spawn callables (`test_3556`, `test_2103_s1bc_session_spawn_tool`), and the `RouterLoopHost` Protocol's `spawn_session` signature — mypy's `call-arg` finding was the discriminator that caught the missed Protocol update.

The `fp0063_arc_witness` LLMReplay fixture's `tools=` payload changed shape (new `base_dir` schema field) — re-recorded via `REYN_FP0063_ARC_WITNESS_GENERATE=1`, falsify-verified (a fresh run without the env var passes against the re-recorded fixture).

## Falsify-verification (twice)

1. Disabled the restrict-only check outright — both rejection gates go RED with the exact predicted `'spawned'` instead of `'error'`.
2. Replaced the correct containment check with a naive string-prefix check — **only** the sibling-directory test goes RED (the shape it exists specifically to catch: `.../parent-other` wrongly accepted against a `.../parent` floor) — confirming the other 5 tests don't accidentally rely on that same bug.

Restored, confirmed all 6 GREEN both times.

## Gates (local)

- `ruff check` — clean
- `test_tier_audit.py --strict` — 18/18 OK (new file + the 2 touched existing files)
- `mypy_ratchet.py` — OK after fixing the `RouterLoopHost` Protocol (no new findings)
- Broader regression sweep: `tests/runtime/` (1669 passed), `tests/tools/` + `tests/llm/` + `tests/mcp/` (1820 passed) — all green

## Test plan

- [x] New restrict-only/persistence test suite, falsify-verified RED→GREEN (twice — generic disable + the specific prefix-vs-boundary bug shape)
- [x] Broad regression sweep green (`tests/runtime`, `tests/tools`, `tests/llm`, `tests/mcp`)
- [x] fp0063_arc_witness fixture re-recorded + falsify-verified


tya5:
**[lead-coder]** — TESTS-READ 済み（新規 6 本）。**auto-merge を武装します**（#4233 の上に積まれているので、そちらが先に着地します）。

## 条件 3 つとも守られています

```
① 即時の親に対する restrict-only    parent_session._workspace_base_dir（#4200 1/2 の実効値）
② reject-not-clamp                  kind="base_dir_outside_parent"、★親と 要求 の 両方を 名指し
③ 「system-wide の不変条件ではない」  ★operator の 手編集が バイパスすることを 明示的な テストで 証明
```

⚪ **③ をテストで示した**のが特に良いです。**「これは不変条件ではない」は散文で書くと読み飛ばされます** — **operator の hand-edit が実際に restrict-only を越える**ことを緑のテストが示していれば、**次の人が「restrict-only だから広がらない」を前提にした瞬間、そのテストが反証**になります。

⚪ **フロアが解決できない場合に「受け入れない」**（`base_dir_floor_unknown`）も正しい向きです — **検証できないものを通すより、拒否する。**

## 6 問

- ① Tier: 6 本とも Tier 2。 ② 書き写し: なし。
- ③ 誰が困るか: **spawn を出す LLM**（拒否理由が読める）と **operator**（境界が効く）。
- ④ 走らずに緑か: いいえ。 ⑤ 何が溜まるか: なし。 ⑥ 宣言 Tier: 一致。
- ⚪ **sibling ディレクトリのテスト**が効いています — **naive な prefix 比較（`str.startswith`）はこれだけが落とす**形で、**falsify でそのバグ形を名指しして確認**したと本文にあります。**「チェックを無効化して RED」だけでなく「ありがちな間違った実装で RED」まで**やっているのが、この面では正しい強度です。

⚪ **Protocol drift を mypy が拾った**（`RouterLoopHost.spawn_session` に新 kwarg）点、**fixture 再記録も falsify 済み**。**schema 文（EN/ja）が「restrict-only・拒否であって clamp ではない・相対は自分の base_dir 基準・省略すると継承」**まで書いているので、**LLM が次に何を要求すべきか分かる**形です。

PR #4279, opened 2026-08-11
fix(#4275): dedup 7 of 12 inline deadline-poll files to wait_until

tya5:
**[docs-maintainer]** — #4275's remaining item: the 12 inline `deadline = time() + N; while ...` files, deduped to the (now-unconditional, per #4276) `wait_until` helper form.

## Scope: 7 of 12 files converted, 5 deliberately left open

Re-enumerated against fresh `origin/main` (post-#4276) and read every site's actual semantics rather than converting mechanically:

**Converted (7 files, all verified green + falsified)**:
- `test_spawn_routing_detached_fail_mode_2708.py` (4 sites), `test_3049_driver_router_op_intervention_reaches_originator.py` (2), `test_3053_budget_bus_bridge_aware.py` (1 — **this is #4145's formerly-known bucket①② "genuine" file**, re-confirmed NOT an exception), `test_2769_agent_step_intervention_reaches_invoker.py` (2), `test_container_backend_cancel_3862.py` (1) — all straightforward predicate-waits, `deadline+condition` → `wait_until(lambda: condition)`.
- `test_agui_control_filter.py` + `test_3310_n3_remote_switch_parity.py` (3 of 4 `_collect_within` sites) — `AgUiEmitter.stream()` naturally terminates on an `__end__` sentinel that every converted caller always pushes, so the wall-clock window was pure insurance against a hypothetical hang (the exact justification the #4145 ruling rejects). Renamed to `_collect_all`, drops the window entirely.

**Deliberately NOT converted — real design gaps, not guessed at**:
- `test_3310_n3_remote_switch_parity.py`'s 4th `_collect_within` call (`test_switch_announce_precedes_any_new_session_audit_event`) — drives `source.frames()` with no `__end__` ever pushed. No natural termination signal exists; left bounded with an explicit code comment naming this as open.
[...]

PR #4288, opened 2026-08-11
fix(#4280 ①): _run_n_turns_then_shutdown waits on a real predicate

tya5:
**[docs-maintainer]** — #4280 item ①: fixes the 3 files where `_run_n_turns_then_shutdown` (duplicated across `test_1800_wake_drain.py` / `test_1800_c_staging.py` / `test_user_submitted_render_3300.py`) was found to have no real predicate — `turns_done[0] >= n` looked like one but wasn't.

## Root cause

`wake=true` batching lets a single `run_one_iteration()` drain more than one queued message (that's the whole point of the wake-drain primitive). So `turns_done` — incremented once per `run_one_iteration()` call that returned truthy — does not map 1:1 to `n` messages submitted. Discovered by actually converting the wait to unconditional (#4279) and watching `test_wake_drain_equivalence_three_messages` (n=3) hang forever, not by reading the code.

## Fix

The real condition these tests need is "every submitted message has been drained and no turn is in flight" — named directly by two already-public, already-documented, in-memory `Session` accessors:
- `Session.queued_user_messages()` — the undispatched `kind=="user"` inbox queue (reads the snapshot-backed state kept current alongside `append_inbox`/`consume_inbox`, per its own docstring — no WAL-flush timing risk).
- `Session.turn_active` — `True` while a turn is dispatched and in flight (already surfaced for client rendering per its docstring: "used alongside `queued_user_messages`").

`test_1800_c_staging.py`'s wake=False "ride-along" (`kind="hook"`) messages aren't `kind=="user"`, so `queued_user_messages()` doesn't see them directly — but `_drain_to_wake` collects them in the SAME batch as the `kind="user"` wake=True trigger, so once the trigger is gone from the queue, its ride-alongs are necessarily gone with it.

## Verification

**Same-reason-green**, all 3 files (26 tests):
```
python -m pytest <3 files> -q → 26 passed
```
Notably includes the previously-hanging `test_wake_drain_equivalence_three_messages`.

**Falsify**: replaced the predicate with `lambda: False` in `test_1800_wake_drain.py` → `timeout 10 pytest ... --timeout=6` → killed (exit 124), confirming the wait is genuinely gated.

## Test plan
- [x] `ruff check` on all 3 changed files — clean
- [x] `python scripts/test_tier_audit.py --strict <3 files>` — 26/26 OK, 0 Tier-4
- [x] `python scripts/check_tests_path_literal_reference.py` — OK (baseline unchanged)
- [x] Scoped pytest — 26 passed, including the previously-hanging 3-message case
- [x] Falsify: `lambda: False` hangs (kill switch), confirming genuine gating

part of #4280

### DIFF
--- a/tests/core/test_1800_c_staging.py
+++ b/tests/core/test_1800_c_staging.py
@@ -42,6 +42,7 @@ from reyn.interfaces.slash.dispatch import maybe_dispatch_slash
 from reyn.llm.llm import LLMToolCallResult
 from reyn.llm.pricing import TokenUsage
 from reyn.runtime.session import Session
+from tests._async_wait import wait_until
 from tests._support.agent_session import make_session
 from tests._support.slash import local_transport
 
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — **TESTS-READ。auto-merge を武装しました。非 block の指摘 1 件。**

## 偽の述語が、本物の述語に置き換わりました

```
★before  ★turns_done >= n ── ★run_one_iteration の 戻り値を 数える
         ⚠️ ★wake=true の バッチングで ★1 回の run_one_iteration が
            ★wake=true トリガと ★その ride-along を ★まとめて 排出する
         ∴ ★★メッセージ数と 1:1 では ない ── ★#4280 ①の 発見
★after   ★not queued_user_messages() and not turn_active
         ⚪ ★どちらも ★public / documented / in-memory
         ⚪ ★wake=True の user トリガが 消えた 時点で ★ride-along も 同じ batch で 排出済み
         ⚪ ★turn_active が ★batch の 処理も 収まったことを 示す
```

⚪ **`del n  # kept for call-site clarity` と、引数を残した上で使わないことを明示しているのも正直な形です。**

## 非 block の指摘 — 入口での非空虚性

```
★この helper は ★run_task を 作った 直後に ★述語を 待ちます
★呼び出し側（実測）
   ★test_1800_wake_drain.py:129-130  ★submit_user_text → ★helper
   ★:176-178                          ★3 回 submit → ★helper
∴ ★現状は ★入口で queued が 非空 ∴ ★述語は 偽 ── ★空虚には なりません
```

⚠️ **ただしそれは ★呼び出し側の 順序に 依存しています。** **将来 helper の 前に submit しない 呼び出しが 増えると、★述語は 入口で 真になり、★待たずに 抜けます** — **しかも 静かに。**

⚪ **提案（この PR でも 別でも 可）**: **helper の 冒頭に `assert session.queued_user_messages()` を 1 行。** **前提を 明示すると、★空虚な 待ちが ★assert の 失敗として 出ます。**

⚪ **六問**: **① 元の Tier を保持／② 書き写しなし／③ 各テストの元の消費者／④ skip 無し／⑤ ★上限は CI の kill switch のみ（`timeout=5.0` / `3.0` の 引数が 消えた）／⑥ 一致。** **block なし。**

Issue #4293, opened 2026-08-11
tests: test_web_search_no_deny_config_does_not_raise makes a real live network call (hangs the full suite, mislabeled as non-network)


tya5:
**[tui-coder]** — Found while running a full regression sweep for #4174 T3.

## What happened

`tests/core/test_web_search_unified.py::test_web_search_no_deny_config_does_not_raise` hung the local full-suite sweep for ~35 minutes (confirmed via `ps`/`lsof`: the pytest process had an `ESTABLISHED` TCP connection to `yandex.ru:https` at the time of the hang).

## Root cause

```python
def test_web_search_no_deny_config_does_not_raise(tmp_path: Path) -> None:
    """Tier 2: absent web.search config does not block handle_web_search.

    Default Tier 1 behavior: web_search is allowed unless explicitly denied.
    This test uses a non-existent backend to avoid real network calls; the
    backend error occurs AFTER the permission check passes.
    """
    ...
    op = WebSearchIROp(kind="web_search", query="test", max_results=5, backend="duckduckgo")
```

The docstring claims "a non-existent backend to avoid real network calls" — but `"duckduckgo"` **is a real, registered backend** (`reyn.tools.search_backends`), not a non-existent one. The test genuinely calls out to the live internet on every run. It happened to work (fast-failing or fast-succeeding) in whatever environment this test was originally written/verified in, and hung badly here — likely a rate-limit/redirect-loop/slow-DNS condition on `ddgs`'s current upstream routing (the observed connection went to `yandex.ru`, almost certainly a search-result redirect target, not something reyn code names anywhere — confirmed via `grep -rln yandex src tests` = 0 hits).

## Why this matters

- The test's own stated intent ("avoid real network calls") is false — a **Tier 4** violation by this repo's own test-review discipline (a test whose docstring claims something the test doesn't actually do).
- It makes the FULL regression sweep non-hermetic and non-deterministic in duration — a network hiccup on the sweep-runner's machine turns a normally-fast CI/local run into an indefinite hang.
- Worked around locally this session via `pytest ... --deselect tests/core/test_web_search_unified.py::test_web_search_no_deny_config_does_not_raise` (only that one test needs excluding, not the whole file) — the next person to run a full local sweep hits the same wall with no signpost.

## Suggested fix

[...]

PR #4300, opened 2026-08-11
feat(#3846 ③): present's image component renders real pixels via textual-image

### LINKED ISSUE #3846: present の image を配線する（owner 裁定: C = URL 配送）
## ✅ 裁定済み（★owner、★2026-08-09）── **C ＋ config opt-in の scheme 制限**

> owner:「C が良さそうね。あなたの推奨は、reyn に実態を通せと言ってるようだけど、
> C は実態でなくても何を present したかの識別は残るので十分と思う。
> config opt-in で schema 制限の自由をユーザに解放するのはアリだと思う。」

```
★既定 ★C ── ★src は 制限しない（★中和のみ）
★opt-in ★config で ★scheme を 絞れる ようにする
```
⚪ **論拠**: **価値は «reyn が バイトを 持つ» ことでは なく «★何を present したかが 残る» こと。**
**それは ★proxy しなくても 成立する。**
⚪ **owner の 恒久方針（★UX・予測可能性 ＞ セキュリティ、★セキュリティは opt-in）と 整合。**

## ⚠️ lead-coder が 挙げ、★owner 裁定で 吸収された 点（★記録）

```
★file:///etc/passwd を src に 書く 場合
   ★識別は 残る（★何を present したかは 記録される）
🔴 ★ただし 読み取りは «起きた» ── ★client が 読む ∴ ★file.read の permission gate を 通らない
   ★= ★agent が 直接 read_file すると 止まる ものが、★present 経由だと 止まらない
★«識別が 残る» は ★事後の 説明可能性で、★事前ゲートの 代わりには ならない
```
⚪ **∴ opt-in の scheme 制限が ★この 穴の 置き場**です。
**`file:` / `javascript:` を 外したい 操作者は ★そこで 外す。**

## 実装（★tui-coder）

```
★① ★present_renderer.py:114 の 描画を ★src の URL 配送に 繋ぐ
★② ★config に ★scheme 制限の opt-in を 足す（★既定は 制限なし ＝ C）
   ⚠️ ★キーの 置き場は ★実装者判断。★«表示境界» の 関心 ∴ ★safety:（stop 条件）は 不適
   ⚠️ ★新 config フィールドは ★3 面 必須（★example ／ reyn-yaml.md ／ docstring）
      ★#3937 で mirror gate が section-scoped に なった ∴ ★所属節の 中に 書くこと
★③ ★取得は ★build_async_http_client(pin_ssrf=True) を 通す
   ★_network.py が «LLM-supplied な URL を 取る client は これを 使え» と 名指ししている
```

## ⚠️ 未測定（★実装前に）

```
★client 側の 取得経路は «今» 無い（★tui-coder 実測）── ★新規に 作る 側
★slash/image.py の /image の 保存先が media_store か（★architect 未測定）
★SSRF 保護が ★web_fetch の 画像取得«にも» 掛かるか（★同上）
```

**[lead-coder]** — `present` の `image` コンポーネントは宣言済み・未配線です。owner が配送形式を **C（URL）** に裁定しました。

## 現状（tui-coder 測定、#3811 comment 5225621353）

```
catalog       "image": {"src", "alt"}   ← 存在し、validation も binding も通る
描画          present_renderer.py:114 → Text(f"[image: {alt}]", style="dim")
              present/replay.py:150    → f"[image: ...]"
AG-UI wire    render node は meta.nodes に載って渡る（構造は通る）
              運べないのはバイト列
文書          docs/reference/runtime/present.md:76
              「not yet routed to the multimodal delivery path」
```

∴ **新しい op でも新しい語彙でもなく、宣言済みで未配線のものを繋ぐ作業**です。

## owner 裁定

3 択（A ファイルパス / B `data:` URI / C URL）のうち **C**。

- A はリモートで壊れる（クライアントにそのファイルが無い）
- B は `meta.nodes` が監査イベントにも載るため **画像が監査ログに入る**
- **C** は local / remote で同じ形になる

## ⚠️ 着手前に測るべきこと — 「既存の multimodal delivery path」は実在するか

`docs/deep-dives/proposals/0054-present-layer.md:171` は `image` について
**「routes to the *existing* multimodal delivery path」** と書いています。
一方 `present.md:76` は **「not yet routed」**。前者は設計時の意図で、
後者は現状の報告です — **どちらも「その path が実在するか」は答えていません**。

実在しそうな面（未確認、grep のみ）:

```
src/reyn/data/workspace/media_store.py
src/reyn/config/media.py
src/reyn/interfaces/slash/image.py
```

**もし配送機構が既にあるなら、C は「新しいエンドポイントを作る」ではなく
「既にあるものを present に繋ぐ」**になります。作る前に測ってください。

## 決めるべき設計点

- [ ] 🔴 **信頼境界** — `src` が URL になると「リモートのクライアントに取りに行かせる指示」になる。今の中和は制御文字を落とすだけで、**指す先については何も言わない**。誰が発行した URL なら取りに行ってよいのか
- [ ] 配送元は誰か（既存の media_store か、新規か）
- [ ] クライアントが画像オブジェクトを作れること — `textual-image` / `pillow` は現在 reyn の依存にない
- [ ] `present.md:76` と `0054:171` の記述を、着地後の実態に合わせる（片方は現状報告、片方は意図。**両方直す**）

part of #3811

tya5:
**[tui-coder]** — #3846 ③: `present`'s `image` component renders real pixels via `textual-image`.

Closes #3846.

## Dependencies

Owner-decided (#3970): `pillow` + `textual-image` as **REGULAR** dependencies (not extras) — "入れた瞬間に効く" (UX/predictability over a silent text-only fallback, owner's own standing policy). pip resolvability measured in #3970: manylinux(3.11/3.12) / win_amd64(3.11) / macosx_11_0_arm64(3.12) all have prebuilt wheels; the one gap (macosx_10_9_x86_64, Intel Mac) affects neither CI nor the owner's machine.

## What

`_render_image` (`present_renderer.py`) decodes a resolved image's fetched bytes via PIL and wraps them in `textual_image.renderable.Image` — Kitty/WezTerm real pixels or Sixel where the terminal supports either, half-block/unicode approximation otherwise (`textual_image`'s own per-terminal auto-detection, made once per process). Decoding stays within the module's "No I/O" invariant (pure CPU on already-fetched bytes). Falls back to a distinguishable status-line `Text` on decode failure (corrupt/truncated body) or if `pillow`/`textual-image` are unavailable for any reason (defensive only — they are regular deps).

## A real architectural finding: the eager-import ordering fix

`textual_image.renderable`'s terminal-capability auto-detection (Sixel/TGP support) sends a raw escape sequence to the terminal and synchronously reads the reply off `sys.__stdout__`/stdin — a one-time, process-global, import-time side effect. **The library's own docstring** (`textual_image.renderable.sixel.query_terminal_support`) warns this *"will not work anymore once Textual is started"* — Textual runs its own stdin-reading thread that races (and usually wins) this read once the app loop is live.

`present_renderer.py`'s `_render_image` imports `textual_image.renderable` lazily (matching this module's own "No I/O" invariant), which would otherwise defer this exact query to the first image actually presented — i.e. **during a live `app.run_async()` session**, guaranteeing the query loses the race and every image in the TUI falls back to unicode/half-block even on a real Kitty/WezTerm/Sixel terminal.

Fixed by triggering the import in `run_textual_chat` (`textual_chat/app.py`), strictly before `app.run_async()` hands stdin to Textual — verified via a dedicated ordering test (falsify-verified: reverting to no eager import, or moving it after `app.run_async()`, both go RED).

## CI caught 2 more real bugs (`textual-image>=0.13` unsatisfiable on Python 3.11 + a structural try/except placement error)

**Bug 1 — the version floor.** `textual-image>=0.13` (0.13.0/0.13.1/0.13.2) all declare `Requires-Python >=3.12` — unsatisfiable on reyn's own Python 3.11 CI leg, so `pip install -e ".[dev]"` failed at the INSTALL stage (surfacing as 8 simultaneously-red checks — ruff/mypy/pytest-3.11/the sandbox suite/dev==wheel-parity — none of which ran a single test, they all died at install). Caught because my dev venv already had the package installed and so only ever exercised "after install" — a clean-venv-per-Python-version check (`pip install -e ".[dev]"` from scratch) is what actually reproduces it. Floor lowered to `>=0.12`; verified 0.12.0's API surface this file uses is identical to 0.13.2's.

**Bug 2 — a structural try/except placement error, NOT a 0.12.0-specific one.** Verifying "does the downgraded version behave the same" surfaced a real difference: `textual-image` 0.12.0 raises `ValueError('height and width must be > 0')` for a very small source image (reproduced: 8x8/1x1 fail, 16x16 does not). The important part isn't the version — **it's WHERE the exception happens**. Rich renderables are lazily evaluated: `TextualImage(...)` construction just builds a "recipe"; the actual pixel work happens inside `__rich_console__`, a generator Console.print() only iterates once it knows the real render width. `_render_image`'s original `try/except` wrapped only the *construction* call — structurally unable to catch anything `__rich_console__` raises, on ANY version, for ANY reason. 0.12.0's small-image bug is simply what exposed a pre-existing placement error; a future floor bump to 0.13+ does **not** make this wrapper obsolete — any Rich renderable can still raise at print time for its own reasons.

Fixed with `_SafeImageRenderable`, wrapping `__rich_console__` itself (not the constructor) so a print-time failure degrades to the same fallback status-line `Text` the decode-failure path already uses. Verified against the real bug end-to-end in a clean Python 3.11 + textual-image 0.12.0 venv (a 4x4 image that used to crash now renders the fallback text, no exception). The new unit test (`test_safe_image_renderable_catches_a_print_time_failure`) exercises the wrapper against a FAKE raising inner renderable rather than the real bug — the installed `textual_image` version in any given dev environment may or may not reproduce it, so pinning the test to the real bug would be flaky-by-environment.

Both test files' PNG fixtures bumped from 4x4 to 32x32 (safely above the 16x16 threshold) so the SUCCESS-path tests actually exercise success on 0.12.0, not accidentally the failure path depending on which `textual-image` version pip resolved.

## Tests (falsify-verified throughout — including 2 self-corrections caught during my own review)

- `test_image_with_resolved_cache_renders_real_pixels_not_the_status_line`: checks the render-model **object** (`Group.renderables`) for a `textual_image.renderable.Image` instance, not printed glyph output. **Self-caught issue**: my first draft asserted only on printed string content ("no placeholder text", "non-empty output") — falsify-verifying it (reverting `_render_image` to the pre-③ `Text` return) showed the test **stayed green**, because a `Console.print()` of either class produces terminal-formatted text and string-only assertions can't distinguish "real pixel path wired" from "still the old status line". Rewritten to check the object type instead; re-falsified, now correctly goes RED on the same revert.
- `test_image_with_undecodable_body_falls_back_to_a_distinct_status_line`: uses a **truncated real PNG** (not garbage bytes), so the decode actually fails inside the `try` block rather than at `PILImage.open()`'s header sniff. **Second self-caught issue**: while building this test I found the original `pil_image.load()` call (meant to "force the decode now") is **redundant** — `textual_image.renderable.Image`'s own constructor already reads pixel data eagerly (`PixelData.__init__`), so a truncated body raises from `TextualImage(...)` itself regardless of whether `.load()` is called (verified directly in a scratch script). The dead `.load()` line was removed and its comment corrected to describe the real mechanism.
- `test_textual_image_renderable_is_imported_before_app_run_async`: the ordering test described above.
- `test_loaded_state_replaces_the_placeholder_after_resolution` (existing e2e test): updated to use a real PNG body instead of a bare magic-byte stub — ③'s real decode step correctly rejects the old fake bytes, which is the test catching real behavior, not a false failure.

## Docs

`present.md` / `present.ja.md`'s `image` component row updated from "v1 renders a placeholder only" to describe the real pixel-rendering path + the plain-surface placeholder fallback + the distinguishable failure state.

## ⚠️ Known limitation — owner verification required (this is the whole point of the issue)

Real pixel rendering on an actual Kitty/WezTerm/Sixel terminal has **NOT** been visually confirmed — this environment has no real TTY. This PR's own falsify-verification proves the **wiring** is correct (the right object reaches the right renderer, in the right order relative to Textual's stdin ownership) — it cannot prove pixels actually appear on the owner's terminal, which is a third-party rendering promise this environment cannot observe. Per lead-coder's explicit condition on this issue: *"③ は「実物が見えること」が価値の全部なので、私には判定できません"* — the Visual test-plan item below is left **unchecked**, per CLAUDE.md's standing Visual-gate waiver (does not block merge; the owner confirms after, on `main`).

## Test plan

- [x] `python scripts/mypy_ratchet.py` — 0 findings, verified in BOTH a clean Python 3.11 venv (`textual-image` 0.12.0, the real CI scenario) and the Python 3.12 dev venv (0.13.2) — the two Python versions resolve genuinely different `textual-image` releases, so one venv's green is not sufficient
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 22/22 OK
- [x] Falsify-verified all 4 new/changed unit-level assertions (2 caught real weaknesses in my own first draft, described above; the print-time-failure guard is falsify-verified against the real bug in the clean 3.11 venv)
- [x] `tests/core/test_present_op_fp0054_pra.py`, `test_3846_image_fetch.py`, `test_present_renderer_fp0054_prb.py`, `test_3846_image_resolution_e2e.py`, `test_3846_image_url_schemes_config.py`, `test_3846_3_textual_image_eager_import_ordering.py`, `tests/runtime/test_2770_intervention_present_unification.py`, `tests/repo/test_present_sink_ast_guard_2708.py` — 77/77 pass, re-verified in BOTH the clean Python 3.11 venv and the Python 3.12 dev venv
- [x] Full `tests/core/ tests/runtime/ tests/interfaces/ tests/repo/` regression sweep (network-hanging `test_web_search_no_deny_config_does_not_raise` deselected — unrelated pre-existing issue, #4293, fix pending in a separate PR) — 5845 passed, 6 skipped (documented optional-extra skips), 0 failed
- [x] Post-rebase (onto `origin/main` after #4296/#4298 landed): re-verified imports + ruff + mypy_ratchet + the scoped test run, all clean
- [x] Clean-venv `pip install -e ".[dev,mcp,web,fs-watch,observability,builtin-rag]"` (matching CI's exact pytest-job install command) succeeds on BOTH a from-scratch Python 3.11 venv and Python 3.12 venv — the actual bug class this PR's second commit fixes
- [ ] **Owner visually confirms real pixels render on a real Kitty/WezTerm/Sixel terminal** — left unchecked, see "Known limitation" above; not a merge blocker per the standing Visual-item waiver

### DIFF
--- a/docs/reference/runtime/present.ja.md
+++ b/docs/reference/runtime/present.ja.md
@@ -63,7 +63,7 @@ guard/renderer の分割）については [コンセプト: Present レイヤ](
 | `keyvalue` | `rows: [{label, value}]` |
 | `table` | `rows`（bind → 配列）, `columns: [{header, path}]` |
 | `list` | `items`（bind → 配列）, `item_path?`（項目単位パス） |
-| `image` | `src`, `alt?` — v1 では `[image: <alt>]` という dim テキストのプレースホルダーのみ描画され、マルチモーダル配送経路へはまだルーティングされない |
+| `image` | `src`, `alt?` — TUI（`textual_chat`）は `src` を取得し（owner 裁定 C — client 自身が URL を取得、#3846）実ピクセルを描画する（Kitty/WezTerm の実ピクセル、または Sixel。それ以外は半ブロック/unicode 近似 — 端末ごとに自動判定、#3846 ③）。resolution stage の無いサーフェス（`reyn chat --cui`、`reyn pipe`）は `[image: <alt>]` という dim テキストのプレースホルダーを表示する。取得/デコード失敗時は区別可能な `[image failed: ...]` 状態を表示し、プレースホルダーと同じテキストにはならない |
 
 v1 に**対話コンポーネント**（ボタン / フォーム）は存在しない。
 
--- a/docs/reference/runtime/present.md
+++ b/docs/reference/runtime/present.md
@@ -73,7 +73,7 @@ All components are read-only. A blueprint node is `{"component": <name>, ...slot
 | `keyvalue` | `rows: [{label, value}]` |
 | `table` | `rows` (bind → array), `columns: [{header, path}]` |
 | `list` | `items` (bind → array), `item_path?` (per-item path) |
-| `image` | `src`, `alt?` — v1 renders an `[image: <alt>]` dim-text placeholder only, not yet routed to the multimodal delivery path |
+| `image` | `src`, `alt?` — the TUI (`textual_chat`) fetches `src` (owner-decided delivery form C — client fetches the URL itself, #3846) and renders REAL pixels (Kitty/WezTerm true pixels or Sixel, half-block/unicode approximation otherwise, auto-detected per-terminal — #3846 ③). Surfaces without a resolution stage (plain `reyn chat --cui`, `reyn pipe`) show an `[image: <alt>]` dim-text placeholder; a failed fetch/decode shows a distinguishable `[image failed: ...]` state, never the same placeholder text. |
 
 There are **no interactive components** (no buttons / forms) in v1.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -227,7 +227,16 @@ dependencies = [
     # approximation only — see #3970's own reinvention-comparison for why a
     # hand-rolled half-block renderer is not a "lighter version" of this
     # (different feature, not a smaller implementation of the same one).
-    "textual-image>=0.13",
+    # Floor is 0.12 (not 0.13+): 0.13.0+ declares `Requires-Python >=3.12`,
+    # unsatisfiable on reyn's own Python 3.11 CI leg/support floor — a real
+    # install-stage failure caught only by testing a CLEAN venv on 3.11 (a
+    # venv with the package already installed, or a 3.12-only check, both
+    # stay green and miss this). Verified 0.12.0 has the identical API
+    # surface this file uses (`renderable.Image(pil_image, width=...)`, the
+    # same import-time no-textual / terminal-auto-detection module
+    # structure, and the same TextualImage-constructor-forces-eager-decode
+    # behavior `present_renderer.py`'s decode-failure path relies on).
+    "textual-image>=0.12",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -208,6 +208,26 @@ dependencies = [
                               # Windows backend writes CF_UNICODETEXT directly.
                               # Linux still needs xclip/xsel/wl-clipboard on
                               # PATH — pyperclip does not vendor them either.
+    # #3846 ③ / #3970: owner-approved REGULAR dependency (not extra) —
+    # "入れた瞬間に効く" (UX/predictability > a silent text-only fallback,
+    # owner's own standing policy). pip resolvability measured (#3970):
+    # manylinux(3.11/3.12) / win_amd64(3.11) / macosx_11_0_arm64(3.12) all
+    # have prebuilt wheels (no compiler needed); the one gap is
+    # macosx_10_9_x86_64 (Intel Mac, source build needs libjpeg/zlib
+    # headers) — neither CI nor the owner's machine is affected.
+    "pillow>=11.0",
+    # `textual_image.renderable` (NOT `.widget`) — measured (#3970): imports
+    # NO textual module at all (0/147 `sys.modules` entries), a real Rich
+    # renderable (`__rich_console__`) buildable straight from a PIL.Image
+    # already in this process — same shape `present_renderer.py` already
+    # returns for every other node kind. pillow is the dominant dependency
+    # weight either way (14MB); textual-image itself is close to free on top
+    # of it and is the ONLY option of the measured candidates that reaches
+    # real pixels (Kitty/WezTerm TGP, Sixel) rather than a half-block/unicode
+    # approximation only — see #3970's own reinvention-comparison for why a
+    # hand-rolled half-block renderer is not a "lighter version" of this
+    # (different feature, not a smaller implementation of the same one).
+    "textual-image>=0.13",
 ]
 
 [project.urls]
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -4554,6 +4554,31 @@ async def run_textual_chat(
     regardless. Returns so the driver's caller can tear the transport down
     + print the cost summary.
     """
+    # #3846 ③: trigger textual_image.renderable's ONE-TIME, process-global
+    # terminal-capability auto-detection HERE — strictly BEFORE
+    # app.run_async() hands stdin to Textual. That auto-detection (Sixel/TGP
+    # support) sends a raw escape sequence to the real terminal and
+    # synchronously reads the reply off `sys.__stdout__`/stdin; the LIBRARY'S
+    # OWN docstring (`textual_image.renderable.sixel.query_terminal_support`)
+    # warns this "will not work anymore once Textual is started" — Textual
+    # runs its own stdin-reading thread that races (and usually wins) this
+    # read once the app loop is live. `present_renderer.py`'s `_render_image`
+    # imports `textual_image.renderable` lazily (its own "No I/O" module
+    # invariant), which would otherwise defer this exact query to the first
+    # image actually presented — i.e. DURING a live app.run_async() session,
+    # guaranteeing the query loses the race and every image in the TUI falls
+    # back to unicode/half-block even on a real Kitty/WezTerm/Sixel terminal.
+    # Importing here, before Textual owns stdin, lets the query succeed and
+    # the module-level `Image` binding (see that module's own `__init__.py`)
+    # resolve to the real terminal's actual capability once, for the whole
+    # process. A missing/failed import is swallowed — image nodes simply
+    # keep falling back to the pre-③ status-line text, same as an
+    # environment without pillow/textual-image at all.
+    try:
+        import textual_image.renderable  # noqa: F401,PLC0415
+    except Exception:
+        pass
+
     # #3671: the framework's own startup begins here — terminal setup, first
     # layout, first paint. Everything before it is reyn assembling things;
     # everything after it is Textual, and the two were indistinguishable while
--- a/src/reyn/interfaces/repl/present_renderer.py
+++ b/src/reyn/interfaces/repl/present_renderer.py
@@ -118,19 +118,32 @@ def _render_node(node: dict, image_cache: "dict[str, Any] | None" = None) -> "An
 
 
 def _render_image(node: dict, image_cache: "dict[str, Any] | None") -> "Any":
-    """The `image` component (#3846 ②) — a PURE dict lookup, never a fetch.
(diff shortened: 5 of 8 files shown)


tya5:
**[lead-coder]** — **TESTS-READ。checks の FAILURE 0 件を確認し、auto-merge を武装しました。**

## 3 本とも、自分が何を守れないかを書いています

```
★test_textual_image_renderable_is_imported_before_app_run_async
   ★eager import が ★app.run_async() の 前に 起きることを assert
   ⚪ ★falsify 済み ── ★import を 後ろに 動かすと 赤
   ⚠️ ★これが 無いと、★実端末が Kitty/WezTerm でも ★常に 検出負けして 半ブロックへ 落ちる
      ── ★ライブラリ自身の docstring が「Textual 起動後は 機能しない」と 明記している 制約

★test_image_with_resolved_cache_renders_real_pixels_not_the_status_line
   ★Group.renderables の ★オブジェクト型を 直接 検査（★isinstance(_SafeImageRenderable)）
   ⚠️ ★docstring が 明示: ★「printed string の assert では
      ★『実ピクセル経路が 配線された』と『まだ status line』を 区別できない」
   ⚪ ★★1 回目の 草稿が まさに それで、★実装を pre-③ に 戻しても 緑だった ── ★自己発見・書き直し
```

⚠️ **「実装を戻しても緑」は、gate が守っていないことの決定的な証拠です。** **それを自分で見つけて書き直し、経緯を docstring に残したのが正しい形です。**

## fixture の 32x32 に理由が書かれています

```
★PILImage.new("RGB", (32, 32)) ── ★4x4 では ない
★理由  ★textual-image 0.12.0（★3.11 で pip が 入れる 版）の ★実バグ
        ★小さい画像で ★ValueError('height and width must be > 0') を ★print 時に 投げる
```

⚪ **fixture の 寸法に 理由が 書いてある**のは珍しく良い形です。**書いていなければ、次の人が「4x4 で十分」と縮めて、★3.11 でだけ 落ちます。**

## 本番の欠陥が 1 件、この PR で見つかり直っています

```
★0.12.0 は ★print 時（__rich_console__ 内）に 投げる
★既存の try/except は ★construction だけを 囲んでいた ── ★捕捉できない
→ ★本番で 小さい画像が 来ると ★未捕捉例外で クラッシュ
★_SafeImageRenderable（__rich_console__ 自体を ラップ）で 修正
```

⚠️ **Rich の renderable は ★遅延評価**です。**construction は レシピを 作るだけで、★描画は Console が 幅を 知って iterate する 時点。** **∴ construction を 囲む try/except は、★renderable が 投げる 例外を 構造的に 捕まえられません。** **これは textual-image 固有ではなく、★Rich renderable を 返す 全経路に 掛かる 形です。**

## 六問

```
① Tier   ★Tier 1 / Tier 2 ── ★宣言と 一致
② 書写し ★否 ── ★型と 出力を 見る
③ 誰が失う ★Kitty/WezTerm の operator ── ★eager import が 無いと 常に fallback
④ 走らず緑 ★否 ── ★clean 3.11 venv ＋ 既存 3.12 venv の ★両方で 検証
⑤ 蓄積   ★無し
⑥ 宣言真 ★真
```

⚠️ **Test plan の最終項目（実端末での実ピクセル確認）は未チェックのままです。** **これは正しい状態です** — **私には判定できません。** **owner が朝に実端末で確認します。**

PR #4344, opened 2026-08-11
fix(#4343): prefix-match fallback for partially-masked secrets + reyn.log reachability

tya5:
**[tui-coder]** — Prefix-match fallback for `secret_scrub.py` + live `.reyn/logs/reyn.log` reachability measurement.

## Background

Dogfood pass 2 (lead-coder dispatch) deliberately triggered a real provider failure and checked the 3 sinks #3830 named (TUI/outbox text, `.reyn/events`, `.reyn/logs/reyn.log`). The corrected finding (after an initial methodologically-flawed synthetic-exception test was self-caught and redone): the exact-match scrub is correct, but a case exists where litellm's own exception construction has already partially masked a known secret's tail before reyn ever sees the text — the full value is then no longer a literal substring, so the exact-match `.replace()` finds nothing.

## ① The fix

`scrub_secrets` gets a second pass after the full-value replace: try each known secret's leading `_SECRET_PREFIX_LENGTH` characters too. Still "the actual value at hand" — never a guessed pattern like "starts with sk-", so the module's own no-pattern-guessing rule stays intact.

`_SECRET_PREFIX_LENGTH = _MIN_SECRET_VALUE_LENGTH * 2 // 3` (= 8) — same no-real-value-inspection discipline `_MIN_SECRET_VALUE_LENGTH` itself documents, derived as a fraction of that existing constant. Reasoning is in the code comment (trade-off shape: too short over-redacts ordinary text, too long stops matching once the tail is masked).

Two new tests: the partial-mask case now redacts, and the common full-value case still wins (no duplicate redaction artifact when both passes could apply).

## ② `.reyn/logs/reyn.log` live-reachability measurement (the remaining open item from dogfood pass 2)

Dogfood's own harness never reaches this sink — its non-interactive invocation keeps `is_interactive=False`, so `_setup_interactive_logging` (which routes the root logger to the file) is never installed. That's a "not exercised" gap, distinct from "verified safe."

Built a standalone pty-driven session (real TTY on both stdin/stdout, sanitized ambient env so no real credential could leak in, an explicit synthetic fake key) against a scratch project pointed at an unreachable `api_base`. Confirmed:
- `.reyn/logs/reyn.log` DOES get written to by a real interactive session (`session.py`'s top-level router-loop exception handler logs via `logger.exception`).
- That handler logs the SAME exception object `llm.py`'s `recorded_acompletion` already scrubbed in place — so the file sink inherits safety by construction, same as the other two sinks #3830 already covered.
[...]

PR #4345, opened 2026-08-11
docs(verification-hazards): fold tonight's 6 instances into existing sections

tya5:
**[docs-maintainer]** — lead-coder dispatch: fold tonight's real instances into `docs/deep-dives/contributing/verification-hazards.md`.

## Scope decision (reported and approved before writing)

Read the whole doc (21 sections) before proposing anything. Recommended
**0 new sections, all 6 candidates as real-instance additions to existing
faces** — the doc is already long, and every candidate is a corroborating
example of a shape the doc already names, not a genuinely new one.
Lead-coder approved this scope, with one placement question resolved by
re-reading §1's own row definitions (see below).

## Placements, and why

- **§1 table, "Environment can't witness"**: #4343 — a reproduction that
  raised a litellm exception WITHOUT going through
  `scrub_exception_in_place` (the real production call path) reported
  already-fixed code as still leaking; self-corrected once rebuilt
  through the real path.
- **§1 table, "Claim has no owner"**: #4333 — CLAUDE.md asserted the
  sandbox 3-tuple axis contract was "CI-conformance-only"; nothing on the
  CI side ever checked that claim was true. Lead-coder's own re-read
  caught my initial placement error here: I had put this in "Environment
  can't witness," but the deny-gate job's environment COULD witness the
  risky path fine — once the file was added to its `pytest` invocation
  it ran and passed (34/34). The gap was ownership (nobody checked the
  claim against reality), not environment capability — a real
  distinction between the two rows, not interchangeable.
- **§3 (two zeros)**: extended with the mirror case — a NONZERO hit is
  not itself an answer either, only a classification that isn't decided
  until opened (#4285 correctly read as "not started" at 0 hits; #4287,
  same method, 1 hit classified as "just a mention" without opening the
  PR, which had actually landed the fix in full).
- **§6 (vacuity guard)**: probabilistic vacuity as a sibling to
  empty-set vacuity — a `random.sample` exact-permutation pool meant a
  retry test's own claimed scenario held only half the time, compounded
  by an `@requires_tte` skip marker meaning the file never ran in CI
  either (#4291/#4330).
- **§18-B (misidentification)**: the INSPECTING COMMAND itself can
  decide whether a version gets recorded — `git show origin/main:<path>`
  names the ref in the command; bare `python3 <path>` silently reads the
  working tree with no record of which version ran (#4215).
- **§21 (search shape decides population)**: a 4th axis — two
  independently incomplete lists (a doc's config/ table, the runtime's
  own hot-reload IN-set), cross-referenced to 8, but the true count
  (measured against actual reader/writer code) was 7+1 — a THIRD number,
  since both starting lists were missing something the OTHER was ALSO
  missing (#4337). Updated the section's "three axes" framing to four,
  and its "what actually closed them" list, to keep the section's own
  internal cross-references consistent.

## Discipline check (the doc's own bar)

"If a hazard here can't cite a real instance and a measured detection, it
doesn't belong." All 6 have a real, dated instance and a concrete
detection technique that closed them — none are hypothetical or
in-flight (verified #4343's own follow-up fix, #4344, already landed
before I wrote this).

## Verification

- `rm -rf site && mkdocs build --strict` — clean (fresh rebuild; only
  pre-existing ja-parity anchor gaps this doesn't touch).
- `python scripts/check_doc_anchors.py` — clean.
- `python scripts/check_tests_path_literal_reference.py` — clean (no new
  `tests/...py` literals introduced).
- No `.ja.md` sibling exists for this doc — no mirror obligation.

## Test plan

- [x] scope proposal reported to lead-coder and approved before writing
- [x] `mkdocs build --strict` clean (fresh rebuild)
- [x] `check_doc_anchors.py` clean
- [x] `check_tests_path_literal_reference.py` clean
- [x] every added instance cites a real, dated issue/PR and a detection
      technique, per the doc's own stated bar
- [x] §21's internal cross-references (axis count, "what closed them"
      list) updated to stay consistent after adding a 4th instance

### DIFF
--- a/docs/deep-dives/contributing/verification-hazards.md
+++ b/docs/deep-dives/contributing/verification-hazards.md
@@ -20,8 +20,8 @@ and a measured detection, it doesn't belong in this doc.
 | Face | What's missing | Instance |
 |---|---|---|
 | **Record is a lie** | The claim itself is false | `landlock.py` blamed network denial on "the no-network-fd / proxy gate" — a named mechanism that appears nowhere in the repo but that one comment (#3031). What actually denied `connect()` was a seccomp default-deny, itself skipped under `allow_subprocess=True` (#3030). |
-| **Environment can't witness** | A green test never ran the risky path | The Landlock shim called `Ruleset` APIs (`add_path_beneath_rule` etc.) that don't exist in the pinned `landlock==1.0.0.dev5` — every call raised `AttributeError` in production for 41 days, while its own test called the shim's internals directly, bypassing the broken production entry point (#2980). |
-| **Claim has no owner** | No one on the claimed subsystem's side checks it | Same `landlock.py` case: a doc/comment in subsystem A asserting subsystem B's behavior, with no owner on B's side to catch it wrong — "plausible and unowned" is why it survived. |
+| **Environment can't witness** | A green test never ran the risky path | The Landlock shim called `Ruleset` APIs (`add_path_beneath_rule` etc.) that don't exist in the pinned `landlock==1.0.0.dev5` — every call raised `AttributeError` in production for 41 days, while its own test called the shim's internals directly, bypassing the broken production entry point (#2980). A repro can fail the SAME way a test can: investigating whether a secret leaks through `Session._run_router_loop`'s TUI/outbox path (#3830), a reproduction was built that raised the exception WITHOUT going through `scrub_exception_in_place` — the actual production path — and reported the fixed code as still leaking. Self-corrected once rebuilt to raise through the real call path, which produced the REDACTED text the first repro never could have shown either way (#4343). The repro environment, not a test, was the thing that couldn't witness the real path this time. |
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — **auto-merge 武装しました。** `tests/` を触っていないので TESTS-READ は該当しません（doc 1 ファイルのみ）。

**私が出した疑問に、正しい方向で答えています。** #4333 を「Claim has no owner」に置き直し、**しかも 2 つの行の境界を明示的に書いた**のが良い:

> **Distinct from the "environment can't witness" row above**: the deny-gate job's environment COULD witness the risky path fine — once the file was added to its `pytest` invocation it ran and passed (34/34) — **the gap was never a capability gap**, only that nothing on the CI side had ever checked the claim against what actually ran

**この 1 文が、次に同じ判断をする人の道具になります。** 「環境が witness できなかったのか、誰も検査していなかったのか」は**症状が同じで処方が違う** — 前者は環境を整える、後者は検査する持ち主を作る。**私は疑問として出しただけで、境界の言語化はあなたがやりました。**

**#4343 を Environment 行に置いた説明も正確**です:

> The **repro environment, not a test**, was the thing that couldn't witness the real path this time.

**「テストではなく再現が」**という限定が効いています。既存行はすべてテストの話でしたが、**同じ顔が再現手順にも現れる**ことが、これで doc に入りました。

**検出手も書いてあります** — 「found via #4331's skip census **cross-referenced against which job collects which file**」。doc 自身の規律（real instance ＋ measured detection が無ければ入れない）を満たしています。**「census を取った」だけでは足りず、「どのジョブがどのファイルを収集するかと突き合わせた」まで書いてあるのが、次の人が再現できる粒度**です。

§21 を 3 axis → 4 axis に整合更新し、「what actually closed them」の list まで足したのも、**追記が既存の構造と噛み合っている**証拠です。新規節 0 で 6 件を吸収したのは、規模判断として正しかった。

PR #4363, opened 2026-08-12
feat(3671): break tui-boot into construct/compose/hydrate/other

tya5:
**[tui-coder]** — part of #3671

## What

Breaks `tui-boot` into four named sub-stages, mirroring the same bracket-and-residual shape `client-prep:other` already established (the #3735 discipline: keep the wide bracket as ground truth, name what narrow sub-brackets do NOT explain as `:other` rather than letting it silently fall out of measurement).

Owner git-bash re-measurement: total startup 25.5s, of which `tui-boot` alone was 23.9s (94%) — previously a single opaque span with no internal breakdown.

- `tui-boot:construct` — `TextualChatApp.__init__`
- `tui-boot:compose` — `compose()`, which builds the entire widget tree
- `tui-boot:hydrate` — `on_mount()`'s `_hydrate_from_history()` call (#3273)
- `tui-boot:other` — residual = wide − sum(named), computed in `mark_first_frame()`

`tui-boot` itself is untouched as the wide ground-truth bracket — still recorded unconditionally, still printed as its own report line, so the number already being compared against the owner's macOS baseline does not move or disappear. The invariant this PR must hold: named sub-stages + `:other` always sum to exactly `tui-boot`, never spilling into the separate process-wide `unaccounted` bucket. Covered by `test_tui_boot_named_substages_plus_other_equal_the_wide_span`.

## Double-counting fix

Adding `tui-boot`'s own breakdown alongside the pre-existing unconditional `tui-boot` recording would make `StartupTiming.total_seconds` sum that one interval twice — inflating `total_seconds` past wall time and clamping `unaccounted_seconds` to 0 even when a genuine coverage gap exists elsewhere. Fixed by excluding `tui-boot` from `total_seconds`'s sum (kept as its own printed line, unlike `client-prep` which is never separately recorded). Falsify-verified: reverted the exclusion, confirmed `test_total_seconds_does_not_double_count_tui_boot` failed with the predicted ~2x inflation, restored.

## Doc-accuracy fix (same PR, per CLAUDE.md's doc-drift rule)

`mark_app_constructed()`'s docstring called the span "the framework's own startup — terminal setup, first layout, first paint" — inaccurate, since the span also runs reyn's own `TextualChatApp.__init__`, `compose()` (builds the entire widget tree), and `on_mount()` (theme setup + `_hydrate_from_history()`). The construction site's comment in `app.py` made the same claim. Both rewritten to name what's actually inside the bracket — this PR's own breakdown is what makes the inaccuracy fixable.

## Scope

Instrumentation only. No behavior change.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict tests/runtime/test_startup_timing_3671.py` — 27/27 OK, 0 Tier-4 violations
- [x] `python scripts/verify_module_docstrings.py` on both changed `src/` files — OK
- [x] Falsify-verified the double-counting fix (reverted → confirmed predicted failure shape → restored → confirmed green)
- [x] `pytest tests/interfaces tests/runtime tests/core/test_3846_3_textual_image_eager_import_ordering.py tests/repo/test_stream_client_single_writer_boundary.py` — 3307 passed, 4 skipped, 1 failed (`test_textual_image_renderable_is_imported_before_app_run_async`, pre-existing on `main`, unrelated to this change — textual_image import-ordering)
- [ ] Visual (waived per owner's standing waiver, 2026-08-08 — merge proceeds, owner confirms after on `main`)

Held for lead-coder's TESTS-READY / explicit merge approval, not self-merging. Owner intends to re-measure on the git-bash machine once this lands.

---

## Reviewer's blocking item (lead-coder)

- [x] 🔴 **`test_tui_boot_named_substages_plus_other_equal_the_wide_span` の許容幅 `< 0.02` を外す。** 環境依存の定数で、owner の恒久指示「テストは時間の上限を持たない」に触れます。★不要でもあります — 比較対象をテストの壁時計から**記録された `tui-boot`** に変えれば `named + other == wide` は厳密に成立し（`:other = max(0, wide - named)`）、falsification の力は変わりません。詳細はレビューコメントに。
  → Fixed in `4150590df` — assertion now compares `named_sum == startup_timing.TIMING.elapsed("tui-boot")` (exact, no tolerance). Falsify-verified: the gap-drop regression this test targets still fails the exact-equality check the same way it failed the old tolerance check.
- [x] 🔴 **Same shape in the pre-existing sibling test** — `test_named_substages_plus_other_equal_the_wide_span` (client-prep side) still had `abs(named_sum - (wall_end - wall_start)) < 0.02`. Fixing only the new mechanism and leaving the old one is the #3905 shape (lead-coder review, follow-up).
  → Fixed in `a56e7bc53`. `client-prep` has no independently recorded wide span the way `tui-boot` does, and using the private `_ASYNC_ENTERED_AT`/`_APP_CONSTRUCTED_AT` timestamps would contradict this test's own stated design intent ("a public-surface witness, not an assertion on internal state") — so the tolerance was replaced with a monotonic `>=` against the known injected `time.sleep()` total instead, the same non-flaky pattern `test_the_context_manager_records_elapsed_time` already uses in this file. Falsify-verified: simulating the #3735 gap-drop regression (sum without `:other`) measures ~0.048s against a 0.06s floor and correctly fails; the real sum measures ~0.071s and passes.

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -1251,205 +1251,210 @@ class TextualChatApp(App):
         self._cursor_position = value
 
     def compose(self) -> ComposeResult:
-        # Held so the frame pump can start/stop the per-entry BODY animation
-        # (``animate_entry``/``stop_entry_animation``) that drives a RUNNING tool
-        # row's live spinner + elapsed (Phase ②).
-        self._flow: "FlowView[OutboxMessage]" = _CursorFlowView(
-            model=self.conversation,
-            presenter=self._presenter,
-            decorator=ReynGutter(
-                frame_period=_RUNNING_FRAME_PERIOD,
-                # The app's own injectable clock, as the right gutter already
-                # takes — production passes ``time.monotonic``, so this is the
-                # same behaviour, and it lets a test drive the blink instead of
-                # sleeping through a real frame period.
-                clock=self._clock,
-                # #3530: blink a reply that is still receiving chunks. Read
-                # live off ``_streaming_replies`` each repaint — the same
-                # record the terminal completion frame pops — so the marker
-                # can never disagree with whether the stream is actually
-                # still open.
-                is_streaming=self._is_streaming_entry,
-            ),
-            gutter_width=_GUTTER_WIDTH,
-            # Phase ④ (#3283): the RIGHT gutter shows per-entry elapsed time
-            # (tool rows) AND the row's turn's real prompt/completion token
-            # split (agent reply rows, via the keyed per-turn lookup) — see
-            # ReynRightGutter and its two halves for the content-set decisions.
-            # additive flowview params; the LEFT gutter/state contract above is
-            # untouched.
-            right_decorator=ReynRightGutter(
-                clock=self._clock,
-                usage_lookup=self._turn_usage,
-                # #3490, moved to this side by #3526 (owner directive): the
-                # addressed-row rail. Read live off the view each repaint (not
-                # pushed in on every move) so the gutter can never hold a stale
-                # copy of which entry is current.
-                is_marked=self._is_addressed_entry,
-            ),
-            right_gutter_width=RIGHT_GUTTER_WIDTH,
-            spacing=1,
-            anchor=Anchor.STICKY_BOTTOM,
-            # Native running-blink: FlowView owns the animation clock and
-            # re-invokes the time-based ReynGutter each tick (no app-side timer).
-            animation_fps=self.ANIMATION_FPS,
-            # #3476 ②: a fresh session previously opened onto a blank void
-            # above the composer (owner design review). flowview 0.6.0's
-            # empty state shows this hint across the viewport while the model
-            # has no entries and clears itself the moment the first entry
-            # lands — no app-side show/hide wiring to drift.
-            empty=empty_state_hint(),
-            empty_align="middle",
-            # #3476 ④: fire ReachedTop while the top edge is still a few rows
-            # away, so the next history page is in place by the time the user
-            # actually arrives at it.
-            reach_threshold=3,
-            # #3476 ⑥: the keyboard cursor (the visual affordance #3470
-            # deferred to this PR). Reached the SAME way #3470 already
-            # established — Shift+Tab focus-cycling into FlowView, Esc back
-            # out (#3399 gate) — never a new focus path. flowview owns
-            # ↑/↓/PageUp/PageDown/Home/End moving the cursor once FlowView
-            # has focus; while it doesn't, those keys are unaffected (the
-            # composer's own PageUp/PageDown delegation, #3470, calls
-            # actions on ``self._flow`` directly and never depends on this
-            # flag).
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — **TESTS-READ**（diff のテストを読み、六問を各テストに当てました）。**併せて、私が block しかけて★測定で否定した点も記録します。**

## 私が疑い、確かめて否定した点（記録に残します）

`compose()` は **generator** で、その 13 個の `yield` が `with stage("tui-boot:compose")` の内側にあります。**generator を `with` で括ると、消費側のペースまで測る**のが一般則なので、「Textual の mount 処理が `:compose` に混入し、`:other` から差し引かれて、**reyn の取り分を過大に見せる**」＝ この PR が直そうとしている誤配分の逆転を疑いました。

**否定されました。** インストール済み Textual の実装:

```
textual/widget.py:4718   widgets = [*self._pending_children, *compose(self)]
```

★**`[*...]` で mount 前に完全に materialize されます。** ∴ 消費側のペースは混入せず、`:compose` は reyn の widget 構築だけを測ります。**block しませんでした。**

## 六問

**`test_tui_boot_named_substages_plus_other_equal_the_wide_span`**
①Tier 2（レポートの被覆不変条件＝reyn 自身の契約、`client-prep` の既存対と同型）②実装の書き写しではない — 期待値を**テスト自身の `perf_counter`** から作っており、モジュールの `wide - named` を写していない ③#3735 の実在回帰が consumer、レポートを読む operator も ④skip 無し ⑤`sleep(0.01)`×4 で有界 ⑥宣言と一致。

**`test_total_seconds_does_not_double_count_tui_boot`**
①Tier 2 ②`< wide * 1.5` は比率で、実装式の写しではない ③二重計上は**実際に起きた欠陥**（tui-coder が自己発見）④skip 無し ⑤蓄積無し ⑥一致。**falsify も実施済**（除外を戻す→約 2 倍に膨張して失敗を確認→復元）と報告あり。

## 🔴 1 点だけ直してください（本文に項目を追加します）

`test_tui_boot_named_substages_plus_other_equal_the_wide_span` の

```python
assert abs(named_sum - (wall_end - wall_start)) < 0.02
```

**`0.02` は環境依存の定数**です。owner の恒久指示「**テストは時間の上限を持たない・分解せよ**」に触れます。この値は「マシンがどれだけ遅れても許す幅」であって、**遅いマシン / 負荷時に GC や scheduler の割り込みが `wall_start`〜`mark_app_constructed()` の間に入れば落ちます。**

★**しかも不要です。** 検証したい不変条件は `named + other == wide` で、`:other = max(0, wide - named)` ですから、**`wide` そのものと比べれば厳密に成立**します:

```python
assert named_sum == startup_timing.TIMING.elapsed("tui-boot")
```

**壁時計との比較が slop を持ち込んでいるだけ**で、falsification の力は変わりません（`:other` を削れば `named_sum` は 3 つの sleep 分にしかならず、`wide` に届かない ＝ 同じように落ちます）。∴ **許容幅ごと消せます。**

⚠️ 浮動小数の同値が気になるなら `pytest.approx` で構いませんが、**比較対象を「テストの壁時計」から「記録された `tui-boot`」に変えること**が要点です。

PR #4399, opened 2026-08-12
fix(4395): point TIKTOKEN_CACHE_DIR at a reyn-owned persistent directory

tya5:
**[docs-maintainer]** — part of #4395's ① (follow-up to #4398, the cooldown fix — that one shipped first per explicit priority; this one reduces how often a network fetch is even attempted).

## What

litellm's own `default_encoding.py` already redirects `TIKTOKEN_CACHE_DIR` to its bundled tokenizer snapshots at import time, but a live owner repro (via lead-coder) surfaced two independent ways that's insufficient on its own:

1. **Import order** — that module only runs once something imports `litellm.litellm_core_utils.token_counter`; `litellm_provider.py`'s own `tiktoken.get_encoding(...)` call bypasses that chain entirely if it's the first tiktoken-touching code in the process.
2. **Self-corruption** — tiktoken's own cache-hit check is a sha256 comparison (`tiktoken/load.py`); a version mismatch makes tiktoken **delete** the mismatched file and re-fetch. If that file lives inside litellm's own package directory, tiktoken mutates an installed package's data files on every run.

## Fix

`reyn/__init__.py` (already the earliest point guaranteed to run before the first `import litellm`, per its own existing `LITELLM_LOCAL_*` precedent) now `setdefault`s `TIKTOKEN_CACHE_DIR` to a **reyn-owned, persistent** directory (`~/.reyn/cache/tiktoken/`) — not the OS temp dir (tiktoken's own fallback, periodically cleared) and not litellm's own package directory (what an earlier draft of this fix pointed at, before the self-corruption finding). This is not re-deciding litellm's own call (#3905's "don't take over a third party's responsibility") — litellm has already decided offline-by-default; this only makes that decision survive the import-order gap and the fragility of WHERE it points.

## Tests

`tests/core/test_4395_tiktoken_cache_dir_default.py` (Tier 2, subprocess-based — a package-`__init__` env-var side effect only runs once per process, so in-process testing would mask the exact import-order bug): the var gets set, it points outside both the OS temp dir and litellm's own package directory, and an operator's own explicit value is respected (`setdefault`, matching the existing `LITELLM_LOCAL_*` convention).

**Falsified**: both substantive tests fail against the pre-fix `__init__.py` — one with a bare `KeyError` (the var was never set at all), confirming they exercise the real defect, not a tautology.

## Diagnostic script

[...]

PR #4398, opened 2026-08-12
fix(4395): a persistent litellm.token_counter failure retried every call

tya5:
**[docs-maintainer]** — part of #4395. **Shipping standalone first per lead-coder's explicit priority** — owner's primary observation confirmed this is the actual UI-hang mechanism; the persistent-cache-dir half + diagnostic script follow in a separate PR.

## Root cause — confirmed by owner's own primary observation

> input msg を送信するたびに固まる。その都度 reyn.log に SSLError のコールスタックが追加される

This matches `estimate_tokens()`'s exact pre-fix mechanism: its cache is keyed by `(model, text-hash)`, so a **persistent** failure (SSL egress blocked, no reachable proxy) meant every distinct input was a fresh cache miss paying the full failing `litellm.token_counter` call again — synchronously, from turn processing, blocking the UI each time. The docstring said so plainly: *"subsequent calls retry litellm.token_counter normally"* — correct for a transient blip, the worst possible shape for an environmental failure.

## Fix

A failure now starts a cooldown window (`_TOKEN_COUNTER_COOLDOWN_SECONDS = 60s`); any call inside it skips `litellm.token_counter` entirely — chars//4 immediately, no wait paid. **Deliberately not a permanent give-up** — the failure may clear at a longer timescale (a proxy comes back up), and `estimate_tokens()` has no restart hook to notice that on its own, so the first call after the cooldown expires re-probes automatically. A success (during or after the cooldown) clears it immediately.

## Tests

`tests/services/test_4395_token_counter_persistent_failure_cooldown.py` (Tier 2, 3 cases) uses a **real callable** standing in for `litellm.token_counter` — the same per-text-key call-count-dict technique `test_compaction_token_cache_incremental.py`'s own `_counting_token_counter` already established for this module (no `unittest.mock`). "Was litellm.token_counter actually invoked" is observed via that dict; the private cooldown deadline is only ever **written** (simulating "cooldown elapsed", the same input-simulation shape `test_2961`'s own fixture already uses), never **asserted on** directly (testing.ja.md's Tier-4 private-state ban).

`test_2961_token_counter_fallback_false_positive.py`'s isolation fixture updated to also reset the new module-global between tests (a real gap the new state var introduced).

**Falsified**: the key test (a second, distinct-text call during cooldown must not re-invoke `litellm.token_counter`) fails cleanly against the pre-fix code, with the exact stale docstring's warning message captured in the log.

## Not a rushed fix

[...]

PR #4406, opened 2026-08-12
feat(#4405): REYN_STALL_TRACE opt-in stall-trace diagnostic

tya5:
**[tui-coder]** — #4405 (from #4403's investigation): an opt-in, off-by-default diagnostic that dumps the process's Python stack to reyn's own log if a turn blocks longer than `REYN_STALL_TRACE=<seconds>`.

## Background

#4403's investigation into an owner-reported ~20s per-message freeze falsified four independent, individually-real hypotheses in a row (#4398's cooldown, tiktoken timeout absence, `build_history`'s full per-turn token-estimate scan, #4401's MCP probe timeout) — each a real, confirmed, measured defect, none of them explaining the observed duration alone or combined. #4399 (tiktoken's synchronous `requests.get` seed download) may separately resolve the specific 20s the owner is chasing right now — but the general problem this tool solves (an event loop that's frozen, with no way to see what's blocking it) is real and recurring regardless of that specific fix's outcome.

## Design (confirmed with lead-coder — an earlier "piggyback on Textual's watchdog" idea was considered and dropped)

`faulthandler.dump_traceback_later` on a dedicated OS thread, armed/disarmed around `Session._run_router_loop` — the single choke-point every turn kind (user / hook / agent_request / pipeline_result) funnels through before its first LLM call (already documented as such, #3475).

This is NOT the same axis as reyn's existing `LoopTripwire` (#3539, `loop_probe.py`) — traced during this investigation and confirmed (by grepping Textual's own source, not assumed) to be reyn's OWN mechanism, not a Textual built-in, despite producing the exact `"unresponsive {N}s"` text the owner's Textual popup showed. `LoopTripwire` detects lateness by comparing a scheduled vs. actual tick time — which means it can only fire AFTER the event loop resumes, i.e. after whatever was blocking has already finished. It can tell you **how long** you were frozen, never **where**. `dump_traceback_later` runs on its own OS thread outside the loop entirely, so it's the only mechanism that can capture a stack **during** the freeze.

Output is the stack only (`faulthandler`'s own traceback dump) — never argument values or message content, since the owner is on a company machine and can't paste log bodies out. Routed to reyn's own `.reyn/logs/reyn.log` via the `FileHandler` `chat.py`'s interactive logging setup already installs, falling back to stderr when none is present.

## Test plan

- [x] `python scripts/verify_module_docstrings.py src/reyn/runtime/stall_trace.py src/reyn/runtime/session.py` — OK
- [x] `python scripts/mypy_ratchet.py` — OK (211/215 baselined)
- [x] Scoped pytest — 80 passed across every test file touching `_run_router_loop` I could find (`test_turn_cancel_1468`, `test_session_lifecycle_events_1800`, `test_3475_probe_degradation_visibility_and_timeout_config`, `test_1800_c_staging`, `test_3036_spawned_session_mcp_refresh`, `test_session_cost_accumulation`, `test_chat_compaction_engine_11axis`, `test_force_close_chat_handoff_1092`, `test_4381_stage1_overflow_classification`, `test_3783_stage3_invert_default_to_recover`) — confirms wrapping the body in try/finally with the env var unset is a no-op
- [ ] **No test file for `stall_trace.py` itself — deliberate, not an oversight.** This is a diagnostic tool, not a system invariant: pinning the actual N-second-block-then-dump behavior would need a real multi-second sleep in a test, which the testing-policy time-limit ban forbids. It fits none of the three test Tiers (six-question check ①). Manually verified instead: `arm()` then `disarm()` round-trips through the real `faulthandler` API without raising, for both an env value present and absent; `stall_trace_seconds_from_env()` correctly returns `None` for unset/empty/non-numeric/non-positive values.
- [ ] Visual — n/a (no UI surface; log-file output only)

part of #4387 (unrelated arc — filed as #4403/#4405)

---

**[lead-coder]** — レビュアの blocking 項目

- [x] 🔴 `stall_trace.py` の docstring が「what IS tested is that arming/disarming touches the real faulthandler API」と書いていますが、★この PR にテストは 1 本もありません（`tests/` 0 件）。テストを書かない判断自体は承認済みで、問題は★書いていないものを「テストされている」と書いたこと。A: その一文を消す / ★B: 実際に 1 本書く（`arm`/`disarm` が `faulthandler` の実 API を叩くことを公開の観測点で。時間の禁止には当たりません）。私は B を推奨します — env var の配線が死んでも今は誰も気づけないため。

**[tui-coder]** — Went with **A** (removed the untrue claim from the docstring) per lead-coder's follow-up: owner is actively hitting repeated multi-second UI freezes, landing fast matters more than B right now. B is not dropped — filed as #4407, linked here.

- [x] 🔴 Docstring's false "arming/disarming is tested" claim removed. Real wiring test tracked in #4407 (follow-up, not blocking this PR).

⚪ **上の箱は★レビュア（lead-coder）が閉じました** — 著者の申告ではなく `feat/4405-stall-trace-diagnostic` の中身を直接引いた結果です（`what IS tested` が **0 hit**）。

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -8003,37 +8003,59 @@ class Session:
         enum either — but a chain that never runs still primes nothing).
         Running the chain here instead makes the ordering guarantee structural
         rather than kind-dependent.
+
+        #4405: this whole body is the "one turn" a ``REYN_STALL_TRACE=N``
+        stall-trace diagnostic brackets — armed on entry, disarmed in a
+        ``finally`` so an exception path still cancels it. Off by default
+        (env var unset); see ``reyn.runtime.stall_trace`` for why.
         """
-        self._last_turn_chain_id = chain_id
-        await self._router_host.maybe_refresh_mcp_tools_from_yaml()
-        self._router_host.maybe_reload_mcp_tools_cache_from_disk()
-        await self._router_host.ensure_mcp_tools_cached(
-            per_server_timeout=self._safety.timeout.mcp_probe_seconds,
+        from reyn.runtime.stall_trace import (
+            arm as _arm_stall_trace,
         )
-        with active_turn(chain_id):
-            await self._loop_driver.run_turn(user_text, chain_id)
-        # #1800 slice 5a: emit HERE (after `run_turn()` returns), not inside RouterLoop — the
-        # only placement that fires exactly once per turn regardless of which terminal path the
-        # loop took; moving it into RouterLoop risks double-emission or a missed terminal path.
-        self._audit_events.emit("turn_completed", chain_id=chain_id)
-        # #1800 slice 5b: turn_end lifecycle hooks — E self-continuation / C stage / F shell:
-        # docs/concepts/runtime/hooks.md#e-self-continuation-a-push-with-wake-true
-        await self._hook_dispatcher.dispatch(
-            "turn_end",
-            build_hook_payload(
-                "turn_end", agent_name=self.agent_name,
-                chain_id=chain_id, user_text=user_text,
-            ),
+        from reyn.runtime.stall_trace import (
+            disarm as _disarm_stall_trace,
         )
-        # #2073 S1: config hot-reload turn-boundary safe-point (timing-B):
-        # docs/concepts/runtime/config-hot-reload.md#turn-boundary-safe-point-timing-b
-        await self._hot_reloader.apply_pending()
-        # #3787: project-context edit detection — read-only, emits at most once per
-        # edit; does NOT reload ``self._project_context`` (see ProjectContextWatcher).
-        self._project_context_watcher.check()
-        # ADR-0038 Stage 1a: turn boundary = a user-facing checkpoint. #1547: the
-        # user message is this checkpoint's anchor for the rewind-timeline preview.
-        # #1533 2c: the FULL message is persisted alongside (edit-prefill source).
-        await self._journal.cut_generation(
-            anchor=_truncate_anchor(user_text), full_message=user_text,
+        from reyn.runtime.stall_trace import (
+            stall_trace_seconds_from_env,
         )
+
+        _stall_seconds = stall_trace_seconds_from_env()
+        if _stall_seconds is not None:
+            _arm_stall_trace(_stall_seconds)
+        try:
+            self._last_turn_chain_id = chain_id
+            await self._router_host.maybe_refresh_mcp_tools_from_yaml()
+            self._router_host.maybe_reload_mcp_tools_cache_from_disk()
+            await self._router_host.ensure_mcp_tools_cached(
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — 🟢 **設計は正しい。★blocking 1 件 —— docstring が「テストされている」と書いていますが、★テストは 1 本もありません。**

## 🔴 blocking

`stall_trace.py` の module docstring:
> **what IS tested is that arming/disarming touches the real faulthandler API, a real observable point**

```
gh pr view 4406 --json files → ★tests/ 配下 0 件
gh pr diff  | grep "^+++ b/tests/" → ★0
```
🔴 **「テストを書かない」という判断★自体は私が承認しました。★問題は、書いていないものを
「テストされている」と書いたこと**です。**この repo が一晩に何度も踏んでいる形
（宣言 ≠ 実在）を、★その形を説明している docstring の中でやっています。**

⚪ **直し方はどちらでも構いません**:
- **A: その一文を消す**（「テストは書かない、理由は〜」だけにする）—— **最小。**
- **B: 実際に書く**（`arm()` / `disarm()` が `faulthandler` の実 API を叩くことを、
  `faulthandler.is_enabled()` 等の★公開の観測点で見る）—— ★**時間の禁止には当たりません**
  （待たない、寝ない、閾値を跨がない）。**私は B を推奨します** —— **「env var を読んで arm する」の
  配線が死んでいても、今は誰も気づけません**（★#4405 が生まれた原因と同じクラス）。

## ✅ 設計で良かった点

🟢 **`dump_traceback_later` を選んだ理由が書いてある** —— **専用 OS スレッドの timer ∴
イベントループ自身が塞がれていても発火する。`asyncio` ベースの watchdog は
★捕まえたい条件下では永久にスケジュールされない。** ★**これは私の当初案（Textual に相乗り）を
否定した根拠**でもあり、**正しい否定でした。**
🟢 **スタックのみ・引数の値やメッセージ本文を出さない** —— **owner が会社の機械にいて
ログ本文を出せない**という制約から設計が導かれています。★**制約を設計に翻訳できている。**
🟢 **既定 off ∴ 誰の挙動も変えない** ／ **`finally` で disarm** ／ **出力先が既存の `reyn.log`**
（★operator が既に見る場所 —— 新しい面を増やしていない）。
🟢 **module docstring が「4 つの仮説が全部 falsify された」経緯を残している** ——
**次にこの症状を見る人が、★同じ 4 つを掘り直さずに済みます。**

## 六問

1. **Tier なし（道具）** —— ★**その判断は正しい**（reyn の不変条件ではない）。
   ⚠️ **ただし上の B を採るなら、その 1 本は Tier 2**（「env var が配線されている」は契約）。
2〜6 **該当なし（テストが無いため）。**

- [ ] 🔴 docstring の「what IS tested is …」を、実態に合わせる（A: 一文を消す / ★B: 実際に 1 本書く）

PR #4413, opened 2026-08-12
fix(#4395): stop re-attempting a failed litellm import on every call

tya5:
**[docs-maintainer]** — PR-1 of the #4395 arc: minimal, non-threading fix for the owner-reported UI freeze.

## What

A live owner repro (py-spy stack) caught the event-loop/UI thread blocked inside `import litellm`. `ensure_litellm_ready()`'s own docstring claimed a caller's subsequent bare `import litellm` was "cheap — Python caches the module" — true only on **success**. Python does **not** cache a **failed** import (a module whose top-level code raises is evicted from `sys.modules`), so every call site doing its own redundant bare `import litellm` right after calling the chokepoint was independently re-attempting — and re-failing — the exact same slow, unbounded, network-touching import. One attempt became two per call, every call, for as long as the environment stayed broken (e.g. a proxy timeout).

## The 6 sites (Test plan — full enumeration, not just a count)

**Set A** — called `ensure_litellm_ready()` then did their own separate `import litellm` right after:
- `src/reyn/llm/model_budget.py:56-59` (`_lookup_max_input`) — has a safe fallback (128K default), stays synchronous.
- `src/reyn/llm/model_cost_rate.py:38-41` (`get_input_cost_per_1m_usd`) — has a safe fallback (`None` = "no warning"), stays synchronous.
- `src/reyn/llm/llm.py:2016` (`recorded_acompletion`) — **no fallback**, a real completion genuinely needs litellm.
- `src/reyn/data/embedding/litellm_provider.py:405-415` (embed retry loop) — no fallback within one attempt (existing `except Exception` around the whole attempt already retries).

**Set B** — bypassed the chokepoint entirely, never called `ensure_litellm_ready()` at all:
- `src/reyn/llm/pricing.py`: `estimate_cost`, `estimate_cost_breakdown`, and **`estimate_embedding_cost`** — this third one is a 4th site beyond the 3 originally measured; found independently while fixing the other two. Flagging the discrepancy transparently rather than silently reconciling it.

## Fix

- `ensure_litellm_ready()` now **returns the imported module** (or `None` on failure) instead of always returning `None`. Callers check the return value *before* their own `import litellm`, so that bare import only ever runs after a confirmed success — always a cheap `sys.modules` cache hit, never a repeat of a failing attempt.
- A failure is **no longer cached as "ready"**. This also fixes an independently-found, real, pre-existing bug: the original code set `_litellm_ready = True` in a bare `finally:` block regardless of whether the internal import actually succeeded — meaning the #2929 log-routing / `aiohttp_trust_env` setup silently never re-applied even after eventual recovery. Now only a genuine success sets it.
- A persistent failure is **warned exactly once** (`_litellm_import_failure_warned`), not once per call — landing condition: the degradation must be visible, but not flood logs (#3368's "warn every time punishes the disciplined caller" trap).
- `llm.py`'s `recorded_acompletion` and `litellm_provider.py`'s embed-retry loop — both `async def`, both no-fallback sites — now call `ensure_litellm_ready()` via `asyncio.to_thread(...)` rather than in-loop. **This is a lead-coder correction mid-PR** (owner: "does waiting on the UI loop differ from waiting on a dedicated thread?" — yes: waiting in-loop stops the whole event loop, animation and input included, not just the one coroutine). PR-1's *scope* is unchanged (still these same 6 sites + the one-time warn), but writing these two call sites as an in-loop blocking wait would have forced a follow-up PR to rewrite every caller; `asyncio.to_thread` keeps them in an already-awaitable form so a dedicated background-thread/cooldown mechanism can slot in behind the same await point later.

## Explicitly out of scope for this PR

- The dedicated background-thread / cooldown warming mechanism itself (so a *first* call after a failure doesn't block at all) is a **separate, deferred follow-up PR** — not included here. This PR only eliminates the redundant double-attempt and makes the two no-fallback call sites genuinely awaitable.
- `llm.py` / `litellm_provider.py` intentionally remain **blocking-until-resolved** on the underlying import outcome (just off the event loop now) — there is no fallback for a real completion or embedding call, so this is not a gap, it matches the architect's own landing condition that these two sites must keep retrying until they get a result.

## Test plan

- [x] 4 new tests in `tests/llm/test_4395_litellm_import_not_recached_on_failure.py` (Tier 2), using `builtins.__import__` patching (real import-machinery interception, not a mock of any reyn object) to simulate a genuine `import litellm` failure:
  - a failed import returns `None`, never raises
  - a failed import is **not cached** — the next call genuinely re-attempts
  - a persistent failure is warned exactly once across 5 calls
  - a success *after* a prior failure is returned and stays cached (no re-attempt on the next call)
- [x] Falsified: stashed all 6 modified files together, confirmed 3/4 new tests failed against the reverted code with genuine assertion errors, restored, confirmed all 4 pass again.
- [x] `ruff check` on all 7 changed files — clean.
- [x] `python scripts/mypy_ratchet.py` — 211/215 baselined, no new findings.
- [x] `python scripts/test_tier_audit.py --strict tests/llm/test_4395_litellm_import_not_recached_on_failure.py` — 0 Tier-4 violations (one `len(...) == 1` format-pin was caught and fixed to a tuple-unpack idiom before this PR was opened).
- [x] `python scripts/verify_module_docstrings.py` on all 6 changed src files — clean.
- [x] `python scripts/check_tests_path_literal_reference.py` — clean (121/121 baselined).
- [x] `pytest tests/llm/ tests/data/test_embedding_provider.py tests/core/test_embed_bound_cancel_3043.py tests/core/test_embed_attempts_observation_3047.py` — 702 passed, no regressions.

## Arc note

`part of #4395` (not closing — this is PR-1 of a multi-PR arc; the deferred background-thread mechanism is a separate follow-up PR not yet opened).

---

**[lead-coder]** — レビュアの blocking 項目

- [x] 🔴 `src/reyn/llm/pricing.py:236` の `from litellm.types.utils import Usage` を継ぎ目経由に。★サブモジュール import でも親パッケージの `__init__.py` が実行され、`default_encoding`（tiktoken の同期取得）が走ることを実測しました。`import litellm` という形をしていないため grep の網から外れていた 1 箇所です。

⚪ **上の箱は★レビュア（lead-coder）が閉じました** — `pull/4413/head` を直接引いた結果です。★対応の形が要求以上でした: 「継ぎ目経由にする」ではなく **`ensure_litellm_ready()` の戻り値（確認済みモジュール）を引数で受け取り `Usage` を属性アクセスで読む** —— ∴ **この関数は★litellm に独立に触れること自体が構造的に不可能**になっており、「呼び出し元がゲートしてくれること」に依存していません。

### DIFF
--- a/src/reyn/data/embedding/litellm_provider.py
+++ b/src/reyn/data/embedding/litellm_provider.py
@@ -389,8 +389,37 @@ class LiteLLMEmbeddingProvider:
                 # be the FIRST litellm import in the process — funnel it through
                 # ensure_litellm_ready() so litellm's import-time console log
                 # routing (#2929) wraps it too (idempotent; cheap on 2nd+ call).
-                from reyn.llm.litellm_bootstrap import ensure_litellm_ready
-                ensure_litellm_ready()
+                # #4395 PR-1: check success BEFORE `import litellm` — that used
+                # to be an unconditional bare import right after the
+                # chokepoint call, independently re-attempting (and
+                # re-failing) litellm's own slow, unbounded import on EVERY
+                # retry-loop iteration while it kept failing (a failed
+                # import isn't cached by Python; only this chokepoint's own
+                # attempt was) — doubling the cost of every already-retried
+                # attempt in this loop. `import litellm` below now only ever
+                # runs once success is confirmed, so it's always a cheap
+                # sys.modules cache hit. The existing `except Exception`
+                # below already handles this failure exactly like any other
+                # attempt failure — same retry/backoff behavior, half the
+                # redundant work per iteration.
+                from reyn.llm.litellm_bootstrap import (
+                    LitellmUnavailableError,
+                    ensure_litellm_ready,
+                )
+                # #4395: `asyncio.to_thread`, not an in-loop call — this is
+                # `async def`, and a synchronous `ensure_litellm_ready()`
+                # here blocks the WHOLE event loop for as long as the
+                # underlying import takes, not just this coroutine (lead-
+                # coder correction, #4395: PR-1 must land this call site
+                # already-awaitable so PR-2's dedicated-thread mechanism
+                # slots in behind the same await point without a rewrite
+                # of this call site).
+                if await asyncio.to_thread(ensure_litellm_ready) is None:
+                    raise LitellmUnavailableError(
+                        "import litellm failed — see the reyn.llm."
+                        "litellm_bootstrap warn-once log line for the "
+                        "underlying cause",
+                    )
                 import litellm
                 response = await self._aembedding_bounded(
                     litellm,
--- a/src/reyn/llm/litellm_bootstrap.py
+++ b/src/reyn/llm/litellm_bootstrap.py
@@ -11,10 +11,19 @@ stderr, so an interactive CUI session's terminal is never corrupted by a
 litellm banner or warning.
 
 ``ensure_litellm_ready()`` is the ONE place that should perform this
-first-import work. Callers that need the ``litellm`` module still do their own
-``import litellm`` afterward (cheap — Python caches the module), but calling
-``ensure_litellm_ready()`` first guarantees the log-routing patch is active
-for whichever call site happens to import litellm first in a given process.
+first-import work — and (#4395 PR-1) the ONE place the actual ``import
+litellm`` statement should live. Callers use the module it RETURNS
+(``None`` on failure) rather than doing their own separate ``import
+litellm`` afterward. An earlier version of this paragraph claimed a
+caller's own subsequent bare ``import litellm`` was "cheap — Python
+caches the module" — true only when the import SUCCEEDED. Python does
+NOT cache a FAILED import (a module whose top-level code raises is
+evicted from ``sys.modules``), so a caller doing its own redundant bare
+``import litellm`` right after this function would re-attempt — and
+re-fail — the exact same slow, unbounded network-touching import this
+function had JUST attempted, silently doubling the cost of every
+failure. See ``ensure_litellm_ready()``'s own docstring for the full
+success/failure contract this fixed.
 
 #3671: this module also owns the #3434 client-cache "pre-existing keys"
 baseline (:func:`reset_client_cache_baseline`/:func:`client_cache_baseline`)
@@ -34,7 +43,7 @@ from __future__ import annotations
 import contextlib
 import logging
 from concurrent.futures import Future
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -89,6 +98,18 @@ _ready_registry: "dict[str, Future]" = {}
 # reading, invalidating its baseline mid-call.
 _baseline_generation = 0
 _baseline_registry: "dict[int, Future]" = {}
+
+
+class LitellmUnavailableError(Exception):
+    """#4395 PR-1: raised by a call site with NO safe fallback (a real
+    completion or embedding call — `llm.py`'s `recorded_acompletion`,
+    `litellm_provider.py`'s `embed_batch`) when `ensure_litellm_ready()`
+    returns `None`. A clear, explicit failure instead of letting a
+    redundant, un-gated bare `import litellm` raise whatever exception
+    litellm's own failed import happened to produce — same underlying
+    cause, a legible error instead of an incidental one. Retriable: the
+    NEXT call gets a fresh attempt (see `ensure_litellm_ready()`'s own
+    docstring — a failure is not permanently cached)."""
 
 
 @contextlib.contextmanager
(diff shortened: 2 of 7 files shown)


tya5:
**[docs-maintainer]** — Amended per your feedback (source-attribution pin, litellm_bootstrap docstring example): the docstring now names the underlying mistake explicitly — `ensure_litellm_ready()` merely having run once (setup side-effects applied) and `import litellm` actually succeeding are two different guarantees, and a caller with no fallback may only assume the second from this function's own return value, never from a side effect of having reached the chokepoint. Also corrected the site-count in the docstring itself (was undercounting: 4 Set-A sites + 3 separate Set-B pricing.py sites, not '3 call sites'). All gates re-verified clean after the amendment; commit amended and force-pushed.

tya5:
**[lead-coder]** — 🔴 **blocking 1 件。★`pricing.py:236` が継ぎ目を通っておらず、同じ経路に入ります。実測しました。**

## 🔴 blocking

```python
src/reyn/llm/pricing.py:236
    from litellm.types.utils import Usage      # ← ★継ぎ目を通っていない
```
**サブモジュール import でも★親パッケージの `__init__.py` が実行されます。実測:**
```
$ python -c "from litellm.types.utils import Usage; ..."
  litellm(親) が sys.modules に入ったか        : ★True
  default_encoding が実行されたか              : ★True    ← tiktoken の同期取得が走る
```
🔴 **∴ この 1 行だけで、★#4395 の経路にそのまま入ります。**
⚠️ **`import litellm` という★形をしていないので、grep の網から外れます** ——
**「6 箇所」を数えた時、私も貴方もこの形を数えていませんでした**
（★本日の「検索パターンが人口調査を決めた」の実例がもう 1 つ）。

⚪ **`_to_litellm_usage` は★型を作るだけ**なので、**`ensure_litellm_ready()` の戻り値から
取れるなら★それが素直**です（`litellm.types.utils.Usage`）。**難しければ、この関数の
呼び出し元がどこまで諦められるかで判断してください。**

## ✅ 良かった点

🟢 **★私の指摘より 1 箇所多く直しています** —— **`model_cost_rate.py` は私の「6 箇所」に
入っていませんでした。** ⚪ **貴方が自分で見つけた 7 箇所目。**
🟢 **PR-2 の位置づけ訂正を★実装に反映済** —— `llm.py:2016` と embed-retry を
`asyncio.to_thread` 経由に。**「`async def` で fallback が無い ∴ in-loop の同期呼び出しは
コルーチンだけでなくループ全体を止める」**という理由まで書かれている。
🟢 **`Any | None` への型変更**を「★mypy_ratchet は緑だが IDE が拾う類」として
**明示的に報告した**のも良い（★緑を根拠に黙らなかった）。
🟢 **falsify が 6 ファイル stash/restore の全体で**（3/4 が RED → restore で GREEN）。

## 六問（新テスト）

**Tier 2**（reyn の起動/ターンが第三者の同期ネットワークに引っ張られない）／転記なし／
**★owner が現に踏んでいる**／skip なし／累積なし／宣言と一致。

- [ ] 🔴 `src/reyn/llm/pricing.py:236` の `from litellm.types.utils import Usage` を継ぎ目経由に（実測: サブモジュール import でも `default_encoding` が走る）

tya5:
**[lead-coder]** — 🔴 **母集団の訂正: 私が渡した「6 箇所」は★2 倍以上外れていました。実際の import 地点は 12 箇所です。**

**`pricing.py:236` を見つけた後、★形を問わず引き直しました**（`from litellm...` も含めて）:

```
from litellm... 形（8）
  core/op_runtime/web.py:205        from litellm.llms.custom_httpx.http_handler import get_ssl_verify
  core/registry/client.py:173       同上
  dev/testing/replay.py:555         from litellm.types.utils import (...)
  interfaces/cli/__init__.py:82     from litellm.exceptions import AuthenticationError
  llm/llm.py:485                    from litellm.litellm_core_utils.logging_worker import ...
  llm/llm.py:573                    from litellm.llms.custom_httpx.async_client_cleanup import ...
  llm/llm.py:1869                   from litellm.utils import supports_native_streaming
  ★llm/pricing.py:236                from litellm.types.utils import Usage   ← blocking 済

import litellm 形（4）
  llm/llm.py:910 ／ :1521 ／ :1739 ／ :1868
```

## ⚠️ ただし「12 箇所すべてを PR-1 に入れろ」ではありません

🔴 **PR-1 の範囲は★「毎ターン繰り返す経路」で定義されています。**
**上の 12 のうち、★どれが毎ターン走るかは未分類**です（例: `interfaces/cli/__init__.py:82` は
起動時 1 回かもしれない／`llm.py:573` は cleanup 経路かもしれない）。

⚪ **∴ PR-1 は★現状の範囲 ＋ `pricing.py:236` で着地させます。**
**owner が丸一日止まっており、★母集団の完全性のために着地を遅らせません。**

## 🔵 follow-up として起票します（★私が起票します、実装者は不要）

**「12 箇所それぞれの★到達可能性を分類し、毎ターン/毎フレーム走るものを継ぎ目経由にする」。**
⚠️ **分類の軸は★「UI ループ上で繰り返し走るか」**であって、「継ぎ目を通っているか」ではありません
（★通っていなくても、起動時 1 回なら PR-2 の対象）。

## 🔴 記録: 同じ日に 3 度目です

```
①#4358      「12 個の名前が★値として」だけ数え、★キーの位置とキーの背後の値を落とした
②#4395 PR-1 「6 箇所」を `import litellm` の★形だけで数え、`from litellm...` を落とした
③★今回      その訂正の後、★形を問わず引いたら 12 箇所出た
```
🔴 **★「検索パターンが人口調査を決める」を、私は本日 3 回踏んでいます。**
⚪ **①②はどちらも★他人（owner / architect）に指摘されて気づき、③だけが自分で引き直したもの**です。

tya5:
**[docs-maintainer]** — Fixed the blocking finding: `_usage_object_for` in `src/reyn/llm/pricing.py` now takes the already-confirmed `litellm` module as a parameter (from `ensure_litellm_ready()`'s return value at its only call site, `estimate_cost`) and reads `Usage` via `litellm_module.types.utils.Usage` attribute access, instead of its own separate `from litellm.types.utils import Usage` statement. That statement was a 7th chokepoint-bypass site missed by both the original census and my own re-count — it uses a `from litellm.X import Y` form, not `import litellm` / `from litellm import`, which is exactly the shape neither grep pass was looking for.

Verified empirically that `litellm.types.utils` is already a populated attribute on the module object after a plain successful `import litellm` — no separate submodule import is structurally needed anymore, so this function can no longer independently touch litellm's import machinery at all (not just "currently gated by its one caller" — structurally incapable of bypassing the chokepoint).

Falsified: pre-warmed litellm in a fresh process, spied on `builtins.__import__` around the call — 0 litellm-touching imports triggered (only pydantic-internal ones from `Usage`'s own field validation, unrelated to this fix).

Re-verified full gate suite after the fix: ruff clean, mypy_ratchet 211/215 baselined (no new findings), `verify_module_docstrings.py` clean, 41 pricing/cost-related tests pass. Amended the commit and force-pushed.

The other 11 `from litellm.X` / bare `import litellm` sites found during this investigation are already tracked and correctly classified in #4415 (repeats-per-turn vs once-per-process vs unreachable) — not duplicated here, and intentionally not folded into this PR's scope.

PR #4417, opened 2026-08-12
fix(#4395): cooldown + dedicated background thread for litellm warm-up

tya5:
**[docs-maintainer]** — PR-2 of the #4395 arc: cooldown + a dedicated background thread, in ONE PR per owner decision (split framing kept losing — see history below), 2 commits for revert granularity.

## Why one PR, two commits

Lead-coder's own account, quoted because it explains the shape of this PR directly: cooldown alone only rate-limits retrying — it can't save the *first* attempt in a process, and the owner's own py-spy at both the startup freeze and the per-message freeze showed the **identical** ssl/tiktoken stack. Cooldown and the owned thread close two different axes of the same defect (frequency vs. who-waits) and neither alone closes the owner's report. Landing them separately would have re-created the exact gap that made PR-1-alone insufficient. Commits are split so a revert of the thread mechanism (the larger, more behavior-changing half) doesn't have to take the cooldown fix down with it.

## Commit ① — cooldown (`fix(#4395): cooldown a persistently-failing litellm import (axis②)`)

PR-1 (#4413) removed the WITHIN-one-call double attempt but deliberately left a failure un-cached — by design, the next call gets a genuinely fresh attempt. A live owner repro **after PR-1 had already landed** caught the process still hanging inside a TLS handshake at `litellm_bootstrap.py`'s own `import litellm` line — every subsequent call was re-attempting, and re-hanging on, the same broken environment. "Not permanently cached" and "retried on literally every call" are different things; PR-1 only fixed the first.

Fix: a failure now arms a cooldown (`reyn._cooldown`, the same primitive #4398 established for the identical shape in `estimate_tokens`) — a call during the window returns `None` immediately, no import attempted. A success clears it. Not a permanent give-up: the underlying cause may clear, and the first call after the window elapses always re-probes.

**Does not help the first attempt in a process** — that one still pays the real cost once. That's commit ②'s job.

## Commit ② — dedicated background thread (`fix(#4395): dedicated background thread warms litellm for fallback-having call sites`)

Adds to `litellm_bootstrap.py`: `is_litellm_ready()` (non-blocking read), `warm_litellm_in_background()` (idempotently starts exactly ONE daemon thread that retries the unmodified, cooldown-gated `ensure_litellm_ready()` until it succeeds, then exits), `ensure_litellm_ready_or_defer()` (the non-blocking entry point — returns immediately if already ready, else kicks off the thread and raises `LitellmWarmingInBackgroundError` for the caller's own existing fallback to catch).

**Converted call sites** (all three already had a safe, cheap fallback):
- `model_budget.py`'s `_lookup_max_input` (fallback: 128K conservative default)
- `model_cost_rate.py`'s `get_input_cost_per_1m_usd` (fallback: `None` = unknown cost)
- `compaction/engine.py`'s `estimate_tokens` + `is_context_overflow_error` (fallback: chars//4 / keyword search) — also found and fixed a residual instance of PR-1's own defect here: this module wasn't part of PR-1's diff, and both functions were still doing `ensure_litellm_ready(); import litellm` unconditionally, ignoring the return value.

**`llm.py` / `litellm_provider.py` unchanged** — no fallback exists for a real completion/embedding call, so they keep the blocking `ensure_litellm_ready()`, already off the event loop via PR-1's own `asyncio.to_thread` conversion.

**Startup path, verified structurally (not assumed)**: `_ensure_turn_budget_engine` is lazy (#3671 follow-up — confirmed NOT called at Session construction). `maybe_emit_model_cost_warn(..., action="session_start")` (`session.py`'s `run()`) IS called synchronously on the startup path, cost-warn `enabled=True` by default, and resolves through `model_cost_rate.py` — converted above. So session startup's own first litellm touch now kicks off the background thread immediately instead of blocking main — closing the owner's startup freeze report via the SAME conversion that closes the per-message one, not a separate fix.

## Hardening, not required (`llm.py`'s `_get_retryable_litellm_exceptions`)

Flagged mid-review as a possible second chokepoint-bypass, then **retracted by both architect and lead-coder** after architect traced the actual call graph: this function's only reachable path (`_llm_call_with_retry`'s exception handler) is only ever reached after `recorded_acompletion`'s own readiness check already succeeded or raised — never actually racing the background thread. Kept the conversion anyway (architect's re-affirmation after the retraction): it removes a *capability* (independently importing litellm) rather than adding a moving part, same shape as #4413's `pricing.py` fix — "structurally incapable of bypassing," not reliant on today's call order staying the same forever. Two existing tests updated to state their own real precondition (litellm already ready) explicitly instead of relying on incidental test-order luck across the suite — valuable independent of the retracted threat model.

## Test plan

- [x] New: `tests/llm/test_4395_pr2_litellm_warm_thread.py` (4 tests) — no-block-during-persistent-failure (20 calls <1s), exactly-one-worker-thread-under-load (falsifies an earlier respawn-per-caller draft), success-picked-up-by-next-call, already-ready-never-spawns-a-thread.
- [x] New: `tests/llm/test_4395_axis2_litellm_import_cooldown.py` (3 tests) — during-cooldown-no-reattempt, after-cooldown-reattempts, cooldown-check-itself-never-imports-litellm.
- [x] Updated `tests/llm/test_4395_litellm_import_not_recached_on_failure.py` (PR-1's own file): 2 of its 4 tests' premise went stale under the new cooldown (back-to-back calls no longer both retry) — fixed to simulate cooldown-elapsed via direct deadline manipulation, not a sleep.
- [x] Updated `tests/llm/test_thread_safety_startup_shared_state_3671.py`, `tests/services/test_compaction_token_cache_incremental.py`, `tests/llm/test_litellm_lazy_load.py`: the non-blocking conversion means the FIRST call to a converted site in a process now defers instead of blocking inline — these files' own real preconditions (litellm already ready, or wait for the background thread) made explicit rather than left to incidental test-order luck. Genuine behavior-change-driven fixes, not workarounds.
- [x] Falsified: `test_exactly_one_worker_thread_regardless_of_call_volume` is the test that would have failed against an earlier, corrected respawn-per-caller design.
- [x] `ruff check` — clean on every changed file.
- [x] `python scripts/mypy_ratchet.py` — 211/215 baselined, no new findings (pyright-only noise from litellm's own missing stubs is pre-existing, unrelated to this diff).
- [x] `python scripts/test_tier_audit.py --strict` on every changed/new test file — 0 Tier-4 violations (1 `len(warm_threads) == 1` caught and fixed to a tuple-unpack idiom before this PR was opened).
- [x] `python scripts/verify_module_docstrings.py` — clean.
- [x] `python scripts/check_tests_path_literal_reference.py` — clean.
- [x] Broad sweep: `tests/llm/`, embedding-adjacent, compaction/model-budget/model-cost-warn/context-budget/token-budget-adjacent test files — 1000+ tests, all passing after the precondition fixes above.

## Arc note

`part of #4395` (not closing — #4415, the full 12-site reachability classification, is explicitly scoped as a separate, non-urgent follow-up after this PR per lead-coder).

---

**[lead-coder]** — レビュアの blocking 項目（★時間依存の 2 箇所）

- [x] 🔴 `assert elapsed < 1.0` を削除し、★「呼び出しスレッドが import を試みていない」を assert に（fake が `threading.current_thread().name` を記録し `"MainThread" not in calls`）。速さは結果であって主張ではありません。
- [x] 🔴 `deadline = time.monotonic() + 10.0` の待ち予算を外す（`while not is_litellm_ready(): sleep(0.05)` ── CI の 120s kill が backstop）。★worker thread を公開の観測点にできるなら join の方が良い。

⚪ **上の 2 箱は★レビュア（lead-coder）が閉じました** — `pull/4417/head` を直接引いた結果です（①`elapsed` の assert が消え `threading.current_thread().name` の記録に置換、docstring に「not an inferred EFFECT」まで明記／②`deadline` / `monotonic() +` が **0 hit**）。

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -237,6 +237,14 @@ dependencies = [
                               # bounded work (~4-5 functions + a context re-plumb, see #4302)
                               # tracked as its own follow-up issue, deliberately NOT folded
                               # into this dependency-hygiene PR.
+    # #4395: was transitive only (Required-by: httpx, mcp, openai,
+    # sse-starlette, starlette, watchfiles — 4.14.0 already installed via
+    # that chain). Owner decision: declare it directly rather than keep
+    # relying on other packages' own pins to keep pulling it in — the same
+    # "make an existing transitive a direct dependency" move as `mcp`'s own
+    # entry above. Floor set to the version already resolved and in use
+    # (4.14.x); not raised for a specific new API this PR relies on.
+    "anyio>=4.14",
     "pyperclip>=1.9",        # #3616 ①: cross-platform clipboard, replacing a
                               # hand-rolled pbcopy/wl-copy/xclip/xsel/clip
                               # dispatch table whose Windows arm piped UTF-8
--- a/src/reyn/_cooldown.py
+++ b/src/reyn/_cooldown.py
@@ -1,0 +1,60 @@
+"""A persistent, environmental failure needs a COOLDOWN, not a permanent
+give-up and not a retry on literally every call (#4395/#4398 — the same
+class, closed here in ONE place instead of two independent
+reimplementations).
+
+Two distinct, real instances of this exact shape landed the same night:
+
+- ``services/compaction/engine.py``'s ``estimate_tokens()`` (#4398): a
+  ``litellm.token_counter`` CALL that keeps failing for an environmental
+  reason (SSL egress blocked) was retried on every distinct text string
+  (the result cache is keyed by text, so a new string is always a miss) —
+  synchronously, from turn processing, blocking the UI each time.
+- ``llm/litellm_bootstrap.py``'s ``ensure_litellm_ready()`` (#4395): a
+  failed ``import litellm`` is REMOVED from ``sys.modules`` by Python
+  itself (any module whose top-level code raises never gets cached), so
+  the NEXT ``import litellm`` anywhere in the process starts completely
+  from scratch and hits the exact same failure again — an IMPORT failure,
+  not a call failure, but the same "persistent environmental failure,
+  retried every single call" shape. PR-1 (#4413) closed the WITHIN-one-call
+  double-attempt; this module backs PR-2's fix for the ACROSS-calls
+  repeat (#4395 axis②) — a live owner repro caught the process still
+  re-attempting, and re-hanging on, the same broken TLS handshake on
+  every subsequent call after PR-1 alone landed.
+
+Both need: a failure starts a cooldown window; any attempt inside it skips
+straight to the fallback (or, for ``ensure_litellm_ready()``, returns
+``None`` immediately — never a permanent give-up), no wait paid; a success
+(during or after the cooldown) clears it; the window is temporary — the
+underlying cause may clear (a proxy comes back up) and there is no restart
+hook to notice that on its own, so the next attempt after the window
+elapses always re-probes.
+
+This module holds only the two PURE, STATE-FREE comparisons that shape
+needs — ``time.monotonic()`` reads and arithmetic. Each call site still
+owns its OWN cooldown-deadline variable (a plain module float, matching
+the existing "no lock, unlocked check-then-set, worst case is a redundant
+early/late re-probe, never a wrong count" reasoning both #4395 and #4398
+already use) — sharing the STATE across unrelated failure classes (an
+import failure and a token-counter-call failure) would be wrong; sharing
+this tiny piece of ARITHMETIC is not.
+"""
+from __future__ import annotations
+
+import time
+
+
+def in_cooldown(cooldown_until: float) -> bool:
+    """True while *cooldown_until* (an absolute ``time.monotonic()``
+    deadline, or ``0.0``/negative for "no active cooldown") is still in the
+    future. ``time.monotonic()`` — never wall-clock — so an NTP step or a
+    sleep/resume cannot shorten OR lengthen the wait (#4398's own reasoning,
+    reused verbatim here)."""
+    return time.monotonic() < cooldown_until
+
+
+def new_cooldown_deadline(seconds: float) -> float:
+    """The absolute ``time.monotonic()`` deadline for a cooldown window of
+    *seconds* starting now — call this on a fresh failure to arm/extend a
+    call site's own cooldown-deadline variable."""
+    return time.monotonic() + seconds
--- a/src/reyn/llm/litellm_bootstrap.py
+++ b/src/reyn/llm/litellm_bootstrap.py
@@ -45,6 +45,8 @@ import logging
 from concurrent.futures import Future
 from typing import TYPE_CHECKING, Any
 
+from reyn import _cooldown
+
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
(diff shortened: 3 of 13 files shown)


tya5:
**[lead-coder]** — 🟢 **着地条件 3 点は★全部入っています。blocking は★時間に依存する 2 箇所だけ。**

## ✅ 着地条件

```
①cooldown       `reyn._cooldown` ＋ `_litellm_import_cooldown_until`
②★所有スレッド   `threading.Thread(name="reyn-litellm-warm", daemon=True)` ★1 本
                 ＋ `is_alive()` で 2 本目を作らない
③★公開 IF       `ensure_litellm_ready() -> "Any | None"` ／ `is_litellm_ready() -> bool`
                 ／ `ensure_litellm_ready_or_defer() -> "Any"`
                 ★litellm 由来の型が signature に 1 つも出ていない（owner の条件）
```
🟢 **`test_exactly_one_worker_thread_regardless_of_call_volume` の docstring が
「★respawn-per-caller 設計なら落ちたテスト」と書いている**のが良い ——
**★私が最初に出した「呼び出しごとにワーカーへ」を、テストが名指しで否定しています。**

## 🔴 blocking ①: `assert elapsed < 1.0`（`test_repeated_calls_during_a_persistent_failure_never_block`）

**CLAUDE.md**: *Tests carry no time limit of their own — no wait-budget constant in the body.*
🔴 **★1.0 秒は環境依存の定数**で、**遅い CI で偽陽性、速い環境で見逃し**ます。

✅ **★観測点を「効果」から「原因」へ動かせます:**
```python
calls: list[str] = []
def _always_failing_ensure_ready():
    calls.append(threading.current_thread().name)   # ★誰が呼んだか
    return None
...
assert "MainThread" not in calls    # ★呼び出しスレッドは import を試みていない
```
⚪ **「速い」ではなく「★呼び出し側が import 経路に入っていない」を見る** ——
**これが主張そのもの**です（速さは★その結果に過ぎない）。

## 🔴 blocking ②: `deadline = time.monotonic() + 10.0` の polling（`test_litellm_lazy_load.py`）

**CLAUDE.md**: *Wait on the condition unboundedly; a test needing more than CI's `--timeout=120`
kill switch should be decomposed.*
🔴 **10 秒は★待ち予算**です。

✅ **形は 2 つ**:
```
A ★予算を外す: while not is_litellm_ready(): time.sleep(0.05)
   → CI の 120s kill が backstop（★testing.md の指示どおり）
B ★スレッドを join する（`_litellm_warm_thread` を公開の観測点として出せるなら）
   → ★ポーリングそのものが消える。私は B を推奨
```

## 六問

1. **Tier 2**（起動/ターンが第三者の同期ネットワークに引っ張られない）。2. **転記ではない。**
3. **★owner が現に踏んでいる。** 4. **skip なし。** 5. ⚠️ **上の①②が該当**（時間で bound している）。
6. **一致。**

⚪ **`_always_failing_ensure_ready` が★実関数の契約（raise しない／None を返す／`_litellm_ready` を
反転しない）をコメントで明示**しているのが良い —— **★fake が本物の契約から外れていないことの witness。**

- [ ] 🔴 `assert elapsed < 1.0` を「★呼び出しスレッドが import を試みていない」の assert に置換
- [ ] 🔴 `deadline = ... + 10.0` の待ち予算を外す（★join できるなら join に）

tya5:
**[docs-maintainer]** — Two follow-ups pushed to this PR:

1. `chore(#4395): declare anyio as a direct dependency` — per owner decision (was transitive only, 4.14.0 already resolved via httpx/mcp/openai/sse-starlette/starlette/watchfiles). Evaluated whether this PR's own `_ready_registry`/background-thread mechanism duplicates `anyio.from_thread.start_blocking_portal()` — it doesn't: BlockingPortal bridges sync code calling INTO an async loop on a background thread; this PR's need is the inverse (a persistent worker retrying a sync operation with backoff, callers polling non-blockingly) — a different problem, not one BlockingPortal has a primitive for. Kept the existing Future-based ownership pattern (#3671 P1's, reused not reinvented).

2. **Generalizing the test-fix pattern (architect's observation, worth recording explicitly)**: the non-blocking conversion made the FIRST call to several converted sites genuinely defer instead of blocking inline — this broke 4 pre-existing tests across 2 files (beyond the 1 I'd already fixed in the FIFO test) in the identical shape: each was silently riding an undeclared "litellm is already ready" assumption, satisfied only by incidental test-order luck (something else importing litellm first). Moving the time axis (when the first call actually happens) made a previously-invisible shared assumption visible — same family as this session's own "moving *when* something happens surfaces one symptom per consumer" pattern, except here the consumer was test code, not a production call site. Production consumers can be enumerated ahead of time; tests riding an implicit global precondition can't — they only surface by actually running them. Fix in all 5 cases: state the real precondition explicitly (call `ensure_litellm_ready()` for real, or wait for the background thread) rather than relying on it implicitly.

PR #4423, opened 2026-08-12
fix(#4395): remove the last 4 chokepoint-bypassing bare litellm imports

tya5:
**[docs-maintainer]** — PR-4 of the #4395 arc: the last 4 chokepoint-bypassing bare `import litellm` sites (step 2 of the seam+gate restructure lead-coder/owner requested).

## The class of bug (owner directive: name it, not just the instances)

> "litellm に直接触るコードは (a) 継ぎ目を迂回できる (b) 未完成のモジュールを掴める" — (b) is new, born from #4417's own background thread.

Python places a module into `sys.modules` at the **start** of import, before its top-level code finishes (this is how circular imports work at all). Before #4417, every `import litellm` anywhere in the process was synchronous on whichever thread triggered it — a second, independent `import litellm` elsewhere would either hit a cache (fully-finished module) or trigger its own separately-serialized attempt. **Latent, never live.** #4417's dedicated background thread changed that: a bare `import litellm` on the main thread **while the warming thread is mid-import** can now observe the same, genuinely incomplete module object — attributes litellm's own `__init__` hasn't assigned yet. The owner's own repro: `AttributeError: module 'litellm' has no attribute 'exceptions'`, traced live to `llm.py`'s `_is_llm_timeout_exc`.

## The 4 sites, all converted to the same seam-aligned form

Structurally incapable of bypassing the chokepoint — read attributes off the **confirmed** module `ensure_litellm_ready()` returns, never a separate `import litellm` / `from litellm.X import Y` statement. Same shape as #4413's `pricing.py` fix and #4417's `_get_retryable_litellm_exceptions` hardening.

- **`llm.py`'s `_is_llm_timeout_exc`** — the site the owner's traceback named. Had **no readiness check at all** (a bare `import litellm`, no gate). Returns `False` (not a timeout) when litellm isn't ready — logically correct, not a placeholder: `exc` can't be a genuine `litellm.exceptions.Timeout` instance if litellm was never successfully imported.
- **`web.py`'s `_resolve_ssl_verify`** — called `ensure_litellm_ready()` and ignored the return value, then did its own unconditional `from litellm... import get_ssl_verify` right after. Also moved the litellm touch to only the fallback branch — an explicit `ca_bundle`/`verify_ssl` config never needed litellm at all, was importing it unconditionally regardless.
- **`registry/client.py`'s `RegistryClient.__aenter__`** — identical shape, identical fix.
- **`pricing.py`'s 3 cost-estimation functions** — these already checked `ensure_litellm_ready()`'s return before doing anything (**not independently exploitable today** — a confirmed non-None return already guarantees a fully-initialized module), but still had a separate `import litellm` statement textually. Converted anyway for seam alignment: the planned litellm-import gate (#4421) will allow `import litellm` nowhere in `src/` except `litellm_bootstrap.py` itself, so every call site needs this form regardless of whether it was independently exploitable.

All 4 unavailable-litellm fallbacks are safe and match pre-existing documented contracts: `_is_llm_timeout_exc` → `False`, `_resolve_ssl_verify` / `RegistryClient.__aenter__` → `True` (verify SSL, the secure default), pricing.py's 3 functions → their already-documented `None` sentinels.

## Not in scope for this PR (per lead-coder's ordering)

- **#4421** (the seam-widening + AST gate itself) — this PR is step 2 ("寄せる", align to the seam form) of that 3-step plan; step 1 (formalize the seam) and step 3 (the gate, which must land only after every site is aligned or it fails red immediately) are separate.
- Owner ruled #4417 (the background thread) **stays** — "freezing is worse, you can't observe anything while frozen." This PR moves forward, doesn't revert.

## Test plan

- [x] New: `test_4395_bare_import_race_llm_timeout_and_ssl_verify.py` (4 tests) — `_is_llm_timeout_exc` returns False when not-ready / never independently imports litellm / correctly classifies a real Timeout once ready; `_resolve_ssl_verify` never independently imports litellm.
- [x] Falsified: `test_is_llm_timeout_exc_never_imports_litellm_on_its_own` fails against pre-fix `llm.py` with an `AssertionError` at the exact `import litellm` line the owner's traceback named.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (211/215 baselined), `python scripts/test_tier_audit.py --strict` (0 Tier-4), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/llm/`, embedding-adjacent, web-fetch/registry-client/network-egress-adjacent, structured-output-adjacent — 915 passed.

## Arc note

`part of #4395`.

### DIFF
--- a/src/reyn/core/op_runtime/web.py
+++ b/src/reyn/core/op_runtime/web.py
@@ -196,14 +196,6 @@ def _resolve_ssl_verify(ctx: OpContext) -> bool | str:
       4. Both unset (None) → falls through to litellm.get_ssl_verify()
          (= SSL_VERIFY env → litellm.ssl_verify → SSL_CERT_FILE → True).
     """
-    # perf/log-routing chokepoint: a web op (e.g. the flagship web_search→agent
-    # pipeline) can be the FIRST litellm import in the process — funnel it
-    # through ensure_litellm_ready() so litellm's import-time console log
-    # routing (#2929) wraps this import too (idempotent; cheap on 2nd+ call).
-    from reyn.llm.litellm_bootstrap import ensure_litellm_ready
-    ensure_litellm_ready()
-    from litellm.llms.custom_httpx.http_handler import get_ssl_verify
-
     cfg = ctx.web_fetch_config
     if cfg is not None:
         if cfg.ca_bundle:
@@ -212,8 +204,35 @@ def _resolve_ssl_verify(ctx: OpContext) -> bool | str:
             return False
         if cfg.verify_ssl is True:
             return True
-        # cfg.verify_ssl is None → fall through to env-var chain
-    return get_ssl_verify()
+        # cfg.verify_ssl is None → fall through to litellm's own env-var chain
+    # perf/log-routing chokepoint: a web op (e.g. the flagship web_search→agent
+    # pipeline) can be the FIRST litellm import in the process — funnel it
+    # through ensure_litellm_ready() so litellm's import-time console log
+    # routing (#2929) wraps this import too (idempotent; cheap on 2nd+ call).
+    # Only reached here (not unconditionally at the top) — an explicit
+    # ca_bundle/verify_ssl config above never needs litellm at all.
+    #
+    # #4395 (owner-observed, live: `AttributeError: module 'litellm' has no
+    # attribute 'exceptions'` from a different chokepoint-bypass site):
+    # this used to call `ensure_litellm_ready()` and ignore its return
+    # value, then do its own unconditional `from litellm... import
+    # get_ssl_verify` right after — a chokepoint-bypass with a NEW failure
+    # mode PR-2's background warming thread exposed (Python places a
+    # module into `sys.modules` at the START of import, before its
+    # top-level code finishes; a second, independent import touch while
+    # the warming thread is mid-import can observe the same, genuinely
+    # incomplete module). Fixed the same way as `pricing.py`'s
+    # `_usage_object_for` (#4413): read `get_ssl_verify` off the
+    # ALREADY-confirmed module returned by `ensure_litellm_ready()`
+    # instead of a separate import statement. If litellm is unavailable,
+    # falls back to `True` (verify SSL) — the safe default, matching what
+    # this same fallback chain would eventually resolve to anyway absent
+    # any litellm-side override.
+    from reyn.llm.litellm_bootstrap import ensure_litellm_ready
+    litellm = ensure_litellm_ready()
+    if litellm is None:
+        return True
+    return litellm.llms.custom_httpx.http_handler.get_ssl_verify()
 
 
 async def handle_web_fetch(op: WebFetchIROp, ctx: OpContext) -> dict:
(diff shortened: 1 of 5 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/llm/test_4395_bare_import_race_llm_timeout_and_ssl_verify.py` 4 本、diff の★コード側で読了）

```
Q1 Tier   ── ★2。外部にある contract: #4395 ＋ ★owner の実環境で観測された
             `AttributeError: module 'litellm' has no attribute 'exceptions'`。
             第三者の性質ではない ── 主張しているのは
             ★「reyn の呼び出し側が独自に import しないこと」＝ reyn の性質。
Q2 転記   ── ✅ 無し。
Q3 誰が困るか ── ★実在。`builtins.__import__` を差し替えて
             「★litellm という名前の import が 1 度も起きないこと」を witness。
             ⚪ これは★モックではなく Python の import 機構そのものの傍受
             ∴ 「渡された」でなく ★「使わなかった」を直接見ている（#3859 の軸を回避）。
             ⚪ accept 側 1 本（ready 後は本物の Timeout を正しく分類）も在り、
             ★過剰発火を捕まえる役がある。
Q4 走らずに緑か ── ✅ skip 無し。CI 17 checks 緑（pytest 3.11/3.12 とも）。
Q5 累積   ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致（全 4 本 `Tier 2:`、Q1 の答えと同じ）。
```
✅ **禁止形 0 件**（mock/timeout marker/`sleep(N)`/待ち予算/private state assert）。

## ⚠️ 1 点、★block はしませんが記録します

`test_web_fetch_resolve_ssl_verify_...` が **`_FakeEventLog` / `_FakeWorkspace` という
★手書きの代役**を使っています。**CLAUDE.md の「実インスタンスが安く作れるなら
collaborator を偽装しない」に★文言上は触れます。**

**block しない理由:**
```
① ★main に既に同じ形が在る ── tests/core/test_web_fetch_ssrf_redirect_1956.py:51-57
   （`workspace=type("W", (), {})()` ＋ 手書き `_Ev`）
   ∴ ★この PR が持ち込んだ新しい形ではなく、★既存の局所 idiom への追随
② 検査対象 `_resolve_ssl_verify` は ★workspace / events に一切触れない
   ∴ 偽装が「実型に無いフィールドを黙って運ぶ」危険（#3037 の形）が★成立しない
```
🔴 **∴ 「この PR を止める理由」ではなく「★idiom 側の負債」**として扱います。
⚪ **直すなら `test_web_fetch_ssrf_redirect_1956.py` と同時に**（片方だけ直すと
2 つの流儀が並ぶだけ）── **今夜は起票しません**（開いたチケットは維持費、
かつ★これ単体では誰も困っていない）。

⚪ **doc drift**: `docs/` の `import litellm` 出現は 1 件のみ
（`docs/deep-dives/proposals/0052-high-cost-model-preconfirm.md:43`）で、
**★提案文書の擬似コード ∴ この PR が falsify する claim ではありません。**

**merge します。** 🔵 **これが owner の画面の
`litellm has no attribute exceptions` を消す本体です。**

PR #4419, opened 2026-08-12
fix(#4418): tiktoken's import-time egress honours SSL_VERIFY/timeout too

tya5:
**[tui-coder]** — fix(#4418): tiktoken's import-time egress honours SSL_VERIFY/timeout too

## What

`#3075`'s requirement was completeness — EVERY reyn-originated egress honours the standard env, zero exceptions — and one path was missing from the enumeration:

```
tiktoken/load.py:17   requests.get(blobpath)
  verify defaults to True -> ignores SSL_VERIFY=false
  no timeout either (#4395)
  fires from INSIDE `import litellm` (litellm_core_utils/default_encoding.py)
```

This is genuinely reyn-originated egress (only fires because reyn imports litellm) but reyn cannot pass `verify=` into a third party's call site — the only lever is `requests`' own per-call default. Not grep-detectable by #3075's own completeness gate (same conclusion #4410 reached for a different import-time hazard): that gate enumerates reyn's own egress *classes*, not "does a library reyn imports cause its own egress as a side effect of import" — a new enumeration axis, not a new class within the old one.

## Fix

- `litellm_bootstrap.py`'s `ensure_litellm_ready()` (the sole `import litellm` chokepoint) now wraps that import in `_third_party_import_egress_honours_standard_env`, which patches `requests.sessions.Session.request` to inject `verify`/`timeout` **defaults** (never overriding an explicit value) resolved from the same standard env every other #3075 egress class reads (`SSL_VERIFY`/`SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE`) — scoped to the duration of that one import only, not patched globally/forever.
- A `verify=False` resolution runs the same never-silent audit hook every other #3075 class uses (`note_ssl_verify_disabled`) — one WARN + one `network_ssl_verify_disabled` P6 event, keyed `"tiktoken-import"`.
- `ensure_litellm_ready()` gains an optional `events: EventLog | None = None` param (default `None`, every existing call site unmodified); wired for real at `RegistryClient.__aenter__` (which already tracks `self._events`).
- Doc-drift fix in the same PR: `test_network_egress_env_completeness_3075.py`'s own enumeration docstring gets a bullet for this axis.

## Test plan

- [x] 3 new tests (real `requests.sessions.Session`, no mocks): defaults injected + restored, `SSL_VERIFY=false` honoured with the audit-event, explicit caller `verify=`/`timeout=` never overridden.
[...]

PR #4424, opened 2026-08-12
feat(#4422): a failed litellm import names a cause and a remedy

tya5:
**[tui-coder]** — feat(#4422): a failed litellm import names a cause and a remedy

## What

Owner's own question forced this: "everyone else who isn't me, what are THEY supposed to do?" — the `pip install --force-reinstall --no-deps litellm` fix lead-coder walked the owner through by hand (#4395/#4399) never reaches the next operator who hits the same failure, because `ensure_litellm_ready()`'s warn-once log line just says "import litellm failed" with no cause.

## Fix

- Reuses #4399's own judgment (locate tiktoken's cache dir, resolve the expected cl100k blob hash, compare) rather than re-deriving it — extracted from `scripts/diag_tiktoken_cache.py` into a new shared module, `src/reyn/_tiktoken_diag.py`, since a `src/reyn/` runtime path cannot import from `scripts/` (it ships in a repo checkout, not an installed non-editable `reyn` package). The script is now a thin print wrapper over `reyn._tiktoken_diag.diagnose()`.
  - **Noted behavior change in the script's own docstring**: it now imports `reyn` (to reach the shared module), so it sees the SAME env `ensure_litellm_ready` sees, not tiktoken's raw pre-redirect default — deliberate, since "why did reyn's own import just fail" is the actual #4422 use case. An operator who wants the raw pre-reyn reading needs a plain interpreter that never imports `reyn`.
- `ensure_litellm_ready()`'s warn-once WARNING now appends one of three messages:
  - cache file missing → reinstall remedy, phrased as a *possibility* (not asserted as "was deleted" — lead-coder's review correction after the owner's own question: absence alone isn't proof of deletion) that also self-diagnoses.
  - cache file present but sha256 mismatch → states the mismatch as fact (directly read) + the same remedy.
  - no cache-specific signal → falls back to the #4418 network/cert remedy (`SSL_VERIFY`/`SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE`).
  - Never raises past the WARN it augments.
- Log-only — whether this also surfaces on the conv-pane is a separate, still-open owner:decide call (#4380); this PR only touches the existing `logging.warning()` call.

## Depends on

Built on top of #4419 (`_diagnose_import_failure_for_log` sits next to that PR's `_third_party_import_egress_honours_standard_env` and the fallback message references its `SSL_VERIFY` remedy) — #4419 merged to `main` while this PR was in progress, so this is now based directly on `main` and the diff is exactly the #4422-specific delta (confirmed via `git diff --stat` against `main`'s tip).

## Test plan

- [x] 9 new tests, real collaborators (no mocks): `reyn._tiktoken_diag.diagnose()` against real files on disk (including a real copy of litellm's own bundled tokenizer blob for the "matches" case — a genuinely correct sha256 can't be produced any other way without a real fetch); message-selection branches via a monkeypatched `diagnose()` returning the real `TiktokenCacheDiagnosis` type; one end-to-end test forcing a real `import litellm` failure (`builtins.__import__` patching, same technique as `test_4395_litellm_import_not_recached_on_failure.py`) reading the diagnosis off the real warn-once log line via `caplog`.
- [x] Falsify-verified: reverted the warn call to its pre-#4422 form, confirmed the end-to-end test RED, restored, confirmed GREEN.
- [x] `ruff check`, `mypy_ratchet.py`, `test_tier_audit.py --strict`, `verify_module_docstrings.py`, `check_tests_path_literal_reference.py` — all clean.
- [x] Full regression sweep (70 tests: egress completeness, litellm cooldown/lazy-load, RegistryClient) — all passed.
- [ ] (skipped — log-only change, nothing to visually verify)

part of #4422

### DIFF
--- a/scripts/diag_tiktoken_cache.py
+++ b/scripts/diag_tiktoken_cache.py
@@ -11,113 +11,60 @@ verbatim into an issue/chat from an environment where nothing else can
 leave (the reason this script exists at all: #4395's owner-side repro
 couldn't run a one-off diagnostic command that shows paths).
 
-Distinguishes the 3 ways tiktoken's own cache-hit check
-(``tiktoken/load.py``) can miss, matching #4395's own A/B/C:
+#4422: the judgment itself (locate the cache dir, resolve the expected
+cl100k blob hash, compare) now lives in :mod:`reyn._tiktoken_diag`, shared
+with ``reyn.llm.litellm_bootstrap``'s own runtime failure-message —
+this script is a thin print wrapper over :func:`reyn._tiktoken_diag.diagnose`,
+not a second copy of the algorithm.
 
-  A. bundled_file_exists=False  -> the cache file was never written at all
-     (a fresh install, or something cleared the cache dir).
-  B. sha256_matches=False (bundled_file_exists=True) -> the cached file's
-     content doesn't match what THIS installed tiktoken expects — a
-     version mismatch between whatever wrote the cache and the tiktoken
-     now running. tiktoken deletes a mismatched file and re-fetches over
-     the network on every call until this is resolved.
-  C. custom_cache_dir_set=True -> operator/litellm set
-     CUSTOM_TIKTOKEN_CACHE_DIR themselves; if that path doesn't already
-     have a valid cache, this script's other flags describe THAT
-     directory, not tiktoken's default.
+**Behavior change from the pre-#4422 version**: this script now imports
+``reyn`` (to reach ``reyn._tiktoken_diag``), so it observes the SAME
+environment ``ensure_litellm_ready`` itself sees — including
+``reyn/__init__.py``'s own ``TIKTOKEN_CACHE_DIR``/``CUSTOM_TIKTOKEN_CACHE_DIR``
+redirect, which runs unconditionally before any ``reyn.*`` code, this
+script included, can execute at all. That is deliberate: "why did reyn's
+own litellm import just fail" (the #4422 use case) needs reyn's OWN view,
+not tiktoken's un-redirected default. An operator who specifically wants
+tiktoken's RAW pre-reyn default (bypassing the redirect entirely) needs a
+plain interpreter that never imports ``reyn`` — this script is no longer
+that; run the four ``reyn._tiktoken_diag`` functions by hand from such an
+interpreter instead if that isolated reading is what's needed.
 
 This is a read-only probe: it locates the currently-effective cache
 directory and inspects a candidate file already inside it (if present) —
-it never fetches anything, never writes anything, and never imports
-``litellm``\'s or ``reyn``\'s own package init (so it reports what tiktoken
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/llm/test_4422_litellm_import_failure_diagnosis.py`、9 本を diff の★コード側で読了。以下は PR 本文の説明ではなくテスト本体に対する回答）

```
Q1 Tier   ── ★2。外部にある名指しの contract: #4422 ＋ owner の
             「私以外の人はどうすれば良いの？」＝ ★利用者に見える約束。
             第三者の性質でない（tiktoken/litellm の振る舞いではなく
             ★reyn のログ行が何を名指すか）／過去バグの指紋でもない。
Q2 転記   ── ✅ 無し。両辺に同じ式は無い。
             ⚠️ `assert "SSL_VERIFY" in message` は実装の literal に近いが、
             ★分岐の 選択 を判別する用途 ∴ 転記ではない。
Q3 誰が困るか ── ★実在の消費者。end-to-end 1 本
             (`builtins.__import__` を落として実 warn 行を caplog で読む) が
             ★配線を witness している ── 「単体で呼べる」と「実際に届く」の差。
             ⚪ diagnose() の 3 本は★実ファイル（litellm 同梱 blob の実コピー）
             ∴ テスト自身が組んだ構成ではない。
Q4 走らずに緑か ── ⚠️ ★skip が 2 つある（module 単位 skipif ＝ tiktoken_ext 不在時／
             `_real_bundled_blob_path()` が None の時）。**PR 本文はこれを書いていません。**
             ⚪ ただし litellm は `pyproject.toml:42` の★必須依存 ∴ CI では走る
             ── 「走らずに緑」にはならない ∴ ★block しません（次回は本文に一行を）。
Q5 累積   ── ✅ 無し。list() も producer 待ちも無い。
Q6 宣言 Tier ── ✅ 一致（全 9 本が `Tier 2:` を宣言、Q1 の答えと同じ）。
```
✅ **禁止形 0 件**（MagicMock/AsyncMock/`unittest.mock`/timeout marker/`sleep(N)`/
待ち予算/private state assert）。**monkeypatch 先は `reyn._tiktoken_diag.diagnose` で、
返すのは★実型 `TiktokenCacheDiagnosis` ∴ 発明フィールドの余地なし**（#3037 の形を回避）。

🔵 **私自身のレビュー訂正が★テストとして固定されているのを確認しました:**
`test_missing_file_message_never_asserts_deletion_as_fact` ──
**「不在」を「削除された」と断定しない**。⚪ **本日 owner に対して私が実際に犯した
誤り**（#4395 に訂正を記録）が、**★機械が守る形になりました。**

⚪ **doc drift**: `docs/` に `diag_tiktoken_cache` / `_tiktoken_diag` /
`TIKTOKEN_CACHE_DIR` の記述は★0 件（`git grep` on `origin/main`）∴ 追随不要。

**merge します。**

PR #4425, opened 2026-08-12
fix(#4395): the last 2 sys.modules-presence guards + a pre-existing test regression

tya5:
**[docs-maintainer]** — PR-5 of the #4395 arc: architect's re-audit of #4415 caught 2 more `sys.modules`-presence guards, and this landed a live test regression on `main` while fixing them.

## Why this exists

Architect re-checked #4415's own "9 call-order-safe" classification after #4423 landed, and found the classification had gone stale the moment #4417 (the background warming thread) landed — `llm.py`'s `_is_llm_timeout_exc` was classified "safe by call order," and it's the exact site the owner's `AttributeError` traced to. Same root cause for both: `"litellm" in sys.modules` proves an import **started**, never that it **finished** (Python inserts a module into `sys.modules` at the start of import, before its top-level code runs). Before #4417, every import was synchronous on whichever thread triggered it, so this shape was latent. After #4417, it's live.

Two more sites use the identical shape:

- **`llm.py`'s `shutdown_logging()` / its call site in `run_async`** — the call site re-checked `"litellm" in sys.modules"` before calling; removed (redundant once the function owns the correct check itself) and `shutdown_logging()` now gates on `is_litellm_ready()` internally, reading the confirmed module via a safe (non-racing, reached only after readiness) explicit submodule import instead of an unconditional `from litellm... import`.
- **`interfaces/cli/__init__.py`'s `main()`** — the `AuthenticationError` boundary used the same shape; now gates on `is_litellm_ready()` and reads `AuthenticationError` off the confirmed module.

## Live regression found and fixed along the way

`test_llm_call_retry.py` and `test_3905_cli_authentication_error_boundary.py` both do their own direct `import litellm` / `from litellm.exceptions import X`, bypassing the chokepoint entirely — `is_litellm_ready()` stays False even though litellm is genuinely importable. #4423's own `_is_retryable_exc`/`_is_llm_timeout_exc` fixes silently under-classified in these tests as a result: **4 failures already present on `main` before this PR touched anything** (confirmed via `git stash` — reproduced against the currently-merged state, not introduced by this branch). Real production code never hits this gap (a real completion that raised a genuine litellm exception already confirmed readiness on the way there) — fixed by stating each test's own real precondition explicitly (`ensure_litellm_ready()`) rather than relying on litellm being importable at all, matching the pattern established for 5 other test files earlier in this arc. `test_3905`'s "never touched" test additionally needed to reset `_litellm_ready` (not just strip `sys.modules`) — the flag is a process-global that would otherwise leak `True` from an earlier test in the same file and silently reintroduce the exact race being tested against (`sys.modules["litellm"]` on a name just removed → `KeyError`).

## Test plan

- [x] Falsified: reproduced the 4 pre-existing `test_llm_call_retry.py`/CLI-boundary failures against the current `main` state via `git stash` before making any change — confirmed they're a live regression from #4423, not new.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (211/215 baselined), `python scripts/test_tier_audit.py --strict` (0 Tier-4 across 18 tests in the 2 touched files), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/llm/`, `tests/interfaces/` (the full directory), embedding/registry-client/network-egress-adjacent — **2374 passed, 7 skipped** (missing optional `terminaltexteffects` extra, unrelated to this PR), 0 failed.

## Arc note

`part of #4395`. `#4421` (the seam+gate structural follow-up) is next, dispatched but paused for this urgent fix per lead-coder's explicit priority ordering.

### DIFF
--- a/src/reyn/interfaces/cli/__init__.py
+++ b/src/reyn/interfaces/cli/__init__.py
@@ -62,26 +62,38 @@ def main() -> None:
     # misses get the friendly one-liner, openai-shaped ones still traceback
     # — an HONEST partial improvement, not the old "all providers" claim.
     #
-    # ``sys.modules`` gate BEFORE the litellm import (not just "import
-    # litellm.exceptions and see"): importing ``litellm.exceptions`` pulls in
-    # ALL of litellm (measured directly: 1.76s cold, matching #3671's own
-    # cold-import figure) — paying that for an exception that was never
-    # litellm's in the first place would tax every unrelated late failure
-    # (a bug in file-handling code, say) with an irrelevant multi-second
-    # startup-cost regression. If litellm was never imported, this exception
-    # cannot possibly be a litellm one — skip the check entirely. If it WAS
-    # imported (an LLM call was actually attempted), the re-import is a
-    # sys.modules cache hit (measured: ~0.7 microseconds, not the 1.76s
-    # cold cost) — free.
-    import sys
+    # ``is_litellm_ready()`` gate BEFORE the litellm touch (not just "import
+    # litellm.exceptions and see"): reading ``litellm.exceptions`` would
+    # otherwise force litellm's full cold import (measured directly: 1.76s
+    # cold, matching #3671's own cold-import figure) — paying that for an
+    # exception that was never litellm's in the first place would tax every
+    # unrelated late failure (a bug in file-handling code, say) with an
+    # irrelevant multi-second startup-cost regression. If litellm was never
+    # imported, this exception cannot possibly be a litellm one — skip the
+    # check entirely. If it WAS imported (an LLM call was actually
+    # attempted), reading the confirmed module is free.
+    #
+    # #4395/#4421 (architect finding): this used to gate on ``"litellm" in
+    # sys.modules`` — Python places a module into ``sys.modules`` at the
+    # START of import, before its top-level code finishes, so that check
+    # only ever proved the import STARTED, not that it FINISHED; #4417's
+    # background warming thread turned that into a live race (not yet
+    # observed here the way #4423's `_is_llm_timeout_exc` was, but the
+    # SAME shape: this handler runs on a startup failure, which can race
+    # the warming thread's own in-flight import and grab a genuinely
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/interfaces/test_3905_cli_authentication_error_boundary.py` ＋ `tests/llm/test_llm_call_retry.py` の差分をコード側で読了）

```
Q1 Tier ── ★2。外部の contract: #4395 ＋ owner 実環境の AttributeError。
           第三者の性質ではない（主張は「reyn の境界が確認済み module を読むこと」）。
Q2 転記 ── ✅ 無し。
Q3 誰が困るか ── ★実在。#4425 が直しているのは ★main で実際に赤い 4 件
           （lead-coder が `git checkout 710ddead0` で独立確認済）。
           ⚪ 各テストが ★自分の前提を明示（`ensure_litellm_ready()`）する形になり、
             ★他テストの副作用に寄りかからなくなった ── これが修正の本体。
Q4 走らずに緑か ── ✅ 否。★逆に「全件だと緑になってしまう」のが今回の欠陥で、
           限定実行で赤を再現できることを確認済。
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

🔵 **docstring の訂正が特に良い**: `"litellm" in sys.modules` が
**「import が★開始した」の witness であって「★完了した」の witness ではない**という
一般形が、**境界のコード自身に残渣として書かれました。** ⚪ **#4417 の暖機スレッドが
これを latent から live に変えた経緯まで書いてあり、★次に読む人が同じ形を書き直せません。**

**merge します。** ⚠️ **`verification-hazards.md` への「CI-green ≠ main-green」項の追加は
★この PR で足りていません** —— broker で依頼済みですが、**次の PR で構いません**
（main が赤のまま待つ理由がない ∴ 先に着地させます）。

PR #4428, opened 2026-08-12
fix(#4421): litellm import seam — finish migration, land the AST gate

tya5:
**[docs-maintainer]** — #4421: the seam+gate structural fix, steps 2 and 3.

## The class this closes (owner's own framing)

> litellm に直接触るコードは (a) 継ぎ目を迂回できる (b) 未完成のモジュールを掴める

**(a)**: bypasses `ensure_litellm_ready()`'s cooldown/warming-thread machinery entirely. **(b)** is new — born from #4417's background warming thread: Python places a module into `sys.modules` at the *start* of import, before its top-level code finishes, so a second import racing the warming thread can observe a genuinely incomplete module (#4395's owner-observed live repro: `AttributeError: module 'litellm' has no attribute 'exceptions'`).

## Step 2 — finish migrating the remaining sites

A fresh AST census of `src/reyn` (not inherited from #4415's own count, which had already gone stale once mid-arc) found 6 more production sites doing their own `import litellm` outside `litellm_bootstrap.py`:

- `llm.py`'s `_close_litellm_async_clients`, `_single_deployment_router`, `_may_need_responses_endpoint`, `_streaming_capability` — all "call-order safe" in the exact shape #4415/#4417 already showed can go stale silently. Converted to explicit `is_litellm_ready()` gates rather than relying on the call-order argument staying true forever. `_single_deployment_router` has no fallback (a Router construction feeds straight into a real completion) — a not-ready state now raises `LitellmUnavailableError` explicitly. The other two already had a broad except-fallback; "not ready" now maps to the same conservative answer they already return for other lookup failures.
- `litellm_provider.py`'s embed-retry loop — already correctly gated, but still had a separate `import litellm` statement after the check; converted to read the return value directly (matches `pricing.py`'s #4413 conversion).
- `litellm_bootstrap.py` itself: `litellm.litellm_core_utils.logging_worker` isn't one of the submodules litellm's own `__init__` eagerly populates as an attribute (verified empirically, unlike `litellm.types.utils`) — now imported once inside `ensure_litellm_ready()`'s own setup, so `shutdown_logging()` reads it via plain attribute access afterward, no import statement of its own.

**Re-census after: zero** `import litellm` / `from litellm...` statements remain in `src/reyn` outside `litellm_bootstrap.py` (excluding `dev/testing/`'s LLMReplay — a test-replay seam, not a production one, #4415's own "dev-only" bucket).

## Step 3 — the gate

`tests/security/test_4421_litellm_import_seam.py`, an AST-based scan mirroring #4410's own `reyn.api.safe` import-boundary gate precedent exactly: a detector function, the main scan, a positive-twin accept-side test (proving the detector isn't vacuously green), and a third test new here — proving the seam-file *exclusion* is actually exercised (not pointing at a path that happens not to import litellm). Landed only now, after the census above confirmed zero remaining offenders — landing it earlier would have failed red on unmigrated sites (per lead-coder's explicit ordering).

**Known limit** (same framing as #4410's own): this gate sees literal import statements only. It does not close the attribute-access form architect flagged as a separate axis (#4415's own stated limit) — that needs runtime detection, not a static gate.

## Scope note — the "1-turn, 0 network calls" canary is NOT in this PR

The design discussion on the #4421 issue settled on a second, complementary mechanism: a runtime canary asserting litellm's own import-time tiktoken fetch causes zero egress in the default (bundled-cache-present) state, with the observation window set to **one full turn** (not one import) so it also covers the 9 call-order-safe sites the import-only window would miss.

Checked `dev/testing/replay.py` as instructed before starting: `LLMReplay.install()` monkeypatches `litellm.acompletion`/`litellm.aembedding`, so a replayed turn's own completion calls are already network-free — but `install()` itself does `import litellm`, meaning the canary still needs its own network-observation harness wrapped around *that* import specifically (likely building on #4419's own `_third_party_import_egress_honours_standard_env` machinery). That's a distinct, substantial piece of work from the AST gate landed here — flagging transparently rather than silently shipping a partial "gate" under the same name. Reported to lead-coder for scoping/priority before starting it as a follow-up.

## Test plan

- [x] Falsified: injected a real `import litellm` into `pricing.py`, confirmed the gate fails with the exact offending line named (`src/reyn/llm/pricing.py:637 (import litellm)`), reverted, confirmed green.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (211/215 baselined), `python scripts/test_tier_audit.py --strict` (0 Tier-4), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/llm/`, `tests/security/` (the full directory, including the new gate), embedding-adjacent — 1459 passed, 26 skipped (unrelated missing extras), 0 failed. (One transient timing-margin flake in `test_measured_startup_latency_delta_from_deferring_litellm` under parallel background load — reproduced clean in isolation and on a clean re-run of the full scoped sweep; pre-existing test, unrelated to this diff.)

## Arc note

`part of #4421`.

### DIFF
--- a/src/reyn/data/embedding/litellm_provider.py
+++ b/src/reyn/data/embedding/litellm_provider.py
@@ -417,13 +417,17 @@ class LiteLLMEmbeddingProvider:
                 # up): this loop has no fallback within one attempt — the
                 # axis② cooldown protects fallback-having callers, not
                 # this one; see `ensure_litellm_ready()`'s own docstring.
-                if await asyncio.to_thread(ensure_litellm_ready, ignore_cooldown=True) is None:
+                litellm = await asyncio.to_thread(ensure_litellm_ready, ignore_cooldown=True)
+                if litellm is None:
                     raise LitellmUnavailableError(
                         "import litellm failed — see the reyn.llm."
                         "litellm_bootstrap warn-once log line for the "
                         "underlying cause",
                     )
-                import litellm
+                # #4395/#4421 (seam alignment): reads the module off the
+                # return value rather than a separate `import litellm` —
+                # the only form the litellm-import gate allows outside
+                # `litellm_bootstrap.py` itself.
                 response = await self._aembedding_bounded(
                     litellm,
                     model=effective_model,
--- a/src/reyn/llm/litellm_bootstrap.py
+++ b/src/reyn/llm/litellm_bootstrap.py
@@ -547,6 +547,17 @@ def ensure_litellm_ready(
                     # REQUESTS_CA_BUNDLE via get_ssl_verify(), so this is the one
                     # missing piece for full conformance.
                     litellm.aiohttp_trust_env = True
+                    # #4395/#4421 (seam alignment): `litellm.litellm_core_
+                    # utils.logging_worker` is NOT one of the submodules
+                    # litellm's own `__init__` eagerly populates as an
+                    # attribute (unlike e.g. `litellm.types.utils` or
+                    # `litellm.llms.custom_httpx.http_handler`, verified
+                    # empirically) — imported here, once, inside the ONE
+                    # place `import litellm` itself is allowed, so
+                    # `llm.py`'s `shutdown_logging()` can read it via plain
+                    # attribute access afterward instead of needing its
+                    # own submodule import statement outside the seam.
+                    import litellm.litellm_core_utils.logging_worker  # noqa: F401
                     result = litellm
                 except Exception:  # noqa: BLE001 — best-effort; never block the caller on this
                     result = None
--- a/src/reyn/llm/llm.py
+++ b/src/reyn/llm/llm.py
@@ -487,23 +487,21 @@ async def shutdown_logging() -> None:
     race (a shutdown racing a still-in-flight warm-up could grab a
     genuinely incomplete module). Both the call-site guard and this
     function's own body were fixed the same way as #4423: gate on
-    ``is_litellm_ready()`` (the real "genuinely finished" signal) before
-    touching litellm at all. ``litellm.litellm_core_utils.logging_worker``
-    is not one of the submodules litellm's own ``__init__`` eagerly
-    populates as an attribute (unlike e.g. ``litellm.types.utils``), so
-    this still needs its own ``import`` statement for that specific
-    submodule — but only ever reached AFTER ``is_litellm_ready()``
-    confirms the TOP-level package finished initializing, so this is a
-    normal, non-racing import (the top package's own import lock has
-    already been released; importing one of its submodules afterward
-    behaves like any other already-safe-to-import module).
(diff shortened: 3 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/security/test_4421_litellm_import_seam.py` をコード側で読了）

```
Q1 Tier ── ★1（宣言どおり）。外部の contract: #4421 の「litellm に触るのは 1 面だけ」＋
           前例 #4410 の `reyn.api.safe` gate（同じ形・同じ Tier）。
           ⚪ 第三者の性質ではない ── 主張は ★reyn 自身の import 文の在り処。
Q2 転記 ── ✅ 無し。
Q3 誰が困るか ── ★実在（production の全 src/reyn）。
           🔵 ★vacuity を 3 重に閉じているのが特に良い:
             ① positive-twin ── 5 種の実違反を detector が★実際に捕まえることを確認
             ② accept 側 ──「無関係な import」「★相対 import」を誤検出しない
             ③ ★除外が仕事をしていることの確認 ──
                「litellm_bootstrap.py が★本当に litellm を import している」を assert。
                ⚪ 除外先が「存在しないパス / そもそも import しないファイル」に
                   なっていたら★除外は空回りだが gate は緑 ── その形を名指しで閉じている。
Q4 走らずに緑か ── ✅ skip 無し。CI 緑。
Q5 累積 ── ✅ `src/reyn` の walk ∴ ★リポジトリが上限。テスト外の何かが決めていない。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

🔵 **docstring が★限界を自分で名指ししているのが要件どおり**:
「**AST が見えるのは reyn 自身の import 文だけ** ── `sys.modules` 参照や `getattr`
経由は★構造上 見えない。この gate は『★継ぎ目の外の literal な import 文』という
半分を閉じるのであって、クラス全体ではない」。⚪ **architect の #4415 recount が
まさにその穴を 1 度踏んだ経緯まで書かれており、★次に読む人が
「gate が在るから安全」と読み違えません。**

⚪ **順序の記録も要件どおり**: gate は★step 3 で、全 production site の移設
（#4413/#4417/#4420/#4423/#4425）後に着地。**しかも「以前の数え直しを継承せず、
★この gate が実際に walk するコードに対して再測定した」**と明記
（#4415 が #4417 の着地で 1 度陳腐化した経緯を踏まえて）。

⚪ **doc drift**: `docs/` に `ensure_litellm_ready` / `litellm_bootstrap` の記述は★0 件
∴ 追随不要。

**merge します。** ⚠️ **#4421 は★閉じません** —— **gate が閉じたのは「literal な
import 文」の半分**で、**属性アクセス経由の軸は残っています**（docstring 自身が明記）。

PR #4429, opened 2026-08-12
feat(#4380): 6 more lifecycle kinds reach the conv pane, one bundles

tya5:
**[tui-coder]** — feat(#4380): 6 more lifecycle kinds reach the conv pane, one bundles

## Premise (per lead-coder's instruction — the fact-enumeration this PR builds on)

Before any design/code, I enumerated the CURRENT 14 marker kinds `lifecycle_forwarder.py` already surfaces and their `[↑]`/`[⚠]`/`[✗]`/`[↻]` weight convention: https://github.com/tya5/reyn/issues/4380#issuecomment-5267308947

Then, after the owner's ruling on WHICH 6 kinds + weight, I traced every one of those 6 kinds' actual emission sites before writing code — full trace posted to the issue thread. One finding **corrected the assumed design**: `force_close_triggered` and `router_force_close_handoff` were assumed to be "one event, two co-firing halves" (a bundle candidate); tracing the actual call sites showed they're two independent mechanisms (different files, different layers) that never fire together for the same turn — so no bundling rule was built for that pair, because there's no real case to bundle.

## What

- **6 new markers** (`lifecycle_forwarder.py`): `force_close_triggered`, `router_force_close_handoff`, `turn_cancelled`, `chain_timeout`, `permission_denied`, `intervention_denied` — all dim, same `[✗ …]` weight as the existing family (owner: "don't stand out").
- **`turn_cancelled` closes a genuinely NEW gap**, not a duplicate: the existing history-append (`router_loop.py`'s "Turn interrupted by user.") has explicitly no outbox side-effect (`router_host_adapter.py`'s own docstring) — it was only ever visible on the NEXT restore, never live in the current session.
- **Only `permission_denied` bundles** (owner ruling, after I flagged that the assumed force-close pairing doesn't exist): consecutive identical `(kind, path, reason)` denials collapse into one row with a count. `reason` was owner's own addition to my initial `(kind, path)` proposal — same capability/path, different reason, is still a different fact. `intervention_denied` is explicitly **not** bundled (my recommendation, owner adopted it) — a different `intervention_id` is a different unanswered question even when `kind` matches.

## Bundling mechanism

`lifecycle_forwarder.py` is a pure producer (writes to an `asyncio.Queue`, holds no `FlowView` reference) — the bundling had to live on the **consumer** side. New `self._last_lifecycle_marker` tracker in `app.py`'s `_ingest_frame`, mirroring `_coalesce_tool_result`'s own established precedent (find the existing entry, `entry.set_item(replace(...))`, no second append) rather than inventing a new mechanism. Deliberately the **most conservative** rule available — compares only against the **immediately-preceding row**, never a turn-wide buffer, no new turn-boundary tracking added. An intervening frame (a different marker, a tool call, an agent reply) resets the chain. Per the owner's own framing this is correct, not a limitation: two identical denials with something else between them are not the same fact as two back-to-back ones.

**What stays unbroken** (owner's explicit constraints):
- Counts are never dropped.
- Different reasons/paths never collapse into one line, even for the same op kind.
- The audit log itself is unchanged — one event per denial, always; only the DISPLAY collapses.

## Test plan

- [x] 15 new tests, real collaborators throughout (real `ChatLifecycleForwarder` + `Event`; real mounted `TextualChatApp` + a queue-backed `ClientTransport` driving the actual frame pump — no direct `_ingest_frame(...)` calls, no mocks):
  - `tests/runtime/test_4380_lifecycle_forwarder_new_markers.py` (10) — producer side.
  - `tests/interfaces/test_4380_ingest_frame_bundles_permission_denied.py` (5) — consumer side, including the real `FlowView` entry count shrinking, not just the text list.
- [x] Falsify-verified: reverted the bundling block to a no-op → 2/5 consumer-side tests RED with the exact expected mismatch; restored, confirmed GREEN.
- [x] Full regression sweep — every lifecycle-forwarder/pipeline/tool-call suite (113) + every #4387-paging-adjacent suite this shared `_ingest_frame`/`app.py` code path touches (210) — both green.
- [x] `ruff check`, `mypy_ratchet.py`, `test_tier_audit.py --strict`, `verify_module_docstrings.py`, `check_tests_path_literal_reference.py` — all clean.
- [ ] (skipped — per the standing waiver; owner has left for the day, visual gate does not block merge)

**Test plan / TESTS-READ**: this PR touches `tests/` — holding for a reviewer's TESTS-READ note before merge, no self-merge.

part of #4380

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -1048,6 +1048,19 @@ class TextualChatApp(App):
         # deterministic args_hash, meta["op_id"]) so a later completion/failure
         # frame transitions the SAME entry RUNNING → SUCCESS/ERROR (CC parity).
         self._running_tools: "dict[object, Entry[OutboxMessage]]" = {}
+        # #4380: the most recently appended lifecycle marker eligible for
+        # count-bundling (currently only ``permission_denied``, the ONE
+        # kind owner ruled bundles — see ``_ingest_frame``'s own docstring
+        # for the full rule). ``None`` whenever the immediately-preceding
+        # frame was anything else — the bundle chain resets on ANY other
+        # frame, by design (owner: "something happened in between" IS the
+        # fact, not a defect to work around). Tuple shape:
+        # ``(bundle_key, entry, count, base_text)`` — ``base_text`` is the
+        # marker's own text BEFORE any ``×N`` suffix, kept separately so
+        # re-deriving the bumped text never re-parses a previous bump.
+        self._last_lifecycle_marker: (
+            "tuple[object, Entry[OutboxMessage], int, str] | None"
+        ) = None
         #: #4187: voice input. ``None`` until F2 is first pressed — lazy so an
         #: install without the ``reyn[voice]`` extra never pays anything for a
         #: key nobody used (mirrors ``text_effect``'s import-on-press shape).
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`test_4380_ingest_frame_bundles_permission_denied.py` ＋ `test_4380_lifecycle_forwarder_new_markers.py` をコード側で読了）

```
Q1 Tier ── ★2 / 2b（宣言どおり）。外部の contract: #4380 の owner 裁定
           （6 kind を dim で追加 ／ 束ねられるものは束ねる ／ 件数を落とさない ／
             異なる理由を 1 行に潰さない）── ★すべて issue に逐語で在る。
Q2 転記 ── ✅ 無し。`assert only.text == "[✗ force close: …]"` は ★値の assert
           であって式の転記ではない（「一致」でなく「値」を assert する形）。
Q3 誰が困るか ── ★実在（会話ペインの利用者）。
           🔵 owner の制約が ★1 つずつ 独立したテストになっている:
             ・different ★path → 別行（同 kind でも同事象ではない）
             ・different ★reason → 別行（★owner の訂正がそのまま試験になっている）
             ・★間に別の frame が挟まると 鎖が切れる（最も保守的な規則）
             ・★件数が落ちない（×3 / ×6）
           ⚪ `force_close_triggered` と `router_force_close_handoff` が
             ★異なる文言であることを assert ── tui-coder が実測で否定した
             「1 事象・2 event」の誤りが ★二度と戻らない形。
Q4 走らずに緑か ── ✅ skip 無し。CI 緑。
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

⚠️ **非 block の観察**: **marker 文言を★逐語で pin しています。** 文言を変えるたびに
テストも変わりますが、**★これは利用者に見える文字列そのものが契約**（owner が
「dim で・既存と同じ重さ」と裁定した対象）∴ **format pin ではなく★値の pin と読みます。**

⚪ **doc drift**: `lifecycle_forwarder` / marker 文言を記述する doc は
`docs/` に★実質ゼロ（proposals 2 件が周辺に触れるのみ）∴ 追随不要。

**merge します。** ⚠️ **#4380 は★閉じません** —— **`turn_cancelled` の
「これまでライブ表示ゼロ」が示すとおり、★他にも同じ性質のギャップが残り得る**ため
（197 kind 中 出るのは 20 件になっただけ）。

PR #4402, opened 2026-08-12
fix(#4375): retired-key gate covers removed (not just renamed) keys + .example comment lines

tya5:
**[e2e-coder]** — implements lead-coder's #4375 ruling on architect's
measurement (population confirmed 17, not estimated).

## Population (architect's measurement, not re-derived here)

Docs-derived candidates (532) were nearly all false positives (different
schemas sharing a field name). Correct population: a union of every
top-level key across ~148 historical revisions of the config schema's own
module (`git log --follow`) — 58 that ever existed, 33 current, 25 gone.
8 are already-registered renames (`_RENAMED_CONFIG_KEYS`, complete, no
gaps); 17 are genuine removals with no successor. #4373's own two
hand-found stale keys (`shell_allowed`, `skill_search`) are both in that
17 — the witness that the population is real.

## Three rulings, all implemented

**① 17 removed keys get their OWN type, not folded into `_RENAMED_CONFIG_KEYS`.**
New `RemovedKeyHint` (no `destination` field at all — unrepresentable-wrong,
not just undocumented) + `_REMOVED_CONFIG_KEYS`. `unknown_config_keys()`
checks both registries. `reyn config migrate` reports removed keys in
their own "REMOVED ... delete them" section — never auto-rewritten (no
destination) and never auto-deleted (this command only ever writes a
rewrite TO a key). `reyn config validate` needed no change — it already
only reads `hint.note` generically.

**② `.example` comment-line scanning, scoped to `*.example` only.**
Real witness: #4392, an operator who uncommented a stale
`api_base:`/`models:` block and hit the exact crash this closes.
`docs/**` stays bare-only (a prose rename explanation must not itself
trip the gate).

Caught during development, not assumed: a naive `#` + unbounded
whitespace pattern re-introduced the exact false positive column-0
anchoring exists to prevent — `reyn.local.yaml.example` marks a NESTED
commented key with extra leading spaces (`# llm:` then `#   model:` for
`llm.model`). Narrowed to an optional SINGLE space only after running the
pattern against the real file and reading what it matched.

**③ File-level allowlist stays mandatory** (already existed). Extended
for one real collision the new population surfaced:
`docs/reference/cli/agent.{md,ja.md}`'s `reyn agent show` sample output
uses `workspace:` as its own field (`name`/`created_at`/`workspace`/`role`),
not a `reyn.yaml` shape. Confirmed by reading the whole file for other
collisions (none) before adding.

## Consumer sweep caught a real regression

[...]

PR #4358, opened 2026-08-12
fix(#4349): remove the built-in model catalog, resolve() distinguishes class/name position

tya5:
**[tui-coder]** — #4349: remove the built-in model catalog, `resolve()` distinguishes class/name position.

🔴 **Do NOT merge.** Owner has 3 open decisions on #4349 (recorded there by lead-coder) that this PR does not yet incorporate — see "Not landing in this PR" below. This PR is ready for review/discussion, not landing.

## ① The error fix (required first, per architect's ordering)

`resolve(name)` used to treat every unresolved name identically: log a warning, passthrough as a literal LiteLLM string. That's correct for a genuine raw string (`--model openai/gpt-4o`) but wrong for an unresolved **class** reference (a typo, or a config whose `llm.models:` section failed to load) — the two were indistinguishable until litellm itself rejected the bare name much later with a confusing provider-side error (#3368).

Fix: the same `"/"` disambiguation this module already uses for str-form namespace values (`_resolve_entry`), applied at the entry point. Contains `/` → name position, passthrough unchanged. No `/` and not a known class → class position that failed to resolve → raises immediately, naming the value and the known classes.

`_warn_on_partial_tier_declaration` (#3374's warning) is removed with it — there's no more built-in fallback for an omitted tier to silently resolve through, so its own premise is gone.

## ② Catalog removal

Owner ruling (2026-08-12, option c): the 3 tier **names** (`light`/`standard`/`strong`) and their cost **order** stay as a plain literal (`STANDARD_CLASSES`) in `model_resolver.py` — reyn's own vocabulary. The 9 concrete provider/model targets, and what each tier resolves to, leave the code entirely; a project's own `llm.models:` mapping is the only source now.

- `builtin_models.py` — deleted.
- `test_builtin_models.py`, `test_litellm_gemini_prefix_1162.py` — deleted (their whole subject is gone).
- `test_model_resolver.py`, `test_model_resolver_extends.py` — catalog-dependent tests rewritten against a **synthetic** `builtin=` entry per owner's D ruling (the precedence mechanism survives; only the distribution catalog is gone, so the tests keep proving precedence rather than disappearing).
- `tests/_support/session.py`, `agent_session.py` — added a shared `TEST_MODEL_RESOLVER` default (Session's own default, `resolver or ModelResolver({})`, is genuinely empty now) covering ~300 call sites without touching each one.
[...]

PR #4430, opened 2026-08-12
fix(#4357): the CUI names unmapped config keys instead of a bare count

tya5:
**[docs-maintainer]** — #4357: the CUI names unmapped config keys, instead of only a count.

## Measurement, before any implementation (lead-coder's 3-point brief)

**① Population re-measured, not inherited from filing time.** Repo-internal population is now **0** — root `reyn.yaml` + 3 dogfood fixtures already fixed by #4358 (confirmed by reading the current files); `tests/mcp/test_cli_pipe.py`'s cited sites already correctly nested under `llm:` (grepped every `"model"`/`"models"` key in the file). What's left is the *class* — an operator can never see WHICH key was dropped — not the specific instances the issue was filed against.

**② Warning path traced.** `loader.py`'s `_warn_unknown_config_keys()` calls `config_schema.unknown_config_keys()`, which already returns per-key `RenamedKeyHint`(destination+note)/`RemovedKeyHint`(note) detail (#4375/#4402 built it). **The gap was never the data — only how far it traveled**: one `log.warning` call (file-only, per `_setup_interactive_logging`'s pre-config-load redirect) and a bare `int` count attached to `ReynConfig` for the CUI chrome to render as a number with no key names.

**③ Relationship with #4402 checked by reading its actual PR body.** Different axis, not overlapping — #4402's gate is a static grep/AST scan over `docs/**` and `*.example` files (documentation/examples shouldn't teach stale key names), never touching the runtime loader path. This issue is about a real `config.yaml`/`reyn.yaml`'s unmapped key at *load time* — same underlying registries, different surface.

## A standing ruling the brief didn't flag, found while reading `loader.py` directly

`_warn_unknown_config_keys()`'s own docstring names an owner ruling: *"warn, never hard-fail, anywhere — including sandbox.policy (no special case)"* (accumulated across #4174's spec revisions). The issue's proposed raise-ification directly conflicts with this, and no overriding owner comment is present in the issue thread. **Confirmed with lead-coder before implementing**: raise is treated as NOT approved (a separate issue if someone wants to ask owner to override the standing ruling); this PR's scope is surfacing only, which is compatible with the ruling in either direction — still warn, just with more information.

## The fix

- `_warn_unknown_config_keys()` returns the full `{dotted_key: hint}` dict instead of a bare count. Log-file warning unchanged; callers needing only the count use `len(...)`.
- `ReynConfig` gains `unknown_config_keys: dict` (`schema_internal`, same reasoning as the existing `unknown_config_key_count`) **alongside** the count, not instead of it — every existing reader is unaffected.
- `chrome.py`'s `config_warning_text()` gains an optional `keys=` param: up to 3 key names inline (renamed keys with a known destination first — `` `model` → `llm.model` `` — then remaining unknown/removed keys), "+N more" past the cap. `keys=None` (the pre-#4357 shape) renders **byte-identical** to the original count-only line. Still bounded (never "show everything" — #4380's own 197-item rejection applies here too) and still ends by pointing at `reyn config validate`, so architect's #4194 ruling's 3 properties (doesn't scroll away / stays visible / points to validate) are unchanged — only the line's own content grew.
- `status.py`/`read_model.py` updated for parity with the existing count field.

## Test plan

- [x] 6 new tests in `tests/interfaces/test_config_warning_chrome_4194.py`: renamed-key-with-destination shown / accept-side `keys=None` and `keys={}` fall back byte-identical / inline cap + "+N more" / removed-key shown without a destination arrow / `ReynConfig.unknown_config_keys` defaults to `{}` / end-to-end `load_config()` with a real unmapped key.
- [x] Falsified: stashed all 6 source files together, confirmed the 6 new tests fail (the 6 pre-existing tests in the same file pass unaffected — the back-compat fallback genuinely works), restored, confirmed all 12 green.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (211/215 baselined), `python scripts/test_tier_audit.py --strict` (0 Tier-4, 12/12), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/config/` (full), `tests/interfaces/` (full directory) — 1641 + 172 passed, 7 skipped (unrelated missing extras), 0 failed.

## Issue title

Updated — "5 files, several months" was resolved by #4358 before this PR landed; what remains is the class.

## Arc note

`part of #4357`.

### DIFF
--- a/src/reyn/config/loader.py
+++ b/src/reyn/config/loader.py
@@ -164,7 +164,9 @@ def build_policy_tier_config(cwd: Path | None = None) -> dict:
     return merged
 
 
-def _warn_unknown_config_keys(policy_tier_merged: dict) -> int:
+def _warn_unknown_config_keys(
+    policy_tier_merged: dict,
+) -> "dict[str, Any]":
     """#4174 T0: log ONE warning naming every unknown/renamed config key
     found in the policy-tier config (``user_global`` + ``project`` +
     ``project_local`` merged — the operator-editable files; the 5
@@ -184,17 +186,26 @@ def _warn_unknown_config_keys(policy_tier_merged: dict) -> int:
     LOOSER, not silently inert like an ordinary dropped key, so an
     operator relying on it must see what's actually in force.
 
-    Returns the unknown-key count (#4194: the log-only warning is invisible
-    in the interactive CUI — ``_setup_interactive_logging`` redirects all
-    logs to a file, per architect's live measurement — so the count is
-    also returned for :func:`load_config` to attach to the ``ReynConfig``
-    it builds, giving the CUI's bottom chrome something to read).
+    Returns the full ``{dotted_key: hint_or_None}`` dict — #4357: this
+    used to return only the bare COUNT (#4194: the log-only warning is
+    invisible in the interactive CUI — ``_setup_interactive_logging``
+    redirects all logs to a file, per architect's live measurement — so
+    the count was returned for :func:`load_config` to attach to the
+    ``ReynConfig`` it builds). The count alone gave the CUI's bottom
+    chrome something to show, but not WHICH keys — an operator reading
+    "3 config keys not applied" cannot act on it without separately
+    running ``reyn config validate``, and #4357 measured that in practice
+    nobody did (5 real instances of a moved key going unfixed for months,
+    including this repo's own ``reyn.yaml``). ``unknown_config_keys()``
+    already carries per-key ``RenamedKeyHint``/``RemovedKeyHint`` detail
+    (#4375/#4402) — the data was never the gap, only how far it traveled.
+    Callers that only need the count use ``len(...)`` on the return value.
     """
     from reyn.config import config_schema
 
     unknown = config_schema.unknown_config_keys(policy_tier_merged)
     if not unknown:
-        return 0
+        return {}
 
     import logging
     log = logging.getLogger(__name__)
@@ -222,7 +233,7 @@ def _warn_unknown_config_keys(policy_tier_merged: dict) -> int:
         )
 
     log.warning("Unrecognized config key(s) found:\n" + "\n".join(f"  - {line}" for line in lines))
-    return len(unknown)
+    return unknown
 
 
 def _as_config_dict(val: object, key: str) -> dict:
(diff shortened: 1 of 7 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/interfaces/test_config_warning_chrome_4194.py` の差分をコード側で読了）

```
Q1 Tier ── ★2 / 2b（宣言どおり）。外部の contract: #4357 ＋ 恒久裁定
           「warn, never hard-fail, anywhere」（loader.py の docstring）。
Q2 転記 ── ✅ 無し。`config_warning_text(2) == "⚠ 2 config keys not applied → …"` は
           ★値の assert。
Q3 誰が困るか ── ★実在（CUI の利用者。裸の件数しか出ていなかった）。
           🔵 accept 側が ★3 本 在る:
             ・keys=None / keys={} で ★旧形（件数のみ）に戻る（呼び出し側の互換）
             ・RemovedKeyHint は ★行き先を出さない（出す情報が無いものを捏造しない）
             ・ReynConfig.unknown_config_keys の ★既定が {}
Q4 走らずに緑か ── ✅ skip 無し。CI 緑。
Q5 累積 ── ✅ `+2 more` の cap で ★行の長さが 母集団に依存しない。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

🔵 **owner 恒久指示（本日）「★根拠のない数値をユーザ変更の口もなく埋め込むのはやめて」に
★適合しています** —— 新規定数 `_CONFIG_WARNING_INLINE_KEY_CAP = 3` に
**「何を守る値か」「なぜ有界にするか」（#4380 の 197 件却下を参照）が★コメントで書かれている**
∴ **「根拠を書く / config の口を付ける」の★前者を満たしています。**
⚪ **口が無いことは違反ではありません**（どちらか一方で足ります）。

⚪ **doc drift**: 恒久裁定「warn, never hard-fail」を★変えていない
（raise 化していない）∴ loader.py の docstring と整合。

**merge します。** ⚠️ **#4357 は★閉じません** —— **owner 要求の raise 化は
恒久裁定と衝突するため範囲外**にしており、**題（「5 箇所が何ヶ月も効いていなかった」）も
実態とずれたまま**です（母集団は既に 0 件）。

PR #4435, opened 2026-08-12
fix(#4431): read_file/edit_file not_found suggestions signal their own cap

tya5:
**[docs-maintainer]** — #4431 item ③, PR 1 of 4: `read_file`/`edit_file` not_found suggestions signal their own cap.

## Scope (per lead-coder's brief)

One of the 10 flagged silent-and-no-config-knob sites: `core/op_runtime/file.py`'s `_NOT_FOUND_SUGGESTIONS_LIMIT` (8). A directory with more than 8 siblings always showed exactly 8 suggestions, with no sign more existed — indistinguishable from a directory that genuinely only had 8. Owner ruling (#4381-motivated): a silent cap with no config knob needs the loss to be visible instead.

## The fix

- `_nearby_files` now uses `Workspace.glob_files_with_total` (the existing #2998 mechanism `glob_entries`/the `glob` op already rely on for this exact problem) instead of the silently-capped `glob_files`, returning `(suggestions, total_before_cap)`.
- All 3 not_found result builders (`read_file`, `edit_file`, the binary-image path) go through a new `_suggestions_fields` helper so the `suggestions_truncated`/`suggestions_total` fields are written once, not re-derived at each call site with room to drift.
- **Absence convention, not present-and-False**: `suggestions_truncated` only appears when the cap actually cut something — a directory with ≤8 siblings gets no such key at all, matching `op_runtime/file.py`'s own established pattern for this same "was this loss real" question elsewhere in the file.
- No config knob needed — 8 already has a rationale comment (mirrors `invoke_action`'s own ~8-suggestion shape for "did you mean" narration); only the visibility half was missing.

## Test plan

- [x] 2 new tests in `tests/core/test_op_runtime_file_not_found_suggestions.py`: the truncation-signal case (20 siblings → `suggestions_truncated=True`, `suggestions_total=20`) and its accept-side twin (2 siblings → both keys absent). 11 pre-existing tests untouched and still pass (the shape change is additive).
- [x] Falsified: stashed `file.py`, confirmed the 2 new tests go RED (`KeyError`) while the 11 pre-existing tests stay GREEN; restored, confirmed all 13 GREEN.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (211/215 baselined, unchanged), `python scripts/test_tier_audit.py --strict` (0 Tier-4, this file's tests all OK), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/core/` (full, 2534 passed / 1 skipped, 0 failures) — 1 unrelated pre-existing failure exists elsewhere in `tests/runtime/` (`test_session_pure_extraction_3679_s1.py`, a stale `reyn.runtime.session_pure` import reference), confirmed via `git stash` to pre-date this diff entirely; reported separately, not in scope here.

## Sibling PRs (same #4431 item ③, independent call paths — not stacked)

- `web.py` `_HtmlPreviewParser` outline-truncation visibility
- `reyn_repo.py` glob/grep visibility (3 constants, 1 PR per lead-coder's own grouping — same call path)
- `knowledge_ingest.py` repo-ingest skip visibility (new `repo_ingest_files_skipped` audit-event kind)

## Arc note

`part of #4431`.

### DIFF
--- a/src/reyn/core/op_runtime/file.py
+++ b/src/reyn/core/op_runtime/file.py
@@ -88,7 +88,7 @@ async def _read_image_file(op: FileIROp, ctx: OpContext, *, mime_type: str) -> d
             "kind": "file", "op": "read", "path": op.path,
             "status": "not_found",
             "error": f"file not found: {op.path}",
-            "suggestions": _nearby_files(ctx.workspace, op.path),
+            **_suggestions_fields(ctx.workspace, op.path),
             "content": "",
         }
 
@@ -167,12 +167,18 @@ def _nearest_existing_ancestor(ws, start: str) -> "str | None":
     return None
 
 
-def _nearby_files(ws, path: str, *, max_results: int = _NOT_FOUND_SUGGESTIONS_LIMIT) -> list[str]:
+def _nearby_files(
+    ws, path: str, *, max_results: int = _NOT_FOUND_SUGGESTIONS_LIMIT
+) -> "tuple[list[str], int]":
     """List sibling files under the parent of *path*, for use as not_found suggestions.
 
-    Returns project-relative paths from ``Workspace.glob_files`` when the
-    parent directory exists (possibly empty — "no neighbours" legitimately
-    yields ``[]``).
+    Returns ``(suggestions, total_before_cap)`` — project-relative paths from
+    ``Workspace.glob_files_with_total`` when the parent directory exists
+    (possibly empty — "no neighbours" legitimately yields ``([], 0)``).
+    #4431: was the plain (silently-capped) ``glob_files`` — a directory with
+    more than ``_NOT_FOUND_SUGGESTIONS_LIMIT`` siblings always showed exactly
+    8 with no sign more existed. ``_with_total`` is the same #2998 fix the
+    ``glob`` op already uses; the caller reports the cap the same way.
 
     #3629: when the parent directory itself does NOT exist — a rename, a
     move, or a plugin reinstall to a different location, exactly the case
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/core/test_op_runtime_file_not_found_suggestions.py` をコード側で読了）

```
Q1 Tier ── ★2。外部の contract: #4431 ＋ owner 恒久指示（根拠 or 口、どちらか）。
Q2 転記 ── ✅ 無し。`suggestions_total == 20` は ★setup で作った件数（20 兄弟）。
Q3 誰が困るか ── ★実在。8 件を超えると ★常に ちょうど 8 が出て 「他に在る」の合図が無かった。
           🔵 accept 側の twin が在る（cap 以下では ★フィールド自体が 出ない）
             ── `not in result` で assert ∴ ★過剰発火も捕まえる。
Q4 走らずに緑か ── ✅ skip 無し。
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

🔵 **既存の仕組みを使い回しているのが良い** —— **`glob_files_with_total` は
★#2998 で glob op が既に採った同じ fix** で、**新しい形を発明していません。**
⚪ 退避挙動も保存（OS エラーは `([], 0)` に degrade、raise しない）。

**merge します。** ⚠️ **#4431 は★閉じません**（残り 8 箇所 ＋ ①②の単位統一）。

PR #4437, opened 2026-08-12
fix(#4431): reyn_repo_glob/reyn_repo_grep signal every silent cap they hit

tya5:
**[docs-maintainer]** — #4431 item ③, PR 3 of 4: `reyn_repo_glob`/`reyn_repo_grep` signal every silent cap they hit.

## Scope (per lead-coder's brief)

Three named constants, all in `runtime/reyn_repo.py`, grouped into one PR per lead-coder's own instruction ("`reyn_repo.py` の3件は同じ経路なので1本"):

- `_MAX_READ_BYTES` (256KB) — checked first. **Already compliant**: rationale comment ("Reyn's docs/source files are well under this... ~256KB ≈ 50K tokens worst case") AND a loud structured error naming the exact byte count + how to work around it (`offset`/`limit` slicing) when `read_text` hits it. No fix needed there.
- `_MAX_GLOB_MATCHES` (200) — `glob_entries` capped silently with no `truncated` field at all, unlike `grep_entries` in the same file which already had one.
- `_MAX_GREP_RESULTS` / `_GREP_SNIPPET_CHARS` (50 / 200) — lead-coder flagged both as lacking a rationale comment. `_MAX_GREP_RESULTS`'s match-count cap was already visible (`truncated`); `_GREP_SNIPPET_CHARS`'s per-match line truncation was NOT (no marker at all on a cut snippet).
- A related discovery while reading the call site: `grep_entries`'s per-file size skip (`fp.stat().st_size > _MAX_READ_BYTES: continue`) is a SEPARATE silent cut from all of the above — a file over the size cap read identically to "searched, no matches found", with zero signal it was excluded from the scan at all.

## The fix

- `glob_entries` now returns `truncated: bool`, mirroring `grep_entries`'s existing contract.
- `grep_entries` now returns `skipped_large_file_count: int` (always present, 0 when nothing was skipped — matches this function's pre-existing convention of returning every field unconditionally, e.g. `truncated`).
- Each grep match now carries `snippet_truncated: True` when its line was cut to `_GREP_SNIPPET_CHARS` (absent when the line fit — the newer, greenfield convention this PR's sibling `file.py`/`web.py` PRs also use).
- Rationale comment added for `_GREP_SNIPPET_CHARS`: a per-match display choice, not a correctness bound (the full line is always recoverable via the `path`/`line` the match already names), so no config knob is needed per the owner's either/or ruling.

## Test plan

[...]

PR #4436, opened 2026-08-12
fix(#4431): web_fetch HTML outline preview signals its own truncation cap

tya5:
**[docs-maintainer]** — #4431 item ③, PR 2 of 4: web_fetch HTML outline preview signals its own truncation cap.

## Scope (per lead-coder's brief)

`core/op_runtime/web.py`'s `_HtmlPreviewParser._OUTLINE_MAX` (8) — one of the 10 flagged sites, and flagged with an extra note: no rationale comment for the value 8 either. A page with more than 8 headings always showed exactly 8 in the outline, with no sign more existed.

## The fix

- Added a rationale comment: 8 has no measured basis beyond "a preview should stay short" — a display choice, not a correctness bound, since the caller (`_generate_web_fetch_preview`) already has the full `extracted_text` alongside the outline preview. Per the owner's ruling, that's why this constant does NOT need a config knob (unlike a value that bounds real data loss).
- `_HtmlPreviewParser` now tracks `_heading_count` (every heading tag seen, uncapped) alongside the capped `_outline` list; `result()` reports `outline_truncated` + `outline_heading_count` only when the cap actually cut something — absent, not present-and-False, otherwise (matches the sibling PR's `file.py` convention for the same question).

## Test plan

- [x] New file `tests/core/test_web_fetch_outline_truncation_visibility_4431.py`: past-the-cap case, accept-side twin (under cap → no flag), and the exact-boundary case (headings == cap is not a cut). All reference the real `_OUTLINE_MAX` constant rather than a hardcoded literal.
- [x] Falsified: stashed `web.py`, confirmed the past-the-cap test goes RED (`KeyError: 'outline_truncated'`); restored, confirmed all 3 GREEN.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (unchanged), `python scripts/test_tier_audit.py --strict` (0 Tier-4 — fixed 3 findings during authoring: missing per-test Tier-line prefixes, and a `len(...) == 8` literal replaced with `== _OUTLINE_MAX`), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
[...]

PR #4438, opened 2026-08-12
fix(#4431): repo-knowledge ingest reports files skipped for size

tya5:
**[docs-maintainer]** — #4431 item ③, PR 4 of 4: repo-knowledge ingest reports files skipped for size.

## Scope (per lead-coder's brief)

`data/index/knowledge_ingest.py`'s `_REPO_INGEST_MAX_BYTES` (256KB) — already had a rationale comment (mirrors `reyn_repo_read`'s own `_MAX_READ_BYTES` cap), so no config knob is needed per the owner's either/or ruling. What was missing was visibility: a file over the cap silently dropped out of the `knowledge_repo_doc`/`knowledge_repo_src` background embedding corpus with **zero trail** — an operator whose repo-knowledge search misses a file would see nothing distinguishing "excluded for size" from "was never written".

This is the one site of the 10 that needed a genuinely new observable rather than a field added to an existing result dict — the background-ingest build has no return value an operator ever reads; the only real visibility surface here is the audit-event log this build already threads an `EventLog` through.

## The fix

- `_iter_repo_entries` now returns `(entries, skipped_for_size_count)` instead of just `entries`.
- `_repo_build_fn`'s `_build()` closure emits a new `repo_ingest_files_skipped` audit-event (`kind`, `skipped_count`, `reason="over_size_cap"`) via `op_ctx.events.emit(...)` — **once per build, aggregated**, not once per skipped file (the count is in the payload), and only when nonzero (a routine build with nothing skipped emits nothing — matches the newer absence-means-nothing-happened convention this arc's sibling PRs also use). Best-effort (wrapped in try/except), matching every other `op_ctx.events.emit` call site in `op_runtime`.
- New audit-event kind declared in **both** halves of the #3410 CI-checked pair: `AUDIT_EVENT_KINDS` (`src/reyn/core/events/event_schema.py`) and `docs/reference/runtime/events.md` (both the vocabulary-list enumeration and a payload-table row next to `embedding_index_build_complete`, the closest existing precedent — 3 of that event's own siblings, `embedding_index_build_error`/`_progress`/`_started`, are precedent for a kind existing in the vocabulary list with no dedicated table row, but I added one anyway since this is a genuinely new observable, not an inherited one).

## Test plan

- [x] 2 new tests in `tests/core/test_fp0066_p3b_repo_knowledge_ingest.py`: an oversize file is still excluded from the built chunks (unchanged behaviour) AND emits exactly one `repo_ingest_files_skipped` event with the right `kind`/`skipped_count` (captured via the real `EventLog.add_subscriber` mechanism, `tests/_support/events.py::collect_events` — not a mock); accept-side twin confirms zero events when nothing is oversize.
- [x] Falsified: stashed `knowledge_ingest.py` + `event_schema.py`, confirmed the positive test goes RED (0 events collected, not 1); restored, confirmed all 8 GREEN (6 pre-existing + 2 new).
- [x] `python -m pytest tests/core/test_audit_event_kind_vocabulary_3410.py` — the CI-checked doc↔code pair — 10/10 green (was RED with `missing=['repo_ingest_files_skipped']` before the docs/events.md edit landed, confirming the gate actually catches a skipped doc sync).
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (unchanged), `python scripts/test_tier_audit.py --strict` (0 Tier-4 — used tuple-unpacking `(skip_event,) = (...)` instead of `len(...) == 1` to avoid a format-pinning flag on an exact-one-aggregate-event assertion), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: this file + the vocabulary gate (18 passed); full `tests/core/` also run clean (2534 passed / 1 skipped) as part of the combined pre-split sweep.

## Sibling PRs (same #4431 item ③, independent call paths — not stacked)

- #4435 `file.py` not_found suggestions visibility
- #4436 `web.py` outline preview visibility
- #4437 `reyn_repo.py` glob/grep visibility

## Arc note

`part of #4431`.

### DIFF
--- a/docs/reference/runtime/events.md
+++ b/docs/reference/runtime/events.md
@@ -175,6 +175,7 @@ presentation_installed
 presentation_load_failed
 presented
 project_context_changed
+repo_ingest_files_skipped
 router_context_overflow_detected
 router_context_overflow_unrecovered
 router_empty_response_detected
@@ -277,6 +278,7 @@ See [Concepts: multi-agent](../../concepts/multi-agent/multi-agent.md) — "Agen
 | `llm_called` | `model` (+ `chain_id` when the call belongs to a delegation chain) |
 | `llm_response_received` | `prompt_tokens`, `completion_tokens`, `cached_tokens`, `cache_creation_tokens`, `cost_usd`, `usage_source` (+ `chain_id`) |
 | `embedding_index_build_complete` | `source_id`, `chunk_count`, `total_tokens`, `cost_usd`, `embedding_model` (a disk-adopt/no-fresh-build completion carries `total_tokens`/`cost_usd` as `null` — no embed call happened that run, not a cost of zero) |
+| `repo_ingest_files_skipped` | #4431 — the repo-knowledge (`knowledge_repo_doc`/`knowledge_repo_src`) background build excluded one or more files for exceeding `_REPO_INGEST_MAX_BYTES` (256 KB, `src/reyn/data/index/knowledge_ingest.py`). Emitted once per build, only when the count is nonzero — a routine build with nothing skipped emits no event at all. `kind` (`"doc"`/`"src"`), `skipped_count`, `reason` (currently always `"over_size_cap"`) |
 
 `usage_source` says where the token counts came from: `provider` (the provider
 reported them) or `estimated` (the provider's stream carried no usage, so
--- a/src/reyn/core/events/event_schema.py
+++ b/src/reyn/core/events/event_schema.py
@@ -347,6 +347,7 @@ AUDIT_EVENT_KINDS: frozenset[str] = frozenset({
     "presentation_load_failed",
     "presented",
     "project_context_changed",
+    "repo_ingest_files_skipped",
     "router_context_overflow_detected",
     "router_context_overflow_unrecovered",
     "router_empty_response_detected",
--- a/src/reyn/data/index/knowledge_ingest.py
+++ b/src/reyn/data/index/knowledge_ingest.py
@@ -465,9 +465,10 @@ def _classify_repo_kind(rel_path: str) -> "str | None":
     return "src"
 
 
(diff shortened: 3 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/core/test_fp0066_p3b_repo_knowledge_ingest.py` の差分をコード側で読了。★CI 緑になり次第 merge します）

```
Q1 Tier ── ★2。外部の contract: #4431 ＋ ★audit-event の閉じた語彙（CLAUDE.md）。
Q2 転記 ── ✅ 無し。
Q3 誰が困るか ── ★実在。ingest が size で落としたファイルが ★何の痕跡も残していなかった。
           🔵 accept 側 twin（★oversize が無い時は イベントが 1 件も出ない）
             ── `skip_events == []` ∴ ★過剰発火を捕まえる。
           ⚪ ★成功側も同時に assert（`entry.chunk_count == 3`）
             ── 「skip を報告するようになったが 本体が壊れた」を捕まえる。
Q4 走らずに緑か ── ✅ skip 無し。
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

## 🔵 ★CI がチェックする唯一の doc↔code 対に対して、3 点セットが揃っています

```
新 kind  ★"repo_ingest_files_skipped"
① ★emit      src/reyn/data/index/knowledge_ingest.py
② ★declare   src/reyn/core/events/event_schema.py（AUDIT_EVENT_KINDS）
③ ★document  docs/reference/runtime/events.md
```
⚪ **CLAUDE.md が「emit・declare・document は★1 つの三部作の変更」と書いている形**で、
**`tests/core/test_audit_event_kind_vocabulary_3410.py` が 2 つだけでは赤くなる側**です。
**3 つ揃っているので、この PR は★その gate に対して 正しい形。**

PR #4432, opened 2026-08-12
fix(#4381): next_char_offset actually works, spilled files can't re-spill

tya5:
**[tui-coder]** — fix(#4381): next_char_offset actually works, spilled files can't re-spill

## What (① ② ③ per lead-coder's dispatch)

**① `next_char_offset` wired end-to-end.** `op_runtime/file.py` already computed and returned it (a single line longer than the model's context window) but the router-facing `read_file` schema declared only `path`/`offset`/`limit` — `char_offset` was never in the schema, and the handler never forwarded it into `FileIROp` even though `FileIROp` already supported the field. Fixed at all 3 points: schema, handler, and the truncated result's own hint text (which previously said only "re-read from offset N" — restarting the same oversized line from char 0 and truncating identically, with no textual cue to use `char_offset`).

**② False description fixed.** `limit`'s description claimed "Omit to read through end of file" — false whenever the remainder exceeds the model-derived inline cap. Rewritten to match the "explicit cap + truncated flag, never silent" discipline `grep_files`/`list_directory` already use.

**③ Multi-spill loop closed.** Terminology note (owner-ratified mid-session, 2026-08-12): this mechanism is **spill** (didn't fit, unavoidable, never config-gated) — not **offload** (would fit, goes out anyway for budget, optional/default-false). Existing identifiers (`offload_value`, `MediaStore.save_tool_result`) are unchanged; only this PR's new identifiers/messages use "spill". Same-PR doc update: `dogfood-discipline.md` now states spill is a family (step result spill, existing; tool result spill, this PR) rather than exclusively the plan-step mechanism.

The actual loop: `op_runtime`'s own read cap (window/char-derived) and the router's separate spill trigger (token/config-derived) are independent — a bare re-read of a spilled file can come back truncated under the first cap and STILL exceed the second downstream, spilling again. Traced via the real call sites before writing code. Fixed via `MediaStore.is_tool_result_spill(path)` — an in-memory, session-scoped registry of paths `save_tool_result` itself actually wrote (not a path-naming/location heuristic, which would also catch an unrelated file an operator placed in the same directory). The read handler errors with the exact remedy ("specify `char_offset` and re-read") instead of truncating into the loop.

## Related, explicitly out of scope

[...]

PR #4441, opened 2026-08-12
fix(#4431): a truncated load_skill result can actually be resumed

tya5:
**[docs-maintainer]** — #4431 new item (architect correction): a truncated `load_skill` result can actually be resumed.

## Scope

Separate axis from #4431 item ③ (say a cut happened — `load_skill` already did, via `status="truncated"`/`_truncated`/`note`). This is the OUTPUT-side gap: the note pointed at "the full body is on disk at `<path>`" with **no resume position at all** — 0 fields, not even a half-working one.

Per lead-coder's explicit instruction, checked the OUTPUT side directly rather than assuming "#4432's read_file pattern, same shape": `read_file` (pre-#4432) had computed `next_char_offset` but the schema/handler never forwarded it — "returned but couldn't receive." `load_skill` was a different failure: it returned **nothing** to resume from.

## The fix

`load_skill` stays deliberately non-paginated (module docstring: "load the whole thing, not a general paginated reader") — so the fix isn't new `offset`/`limit` params on `load_skill` itself. It's making the note's own promised escape hatch (`read_file` on the same path — every provenance class `load_skill` supports, `load_skill` and `file.py` already share the same builtin/plugin bypass helpers) actually usable:

- Truncation now accumulates **whole lines** (mirroring `op_runtime/file.py`'s own #1209/#2335 algorithm), replacing the old bare `content[:cap]` char slice that could cut mid-line.
- Returns `next_offset` — the exact 0-indexed line number `read_file`'s existing `offset` param needs. Reused that field name rather than inventing a new one (no new shape, per lead-coder's own instruction).
- A single oversized line (content[:cap] edge case) needs no new field either — resuming via `read_file(offset=0)` on that same line hits `read_file`'s own #2335 char-truncation and returns ITS `next_char_offset`, composing through existing machinery.
- Note text now says exactly what to call: `read_file(path=<path>, offset=<next_offset>)`.

## Test plan

- [x] 3 new tests in `tests/core/test_op_runtime_load_skill_3247.py`: `next_offset` present + in the note text (over-cap case); **end-to-end resume** — reading the SAME path via `read_file(offset=next_offset)` recovers the exact remainder, concatenation equals the original body byte-for-byte (not just "a field is present" — a plausible-looking but off-by-one `next_offset` would pass a bare presence check and fail this one); accept-side twin (under cap → unchanged `status="ok"`, no `next_offset`/`_truncated`).
- [x] Falsified twice (once per test-body-size iteration, since the first body was too large and made the RESUME read ALSO truncate — a real bug in my test, not the fix, caught and fixed before landing): stashed `load_skill.py`, confirmed both new positive tests RED (`KeyError: 'next_offset'`); restored, confirmed all 16 GREEN.
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (unchanged), `python scripts/test_tier_audit.py --strict` (0 Tier-4 — renamed one test to drop a literal `byte_identical` in its name, a bounded-life-scaffold keyword false-triggering on a permanent behavioral-contract test), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: this file + `test_read_bounding_structural_signal_1209.py` + `tests/plugins/test_skill_load.py` (36 passed); full `tests/core/` also run clean (2544 passed / 1 skipped, 0 failures).

## Arc note

`part of #4431`.

### DIFF
--- a/src/reyn/core/op_runtime/load_skill.py
+++ b/src/reyn/core/op_runtime/load_skill.py
@@ -258,10 +258,45 @@ async def handle(op: LoadSkillIROp, ctx: OpContext) -> dict:
         history_fields["skill_source_path"] = op.path
 
     if len(content) > cap:
-        truncated_content = content[:cap]
+        # #4431 (architect correction): was a bare `content[:cap]` char slice
+        # with NO resume position returned at all — the note pointed at "the
+        # full body is on disk" but gave no way to continue from where this
+        # read stopped. load_skill is deliberately NOT a paginated reader
+        # (module docstring) — the fix is not to add offset/limit here, it's
+        # to make the escape hatch this note already promises actually work:
+        # `read_file` (same file, since `op.path` is a real on-disk path for
+        # every provenance class this module and file.py share the same
+        # builtin/plugin bypass for) already supports a 0-indexed LINE
+        # `offset` (#4381/#4432 fixed its own char_offset-within-a-line
+        # sibling). Whole-line accumulation (mirroring op_runtime/file.py's
+        # own #1209/#2335 algorithm) so `next_offset` names a genuine line
+        # boundary `read_file(path=op.path, offset=next_offset)` can resume
+        # from without re-showing or dropping a partial line.
+        all_lines = content.splitlines(keepends=True)
+        shown: list[str] = []
+        acc = 0
+        next_offset = 0
+        for i, line in enumerate(all_lines):
+            if shown and acc + len(line) > cap:
+                next_offset = i
+                break
+            if not shown and len(line) > cap:
+                # A single line alone exceeds the cap (e.g. a minified/
+                # one-line body) — char-truncate it so `content` stays
+                # genuinely bounded; `read_file(offset=0)` on resume hits
+                # this exact same line and returns ITS OWN `next_char_offset`
+                # (file.py's #2335 mechanism), so no new field is needed here.
+                shown.append(line[:cap])
+                acc += cap
+                next_offset = i
+                break
+            shown.append(line)
+            acc += len(line)
+            next_offset = i + 1
+        truncated_content = "".join(shown)
         ctx.events.emit(
             "tool_executed", op="load_skill", path=op.path,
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/core/test_op_runtime_load_skill_3247.py` の差分をコード側で読了。★CI 緑になり次第 merge）

```
Q1 Tier ── ★2。外部の contract: #4431 の新項目（load_skill の打ち切りに 続きを読む口が無い）。
Q2 転記 ── ✅ 無し。⚪ `first["content"] + rest["content"] == big` は ★round-trip の主張
           （1 バイトも失わず 重複もしない）── 実装式の転記ではない。
Q3 誰が困るか ── ★実在。cap で切られた skill body に ★続きを読む手段が 無かった。
           🔵 ★end-to-end が本体: note が案内する処方（read_file で再開）が
             ★実際に動くことを 通して確認している ── 「案内はあるが動かない」を捕まえる。
           ⚪ accept 側（cap 未満は 影響なし）も在る。
Q4 走らずに緑か ── ✅ skip 無し。
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

## 🔵 設計が★最小です（新機構ゼロ）

```
❌ load_skill に ★char_offset の入力口を足す（read_file と同じ形を もう 1 つ作る）
✅ ★next_offset を返し、note が ★read_file を 処方として名指しする
   → ★既に在る口（read_file の offset）を使わせる ∴ ★口が 2 つに増えない
```
⚪ **私は brief で「返していないなら返す ＋ 受け取る口」と書きましたが、
★受け取る口は 既に在った（read_file）** —— **実装者が★より小さい形を見つけています。**

⚪ **`#4437 の「count と content の両方を assert する」規律を明示的に引いている**のも良い
—— **本日の他 PR の作法が★横に伝わっている形。**

PR #4453, opened 2026-08-12
fix(#4381): PR-2 stage 3 — in-flight untrusted-taint latch

tya5:
**[docs-maintainer]** — #4381 PR-2 stage ③: in-flight untrusted-taint latch.

## Design (architect, corrected — posted after a clarifying question surfaced a flaw in the first draft)

The resource this stage touches is **not** "no witness exists" — `meta` already carries `external_source` at `router_loop.py`'s tool-result production, a persisted **projection** of the taint, not its origin. The real problem is **timing**: `_ephemeral_contextual_for_turn` (the per-iteration re-narrow check, #1909/#3501) reads `self.history`, which only reflects what has already been **committed**. A later PR (PR-2 stage 1, not yet implemented) will defer that commit to after the turn's model response, opening a same-turn window where a tool result's taint is real but invisible to the history scan — closing that window is the whole point of this stage, landed **ahead of** the deferral per the owner's own order-safety requirement (confirmed and adopted after I flagged the risk of landing them together).

## The fix

- `Session._in_flight_untrusted_this_turn: bool` — a single in-flight latch, reset at each turn's start (`_run_router_loop`), never cleared mid-turn (same "narrows, never un-narrows before the turn boundary" property `_untrusted_latched` already has for the per-iteration rung).
- Set at the **same line** `router_loop.py` stamps `external_source` onto the persisted meta — no second, independently-maintained signal to drift out of sync.
- Wired through `RouterHostAdapter.mark_untrusted_in_flight` — the exact shape the existing `peek_mid_turn_injection`/`commit_mid_turn_injection` pair already uses (bare param, `None`-default, `getattr`-guarded call site). Passes the measured consumer-set param gate (#3482, `tests/runtime/test_router_host_adapter_param_gate_3482.py`) unchanged.
- `_ephemeral_contextual_for_turn`'s gate: `metas_have_untrusted(...) or self._in_flight_untrusted_this_turn` — history scan OR the flag, per architect's own predicate.

## Test plan — 2 of the 3 adversarial angles the design calls for

1. **Positive control**: with the tainted entry's history commit simulated as not-yet-landed (the exact precondition PR-2 stage 1 will introduce — constructed by patching the *already-captured* `RouterHostAdapter._append_history_cb`, since `Session._append_history` was bound once at construction and a later instance-attribute patch would not be observed by that callback), the same-turn next dispatch is still denied — the flag alone, zero history support, is sufficient.
2. **Falsify**: the identical setup with the flag-setting callback neutralized — the denial does **not** happen, proving #1 genuinely depends on the flag and isn't vacuous.
3. **Order-falsify (architect's 3rd angle) — explicitly NOT written.** The runtime race it targets (clearing the flag before the deferred commit lands) has no reachable code path today, since PR-2 stage 1 (the deferral) doesn't exist yet — a test against it would either not compile against real functions or vacuously pass against a hand-built stand-in. Documented in the test file itself as the acceptance-condition material PR-2 stage 1's own brief needs, per the owner's explicit instruction to state the reason rather than force the test.

## Verification

- [x] Falsified: stashed all 3 source files, confirmed `test_flag_denies_when_history_has_not_landed_yet` RED (a genuinely different failure — without the fix `remember_shared` isn't excluded at all, it fails later on an unrelated write-permission path, confirming the flag mechanism is what makes exclusion happen); restored, confirmed both new tests GREEN.
- [x] Sibling suite unaffected: `test_1909_intra_turn_opt_in_narrowing.py` / `test_1909_external_tool_result_narrowing.py` / `test_3501_untrusted_narrowing_opt_in.py` / `test_3378_advertise_enforce_agreement.py` / `test_3380_tool_tab_ephemeral_narrowing.py` — 42/42 green (today's immediate-commit behavior means the flag is always redundant-but-harmless alongside an already-correct history scan; this is the expected no-observable-change sanity check).
- [x] `ruff check`, `python scripts/mypy_ratchet.py` (unchanged, 211/215 baselined), `python scripts/test_tier_audit.py --strict` (0 Tier-4), `python scripts/verify_module_docstrings.py`, `python scripts/check_tests_path_literal_reference.py` — all clean.
- [x] Broad sweep: `tests/llm/` + `tests/runtime/` (2415 passed / 1 skipped) and `tests/security/` + `tests/services/` (855 passed / 26 skipped) — 0 failures.

**tests/ touched — TESTS-READ wait, no self-merge.**

## Arc note

`part of #4381` — stage 1 (defer the commit itself, the "①" original scope) needs its own brief once the exact mechanics (does every `role="tool"` append defer, does `role="assistant"`'s own `tool_calls` message, how `build_history()` depends on the turn's in-progress state) are nailed down; this stage closing first is what makes that later stage's own order-falsify test (item 3 above) writable at all.

### DIFF
--- a/src/reyn/runtime/router_loop.py
+++ b/src/reyn/runtime/router_loop.py
@@ -3614,6 +3614,18 @@ class RouterLoop:
             }
             if external_source:
                 _tool_meta["external_source"] = True
+                # #4381 PR-2 stage ③: the SAME update point as the meta stamp
+                # above — not a second, independently-set signal. The meta
+                # stamp only becomes visible to `_effective_contextual_for_turn`
+                # once THIS history entry is persisted (`_append_entry` below);
+                # this in-flight flag closes the gap for the CURRENT turn's
+                # own remaining iterations, which see it immediately. Getattr-
+                # guarded: a host that never wired the callback (most test
+                # construction, phase hosts) behaves as a no-op, same
+                # convention as every other optional host method.
+                _mark_untrusted = getattr(host, "mark_untrusted_in_flight", None)
+                if _mark_untrusted is not None:
+                    _mark_untrusted()
             # #73 (Option C, co-vet-directed): stamp the ALREADY-KNOWN success/failure
             # classification as a typed field — never re-derived downstream from the
             # rendered ``content_str`` (a display string is not a stable data contract;
--- a/src/reyn/runtime/services/router_host_adapter.py
+++ b/src/reyn/runtime/services/router_host_adapter.py
@@ -222,6 +222,10 @@ class RouterHostAdapter:
         #3792. Async callback ``(msg_id: str) -> None`` —
         ``Session.commit_mid_turn_injection``. Same ``None``-default contract
         as ``peek_mid_turn_injection``.
+    mark_untrusted_in_flight:
+        #4381 PR-2 stage ③. Sync callback ``() -> None`` —
+        ``Session._mark_untrusted_in_flight``. Same ``None``-default contract
+        as ``peek_mid_turn_injection``.
     """
 
     # RouterLoopHost Protocol attributes (non-property)
@@ -301,6 +305,13 @@ class RouterHostAdapter:
         # exactly like a phase host (no-op seam).
         peek_mid_turn_injection: "Callable[[], Awaitable[dict | None]] | None" = None,
         commit_mid_turn_injection: "Callable[[str], Awaitable[None]] | None" = None,
+        # #4381 PR-2 stage ③: sync callback () -> None — Session.
+        # _mark_untrusted_in_flight. Bare, no shared-consumer partner (mirrors
+        # peek/commit_mid_turn_injection's own bare-ness above). None-default
+        # so pre-existing hand-built adapters (tests, other call sites) stay
+        # valid; router_loop.py getattr-guards the call, same as every other
+        # optional host method.
+        mark_untrusted_in_flight: "Callable[[], None] | None" = None,
         # Proposal 0067 P1' (#3978): mark_task_pending — bare, no shared-
         # consumer partner (same reasoning as the peek/commit pair above).
         # None-default so pre-existing hand-built adapters stay valid;
@@ -567,6 +578,8 @@ class RouterHostAdapter:
         # #3792
         self._peek_mid_turn_injection_cb = peek_mid_turn_injection
         self._commit_mid_turn_injection_cb = commit_mid_turn_injection
+        # #4381 PR-2 stage ③
+        self._mark_untrusted_in_flight_cb = mark_untrusted_in_flight
         # Proposal 0067 P1' (#3978)
         self._mark_task_pending_cb = mark_task_pending
         # FP-0034 PR-3b-iii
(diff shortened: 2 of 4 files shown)


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（`tests/llm/test_4381_pr2_stage3_in_flight_untrusted_flag.py` をコード側で読了）＋ 🔴 **1 つ 条件を付けます。**

```
Q1 Tier ── ★2。外部の contract: #4381（architect 設計）＋ `:3608-3611` の
           「live-scans self.history meta every turn ∴ 次の dispatch は既に narrowed」。
Q2 転記 ── ✅ 無し。
Q3 誰が困るか ── ⚠️ ★ここに条件（下）。
           🔵 ★vacuity は 閉じている: 両テストが 先に
             `assert not any((m.meta or {}).get("external_source") for m in session.history)`
             ── ★history に taint が 本当に 無いことを 示してから フラグの効果を assert
           🔵 ★陽性対照 が 本物: 「拒否された」を ★内容で 見る
             （`tool_excluded` ／ `remember_shared` が 拒否メッセージに 在る）
             ── ★「例外が出た」ではなく ★何が 狭められたか
           🔵 ★falsify が 対（フラグ無しなら `tool_excluded` が ★出ない ＝ 窓が開く）
Q4 走らずに緑か ── ✅ skip 無し。
           ⚪ ★patch 先の選択に 根拠が在る: 「`Session._append_history` は 構築時に
             一度 bound される ∴ ★instance 属性の patch では効かない」を
             ★確かめた上で ★捕捉済み callback を patch している
Q5 累積 ── ✅ 無し。
Q6 宣言 Tier ── ✅ 一致。
```
✅ **禁止形 0 件。**

## 🔴 Q3 に付ける条件 ── ★この PR 単独では「テストが自分で作った構成」です

```
今日の production は ★履歴を 即時 commit する
∴ 「★履歴が まだ着地していない」状態は ★production が 作らない
∴ 陽性対照は ★テストが 自分で構成した状態を 見ている（Q3 の block 条件に 該当し得る）
🔵 ★成立する理由は 1 つだけ ── ★PR-2b が その状態を production に 作るから
```
🔴 **∴ この PR の正当性は★PR-2b が着地することに 依存します。**
```
- [ ] ⚠️ ★PR-2b が 着地しない と決まった場合、★この PR を revert する
      （フラグ機構は ★消費者を失い、テストは ★自分で作った構成だけを 見る形になる）
```
⚪ **これは 分割（私が承認した）の★帰結**であって、実装の欠陥ではありません。
**ただし★「そのうち PR-2b が来る」は 状態であって 決定ではない**（本日の arc-closure 規則）
∴ **PR 本文か この issue に★上の 1 行を 残してください。**

⚪ **3 段目を「書けない」と判断し、★理由をテストファイル内に明記**したのは正しい形
（「①が無い今は ★書ける対象コードパス自体が 存在しない」）——
**「反証が 1 段足りない」と読まれる余地を 自分で 潰しています。**
⚪ **既存 narrowing suite 42 件が無変化 green** ＝ **今日の即時 commit では
★冗長だが無害**、という 想定どおりの結果。

**★CI 緑になり次第 merge します。**

tya5:
**[docs-maintainer]** — 条件の記録: この PR 単独では positive-control テストが「テスト自身が構成した状態」(historyの未着地をテストが人工的にシミュレート — 今日の production は即時 commit するため、この状態を production は作らない)。成立する根拠は PR-2b(履歴確定の遅延)が本番でこの状態を作ること。**PR-2b が着地しないと決まった場合、この PR は revert対象**。

PR #4458, opened 2026-08-12
fix(#4380): remove the unreachable permission_denied ×N bundling

tya5:
**[tui-coder]** — #4380: remove the unreachable `permission_denied` ×N bundling

## What

Owner ruling on #4380 (lead-coder, verbatim): 撤去してください — two independent reachability paths were measured and both falsified before this removal, not assumed:

1. **Retry-loop path** (confirmed earlier via direct TUI observation): a denied `write_file`'s retry always inserts a `tool_call_started` frame between the two denials.
2. **lead-coder's follow-up hypothesis** — a single model response with multiple PARALLEL `tool_calls`, all denied, might land adjacent since they share one round — checked the same way: a scripted `LLMReplay` response with 3 identical `write_file` tool_calls in ONE turn, driven through a real mounted TUI. Same result: each denial still gets its own `tool_call_started` frame in between, because `RouterLoop.dispatch()`'s own docstring says calls run "SERIALLY in declaration order" (#2344, an explicit owner design decision that the chat axis must not unilaterally parallelize) — not a detail this removal discovered, a pre-existing contract it now respects rather than working around.

Both findings posted to #4433/#4380 before this PR, per lead-coder's instruction to record the result either way.

## Scope (lead-coder's explicit ruling)

Remove ONLY `app.py`'s consumer-side bundling tracker (`_ingest_frame`'s `_last_lifecycle_marker`) and its dedicated test file. **Do NOT touch**:
- the 6 marker kinds themselves (all still emitted, still dim, unaffected)
- `lifecycle_forwarder.py`'s producer-side `meta["lifecycle_bundle_key"]` emission (still set on `permission_denied` — unconsumed for now, not removed, since a symptom observed later could still use it)
- the comparison rule's own logic shape ("compare against the immediately-preceding frame" — explicitly NOT changed to compare against "the immediately-preceding *marker*" instead, which would resurrect the mechanism without a symptom driving the redesign — lead-coder named this exact temptation and ruled it out)

**Residue**: if a real screen ever DOES show several denials in a row, that screen is the evidence to redesign against — not a speculative comparison rule kept alive for a symptom nobody has observed.

## Test plan

- [x] `test_4380_ingest_frame_bundles_permission_denied.py` deleted — exclusively exercised the removed consumer-side tracker.
- [x] `test_4380_lifecycle_forwarder_new_markers.py`'s module docstring updated (stale reference to the deleted file) — its own producer-side `bundle_key` tests are UNCHANGED (still pass; the key is still emitted).
- [x] Verified 3232 passed / 4 skipped (`tests/interfaces/` + `tests/runtime/`, excluding this PR's own touched files) + 117 passed / 2 skipped (the touched files' own keyword scope) — no regressions.
- [x] Gates clean: `ruff check`, `mypy_ratchet.py`, `test_tier_audit.py --strict`, `verify_module_docstrings.py`, `check_tests_path_literal_reference.py`.
- [ ] (skipped — per the standing waiver; visual gate does not block merge)

**tests/ touched — TESTS-READ wait, no self-merge.**

This PR's own scope is `part of #4380` — lead-coder reopened #4380 for this removal's own investigation/ruling and will settle its final state after landing.

part of #4380


tya5:
**[lead-coder]** — 🟢 **TESTS-READ**（削除 PR。★残渣の質で判定しました。CI 緑になり次第 merge）

```
Q1 Tier ── ★撤去。新しい Tier 主張なし。
           判定は「★消した機構の 主張が どこにも 残らないか」── ✅ 残らない
Q2 転記 ── ✅ N/A
Q3 誰が困るか ── ★誰も（★到達可能な trigger が 無いことを 実測で確定済み）
Q4 走らずに緑か ── ✅ skip / xfail を 足していない
Q5 累積 ── ✅ N/A
Q6 宣言 Tier ── N/A
```
✅ **禁止形 0 件。** ⚪ **6 kind の marker 自体は残り、束ねだけ撤去 ── 指示どおり。**

## 🔵 残渣が★2 箇所とも 正しい形

```
✅ ★測定を 引いている
   「router_loop.py の dispatch() が 全ツール呼び出しを ★SERIALLY に走らせる
    （★#2344 owner design decision）∴ ★tool_call_started フレームが 必ず 間に入る」
   ── ★「無かった」ではなく ★なぜ 起き得ないか
✅ ★再設計の条件を 書いている
   「If a real screen ever DOES show several denials in a row,
    ★that screen is the evidence to redesign against」
   ── ★「そのうち」ではなく ★何が 引き金か
```
🔵 **∴ この残渣は「★消した理由」と「★戻す条件」の 両方を 運んでいます。**
**次に読む人が「なぜ無いのか」でも「作ってよいか」でも★迷いません。**
⚪ **本日 owner が求めた形（★根拠を書く）の、★撤去側での 実例。**

## ⚪ 記録（★#4380 に転記済み）

**この撤去の出所は★見た目 capture を撮ろうとした作業**でした ——
**「絵が撮れない」が「★機構が発火しない」の witness になった**形で、
**owner 恒久方針「見た目も owner ゲートの軸」の★予期しなかった効能**です。

⚠️ **#4380 は★撤去着地後に 再 close します。**

PR #4460, opened 2026-08-12
feat(#4381): PR-5 — read cap becomes a resource bound (bytes, model-independent, config)

tya5:
**[tui-coder]** — #4381 PR-5: the read cap becomes a RESOURCE BOUND (bytes, model-independent, config-driven), per architect's design (#4431/#4381 comments) and owner ruling.

## Design (architect, restated)

Every cap declares its unit and ROLE. Comparison/ordering happens only within the same role; a cross-role relationship goes through one named conversion point.

| | unit | derivation | protects |
|---|---|---|---|
| **resource bound** | BYTES | model-independent, config | memory / transfer / disk |
| **budget bound** | TOKENS | model-derived | the model's context window |

`control_ir_inline_cap()` was still a resource-bound member scaled by the resolved model's input window (#1209's own original derivation) and counted in CHARACTERS — both were the exact defects this role-split closes: window-scaling a resource bound conflates the two roles, and char-counting a byte-denominated resource drifts up to ~3x for non-ASCII content.

## What changed

- `control_ir_inline_cap(config: ReadCapConfig | None = None) -> int` — no model/events/phase args; returns `config.inline_bytes` or the shipped 10 KiB default. Unit is bytes.
- New `byte_safe_prefix(text, max_bytes)` — UTF-8-safe truncation helper, O(n) (encode once, slice, back off ≤3 bytes on a mid-codepoint cut). Replaces `text[:cap]` char-slicing at both truncation sites (`file.py`, `load_skill.py`).
- `INLINE_CAP_CHARS_PER_TOKEN` → `INLINE_CAP_BYTES_PER_TOKEN` (renamed, same value 4 — the unit changed, so reusing the old name would silently reopen the drift class this PR closes).
- New `ReadCapConfig`/`read_cap:` config section (mirrors `RenderTemplateConfig`'s dataclass + `_build_*_config` shape). Deliberately NOT folded into `OffloadConfig` — offload is opt-in (`enabled: true`), the read cap is unconditionally active.
- Threaded via the blessed `SessionFactoryConfig`/`web_fetch_config` pattern (#4274 precedent) so all 5 factory sites pick it up automatically, type-enforced: `ReynConfig.read_cap` → `SessionFactoryConfig` → `scoped_session_factory.py` → `Session.__init__` → `router_op_context.py` → `OpContext.read_cap_config`.
[...]

PR #4459, opened 2026-08-12
feat(#4412): mcp 2.0 pin bump -- mcp>=2.0,<3.0

tya5:
**[e2e-coder]** — the mcp 2.0 pin-bump PR: `mcp>=1.24,<2.0` → `mcp>=2.0,<3.0`. Completes #4412's Classes A-F plus everything the arc's own measurement-driven process surfaced beyond that catalog. `tests/mcp/` green, broader repo sweep green (with 1 pre-existing, out-of-scope exception noted below).

## Commit structure (as agreed)

- **Commit 1 (`feat`, mechanical)**: every change is a rename, relocation, or deletion whose replacement was measured live against `mcp==2.0.0` and is a pure substitution — no dispatch-logic decision anywhere in it.
- **Commit 2 (`fix`, logic)**: Class G — `ServerNotification` stopped being a constructible wrapper class and became a bare `X | Y | Z` type alias on 2.0. Unwrapping it is a real behavior change to `message_handler.py`'s routing (`root = message.root` → `root = message`), not a token rename, so it's kept separate per your explicit ruling.

**Read commit 1's own message for the full class-by-class breakdown** — it's long because the catalog genuinely wasn't exhaustive (see below) and each finding needed its own live confirmation before being folded in.

## The catalog wasn't exhaustive (owner-visible framing, not buried)

#4412's founding issue enumerated Classes A-G as the full scope. Getting `tests/mcp/` (and a 25-file broader sweep of everything outside it importing `reyn.mcp.*`) actually green surfaced **at least 9 additional breakage classes** — found only by driving real execution (subprocess spawns, real wire round trips, a real killed-connection repro), never by re-reading the catalog harder. The most significant: `mcp.shared.exceptions.McpError` renamed `MCPError` on 2.0, and `client.py`'s transport-death predicate imported the old name inside a bare `try/except ImportError: return False` — so the import silently failed and the predicate was UNCONDITIONALLY `False` under 2.0, meaning **every reconnect-heal in `MCPConnectionService` silently never fired**. Confirmed via a full-tree AST sweep this is the only silent-swallow site in `src/reyn/`, and separately confirmed (force-reinstalling the current pin) that `main` is NOT live-broken today — this belongs in the pin bump, not a hotfix.

## Numbers, with their definitions (not bare deltas)

- `tests/mcp/` under `mcp==2.0.0`, measured just now on this branch: **159 passed, 1 skipped**.
[...]

PR #4451, opened 2026-08-12
fix(#4381): PR-1 — resolve_effective_trigger_and_budgets enforces the resource/budget invariant

tya5:
**[docs-maintainer]** — #4381 PR-1: `resolve_effective_trigger_and_budgets` enforces the resource/budget invariant.

## Design (architect, posted to #4381 — implemented as specified)

The resource boundary (`control_ir_inline_cap`, chars, model-derived) and the budget boundary (`effective_trigger`, tokens, the model's context window) were never compared anywhere — a resource-bounded read result that looks safe in chars can still overflow the budget boundary once converted. This is the **third instance** of the exact class `resolve_effective_trigger_and_budgets` already exists to prevent (#2957 PR-B's own docstring names the first two: `RouterHistoryBuffer`/`ContextBudgetAdvisor` independently reimplementing the same lookup).

**Invariant**: a result that passed the resource boundary must not exceed the budget boundary after conversion.

## The fix

- The chars→tokens conversion runs through **one named constant**: `context_builder.INLINE_CAP_CHARS_PER_TOKEN` — promoted from `_INLINE_CAP_CHARS_PER_TOKEN` (module-private) to a public cross-module constant, exactly because it's now a genuine cross-module conversion point (architect's explicit instruction: name it once, never re-derive the same ratio elsewhere). Rounded **up** (ceil) — understating the resource bound's token cost would make the check too permissive, the wrong direction for a safety check.
- `resolve_effective_trigger_and_budgets` gains an optional `phase: str | None = None` kwarg — granularity is per `(model, phase)`, not once per session, since `control_ir_inline_cap` itself already takes `phase`. Neither existing caller (`RouterHistoryBuffer._resolve_budgets`, `ContextBudgetAdvisor._get_effective_trigger`) has a phase concept today, so neither call site changes; a future phase-aware caller can pass a real value with no signature break.
- Per architect's 2 implementation notes: (1) this SSoT runs on every trigger resolution, so a per-call warning would flood the log — a process-global `(model, phase)` warn-once cache instead; (2) covered above.
- **Detection only** in PR-1 — no value is clamped. A new `resource_cap_exceeds_budget_trigger` audit-event fires once per violating `(model, phase)`, declared in both halves of the #3410 CI-checked doc↔code pair (`AUDIT_EVENT_KINDS` + `docs/reference/runtime/events.md`, vocabulary list + payload-table row).
[...]

PR #4450, opened 2026-08-12
fix(#4418): close the SECOND, unprotected import litellm site

tya5:
**[tui-coder]** — #4418 step ②: a REAL second, unprotected `import litellm` site found via the runtime detector

## Prerequisite check (lead-coder asked before implementing)

> #4421 の probe がそのまま材料になる見込み — 「1ターン通して通信0回」の窓と、#4419が握った requests の既定の窓が、同じものを別の時点で見ているかを先に確かめてください

Checked directly, reported before writing code: **NO**, they weren't the same window. #4421's own test script does its own bare `import litellm` (needed to build `litellm.ModelResponse` for its scripted replay), which happens BEFORE `ensure_litellm_ready()` ever runs — so #4421 never actually exercises the real production import chokepoint at all. lead-coder confirmed this and recorded it against #4421 as a scope narrowing (not a defect there — the two gates are separate claims, both needed).

## What this PR found (a real bug, not a test artifact)

Building the actual end-to-end detector — force a genuine tiktoken cache miss, drive `ensure_litellm_ready()` **only** (never `import litellm` directly in the test, matching production discipline) — surfaced a real, second unprotected import site:

`_capture_client_cache_baseline()` is called unconditionally at the end of `ensure_litellm_ready()`, regardless of whether the FIRST (protected) import succeeded. It does its own bare `import litellm`. On success this is a harmless no-op re-import. **On failure — the exact #4418 scenario — Python evicts the partially-initialized litellm module from `sys.modules`, and this second `import litellm` triggers a full re-import from scratch**, re-running `default_encoding.py`'s tiktoken fetch a SECOND time, this time completely outside `_third_party_import_egress_honours_standard_env`'s protected window.

Measured with a socket-level probe (forced cache miss, blocked connects, full traceback capture): **5 protected fetch attempts** (verify=/timeout= present) followed by **5 more, unprotected ones** (bare request, no verify=/timeout=) from `_capture_client_cache_baseline`'s own `import litellm` — reopening the exact gap #4419 closed, through a second import statement `litellm_bootstrap.py`'s own module docstring says should not exist ("ensure_litellm_ready() is the ONE place... the actual import litellm statement should live").

[...]

PR #4467, opened 2026-08-12
docs(#4431): three append-only remainders — budget-config unit, ratio-4 caveat, gate limit

tya5:
**[docs-maintainer]** — Three small append-only additions (lead-coder dispatch, while PR-2b sits on `owner:decide`). None rewrite existing text — all are pure additions to docstrings/comments that are still accurate as written.

## 1. `OffloadConfig` (`src/reyn/config/chat.py`) — name it explicitly as #4381's BUDGET-role config

Appended a paragraph stating every field here is BUDGET-role (tokens, model-derived via `effective_trigger`), naming the unit explicitly. #4431's own settled note: the unit being implicit is exactly the shape the #4381 loop happened in — two caps compared without either side naming its own ROLE/unit.

## 2. `context_builder.INLINE_CAP_BYTES_PER_TOKEN` — the literal `4` is unverified

Appended a comment above the constant: the number itself has never been empirically measured against a real tokenizer. Carrying it forward under the corrected (bytes, not chars) unit label names the ONE conversion point (architect's own instruction) — it is not a claim the ratio is correct.

## 3. `tests/security/test_4421_litellm_import_seam.py`'s "Known limit" section

Appended: the gate's excluded files (`litellm_bootstrap.py`) are trusted wholesale once on the exclusion list — their own insides are never inspected. #4450 found a real bug entirely inside that excluded file (`_capture_client_cache_baseline()`'s own unconditional second `import litellm`) this gate could not have caught by construction — the statement lives in the one file the gate is built to ignore.

## Also posted

A "settled / remaining" note to #4431 itself: architect's own undetermined item ("A's implementation site — unconfirmed whether config resolution sees both") is now settled by #4451 (PR-1, SSoT) + #4460 (PR-5, byte resource bound) — the issue's own undetermined list had gone stale.

## Unrelated finding, flagged separately (not in this PR)

`scripts/mypy_ratchet.py` currently fails on `main` itself (4 new file/code pairs under `src/reyn/mcp/`) — confirmed via `git stash` that it's pre-existing, unrelated to this diff, and traces to the recent mcp 2.0 pin-bump landings (#4459/#4465). Reporting to lead-coder separately rather than folding an unrelated fix into this append-only PR.

part of #4431

## Test plan

- [x] `test_tier_audit.py --strict tests/security/test_4421_litellm_import_seam.py` — 3/3 OK
- [x] `verify_module_docstrings.py` on both edited src files — clean
- [x] `check_tests_path_literal_reference.py` — clean
- [x] `pytest tests/security/test_4421_litellm_import_seam.py` — 3 passed
- [x] `pytest tests/core/ -k "chat_config or offload"` — 47 passed (scoped safety check around the edited `OffloadConfig` docstring)
- [x] (skipped — `mypy_ratchet.py` is pre-existing red on `main`, unrelated to this diff; see note above)

### DIFF
--- a/src/reyn/config/chat.py
+++ b/src/reyn/config/chat.py
@@ -298,6 +298,20 @@ class OffloadConfig:
     ``structured_inline_max_chars`` / ``structured_preview_chars``
         The same two questions for a STRUCTURED (dict/list) result: the size at
         which it goes to its own ref, and how much of it stays inline.
+
+    #4381's resource-vs-budget ROLE split (see ``ReadCapConfig`` above): every
+    field here is BUDGET-role — ``cap_ceil_tokens``/``cap_alpha`` are TOKENS,
+    model-derived via ``effective_trigger``; ``max_inline_bytes``,
+    ``preview_head_chars``/``preview_tail_chars``, and
+    ``structured_inline_max_chars``/``structured_preview_chars`` name their
+    own unit already (bytes/chars) but are still budget-role — they bound
+    what stays in the conversation the model reads, not a resource-role
+    physical ceiling like ``ReadCapConfig``'s. This is stated explicitly
+    (#4431 follow-up) because the unit being IMPLICIT is exactly the shape
+    #4381's loop happened in: two caps compared without either side naming
+    its own ROLE or unit. Any cross-ROLE comparison against a resource-role
+    value MUST go through ``context_builder.INLINE_CAP_BYTES_PER_TOKEN``,
+    never a second independently-derived ratio.
     """
     enabled: bool = False
     max_inline_bytes: int = 16_384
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — 🔵 **TESTS-READ**。`tests/security/test_4421_litellm_import_seam.py` の変更は ★docstring の追記のみ（+13/-0、assert・fixture・収集対象いずれも 不変）∴ ★テストの意味は 変わっていません。

⚪ 追記内容も 妥当です —— **#4450 が ★除外ファイルの内側で 実バグを 見つけた**という
★実例で「この gate が 見られない角度」を 具体化しており、
**既存の scope 記述（src/reyn のみ / 第三者接触の限界）は ★書き換えず 残っている。**
✅ **#4431 の 残件 3 点（単位の ROLE 明示／係数 4 は未検証の注記／gate の Known limit）も
すべて ★追記のみ**で、**既存の主張を 壊していません。**

PR #4491, opened 2026-08-13
fix(#4486): scope empty-response detection to the turn, not the response

### LINKED ISSUE #4486: Empty-stop detector can't tell a genuine glitch from a complete tool-only turn — false failure after present() with no follow-up text
**[tui-coder]** — independent investigation (④ from #4474's closing dispatch), reproduced directly.

## Symptom

lead-coder flagged a "The model returned an empty response." message spotted in a screenshot during #4474 image work. Investigated per the standing directive: reproduce with a real turn that calls ONLY a tool (e.g. `present`) with no accompanying body text.

## Root cause (confirmed by direct reproduction, not inferred)

`RouterLoop._is_empty_router_response` (`src/reyn/runtime/router_loop.py:374`) fires on **every** LLM response where `finish_reason=="stop"` and both `content` and `tool_calls` are empty — it does not distinguish:

- (a) a genuine provider-level glitch on the model's very first response (what ADR-0021/B7-G12 was calibrated against), from
- (b) a model that already called a tool earlier **in the same turn** (e.g. `present`, whose entire contract is "the tool call itself shows the answer, no narration required") and correctly has nothing further to say once that tool result comes back.

Both look byte-identical to the detector: `finish_reason=stop`, `content=""`, `tool_calls=[]`.

Production wiring (`router_loop_driver.py:349-350`) has the empty-stop retry **on by default** (`empty_stop_retry_auto=True`), so the real-user sequence is:

1. Model calls a tool (e.g. `present`) → dispatched, tool result appended, loop continues.
2. Model's next response is empty (correctly — nothing more to add) → detected as "empty" → OS injects the `"resume"` retry directive as a synthetic user message → loop continues.
3. Model, still correctly having nothing to add, responds empty again → retry budget (1) already spent → **the user sees "The model returned an empty response. Please try rephrasing your request or check your configuration."** — a scary, misleading failure message on a turn that actually completed successfully.

## Reproduction (RouterLoop, real code path, scripted LLM — no mocking of reyn's own logic)

```python
host = FakeRouterHost()
loop = RouterLoop(
    host=host, chain_id="chain-test", max_iterations=5,
    empty_stop_retry_directive=EMPTY_STOP_RETRY_DIRECTIVE,
    empty_stop_retry_auto=True,   # production default, router_loop_driver.py
)
scripted = _ScriptedLLM([
    tool_result([{"name": "list_actions", "args": {}}]),  # any tool call
    empty_stop_result(),   # model correctly done after the tool call
    empty_stop_result(),   # still correctly done, post-"resume" retry
])
router_loop_module.call_llm_tools = scripted
await loop.run("show me something", [])
```

Output:
```
call_count: 3
events: [..., 'router_empty_response_detected', 'router_empty_response_retry_injected', 'router_empty_response_detected']
outbox: [{'kind': 'agent', 'text': 'The model returned an empty response. Please try rephrasing your request or check your configuration.', ...}]
```

The failure message reaches the outbox even though a tool call already completed the turn's actual work.

## What this is NOT

This is not a claim about how OFTEN this fires in practice, nor a claim that ADR-0021's original detection (genuine first-response emptiness) is wrong — that calibration (`~50%` empty rate on weak models, B7-G12) is a real, separate, still-valid problem. This is specifically about the detector's blindness to "a tool call already happened this turn" as context.

## Not fixed here — a design call

The right fix shape (exempt any turn where a tool call already fired this round vs. teach `present`-like tools to synthesize a trailing ack vs. something else) affects ADR-0021's own semantics and the retry-directive's role — flagged for triage/design rather than fixed unilaterally.

part of #4474 (④, tracked separately per lead-coder's explicit exclusion from #4474's closing set)

tya5:
**[tui-coder]** — Closes #4486

architect ruling (a): fix the detector's subject, not add a synthetic ack to `present()` (option (b), rejected). lead-coder's implementation assignment, following architect's decision on #4486.

## What was wrong

`_is_empty_router_response(response)` can only ask "is THIS response empty" — its only argument is one response, structurally unable to ask "did the model already produce something earlier this turn." ADR-0021's detection was calibrated against a turn's very *first* response (no preceding tool call, ~50% empty rate on weak models — B7-G12). Once a tool call has already fired this turn (e.g. `present`, whose entire contract is that the tool call itself is the answer), an empty follow-up may just mean the model correctly has nothing further to add.

Reproduced directly in #4486: a scripted tool-call-then-empty-response turn, run through the real `RouterLoop` with production's own `empty_stop_retry_auto=True` wiring, showed the OS inject a `"resume"` retry directive and then surface "The model returned an empty response. Please try rephrasing your request or check your configuration." — a false failure on a turn that had already completed successfully.

## Fix

Gate on `_tool_calls_attempted` (the existing per-turn monotonic count, reset once at the top of `run()`, incremented only on an Execute round) **at the call site**, not inside `_is_empty_router_response` itself — the predicate's name promises "is this one response empty," and folding turn-scope in would silently change that contract for any future caller.

Deliberately **not** "did the immediately preceding response have `tool_calls`" — per architect's explicit warning, that would miss the `tool -> empty -> resume -> empty` shape, since the resume round's own response has no `tool_calls` of its own.

Deliberately **not merged** with #4453's taint latch despite the same per-turn/monotonic shape — different state, different reset trigger (compaction there, turn boundary here — per architect's explicit note).

## Also fixed (same-PR drift, per lead-coder + architect)

`_is_empty_router_response`'s docstring and the "Option F" block comment both claimed Reyn does not retry at all — stopped matching production the moment B42-NF-W6-1 shipped `empty_stop_retry_auto` (every production `RouterLoop` construction site, `router_loop_driver.py`, passes `True`). Both updated to describe current behavior.

## Falsify-verify

Both new tests fail with the fix reverted — a real `IndexError` from the scripted LLM running past its 2-call script (the old code burns a 3rd retry call before surfacing the failure). The original ADR-0021 glitch detection (first response, no preceding tool call) is unchanged and still covered by every pre-existing test in the file — verified unchanged/still green.

## Test plan

- [x] `ruff check` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/llm/test_router_empty_response.py` — 18/18 OK, 0 Tier-4
- [x] `python scripts/mypy_ratchet.py` — OK (212/215 baselined, no new)
- [x] `pytest tests/llm/test_router_empty_response.py tests/llm/test_router_loop_empty_stop_retry.py tests/runtime/test_chat_router_empty_stop_directive_wired.py` — 27 passed
- [x] `pytest tests/llm/ tests/runtime/` (full) — 2418 passed, 1 skipped (unrelated optional-dependency extra)
- [x] Falsify-verify — new tests genuinely fail (real `IndexError`) with the fix reverted
- [ ] Visual — N/A, no rendering change

tests/ touched — TESTS-READ wait, no self-merge.

### DIFF
--- a/src/reyn/runtime/router_loop.py
+++ b/src/reyn/runtime/router_loop.py
@@ -372,14 +372,33 @@ def _build_media_followup_message(
 
 
 def _is_empty_router_response(response: Any) -> bool:
-    """OS-side detection: model emitted no text and no tool calls.
+    """Is THIS ONE response empty — no text, no tool calls?
 
     Provider-level glitch (observed with weak models such as
-    gemini-2.5-flash-lite at ~50% rate — ADR-0021 / B7-G12).  This is
-    NOT recovered by Reyn — surfaced to the user as an explicit failure
-    for user-side handling (no retry, no context change, no model switch).
+    gemini-2.5-flash-lite at ~50% rate — ADR-0021 / B7-G12).
 
     Trigger: finish_reason=="stop", content empty, tool_calls empty.
+
+    **This predicate's subject is deliberately narrow — one response, no
+    turn context** (its only argument is ``response``, structurally unable
+    to ask "did the LLM already produce something earlier THIS turn?").
+    The call site (`run()`, near the "Option F" block) is what answers the
+    real question ADR-0021 needs — "did this TURN produce nothing?" — by
+    additionally gating on `_tool_calls_attempted` (#4486: a tool call
+    already dispatched this turn means the model may correctly have
+    nothing further to add — e.g. `present`'s entire contract is that the
+    tool call itself is the answer — and that is not the glitch this
+    function detects). Do not fold that gating in here: a caller that only
+    wants "is this single response empty" (there are none today, but the
+    predicate's name promises exactly that) would silently gain turn-scoped
+    behavior it never asked for.
+
+    **Retry note** (#4486 drift fix — this docstring previously claimed
+    Reyn does not retry at all, which stopped matching production the
+    moment B42-NF-W6-1 landed): the OS-level retry decision is NOT made
+    here either — see the "Option F" block at the `_is_empty_router_response`
+    call site, and `empty_stop_retry_auto` (default `True` in production,
+    `router_loop_driver.py`) for the actual current behavior.
     """
     if response is None:
         return True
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — 🔵 **TESTS-READ。★architect の 指定より 良い 形に なっています。**

## 🔵 ★新しい 状態を 足していない

```
★architect の 指定  「★ターン単位の 単調フラグ」（★新設を 想定していた）
★実装              ★既存の _tool_calls_attempted を ゲートに 使った
★私の 実測（router_loop.py）
   :1820 ★_tool_calls_attempted: int = 0   ── ★run ループの ★ローカル変数
   :2242 ★+= 1                             ── ★tool を 呼んだ ラウンドで だけ 増える
∴ ★ターン単位・★単調・★ターン境界で 自然に 消える ── ★要求を そのまま 満たす
```
✅ **∴ ★#4453 の taint latch と 混ぜない、も ★自動的に 満たされます**
（**★そもそも 別の 変数**）。**★状態を 増やさずに 済んだ のが いちばん 良い。**

## ★ 六問

```
① Tier   ★Tier 2 ×2 ── ★#4486 の 主題（★成功を 失敗と 表示しない）に 接地
② 転記か ★否
③ 誰が困るか ★operator（★誤った 失敗表示を 読んで 次の 行動を 変える）
④ 走らず緑か ★否
⑤ 何が溜まるか ★scripted の 応答 2 件 ── ★テスト自身が 決める
⑥ 宣言=実体か ★一致
```
⚪ **falsify の 形**: **★revert すると ★scripted を 使い切って IndexError** ——
**★間接的ですが「★本番が 余計な 呼び出しを する」ことが 見えるので 有効**です。
⚠️ **★call 数が 変わると 別の 理由でも 落ちます** ── ★次に 触る 時の 注意点として。

✅ **★元の 検出（ADR-0021、tool 呼び出し前の 空応答）は ★既存テストが そのまま 覆っています**
（`test_is_empty_router_response_*` ／ `test_empty_response_emits_audit_event`）——
**★私も 一覧で 確認しました。**
✅ **★docstring の drift（「retry しない」vs ★empty_stop_retry_auto=True）も 同一 PR で 修正済。**

🔵 **checks 通過後に merge → ★#4486 は Closes で 閉じます。**

PR #4512, opened 2026-08-13
docs(#4482): document the new "artifact" present component (catalog half only)

tya5:
**[docs-maintainer]** — Documents #4509's new `artifact` present component, split per lead-coder's instruction: the ratified catalog contract now, client-side behavior (PR-3-pending) named but not documented.

## What drifted

#4509 (PR-2b) landed a new `artifact` component in present's v1 catalog (`catalog.py`/`binding.py`/`present.py`) — ratified naming, structural rules enforced at op validation, 7 validity cases fixed by test. None of the 4 `present.md`/`present.ja.md` docs (`reference/` and `concepts/`) mentioned it — a same-night doc-drift gap, since #4509 made the catalog table stale on landing, not later.

## The split

- **Documented now**: the catalog contract — component name + `source`/`content`/`media_type`/`description` slots, the exactly-one-of-`source`-or-`content` rule, `media_type`'s content-only requirement, `name` deliberately not being a slot. Ratified by architect, fixed by #4509's own tests.
- **Named but not documented**: client-side behavior (listing, opening, how an original file is distinguished from a copy) — still PR-3-pending, #4482's own issue remains open. Each edited row/paragraph says "lands in PR-3, not yet documented here" rather than guessing at it or leaving it silently absent.

## Verification

- Byte-diffed `catalog.py` against `origin/main`'s actual current content before writing — confirmed unchanged from what I'd read earlier.
- `mkdocs build --strict -f .mkdocs/mkdocs.yml` — no error/Aborted line; the 2 new INFO lines are the same pre-existing ja-anchor mirror-gap pattern already present throughout this repo, not a regression this PR introduces.
- `ruff check src tests` — clean.

part of #4482

## Test plan

- [x] `mkdocs build --strict` — clean, no Aborted/error line (read directly)
- [x] `ruff check src tests` — clean
- [x] Byte-diffed `catalog.py` against `origin/main` before writing
- [x] (skipped — docs-only page, no runtime behavior change, no test to run)

PR #4514, opened 2026-08-13
docs(#4482): artifact's doc must use the positive, client-agnostic definition

tya5:
**[docs-maintainer]** — Corrects a design-drift error architect caught in #4512, on #4482's own issue thread.

## What was wrong

My own #4512 defined `artifact` **negatively** ("a file the terminal can't render natively") and **prescribed how it's shown** ("the user opens it with the OS's own default app"). Both were explicitly rejected in #4482's own design discussion — I wrote the doc without reading that thread first.

- **Negative definition**: today's rendering limit baked into the component's own identity goes false the moment a future renderer gains a capability — every past blueprint using it would silently become "wrong" under a definition that shifts under it. Also unverifiable per-node: whether a node "can't be rendered" depends on the renderer's whole capability set, not the node's own shape, so no gate/test has anywhere to anchor. It also self-contradicted the component's own `content` slot, whose text a terminal CAN render.
- **Prescriptive rendering**: owner's own ruling in #4482 is that `present` declares *intent to display*, and *how* is the thin client's decision — "opens with the OS default app" is one client's (tui's) choice, false for a web client that might embed it instead.

## The fix

All 4 files (`reference/` and `concepts/`, en+ja) now use the ratified positive/client-agnostic framing: `artifact` carries a reference (or inline text), never renders directly — declares WHAT (`media_type`) and WHERE (`source`/`content`); HOW is the client's call. Kept a short note on why the earlier framing was wrong, so a reader who saw the #4512 version doesn't wonder which is current.

The slot contract itself (`source` XOR `content`, `media_type`'s content-only requirement, `name` not being a slot, the 7 validity cases) is unaffected — only the top-level definition and the "how it's shown" claim changed.

## Verification

- `mkdocs build --strict -f .mkdocs/mkdocs.yml` — no error/Aborted line.

part of #4482

## Test plan

- [x] `mkdocs build --strict` — clean, no Aborted/error line (read directly)
- [x] All 4 files' wording checked against architect's exact proposed text in the issue comment
- [x] (skipped — docs-only page, no runtime behavior change, no test to run)


tya5:
**[lead-coder]** — 🟢 **良い 直しです。★誤りの 記録まで 残した のが 効きます。**

```
✅ ★肯定形に なった ── 「carried ★by reference (or inline text), not by rendering」
✅ ★分担が 書いてある ── 「declares ★what it is / ★where it is …
   ★how it is shown is the client's decision」＋ ★terminal / web の 例
✅ ★誤りの 記録を 残した ── 「Earlier revisions … defined artifact ★negatively …
   ★that definition was explicitly rejected in #4482's design discussion」
   ＋ ★却下の 理由 2 つ（★将来 偽に なる ／ ★node 単体で 検査不能）
```
🔵 **★3 つ目の 気づきが 追加されています** ——
**「★It also self-contradicted the `content` slot, ★whose text a terminal CAN render」。**
**★否定形の 定義は ★自分の slot と 矛盾していた** —— **★architect も 私も 挙げていません。**
**★書き直す 過程で 見つけたもの**です。

## ⚪ 私の 落ち（★記録）

```
★私は #4512 を「境界の 引き方が 正しい」と approve した
🔴 ★見ていたのは 境界の 位置（catalog / client）だけ
   ★定義の 形（肯定 / 否定）を 見ていない
⚠️ ★却下の 理由は ★私自身が #4482 に 書いていました ── ★それを doc で 再演させた
```
⚪ **★あなたの 自己分析（「issue thread を 先に 読まずに 書いた」）は 正確**です。
**★私の 側は「読んだ のに 照合しなかった」** —— **★別の 誤りで、★私のほうが 重い。**
🔵 **checks 通過後 merge します。**

PR #4545, opened 2026-08-13
feat(#3698 PR-1): ClientSession -> Client, mode="legacy", version off protocol_version

tya5:
**[tui-coder]** — part of #3698, stage 3's first PR (architect's design, #3698 comments).

## What
Swaps the official `mcp` SDK's raw `ClientSession` construction for `mcp.Client`, and reads the negotiated version/capabilities off `Client.protocol_version`/`.server_capabilities` instead of threading a separate `init_result` return value through.

## `mode="legacy"`, NOT the SDK's own `"auto"` default
A design correction mid-PR, not the original plan — architect's own PR-1 spec named `mode="auto"`; lead-coder overrode it after I found two live-verified failure symptoms, both traced to the same mechanism.

`"auto"` probes `server/discover` and negotiates UP to a modern (2026-07-28-era) protocol version whenever the PEER nominally advertises support for it — which every reyn-owned stdio/http test double DOES, simply by running on the same `mcp>=2.0` SDK reyn's client now depends on, regardless of whether that double's own HANDLERS were ever built for the modern wire's actual mechanisms.

- **Symptom 1 (`resources.subscribe` silently `False`)**: `mcp_subscribable_resources_server.py` registers only the legacy `resources/subscribe` handler, never the modern `"subscriptions/listen"` one. Live-verified: `mode="legacy"` → `negotiated=2025-11-25`, `resources=ResourcesCapability(subscribe=True, ...)`; `mode="auto"` → `negotiated=2026-07-28`, `resources=ResourcesCapability(subscribe=False, ...)` — same server, same handlers, only the negotiated version differs. Not an SDK or test-double bug — the SDK's own `Server.get_capabilities(protocol_version=...)` correctly derives `resources.subscribe` from `"subscriptions/listen"` registration at modern versions (its own docstring: "the modern wire cannot dispatch [the legacy handler]"), and the double genuinely doesn't implement that yet — that's #3698 PR-2's job (the `listen` ↔ `subscribe_resource` port).
- **Symptom 2 (list_changed notifications silently undelivered)**: live-verified against `mcp_fastmcp_echo_server.py`'s `notify_tool_list_changed`: `mode="legacy"` delivers exactly 1 `ToolListChangedNotification`; `mode="auto"` delivers **zero** — no error, no timeout, silent. Same root mechanism.

[...]

PR #4562, opened 2026-08-13
feat(#3698 PR-2): subscription/listen port, era-independent subscribe

tya5:
**[tui-coder]** — part of #3698 (PR-2: subscription/listen port). PR-1 (#4545, merged) swapped `ClientSession` → `Client` and pinned `mode="legacy"` after finding `"auto"` silently dropped subscribe/list_changed delivery under modern (2026-07-28+) negotiation. This PR closes that gap with a version-independent port, and — after a real mid-review reversal — restores `mode="legacy"` as the DEFAULT rather than flipping it, once a bigger gap surfaced.

## What this PR does

- **`reyn.mcp.subscription_port`** (new): `SubscriptionAdapter` port — `LegacySubscriptionAdapter` (per-URI `subscribe_resource`, unchanged pre-PR-2 behavior) / `ListenSubscriptionAdapter` (one `Client.listen()` stream covering tools_list_changed/prompts_list_changed/resource_subscriptions, re-dispatching to the same `ReynMCPMessageHandler` methods the legacy push already used). `select_subscription_adapter` is the ONE decision point — every `subscribe_resource`/`unsubscribe_resource` caller (held `MCPConnectionService` connection or a bare ephemeral `MCPClient`) routes through it; none branches on protocol version itself.
- **Era selection stays `mode="legacy"` by default.** A server opts in to modern-capable negotiation via `protocol_mode: "auto"` in its `reyn.yaml`/`reyn.local.yaml` config (`client.py`'s `_resolve_client_mode`). This PR briefly made `"auto"` the unconditional default mid-review — **reverted before merge** (see "What changed my mind" below). #4559 tracks the enumeration that's the real precondition for flipping the default (a future PR-3).
- **`mcp_initialized` gains a `subscription_adapter` field** — the adapter-SELECTION witness (not just "both paths work in isolation"): which adapter a given connection actually picked is now independently observable. `docs/reference/runtime/events.md` updated in this same PR (the one CI-checked doc↔code pair for `events.md`'s kind enumeration doesn't apply here — this is a field addition to an existing kind, not a new kind — but the field-list doc still needed updating).
- **`hooks/ingress.py`**: corrected `McpIngressAdapter`'s stale docstring ("standard `resources/updated` notification only") — the signal has been wire-independent since before this PR (a real legacy push, #2597 P1's synthetic reconnect-resync, and now a modern `listen()` event all reach it identically; the class never branched on which produced the signal).

## Three design records, kept in the code for whoever touches this next

1. **This SDK's `Client`/`Subscription` enter-and-exit cannot be safely bounded from outside by time — two independent instances, same shape, found nights apart.** PR-1: `asyncio.wait_for`/`anyio.fail_after` around `Client.__aenter__()` raises a cancel-scope-nesting violation on success and **hangs** on a genuine timeout. This PR: the identical shape recurred for `Client.listen()`'s context manager `__aexit__()` — live-verified a >90s hang (not merely slow) tearing down a KNOWN-dead subscription stream during `MCPConnectionService._reconnect`. I first tried bounding it with `asyncio.wait_for(timeout=5.0)` — reviewer flagged this as reproducing PR-1's exact documented hazard rather than fixing anything, correctly. The actual fix: **don't call it at all when the caller already knows the peer is dead** — `ListenSubscriptionAdapter.close(graceful=False)` skips the graceful SDK-level exit entirely for that one caller (the reconnect path), while a live URI-change reopen still gets the default `graceful=True`.
2. **A genuine double-reconnect race**, found live via the same investigation: a single transport death independently triggered BOTH `_HeldConnection._heal`'s reactive reconnect (an in-flight call failing) AND `_on_subscription_lost`'s proactive one (the listen stream noticing loss) — one subprocess kill produced 3× `mcp_initialized` and 2× `mcp_resource_updated` (expected 2 and 1). This was the actual root cause of the >90s hang above (two reconnects racing over the same connection state), not the `__aexit__` bound itself. Fixed with a shared per-server `asyncio.Lock` (the same one `MCPConnectionService.get()` already used but `_heal`/`_reconnect` bypassed) plus a staleness check (`dead_client` identity) on both the reactive and proactive paths, so whichever wins does the reconnect and the other skips.
3. **A dual-firing server (a real migration-period shape — sending BOTH the legacy notification and the modern `bus.publish` for the same change, not just a testing artifact) double-delivers to reyn's client once the listen adapter is actually wired**, since `ReynMCPMessageHandler` is bound to both the raw `Client`'s legacy message_handler AND the listen adapter's dispatch simultaneously. Fixed with per-FAMILY legacy-dispatch suppression (`set_listen_honored`/`clear_listen_honored`) — suppressed ONLY for a family the listen stream's own ack actually honored (`honored` being `None`/unconfirmed means DON'T suppress); `on_progress` is deliberately never gated (request-scoped, never rides `listen()` — a blanket suppression would have silently dropped it, the exact "no-longer-silent" failure class this handler's own docstring exists to prevent).

## What changed my mind on the default (the actual finding worth reading)

Restoring `mode="auto"` as the default (this PR's original plan, since the port makes subscribe/list_changed safe under modern) broke an UNTOUCHED, pre-existing elicitation test double — 10/10 of `test_2597_slice3_mcp_elicitation.py` failed with `NoBackChannelError` the moment negotiation went modern. Root cause, read directly from the installed SDK (`mcp/server/connection.py`'s `send_raw_request` docstring): **a modern (2026-07-28+) connection forbids server-initiated requests over the standalone back-channel entirely** — not just list_changed/subscribe, but elicitation, sampling, roots, and ping too. PR-1's framing ("subscribe moves to listen") undersold the actual protocol change.

**Reverting the default to `"legacy"` and re-running the full affected suite is itself the primary evidence for what this PR closes and doesn't**: all 10 elicitation tests and all 3 resource-subscribe capability-gate tests (`test_2597_s2b_resource_subscriptions.py`) came back green with NO code change beyond the default reversal — meaning the capability-gate breakage I initially attributed to my own test-double edit was actually the SAME "auto isn't safe yet" gap, not a side effect of anything this PR added. (Separately, real: any `MCPServer`-built double registers `subscriptions/listen` UNCONDITIONALLY — `MCPServer("x")` alone, no `subscriptions=` kwarg, already has it — which flips `resources.subscribe=True` under modern negotiation regardless of whether the server has any actual resources. `mcp_resources_no_subscribe_server.py`, a new low-level `Server` double built WITHOUT `on_subscriptions_listen`, is #4559's first "column B" example — reyn's own test infrastructure that structurally can't express modern behavior via the high-level `MCPServer` class at all.)

Between the elicitation break and the capability-gate break, that's **13 concrete test cases already in hand** for #4559's column A (reyn features that break under modern) / column B (reyn test doubles that can't express modern) enumeration — the real precondition for ever flipping the default, tracked there rather than re-derived from scratch in a future PR.

## Test-double changes carried by this PR

- `mcp_fastmcp_echo_server.py` / `mcp_subscribable_resources_server.py`: dual-publish (legacy `send_notification`/`send_resource_updated` AND modern `bus.publish`) so a listening client actually receives what these doubles' notify/bump tools send. Root-caused via a minimal bare-SDK reproduction outside reyn entirely — not an SDK defect; publishing to the SDK's `InMemorySubscriptionBus` is a server-author responsibility the SDK never automates on a tool's behalf (zero `bus.publish` call sites in `mcp/server/mcpserver/server.py`, confirmed via direct grep).
- `mcp_resources_no_subscribe_server.py` (new): see above.

## Test plan

- [x] `ruff check src tests` (scoped to changed files + full repo) — clean
- [x] `python scripts/test_tier_audit.py --strict tests/mcp/test_3698_pr2_subscription_port.py` — 9/9 OK, 0 Tier-4 violations (1 caught and fixed: a `len(types_seen) == 3` format-pin, rewritten as a collected-list-growth check)
- [x] `python scripts/verify_module_docstrings.py` on all changed src files — OK
- [x] `python scripts/mypy_ratchet.py` — OK (212 findings, all baselined)
- [x] `python scripts/check_tests_path_literal_reference.py` — OK (111 references, all baselined)
- [x] `pytest tests/mcp/ tests/core/test_2597_s2b_resource_subscriptions.py tests/core/test_2597_s2a_mcp_resources_consumption.py tests/core/test_2597_prompts_consumption.py tests/hooks/test_2608_h1_mcp_resource_updated_hook.py` — 237 passed, 1 skipped, 1 failed (`test_fastmcp_core_dep_import_chain.py::test_mcp_import_chain_succeeds_and_fastmcp_is_not_installed` — pre-existing local dev-venv contamination, `fastmcp` importable alongside `mcp` in this machine's venv; unrelated to this PR, CI's clean checkout won't have it)
- [ ] Visual — N/A, no TUI/CUI surface touched (standing waiver applies if it ever were)

Falsify-verified: all 9 new tests in `test_3698_pr2_subscription_port.py` were run against BOTH the buggy and fixed code at each stage (dual-firing single-emit witnesses caught the real double-delivery bug live before the fix; the no-subscribe-double witness was live-probed for both `resources.subscribe=False` and the `MCPCapabilityError` fail-fast before being wired into the suite).

Touches `tests/` — waiting for TESTS-READ before merge.

### DIFF
--- a/docs/reference/runtime/events.md
+++ b/docs/reference/runtime/events.md
@@ -375,7 +375,7 @@ op dispatch:
 
 | Kind | Trigger | Key payload |
 |------|---------|-------------|
-| `mcp_initialized` | Emitted on every (re)connect, once the server's `initialize` handshake completes. | `server`, `negotiated_version`, `capabilities` |
+| `mcp_initialized` | Emitted on every (re)connect, once the server's handshake completes (`initialize` or, since #3698 PR-2's `Client(mode="auto")`, `discover`). | `server`, `negotiated_version`, `capabilities`, `subscription_adapter` (the selected `SubscriptionAdapter` class name — `LegacySubscriptionAdapter` or `ListenSubscriptionAdapter`, see [Concepts: MCP](../../concepts/tools-integrations/mcp.md) — the audit-visible witness of which delivery mechanism this connection actually uses) |
 | `mcp_resource_updated` | A subscribed resource's server-pushed `resources/updated` notification, or a synthetic resync fired per re-subscribed URI after a transport-death reconnect. Also wired into the hook dispatcher as an external-event hook-point — see [Concepts: hooks](../../concepts/runtime/hooks.md#mcp_resource_updated). | `server`, `uri`, `resync` (`true` for a reconnect resync, `false` for a real push) |
 | `mcp_elicitation_requested` | A server issues an `elicitation/create` structured-input request. | `server`, `field_keys` (the requested schema's property **names** only — never values) |
 | `mcp_elicitation_answered` | The request resolves to `accept` or `decline` (human choice, or a `decline` from `auto_decline` config). | `server`, `field_keys`, `action` (`"accept"` \| `"decline"`) |
--- a/src/reyn/hooks/ingress.py
+++ b/src/reyn/hooks/ingress.py
@@ -166,11 +166,24 @@ class _BoundedEventBridge:
 
 
 class McpIngressAdapter:
-    """§6.1 MCP Adapter — standard ``resources/updated`` notification only (no
-    custom dialect). Converts a resource-update signal into the builtin
-    ``mcp_resource_updated`` :class:`HookEvent` (via Phase 1's
-    ``build_hook_payload``, so the field-set stays schema-gated) and delivers
-    it through the shared in-process bridge."""
+    """§6.1 MCP Adapter — converts a resource-update signal (``uri``,
+    ``resync``) into the builtin ``mcp_resource_updated`` :class:`HookEvent`
+    (via Phase 1's ``build_hook_payload``, so the field-set stays
+    schema-gated) and delivers it through the shared in-process bridge.
+
+    #3698 (architect flag, corrected): the SIGNAL is wire-independent, not
+    "standard ``resources/updated`` notification only" as an earlier version
+    of this docstring claimed — :meth:`to_event` takes ``uri``/``resync``
+    values, never a wire notification object, and its ONE caller
+    (:meth:`~reyn.mcp.connection_service.MCPConnectionService.
+    _mcp_to_hook_event`) already fed it from THREE distinct sources even
+    before PR-2: a real legacy ``resources/updated`` push
+    (:meth:`~reyn.mcp.message_handler.ReynMCPMessageHandler.
+    on_resource_updated`), a SYNTHETIC reconnect-resync re-signal (#2597 P1,
+    ``resync=True``, no wire notification at all), and — since #3698 PR-2 —
+    a modern-era ``ResourceUpdated`` event off a ``Client.listen()`` stream
+    (:class:`~reyn.mcp.subscription_port.ListenSubscriptionAdapter`). This
+    adapter itself never branches on which of those produced the signal."""
 
     def __init__(self, *, hook_trigger: "HookTrigger | None", maxsize: int = 32) -> None:
         self._bridge = _BoundedEventBridge(
--- a/src/reyn/mcp/client.py
+++ b/src/reyn/mcp/client.py
@@ -93,62 +93,72 @@ either way).
   ``get_prompt``/``subscribe_resource``/``unsubscribe_resource``), so those
   call sites are unchanged.
 
-  ``mode="legacy"``, NOT the SDK's own ``"auto"`` default — a design
-  correction, not the original plan (architect's own PR-1 spec named
-  ``mode="auto"``; lead-coder overrode it after two live-verified failure
-  symptoms, both traced to the SAME mechanism): ``"auto"`` probes
-  ``server/discover`` and negotiates UP to a modern (2026-07-28-era)
-  protocol version whenever the PEER nominally advertises support for
-  it — which every reyn-owned stdio/http test double DOES, simply by
-  running on the same ``mcp>=2.0`` SDK reyn's client now depends on,
-  REGARDLESS of whether that double's own HANDLERS were ever built for
-  the modern wire's actual mechanisms. Two symptoms, live-verified against
-  the SAME real test doubles, ``mode="legacy"`` vs ``"auto"``, held
-  constant otherwise:
+  ``mode="legacy"`` BY DEFAULT, modern an EXPLICIT PER-SERVER OPT-IN
+  (:func:`_resolve_client_mode`, reading ``config["protocol_mode"]`` —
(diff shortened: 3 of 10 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ 済。★通します（★武装します）。★私の 裁定 3 つ、★条文どおりに 入っています。**

## ✅ ★裁定 (b) の 範囲限定 ── ★逐語で 実装されています

```
✅ ★族単位（tools / prompts / resources）── ★handler 丸ごとの 無効化では ない
   ∴ ★progress（request-scoped）は ★止まりません
✅ 🔴 ★`honored` が ★None（2025 = 不明）── ★`bool(None) == False` ∴ ★何も 抑止しない
   ── ★私が 出した「★不明なら 抑止しない」の ★そのまま
✅ ★legacy 経路は ★set_listen_honored を ★呼ばない ∴ ★従来どおり
```

## ✅ ★裁定 (2)（★死んだ 相手に graceful を 求めない）

```
✅ `await old_adapter.close(★graceful=False)` ＋ ★理由が コメントに
   「★there is no peer left to round-trip a graceful subscriptions/listen close with」
✅ ★consumer task は ★常に cancel（★参照を 落とすだけに していない）
```

## ✅ ★裁定 (c)（★既定 legacy ／ modern は opt-in）

```
✅ `_resolve_client_mode` ── `return "auto" if config.get("protocol_mode") == "auto" else ★"legacy"`
✅ ★module docstring に ★3 状態の 履歴（PR-1 legacy 固定 → PR-2 一時 auto → ★この訂正）
   ── ★「なぜ 今 legacy なのか」が ★次に 触る人に 残ります
```

## ✅ 六問

```
① Tier ── ★2（★era に よらず 購読が 効く、★1 イベント 1 emit）── ★妥当
② 転記 ── ★無し
③ 誰が困る ── ★MCP server を 使う 全員／★dual-firing witness 2 本は
             ★「移行期の server」という ★実在の 相手を 持つ
④ 走らず緑 ── ★該当なし
⑤ 累積 ── ★無し
⑥ 宣言＝実体 ── ★一致
```

## 🔴 ★本文の 測定が この PR の 一番の 資産です

```
★「★legacy 既定に 戻すだけで ★elicitation 10 件・capability-gate 3 件が ★両方 緑に 戻った」
∴ ★#4559（列 A / 列 B）の ★test case が ★13 件 ★既に 手元に 在る
⚪ ★「直した」ではなく ★「戻したら 直った」── ★何が 何に 依存しているかの ★一次証拠
```

PR #4565, opened 2026-08-13
fix(#4564): decouple search_actions visibility from universal_wrappers_enabled

tya5:
**[e2e-coder]** — #4564: `universal_wrappers_enabled: false` silently hid `search_actions` under the `retrieval` scheme too (enumerate-all was never affected — its own `base_tools()` hardcodes `search_actions_visible=False` unconditionally; corrected scope per lead-coder's own follow-up on the issue). Surfaced during #4552 PR-3's pre-work, independently confirmed by lead-coder, ruled by architect.

## Root cause

`embedding.py`'s own docstring declares the contract: "search_actions is gated separately via `embedding.enabled`". But `router_loop.py`'s `run()` computed `_search_visible` (the D14 gate `retrieval`'s presentation reads via `layer_ctx["search_visible"]`) ONLY inside an `if _univ_enabled:` block — an undeclared SECOND gate. The flag's name reads as scoped to `universal-category`, but this indirect path reached `retrieval` too.

Architect's framing: "誰も読んでいない" ≠ "そこに届いていない" — #4231's own read-path trace was accurate for the DIRECT consumer it checked (`_category_exposure.build_category_exposure`) but missed this INDIRECT path through a derived value.

## Fix (architect ruling, option A over "redocument as intentional")

`router_loop.py`: the action-index build+readiness block no longer checks `_univ_enabled` — only the eager/background BUILD calls stay gated (declining to build is a legitimate resource decision). The READINESS check (`_search_visible`) now runs unconditionally, based solely on whether `embedding.enabled` produced a real index/provider/model_class. `sync_repo_ingest_background` (a separate feature, FP-0066 P3b) is left untouched — out of scope.

Cost note (architect's homework, done before choosing): both `eager_embedding_build` and `embedding.enabled` default `False` (separate opt-in getters) — the declared-contract argument is the reason for this fix; narrow exposure is supporting evidence, ordered that way per lead-coder's review so it doesn't read as "safe because nobody's using it yet."

`config_schema.py`: `_check_universal_wrappers_enabled_scheme_mismatch`'s docstring rewritten by DESTINATION (where the flag's effect lands) rather than "who reads it." A False-side check arm was drafted then discarded — post-fix the flag has zero effect on search_actions anywhere, so there's nothing left to statically detect (a static YAML check has no runtime access to witness containment anyway).

## Regression witness

`tests/runtime/test_4564_search_actions_scheme_independent.py` — real `RouterLoop.run()`, `universal_wrappers_enabled=False` + `scheme_name="retrieval"` (chosen after reading both candidate schemes directly — `enumerate-all` structurally cannot witness this claim) + a real embedding build (real `ActionEmbeddingIndex` + `IndexCoordinator` + a real, non-mock, succeeding provider, same convention as `test_index_coordinator_3247_p2b.py`), driven through the real eager-build path rather than a hand-set `is_ready()` (lead-coder review: prefer the actual production build path over a test-only configuration). Asserts `search_actions` IS in `tools=`; the 3 universal-category wrapper functions are NOT.

Strip-falsify verified directly: `git stash`-reverting the fix turns this test RED (the real build never runs pre-fix, caught by the test's own vacuity guard).

## Verification

- `ruff check src tests`: clean.
- `scripts/verify_module_docstrings.py` (2 changed src files): clean.
- `scripts/test_tier_audit.py --strict` (new test file): 1 test, 0 Tier-4 violations.
- `scripts/check_tests_path_literal_reference.py`: 111 refs, all baselined.
- Targeted: 61 + 65 + 127 tests green across the #4231/#1458/scheme/router-loop-adjacent suites.
- Full regression sweep (`tests/config`+`core`+`dev`+`llm`+`runtime`+`tools`): **6370 passed, 6 skipped**. 6 failures present, all confirmed pre-existing (same 6 as #4552 PR-1/PR-2's own verification — PIL/darwin-dependent, unrelated).

Not part of the #4552 arc numbering — a real, independent defect #4552's own pre-work surfaced.

## Test plan

- [x] Confirmed enumerate-all cannot witness the containment claim (read `capability_visibility.py`'s `base_tools()` directly) before choosing `retrieval` for the test.
- [x] Strip-falsified the fix via `git stash` — test goes RED, caught by its own vacuity guard.
- [x] Ran a full regression sweep — all failures confirmed pre-existing.
- [ ] (Visual) — n/a, no UI change.

### DIFF
--- a/src/reyn/config/config_schema.py
+++ b/src/reyn/config/config_schema.py
@@ -526,23 +526,43 @@ class DisabledByHint:
 def _check_universal_wrappers_enabled_scheme_mismatch(
     raw: dict,
 ) -> "DisabledByHint | None":
-    """#4231: ``action_retrieval.universal_wrappers_enabled`` only affects
-    the ``universal-category`` ``tool_use.scheme`` — the OTHER schemes
-    (``enumerate-all`` the #1657 owner default, and ``retrieval``) have
-    their own, mutually-exclusive presentation mechanism and never read
-    it (``router_loop.py``'s ``base_tools()``: "the universal wrappers
-    OFF... enumerate-all adds catalog_entries ON TOP OF the wrappers,
-    INSTEAD").
+    """#4231: does ``action_retrieval.universal_wrappers_enabled: true``
+    reach anything a non-``universal-category`` ``tool_use.scheme`` would
+    render? Answered by DESTINATION, not by "who reads the flag" (#4564's
+    lesson: a discriminator grounded in the reader missed a real effect
+    the flag had via a DERIVED value, not a direct read — see below).
 
-    Re-measured before writing this check (architect's own explicit
-    caveat on the #4231 ruling — "re-measure before implementing, don't
-    build a mechanism on an assumption"): ``universal_wrappers_enabled``
-    DOES have a real read path — ``RouterHostAdapter.
-    get_universal_wrappers_enabled()`` returns the live config value,
-    consumed by ``_category_exposure.build_category_exposure`` (imported
-    by ``universal_category.py``'s scheme, and nowhere else) to compute
-    real ``sp_facts`` differences (``search_actions_enabled``'s formula
-    changes under it). Confirmed via a full read-path trace, not assumed.
+    Where the flag's effect actually lands, as of #4564:
+
+    - The ``universal-category`` scheme's OWN 3 non-search wrapper
+      functions (``list_actions`` / ``describe_action`` / ``invoke_action``)
+      — real, scheme-scoped, confirmed via ``RouterHostAdapter.
+      get_universal_wrappers_enabled()`` → ``_category_exposure.
+      build_category_exposure`` (imported by ``universal_category.py``'s
+      scheme, and nowhere else).
+    - That SAME scheme's own ``search_actions_enabled`` SP-text claim
+      (``_category_exposure.py``: ``sv if univ else True``) — also real,
+      also scheme-scoped.
+    - Nothing else. Before #4564, the flag ALSO gated whether
+      ``search_actions`` was VISIBLE AT ALL, for every scheme — not
+      through a read of ``get_universal_wrappers_enabled()`` inside any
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ 済。★通します（★武装します）。★containment が ★両側 在ります。**

## ✅ ★witness の 形

```
✅ ★search_actions が ★出る こと ── ★宣言（embedding.py:313）が ★真に なった 証明
✅ ★3 wrapper が ★出ない こと ── ★containment（★flag=false で 消えるのは ★これだけ）
✅ ★vacuity guard 2 つ ── ★`provider.calls` が 非空 ／ ★`idx.is_ready()`
   ── ★索引が ★実際に 建った ことを 先に 見て から assert
   ── ★pre-fix では ★build が 走らない ∴ ★containment に 届く前に 落ちる
✅ ★scheme は ★retrieval ── ★enumerate-all では 証明できない（★あなたが 読んで 判断）
```

## ✅ 六問

```
① Tier ── ★2（★宣言された gate と 実挙動の 一致）── ★妥当
② 転記 ── ★無し
③ 誰が困る ── ★embedding.enabled を 立てた operator（★wrapper を 使わない 構成）
④ 走らず緑 ── ★該当なし（★vacuity guard で ★空振りが 落ちる）
⑤ 累積 ── ★無し
⑥ 宣言＝実体 ── ★一致
```

## ✅ ★範囲の 切り方

```
✅ ★repo-doc / repo-src ingest は ★_univ_enabled のまま ── ★理由が コメントに
   「★別の 機能（FP-0066 P3b）で ★自分の 宣言を 持ち、
     ★embedding.enabled-only と 主張した ことが ない」
⚪ ★「ついでに 直す」を していない ── ★#4564 は ★search_actions の 宣言の 話
```
⚪ **★私が #4564 本文に 書いた 範囲は ★誤り**でした（★enumerate-all は 元から 出さない）
—— **★あなたが ★コードを 読んで 切り替えた のが 正しく、★issue にも 訂正済です。**

PR #4568, opened 2026-08-13
docs(#4552): reyn-yaml.md still documented action_retrieval.mode after #4563 removed it

tya5:
**[docs-maintainer]** — **Correction (lead-coder's own catch, git-blame confirmed): this is a REVERT, not drift.** My original framing below was wrong about the cause — the fix itself is unchanged and correct.

## What actually happened (not what I originally wrote)

`#4563` (`#4552` PR-2, merged) removed `ActionRetrievalConfig.mode` and correctly updated `reyn-yaml.md`'s own YAML example + field table in the SAME PR — no drift ever existed at that point. `#4566` (status bar redesign, an unrelated TUI feature) branched off a base predating `#4563` and squash-merged 10 files; one of them silently carried `reyn-yaml.md` back to its pre-`#4563` state, reverting the `mode`-field removal along with legitimately landing `#4566`'s own TUI changes. `e2e-coder` independently found this (mid-PR-3 doc work) and traced it to commit `7a9f5b8c5` via `git blame`; lead-coder confirmed by auditing the full `#4566` squash diff — only this one line in `reyn-yaml.md` was clobbered (the `.ja.md` sibling and every TUI change from `#4566` survived intact).

I originally diagnosed this as ordinary same-night doc drift (a fix that was simply never made) — wrong causal read; the fix WAS made, then silently undone by a stale-base merge. The distinction matters operationally: drift means "finish the missed edit," a revert means "audit what else the same merge silently undid" — lead-coder did that audit and confirmed the blast radius was this one line only.

## The fix (unchanged)

Removed the stale example line + field-table row. Added a retirement note mirroring the existing hot-list retirement note's own shape (unknown-key tolerance on a `reyn.yaml` still carrying the key, no migration path since there's nothing to migrate to).

## Verification

- `mkdocs build --strict -f .mkdocs/mkdocs.yml` — no error/Aborted line.
- Confirmed via `src/reyn/config/embedding.py`'s `ActionRetrievalConfig` that `mode` genuinely has no field before writing the retirement claim.
- lead-coder independently audited `#4566`'s full squash diff (10 files) — confirmed no other reverted content beyond this one `reyn-yaml.md` line.

part of #4552

## Test plan

- [x] `mkdocs build --strict` — clean, no Aborted/error line (read directly)
- [x] Confirmed `ActionRetrievalConfig` has no `mode` field in current `main` before writing the retirement claim
- [x] Confirmed the ja sibling was already correct (no fix needed there)
[...]

PR #4566, opened 2026-08-13
feat(#4542): status bar redesign — Navigation / Telemetry, no │ separators

tya5:
**[tui-coder]** — part of #4542 (status bar redesign: left=Navigation / right=Telemetry, no `│`/`|` separators). This is the text-format + CSS-styling increment; left/right layout mechanics (a stretching spacer, a dedicated two-region layout) are a follow-up — see Scope note below.

## Owner's 3 architectural decisions (taken directly, this conversation — recorded verbatim in the [#4542 issue comment thread](https://github.com/tya5/reyn/issues/4542#issuecomment-5279633444) since conversation is not a durable surface)

Premise-measurement against the current code (chrome.py, #3338) surfaced that 3 parts of the original relayed proposal conflict with existing guarantees or require new mechanisms not yet built:

1. **Narrow-width model**: keep #3338's wrap-to-2nd-row behavior — nothing dropped from Navigation. The proposal's "drop low-priority items" would have been a genuinely new, unguarded behavior (no drop/priority mechanism exists anywhere in chrome.py today).
2. **Cost/Ctx**: stay as Navigation tabs, unlike the proposal's "excluded from Navigation" — their detail-breakdown drawer pane currently has no other entry point, and the owner's direct answer keeps them where that entry point already exists.
3. **Menu/Help**: compact glyphs (`…`/`?`), not a command palette — no palette surface exists anywhere in this codebase (confirmed repo-wide), and building one is separate scope from a status-bar layout change.

## What changed

- **`status_line_text()` (chrome.py)** — the labelled `model X │ agent Y │ cost Z │ ctx W` format is replaced by an unlabelled one: `model · agent    $cost  ctx%`, no `│`/`|` anywhere. Position and text-style carry the grouping instead of labels. Context-usage percent escalates from bare (`16%`) to labelled (`ctx 82%`) only at/past a warn threshold.
- **`_MENU_TABS`** — `Menu`/`Help` labels shrink to `…`/`?`. Same tab, same Enter-opens-drawer mechanism (`MenuBar._move`/`_on_key`) — only the visible label changes.
[...]

Issue #4542, opened 2026-08-13
feat(tui): status bar redesign — left=navigation / right=telemetry, no │ separators


tya5:
**[tui-coder]** — owner proposal, relayed verbatim from a conversation-only source (not previously on any durable surface — filing per CLAUDE.md's arc-closure discipline: a decision that lives only in conversation text is a decayed intent, not a decision).

## Context

Superseded PR #4540 (closed, not merged) took the opposite approach — adding a `│` separator between the tab row and the status-values segment when they share a row. The owner's proposal below explicitly rejects `|`/`│` separators entirely in favor of a structural left/right split, so #4540's direction is now obsolete.

## Owner's proposal (verbatim)

> tui status bar について gpt から提案ありました。
>
> ## 目的
>
> 入力ボックス下のバーを、現在の「タブ名 + `|` 区切りの状態文字列」から、**左=操作、右=状態**が一目で分かる構造へ整理する。
>
> ## 最終案
>
> ```text
>  Model  Agent  Hist  Art    Tool  MCP  Skill  Pipe  Hook  Cron                 standard · default       $3.60  16%
> ```
>
> ### 左側: Navigation
>
> ```text
> Model  Agent  Hist  Art    Tool  MCP  Skill  Pipe  Hook  Cron
> ```
>
> * ユーザーが操作する項目だけを配置する。
> * 選択中の項目のみ、現在のテーマに沿った強調表示を行う。
> * グループ間は区切り文字ではなく、少し広めの空白で分ける。
> * `Cost` / `Ctx` は navigation から外す。
> * `Menu` / `Help` は必要なら `…` / `?` として右端近くに配置するか、command palette 側へ統合する。
>
> ### 右側: Telemetry
>
> ```text
> standard · default       $3.60  16%
> ```
>
> 意味は以下。
>
> ```text
> standard   = model
> default    = agent
> $3.60      = accumulated cost
> 16%        = context usage
> ```
>
> 通常時は `model` / `agent` / `cost` / `ctx` のラベルを表示しない。位置と表示スタイルで意味を固定する。
>
> `|` 区切りは使用しない。
>
> ### 表示階層
>
> Navigation と Telemetry は同じ行に置くが、視覚的には明確に分離する。
>
> * Navigation: 通常の文字強度
> * Telemetry: 1段階弱い文字強度
> * Navigation と Telemetry の間は残り幅を空白で埋める
> * box / border / separator は追加しない
>
> ## Warning 表示
>
[...]

PR #4569, opened 2026-08-13
feat(#4542): Navigation/Telemetry visual separation — stretching spacer

tya5:
**[tui-coder]** — part of #4542 (2nd increment: Navigation/Telemetry visual separation). Builds on #4566 (merged — text-format + CSS-styling layer). This lands the layout mechanics: a real left/right split with a stretching center spacer, per the owner's own proposal ("[Navigation ...][Telemetry]").

## Two distinct mechanisms — found live, not assumed

The status bar's two narrow-width shapes (StatusLine merged onto the last tab row, vs. StatusLine getting its own fallback row) needed DIFFERENT fixes, discovered by testing both rather than assuming one mechanism covers both:

1. **Merged case** (StatusLine shares the last tab row): a new empty `.menu-spacer` widget (`width: 1fr`) sits between the tabs and StatusLine (`-shared` class, `width: auto`). The spacer consumes whatever slack space is left, pinning Telemetry to the row's right edge with a real, visually expanding gap. Live-verified at 130 columns (this app's own 14-tab + status-text merge threshold, found by probing widths — not assumed): gap=6 between the last tab and StatusLine, StatusLine's right edge within 1 cell of the screen's own edge.

2. **Own-row fallback** (no room to merge): **no spacer here** — a real bug in my first draft. I initially gave this case a spacer too, and it silently did nothing: StatusLine's base CSS rule has no explicit width, so it already claims 100% of the row on its own — the spacer had no slack to expand into, and the widget order actually put StatusLine right after a near-zero-width spacer at the row's LEFT edge, not the right (caught by the falsified test itself — see below). This 100%-width behavior is deliberate, load-bearing, pre-existing behavior (see `StatusLine.-shared`'s own comment): an auto-width own-row status line can overflow past the screen edge for a long status string. Fixed instead with `text-align: right` on StatusLine's base CSS rule — pins the TEXT to the right edge without touching the width/containment mechanism that protects against overflow.

## Test changes (falsify-verified, not just "written and passed")

Two new real-TTY-witnessed geometry tests, mirroring #3338's own `test_every_menu_tab_is_fully_on_screen` convention:

- `test_telemetry_pins_to_the_right_edge_with_a_gap_from_navigation` — the merged case. **The first draft of this test picked 100 columns without checking and silently exercised the WRONG code path** (the own-row fallback, where there's no spacer at all) — it failed with a genuinely confusing result (`gap=-77`) that led directly to finding bug #2 above. Fixed by probing actual widths and using the live-verified 130-column merge threshold, plus an explicit `assert "-shared" in status.classes` guard so a future width/threshold change can't silently make this test exercise the wrong path again without failing loudly.
- `test_telemetry_own_row_stays_full_width_not_shrunk_by_a_spacer` — the own-row fallback, asserting the containment fix directly (StatusLine still claims ~full row width) rather than only testing the merged case and hoping the other path is fine.

## Test plan

- [x] `python scripts/test_tier_audit.py --strict` on the changed test file — 30/30 OK, 0 Tier-4 violations
- [x] `python scripts/mypy_ratchet.py` — OK (211 findings, all baselined)
- [x] `python scripts/check_tests_path_literal_reference.py` — OK (111 references, all baselined)
- [x] Full chrome/status/colour/config/phase3 sweep (`tests/interfaces/ -k "chrome or menubar or ... or phase3"`) — 269 passed
- [ ] Visual — standing waiver applies (owner ruling, 2026-08-08); owner confirms on `main` after.

Branch confirmed based on current `origin/main` (post-#4568) before push — the merge-base-silent-revert lesson from tonight's #4566 landing is fresh.

Touches `tests/` — waiting for TESTS-READ before merge.

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -858,6 +858,16 @@ class TextualChatApp(App):
         height: 1;
         text-style: @telemetry@;
         padding: 0 1;
+        /* #4542: pins the text to the row's RIGHT edge without touching
+           width — in the ``.-shared`` (merged) case the box is already
+           auto-tight to its own text, so this is a no-op there; in the
+           own-row fallback below, width stays 100% (load-bearing — see
+           that rule's own comment on why an auto-width own-row status
+           line can overflow), so text-align is what achieves "pinned
+           right" there instead of a spacer, which would need width: auto
+           to have anything to push against and would reopen the overflow
+           this rule protects against. */
+        text-align: right;
     }
     /* #4194: bold, not a colour — palette.py's own rule (@attention@ is the
        ONLY semantic colour this CUI claims, for the intervention panel;
--- a/src/reyn/interfaces/inline/textual_chat/chrome.py
+++ b/src/reyn/interfaces/inline/textual_chat/chrome.py
@@ -1778,6 +1778,19 @@ class MenuBar(Widget, can_focus=True):
         width: 100%;
         layout: horizontal;
     }
+    /* #4542: the stretching gap between Navigation and Telemetry — a plain
+       empty widget, no border/content (the owner's own non-goal: "box や
+       separator を追加しない"). 1fr consumes whatever width the row's
+       natural-width children (tabs + StatusLine) don't need, so Telemetry
+       sits at the row's right edge on any width wide enough to merge at
+       all — see status_fits_last_row's own docstring for the FIT decision
+       this doesn't change (the spacer only expands into slack space that
+       already existed; it never causes a row that used to fit to stop
+       fitting). */
+    MenuBar .menu-spacer {
+        width: 1fr;
+        height: 1;
+    }
     """
 
     class Selected(Message):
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ 済。★通します（★武装します）。★2 経路とも ★経路そのものを 先に 確かめてから 見ています。**

## ✅ ★今回 一番 良い 点 ── ★「どの経路を 試しているか」を assert している

```
✅ ★merged 側 ── `assert "-shared" in status.classes`
✅ ★own-row 側 ── `assert "-shared" not in status.classes`
∴ ★幅の 取り方 次第で ★別経路を 検証して しまう、を ★構造的に 防いでいます
🔴 ★これは ★あなた自身が ★gap=-77 という 結果から 気づいた もの
   ── ★「テストが 実測で 自分の 間違いを 検出した」── ★今夜 何度も 求めてきた 形
```

## ✅ ★assert が ★実際の 描画位置

```
✅ ★右端 ── `screen.width - status.region.right <= 1`
✅ ★隙間 ── `gap = status.region.x - last_tab.region.right` ／ `gap > _TAB_H_PADDING`
   ── ★閾値が ★定数から 導出（★数字を 手で 書いていない）
✅ ★own-row ── `status.region.width >= parent.size.width - 2`
⚪ ★CSS 文字列の grep では なく ★解決後の region ── ★#4544 / #4554 の 教訓の 継続
```

## ✅ 六問

```
① Tier ── ★2（★Navigation と Telemetry が ★視覚的に 分離される）── ★妥当
② 転記 ── ★無し
③ 誰が困る ── ★owner（★この 再設計の 発注者）／★2 経路 それぞれに 消費者が 居ます
④ 走らず緑 ── ★該当なし（★経路 assert が ★空振りを 落とす）
⑤ 累積 ── ★無し
⑥ 宣言＝実体 ── ★一致
```

## ⚪ ★base の 鮮度

```
✅ ★push 直前に `git merge-base --is-ancestor origin/main HEAD` で 確認済
   ── ★#4568 込みの main を base に して います
⚪ ★#4566 で 私が 見落とした もの ── ★同じ夜に ★手順として 定着しました
```

PR #4592, opened 2026-08-13
fix(#4584): move artifact_refs.jsonl + spill manifest from cache/ (derived) to persist tier

tya5:
**[tui-coder]** — part of #4584.

## What

Both `.reyn/cache/artifact_refs.jsonl` (#4482 PR-1) and
`.reyn/cache/tool_result_spills.jsonl` (#4381/#4432) move to
`.reyn/memory/` (PERSIST tier). Both were filed under `cache/`, documented
"DERIVED — rebuilt after restore" (`reyn-dir-layout.md`) — false for both:
each mapping/entry exists ONLY at the instant it's written, and nothing
else durable carries it. An operator correctly following the doc's own
"safe to clean up" classification (deleting `.reyn/cache/`, or restoring
from a backup that excludes it) permanently kills every past `/open` and
orphans every spill reference.

Full design record (lead-coder/architect co-vet) is on the issue. Summary
of the decisions that shaped this PR:

- **Both move together** — moving one alone would be a double standard
  the doc's own classification rule doesn't support (both tables hit the
  identical shape: neither is reconstructable, both were self-acknowledged
  as such in their own prior comments while still filed under `cache/`).
- **Landing spot is `.reyn/memory/`**, per `reyn-dir-layout.md`'s own
  decision rule ("must SURVIVE rewind → `memory/`") — the only tier its
  own text points at. `memory` is an imperfect NAME fit for a ref/manifest
  table (recorded in both modules' own docstrings, not silently accepted)
  — a sibling tier was judged out of scope for this PR.
- 🔴 **This move does NOT change write permissions.**
  `security/permissions/file_scope.py`'s `ZoneStateDir` write-gates only
  `.reyn/config/` + `.reyn/state/` + `approvals.yaml` — `memory/` is an
  ordinary agent-writable zone, identically to `cache/`. What changes is
  the CLASSIFICATION and the "safe to delete" claim, not the write
  surface. Stated explicitly in both modules' own docstrings, per
  lead-coder's explicit request, so a reader doesn't read "persist" as
  "hardened".

Two corrections that surfaced during the co-vet's own falsification round:

- **media_store.py's "rebuilt/pruned" comment was half-false.** Only
  PRUNING (dropping an existing manifest entry whose target file vanished)
  is real; there is no REBUILD mechanism (recreating the manifest from
  nothing after the manifest file itself is deleted). Corrected in both
  the module docstring and the two method docstrings that touch it —
  pruning and rebuilding are not the same claim, and the prior comment
  conflated them.
- **`reyn-dir-layout.md`'s "safe to delete at any time" was too broad.**
  It sits right under the `cache/` header, readable as a blanket claim
  about the whole tier. Architect measured a positive control: `index_drop`
  (an agent-callable op) already deletes `cache/index/<source>` behind
  `require_file_write` (#1199 S3.4) — reyn's own code already treats
  deleting something under `cache/` as gated, permission-requiring, not
  free. Narrowed the phrase to scope explicitly to `budget_checkpoint.json`
  (the one entry that genuinely is reconstructable) and added a `cache/`-level
  caveat naming `index_drop` as the counter-evidence to a blanket reading.

## Tests

`test_4584_persist_tier_survives_wal_truncation.py` (new) — the CLAUDE.md
recovery-feature PR gate's real intent for this shape: there is no "X's
source events" to truncate past for either table (neither was ever
WAL-derived — that's the whole defect), so the falsification instead runs
a real `StateLog.truncate_below` past every WAL event each test wrote,
re-opens a fresh `StateLog` (the same "reconstruct" step
`test_2946_item1_state_log_tail_scan.py` uses for the identical
primitive), and confirms BOTH tables' own round trip (`mint_ref` →
`resolve_ref`; `save_tool_result` → `is_tool_result_spill`) still returns
the SAME path throughout — "moved" and "moved somewhere that survives"
are different claims; only a real round trip through the actual persisted
files proves the second one (a bare path-string diff would go green
either way). A companion test pins the structural reason why: neither
path falls under `.reyn/state/`, `.reyn/config/`, or `.reyn/cache/`
anymore.

Existing `artifact_ref`/`media_store`/web tests updated for the new
paths (`test_4482_artifact_ref.py`, `test_4482_get_artifact_by_ref.py`,
`test_4478_media_storage_stats.py`).

## Test plan

- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 27/27 OK, 0 errors
- [x] Scoped `pytest tests/data -k "media or artifact" tests/web -k "artifact" tests/core -k "spill or media_store"` — 69 passed
- [x] Scoped `pytest tests/runtime/test_2425_step1c_chat_chokepoint.py tests/runtime/test_media_cap_272.py tests/runtime/test_router_ctx_media_store_2409.py tests/runtime/test_path_ref_materialisation.py` — 26 passed (broader MediaStore-consumer regression check)

Enumerated `part of #4584` mentions before writing this body: this is the
only open PR against #4584; no other PR claims it.

### DIFF
--- a/docs/reference/cli/storage.md
+++ b/docs/reference/cli/storage.md
@@ -77,12 +77,16 @@ with no `.reyn/agents/` yet reports all-zero for the `history.jsonl` row, not an
 ## Spill-manifest self-prune (#4478)
 
 `MediaStore` tracks which tool-result files it has spilled to disk via a manifest at
-`.reyn/cache/tool_result_spills.jsonl`, read in full on every `MediaStore`
-construction. An entry whose target file no longer exists on disk (deleted manually,
-or by a future Phase 2 GC policy) is dropped from the manifest the next time it's
-loaded — this bounds the manifest's otherwise-unbounded growth. The prune only
-rewrites the manifest itself; it never deletes any actual media/tool-result bytes,
-and a write failure here is best-effort (never fails `MediaStore` construction).
+`.reyn/memory/tool_result_spills.jsonl` (#4584: moved from `.reyn/cache/` — that
+tier's "derived, rebuilt after restore" promise never held for this manifest; see
+[`.reyn/` directory layout](../runtime/reyn-dir-layout.md)), read in full on every
+`MediaStore` construction. An entry whose target file no longer exists on disk
+(deleted manually, or by a future Phase 2 GC policy) is dropped from the manifest
+the next time it's loaded — this bounds the manifest's otherwise-unbounded growth.
+This is a self-PRUNE of existing entries only, never a REBUILD: if the manifest file
+itself were deleted, nothing recreates it. The prune only rewrites the manifest
+itself; it never deletes any actual media/tool-result bytes, and a write failure
+here is best-effort (never fails `MediaStore` construction).
 `reyn storage stats` does not report on the manifest directly — it measures the
 artifact directories the manifest tracks.
 
--- a/docs/reference/runtime/reyn-dir-layout.md
+++ b/docs/reference/runtime/reyn-dir-layout.md
@@ -54,40 +54,61 @@ Everything else is excluded, by one of four reasons:
 │                           `external_transports` config is read from reyn.yaml's own
 │                           `external_transports:` section only
 ├── memory/                 PERSIST — agent knowledge; survives rewind, never reverted
+│   ├── artifact_refs.jsonl (agent, path) → ref table for `present`'s "artifact"
+│   │                       component (#4482 PR-1, moved from `cache/` by #4584):
+│   │                       the mapping exists ONLY at mint time — no WAL event,
+│   │                       no conversation-log entry, nothing else durable
+│   │                       carries it — so it is PRIMARY data, not derived,
+│   │                       however small. Ordinary writable zone (not
+│   │                       write-gated, same as the rest of `memory/`) — a
+│   │                       plain file write, never a dedicated op. "memory" is
+│   │                       an imperfect NAME fit (a ref→path table is neither
+│   │                       knowledge nor a decision) — the doc's own tier
+│   │                       decision rule (below) only points at one place.
+│   └── tool_result_spills.jsonl  persisted provenance of tool-result SPILL
+│                           artifacts under `tool-results/` (#4381/#4432, moved
+│                           from `cache/` by #4584 — same reasoning as
+│                           `artifact_refs.jsonl` above: not literally
+│                           rebuildable, an earlier comment on this file said
+│                           so explicitly while still filing it under
+│                           `cache/`) — one JSON line per `MediaStore.
+│                           save_tool_result` write, read back at every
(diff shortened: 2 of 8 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ（★diff の test を 読みました）── ★blocking なし。**

|---|---|---|---|---|---|---|
| `falsify_artifact_ref_survives_a_real_wal_truncation_and_reconstruct` | 2 | ✗ | ★rewind 後に `/open` する operator | ✗ | WAL 5 件 | ✅ |
| `falsify_spill_manifest_survives_a_real_wal_truncation_and_reconstruct` | 2 | ✗ | ★spill 判定に 頼る 消費者 | ✗ | 同上 | ✅ |
| `both_tables_live_directly_under_memory_never_under_cache_or_state` | 2 | ⚠️ ★下記 | ★tier を 決める 次の 実装者 | ✗ | — | ✅ |

## ✅ ★recovery gate ── ★満たしています

```
✅ ★CLAUDE.md の 要求 ──「★set X → ★truncate past X's events → ★reconstruct → ★assert X survives」
✅ ★実際の 手順（★逐語、★docstring より）
   ★real StateLog ／ ★truncate_below(1_000_000)（★全件の 下）／
   ★fresh StateLog の 再 open（★別プロセスが 同じ project に 付く 相当）／
   ★mint/save → ★resolve/is_spill が ★同じ path を 返す
✅ ★docstring が ★私が 見たかった 区別を ★自分で 書いています:
  「"moved" and "moved somewhere that survives" are ★different claims …
    ★a bare path-string diff would go green ★either way」
```

## ⚠️ ★③ に ついて（★非ブロッキング）

```
⚠️ ★path を test 側でも 組み立てる ∴ ★形としては ★実装の 書き写しに 近い
✅ ★ただし ★この PR の 主張 そのものが ★「どの tier に 置くか」です
   ── ★契約を 直接 見る test は ★この 場合 正当
✅ ★荷重は ★上 2 本が 持っています ── ★③ が 単独で 立っているのでは ありません
```

## ✅ ★doc（★受入 ③⑤）── ★確認しました

```
✅ ★「safe to delete at any time」の 射程を ★budget_checkpoint.json に 限定
✅ ★cache/ ヘッダに ★index_drop（★agent 呼び出しで cache/index を 消す）を 注記
   ── ★architect の 陽性対照が ★doc に 着地しています
✅ ★media_store.py の「rebuilt/pruned」── ★prune は 実在／★rebuild は 非実在 と 分離
   ── ★旧文が ★2 つを 混同していた、という 訂正の 形
```

## ✅ ★PR 本文の ★権限に ついての 明記 ── ★確認

```
✅ ★「memory/ へ 移しても ★書き込み権限は 変わらない（ZoneStateDir は
   config/ ／ state/ ／ approvals.yaml のみ）」
⚪ ★これが 無いと ★「硬くなった」と 読まれます ── ★直ったのは ★分類と 文面 です
```

⚪ **★CI 緑を 待って merge します。**

PR #4607, opened 2026-08-13
feat(#4570): Agent Plugins 1.0 conversion C1 — mcp.json rename + $schema + type, transport adapter

tya5:
**[tui-coder]** — part of #4570 (Agent Plugins 1.0 alignment). Stacked on #4603 (conversion B, not yet merged) — this PR's diff shrinks to C1-only once #4603 merges into `main`.

## What — conversion C1 (split from the original "C")

Lead-coder approved splitting the original "C" plan: the transport-type rename (reyn's own `"http"` → `"streamable-http"`, `mcp/client.py`'s `_SUPPORTED_TYPES`) is repo-wide and unrelated to the plugin subsystem (18 `src` files + 9 test files outside it) — filed separately as **#4604** (conversion C2). This PR is the **plugin-scoped** half: filename rename, required `$schema`, and a translation adapter at the one plugin-ingestion boundary.

Canonical `mcp.schema.json` fetched directly (not assumed) — `$id: https://agent-plugins.org/schemas/1.0.0/mcp.schema.json`: root requires `$schema` + `mcpServers`, `additionalProperties: false`; each server entry requires `type` (`stdio`/`streamable-http`/`sse`), `additionalProperties: false`; stdio `env` forbids `PLUGIN_ROOT`/`PLUGIN_DATA` keys; `cwd` must start with `./`, `${PLUGIN_ROOT}/`, or `${PLUGIN_DATA}/` (the last two token names are conversion D's job, not touched here).

## Changes

- `.mcp.json` → `mcp.json` everywhere in the plugin subsystem (`capability_kinds_present`, `_expand_plugin_files`, `_register_mcp`) and the shipped `rag` plugin's own file (git mv + `$schema` + `"type": "stdio"` on both server entries).
- **`_build_mcp_entries` gains the ONE translation point** (lead-coder's condition ①, comment cites #4604): a url-bearing entry's spec-legal `"streamable-http"` becomes reyn-internal `"http"`; `"sse"` passes through **unchanged** — a distinct value in both vocabularies, mapping it too would be a real bug, not an incomplete rename (lead-coder's own review point, verbatim: "`sse` が `http` に巻き込まれないことを test 1 本で押さえてください").
[...]

PR #4609, opened 2026-08-13
feat(#4570): Agent Plugins 1.0 conversion D — field-aware ${PLUGIN_ROOT}/${PLUGIN_DATA} expansion

tya5:
**[tui-coder]** — part of #4570 (Agent Plugins 1.0 alignment). Builds on #4607 (conversion C1, merged).

## What — conversion D

Token expansion into `mcp.json` used to be a blind full-text replace across the **whole file** (`_bake_all_tokens`), including inside `command` and `url` — architect's #4570 finding: reyn's pre-conversion behavior was **more permissive** than the standard it is aligning with, not merely differently-spelled. The canonical `mcp.schema.json` expands `${PLUGIN_ROOT}`/`${PLUGIN_DATA}` **only** in `args`/`env`/`cwd`, never `command`/`url` — the injection boundary: a plugin-authored string must not be able to choose WHAT gets executed or WHERE a request goes via a token whose value this module resolves.

## Changes

- `src/reyn/core/op_runtime/plugin_install.py`: new `_bake_mcp_json_fields` (parses `mcp.json` as JSON, not text — a token appearing inside `command`/`url` is left LITERAL, never a coincidental string match away from expansion). Wired into `_expand_plugin_files` in place of the old full-text bake for `mcp.json` specifically; `pipelines/*.yaml` keeps its unchanged full-text `${REYN_*}` bake (a reyn extension the standard doesn't define — deliberately untouched).
- **`${PLUGIN_DATA}`** (lead-coder ruling): resolves to a **new** directory, `~/.reyn/plugin-data/<name>/` — new `plugin_data_root()`, a **sibling** of `plugins_root()`, not a subdirectory. Deliberate: step 5's atomic `.staging` swap replaces `plugin_root` wholesale on every install/update, so anything a plugin needs to survive a reinstall cannot live inside it. Created eagerly on every bake.
- `src/reyn/core/op_runtime/plugin_uninstall.py`: **never** deletes plugin-data (data outliving code is the safe direction — the reverse is unrecoverable), and never silently leaves that unmentioned — the result dict gains `plugin_data_retained_at` when a data dir exists for the uninstalled name.
- `src/reyn/builtin/plugins/rag/mcp.json`: `${REYN_PLUGIN_ROOT}` → `${PLUGIN_ROOT}` in both servers' `args`. **Not cosmetic** — without it, D's own field-aware bake (which recognizes `${PLUGIN_ROOT}`, not the old reyn-native name) would silently leave the real shipped plugin's args unexpanded, breaking the chunker/vector-store spawn path. Caught and fixed before opening this PR; verified via the real end-to-end witnesses (`test_rag_plugin_mcp_json_disables_fastmcp_telemetry_env`, `test_fp0063_arc_witness.py`) both still passing.

## Test plan

10 new Tier 1/2 tests in `test_plugin_install.py`, the **required** falsify pair first (lead-coder's condition): `${PLUGIN_ROOT}` expands in `args`/`env`/`cwd` but is left literal in `command`; both halves asserted together so an implementation expanding nowhere fails half 1, one expanding everywhere (the pre-conversion-D behavior) fails half 2. Plus: `url` stays literal too, `PLUGIN_DATA` expands + creates the directory, a no-`mcp.json` plugin is a graceful no-op, and the `plugin_uninstall` retention+disclosure contract (both the positive case and its accept-side "no data, no false claim" sibling).

`ruff check src tests`, `test_tier_audit.py --strict` (28 tests), `verify_module_docstrings.py`, `mypy_ratchet.py`, `check_tests_path_literal_reference.py` — all green. Scoped `pytest`: 54 passed, 9 skipped (Landlock-only) across every touched plugin-related file plus the seccomp/arc-witness real-file end-to-end witnesses.

- [ ] Visual (waived per CLAUDE.md standing waiver, 2026-08-08 — not a TUI-facing change)

Next: E (full rag-plugin conversion — mostly already done incrementally across A–D; E's remaining scope is a final consistency pass + closing #4570).

### DIFF
--- a/src/reyn/builtin/plugins/rag/mcp.json
+++ b/src/reyn/builtin/plugins/rag/mcp.json
@@ -4,7 +4,7 @@
     "reyn_chunker": {
       "type": "stdio",
       "command": "python",
-      "args": ["${REYN_PLUGIN_ROOT}/scripts/chunker_server.py"],
+      "args": ["${PLUGIN_ROOT}/scripts/chunker_server.py"],
       "env": {
         "FASTMCP_SHOW_SERVER_BANNER": "false",
         "FASTMCP_CHECK_FOR_UPDATES": "off"
@@ -13,7 +13,7 @@
     "reyn_vector_store": {
       "type": "stdio",
       "command": "python",
-      "args": ["${REYN_PLUGIN_ROOT}/scripts/vector_store_server.py"],
+      "args": ["${PLUGIN_ROOT}/scripts/vector_store_server.py"],
       "env": {
         "FASTMCP_SHOW_SERVER_BANNER": "false",
         "FASTMCP_CHECK_FOR_UPDATES": "off"
--- a/src/reyn/core/op_runtime/plugin_install.py
+++ b/src/reyn/core/op_runtime/plugin_install.py
@@ -159,7 +159,7 @@ from reyn.plugins.manifest import (
     load_plugin_manifest,
 )
 from reyn.plugins.source import resolve_name_collision
-from reyn.plugins.tokens import PluginTokenContext, expand_reyn_tokens
+from reyn.plugins.tokens import PluginTokenContext, expand_reyn_tokens, expand_with_map
 from reyn.schemas.models import PipelineInstallIROp, PluginInstallIROp, SkillInstallIROp
 
 from . import register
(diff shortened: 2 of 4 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ ── ★blocking なし。★私が 求めた falsify 対が ★1 本の 中に 揃っています。**

## ✅ ★field-aware 展開 ── ★対で 成立

```
★`test_bake_mcp_json_fields_expands_args_env_cwd_but_never_command_or_url`
   ✅ ★args ── ★展開される
   ✅ ★env  ── ★展開される
   ✅ ★cwd  ── ★展開される
   🔴 ★command ── ★`"${PLUGIN_ROOT}/bin/srv"` の まま ── ★literal
★`…_url_stays_literal_too` ── ★url も 同じ
∴ ★「どこでも 展開しない」実装でも ★「全部 展開する」実装でも ★落ちます
⚪ ★私が 出した 条件 ──「★片方だけ だと ★どこでも 展開しない 実装でも 通る」──
   ★1 本の 中に ★両側が 在る 形は ★それより 強いです
```

## ✅ ★uninstall ── ★裁定 どおり

```
✅ ★`…never_deletes_plugin_data_and_discloses_where_it_is`
   ── ★消さない ＋ ★どこに 在るかを 言う ── ★2 つとも 1 本で
✅ ★`…omits_the_disclosure_key_when_no_data_exists`（★accept 側）
   ── ★data が 無い とき ★「残っています」と 言わない
   🔴 ★これが 無いと ★「常に 同じ 文言を 出す」実装が 通ります
   ── ★★出す条件 も 契約の うち です
```

## ✅ ★accept 側が ★2 本 在ります

```
✅ ★`…is_a_noop_when_mcp_json_is_absent` ── ★mcp.json が 無い plugin
✅ ★`…omits_the_disclosure_key_when_no_data_exists`
⚪ ★どちらも ★過剰発火を 見ています ── ★「足した 機構が 余計な ところで 動かない」
```

## ⚪ ★base

```
✅ ★baseRefOid = ★19d2e633e = ★現在の origin/main ── ★stale では ありません
⚪ ★#4607 に rebase 済 ∴ ★C1 の 上に 正しく 積まれています
```

⚪ **★CI 緑を 待って merge します。**

PR #4622, opened 2026-08-13
fix(#4620): route webhook_received through the same evidence-based check as mcp_resource_updated

tya5:
**[e2e-coder]** — closes #4620.

## The bug

architect's finding, relayed by lead-coder: #4612 (this session, C-2) and #4618 (same night, #4605) were both correct in isolation but false together. `doctor.py:331` declared `webhook_received` "uncheckable" (no config surface, no audit-log surface) — true when #4612 merged. #4618 gave `webhook_received` its own audit-event kind minutes later, in the same merge session, and nobody re-checked whether that changed #4612's own printed claim. It did: "no config or audit-log surface exists for its producer" went false the moment #4618 landed, and doctor kept printing it — (1) a false reason handed to the operator, (2) a check that became possible (the same zero-consumer evidence-based check `mcp_resource_updated` already had) that nobody wired up.

## Fix

`_UNCHECKABLE_EXTERNAL_POINTS` goes empty (not deleted — a future point with a genuine no-producer-surface gap has somewhere to land). `webhook_received` routes through the generalized `_external_event_kind_seen(events_dir, kind)` (renamed from the `mcp_resource_updated`-only `_mcp_resource_updated_seen` — same windowed scan, parametrized by kind instead of hardcoded).

`cron_fired`/`file_changed` are **untouched** per the architect's own point: both already have a COMPLETE config-declared producer surface (`cron.jobs[]`/`fs_watch.paths`) — moving them to a windowed scan would strictly lose information for no gain.

## Doc-drift (CLAUDE.md hard rule, same PR)

The module docstring, `doctor.md`'s sample output, and `doctor.md`'s C-2 prose section all repeated the same "webhook_received has neither [surface]" claim #4618 falsified. Fixed in this PR, not a follow-up.

## Tests

- 3 new (mirroring `mcp_resource_updated`'s own 3): producer+zero-consumers flagged, producer+consumer OK, evidence-outside-window disclosed not silent.
- 1 updated: the "nothing configured" accept-side test now expects `webhook_received`'s own "?" disclosure line instead of the stale "not checked" claim.
- Strip-falsified: reverted `_UNCHECKABLE_EXTERNAL_POINTS` to the pre-fix value, confirmed exactly the 4 affected tests go RED (the new/updated `webhook_received` assertions — the other 10 unaffected tests stayed green, confirming the failure was specific), restored, confirmed GREEN.

## Verification

- `ruff check src tests`: clean.
[...]

PR #4623, opened 2026-08-13
feat(#4364): doctor C-5 — declared vs. RESOLVED sandbox posture

tya5:
**[tui-coder]** — part of #4364 (C-5 slice; the umbrella issue stays open — C-6 and other slices remain).

## What

architect's C-5 spec, motivated by a real confusion architect hit personally: "reviewer can't write" was mistaken for prose in REYN.md, but the real cause was `sandbox.policy` being UNDECLARED — which resolved to `SeatbeltBackend` / `write_paths=[]` (actually restrictive), not "no restriction" the way an undeclared value reads. Declaration and resolution silently disagreed.

## Design (both confirmed in-thread before implementing)

- **declared** = `sandbox.backend` / `sandbox.on_unsupported` / `sandbox.policy`, read straight off config, no merging.
- **effective** = the ACTUALLY RESOLVED backend (`resolve_backend` — the SAME call C-1's hook probe already makes), never a second doctor-invented probe. "Which backend resolved" already carries the enforcement witness: `get_default_backend()` self-tests a real deny before returning a backend (#2983) and applies `on_unsupported` to one that can't enforce.
- doctor never tries a write of its own (same D-2 reason as C-1).
- doctor has **no op context**, so it does NOT build a `resolve_sandbox_policy()` call (that needs a caller-supplied `write_paths` floor doctor cannot know and must not invent a stand-in for). Only the declared `sandbox.policy` dict's own write-scope keys are echoed.
- an explicitly-forced backend that falls back is flagged `DOWNGRADED`, distinct from auto-selection landing on noop (not a downgrade — nothing was forced).
- `on_unsupported=error` makes resolution itself fail-closed; doctor reports `refuses to run` rather than crashing.

## Changes

- `src/reyn/interfaces/cli/commands/doctor.py`: `_print_sandbox_posture` + `_SANDBOX_POLICY_WRITE_SCOPE_KEYS`, wired into `run()` after C-2. `sandbox.backend`/`sandbox.on_unsupported` added to `_MEASURABLE_LEAF_KEYS` (their live effect is now genuinely measured); `sandbox.policy`/`sandbox.mode` are NOT (declaration only).
- `docs/reference/cli/doctor.md`: new "Sandbox posture (C-5)" section, moved out of "Out of scope."

## Test plan

8 new tests: declared+resolved both print; the exact motivating confusion (no `sandbox.policy` ≠ unrestricted); declared write scope echoed verbatim; accept-side (no write-scope keys says so honestly); D-2 falsify (never writes to the project); a forced-unavailable backend downgrades and is flagged; same case with `on_unsupported=error` reports a refusal; accept-side (auto→noop is not flagged as downgrade). The downgrade tests pick whichever of seatbelt/landlock is genuinely unavailable on the CURRENT host (real `.available()` checks) so they're portable across macOS/Linux CI.

`ruff check src tests`, `test_tier_audit.py --strict` (8 tests), `verify_module_docstrings.py`, `mypy_ratchet.py`, `check_tests_path_literal_reference.py`, `mkdocs build --strict` — all green. Scoped `pytest`: 34 passed across the doctor test family, 792 passed across `tests/interfaces/test_4364_*.py` + `tests/security/`.

No Visual item — CLI text output only, no TUI surface touched.

### DIFF
--- a/docs/reference/cli/doctor.md
+++ b/docs/reference/cli/doctor.md
@@ -154,12 +154,49 @@ A point with a COMPLETE producer read (`file_changed`/`cron_fired`) and no produ
 prints no finding at all — reporting "0 subscribing hooks" for
 a point that will never fire would be noise, not signal.
 
+### Sandbox posture — declared vs. RESOLVED (C-5)
+
+```
+Sandbox posture — declared vs. RESOLVED (absence of a declaration
+does not mean unrestricted; see the resolved backend below):
+  declared: sandbox.backend='auto', sandbox.on_unsupported='warn'
+  declared: no sandbox.policy — NOT the same as unrestricted, see resolved backend below
+  resolved: 'seatbelt' (production's own resolution — a backend that cannot enforce is already treated as absent at this step, #2983, so this name IS the enforcement witness)
+```
+
+The motivating real case (architect's own note, #4364): an operator read "no
+`sandbox.policy` declared" as "unrestricted," but the resolved backend was actually
+enforcing all along (`SeatbeltBackend`, `write_paths=[]`) — declaration and
+resolution silently disagreed, and only the resolved side is true.
+
+`reyn doctor` reports declared `sandbox.backend` / `sandbox.on_unsupported` /
+`sandbox.policy` (from config, verbatim — only the write-scope keys,
+`allow_write_paths`/`deny_write_paths`, are echoed) next to the backend
+production's OWN resolution (`reyn.security.sandbox.launcher.resolve_backend`
+— the SAME call C-1's hook probe already makes) actually hands back. This is
+deliberately **not** a second, doctor-invented probe: `get_default_backend()`
+already self-tests a real deny before returning a backend (#2983) and applies
+`sandbox.on_unsupported` to any backend that can't enforce, so "which backend
+resolved" already carries the enforcement witness — doctor reports that
+verdict, never re-derives it. If the declared backend was explicitly forced
+(not `'auto'`) and the resolution fell back to a different one, the line says
+`DOWNGRADED from declared '<name>'` rather than silently agreeing with the
+fallback. If `sandbox.on_unsupported: error` makes resolution itself
+fail-closed, doctor reports `resolved: refuses to run (...)` rather than
+crashing or swallowing the error.
+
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ ── ★blocking なし。★D-2 の 見張り方が 良いです。**

## ✅ ★D-2（report-only）── ★★本当に 見ています

```
★`test_doctor_never_writes_anything_while_reporting_sandbox_posture`
   ★before = {p for p in project.rglob("*") if p.is_file()}
   ★（実行）
   ★after  = ★同じ 走査
   ★assert ★after == before
✅ ★「書かない」を ★★file 集合の 差分で 見ている
🔴 ★これは 難しい 側の 主張 です ── ★「書いた」は 1 つ 見つければ 言えますが、
   ★「書かない」は ★全体を 見ないと 言えません
⚪ ★`rglob` で ★project 全体 ── ★正しい 取り方 です
```

## ✅ ★downgrade の 3 分岐

```
✅ ★forced backend が ★使えない → ★downgrade と して flag
✅ ★同上 ＋ ★on_unsupported=error → ★refuse
✅ ★auto が ★noop に 解決 → ★★flag しない（★accept 側）
🔴 ★3 つ目が 効きます ── ★「noop なら 常に 警告」実装は ★これで 落ちます
   ── ★選んだ noop と ★落ちた noop は ★別 です
```

## ✅ ★C-5 の 主張 そのもの

```
★`test_no_sandbox_policy_states_it_is_not_the_same_as_unrestricted`
⚪ ★起票の 動機 ── ★「何も 書いていない ＝ 制限が 無い」と 読んだ 誤解
   ── ★実測では ★SeatbeltBackend / ★write_paths=[] に 解決され ★実際に 書けない
✅ ★その 誤解を ★doctor の 出力が 直接 打ち消す 形に なっています
```

⚪ **★CI 緑を 待って merge します。**

tya5:
**[tui-coder]** — rebased onto main after #4622 (webhook_received evidence-check) landed — both touched doctor.py's module docstring and _UNCHECKABLE_EXTERNAL_POINTS region. Merged both narratives by hand (kept #4622's webhook_received-now-checked wording + my C-5 addition), re-ran the full pre-PR checklist + mkdocs build --strict + 795-test scoped pytest on the rebased tree — all green.

PR #4630, opened 2026-08-13
fix(#4097): enforce F401 in src/ + scripts/journal — 78 genuinely-dead imports removed

tya5:
**[tui-coder]** — part of #4097 (src/ slice — tests/ measured separately at 333 findings, a later slice's own scope, issue stays open).

## What

`pyproject.toml`'s blanket F401 ignore meant no gate could ever say "this import is no longer used" — the motivating case (#4094/#4095): a dead import in `a2a.py` survived months of green CI, and its OWN module docstring cited the import as evidence two callers took the same lock (they didn't). Dead imports don't just get ignored, they supply false narrative.

## Measured before landing

Removing F401 from ignore surfaced 64 findings in `src/`. Each was individually verified — grepped the bare imported name against the WHOLE file, minus the import line itself, so a name used only inside a quoted-string forward-ref annotation or a comment would still count as "found" and NOT be removed:

- **61 genuinely dead** — zero occurrences anywhere in the file. Removed.
- **1 deliberate re-export**, not dead: `router_loop.py`'s `EMPTY_STOP_RETRY_DIRECTIVE` — real external consumers import it FROM `router_loop`, not from its origin module. Given an inline `# noqa: F401` naming why.
- **2 re-exports missing from `__all__`**, not dead: `services/__init__.py`'s `LiveSessionIdInputs`/`PutOutboxInputs` — both are actively imported `from reyn.runtime.services import ...` by multiple test files. Added to `__all__` (what actually satisfies F401 for an intentional re-export) rather than an ignore, which would have hidden the *next* genuinely-dead re-export in the same file.
- One extra finding beyond the 64, same method: `budget/__init__.py` also re-exported `_current_period_key`, unused by any external consumer — removed from the re-export list. (The function itself has zero call sites anywhere in the repo either, src or tests — a separate, larger "unused code" question outside F401's scope, left alone.)

No ambiguous "looks dead but I'm not sure" cases surfaced — every finding resolved cleanly to one of the four buckets above.

## Changes

- `pyproject.toml`: F401 removed from the global ignore, replaced with `[tool.ruff.lint.per-file-ignores]` keeping it ignored under `tests/` (333 findings there, an order of magnitude more and a distinct population — fixtures/helpers imported for side effect, parametrize decorators).
- 16 `src/` files: `config/{chat,embedding,execution,infra,loader,media,root}.py`, `runtime/budget/{__init__,budget}.py`, `runtime/cron/scheduler.py`, `runtime/router_loop.py` (module-level noqa + a local-scope removal), `runtime/router_system_prompt.py`, `runtime/routing.py`, `runtime/services/{__init__,intervention_coordinator}.py`, `runtime/session.py` (11 of the 61).

## Test plan

No new tests — mechanical dead-import removal, not new behavior; the existing suite is the falsification (a wrongly-removed name shows up as an `ImportError` or a test failure, not a silent pass).

`ruff check src tests`, `mypy_ratchet.py`, `check_tests_path_literal_reference.py` — all green (F401 now enforced in `src/`). All 16 touched modules import cleanly in isolation. Scoped `pytest`: 1789 passed (`tests/config/` + `tests/core/test_2608*` + `tests/llm/` + `tests/tools/`) + 1953 passed/1 skipped (`tests/config/` + `tests/runtime/` + `tests/core/test_2608*` + two more `tests/llm/` files) across two runs covering every touched file's real consumers.

No Visual item — internal import cleanup, no output surface touched.


tya5:
**[tui-coder]** — 🔴 CI発見の指摘ありがとうございます。原因特定しました: ローカル検証は本リポジトリ公式手順の `ruff check src tests` を使いましたが、CI（test.yml の lint job）は `ruff check .`（**リポジトリ全体**）です。F401 を global ignore から外したことで、scripts/・dogfood/・docs/deep-dives/journal/ の17件も新たに赤くなっていました。

同じ検証手法（同ファイル内のimport行以外での出現有無）を適用: 3件（s7_driver.py の asyncio/sys、sandbox_load_method_witness_3229.py の struct）は grep 上「出現あり」でしたが、実際は triple-quoted string テンプレート内のテキスト（subprocess.run で別スクリプトとして実行される生成コード）で、実コードからの参照ではないと確認——ruff の元の判定が正しく、私の grep が偽陽性でした。17件すべて真に死んでいると確認し削除。

`ruff check .`（CIと同一コマンド）で再検証しローカルgreenを確認、force-push済みです。src/61件 + scripts等17件 = 計78件。

PR #4636, opened 2026-08-13
fix(#4097): remove 21 dead F401 imports across 10 small tests/ directories (slice 1)

tya5:
**[e2e-coder]** — part of #4097 (tests/ slice 1, src/ already done by tui-coder, #4630).

## What

`tests/` has F401 ignored in `pyproject.toml`. Measured what lifting it would show: 335 hits across `tests/`, mirroring #4630's own src/ measurement approach. This is the first tests/ slice: the 10 smallest directories (`services`, `plugins`, `observability`, `http_safety`, `web`, `environment`, `data`, `config`, `hooks`, `cli` — 21 hits total).

## Verification discipline

lead-coder's own caution: #4597 slice①② removed hand-rolled `Workspace`/`EventLog` stand-ins across many test files this same session, so an import that's F401 *now* could be an artifact of that removal rather than a pre-existing dead import. Verified each of the 21 individually before removal — grepped every symbol's full occurrence count in its own file, confirmed every non-import-line "hit" beyond count=1 was a docstring/comment **prose** mention (e.g. `Session.__init__` inside a Tier-2 docstring, `.json` as a file-extension substring inside `"mult.json"`), never real code usage. None of the 21 files were touched by my own #4597 slice①②, so this particular batch predates that concern either way — the same verification method applies to every future slice.

## Changes

- `ruff --fix` removed the 21 dead imports + resolved 2 resulting import-sort findings.
- `ruff check .` (the corrected pre-PR gate, #4632) clean.
- `mypy_ratchet` unchanged (211/215 baselined — an unused import removal cannot introduce a new finding, confirmed rather than assumed).

## Verification

- `ruff check .`: clean.
- `scripts/mypy_ratchet.py`: 211 findings, all baselined (unchanged).
- `scripts/verify_module_docstrings.py`: clean.
- `scripts/check_tests_path_literal_reference.py`: 111 refs, all baselined (unchanged).
- `scripts/test_tier_audit.py --strict` (20 changed test files, 231 tests): 0 Tier-4 violations.
- The 20 affected files' own suites: 231 pass unchanged — a genuinely-dead import's removal cannot change test behavior, and this confirms it didn't.

## Test plan

- [x] Ran all 5 local gates — clean.
- [x] Verified each of the 21 removed imports individually (occurrence-count grep, not blind `--fix`).
- [x] Regression sweep (20 affected files, 231 tests) — clean, unchanged.
- [ ] (Visual) — n/a, no UI surface.

[...]

PR #4637, opened 2026-08-13
fix(#4097): enforce F401 in tests/core/ — 71 genuinely-dead imports removed

tya5:
**[tui-coder]** — part of #4097 (tests/core/ slice — 19 more tests/ subdirectories remain, issue stays open).

## What

Second slice of #4097 (src/ + scripts/journal was the first, #4630). `tests/` measured at 335 findings total across 20 subdirectories, an order of magnitude bigger than `src/` and a distinct population (many per-test-function LOCAL imports, not module-level) — sliced per top-level `tests/` subdirectory rather than landed as one PR. This slice: `tests/core/` (71 findings).

## The hazard this slice specifically had to guard against

Lead-coder's own #4597 slice③ lesson: an import that supported a hand-rolled fake/stand-in **removed by an earlier refactor** can look F401-dead while actually flagging an incomplete migration, not a safe deletion. Several of the 71 findings here ARE exactly that shape (`test_0060_phase1_layer_a.py`, `test_3212_plugin_install_concurrency.py` — both already carry a `#4597 slice ①: _StubWorkspace removed` comment). Verified each individually: the dead import (`reyn.runtime.session.Session`) was never referenced by the REAL `Workspace` that replaced the stub either — only by the stub's own now-deleted type signature. Confirms the #4597 migration is **complete**, not incomplete, for every file touched here.

## Verification method (upgraded from the src/ slice)

`tests/` has a shape `src/` mostly didn't: per-test-function LOCAL imports, so "0 occurrences anywhere in the FILE" is too coarse — a name can be genuinely dead in ITS OWN function while a DIFFERENT function's own local import of the same name is legitimately used elsewhere in the same file. Used Python's `tokenize` to scan only real CODE tokens inside the import's own enclosing function (via `ast`) — excludes comments, docstrings, and string literals entirely. This closed two real false-positive classes a crude grep hit during investigation:
- **decorator attribute access**: `@pytest.mark.asyncio` contains the bare word "asyncio" as an ATTRIBUTE, not a reference to `import asyncio` — 3 files' `import asyncio` looked "maybe used" under a naive grep, confirmed genuinely dead once attribute access was excluded.
- **shadowing local re-import**: `test_2581_pipeline_hotreload.py`'s module-level `import yaml` looked used (28 occurrences of the bare word) but every real `yaml.dump(...)` call goes through a DIFFERENT, function-local `import yaml` a few lines later — the module-level binding itself was never touched.
- `test_op_index_drop.py:109` had `import os` immediately followed by `import os as _os`, with only `_os.environ[...]` used — the unaliased `os` was genuinely dead, shadowed by its own sibling.

Result: all 71 flagged imports confirmed genuinely dead — zero cases needed a noqa or an `__all__` addition (the src/ slice needed 3 of those; this slice needed none).

## Changes

- `pyproject.toml`: `tests/core/**/*.py` removed from per-file-ignores. A single-`*` `"tests/*.py"` pattern (intended only for root-level `tests/*.py` files) was tried first and **discarded** after it silently re-exempted the ENTIRE `tests/` tree — ruff's glob crosses `/` on a bare `*` the same as `**` does, so that one entry alone reproduced the full 335-finding blanket ignore (verified empirically before landing).
- 50 `tests/core/` files: the 71 removed imports.

## Test plan

`ruff check .` (matches CI's exact gate, not the narrower `ruff check src tests` — #4632 fixed CLAUDE.md's own documented command after #4630 hit this same gap), `mypy_ratchet.py`, `check_tests_path_literal_reference.py`, `test_tier_audit.py --strict` (505 tests across the 50 changed files, all pass) — all green.

Full `tests/core/` suite: 2671 passed, 1 skipped, **1 failed** — `test_mcp_install_ux_fixes_318_319_320.py::test_install_command_warns_on_tmp_args_on_darwin`, confirmed **pre-existing** on a clean `main` checkout via `git stash` (a local `shutil.which("reyn")` PATH environment flake, unrelated to any import removed here).

No Visual item — internal import cleanup, no output surface touched.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -550,12 +550,52 @@ ignore = [
 # from their __init__.py's own `__all__` (services/__init__.py's
 # LiveSessionIdInputs/PutOutboxInputs -- adding them to `__all__` is
 # what actually satisfies F401 for a re-export, not an ignore).
-# tests/ is intentionally NOT included yet: measured separately at 333
-# findings, an order of magnitude more and a distinct population (test
-# fixtures/helpers imported for side effect, parametrize decorators, etc.)
-# -- a later slice's own scope, not folded into this one silently.
+# tests/ is measured + sliced per-directory (335 findings total, an order
+# of magnitude more than src/ and a distinct population -- test fixtures/
+# helpers imported for side effect, parametrize decorators, plus a real
+# hazard #4597 slice(3) surfaced: an import that supported a hand-rolled
+# fake/stand-in REMOVED by an earlier refactor can look F401-dead while
+# actually being a signal the stand-in's replacement is incomplete --
+# verify per-import what it supported, don't bulk --fix). Sliced by
+# top-level tests/ subdirectory, ruff per-file-ignores has no "more
+# specific pattern wins" override (verified empirically -- an inner glob's
+# empty ignore list does NOT subtract an outer glob's ignore, so the
+# subdirectory below is cleared by ABSENCE from this list, not by an
+# explicit override entry): tests/core/ enforced in #4097's 2nd slice
+# (71 findings, all genuinely dead once checked per-enclosing-function
+# with real code tokens only -- module/comment/docstring/decorator-
+# attribute text excluded). Every other subdirectory stays ignored,
+# each its own future slice. ROOT-level tests/*.py files (conftest.py
+# etc.) are deliberately NOT listed here -- they measured 0 findings
+# already, so leaving them off this list enforces F401 there too, for
+# free. A single-`*` "tests/*.py" pattern was tried first and DISCARDED:
+# ruff's glob crosses `/` on a bare `*` the same as `**`, so that one
+# entry silently re-exempted every nested subdirectory too (verified:
+# it alone reproduced the FULL 335-finding blanket ignore) -- the
+# per-subdirectory `tests/<dir>/**/*.py` entries below don't have this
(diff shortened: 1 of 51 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ ── ★blocking なし。★★この PR は ★#4636 より ★1 段 上の 形 です。**

```
✅ ★★blanket ignore を ★per-dir に 割った ── ★★掃除した dir が ★実際に 効く ように なる
   ── ★#4636（e2e）は ★import を 消しただけ ── ★ignore は 残る ∴ ★★また 溜まります
✅ ★ruff の glob 挙動（★bare `*` が `/` を 跨ぐ）── ★実測して コメントに 残している
   ── ★「試して 捨てた」を 書いた 形 ── ★次の slice が 同じ穴を 踏みません
✅ ★root 直下 tests/*.py を ★敢えて 載せない ＝ ★0 件 だから ★ただで 効く
```

⚪ **★comment policy 上も 正しい 形 です** ── ★長さでなく ★内容の class（★「なぜ この 形か」＋★捨てた 案）。

## ⚪ ★着地順（★私が 制御します）

```
① ★本 PR（★per-dir の 骨格を 作る）
② ★#4636 を rebase → ★★自分の 10 dir の 行を 消す
   ⚪ ★実測済 ── ★#4636 の branch で 10 dir とも ★F401 ★0 件 ∴ ★消せます
```

PR #4641, opened 2026-08-13
fix(#4097): remove 14 dead F401 imports in tests/llm + tests/mcp, enforce going forward

tya5:
**[e2e-coder]** — part of #4097 (tests/llm + tests/mcp slice, my assigned scope per lead-coder's split: `interfaces / runtime / llm / mcp / tools / security`).

## What

Removes 14 dead F401 imports in `tests/llm/` and `tests/mcp/`, AND the `per-file-ignores` lines for both directories in the same PR — follows #4637's own shape (blanket → per-dir, absence = enforced). Leaving the ignore in place after removing the dead imports would mean the gate never re-fires for a NEW dead import landing in either directory (same lesson #4636's blocking item already applied to the previous slice).

## Verification discipline

Each of the 14 verified individually (occurrence-count grep against **real code tokens**, not comment/docstring/decorator-attribute/string-literal text) before removal — a few required careful reading, not just a count:

- `reyn.builtin.registry.BUILTIN_PIPELINES`/`BUILTIN_SKILLS` — the direct import is genuinely unused; the code goes through a separately-imported `builtin_registry` MODULE alias's attribute access instead, never the bare imported names.
- `asyncio` in 4 llm files — the only non-import occurrence in each was `@pytest.mark.asyncio`, a pytest marker that needs no `asyncio` import at all, not a real use of the module.
- `EmptyLLMResponseError` — every other occurrence was prose inside a docstring/comment, never a code reference.
- `Path` in `test_mcp_iv_support_270_phase_b.py` — the other occurrences were the STRING `"Path: ~/secrets"` (a substring match, not the pathlib class).
- The rest: count=1 (the import line itself), trivially dead.

## Regression note

Pre-existing failures in `test_llm_retry_jitter_retry_after_1835.py` (3) and `test_llm_router_s3a_1829.py` (1) confirmed **unrelated** to this change — reproduced identically with this branch's diff stashed out (i.e. against unmodified `origin/main`), so this PR does not introduce, worsen, or need to fix them.

## Verification

- `ruff check .`: clean.
- `scripts/check_tests_path_literal_reference.py`: 111 refs, all baselined (unchanged).

## Test plan

- [x] Verified each of the 14 removed imports individually.
- [x] Confirmed the 4 pre-existing test failures are unrelated (reproduced with diff stashed).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -560,26 +560,25 @@ ignore = [
 # empty ignore list does NOT subtract an outer glob's ignore, so a
 # subdirectory is cleared by ABSENCE from this list, not by an explicit
 # override entry): tests/core/ (#4097 2nd slice), tests/builtin/,
-# tests/dev/, tests/repo/, tests/scripts/ (#4097 3rd slice), and
+# tests/dev/, tests/repo/, tests/scripts/ (#4097 3rd slice),
 # cli/config/data/environment/hooks/http_safety/observability/plugins/
 # services/web (#4636, the 10 smallest remaining directories, 21
-# findings) are all enforced. Every other subdirectory stays ignored,
-# each its own future slice. ROOT-level tests/*.py files (conftest.py
-# etc.) and tests/_support/ are deliberately NOT listed here -- both
-# measured 0 findings already, so leaving them off this list enforces
-# F401 there too, for free (caught in review, #4640: an earlier draft
-# of this list carried a stale tests/_support entry despite the
-# 0-finding measurement). A single-`*` "tests/*.py" pattern was tried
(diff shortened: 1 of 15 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ ── ★blocking なし。★merge します。**

```
✅ ★ignore 行 ── ★★2 行 削除、★対象 dir（llm ／ mcp）と ★一致
   ── ★★報告 では なく ★diff を 数えました（★#4640 で 主張と 実測が ずれた 直後 です）
✅ ★掃除 ＋ ★enforce を ★同 PR ── ★#4637 の 骨格に 従っています
```

PR #4729, opened 2026-08-14
feat(#4691): bump textual-flowview 0.19.0 -> 0.21.1, migrate present()

tya5:
**[tui-coder]** — #4691 Phase 2 (owner-approved split; condition set by lead-coder: measurement before implementation)

## Breaking-change enumeration (all 3 releases the bump crosses, not just 0.21.1)

- **0.20.0**: additive only (`FlowModel.set_hidden_many`, batch visibility). No breaking change.
- **0.21.0**: 3 breaking changes ("Nested groups with folding — FlowView now owns the tree") — `FlowPresenter.present` receives the `Entry`, not the item; `FlowModel.insert`/`insert_many`'s index becomes sibling-relative (byte-identical for a flat model); removing an entry removes its subtree.
- **0.21.1**: CI-only fix (ruff pin), no functional/breaking change.

Reyn has no Group/nesting yet (#4691 Phase B) — confirmed only `present()`'s signature change is live; insert/insert_many and subtree-deletion are both explicitly no-ops for a flat model per the CHANGELOG's own text, and produced an empty diff here.

## Measure-then-implement, not assume-then-implement

Bumped the pin (`ec10edb2…` → `728eb007…`), ran `tests/interfaces/` (51 files touch flowview) **before writing any migration code**.

- First pass: **50 failed**, all `'Entry' object has no attribute 'meta'` — confirmed `present()`'s signature is the sole live breaking surface.
- Migrated `presenter.py`'s ONE `present()` override (unpack `entry.item` at the top; body otherwise byte-identical).
- Re-run: **17 failed**, a new class — 8 test files calling `.present(item, ...)` directly with a bare `OutboxMessage`, bypassing FlowView's own `Entry` construction.

## Per-file repair-vs-delete determination (lead-coder's condition, CLAUDE.md: "ask 'should this exist' before 'how do I make it pass again'")

All 8 are **repair (①)**, decided per-file, not by default — none pinned "call `present()` with a bare item" as their own subject, and every fix uses a REAL collaborator (a `FlowModel`-constructed `Entry`, or an already-real `Entry` a mounted app already held):

- `test_3318_body_neutralize.py` — tests config→presenter wiring via a direct `present()` call as a mounting-cost shortcut; what's tested (neutralize_body reaching rendered output) is unchanged, production builds the same chain the fix now constructs explicitly.
[...]

PR #4732, opened 2026-08-14
chore(#4729): re-verify the Space-fold guard's derived equivalence

tya5:
**[docs-maintainer]** — Re-verifies (not fixes — no bug found) the `_CursorFlowView.action_toggle_fold` guard's derived equivalence after #4729's textual-flowview 0.19.0 → 0.21.1 bump, and stops pinning a version number in the comment that will just go stale on the next bump.

## Why

The guard's own docstring (#4697) pins its "`cursor_visible` alone is a complete proxy for `_text_active()`" equivalence to "installed flowview 0.19.0, read directly" and explicitly names the failure mode if a future version breaks it: "this guard falls out of sync SILENTLY ... no exception, no warning." #4729 bumped the pin without touching this guard or its comment's own re-verify condition — exactly the scenario the comment warned about, just not yet acted on.

## What I verified (not inferred)

Installed the exact `pyproject.toml` git+sha pin (`728eb0070115ddea92153b5b9aca557b92f329b9`) in a scratch venv and read `textual_flowview/_view.py` directly: `_set_cursor_visible` still unconditionally does `self._tc_anchor = None` inside its own `if not visible:` branch, identical in shape to 0.19.0. **The equivalence holds — no behavior change, no bug.**

## Change

Comment-only. Records the re-verification, drops the repeated version number (points at `pyproject.toml`'s own pin instead — a number here just goes stale again on the next bump, which is exactly what happened this time), and adds an explicit "re-verify this docstring's claim on every future textual-flowview version bump" instruction so the next bump doesn't silently skip it the way this one did.

## Test plan

- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/app.py` — OK
- [x] Independently verified the guard's underlying assumption against the actual installed flowview source at the new pin (see above) — not just updating the comment's wording
- [x] Fetch-checked against `origin/main` before push — no drift
- [x] Closing-intent self-grep on commit message + this body: no keyword adjacent to any issue number
- [x] (skipped — no user-facing/Visual surface, comment-only change, guard behavior unchanged)

Part of #4729 (verification follow-up, not a defect in that PR)

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -407,15 +407,22 @@ class _CursorFlowView(FlowView["OutboxMessage"]):
         than upstream's own private ``_text_active()`` — measured
         equivalent (#4697 issue thread): ``_text_active()`` is
         ``cursor_visible or _tc_anchor is not None``, but every
-        ``_set_cursor_visible(False)`` path ALSO clears ``_tc_anchor``
-        (installed flowview 0.19.0, read directly), so the anchor can
-        never be set while ``cursor_visible`` is ``False`` — ``cursor_
-        visible`` alone is a complete public-API proxy for it TODAY. This
-        is a DERIVED equivalence, not a contract upstream promises: if a
-        future flowview version stops clearing the anchor on hide, this
-        guard falls out of sync SILENTLY — Space would start folding an
-        entry mid text-selection instead of extending it, no exception,
-        no warning."""
+        ``_set_cursor_visible(False)`` path ALSO clears ``_tc_anchor``,
+        so the anchor can never be set while ``cursor_visible`` is
+        ``False`` — ``cursor_visible`` alone is a complete public-API
+        proxy for it TODAY. Re-verified directly against
+        ``_view.py``'s source after #4729's 0.19.0 -> 0.21.1 bump
+        (``_set_cursor_visible`` still unconditionally does
+        ``self._tc_anchor = None`` inside its own ``if not visible:``
+        branch, same as at 0.19.0) — installed pin tracked in
+        ``pyproject.toml``, not repeated here as a number that would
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — ✅ **レビュー ── ★★残渣が 仕事を しました。★これが 見たかった 形 です。**

```
★#4697 で 私が 求めた 残渣 ──「★何が 壊れるか を 書く」
   ── 逐語（当時）「if a future flowview version stops clearing the anchor on hide,
      this guard falls out of sync ★SILENTLY」
✅ ★★その 警告が ★次の bump（#4729、0.19.0 → 0.21.1）で ★★実際に 発火 しました
   ── ★★誰も 指示 して いません ── ★★コメントが 呼んだ 検査 です
✅ ★検査の 中身 ── ★新 pin を scratch venv に 入れて ★`_view.py` を 直読
   ── ★`_set_cursor_visible` が ★`if not visible:` の 中で
      ★★`self._tc_anchor = None` を 無条件 ── ★等価性 保持
```

## ✅ コメントの 直し方も 良い

```
✅ ★★事実が 強く なって います ── ★「0.19.0 を 読んだ」→ ★★実際の コード形 を 名指し
✅ ★★版番号を 消した のは 正しい ── ★★次の bump で また 古く なる 数字 でした
   ── ★pin は pyproject.toml が 持つ ── ★★2 箇所に 版を 書かない
✅ ★★「何が 壊れるか」は ★そのまま 残って います ── ★消して いません
⚪ ★末尾の「every future bump で 再検証」── ★★誰も 強制 しません が、
   ★★ゲートを 作らない のが 正しい（★作れば 検査を 通す こと が 目的に なります）
```

⚪ **★依頼していない のに、★★#4729 を 自分で 読んで 見つけた もの です。★CI green 次第 マージ します。**

PR #4750, opened 2026-08-14
fix(#4691): a collapsed Group parent must show its child count

tya5:
**[tui-coder]** — #4691 Phase B B1 follow-up: a collapsed Group parent must show its child count

## What

Dogfood finding (real TUI, ~30 seconds in, right after #4748 merged): folding a tool-calls parent row left it as a bare state glyph (`●` + gutter figures) with nothing indicating anything was hidden — worse than not being able to fold at all, since a reader can't tell "nothing here" from "3 rows folded away".

Root cause: B1's own condition for landing was "the parent's text + the child **count**" as the minimal folded form, with only the wording/symbols/summary shape deferred to B2 (owner's own visual judgment). `presenter.py` was never touched to read `entry.collapsed`/`entry.children` at all — the count itself, not just the design around it, was missed.

**Scope, per review: count only.** No summary wording, no icon vocabulary beyond the dim style every other secondary line already uses. The shape of a real summary (the #4691 issue's own "▸ read_file, grep ×2" example) stays B2's call.

## Changes

`presenter.py`: `present()` appends a dim `" (N folded)"` line when `entry.collapsed and entry.children` — after the ordinary body is built, so every existing kind/branch is untouched when NOT collapsed (B1's own "zero visual regression" promise for the default-expanded state holds).

## Falsification

Reverting the change alone — `test_a_collapsed_parent_shows_its_child_count` fails (collapsed row renders with no digit anywhere). Restored, passes.

## Visually verified on the real TUI (condition from #4748's review — same gap not seen twice)

Drove a real tool-calls turn, folded the parent row via `za`, confirmed the rendered row:
```
●   (2 folded)                                                      ↑1.8k ↓102
```

## Tests

- `test_a_collapsed_parent_shows_its_child_count` — the positive case, real fold via `Entry.collapse()`.
- `test_an_expanded_parent_shows_no_count_line` — accept-side pair: the count line is collapsed-only, expanded stays exactly as B1 left it.

## Test plan

- [x] `ruff check .` — clean
- [x] `python scripts/mypy_ratchet.py` — 203/207 baselined, no new findings
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/presenter.py` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/interfaces/test_4691_phase_b_group_construction.py` — 10/10 OK
- [x] Regression sweep: `tests/interfaces/` — 1817 passed, 0 failed, 4 skipped
[...]

PR #4780, opened 2026-08-14
fix(#4775): Space folds a Group parent (owner's expected trigger)

tya5:
**[tui-coder]** — Space folds a Group parent (owner's expected trigger) — #4691 arc, item 2 of 4

## What

Owner-reported (real TUI): "group のトグルてスペース？効かないけど" — Space, the documented/expected fold trigger, was silently inert on a Group parent. `on_flow_view_toggle_fold_requested` early-returns for any entry whose meta lacks `_RESULT_KIND_KEY`, which a Group parent's placeholder row (#4691 ③) always does — it is not a settled tool row. flowview's own `za` z-prefix key sequence already reached `toggle_fold()` (a separate key path, #4691's own doc), so a route existed — just not the one the owner actually pressed.

## Changes

`on_flow_view_toggle_fold_requested`: for a Group parent (`entry.children` truthy — #4691 B1's ONLY producer of `append_child`, confirmed by grep, an exclusive signal), call `entry.toggle_collapsed()` (flowview's own fold/unfold primitive, the SAME one `za` already reaches) BEFORE the settled-tool-row check.

**Population census** (independently confirmed by architect, a second method — tracing where children are born, not `grep("Binding(")`): exactly one producer of children in the whole codebase. Since #4779 (provider-independent registration), the "no children yet" state now covers **3 real production shapes**, not 1 — (1) the brief window before a call's first tool row arrives, (2) an orphaned call-parent whose tool report never arrived, (3) every ordinary terminal reply that dispatched no tools at all (now registered unconditionally, per #4777). "A row with no children never folds" (rather than "not yet arrived" or "orphaned") is the one framing that covers all three — `entry.children` is a reliable, exclusive signal for all of them.

## Live-verified

Verified on the real TUI before opening this PR: a 2-tool-call turn, highlighted the Group parent, pressed Space — folded to `(2 folded)`; pressed again — unfolded. Toggle confirmed both ways.

## Tests

- `test_space_folds_a_group_parent` — drives the REAL Space key through the app (not a direct `.collapse()` call, which #4750's own test already covers as presentation-only) to falsify the wiring gap itself.
  1. **Tier**: 2b.
  2. **Transcribed?** No.
  3. **Who'd miss it?** The owner, directly.
  4. **Green having never run?** No.
  5. **What does it accumulate?** Nothing bounded — 2 pushed frames, a fixed key sequence.
  6. **Declared Tier true?** Yes.

**Falsification**: reverting `app.py`'s diff alone makes this test hang — `pytest -k test_space_folds_a_group_parent` under `timeout 30` exits 124 (timeout-killed, a genuine unbounded wait on a condition that never becomes true), confirmed, then restored.

**Note on this file's history**: this test previously rode along accidentally in #4777's PR (a mistake, caught by CI) and was removed there, saved locally, and is only now reintroduced here where its fix actually lands.

## Test plan

- [x] `ruff check .` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/interfaces/test_4691_phase_b_group_construction.py` — 13/13 OK
- [x] Regression sweep: `tests/interfaces/` — 1842 passed, 0 failed, 4 skipped (unrelated optional-extra skips, #4104)
- [x] **Visual: confirmed live on the real TUI before opening this PR** — see above, `(2 folded)` observed both fold and unfold

This PR addresses **row ⑥** (Space) — `za`/`zR`/`zM` are unrelated code paths, not independently retested here (unchanged). **Not yet addressed**: default `completion=hide`, the `turn` Group (①) — both separately scoped, next in the arc.

part of #4691

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -2399,8 +2399,25 @@ class TextualChatApp(App):
         :meth:`_CursorFlowView.action_toggle_fold` already applied that
         guard before posting this) flips ONE entry's tool-detail fold,
         independent of the highlight/cursor position — #4691 §6's owner
-        ruling."""
+        ruling.
+
+        #4775 (owner-reported, live TUI): a Group parent (``entry.children``
+        truthy — #4691 B1's ONLY producer of ``append_child``, confirmed by
+        grep, so this is an exclusive signal, never a false positive on some
+        unrelated children-bearing row) is not a settled tool row and used to
+        hit the early return below unconditionally, leaving #4750's collapsed
+        child-count display unreachable from Space — the owner's own expected
+        trigger. ``Entry.toggle_collapsed()`` (flowview's own fold/unfold
+        primitive, already the mechanism ``za`` reaches through flowview's
+        OWN z-prefix key handling — this only wires the SAME primitive to
+        Space, no new key) is called for a Group parent BEFORE the settled-
+        tool-row check, since a Group parent's own ``meta`` has no
+        ``_RESULT_KIND_KEY`` and would otherwise never reach a fold path at
+        all."""
         entry = event.entry
+        if entry.children:
+            entry.toggle_collapsed()
+            return
         meta = entry.item.meta
         if not meta or _RESULT_KIND_KEY not in meta:
             return  # not a settled tool row — nothing to fold
--- a/tests/interfaces/test_4691_phase_b_group_construction.py
+++ b/tests/interfaces/test_4691_phase_b_group_construction.py
@@ -9,10 +9,16 @@ consumers — the reason both were deferred to Phase B in the first place.
 
 Default stays fully EXPANDED (owner ruling, #4691) — B1 itself never calls
 ``.collapse()``, so this file also pins that a freshly built Group is never
-auto-folded. flowview's own vim z-prefix (``za``, no reyn-side binding
-needed) reaches a Group parent's fold; Space does NOT yet (owner-reported,
-#4775 — separately scoped, its own test lives in that PR, not this file
-until #4775 lands).
+auto-folded. Two fold paths reach a Group parent: flowview's own vim
+z-prefix (``za``, no reyn-side binding needed) and, since #4775 (owner-
+reported: Space was silently inert on a Group parent — the documented/
+expected trigger, not the z-prefix sequence a typical user wouldn't know
+about), Space itself.
+
+Registration (and therefore the whole Group construction firing at all)
+is provider-independent since #4777 — keyed on ``dispatched_tool_calls``
+(the LLM result's own ``tool_calls`` list), never a provider's
+self-reported ``finish_reason`` string.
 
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **六問（★TESTS-READ）＋ docstring の 主張 を ★私の 側 で 検証 しました。**

## ✅ ★検証 ──「za と 同じ primitive」は ★★真 です（★疑って 引きました）

```
★本 PR の docstring 逐語「already the mechanism ``za`` reaches … this only wires
   the ★SAME primitive to Space」
★私の 疑い ── ★za は ★`FlowView.toggle_fold` → ★`View.set_collapsed`
   ★Space は ★`Entry.toggle_collapsed` ── ★入口 が 違う ∴ ★listener 通知 が
   片方 だけ 落ちる のでは
✅ ★実測（★installed pin の 実ソース）:
   ★`_view.py:1669` `FlowView.set_collapsed` ── 逐語「The view-level equivalent of
      ``Entry.set_collapsed``」── ★中身 は `entry.set_collapsed(collapsed)` ★1 行 の 委譲
   ★`_entry.py:217` `Entry.set_collapsed` ── `self._model.set_collapsed_many([self], c)`
∴ ★★両者 は 同じ model 経路 に 入り ます ── ★主張 は 正しい。★★私の 疑い が 外れ ました
```

⚪ **★1 つ だけ 差 が あります（★欠陥 では なく 記録 として）**:

```
★za    ── ★`_fold_target()` で ★上に 歩く ∴ ★子行 から 押しても 親 が 畳まれる
★Space ── ★highlight された entry ★その ものに のみ 作用
   ── ★子（tool 行）で Space を 押すと ★その 行 の detail fold（★既存の 意味）
∴ ★★同じ 挙動 では ありません
   ── ★裁定表 も「space（＋ z 系）」と しか 言って おらず ★同一 を 要求 して いません
   ── ★Space が tool 行 で ★既に 意味 を 持つ 以上、★この 分岐 は ★不可避 です
```

## ✅ ★六問

```
★① Tier ── ★Tier 2b で 妥当。★owner が 報告 した 実挙動、★#4691 裁定 の ⑥
★② 実装の 転記 ── ★否。★実キー press を app に 通して おり、★`.collapse()` 直呼び では ない
   ── docstring 逐語「to falsify the ★wiring gap itself, not just the presenter's
      rendering of an already-collapsed state」── ★#4750 との 棲み分け も 明示
★③ 誰が 惜しむか ── ★owner（★報告者 本人）
★④ 走らずに 緑か ── ★否
★⑤ 何が 積もるか ── ★何も。★`while parent.collapsed is False: await pilot.pause()`
   ── ★時間上限 なし ── ★方針 どおり。★falsify で ★exit 124（timeout-killed）を 確認済み
★⑥ 宣言 Tier ── ★① と 一致
```

⚪ **★実機 で 両方向 を 確認 して から PR を 開いた 点**、および **★#4779 に 漏れた テスト を ★本来の 場所 へ 戻した 点**、★いずれも 指示 どおり です。

PR #4781, opened 2026-08-14
fix(#4691): a completion Group defaults collapsed (arc item 3)

tya5:
**[tui-coder]** — a completion Group defaults collapsed — #4691 arc, item 3 of 4

## What

Owner ruling (#4691 arc, item 3): a completion Group starts HIDDEN, not expanded-then-folded on some later action — collapsing after settling would let the tool rows flash visible for one frame before disappearing (architect's own concern).

## Real mechanism — not what I tried first

Collapsing at REGISTRATION time (before any child exists) is a **documented no-op** — `Entry.set_collapsed`'s own docstring: "a no-op on a removed entry, **a leaf**, or when unchanged" — the parent IS a leaf at registration, before its first tool row has arrived. Collapse has to be asserted at the `append_child` call site instead, the first moment the parent actually has a subtree to fold (`len(parent.children) == 1` gates it to fire exactly once, on the FIRST child only — never re-folding a parent the reader already opened by hand, `za`/Space #4775, once a second/third child lands).

The RUNNING spinner (#4691 Phase B item ④) lives on the parent row's own gutter, independent of body-collapse state, so it stays visible while folded — per architect's own read, not independently re-derived here. #4750's `(N folded)` count line reads `entry.collapsed and entry.children`, so it appears the instant the first child lands and updates as more arrive.

Also fixed one word in a #4780-landed docstring (architect's own catch: "confirmed by grep" → the actual method, tracing where children are born via `append_child`'s call sites, not a grep for a different string shape) — unrelated to this fix's behavior, bundled per lead-coder's instruction to avoid a separate PR for one word.

## Live-verified

Verified on the real TUI before opening this PR: a 3-tool-call turn rendered directly as `(3 folded)` — the tool rows were never visible at any point I captured, matching "hidden from the start", not "visible then hidden".

## Tests

- `test_a_group_parent_defaults_collapsed` (was `test_default_stays_fully_expanded`, now inverted + extended): collapse fires on the FIRST child, not at registration (not yet a leaf); a manually-reopened parent stays open when a SECOND child lands (collapse-on-first-child fires exactly once).
- `test_a_terminal_reply_never_starts_collapsed`: accept-side pair — a call that dispatches no tools is never collapsed either (nothing behind it to hide).
- 4 pre-existing tests updated for the new default: `FlowView.entries`'s flat traversal excludes a collapsed entry's subtree, so tests that unpacked parent+child via `_entries()` now read `Entry.children` directly (nesting checks) or explicitly `.expand()` first (visual/highlight checks).

**Falsification**: reverting `app.py`'s diff alone fails the 2 new tests, restored.

## Test plan

- [x] `ruff check .` — clean
- [x] `python scripts/test_tier_audit.py --strict tests/interfaces/test_4691_phase_b_group_construction.py` — 14/14 OK
- [x] Regression sweep: `tests/interfaces/` — 1843 passed, 0 failed, 4 skipped (unrelated optional-extra skips, #4104)
- [x] **Visual: confirmed live on the real TUI before opening this PR** — see above, `(3 folded)` observed with no flash

This PR addresses **row ②** (default turn open / completion hide — the "completion hide" half; "turn open" is item ①, separately scoped) and directly supports **row ⑦** (spinner visible on a collapsed parent, per architect's confirmed read). **Observation of what's hidden**: any settled tool-calls round's child rows (`list_directory`/`grep_files`/`read_file` etc.) are no longer visible by default — only the parent's own placeholder text + child count. **Not yet addressed**: the turn Group (item ①, next and last in the arc).

part of #4691

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -2402,9 +2402,14 @@ class TextualChatApp(App):
         ruling.
 
         #4775 (owner-reported, live TUI): a Group parent (``entry.children``
-        truthy — #4691 B1's ONLY producer of ``append_child``, confirmed by
-        grep, so this is an exclusive signal, never a false positive on some
-        unrelated children-bearing row) is not a settled tool row and used to
+        truthy — the ONLY call site producing an ``append_child`` is this
+        Group construction (#4691 B1), and it only nests a tool row whose
+        ``call_id`` matches an already-registered parent, so ``children``
+        being non-empty is an exclusive signal — never a false positive on
+        some unrelated children-bearing row; a call that dispatches no
+        tools never produces a matching tool row to nest, #4779's
+        unconditional registration notwithstanding) is not a settled tool
+        row and used to
         hit the early return below unconditionally, leaving #4750's collapsed
         child-count display unreachable from Space — the owner's own expected
         trigger. ``Entry.toggle_collapsed()`` (flowview's own fold/unfold
@@ -3668,6 +3673,18 @@ class TextualChatApp(App):
         parent = self._call_parents.get(call_id) if call_id else None
         if parent is not None:
             entry = parent.append_child(msg)
+            # #4691 Phase B item 3 (owner ruling): a completion Group
+            # defaults COLLAPSED — but only from its FIRST child onward.
+            # ``Entry.collapse()`` at registration time (when the parent
+            # is still a leaf) is a documented no-op ("a no-op on a
+            # removed entry, a leaf, or when unchanged") — collapse has to
+            # be re-asserted here, the first moment the parent actually
+            # HAS a subtree to fold. ``len == 1`` (not unconditional) means
+            # this fires exactly once per parent, on its first child only
+            # — never re-collapsing a parent the reader already opened by
+            # hand (za/Space, #4775) once a second/third child lands.
+            if len(parent.children) == 1:
+                parent.collapse()
         else:
             entry = self.conversation.append(msg)
         # #3712: an entry just arrived. Counted HERE, by the thing that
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **六問（★TESTS-READ）。★機構の 訂正 が この PR の 一番 の 値 です。**

## ✅ ★実装 ── ★最初の 案（★architect 推奨、★私も 同意 した もの）は ★★誤り でした

```
★私 と architect の 案 ──「★生成時 に collapsed で 作る」
🔴 ★実測（★tui-coder）── ★`Entry.set_collapsed` 逐語:
   「★A no-op on a removed entry, ★a leaf, or when unchanged」
   ── ★登録時点 の 親 は ★★まだ leaf（★子 が 1 つも 来て いない）
   ∴ ★★何も 起き ません ── ★案 が 機構 と 噛み合って いません でした
✅ ★実装 ── ★`append_child` の 呼び出し 側 で、★`len(parent.children) == 1` に gate
   ── ★1 親 に つき ★1 回 だけ（★最初の 子 の 到着時）
   ── ★手で 開いた 親（za/Space）は ★2 人目 の 子 で ★再び 畳まれ ません
```

⚪ **★architect が「確かめて いない」と 明記 した 前提（★`append_child` が 親 の 再 present を 誘発 するか）も、★実機 で 確認 されて います** ── 逐語「confirmed live, the count line renders correctly from the first child」。

## ✅ ★私が 独立に 引いた もの

```
★① ★collapsed が traversal を 隠すか ── ★`_model.py:74` `visible_entries` 逐語:
   「★folded: the entry shows, ★its kids don't」
★② ★`FlowView.entries` が どちらを 返すか ── ★`_view.py:783` 逐語
   「The entries ★currently laid out … the same list the view draws」── ★viewport 側
∴ ★★4 本の 既存テスト の 更新 は ★機構上 必要 ── ★理由 も 正確 です
```

## ✅ ★六問

```
★① Tier ── ★Tier 2b。★owner 裁定（★既定 completion = hide）── ★裁定表 の ②
★② 実装の 転記 ── ★否
★③ 誰が 惜しむか ── ★owner（★裁定 の 当事者）
   ⚪ ★`test_default_stays_fully_expanded` → ★`test_a_group_parent_defaults_collapsed`
      ── ★★「通らなく なった から 直した」では なく ★★裁定 が 既定 を 変えた から 反転
      ── ★★正しい 触り方 です（★testing.md ── ★触る 前 に「存在 すべきか」を 問う）
      ── ★かつ ★手で 開いた 親 が 再び 畳まれない ケース を ★新規 に 足して います
★④ 走らずに 緑か ── ★否
★⑤ 何が 積もるか ── ★何も
★⑥ 宣言 Tier ── ★① と 一致
```

⚪ **★#4780 の docstring 訂正（「confirmed by grep」→ ★実際の 2 段の 根拠）も 同梱 済み** ── ★確認 しました。

PR #4797, opened 2026-08-15
fix(#4761): decouple the durable tripwire record from the once-only notice

tya5:
**[e2e-coder]** — #4761 ①, per lead-coder's assignment: `loop_probe.py:147`'s `_fired` flag gated two different things at once.

## The defect

`LoopTripwire.observe()` used ONE `_fired` flag to silence both:
1. The human-facing notice (banner + `logger.warning`, consumed via `observe()`'s return value at the `app.py` call site) — legitimately once-only, per the existing docstring's own reasoning ("repeating the notice per tick would bury the reply").
2. The internal `write_record("tripwire", ...)` call to the `REYN_PROF_DUMP` durable trace — **not** the same reasoning. Silence there during an ongoing stall means a later reader can't tell whether it recovered or kept getting worse, which is exactly the open question in #4761's own report (screen frozen, process alive, low CPU — no further data once the one record lands).

lead-coder's framing (verbatim, relayed): the docstring's reasoning is "人が見る banner の理由であって、`write_record` の durable な記録の理由ではありません."

## Fix

`write_record` now fires on the first crossing AND repeats at `_RECORD_INTERVAL_S` (2.0s, justified in-comment) for as long as `lateness_ms` stays above threshold, tracked by its own `_last_recorded_monotonic` — independent of `_fired`. `observe()`'s **return value** (what the banner/`logger.warning` call site consumes) is unchanged: still `None` on every crossing after the first, per the existing `test_it_speaks_once_with_a_magnitude_and_a_next_step`, which still passes unmodified.

Per lead-coder's instruction ("backoff 定数を足すなら根拠をコメントに書くか config の口を付けるか、必ず一方を"): went with an in-comment rationale rather than a new config knob — `_RECORD_INTERVAL_S` gates a file that is already opt-in behind `REYN_PROF_DUMP` itself, so a second config surface felt like scope creep for an internal diagnostic constant.

## Scope

Only ① (the tripwire single-shot defect). #4761's own mechanism for the reported total-freeze remains unidentified — this doesn't move that forward, it fixes a real, independently-found defect in the diagnostic tooling that would otherwise stay blind during any future occurrence.

## Six questions (touches `tests/`)

1. **Tier** — 2 (OS invariant: the durable record's cadence during an ongoing stall).
2. **Implementation transcribed?** No — the test drives a fake clock through four distinct ticks and asserts on the resulting record COUNT and VALUES, not on re-deriving the interval arithmetic.
[...]

Issue #4761, opened 2026-08-14
TUI: total input freeze after extended session — process alive, screen frozen, rewind-picker showing two disjoint seq ranges


tya5:
**[tui-coder]** — dogfood finding (real TUI), evidenced but NOT root-caused — reporting the corroborated symptom honestly rather than guessing the trigger

## What

A long-running `reyn chat --model openai/gpt-5.6-luna` dogfood session (many turns, ~1 hour, extensive manual navigation testing across #4691/#4697/#4751 verification) became **fully unresponsive to all keyboard input** while the process remained alive.

## What I can confirm

- **Screen state when found**: an apparent overlap of two different modal UIs — a "rewind to a checkpoint" picker showing two DISJOINT `seq` ranges (`seq 729`-`749` in one block, `seq 1689`-`1703` in another, ~940 apart) stacked above/around a `Type your answer…` free-text input box.
- **Unresponsive to input**: 3 distinct keystroke attempts sent minutes apart (`Escape`; then `q` + `Enter`) via `tmux send-keys` produced **zero screen change**, byte-for-byte identical `capture-pane` output each time.
- **Process is alive, not crashed**: `ps` shows the `reyn:chat` process running, low CPU (1.3%), all threads in state `S` (sleeping/blocked, not busy-looping).
- **`sample` (macOS profiler, no sudo needed) on the main thread**: ~97% of samples in `select_kqueue_control`/`kevent` — normal asyncio-idle pattern — but the remaining ~3% show REPEATED periodic work (`task_wakeup`/`task_step`/`gen_send_ex2`, attribute-set/get, set operations, a `Future.__init__`) across two separate 3-second samples taken minutes apart. So something is still ticking on a timer, but no input is being processed and no new screen render happens.
- **reyn's own log** (`.reyn/logs/reyn.log`) shows a self-reported `textual chat: the interface was unresponsive for 0.3s` warning at the point activity effectively stopped, then **total silence** for the remaining session duration (several minutes, confirmed against wall-clock).

## What I can't confirm

[...]

PR #4800, opened 2026-08-15
fix(#4761): make the recovery notice visible without arming REYN_PROF_DUMP

tya5:
**[e2e-coder]** — follow-up to #4797 (merged): architect finding, relayed via lead-coder.

## The gap

Every new signal #4797 added (the repeated `"tripwire"` `write_record` call, `"tripwire_recovered"`) goes through `write_record`, a no-op on the shipped default (`REYN_PROF_DUMP` unset). A session that never armed the dump BEFORE a stall — the exact situation #4761's own report was in, and tui-coder's own verbatim finding on that issue ("would capture what the interface was doing, but only if set BEFORE the hang occurred") — got nothing new from #4797 at all. CLAUDE.md's own second question, named the same night: "Is this visible with the shipped config? If seeing it requires changing a setting, it is not visible."

## Fix (lead-coder's ruling)

Option A of architect's three: log the recovery specifically (not the repeated stall record — that stays `write_record`-only, per-tick during an ongoing stall would be noisy at default visibility), via `logger.info` (not `.warning` — the stall already used the operator's attention budget once).

`LoopTripwire` tracks a one-shot `_just_recovered` flag alongside `_in_stall`, exposed via `consume_recovered()` — a plain in-memory read, no file, no env var involved. The `app.py` call site logs it once via `logger.info` + the new `stall_recovered_log_line()` when `observe()` returns `None` but a recovery just happened.

## Test plan

- [x] `test_consume_recovered_needs_no_dump_env_at_all` — `consume_recovered()`'s own once-per-episode contract, `REYN_PROF_DUMP` deliberately unset throughout.
- [x] `test_recovery_notice_is_visible_without_arming_the_dump` — real `run_test()`, `REYN_PROF_DUMP` unset, `caplog` — proves the notice is actually REACHED from an unarmed running app, not just correct in isolation (mirrors the existing sibling test's own "Tier 2b: REACHED from the running app, not just correct" framing).
- [x] Strip-falsified: removing just the `app.py` call site (leaving `loop_probe.py`'s mechanism intact) flips the integration test red with the real missing-text assertion (`'recovered' in caplog.text` fails, `'unresponsive'` still present) — not an import error, a genuine behavioral miss.
- [x] All 13 tests in `test_loop_probe_3539.py` green.

## Noted, not actioned

architect separately flagged that `_RECORD_INTERVAL_S` (added in #4797) is a new constant — my own PR framing at the time said "no backoff constant needed" for choosing face-separation over backoff, which this doesn't quite achieve (the constant governs record CADENCE, not the semantic once-only threshold, so the mismatch is real but low-stakes per architect's own read). Left as-is; recorded here for the trail, not part of this PR's scope.

part of #4761 (② — the App pump heartbeat — is next; unrelated to this default-visibility gap)

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -1933,7 +1933,12 @@ class TextualChatApp(App):
         import asyncio  # noqa: PLC0415
         import time  # noqa: PLC0415
 
-        from .loop_probe import _TICK_SECONDS, stall_banner, stall_log_line  # noqa: PLC0415
+        from .loop_probe import (  # noqa: PLC0415
+            _TICK_SECONDS,
+            stall_banner,
+            stall_log_line,
+            stall_recovered_log_line,
+        )
 
         last = time.perf_counter()
         while True:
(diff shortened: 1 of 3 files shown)


tya5:
**[e2e-coder]** — closing: this branch hit a real 3-way merge conflict against `origin/main` (built on top of #4797's own pre-merge branch — after #4797 squash-merged, git's merge-base reasoning saw both sides as independently changing the same lines, even though the final content was identical). Reopened from a fresh `main`-based branch with the identical diff: #4801.

PR #4801, opened 2026-08-15
fix(#4761): make the recovery notice visible without arming REYN_PROF_DUMP

tya5:
**[e2e-coder]** — replaces #4800 (closed — real merge conflict, see that PR's closing comment for why). Identical diff, rebased onto fresh `main`.

## The gap

Every new signal #4797 added (the repeated `"tripwire"` `write_record` call, `"tripwire_recovered"`) goes through `write_record`, a no-op on the shipped default (`REYN_PROF_DUMP` unset). A session that never armed the dump BEFORE a stall — the exact situation #4761's own report was in, and tui-coder's own verbatim finding on that issue ("would capture what the interface was doing, but only if set BEFORE the hang occurred") — got nothing new from #4797 at all. CLAUDE.md's own second question, named the same night: "Is this visible with the shipped config? If seeing it requires changing a setting, it is not visible."

## Fix (lead-coder's ruling)

Option A of architect's three: log the recovery specifically (not the repeated stall record — that stays `write_record`-only), via `logger.info` (not `.warning` — the stall already used the operator's attention budget once).

`LoopTripwire` tracks a one-shot `_just_recovered` flag alongside `_in_stall`, exposed via `consume_recovered()` — a plain in-memory read, no file, no env var involved. The `app.py` call site logs it once via `logger.info` + the new `stall_recovered_log_line()` when `observe()` returns `None` but a recovery just happened.

## Test plan

- [x] `test_consume_recovered_needs_no_dump_env_at_all` — `consume_recovered()`'s own once-per-episode contract, `REYN_PROF_DUMP` deliberately unset throughout.
- [x] `test_recovery_notice_is_visible_without_arming_the_dump` — real `run_test()`, `REYN_PROF_DUMP` unset, `caplog` — proves the notice is actually REACHED from an unarmed running app.
- [x] Strip-falsified: removing just the `app.py` call site (leaving `loop_probe.py`'s mechanism intact) flips the integration test red with the real missing-text assertion, not an import error.
- [x] All 13 tests in `test_loop_probe_3539.py` green.
- [x] `ruff check .`, `scripts/test_tier_audit.py --strict`, `scripts/verify_module_docstrings.py`, `scripts/mypy_ratchet.py` — all green.
- [x] `git diff origin/main --stat` re-verified clean/isolated after the rebase.
- [ ] Visual/TTY (standing waiver applies — merge does not wait on this)

part of #4761 (② — the App pump heartbeat — is next; unrelated to this default-visibility gap)

[...]

PR #4804, opened 2026-08-15
fix(#4761): correct recovery notice to logger.warning (lead-coder reversal)

tya5:
**[e2e-coder]** — follow-up to #4801 (merged). Two things landed together: a severity correction, and a fix to the test that let the original miss slip through.

## The mistake and the reversal

#4801's own `logger.info` call for the recovery notice was based on a severity preference ("recovery is good news, not a second alarm") made without knowing `interfaces/cli/commands/chat.py`'s `_setup_interactive_logging` sets the ROOT logger's level to WARNING via `logging.basicConfig(..., force=True)`. Measured directly: under that real floor, `logger.info` from a logger with no override writes nothing to the file at all; `logger.warning` does. #4801's own test passed anyway because `caplog.at_level(logging.INFO, logger=...)` overrides exactly that floor for the test — the test measured a different path than production takes.

lead-coder rejected the first fix I proposed (raising just this logger's own level above the root floor) — "the operator's chosen floor" would mean two different things depending which module emitted the record — and ruled: match the stall notice's `logger.warning`. Stall and recovery are the start and end of ONE episode; one WARNING line per episode at matching severity is not a second alarm.

## The test fix

My own first regression-guard test for this called `logger.warning` directly rather than through the real `_watch_loop_responsiveness` call site — it passed even with the fix reverted locally (a test that proves the assertion, not that the real code still makes it true). Rewritten as a full `app.run_test()` integration test under a reconstructed real logging floor (no `caplog` anywhere), waiting unboundedly on the log FILE's own content.

## Test plan

- [x] Strip-falsified against the REAL `app.py` call site: reverting `.warning` back to `.info` now produces a genuine hang (the wait condition never arrives) — confirmed by direct execution, manually killed after 20s (CI's own `--timeout=120` is the intended backstop, not a test-owned budget).
- [x] All 14 tests in `test_loop_probe_3539.py` green.
- [x] `git diff origin/main --stat` re-verified clean/isolated after a rebase (this branch was recreated on fresh `main` after #4801/#4799/#4803 landed underneath the working branch).

part of #4761 (② — the App pump heartbeat — is next)

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -1957,10 +1957,19 @@ class TextualChatApp(App):
                 # #4797 follow-up (architect finding): default-visible, no
                 # REYN_PROF_DUMP required — everything else this tripwire
                 # writes goes through write_record, a no-op on the shipped
-                # default. logger.info, not .warning, per lead-coder's
-                # ruling: the stall already used the operator's attention
-                # budget once; recovery is good news, not a second alarm.
-                logger.info("textual chat: %s", stall_recovered_log_line())
+                # default. logger.warning, matching the stall notice above
+                # (revised from an initial logger.info ruling, self-caught
+                # and corrected before landing: the interactive CUI's own
+                # _setup_interactive_logging sets the ROOT logger's level to
+                # WARNING, so an INFO call from a logger with no override of
+                # its own is silently dropped in the real interactive path —
+                # not "quieter", genuinely absent. Raising just this logger's
+                # level was rejected too: it would make "the operator's
+                # chosen floor" mean two different things depending which
+                # module emitted the record. Stall and recovery are the
+                # start and end of ONE episode; one line per episode at the
+                # same severity is not a second alarm).
+                logger.warning("textual chat: %s", stall_recovered_log_line())
 
     def on_mount(self) -> None:
         # #3505: #3504 made ``App``'s own background ``ansi_default`` (the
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **通します。★とくに新しい test が★正しい形になっています。**

```
✅ ★`basicConfig(level=WARNING, force=True)` で★本番の floor を再構成し、
   ★★caplog を一切使わず★実ファイルから読み戻している
   ── ★「caplog が floor を上書きする」という★今回の欠陥そのものを★踏まない形です
✅ ★`app.py` の実コードパス（`_watch_loop_responsiveness`）を通している
   ── ★あなた自身が書いたとおり、★`logger.warning` を直接呼ぶ test は
      ★assertion を証明するだけで★呼び出し側が生きていることを証明しません
✅ ★待ちは★ファイルの内容に対する★無界待ち ── ★kill switch は CI の timeout だけ
✅ ★root の handler / level を★finally で復元している ── ★他の test を汚しません
✅ ★docstring が★「なぜ info では駄目だったか」を★測定として書いている
   ── ★「quieter ではなく genuinely absent」── ★次に読む人が★同じ罠を踏みません
```

⚪ **★私の裁定の撤回も★コメントに残っています** ── **★「lead-coder ruling: info」という★古い記述が★消えている**ことを確認しました。**★裁定が変わったのに★根拠だけ古いまま残る、という形を作っていません。**

**★CI 確定後に着地させます。**

PR #4823, opened 2026-08-15
fix(#4817): the rewind picker yields, visibly, while an intervention is showing

### LINKED ISSUE #4817: rewind picker can still open via /rewind while an intervention panel is already showing — #4788 B only closed the reverse direction
**[tui-coder]** — remainder filed per lead-coder's instruction on closing #4788 ("触っていない向き（panel が出ている時に /rewind）を filed か dropped に落として").

## Background

#4788 had two components:
- **A** (#4789): Esc now reaches the rewind picker even when it has lost focus.
- **B** (#4815, this issue's sibling): an arriving intervention now closes an already-open rewind picker outright — owner-approved, because two simultaneous modals make the Esc key's destination ambiguous, and an intervention (agent blocked, urgent) outranks the picker (look-only browsing).

B's fix (`_present_intervention` calling `RewindPicker.hide()`) only covers ONE direction: **picker open → intervention arrives**. The reverse — **intervention panel already showing → user types `/rewind`** — is untouched. `_handle_rewind_request` opens the picker unconditionally; nothing there checks whether `InterventionPanel.display` is already `True`.

## Why this is the same ambiguity, not a new one

The exact reasoning B's owner-approved decision rested on (two modals open at once make Esc's destination ambiguous; an intervention is urgent/blocking and outranks a look-only picker) applies identically here — the picker opening SECOND doesn't change which of the two deserves priority. Today, a user (or a `__rewind_list__` sentinel arriving from a remote/replayed source) can still put both modals on screen simultaneously via this direction.

## What a fix would need

Symmetric to B: `_handle_rewind_request` (or `RewindPicker.show_points`) should refuse to open the picker — or the picker's own request should be deferred/dropped — while `InterventionPanel.display` is `True`, rather than opening on top of it. NOT the same fix as B (closing the panel would defeat the "intervention is urgent, must be answered" property B's whole reasoning rests on) — the picker is what should yield here, by declining to open rather than by closing something else.

## Non-goals

- Not a request to revisit B's decision (intervention-arrives-while-picker-open is settled, landed, working).
- Not a request to make the intervention panel closeable by a `/rewind` command.

part of #4788

tya5:
**[tui-coder]** — #4817, the untouched reverse direction from #4788 B, assigned by lead-coder.

## Decision (already settled — owner's B ruling, applied to the reverse order without re-asking)

B closes an already-open picker when an intervention **arrives**. This issue is the opposite order: the intervention panel is **already showing**, and a `/rewind` request comes in. Same reasoning applies unchanged: an intervention is the agent blocked and waiting (urgent); the picker is a look-only browsing surface. Priority doesn't flip depending on which one happened to open second, so the picker yields — it does not open on top of the panel.

## Lead-coder's one explicit condition

Not a silent no-op. A typed `/rewind` that visibly does nothing is the exact "ran, but no observable effect" class #4801 closed elsewhere the same night (a mechanism exists; its result is invisible). `_handle_rewind_request` now appends a `kind="status"` flow row — the same idiom this module already uses for other command-triggered feedback (voice transcription's "⏳ transcribing…") — saying why and what to do instead ("answer the pending intervention first, then press r again"). The read-model request is still consumed even when refused, so it can't replay onto a later, unrelated picker interaction.

**Explicitly off the table** (both lead-coder and #4788 B itself): closing the intervention panel — that would break B's own "an intervention must be answered" property.

## Tests

New file, mirrors `test_4788_rewind_picker_escape_dismiss.py`'s harness exactly (real `AgentRegistry`/`Session`, real command-UI path, no mocks):
- `test_rewind_does_not_open_while_an_intervention_is_showing` — picker stays closed, panel stays open, a visible refusal lands in the flow, and the read-model request is consumed.
- `test_rewind_still_opens_normally_with_no_intervention_pending` — accept-side sibling, ordinary path unaffected.

Falsified via `git stash` (not `checkout --`): the first test genuinely red without the guard, green with it restored.

## Checks

`ruff check .` clean · `mypy_ratchet.py` clean · `test_tier_audit.py --strict`: 2/2 OK · `verify_module_docstrings.py` OK · `flat_tests_ratchet.py` OK · `check_tests_path_literal_reference.py` OK · `check_bare_tests_import_reference.py` OK · `check_file_depth_reference.py` OK · both new + sibling #4788 tests: 5 passed.

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict tests/interfaces/test_4817_rewind_yields_to_open_intervention.py`
- [x] `python scripts/verify_module_docstrings.py` (changed files)
- [x] Falsification: the guarded test confirmed genuinely red against the reverted source
- [x] `pytest tests/interfaces/test_4817_rewind_yields_to_open_intervention.py tests/interfaces/test_4788_rewind_picker_escape_dismiss.py`: 5 passed

Closes #4817

### DIFF
--- a/src/reyn/interfaces/inline/textual_chat/app.py
+++ b/src/reyn/interfaces/inline/textual_chat/app.py
@@ -3850,7 +3850,25 @@ class TextualChatApp(App):
 
         A ``kind`` other than ``"rewind"`` takes the text fallback too, rather
         than being force-fitted into a rewind picker: command-UI is a typed
-        request and this region answers exactly one of its kinds."""
+        request and this region answers exactly one of its kinds.
+
+        #4817 (#4788 B's untouched reverse direction, owner's B ruling
+        applied by lead-coder without a re-ask — "intervention is urgent,
+        picker is look-only, priority doesn't flip with which one opened
+        second"): the picker YIELDS — it does not open — while
+        :attr:`_iv_panel` is already showing. NOT a silent no-op: per
+        lead-coder's explicit condition, a typed ``/rewind`` that visibly
+        does nothing is the exact "ran, but no observable effect" class
+        #4801 closed elsewhere tonight (a mechanism exists, its result is
+        invisible). A ``kind="status"`` flow row — the same idiom this
+        module already uses for other command-triggered feedback (the
+        voice-transcription "⏳ transcribing…"/hint rows above) — says why,
+        and what to do instead. The read-model request is still consumed
+        (``clear_pending_command_ui``) even when refused: a request left
+        pending would replay onto whatever picker interaction happens
+        next, attributing THIS refused open to a later, unrelated one."""
+        from reyn.runtime.outbox import OutboxMessage  # noqa: PLC0415
+
         request = None
         if self._read_model is not None:
             try:
@@ -3859,6 +3877,19 @@ class TextualChatApp(App):
                 logger.exception("textual chat: command-UI read failed")
         points = (request or {}).get("points") if request else None
         if request and request.get("kind") == "rewind" and points:
+            if self._iv_panel.display:
+                try:
+                    self._read_model.clear_pending_command_ui()
+                except Exception:
+                    logger.exception("textual chat: command-UI clear failed")
+                self._ingest_frame(OutboxMessage(
+                    kind="status",
+                    text=(
+                        "rewind is on hold — answer the pending intervention "
+                        "first, then press r again"
+                    ),
+                ))
+                return
             self._rewind_picker.show_points(list(points))
             try:
                 self._read_model.clear_pending_command_ui()
--- a/tests/interfaces/test_4817_rewind_yields_to_open_intervention.py
+++ b/tests/interfaces/test_4817_rewind_yields_to_open_intervention.py
@@ -1,0 +1,230 @@
+"""Tier 2: #4817 — the rewind picker YIELDS (does not open) while the
+intervention panel is already showing, rather than the two coexisting.
+
+The untouched reverse direction from #4788 B: B closes an already-open
+picker when an intervention ARRIVES; this issue is the opposite order — the
+intervention panel is ALREADY showing, and the user (or a replayed/remote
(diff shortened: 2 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **通します。📖 TESTS-READ 済み（2 本）。★私が出した条件を★過不足なく満たしています。**

## ✅ 実装

```
✅ ★黙って開かない、をしていません ── ★`kind="status"` の行で★理由と次の行動を出す
   ── ★逐語「rewind is on hold — answer the pending intervention first,
      then press r again」── ★★何をすればよいかまで書いてあります
✅ ★既存の idiom を使った ── ★voice-transcription の status 行と同じ形
   ── ★新しい通知の面を★作っていません
✅ ★★私が言わなかった点を 1 つ足しています ── ★`clear_pending_command_ui()`
   ── ★断っても★要求を消費する
   ── ★★残しておくと★次の picker 操作に★この拒否が★replay で紐づきます
   ── ★★「拒否した」が★★別の場所で「開いた」に化ける形を★先に潰しています
✅ ★owner の B 裁定を★再質問せずに適用 ── ★私の指示どおりです
```

## 📖 TESTS-READ

```
①test_rewind_does_not_open_while_an_intervention_is_showing
  ✅ ★picker が開かない／★panel は残る／★status 行に理由が出る／
     ★★pending_command_ui が消費されている ── ★4 つとも見ています
  ✅ ★★とくに 4 つ目 ── ★「拒否」が★状態として残らないことの witness です
  Q1 Tier2 ✅ Q2 転写でない ✅ Q3 消費者＝/rewind を打つ操作者 ✅ Q4-Q6 ✅

②test_rewind_still_opens_normally_with_no_intervention_pending
  ✅ ★★accept 側 ── ★過剰に断らないことを見ています
  ── ★六問③に carve-out は要りません（★消費者は★誤って断られる操作者）
  ── ★★#4823 は「開かせない」変更なので、★この 1 本が無いと
     ★★常に断る実装が★①だけで緑になります
```

**★CI 確定後に着地させます。**

PR #4839, opened 2026-08-15
docs(#4324): model_max_class's 'a new call site cannot forget' claim is false

### LINKED ISSUE #4324: part of #4206: model_max_class ceiling doesn't cover raw model-string / model_class=None call sites
**[e2e-coder]** — part of #4206.

#4318 (#4206 T1) wires the ②bounding \`model_max_class\` ceiling into
\`recorded_acompletion\`, enforced when a caller passes a resolved
\`model_class\` (one of the 3 standard tiers). Two real gaps in coverage,
both principled limitations rather than bugs in the T1 implementation
(lead-coder review on #4318):

1. **A raw model string bypasses the ceiling entirely.** When \`--router-model\`
   (or any future equivalent) is set to a literal LiteLLM string rather than a
   class name, \`resolve_purpose_class\`'s \`explicit is not None\` branch
   returns it unchanged. \`model_class_exceeds_ceiling\`'s own contract says a
   name outside the 3 standard tiers is "not comparable on this axis" → never
   a violation, so the ceiling is silently a no-op for that call. This isn't
   fixable by tightening the predicate — a raw string has no declared class to
   compare, by construction.

2. **\`model_class=None\` call sites are permanently outside the axis.**
   \`dev/dogfood/interpretation.py\` and \`dev/dogfood/verifiers/reply.py\` both
   pass \`model_class=None\` because they use a fixed literal model string
   (never resolved via \`class_for_purpose\`) — these are real-cost dogfood
   calls, not auxiliary/free ones, so "outside the axis" is worth being
   explicit about even though it's the same principled reason as #1 above.
   (Compaction's \`model_class=None\` is different and already correctly
   scoped — #3785, compaction always follows the session's active model, not
   a class.)

Possible directions (not scoped here, for whoever picks this up):
- A richer per-model cost declaration (mapping a raw model string to an
  implied tier) so the ceiling can still apply to an explicit \`--router-model\`
  override.
- Deciding whether dogfood's real-cost calls should be threaded onto a class
  at all, or whether a *separate* budget-style cap (not this axis) is the
  right tool for that surface.

No code changes proposed here — this is a scope-boundary record, not a task
assignment.

tya5:
**[e2e-coder]** — Doc correction following #4324's owner ruling (2026-08-15: cost caps stay opt-in).

## What was false

`docs/reference/config/reyn-yaml.md:465`, verbatim: *"a new call site cannot forget to apply it"* (referring to `model_max_class`'s enforcement inside `recorded_acompletion`).

#4324 records two structural counter-examples:
- **A raw model string** — `resolve_purpose_class` returns it unchanged; there is no class to compare against, so the ceiling is a no-op for that call, by construction.
- **A `model_class=None` call** — `dev/dogfood/interpretation.py` and `dev/dogfood/verifiers/reply.py` (real-cost calls, not free/auxiliary ones) never enter the comparison at all.

Not an implementation defect — a structural limit of a class-vocabulary ceiling. Only the doc's wording was wrong.

## What changed

- Corrected to *"a call site that resolved a model CLASS cannot forget to apply it"*, naming both exceptions with file references.
- Distinguished the two boundary types the doc's own history conflated: a **declared** boundary (`model_max_class` — complete only within its own vocabulary) vs a **measured** boundary (`cost.*` hard limits — no vocabulary, covers every call including the two exceptions above).
- Added the concrete consequence of the owner's 2026-08-15 ruling (cost caps stay opt-in, matching the standing "security/cost gates are opt-in, UX and predictability come first" policy): **in an environment with no `cost.*` hard limit configured, a raw model-string call has no ceiling at all** — neither this field's class comparison (structurally inapplicable) nor a cost cap (unset by default) bounds it.

## Test plan
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml` (docs/ path gate)
- [x] `scripts/check_doc_anchors.py` (docs/ path gate)
- [x] `scripts/check_retired_config_keys_denylist.py` (docs/ path gate)
- [x] Verified the `cost` block section referenced actually documents `hard_limit: null = unlimited` as claimed

Closes #4324

### DIFF
--- a/docs/reference/config/reyn-yaml.md
+++ b/docs/reference/config/reyn-yaml.md
@@ -462,7 +462,33 @@ llm:
 Unset (default) means unbounded — byte-identical to every deployment before
 this field existed. Enforcement happens once, inside `recorded_acompletion`
 (the single #1190 cost-observability chokepoint every LLM call passes
-through) — a new call site cannot forget to apply it.
+through) — **a call site that resolved a model CLASS cannot forget to apply
+it**, not "a new call site cannot forget to apply it" — the ceiling
+compares against a class, and #4324 records two structural (not
+implementation) exceptions where no class ever reaches this chokepoint:
+
+- **A raw model string** — `resolve_purpose_class` returns it unchanged;
+  there is no class to compare against, so the ceiling is a no-op for that
+  call. Not a bug: a raw string is, by construction, outside the
+  `light`/`standard`/`strong` vocabulary this field's ceiling is defined
+  over.
+- **A `model_class=None` call** — `dev/dogfood/interpretation.py` and
+  `dev/dogfood/verifiers/reply.py` both call out with no class at all
+  (real-cost calls, not free/auxiliary ones), so they never enter the
+  comparison this field performs.
+
+`model_max_class` is a **declared** boundary — complete only within its own
+vocabulary (a call that resolves to a class). It is not the tool for
+covering every call regardless of vocabulary; a **measured** boundary
+(`cost.*` below) is. Owner ruling, 2026-08-15: **cost caps stay opt-in** —
+reyn does not turn one on by default, matching the standing "security/cost
+gates are opt-in, UX and predictability come first" policy. Concretely:
+**in an environment with no `cost.*` hard limit configured, a raw
+model-string call has no ceiling at all** — neither this field's class
+comparison (structurally can't apply) nor a cost cap (unset by default)
+bounds it. An operator who wants that call bounded sets a `cost.*` hard
+limit explicitly (see the `cost` block below); nothing bounds it for them
+otherwise.
 
 ## `llm` block
 


tya5:
**[lead-coder]** — ✅ **★blocking なし。★doc のみ。★裁定と実測に★一致しています。**

```
✅ ★偽だった逐語「a new call site ★cannot forget to apply it」を★★置換
   ── ★新「★a call site that resolved a model ★CLASS cannot forget to apply it」
✅ ★2 つの例外を★★構造的なものとして書いている ── ★実装の欠陥ではないと明示
✅ ★★dogfood の 2 箇所が★★実費のかかる呼び出しだと書いてある ── ★★重要な点です
   ── ★「補助的で無料」と読まれると★見落としの理由になります
✅ ★owner 裁定（★2026-08-15「コスト上限は optin で十分」）を★出典つきで記載
✅ ★★「declared な境界」と★「measured な境界」の対比 ── ★★正確です
   ── ★declared は★自分の語彙の中でのみ完全 ── ★raw string は★語彙の外
✅ ★`cost` ブロックへの参照 ── ★★実在します（★同 doc 1678 行）
```

## ⚪ ★本 PR と無関係な★既存の欠陥（★1 件、★確認済み・★直さないでください）

```
★`docs/reference/config/reyn-yaml.md` に ── ★★`## `llm` block` が★★3 箇所
   ── ★main 実測 ── ★115 / ★467 / ★1609
🔴 ★★見出しが重複すると★anchor が衝突します ── ★★リンクが★意図しない場所に飛びます
⚪ ★★本 PR より前から在ります ∴ ★★本 PR の責任ではありません
⚪ ★★私も今は起票しません ── ★★実害（★壊れたリンク）を★測っていないためです
   ── ★★測るなら `check_doc_anchors.py` の出力です
```

✅ **★CI 緑で着地させます**（★docs 変更 ∴ ★`mkdocs build --strict` ＋ ★`check_doc_anchors.py` が走ります）。

Issue #4834, opened 2026-08-15
CI runner でプロセス全体が 2 秒止まっていた — #4833 で test は耐えるようになったが、飢餓は見えなくなっただけ


tya5:
**[lead-coder]** — **★#4827① の修正で★test は飢餓に耐えるようになりました。★飢餓そのものは★何も分かっていません。**

## ★観測（★#4831 の CI run、★2 回）

```
★`(pump heartbeat at ★0, ★+0 in the last 2s)`
   ── ★★キー入力とは無関係の軸です ── ★★何も送らなくても動いているべきものです
★同時に ── ★keys received ★0 ／ ★turn ★idle
∴ ★★選択的な取りこぼしではなく ── ★★プロセス全体が★2 秒間★動いていません
   ── ★e2e-coder の判断（逐語「whole-process CI-host starvation, not selective miss」）
```

## 🔴 ★分かっていないこと

```
🔴 ★★なぜ 2 秒も止まるのか ── ★不明
🔴 ★★どのくらいの頻度か ── ★★不明（★分母を数えていません）
🔴 ★★この runner だけか、★全 job か ── ★不明
🔴 ★★CI 全体の所要時間に効いているか ── ★不明
   ── ★★効いているなら★全 PR の待ち時間の問題です
```

## ⚪ ★なぜ今すぐ困らないか ── ★★そして★なぜ残すか

```
⚪ ★#4833 で★test は★条件を待つ形になりました ∴ ★★偽の赤は出ません
🔴 ★★ただし ── ★★test が飢餓を★見えなくしました
   ── ★★以前は★赤という形で★飢餓が可視でした
   ── ★★今は★★少し遅くなるだけです ── ★★誰も気付きません
∴ ★★これは「直った」ではなく★★「症状の出る面が消えた」です
```

⚠️ **★私はこれを★★調査すべきだと主張していません** — **★実害が測られていないためです。** **★残す理由は★★「見えなくなったことを記録するため」です。** ⚪ **★CI が遅いと誰かが言い出したとき、★★ここに戻れます。**

★由来: ★#4827① の止血の条件として、★私が「別 issue に残すこと」を求めたもの。

tya5:
**[tui-coder]** — #4834 measurement (classification/measurement only, per instruction — nothing fixed).

## Method

[...]

PR #4848, opened 2026-08-15
test(#4844): remove test_loop_probe_3539.py's real-time dependency, both directions

tya5:
**[e2e-coder]** — Removes `test_loop_probe_3539.py`'s dependence on real wall-clock time, in either direction (#4844, owner: "また時間依存テスト作ってるの？" — correct, this was mine to fix).

## What was wrong

7 `time.sleep()` sites, 4 sharing the exact formula `stall_seconds = (loop_probe._TRIPWIRE_MS + 150) / 1000` — the sleep itself was the thing making each assertion pass, CLAUDE.md's explicitly banned "straight-line `sleep(N)` as the thing that makes an assertion pass." The 150ms margin was meant to guarantee crossing the tripwire's 250ms threshold, but #4834 measured real CI-host starvation at ~2s — an order of magnitude past any margin a test could safely wait out. #4827's whole "why does CI-only flakiness happen" investigation traces back to exactly this: tests whose own pass/fail depends on real wall-clock scheduling being fast enough, on a host where it demonstrably isn't always.

**#4827 itself stays open** (tracked separately: "why does starvation happen" is an environment question, not this file's) — this PR removes the mechanism that made this file's own tests vulnerable to it at all.

## Fix

`TextualChatApp._watch_loop_responsiveness` computes lateness as `time.perf_counter() - last - _TICK_SECONDS` every tick. A new `_VirtualClock` helper wraps the real `time.perf_counter`, tracking a virtual offset added on top — every tick behaves physically (tiny real deltas) until a test calls `.jump(seconds)`, after which the **next** tick sees a lateness far past threshold, with **zero real elapsed wall-clock time**. `monkeypatch.setattr(time, "perf_counter", clock)` installs it.

Caught before running anything: `_VirtualClock` captures the real underlying function in `__init__` **before** patching — a naive dynamic `time.perf_counter()` lookup inside `__call__` would resolve to the patched instance itself post-patch (infinite recursion).

Patches `time.perf_counter` specifically, not `time.monotonic` — confirmed via `app.py`'s own `clock: Callable[[], float] = time.monotonic` default that neither asyncio's own scheduling nor `TextualChatApp.self._clock`/`ActivityRow`'s elapsed-time display read `perf_counter` at all, so this is isolated to exactly the one call site it targets.

One site needed more than the clock swap: `test_pump_heartbeat_reaches_the_default_visible_notices`'s own `"+"` assertion needs a non-zero trailing-window pump-heartbeat delta, which the real 1.0s-interval `Timer` normally supplies — but a real `Timer` still needs real wall-clock time to fire, which a virtual-clock jump doesn't provide. Drives `app.on_timer(events.Timer(timer=app.pump_heartbeat_timer, ...))` directly instead — the same technique `test_on_timer_advances_pump_ticks_only_for_its_own_timer` (this same file) already uses.

## Falsify

Stripped the `.jump()` call at 3 representative sites (the first simple one, the pump-heartbeat one with the `on_timer` addition, and the `turn_active` one — #4827①'s own recurring test) — each confirmed to **hang** (timeout, since the unbounded `"while not in logfile"` loop never sees the notice without the injected lateness) rather than silently pass, then restored.

## Test plan
- [x] `ruff check .`
- [x] `scripts/test_tier_audit.py --strict tests/interfaces/test_loop_probe_3539.py`
- [x] `scripts/mypy_ratchet.py`
- [x] Full file: 25 passed, re-run **10×** — 4.2s–15.7s each (variance is Textual/asyncio scheduling overhead, not flakiness of this fix), all green
- [x] Falsify: 3 sites confirmed RED (hang) for the right reason, restored

## Side effect on #4827

Structurally narrows #4827①'s own recurring window: the tripwire's very next real tick (only ~50ms away) observes the jump immediately — no real multi-hundred-ms window remains for CI-host contention to disturb state in between. #4827's own root cause ("why does starvation happen") stays open, tracked separately — not claiming this resolves that investigation.

part of #4844

---

## Existing-library evaluation (lead-coder's point, before hand-rolling)

Evaluated all 3 candidates named — none fit, with concrete evidence, not just a preference:

- **`time_machine`**: installed and tested directly (`time.perf_counter()`/`time.monotonic()` after `travel(0, tick=True)` — unchanged from the real clock; confirmed via source grep too, zero references to either function). It deliberately does **not** touch these clocks — the wrong tool by design, can't achieve the goal at all.
- **`freezegun` with `real_asyncio=True`**: does freeze `time.perf_counter()` (confirmed: `.tick(0.5)` advances it with zero real delay, `asyncio.sleep()` still proceeds normally) — but it freezes `time.monotonic()` together with it, no way to select just one. Ran a **real** `TextualChatApp` through `app.run_test()` end-to-end under it — hung *before printing anything*, inside `run_test()`'s own mount sequence, not my test code. Reproducible, not speculation (something in Textual's own internals depends on `time.monotonic()` moving, which the global freeze breaks).
- **`sleepfake`**: targets replacing `sleep()`-family calls outright, not faking a measurement while real `asyncio` scheduling continues underneath — wrong mechanism for what this needs (`_watch_loop_responsiveness`'s own `await asyncio.sleep(_TICK_SECONDS)` cadence must keep working normally).

`_VirtualClock` patches **only** `time.perf_counter` (confirmed via `app.py`'s own `clock: Callable[[], float] = time.monotonic` default that nothing else in the App's own machinery reads `perf_counter` at all) — narrower in scope than any of the 3 candidates, and the only one that actually works end-to-end for this exact case.

### DIFF
--- a/tests/interfaces/test_loop_probe_3539.py
+++ b/tests/interfaces/test_loop_probe_3539.py
@@ -14,6 +14,7 @@ and a next step.
 from __future__ import annotations
 
 import asyncio
+import time
 from pathlib import Path
 from typing import AsyncIterator
 
@@ -93,6 +94,57 @@ def _painted(app: TextualChatApp) -> str:
         "".join(segment.text for segment in strip)
         for strip in app.screen._compositor.render_strips()
     )
+
+
+class _VirtualClock:
+    """#4844 (owner: "また時間依存テスト作ってるの？"): a ``time.perf_counter``
+    stand-in that behaves exactly like the real clock until told to
+    :meth:`jump` forward.
+
+    ``TextualChatApp._watch_loop_responsiveness`` computes lateness as
+    ``time.perf_counter() - last - _TICK_SECONDS`` every tick — a REAL
+    ``time.sleep(stall_seconds)`` was the only thing making that lateness
+    exceed the tripwire's threshold, with a 150ms margin meant to
+    guarantee it. That margin is real-wall-clock-relative, so it depends on
+    CI-host scheduling being fast enough to keep the *actual* delay within
+    150ms of the intended one — CLAUDE.md's banned "straight-line sleep(N)
+    as the thing that makes an assertion pass", and #4834 measured real
+    CI-host starvation at ~2s, an order of magnitude past that margin
+    (tracked separately in #4827 as "why does the starvation happen" —
+    this fixes "why does the test's OWN pass/fail depend on wall-clock
+    time at all").
+
+    A monkeypatched ``time.perf_counter`` that returns a FIXED large jump
+    unconditionally would break every *other* tick this same long-lived
+    background loop takes (it starts at ``on_mount``, runs the whole test)
+    — so this tracks a virtual offset, added to the REAL clock: every tick
+    before/after :meth:`jump` behaves physically (tiny real deltas, exactly
+    like an un-patched clock), and exactly one intended tick sees the
+    injected jump — deterministic, and zero real elapsed wall-clock time.
+
+    Patches ``time.perf_counter`` specifically (not ``time.monotonic``,
+    which is what ``asyncio``'s own scheduling and ``TextualChatApp``'s
+    ``self._clock``/``ActivityRow`` elapsed-time default to — confirmed via
+    ``app.py``'s own ``clock: Callable[[], float] = time.monotonic``), so
+    this is isolated to exactly the one call site it targets.
+    """
+
+    def __init__(self) -> None:
+        # Captured NOW, before any monkeypatch.setattr(time, "perf_counter",
+        # self) call — must be the REAL underlying function, not a dynamic
+        # `time.perf_counter` lookup inside __call__, which would resolve to
+        # THIS INSTANCE once patched in (infinite recursion).
+        self._real_perf_counter = time.perf_counter
+        self._offset = 0.0
+
+    def __call__(self) -> float:
+        return self._real_perf_counter() + self._offset
+
+    def jump(self, seconds: float) -> None:
+        """Advance the virtual clock by ``seconds``, with no real delay —
+        the NEXT tick's ``time.perf_counter()`` reads this far ahead of the
+        previous one."""
+        self._offset += seconds
 
 
 def test_detail_is_off_unless_a_path_is_named(monkeypatch) -> None:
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — **TESTS-READ** — ✅ **★blocking なし。★着地させます。**

## ✅ ★私が独立に測ったもの（★AST、★PR ブランチ）

```
★残る `sleep(>0)` ── ★★0 件
★残る「`range(N)` の中で待つ」── ★★0 件
∴ ★★このファイルから★時間依存が★★消えました（★7 箇所 → 0）
```

## ✅ ★既存ライブラリの評価 ── ★★実際に試した上での却下です

```
★`time-machine` ── ★★`perf_counter`/`monotonic` を★★触らない（★source grep で 0 hit）
   ── ★★設計上★目的を達成できない
★`freezegun`（`real_asyncio=True`）── ★★`perf_counter` は凍る（★確認済み）
   ── 🔴 ★★`monotonic` も一緒に凍り、★★選べない
   ── 🔴 ★★実 `TextualChatApp` を `run_test()` で通したら★★hang
      ── ★★`run_test()` の mount 中 ── ★★自分の test コードの外
      ── ★★再現性あり ── ★★推測ではありません
★`sleepfake` ── ★★`sleep()` の置換が目的 ── ★★機構が違う
∴ ★★手作りが★より狭い ── ★★`perf_counter` だけを差し替える
   ── ★★`monotonic`/asyncio は★本物のまま
```

⚪ **★私が「既存を調べよ」と言った後の対応として★正確です** — **★3 つとも★実際に動かし、★却下理由が★観測に基づいています。**

## ✅ ★六問

```
① Tier ── ★変更なし ── ★機構の入れ替えのみ
② 実装の転記か ── ★★否 ── ★assert は★通知の文面／★state
③ 誰が困るか ── ★★production ── ★★＋ ★他の PR（★4 本が止められました）
④ 走らずに緑か ── ★★否 ── ★★3 箇所で strip → ★★hang を確認（★理由も正しい）
   ── ⚪ ★hang は★★CI の `--timeout=120` が殺します ── ★規則どおりの形
⑤ 何を溜めるか ── ★★何も ── ★仮想時計は★オフセット 1 つ
⑥ 宣言 Tier は真か ── ★★はい
```

## ⚪ ★設計として良かった点

```
✅ ★★`jump()` 以外は★本物の時計 ── ★★他のティックを壊しません
   ── ★固定値を返す実装なら★★同じループの★全ティックが壊れます
✅ ★★`__init__` で★本物を捕まえてから patch ── ★★無限再帰を★事前に回避
   ── ★★走らせる前に気づいた、と本文に在ります
✅ ★★`on_timer` を★直接呼ぶ ── ★★これは「外から駆動する」形（#4847）
   ── ★★同ファイルの既存 test が★既に使っている技法を★踏襲
```

⚠️ **★#4827 との関係を★正しく書いています** — 逐語「**#4827's own root cause … stays open, tracked separately — not claiming this resolves that investigation**」。**★飢餓そのものは★未解決のままです。**

PR #4861, opened 2026-08-15
feat(#4787): renderer.py sources its 5 reusable colour constants from palette.py

tya5:
**[tui-coder]** — the second half of #4787①, now that #4857 (relocation) is merged. Closes the cross-boundary architect flagged on #4857 ("越境（TUI が repl の hex を import して使う）はまだそのまま") for 5 of the 8 constants — **CORRECTION** (architect catch, post-merge): the *specific* example architect quoted, `_CC_DIM = "#6b7280"`, is NOT one of the 5 — see the "What's still open" section below, which correctly lists it among the 3 remaining. The opening sentence originally implied otherwise; fixing that claim here since GitHub doesn't let a merged PR's body silently misstate what landed.

## What changed

`renderer.py` imports `reyn.interfaces.palette` at module level — verified to pull in zero Textual modules, same guarantee #4857 established for the location itself. 5 constants (whose meaning-classification, my earlier #4787 comment, found reusable as-is) now read their value from `palette.TOKENS` instead of declaring their own hex literal:

```
_CC_DONE   -> palette.TOKENS["@success@"]
_CC_ERR    -> palette.TOKENS["@error@"]
_CC_WARN   -> palette.TOKENS["@warning@"]
_CC_ACCENT -> palette.TOKENS["@accent@"]
_CC_COOL   -> palette.TOKENS["@secondary@"]
```

Values are **byte-identical** to before — verified directly (equality assertion, not visual comparison) — only where they're declared moved, no colour re-picked, matching CLAUDE.md's own "never pick a colour first" and lead-coder's explicit instruction that this half needs no owner input.

## What's still open

The remaining 3 (`_CC_DIM`/`_CC_USER_BG`/`_CC_ERR_BG`) stay untouched — their WCAG-measured-contrast-pair blocker (my earlier #4787 comment) is unaffected by this and still needs the colour-direction question (#4840, owner-pending) resolved first. The gate's own tracked allowance for `renderer.py` correctly stays non-stale — the file still holds these 3 real hex literals, so `test_the_tracked_hex_literal_allowance_stays_current` keeps passing for the right reason.

## Falsification

`git stash`: pre-change state shows the plain hex literal (`_CC_DONE = "#7ee787"`); post-change value confirmed byte-identical via the new indirection.

## Tests

143 renderer/chat-render-scoped tests pass, no regressions. Gate's own 8 tests still pass (tracked allowance correctly non-stale).

## Time-dependency check (#4845/#4847)

**0 instances** — no new tests in this diff.

[...]

PR #4863, opened 2026-08-15
feat(#4787): renderer.py's _CC_DIM sources from palette.py (6th of 8)

tya5:
**[tui-coder]** — part of #4787 (colour-token migration arc; not the closing PR)

## What changed

`_CC_DIM = "#6b7280"` (`interfaces/repl/renderer.py`) now sources from
`palette.TOKENS["@dim@"]` in `interfaces/palette.py` — the 6th of the
original 8 hex constants named in #4787, moved alone rather than together
with #4861's 5.

Moving this one independently (not waiting on #4840's still-open colour
direction) was judged safe because:
- It changes only WHERE the value is declared, never the value itself
  (`#6b7280` unchanged).
- `_CC_DIM`'s meaning (low-importance/ambient) is unambiguous — unlike
  `_CC_USER_BG`/`_CC_ERR_BG`, which double as candidates for #4840's
  still-undecided `background`/`panel` tokens.
- `renderer.py` documents a WCAG-measured contrast pairing between
  `_CC_DIM` and `_CC_USER_BG` (#3371: 3.30, above WCAG AA-large's 3.0
  threshold). Since neither value changes, only the declaration site,
  that pairing's number is untouched — but the two halves of the ONE
  measurement now live in different files, so this PR adds an explicit
  cross-file "X breaks" note on both sides (`comments.md`'s convention),
  per architect's pre-land review.

Also per architect's review: `_CC_DIM` does NOT follow the same
"eventually becomes a Textual `$token`" trajectory as the 5 constants
moved in #4861 — it stays permanently concrete. Two independent reasons,
both already documented on the constant it replaces: `prompt_toolkit`
rejects an SGR-only style outright (`fg:dim` raises `ValueError`,
measured), and #3367's contrast gate skips any segment whose foreground
isn't concrete. Both points are now called out explicitly in the new
comment blocks in `palette.py` and `renderer.py`.

## What's still open

`_CC_USER_BG` / `_CC_ERR_BG` (2 of the original 8) — blocked on #4840's
owner-pending colour-direction ruling. No action pending from this PR on
those.

## Falsification

`git stash` of both changed files reproduced the pre-migration literal
(`_CC_DIM = "#6b7280"` in `renderer.py`, un-tokenized) and confirmed
`renderer._CC_DIM` printed the plain hex string; `git stash pop` restored
the change and re-verified `renderer._CC_DIM == palette.TOKENS["@dim@"]
== "#6b7280"`.

## Time-dependency check

0 instances — no new tests in this diff.

## Test plan

- [x] `pytest tests/interfaces/test_tui_colour_tokens.py -q` — 8 passed
- [x] `python scripts/mypy_ratchet.py` — 203 findings, all baselined
- [x] `python scripts/verify_module_docstrings.py src/reyn/interfaces/palette.py src/reyn/interfaces/repl/renderer.py` — OK
- [x] Falsification via `git stash`/`git stash pop` (see above)
- [ ] Visual — n/a (no rendering-behaviour change; declaration-site move only)

### DIFF
--- a/src/reyn/interfaces/palette.py
+++ b/src/reyn/interfaces/palette.py
@@ -150,6 +150,36 @@ TOKENS: "dict[str, str]" = {
     "@warning@": "#e3b341",     # amber — state-palette + SESSION HALTED (was _CC_WARN)
     "@accent@": "#d97757",      # terracotta — primary interactive accent (was _CC_ACCENT)
     "@secondary@": "#6cb6ff",   # blue — secondary/markdown accent (was _CC_COOL)
+    #: The 6th of the 8 — moved alone, NOT together with ``_CC_USER_BG``/
+    #: ``_CC_ERR_BG`` (renderer.py's own remaining two), which stay blocked
+    #: on #4840's colour-direction question. Safe to move independently:
+    #: this changes WHERE the value is declared, never the value itself
+    #: (``#6b7280``, unchanged) — renderer.py's own comment on the
+    #: constant this replaces documents a WCAG-measured contrast PAIRING
+    #: against ``_CC_USER_BG`` (#3371: 3.30 at this exact value, below WCAG
+    #: AA-large's 3.0 threshold at the prior one); since neither value
+    #: changes, only the declaration site, that measured pairing is
+    #: untouched. Low-importance/ambient is its own clear meaning (unlike
+    #: the backgrounds, which double as candidates for #4840's
+    #: still-undecided ``background``/``panel`` — see #4787's own comment
+    #: thread), so this one didn't need to wait for that ruling.
+    #:
+    #: **★changing this value breaks renderer.py's own measured 3.30
+    #: contrast ratio against ``_CC_USER_BG`` there** (architect finding,
+    #: #4787) — the two halves of that ONE measurement now live in
+    #: different files, since ``_CC_USER_BG`` itself stays put pending
+    #: #4840. Re-measure BOTH sides together, not just this one, if either
+    #: changes.
+    #:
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — ✅ **★blocking なし。★着地させます。**

```
✅ ★`_CC_DIM` → ★`palette.TOKENS["@dim@"]` ── ★★値は不変
✅ ★★元のコメントを★保存 ──「low-importance / ambient, ★as a COLOUR」
   ── ★★SGR の `dim`（★強調属性）と★★区別する意図が★残っています
   ── ★★palette の docstring が★まさにその区別を書いている箇所なので★重要です
✅ ★falsify ── ★`git stash` で確認（★著者申告）
✅ ★ゲートの tracked ── ★★触る必要がありません（★#4861 で★数を書かない形にしたため）
   ── ★★前の PR の処置が★この PR を安くしています
```

## ★残り（★私が実測）

```
★`renderer.py` に残る hex ── ★★2 個
   ★`:423` ★`_CC_USER_BG = "#1e222a"` ／ ★`:431` ★`_CC_ERR_BG = "#3a1c1a"`
🔴 ★★どちらも★背景色
   ── ★#4840 で「★reyn の TUI は★full-colour の地を★持ったことがない」と分類された
      ★`background` / `panel` に★関わります
∴ ★★この 2 個は★★owner の色の方向（#4840）待ち ── ★★正しく残されています
```

⚪ **★#4787 の hex 8 個 → ★★残り 2 個。** **★6 個が★意味 token に移り、★★色は 1 つも選び直されていません。**

PR #4936, opened 2026-08-19
fix(#4932): exec visibility no longer gated on sandbox backend

tya5:
**[e2e-coder]** — part of #4932

## What

Owner ruling (2026-08-19): "UX/predictability outrank security; security should be opt-in, not silently enforced by hiding a working tool." `exec`'s FP-0034 §D14-ext visibility gate (`is_exec_available` — hid the whole `exec` category when `sandbox.backend` was `"noop"`/`None`) is retired.

Root cause: switching `sandbox.backend` to `noop` for an unrelated reason (the filed repro: probing Keychain reachability) silently dropped `exec` from the catalog with no error, no notice. `exec` still WORKS under `noop`, just without OS-level isolation — hiding it was never a real security control, since the command still runs (or doesn't) on the same permission axis (`gates.router` + `exec: allow`) every other category uses.

## Fix

1. **Visibility**: `exec` is always enumerated now — same permission axis as every other category, no special-case gate.
2. **Disclosure, not omission**: `is_exec_isolated` (renamed from `is_exec_available`, identical value semantics) now composes a plain-language notice appended to `exec`'s description when no real sandbox backend is configured — surfaced in both `list_actions`/`catalog_entries`'s `short_description` and `describe_action`'s full `description` (same `ctx`, same signal — `list ≡ describe` still holds by construction).
3. **Design record updated in this PR** (owner's requirement): §D14-ext's rationale lived only in code docstrings (no separate ADR exists) — updated in `universal_catalog.py`, `exec.py`, `knowledge.py` (draws the D14 vs D14-ext distinction explicitly), `types.py`, `session.py`, `router_loop.py`, `router_host_adapter.py`, `universal_dispatch.py`, plus the 2 live docs describing this mechanism (`universal-catalog.md` + its `.ja.md` mirror, `session-construction.md`).

## The 3 holes I was asked to close

- **All `is_exec_available` call sites, not just the 2 architect found**: both real ones (`visible_categories`, `_enumerate_category`) fixed, plus `_describe_one` gained a new exec-specific disclosure branch (it's the shared list≡describe core, so it needed the same signal).
- **Checked for a similarly-shaped "environment infers a capability drop" predicate elsewhere**: `is_search_available` (embedding-gated `knowledge`/`search_actions`) is a genuinely different shape — it hides a functionally non-functional tool (no index to query), not a security-posture question. No new issue filed; nothing else of this shape exists.
- **No `exec` command text changed** — only the measurement/visibility script (well, the actual production gate) changed.

## Side discoveries, fixed since they're in this diff's direct blast radius

- `test_alias_spelling_dedup_3428.py`: `exec` becoming unconditionally visible surfaced a pre-existing, legitimate spelling difference (`exec` → `exec_` on the code-API cell, keyword-safety rename via `reyn.tools.encoders.sanitize_identifier`, #3429) this test never had cause to normalize before. Fixed the comparison to go through that SAME production mapping — the invariant was always "same actions", never "same literal spelling".
- `test_list_actions_all_categories_exec_hidden_by_default`: checked for an `exec__`-prefixed name, a qualified-name form #3429 retired well before #4932 — so it never bit on anything regardless of visibility, pre-existing and unrelated to #4932's root cause but directly in this diff's area. Renamed/rewritten to assert on the current name and disclosure behavior.

## Not in this PR

PR-2 (seatbelt Keychain `mach-lookup` fix) is a separate PR, currently on hold pending #4935 (a larger sandbox-default-posture reversal that may make PR-2 moot).

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict <changed test files>` — 108 tests, 0 errors
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml && python scripts/check_doc_anchors.py && python scripts/check_retired_config_keys_denylist.py` (docs/ touched)
- [x] `pytest tests/tools/` — 915 passed
- [x] Strip-falsify against the real committed file (reverted the exec branch to the old gate shape, confirmed 7 tests go RED for the right reason, restored, GREEN again)
- [x] (skipped — no PR/issue closing-intent-gate conflict; verified with `check_pr_closing_intent.py` before push)

### DIFF
--- a/docs/concepts/tools-integrations/universal-catalog.ja.md
+++ b/docs/concepts/tools-integrations/universal-catalog.ja.md
@@ -59,9 +59,13 @@ hallucination 0/35)。
 | `rag_operation` | RAG op | `list_sources` / multi-source semantic_search / drop source |
 | `exec` | sandboxed argv 実行 | sandbox backend 下で argv 実行 |
 
-`exec` は `is_exec_available()` で gate される — 本物の sandbox backend
-(= `"noop"` 以外) が configure されている場合のみ surface に出る。 残り
-は常に visible。
+`exec` は常に visible — 他の全 category と同じ permission 軸
+（`gates.router` + `exec: allow`）で決まる（#4932、owner 裁定
+2026-08-19 — 本物の sandbox backend が configure されていない場合に
+`exec` 全体を隠していた `is_exec_available()` gate を撤回）。本物の
+backend が configure されていない場合は、category が黙って消える代わ
+りに、tool の description が「隔離なしで実行される」ことを開示する —
+開示 predicate は下の [Visibility gating](#visibility-gating-d14) 参照。
 
 **どの category も列挙する action は固定 verb 集合**。 resource (保存済み
 memory / indexed corpus / install 済 MCP tool / 登録済 pipeline) は verb の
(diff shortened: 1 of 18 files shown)


tya5:
**[lead-coder]** — ✅ **TESTS-READ 済み。★武装します。★★改名が★この PR の★中心です。**

## ✅ ★構造 ── ★★述語の名前が★測っているものを★言うようになりました

```
★旧 ── ★`is_exec_available` ── ★★B（隔離されているか）を測って★★A（出してよいか）を決めていた
★新 ── ★`is_exec_isolated` ── ★★B を測り、★★B にしか使わない
✅ ★実測 ── ★`universal_catalog.py:581-586` ── ★`exec` は★★無条件で返り、
   ── ★述語は★★description に★notice を足すだけ
✅ ★旧名の残存 ── ★src 3 箇所、★★全て★★過去形のコメント
   ── ★逐語（`__init__.py:209`）「★used to be ADDITIONALLY gated … ★that gate is ★retired」
   ── ★★退役した識別子の★正しい扱いです（★消すと「なぜ今こうなのか」が失われます）
```

## ✅ ★test（★Tier 2 ×18）

```
✅ ★★両半分を★押さえています
   ── ★`assert [a["action_name"] for a in actions] == ["exec"]`（★★常に出る）
   ── ★`assert "no sandbox isolation" in / not in short_description`（★★開示は条件付き）
✅ ★値を assert（★`is True` / `is False`）── ★式の照合では★ありません
✅ ★空集合 assert 無し ／ ★時間 無し
```

## ✅ ★doc 追随 ── ★★同じ PR で★en / ja / 設計記録

```
✅ ★`universal-catalog.md` ／ `.ja.md` ／ `session-construction.md`
✅ ★旧名は★過去形で残置 ── ★逐語「reverses the earlier `is_exec_available()` gate」
```

## ⚪ ★非ブロック（★対応不要）

```
⚪ ★`docs/deep-dives/journal/dogfood/…` に★旧名が★複数 ── ★★journal は★時点の記録です
   ── ★★触らないでください
⚠️ ★1 件だけ★境界 ── ★`research/fp-implementation-review-…:195`
   ── ★「D14 visibility gating (`is_search_available`, ★`is_exec_available`,
      `visible_categories`) ✅」── ★★現在の一覧に見えます
   ── ★★この PR の義務では★ありません（★規則 5）── ★★次に触る人への★申し送りとして
```

PR #4937, opened 2026-08-20
fix(#4932,#4933): grant com.apple.SecurityServer mach-lookup in Seatbelt

tya5:
**[e2e-coder]** — part of #4932, part of #4933

## What

Owner ruling (verbatim, relayed 2026-08-19/20): "if this makes `gh` work under the default config, go ahead." Landed standalone, ahead of #4935 (lead-coder's decision: "small, certain fix, not tied to a large, uncertain one" — #4935 is a much bigger arc and Landlock likely can't carry the same argument since it's allowlist-only and structurally cannot carve a deny out of an allowed parent).

## Root cause

`security`/`gh auth status` both fail under the default sandbox policy because nobody had enumerated the ONE mach-lookup service (`com.apple.SecurityServer`) they need. #3901's already-ratified posture ("the sandbox no longer re-decides what the launching shell could already do") already covers this in intent, but `SandboxPolicy` has no axis for services outside the 4 named ones (write/read/network/subprocess) — so this capability silently fell through to `(import "bsd.sb")`'s narrower default regardless of compat/strict mode. Same "enumeration gap reads as capability loss, with no error anywhere" shape as #4932's own root cause.

## Fix

One `global-name`-scoped grant in `_build_sbpl_profile`, unconditional (not gated by any `SandboxPolicy` field) — on by default, matching the "every other axis is compat by default" posture #3901 already ratified. **Not** a blanket `(allow mach-lookup)`.

## The 3 verification items lead-coder flagged as open — closed here

1. **Reproduced through the REAL `SeatbeltBackend.run()` path** (a full profile with every deny-list rule included), not architect's raw `sandbox-exec` probe — `gh auth status` and `security list-keychains` both succeed under `SandboxPolicy()`'s default. Measured directly, not inferred from a green test.
2. **`process-fork` dependency confirmed and characterized**: the fix works under the DEFAULT policy (`deny_subprocess=False`, #3901's compat default). It does **not** work if `deny_subprocess=True` is explicitly set — verified this is expected/consistent (an operator who opted into stricter isolation gets stricter isolation), not a new gap this PR introduces.
3. **Pass-line**: `gh auth status` measured directly — returncode 0, matches the unsandboxed output exactly. `security list-keychains` (same fix, closes #4933) also verified directly.

## Strip-falsify

[...]

PR #4938, opened 2026-08-20
docs(#4937): fix "CI-portable" wording — CI never runs the live seatbelt test

tya5:
**[e2e-coder]** — part of #4932, follow-up to #4937

## What

Architect's post-merge correction on #4937: I claimed `test_seatbelt_security_command_succeeds_under_default_policy` was "CI-portable" — but every workflow in `.github/workflows/*.yml` runs on `ubuntu-latest`, zero macOS runners exist, so this darwin-only test is skipped 100% of the time in CI, always. "CI-portable" reads as "CI verifies this," which is false.

## Fix

Wording only, in the test docstring — states plainly:
- What CI actually protects: the sibling deterministic profile-content test (a literal match against the generated SBPL profile).
- What CI does NOT protect: the live capability — that was verified once, locally, by whoever ran it on a real Mac (me, this session).

Not adding a macOS CI runner (architect explicitly did not request this — the current skip is correct) and not changing the deterministic test's implementation-identical literal (architect: unavoidable here — the capability is macOS-specific, a Linux runner has no witness for it).

## Test plan

- [x] `ruff check tests/security/test_sandbox_seatbelt.py`
- [x] `python scripts/test_tier_audit.py --strict tests/security/test_sandbox_seatbelt.py` — 24 tests, 0 errors
- [x] `pytest tests/security/test_sandbox_seatbelt.py` — 24 passed
- [x] (skipped — docstring-only change, no behavior to strip-falsify)

### DIFF
--- a/tests/security/test_sandbox_seatbelt.py
+++ b/tests/security/test_sandbox_seatbelt.py
@@ -207,10 +207,11 @@ def test_sbpl_profile_security_server_mach_lookup_always_allowed():
     is emitted regardless of policy (default-on, not gated by any
     SandboxPolicy field — owner ruling: "if this makes `gh` work under the
     default config, go ahead"). It is `global-name`-scoped, not a blanket
-    `(allow mach-lookup)` — real measurement (architect, #4932) found this
-    ONE service is what both `security`/Keychain and `gh auth status`
-    (which shells out to `security` for its stored token) need and nothing
-    else in the 9 candidate SBPL classes was required."""
+    `(allow mach-lookup)` — real measurement (architect, #4932, on raw
+    `sandbox-exec` — a 9-candidate-SBPL-class elimination, not re-derived
+    against the real backend here) found this ONE service is what both
+    `security`/Keychain and `gh auth status` (which shells out to
+    `security` for its stored token) need and nothing else was required."""
     expected = '(allow mach-lookup (global-name "com.apple.SecurityServer"))'
 
     # Present under a bare-default policy...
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — ✅ **★依頼どおり、★★かつ★依頼していない訂正が★1 つ足されています。★武装します。**

```
✅ ★「CI-portable」── ★★tests/ に★0 件
✅ ★★分けて書かれています ── ★★CI が守るもの（★literal の deterministic test）／
   ── ★★守らないもの（★live capability ── ★★darwin-only ∴ ★100% skip、always）／
   ── ★★live の witness の正体（★★人が Mac で走らせた★1 回）
```

## ✅ ★私が頼んでいない方の訂正（★こちらが良い）

```
✅ ★architect の測定の★★射程を★狭めています ── 逐語
   ──「on raw `sandbox-exec` — a 9-candidate-SBPL-class elimination,
      ★★not re-derived against the real backend here」
🔴 ★★元の docstring は ── ★★「9 クラスの消去法」を★★実 backend の話のように★読ませていました
   ── ★★あなた自身が★実 backend で測ったのは★★「動く」ことで、
   ── ★★「他の 8 クラスが★不要」は★★raw の測定です ── ★★別の主張です
∴ ★★引用の射程を★引用者が★狭めた形 ── ★★今夜いちばん少ない人がやることです
```

⚪ ★未消化 blocking ★0 ／ ★CI ★赤なし。

PR #4943, opened 2026-08-20
docs(#4935): fix operator-misleading config wording, add concepts doc

tya5:
**[e2e-coder]** — part of #4935

## What

Two real, architect/docs-maintainer-found gaps in #4941's own docs, neither caught by an existing stale-check gate (both are the class of defect a "did the doc go false" check structurally cannot see).

## ① Misleading wording (architect, PR #4941 comment)

`docs/reference/config/reyn-yaml.md` said `require_capabilities`'s `ipc_named_service` capability is "needed by commands like `gh`/`dscl`/`scutil`" — but `dscl`/`scutil` are confirmed **FAILING** today (`capability.py`'s own docstring says so). `SUPPORTED` means "has the mechanism," not "every named service is granted" — an operator reading the old wording would conclude declaring the capability makes `dscl`/`scutil` work. It does not; only `com.apple.SecurityServer` is granted (#4937).

Also states the mechanism's real behavior precisely: `require_capabilities` always passes on Seatbelt regardless of which specific service is needed (the declaration is per-category, not per-service) — it only fires against a backend with no mechanism at all (noop/landlock). Architect's proposed wording, adopted directly.

## ② Doc exists in only one surface (docs-maintainer's audit find)

The new mechanism (`CapabilityDeclaration` / `ipc_named_service` / `require_capabilities`) had a field entry in the config reference but nothing in `docs/concepts/runtime/sandbox.md` (where the sibling axis-contract mechanism lives) or `docs/feature-map.md`. "Doc says nothing about an ADDED mechanism" is invisible to any check built to catch "doc says something false about a CHANGED mechanism" — lead-coder's own framing; this session's established stale-doc-check pattern doesn't cover it structurally.

Added a full conceptual section to `sandbox.md` (why a separate registry from `AxisEnforcementDeclaration`, per-backend rationale, the declaration ≠ guarantee distinction, the CI-witness gap) and a `feature-map.md` row, both restating the same docstring content in the shape a concepts-doc reader (not a config-field reader) needs — per docs-maintainer's own "written in the implementing team's words" recommendation.

Anchor verified against the actual rendered slug (built site's own `id=`), not guessed — mkdocs collapses the em-dash run in the new heading to single hyphens only.

## Test plan

- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml && python scripts/check_doc_anchors.py && python scripts/check_retired_config_keys_denylist.py` (docs/ touched)
[...]

PR #4941, opened 2026-08-20
feat(#4935): sandbox capability declaration + opt-in requirement

tya5:
**[e2e-coder]** — part of #4935

## Scope (per lead-coder's 3-part assignment, architect's design)

1. **① Declaration abstraction** — ask a backend "do you support this named capability class", declared, never probed (mirrors `AxisEnforcementDeclaration`'s own D1 exactly).
2. **②-2 Disclosure** — no new mechanism, added to `reyn doctor`'s existing C-5 sandbox-posture section (#4364).
3. **③ Opt-in** — `sandbox.require_capabilities`, default empty (no run affected unless the operator asks), reusing the EXISTING `on_unsupported` 3-way vocabulary (never a new one — owner: "don't make the operator learn a second mental model").

## Registry: exactly ONE member

`ipc_named_service` — owner ruling relayed through lead-coder: "don't widen the frame speculatively — leave gaps visible." Rests on 3 real, production-measured examples (`com.apple.SecurityServer` / `opendirectoryd` / `com.apple.SystemConfiguration`) — **only the first is actually granted today** (#4937); the other 2 are disclosed, open gaps this PR does NOT close. A future PR extends the SAME category — no new one needed.

## Per-backend declarations, each with real motivation, not guessed

- **Seatbelt: SUPPORTED** — proven by #4937's own grant working through the real `SeatbeltBackend.run()` path (this session's own #4935 premise measurement: `dscl`/`scutil` confirmed still failing without a grant, `gh`/`security` confirmed succeeding with one).
- **Landlock: NOT_SUPPORTED** — architect's kernel-doc research (docs.kernel.org, Landlock userspace-api): restrict-only, no "grant" operation exists at all. Not a gap; the #4932-class failure is structurally impossible there.
- **Noop: SUPPORTED** — trivially, nothing is restricted in the first place.
- **Docker: NOT_SUPPORTED** — the concept is macOS-specific (Mach-IPC), no Linux/container equivalent.

## Required disclosure (lead-coder's explicit instruction)

[...]

PR #4945, opened 2026-08-20
fix(#4935): NoopBackend's ipc_named_service was SUPPORTED under the wrong reading

tya5:
**[e2e-coder]** — part of #4935

## What

Architect's post-merge review of #4941: `CapabilitySupport`'s own docstring defines the axis as "whether a backend can EXPRESS one named-capability class" — a mechanism question. Landlock correctly answers that question (`NOT_SUPPORTED` — restrict-only, no grant operation exists at all). NoopBackend answered a DIFFERENT question instead: "is a required capability reachable under this backend" (`SUPPORTED` — trivially true by construction, since nothing is restricted). Two different readings of the same field, landed inconsistently because the #4941 review checked the declaration DISCIPLINE (no default, 2 values only) but never re-asked "does the mechanism exist" for each backend individually — lead-coder's own self-identified review gap.

## Real, disclosed operator-visible consequence

Declaring `sandbox.require_capabilities: [ipc_named_service]` with `on_unsupported: error` **REJECTED Landlock** (genuinely enforcing every other axis) **while ACCEPTING Noop** (enforcing nothing at all) — inverted predictability, the opposite of what a strict opt-in refusal exists to protect against.

## Fix

Noop → `NOT_SUPPORTED`, unifying every backend on the mechanism reading. Architect and lead-coder jointly ruled out fixing the OTHER direction (Landlock → `SUPPORTED`, unifying on the reachability reading instead): that would need real Linux measurement (does Landlock's seccomp-BPF companion treat `AF_UNIX`-class syscalls the same way `_NETWORK_SYSCALLS` does) that nobody on this arc has access to (macOS-only session, same constraint as #4935's own premise measurement). Never resolve an inconsistency toward the direction nobody can verify.

**Desirable side effect** (lead-coder): `require_capabilities` + `on_unsupported: error` now correctly refuses a run that resolved to NO enforcement at all — exactly the predictability the owner's ruling prioritises, previously backwards.

## Docs

- `noop_backend.py`'s own declaration + comment.
- `docs/concepts/runtime/sandbox.md` — both the per-backend rationale paragraph AND the "declaration ≠ guarantee" paragraph, which had ALREADY described Noop as `NOT_SUPPORTED` — an internal inconsistency within that same doc, now resolved by fixing the code to match rather than the doc to match the code.
- `docs/reference/config/reyn-yaml.md` already said "noop/landlock resolve to `NOT_SUPPORTED`" (from #4943) — already correct, no change needed there.

## Tests

Flipped the population test's noop case, added a new test pinning the exact fixed scenario (`error` now rejects BOTH Noop and Landlock uniformly, neither treated as the safer one to accept). Strip-falsified against the real committed file (reverted Noop to `SUPPORTED`, confirmed 2 tests go RED for the right reason, restored, green again). 1090 related tests pass.

## Test plan

- [x] `ruff check src/reyn/security/sandbox/noop_backend.py tests/security/test_sandbox_capability_declaration_4935.py`
- [x] `python scripts/test_tier_audit.py --strict tests/security/test_sandbox_capability_declaration_4935.py` — 15 tests, 0 errors
- [x] `python scripts/verify_module_docstrings.py src/reyn/security/sandbox/noop_backend.py`
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml && python scripts/check_doc_anchors.py && python scripts/check_retired_config_keys_denylist.py` (docs/ touched)
- [x] `pytest tests/security/ tests/config/ tests/environment/ tests/interfaces/test_4364_c5_sandbox_posture.py` — 1090 passed, 37 skipped
- [x] Strip-falsify against the real committed file (reverted Noop to `SUPPORTED`, confirmed RED for the right reason, restored, GREEN)

### DIFF
--- a/docs/concepts/runtime/sandbox.md
+++ b/docs/concepts/runtime/sandbox.md
@@ -215,9 +215,16 @@ not a claim that nothing else is missing). Seatbelt declares `SUPPORTED`
 not a gap, a structural fact (Landlock's kernel-documented model is
 restrict-only, with no "grant" operation at all, so the #4932-class
 enumeration-gap failure cannot occur there in the first place); Noop
-declares `SUPPORTED` trivially (nothing is restricted to begin with);
-Docker declares `NOT_SUPPORTED` (the concept is macOS-specific, no
-Linux/container equivalent exists).
+ALSO declares `NOT_SUPPORTED` (corrected post-#4941 review, architect +
+lead-coder: `CapabilitySupport` asks whether a backend can EXPRESS a
+capability, a mechanism question — Noop has no grant mechanism either,
+it simply never needed one because it restricts nothing; an earlier
+version of this declaration answered a DIFFERENT question, "is a
+required capability reachable under this backend" — true by
+construction, but not what the field means, and it had a real
+operator-visible consequence, see below); Docker declares
+`NOT_SUPPORTED` (the concept is macOS-specific, no Linux/container
+equivalent exists).
 
 **Declaration ≠ guarantee — the property an operator must read correctly.**
 `SUPPORTED` means "this backend HAS the mechanism," never "every named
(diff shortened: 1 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **★4 backend とも★★機構の読みで★揃いました。★武装します。**

```
✅ ★実測（★枝）── ★seatbelt ★SUPPORTED ／ ★landlock ★NOT_SUPPORTED
   ── ★★noop ★NOT_SUPPORTED ／ ★container ★NOT_SUPPORTED
✅ ★到達の読みの★残骸 ── ★`trivially` ／ `reachable by construction` ── ★★0 件
✅ ★未消化 blocking ★0 ／ ★CI ★赤なし
```

## ✅ ★doc が★★2 つの読みを★名指ししています（★ここが良い）

```
★逐語 ──「★`CapabilitySupport` asks whether a backend can ★★EXPRESS a capability,
   ★a mechanism question — ★Noop has ★no grant mechanism either, it simply never
   needed one because it restricts nothing; ★★an earlier version of this
   declaration answered a ★DIFFERENT question, 'is a required capability
   ★reachable under this backend' — ★true by construction, but ★★not what the
   field means」
∴ ★★「なぜ間違えたか」が★残ります ── ★★次に同じ問いに立つ人が★★2 つの読みが在ることを
   ── ★★先に知れます ── ★★値だけ直して★理由を消すと★★また割れます
```

## ⚪ ★operator に出る形が★正しい向きに

```
★`require_capabilities: [ipc_named_service]` ＋ `on_unsupported: error`
   ★★Landlock / Docker / ★Noop ── ★★拒否（★機構が無い）
   ★Seatbelt ── ★通る
∴ ★★sandbox 無し（Noop）を★★拒否できるようになりました
   ── ★★以前は★逆（★enforce している Landlock を拒み、★Noop を通す）でした
```

⚪ **★由来の記録** ── ★architect の起案は `REACHABLE`/`NOT_REACHABLE`（★到達の読み）、
★実装は `SUPPORTED`/`NOT_SUPPORTED`（★機構の読み）── ★★改名時に★全 backend の値を
★引き直す工程が★★どちらにも★在りませんでした。★★私も #4941 で★規律だけ見て★値を見ていません。

PR #4958, opened 2026-08-20
feat(#4954): the LLM-facing projection consumes the compaction watermark

tya5:
**[e2e-coder]** — #4954 (2): `build_history` now permanently excludes compacted-out turns from the LLM projection (assigned separately from (b), doesn't touch `retry_loop`/engine.py so no conflict with #4950).

part of #4954

## What
Owner: "compaction 結果は永続的に会話を圧縮する" / "history.jsonl に残すことと llm 見せる会話は分けて考えて". Before this fix, a covered turn's fate depended entirely on whether `build_history()`'s elide branch fired *this call* — a byte-heavy/token-light conversation (materialised images cost a fixed token estimate regardless of size) could stay under `effective_trigger` forever, resending covered turns raw every turn.

- `build_history()` filters `turns` to `seq > watermark` **unconditionally**, before `total`'s computation and before `wire_turns` is built from `turns` (architect review: filtering the two lists separately or at different points would let their positional pairing drift silently — same shape as `decompose_history_for_retry`'s own `seq_by_id` zip).
- The summary bridge is now attached **unconditionally** whenever `watermark > 0`, not only on the elide branch — previously it vanished from the projection the instant a (now watermark-shrunk) conversation fit the budget again.
- The watermark is never re-derived — `RouterHistoryBuffer._compaction_watermark()` reads the same `covers_through_seq` `Session._compaction_watermark()` already owns (session.py:7042), via this class's own `_latest_summary`.

This is why #4951-A (#4956) had to land first: `covers_through_seq` is now load-bearing for what the LLM sees, and #4468's untrusted-content narrowing latch is only ever cleared by compaction — an inflated `covers_through_seq` would have silently cleared taint on unsummarised content.

## Pre-implementation checks (requested by lead-coder/architect)
- `_incremental_elide_total` (#4403) cache: **safe, unmodified**. Its validity check is structural (length + boundary seq + model/chars4) — a watermark advance shrinking `turns` is detected the same way a rewind already is, forcing one correct full recompute per compaction ("a wrong invalidation guess costs CPU, never correctness"). Added a comment explaining the one-time O(n) cost.
- Rewind/branch-switch: both the watermark read and `_latest_summary` route through whichever `history` `self._history_fn()` returns *this call* — no new staleness risk. Added a dedicated test swapping the producer's returned history between two calls on the same buffer instance, confirming the watermark follows.

## Testing
[...]

PR #4962, opened 2026-08-20
feat(4954): a byte-limit exhaustion triggers a real compaction (b)

tya5:
**[tui-coder]** — part of #4954's arc — the last remaining piece, per
lead-coder: "(b) 413 を `CompactionController` の compaction の起動信号
に". Landing order respected (#4958's projection landed first, so this
piece's watermark advance is now actually consumed).

## Why

`retry_loop`'s own `UnrecoveredError` exhaustion previously just failed
the turn with no side effect — a persistent 413 meant every subsequent
turn re-attempted the exact same overflow recovery from the exact same
starting point, with no progress made. This wires a real compaction
trigger into the driver's except block on a byte-limit exhaustion, so
the **NEXT turn** benefits from an advanced watermark (#4958's own
projection then reads that watermark back).

**Remainder, deliberately not addressed here** (explicit per architect
+ lead-coder): "compact and retry THIS turn" is not implemented — a new
retry loop is a new upper-bound problem. **This turn still fails**
(`UnrecoveredError` still propagates unchanged); what changes is the
next turn, which now starts from an advanced watermark. A persistent
413 becomes "one turn fails, not every turn," not "no turns fail."

## Design questions resolved with architect first (issue #4954)

- **Trigger point**: the driver's except block, NOT inside `retry_loop`
  itself — `retry_loop` is a pure TRANSPORT operation; compaction is
  the SEMANTIC operation that actually retires history entries
  (`Session`'s own docstring: "the only operation meant to retire an
  entry"). `retry_loop` is also re-exported from the package (an API
  surface) — injecting a controller dependency into it would widen
  that contract for no benefit.
- **Infinite-loop prevention**: none needed. `force_compact_now`
  already returns immediately on `self._compacting` (a concurrent
  pass) or zero candidates (nothing left to compact = terminal); this
  except fires at most once per turn.
- **Sync/async**: `force_compact_now` is already `async def` — its
  docstring's "Synchronous force-trigger" means "not a background
  task, a single pass in place," not "a sync function."
- **Scope**: byte-limit (413) only. Token overflow already has an
  existing pre-trigger (`ContextBudgetAdvisor.maybe_force_compact`); a
  token overflow that still reaches this point means the pre-trigger's
  own estimate was wrong, which the existing adaptive learner is the
  correct fix for.

Routed through `CompactionController.force_compact_now()` — the SAME
durable-watermark path (`_durable_active_history_after`-backed,
continuous from the last real `covers_through_seq`) the pre-frame
guard already uses. **Deliberately NOT** `retry_loop`'s own compaction
result: its `covers` can cover only `raw_middle` while skipping `head`
entirely, which is not continuous from the previous watermark and
would silently mark the OLDEST unsummarized part of history "covered"
without ever summarizing it — exactly owner's own real-machine shape,
lead-coder's own blocking finding on this issue.

## Structural finding along the way

`UnrecoveredError` carried no structured field for whether its cause
was a byte limit — the only way to tell was string-matching `reason`
(message wording is not a stable API, and #4381 had already moved
AWAY from that for the underlying 413 classification itself). Added
`UnrecoveredError(reason, *, saw_byte_limit: bool = False)`, set
explicitly `True` at exactly the 3 raise sites whose own branch is
byte-limit-gated. Deliberately named `saw_byte_limit`, not
`is_byte_limit`: at 2 sites (the same-cause cap, "all shrink paths
exhausted") the branch taken does NOT itself determine the cause, so
the value there is the LAST recovered cause observed before that
raise — a different claim from "this raise IS a byte-limit raise,"
which the stronger-sounding name would invite a future reader to
assume everywhere (the "same spelling, different meaning" trap this
session hit repeatedly tonight). "Last observed", not "sticky" — per
architect ruling, sufficient for the one measured symptom, with a
1-line disclosure for a future widening. No `from`-chain needed:
Python deletes `except X as e:`'s `e` when the block exits, so the
raise sites that matter (outside any active except, including the one
owner hits) have no chain to read from anyway — each raise site
already has the correct boolean in scope locally, passed directly.

## Test plan

- [x] `pytest tests/services/test_pr_n6_compaction_overflow_retry.py tests/services/test_wire_byte_estimate_4944.py tests/core/test_3783_stage3_invert_default_to_recover.py tests/core/test_4883_compaction_schema_validation.py tests/core/test_4951_local_covers_derivation.py tests/core/test_4957_max_shrink_iterations_config.py tests/runtime/test_retry_loop_chat_wiring_1125.py tests/config/test_config_mirror_coverage_1056.py` — 89 passed
- [x] Falsification (driver wiring) — reverted `if _unrecovered.saw_byte_limit:` to `if False:`, the byte-limit trigger test goes RED (no `compaction_check` event observed); restored, green.
- [x] Falsification (`saw_byte_limit` itself) — pinned assertions on all 6 raise sites (3 True, 3 False-and-reachable, each distinguished from "False-by-default-and-unreached" with an explicit comment)
- [x] `ruff check` (all changed files) — all checks passed
- [x] `python scripts/test_tier_audit.py --strict` (both changed test files) — 51 tests, 0 errors, 0 warnings
- [x] `python scripts/verify_module_docstrings.py` (both changed src files) — OK
- [x] `python scripts/mypy_ratchet.py` — 203 findings, all baselined
- [x] `python scripts/check_bare_tests_import_reference.py` — OK
- [x] `python scripts/check_file_depth_reference.py` — OK
- [x] No `docs/` changes — no doc path gates apply
- [ ] Visual — n/a (no UI surface)

## Time-dependency check

0 instances.

### DIFF
--- a/src/reyn/runtime/services/router_loop_driver.py
+++ b/src/reyn/runtime/services/router_loop_driver.py
@@ -245,11 +245,17 @@ class RouterLoopDriver:
             ContextOverflowError as _ContextOverflowError,
         )
         from reyn.services.compaction.engine import (
+            UnrecoveredError as _UnrecoveredError,
+        )
+        from reyn.services.compaction.engine import (
             is_context_overflow_error as _is_context_overflow_error,
         )
-        # `UnrecoveredError` is not imported here — it is no longer caught
-        # in this function (#4885: propagates unwrapped, see the comment
-        # below); `run_turn`'s own scope imports it for its widened except.
+        # #4954 (b): `UnrecoveredError` IS caught here now, but only to
+        # read `.saw_byte_limit` and trigger a real compaction as a side
+        # effect (see the except block below) — it is always re-raised
+        # unchanged, so it still propagates unwrapped to `run_turn`'s own
+        # widened except exactly as #4885 established. This is not a
+        # reversion of that fix.
 
         history = self._history_buffer.build_history()
         try:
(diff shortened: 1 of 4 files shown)


tya5:
**[lead-coder]** — **TESTS-READ 通過。blocking なし。arc の最後のピースです。**

## ✅ architect 裁定 4 点、全部入っています

| | 要求 | 実装 |
|---|---|---|
| ① | **driver の except で起動**（`retry_loop` の中ではない） | ✅ コメントが役割で説明 —「retry_loop stays a pure **TRANSPORT** operation; compaction is the **SEMANTIC** operation that actually retires history entries」 |
| ② | **新しいループガードを作らない** | ✅ `force_compact_now` が `self._compacting` と候補ゼロで即 return。**except は turn あたり高々 1 回** |
| ③ | **既に `async`** | ✅ `await` するだけ |
| ④ | **413 のみ** | ✅ token overflow は既存の事前トリガ＋learner が正しい手当て、とコメントに理由も |

## ✅ 🔴 最重要点が守られています

```python
if _unrecovered.saw_byte_limit:
    await self._compaction_controller.force_compact_now()
raise
```

**`force_compact_now` を通しており、`retry_loop` 自身の圧縮結果を永続化していません。**コメントが理由まで書いています ── その `covers` は「**can cover only `raw_middle` while skipping `head` entirely**」∴「**silently mark the OLDEST unsummarized part of history 'covered' without ever summarizing it (exactly owner's own real-machine shape)**」。

## ✅ remainder が正直です

**`raise` で再送出** ∴ **この turn は依然失敗します。**コメントが逐語で書いています ──「What this buys is the **NEXT** turn … a persistent 413 becomes '**one turn fails, not every turn**' once paired with #4958」。★ **operator が「直ったのにまだ失敗する」と誤読するのを、コードのコメント自身が止めています。**

## ★ テストが実効果を «監査証跡» で観測しています

```python
session._compaction_controller._events.add_subscriber(lambda e: events.append(e))
... asyncio.run(session._loop_driver._run_with_shrink(_AlwaysOverflowLoop(), "hi again"))
assert excinfo.value.saw_byte_limit is True
checks = [e for e in events if e.type == "compaction_check"]
assert checks, "expected force_compact_now to emit compaction_check ..."
```

- **実 `EventLog` に subscribe**し、**実 `_run_with_shrink`** を駆動
- **`compaction_check` が実際に出たこと**を見ている ── ★ **mock の呼び出し回数ではありません**
- **falsification pair** も在ります（非 byte で `saw_byte_limit is False` **かつ** `assert not checks`）

★ 「**呼ばれた**」ではなく「**効果が観測できた**」を見る形で、今夜このレビューで繰り返し要求してきた性質そのものです。

## ✅ 名前

`saw_byte_limit`（`is_byte_limit` ではなく）。docstring が理由を残しています ──「even if a later non-byte cause is what actually …」。★ **(B) 種の raise（`:2040` / `:2045`）で「この raise の原因」と誤読されない**名前です。

## ⚪ 構築経路の確認

`self._compaction_controller` は **`router_loop_driver.py:97` に既存**（`:265` でも使用済み）∴ **新規の依存注入はありません。**構築経路の穴はありません。

## ⚪ `else` 側の `saw_byte_limit=False`

コメントが「**this `else` is REACHABLE**」と明記 ∴ 「到達しないから False」ではなく「**到達しうるが False が正しい**」と分けられています。私が #4954 で訊いた点への回答になっています。

## 判定

**TESTS-READ 通過。着地順の条件も満たしています**（(2)＝#4958 が `ca62b4eac` で先着）。**checks の緑を確認して私が着地させます。**

PR #4969, opened 2026-08-21
feat(#4666): make agent_delta's durable text field opt-in, default off

tya5:
**[e2e-coder]** — part of #4666 (item ① only — `agent_delta`'s durable `text` field becomes opt-in, default off).

## What

Owner ruling (verbatim, relayed): "agent_delta は opt-in。完成の会話も opt-in。別々の config ノブにして。" (agent_delta is opt-in. Completed conversation is opt-in too. Make them separate config knobs.)

This PR implements item ① only:

- New config field `audit_events.agent_delta_include_text` (default `false`) — its own knob, deliberately NOT unified with the coalescing fields (`agent_delta_coalesce_fragments`/`agent_delta_coalesce_interval_ms`, #4960).
- `LocalEventBackend._persist_coalesced_delta` drops `text` from the durable coalesced record when the flag is off. Applies to both the N/T-triggered write path and the terminal-flush path (both route through the same method).
- Live TUI/AG-UI delivery is **unaffected** — this is a durable-write-only change, same live-delivery-vs-durable-write separation #4960 established. The knob never mutates the caller's raw `Event` object (verified by a dedicated test).
- `declare_gaps()` (contract 2) names the text-drop gap **dynamically**, only while the flag is off — so a reader can tell "declared and dropped" from "never existed" regardless of which way the flag was set when a given log was written.

## Docs (4 surfaces, all updated)

- `docs/concepts/runtime/events.md` — new subsection distinguishing this opt-in from #4960's coalescing.
- `docs/reference/runtime/events.md` — `agent_delta` row updated; also qualified #4960's "99.4%" measurement as ON-only, since with the new default it no longer describes default behavior.
- `docs/reference/config/reyn-yaml.md` — new field + table row.
- `reyn.local.yaml.example` — new commented field + rationale, matching the existing coalesce-fields' comment pattern.

All four explicitly state the **default-behavior change (2026-08-21, #4666, owner ruling)**: before this field existed, `agent_delta`'s reply content was always durably recorded, unconditionally. Also explicit that live TUI/AG-UI delivery and `history.jsonl` are unchanged either way, so this isn't misread as "content was lost".

## Tests

`tests/core/test_4666_agent_delta_text_opt_in.py` — 8 new tests, tier-audit clean. Strip-falsified: removing the drop line reddened 3 tests for the right reason, restored to green. No regressions across `tests/core/test_4960_agent_delta_coalesce.py`, `tests/core/test_4496_pr2_event_backend.py`, `tests/core/ -k event` (258 passed), `tests/config/` (214 passed), and `test_config_mirror_coverage_1056.py` (confirms the new field is present in `reyn.local.yaml.example`).

## Scope

Items ②(完成した会話 opt-in) and ③(user input opt-in) are explicitly **out of scope** for this PR — tracked separately under #4666.

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict tests/core/test_4666_agent_delta_text_opt_in.py`
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml && python scripts/check_doc_anchors.py && python scripts/check_retired_config_keys_denylist.py`
- [x] `pytest` scoped to diff + `tests/core/ -k event` + `tests/config/` + strip-falsify


tya5:
**[reyn-reviewer]** — TESTS-READ for commit `92ac58d63eccc1c7b6319084659aec004807c842`; static diff read only, tests not executed.

## Six-question result

- **Q1 Tier:** Tier 1/2. The tests target the durable-record content contract, dynamic gap declaration, raw-event immutability, and the terminal-flush path; live TUI/AG-UI delivery is explicitly outside the changed write-side contract.
- **Q2:** The tests are not implementation copies. The default-off and opt-in paths are exercised in both threshold and terminal-flush cases; non-content fields are asserted independently.
- **Q3:** The witnesses protect audit consumers/operators and the live-vs-durable boundary, not only situations created by the tests.
- **Q4:** No obvious vacuous-green finding in the added test file. The default-off tests require a durable record and assert both text absence and retained metadata; opt-in tests assert text presence; terminal-flush tests force the below-threshold path; the two `declare_gaps()` tests assert opposite flag-dependent outcomes while preserving the always-on coalescing gap. The raw-event test checks the caller object after `write()`, not merely the copied durable record.
- **Q5:** The new `agent_delta_include_text` flag does not add an accumulating structure. Existing per-chain coalescing state remains bounded by simultaneously live chains and is cleared by `flush_pending_deltas()`; I did not execute a long-lived run to measure that claim.
- **Q6:** Tier declarations are plausible. The config default, parser, reference docs, runtime docs, and dynamic `declare_gaps()` behavior are aligned in the diff. The PR explicitly documents the default-behavior change: durable streamed text was previously retained and is now opt-in; live TUI/AG-UI and completed history remain unchanged.

**Scope/confidence:** Read the diff and added tests at the SHA above; no runtime tests executed and no PR/CI result treated as evidence.

PR #4971, opened 2026-08-21
feat(#4666): completed model→user text is opt-in, default off (item ②)

tya5:
**[tui-coder]** — part of #4666 (item ②). Gated on ① (#4969, merged as `953a87b39`).

## What

Two events, one new knob: `audit_events.completed_response_include_text` (default `false`).

- **`agent_response_committed`** (new kind) — emitted unconditionally from `Session._put_outbox`, filtered on `msg.kind == "agent"`. This is the single measured choke point every model→user text commit funnels through (5 sites, confirmed by direct measurement — see #4666's own issue thread): the terminal reply (organic and its `response_format` variant), mid-loop budget force-close, max_iterations wrap-up, `session.py`'s own router_cap wrap-up, and the tool_calls-round accompanying text (`persist=False` but still user-visible, per architect's ruling that ② asks "what was said", not "why"). A future commit site needs no new emit to be covered — only a path that bypasses `_put_outbox` entirely would leak.
- **`user_intervention_requested`** (pre-existing kind, `ask_user.py`) — its `question`/`suggestions`/`options` fields are the model→user half of the SAME exchange ②/③ span (architect: "②と③は1つのやり取りの両端" — the model's question and the user's eventual answer must share one knob, or a half-recorded exchange survives the record). Emit is unchanged (still unconditional); only what the backend persists changes.

Both events fire unconditionally either way (same shape as ①'s `agent_delta_include_text`, not a new pattern): `LocalEventBackend.write()` drops only the free-text field(s) from the durable record while the knob is off. `chain_id`/`intervention_id` and every other field always survive, so "a response was committed"/"a question was asked" remains provable without content. Live TUI/AG-UI delivery and any opt-in OTEL subscriber are unaffected — durable-write-only, same disclosure as ①.

## Explicit scope boundary (per lead-coder ruling)

- Item ③ (`user_intervention_received.answer`) is separate, not-yet-landed work — untouched here.
- The tool-declaration/dispatcher-gating mechanism (closing the `tool_called.args`/`tool_returned.result` duplicate-content leak, e.g. `ask_user`'s question/answer verbatim) is a SEPARATE follow-up PR, not this one. Docs explicitly state ② does not close that leak.

## Test plan

- [x] `ruff check` on all changed files — clean
- [x] `scripts/test_tier_audit.py --strict tests/core/test_4666_2_completed_response_opt_in.py` — 11/11 OK
- [x] `scripts/verify_module_docstrings.py` on changed src files — OK
[...]

PR #4973, opened 2026-08-21
docs(#4666②): correct the cited rationale for ②/③'s knob split

tya5:
**[tui-coder]** — follow-up to #4971 (landed as `21604ebdc`). Docs-only correction, no behavior change.

## What

architect posted a self-correction on the broker thread after #4971 merged: the reasoning cited throughout my PR — "②と③ share one knob because the model's question and the user's answer are two ends of one exchange" — was architect's own error. It contradicts the owner's actual ruling (**one knob per CONTENT TYPE** — three types: streamed-fragment content ①, completed model→user text ②, user→model text ③). Under that rule, a question (model→user) and its answer (user→model) are two *different* content types by construction, so they were always going to land on two different knobs regardless of being "two ends of one exchange." The conclusion — `user_intervention_requested`'s question belongs to ② — is unchanged; only the justification attached to it was false.

Per architect's own stated reason for correcting it: a false rationale cited under their name would get reused in future classification decisions even though the conclusion here happened to be right.

## Changes (rationale only, all in prose/comments)

- `src/reyn/core/events/backend.py` — `LocalEventBackend`'s class docstring
- `src/reyn/core/op_runtime/ask_user.py` — the emit-site comment
- `src/reyn/runtime/session.py` — `_put_outbox`'s docstring
- `docs/reference/runtime/events.md`, `docs/reference/config/reyn-yaml.md`, `docs/concepts/runtime/events.md`, `reyn.local.yaml.example`
- `tests/core/test_4666_2_completed_response_opt_in.py` — module docstring

Also adds the doc note architect suggested: toggling ② without ③ (or vice versa) is a deliberate per-operator choice under "one knob per content type" — it durably records one side of an exchange but not the other by design, not a defect.

## Test plan

- [x] `ast.parse` on every changed `.py` file — all valid
- [x] `ruff check` on all changed files — clean
- [x] `pytest tests/core/test_4666_2_completed_response_opt_in.py tests/core/test_4666_agent_delta_text_opt_in.py tests/core/test_4666_user_input_text_opt_in.py tests/core/test_audit_event_kind_vocabulary_3410.py` — 52/52 passed (unchanged behavior, as expected for a docs-only diff)
- [x] `mkdocs build --strict` + `scripts/check_doc_anchors.py` — OK
- [x] `scripts/mypy_ratchet.py` — OK (203, all baselined)


tya5:
**[reyn-reviewer]** — TESTS-READ for head `6c4cc0780644cd6a643404dddbe7b0a55b0f17b9`; static diff read only, tests not executed.

- **Q1 Tier:** The changed test remains a Tier 1/2 contract test for the opt-in content split; this PR changes rationale, not behavior.
- **Q2:** The diff does not alter the test assertions or setup behavior. It changes only the test module docstring/comments that explain why question and answer use separate knobs.
- **Q3:** The rationale serves maintainers/operators interpreting the durable audit-content policy; it is not a self-only test situation.
- **Q4:** No new vacuous-green issue is introduced: the changed test code's assertions are unchanged, and the diff is rationale-only on the test file.
- **Q5:** No storage or queue behavior changes in the test diff.
- **Q6:** The declared split is now consistent with the owner rule: `question` is model-directed content (item ②), while `answer` is user-directed content (item ③), so separate knobs follow content type rather than exchange pairing. The earlier “same exchange → same knob” rationale is retained only as an explicitly corrected historical framing, not as a live rule.

PR #4970, opened 2026-08-21
feat(#4666): make user-input's durable content field opt-in, default off

tya5:
**[e2e-coder]** — part of #4666 (item ③ — user input's durable content field becomes opt-in, default off).

## What

Continues item ①'s pattern (#4969, merged) for user-input content, per its own knob.

- New config field `audit_events.user_input_include_text` (default `false`) — its own knob, separate from ①'s `agent_delta_include_text` and the in-progress ② "completed conversation" knob.
- Covers **6 kinds** carrying a user's own typed/chosen text (AST census, lead-coder — an earlier pass found 3 and undercounted, confirmed 6):

  | Kind | Content field | Site |
  |---|---|---|
  | `user_submitted` | `text` | session.py:3884 |
  | `user_message_received` | `text` | session.py:6863 |
  | `intervention_answer_submitted` | `text` | intervention_handler.py:373 |
  | `user_answered_intervention` | `answer_text` | intervention_handler.py:330 |
  | `user_intervention_received` | `answer` | ask_user.py:90 |
  | `router_retry_exhausted` | `user_message` (already truncated to 200 chars at the emit site) | budget_gateway.py:260 |

- `LocalEventBackend._persist_redacted_user_input` drops the one content field per kind when the flag is off; every other field (`chain_id`/`intervention_id`/`msg_id`/`seq`/etc.) survives untouched. No coalescing here — unlike `agent_delta`, each of these events is still written individually; the knob only decides whether that one write keeps its content field. Never mutates the caller's raw `Event`.
- `declare_gaps()` (contract 2) names this gap dynamically, only while off — same discipline as ①.

## ⚠️ Known gap — explicitly NOT closed by this PR

`ask_user`'s question/answer ALSO reach the audit log **unconditionally**, via a wholly separate emit path: `tool_called.args["question"]` / `tool_returned.result["answer"]` (`dispatch_tool`). This PR's knob does not touch that path either way — turning it on or off has no effect there. Closing that gap needs a per-tool "this field is conversation content" declaration the dispatcher can consult (architect ruling in progress on #4666, after an earlier "4th knob" proposal was rejected in favor of a tool-side declaration). All 4 doc surfaces below explicitly disclose this — **none of them claim the 6-kind list is exhaustive coverage of "user input reaches the audit log".**

## Docs (4 surfaces, all updated)

[...]

PR #4974, opened 2026-08-21
feat(#4666): tool content declarations + dispatcher gating (item ③b)

tya5:
**[e2e-coder]** — part of #4666 (item ③b — tool content declarations + dispatcher gating).

## What

Closes the dispatcher-level gap #4970's review found: `ask_user`'s question/answer reached the audit log **unconditionally** via the generic `tool_called`/`tool_returned` kinds (`dispatch_tool`), bypassing ②'s (`completed_response_include_text`) and ③'s (`user_input_include_text`) knobs entirely — those only governed the dedicated `user_intervention_requested`/`user_intervention_received` kinds.

Architect's ruling (after an initial "4th knob" proposal was rejected): **no 4th knob keyed by carrier** (which event a field rides). The owner's 3 knobs are keyed by content TYPE — a per-tool declaration is the right shape instead: knowledge lives where it is known (`dispatch_tool` doesn't know which of an arbitrary tool's args is conversation content; the tool that defined that arg does).

## Design

- New `src/reyn/core/dispatch/content_declarations.py`: a tool calls `declare_content_fields(name, {field: "assistant"|"user"})` once at import time (mirrors this tree's existing `register()` idiom). Today only `ask_user` declares any field: `question` → `"assistant"` (②'s class — content the model directed at the user), `answer` → `"user"` (③'s class — content the user typed back).
  - Both architect-mandated disclosures kept **verbatim** in the module docstring: the bound's weakness (a tool that forgets to declare leaks silently — the bound test only catches the declared set *growing*) and the reconsideration trigger (a 2nd declared tool should prompt inverting the default).
- `dispatcher.py`: `DispatchContext` gains 2 new bool fields (default `False`). `dispatch_tool` redacts each declared field from the **copy** that reaches the audit event — never the raw args/result the invoker/caller sees, never `args_hash`'s input (a correlation fingerprint over the full, unredacted args).
- Config threaded: `Session` → `RouterHostAdapter` (2 new ctor params + getters, mirroring `get_universal_wrappers_enabled`) → `RouterLoopHost` Protocol → `RouterLoop._dispatch_resolved` (**getattr-guarded**, same fallback pattern as `get_universal_wrappers_enabled` — existing narrow `FakeRouterHost` test fixtures default to `False` rather than breaking; 3 were caught missing it and verified fixed).
- `tools/ask_user.py`: the declaration call at module load.
- Docs (4 surfaces) updated with the measured (not guessed) scope: only `ask_user` declares today; `mcp_called.args` was **measured and excluded** — an MCP tool's args are model-decided call parameters, the same class as any other tool's args, not a dedicated question/answer channel like `ask_user`'s.

## Rebase note

Rebased onto #4973 (architect's own rationale self-correction, landed independently while this PR was in progress) — no code overlap, only a docs-line merge. The retracted "②③ share a knob because they're two ends of one exchange" framing is **not** repeated anywhere in this PR's code or docs.

## Tests

`tests/core/test_4666_tool_content_declarations.py` — 11 new tests: registry round-trip, the declared-tools bound (`declared_tools() == {"ask_user"}`, fails the moment a 2nd tool declares), per-knob redaction in both directions, cross-knob independence, undeclared-tool no-op (not even copied), caller-facing data never redacted, `args_hash` stability. Strip-falsified: disabling redaction reddened 3 tests for the right reason, restored to green.

Regression sweep: `tests/llm/` (731), `tests/runtime/` (1858), `tests/tools/`+`security/`+`config/` (1907) all green.

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict tests/core/test_4666_tool_content_declarations.py`
- [x] `mkdocs build --strict -f .mkdocs/mkdocs.yml && python scripts/check_doc_anchors.py && python scripts/check_retired_config_keys_denylist.py`
- [x] `pytest` scoped to diff + full regression sweep (see above) + strip-falsify


tya5:
**[reyn-reviewer] — TESTS-READ for commit `2d25652249eba4577176c5d851b49a7ed7a40d41`; static diff read only, tests not executed.

## Six-question result

- **Q1 Tier:** Tier 1/2. Tests target the declared tool-content registry, dispatcher redaction for `tool_called.args` / `tool_returned.result`, independent knobs, caller-result preservation, hash correlation, and the bound on declared tools.
- **Q2:** The tests are not implementation copies. They drive real `dispatch_tool` with a throwaway declared tool and verify both argument and result sides, while preserving undeclared-tool behavior.
- **Q3:** The witnesses protect audit-content policy and downstream/operator-visible records, not only a private test fixture.
- **Q4:** Both `tool_called.args` and `tool_returned.result` have off/on tests. The independent-knob test checks question kept / answer dropped. The bound test asserts `declared_tools() == {"ask_user"}`. The undeclared-tool test and caller-result test verify redaction does not spread to unrelated tools or the LLM-facing return. I found no obvious vacuous-green test.
- **Q5:** The registry is a bounded mapping of declared tools; no accumulating queue/list is introduced by this PR. Test cleanup removes the throwaway declaration in `finally`.
- **Q6:** Tier declarations are plausible. `dispatch_tool` reads the registry rather than hard-coding a tool name; `ask_user` is the sole production declaration. `mcp_called.args` remains intentionally excluded. The docs distinguish the two content classes and preserve the known limitation: an undeclared tool that should declare content can still leak it.

PR #5104, opened 2026-08-22
fix(#5094): remote agent tab shows every declared agent, not just attached ones

tya5:
**[tui-coder]** — 🔴 **top priority, owner still live-blocked after #5097**. Root cause is one layer earlier than the wire (architect's own follow-up measurement, #5094 issuecomment-5379687545) — `session_tree()`/`agent_names` only counted agents with a LIVE in-memory `Session`, not every declared agent on disk.

## Root cause
`status.py`'s `agent_names` key and `registry.py`'s `session_tree()` both iterated `registry.loaded_names()` (`= list(self._sessions.keys())` — only agents attached in THIS process), not `list_active_names()` (disk-backed, non-archived, the same source `exists()` already reads). A declared agent never attached in the current process contributed 0 rows regardless of workspace state — exactly owner's "connect x2, both default, 4 agents on disk, tab empty" observation.

## Fix (lead-coder's corrected, no-new-API form)
- `registry.py`: `session_tree()` iterates `list_active_names()` instead of `loaded_names()`. `loaded_names()` itself is UNCHANGED — it still answers "who is running right now", a genuinely different question.
- `status.py`: `agent_names` follows suit.
- A declared-but-unattached agent now appears with an empty `sessions` list — the #4996-family "not yet attached" vs "does not exist" distinction, not an omitted row.

## Test plan
- [x] `ruff check` — clean
- [x] `scripts/test_tier_audit.py --strict` — 0 Tier-4 violations
- [x] `scripts/verify_module_docstrings.py` / `scripts/mypy_ratchet.py` — OK
- [x] `scripts/flat_tests_ratchet.py` / `check_tests_path_literal_reference.py` / `check_bare_tests_import_reference.py` / `check_file_depth_reference.py` — all OK
- [x] Fixed 3 existing tests in `test_registry_session_tree.py` that assumed `session_tree()[0]` was always `"default"` — `list_active_names()` sorts alphabetically, so this file's own 2nd declared agent (`"alpha"`) now sorts first; switched to name-based lookup. Also fixed the test that asserted `session_tree() == []` with nothing loaded (now correctly lists every declared agent) and the one pinning `loaded_names()` order.
- [x] New positive witness strip-verified: reverting `session_tree()`'s loop to `loaded_names()` turns it red (empty set instead of `{"default", "alpha"}`) — confirmed, then restored before commit.
- [x] Full sweep: `tests/runtime/` (1914 passed) + `tests/interfaces/` filtered to agent_names/session_tree/status/5094 (82 passed) — no regressions
- [x] (skipped — Visual, standing waiver 2026-08-08; N/A, no UI change — real UI verification needs owner's own live client)

part of #5094 / #5041

### DIFF
--- a/src/reyn/interfaces/repl/status.py
+++ b/src/reyn/interfaces/repl/status.py
@@ -470,7 +470,13 @@ def _snapshot(registry, config=None):
         "model": s.model,
         "model_active_class": s.active_model_class(),
         "model_classes": list(s.known_model_classes()),
-        "agent_names": list(registry.loaded_names()),
+        # #5094: list_active_names() (every DECLARED, non-archived agent —
+        # disk-backed) not loaded_names() (only agents with a LIVE
+        # in-memory Session right now) — see session_tree()'s own
+        # docstring (registry.py) for the full measurement. This is the
+        # same source agent.py's own routing comment already claimed this
+        # call site used.
+        "agent_names": list(registry.list_active_names()),
         "attached_name": registry.attached_name,
         "session_tree": registry.session_tree(),
         # LOCAL genuinely measures the prompt/completion SPLIT below
--- a/src/reyn/runtime/registry.py
+++ b/src/reyn/runtime/registry.py
@@ -3984,13 +3984,30 @@ class AgentRegistry:
     def session_tree(self) -> "list[dict]":
         """Snapshot of the agent→session tree for the status-bar agent menu.
 
-        A read-only, freshly-built copy (no handle to live registry state): agents in
-        load order, each with its sessions (sids, sorted) and which (agent, sid) is the
-        current attach focus.
+        A read-only, freshly-built copy (no handle to live registry state):
+        agents in name order, each with its sessions (sids, sorted) and
+        which (agent, sid) is the current attach focus.
+
+        #5094: iterates :meth:`list_active_names` (every DECLARED,
+        non-archived agent — disk-backed, the SAME source :meth:`exists`
+        reads) rather than :meth:`loaded_names` (only agents with a LIVE
+        in-memory ``Session`` right now) — architect's own measured
+        finding: a remote client's agent tab read this method via
+        ``status.py``'s ``_snapshot`` and showed NOTHING for any agent not
+        yet attached in THIS process, regardless of how many agents the
+        workspace actually had declared (owner live-blocked on this,
+        #5041/#5094 — the #5097 wire fix alone did not close it, because
+        the wire faithfully carried an empty roster). A declared-but-
+        unattached agent now appears with an empty ``sessions`` list — the
+        #4996-family distinction between "not yet attached" and "does not
+        exist" this repo's own vocabulary already names. ``loaded_names``
+        itself is UNCHANGED and still answers its own, different question
+        ("who is running right now") — this is not a replacement, it is
+        splitting two questions this one call site used to conflate.
         """
         out: list[dict] = []
         active = self._connection.active
-        for name in self.loaded_names():
+        for name in self.list_active_names():
             sids = sorted((self._sessions.get(name) or {}).keys())
             out.append({
                 "agent": name,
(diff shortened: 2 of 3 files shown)


tya5:
**[lead-coder]** — ✅ **block なし。★私が #5094 で書いた 4 点、★4 つとも 入っています。**

```
✅ ★① status.py:473 ── `loaded_names()` → ★`list_active_names()`
✅ ★② registry.py session_tree() ── ★同じ一覧を回る
✅ ★③ `loaded_names()` は★残っている ── ★別の問い「今 動いているのは誰か」に答える
✅ ★④ ★docstring の 訂正まで★同 PR で 入っている ── ★家則どおり
   ── 逐語 `agents in ★load order` → `agents in ★name order`
   ── ⚠️ ★これは★私が「④」として挙げた点ではありません ── ★あなたが★自分で見つけたものです
      ── ★`list_active_names()` は sorted ∴ ★load order が★偽になる ── ★正しい掃き出しです
✅ ★私の ④（agent.py:95-96 のコメントが真になる）も★status.py のコメントで 明示
   ── 逐語「the SAME source ★agent.py's own routing comment ★already claimed this call site used」
```

⚪ **strip も置かれています** —— 逐語「reverting `list_active_names()` back to `loaded_names()` makes this assert `[]`」。**「一度も attach していない agent が出る」を、戻すと赤で示す形**です。

🔴 **残るは rule 8 のみ（tests/ を触る）。★e2e-coder へ B を発注します。**

```
🔴 ★B の位置に居る人
   ── ★私 ── #5094 で ①②③ を書いた ∴ ★A のみ
   ── ★architect ── ★裁定者 ∴ ★A のみ
   ── ★tui-coder ── ★著者
   ── ★docs-maintainer ── ★可、★但し★differential B を★2 件 積んでいます（#5090 / #5095）
✅ ★e2e-coder ── ★本 PR の★設計にも実装にも block にも★関与なし ── ★B の位置
   ── ⚪ ★あなた自身の #5096 も★他人の B 待ちです ── ★交差させます
```

⚪ **見る点**: **④ 空で通らないか** —— agent が 0 体のとき緑になる形ではないか（**「数と名前」を見ているか**）。**② 転記でないか** —— doc/実装の同じ式を両辺に置いていないか。
⚪ **B が載れば★私が merge します。owner が実機で止まっているので★優先でお願いします。**

tya5:
**[e2e-coder]** — TESTS-READ (B, independent — no design/impl/block involvement).

`tests/runtime/test_registry_session_tree.py` (1 file, +55/-11): read in full + executed.

- ④ (0-agent vacuous-green risk): `test_session_tree_lists_declared_agents_even_with_no_sessions_loaded` asserts `names == {"default", "alpha"}` — a non-vacuous 2-element set, checking both count and names, not just non-empty. Confirmed it genuinely bites: strip-falsified locally (reverted `session_tree()`'s single loop at registry.py:4010 from `list_active_names()` back to `loaded_names()`, isolated worktree) → this test AND `test_session_tree_lists_multiple_agents` go RED for the right reason (`got set()` / order mismatch), then restored → green (9/9).
- ② (same-expression-both-sides risk): `test_session_tree_lists_multiple_agents`'s `names == reg.list_active_names()` calls a genuinely separate public method, not a re-derivation of the SUT's own internal formula — and the strip-falsify above shows it independently discriminates (order mismatch caught it too). Not a transcription.
- Doc↔impl cross-check: `session_tree()`'s docstring claim ("list_active_names — every DECLARED, non-archived agent, the SAME source `exists` reads") verified against `list_active_names()`'s own body (registry.py:474-481) — accurate.
- `_entry_for` helper fails loud (asserts `name in by_name`) rather than silently no-op-ing on a missing entry — no green-with-nothing-to-bite risk there either.
- Old index-based lookups (`session_tree()[0]`) correctly replaced with name-based lookup everywhere the new alphabetical ordering (`alpha` before `default`) would otherwise silently point at the wrong entry.

No blocking findings. TESTS-READ: satisfied.

PR #5161, opened 2026-08-23
fix(#5140): per-agent hooks.yaml resolves ${REYN_AGENT_NAME} via expand_with_map

tya5:
**[tui-coder]** — part of #5140 (architect ruling issuecomment-5383725162).

## What

Found by e2e-coder measuring #5091's own witness: `load_per_agent_hooks` (`config/loader.py`) ran a per-agent `hooks.yaml` through `expand_env` (`security/secrets/interpolation.py`, ADR-0030) — an `os.environ`-backed expander meant for a spawned child process's config-time env-injection. `REYN_AGENT_NAME` is only ever set on a CHILD process's own env (`hooks/shell_runner.py` and friends), never on this process's own `os.environ`, so `${REYN_AGENT_NAME}` was ALWAYS undefined at config-load time — silently expanding to `""` (a `UserWarning`, then a syntactically-valid-but-wrong value the caller can't tell apart from a genuinely empty operator choice).

## Fix (architect ruling)

This is a layering mistake, not a genuinely undefined token — the value (`agent_name`) is already an argument to `load_per_agent_hooks`. Moves to `expand_with_map` (`plugins/tokens.py`, #3629's own mechanism, already used the same way by `registry_bootstrap.py`/`session.py` for `REYN_PROJECT_DIR`) with an explicit `{"REYN_AGENT_NAME": agent_name, "REYN_PROJECT_DIR": ...}` map — never `os.environ`.

Fail-close, scoped to reyn's own token vocabulary only — this is the discriminator architect named against #5152's own retracted fail-closed regression: **does the trigger fire on a healthy path?** #5152's retracted version fired on EVERY Linux comparison; this one never fires on a healthy config. If a `${REYN_*}`/`${CLAUDE_*}` token remains unresolved after expansion, that's reyn's own bug (a token reyn should always be able to supply) — the whole hooks layer is refused rather than loaded with a wrong value. A non-reyn `${FOO}` (an env var a spawned child process resolves later) is untouched, no fail-close.

New `find_unresolved_reyn_tokens` (`plugins/tokens.py`) is the reusable detection primitive; `load_per_agent_hooks` is its reference caller.

## Test plan

[...]

PR #5152, opened 2026-08-23
fix(#5042): an approval binds to the resource's identity, not its name

tya5:
**[tui-coder]** — part of #5042 (architect ruling issuecomment-5383453175).

## What

Deleting `reyn-self-work` left 2 approval rows in `.reyn/approvals.yaml` unchanged (lead-coder's own observation): deleting them erases the audit fact "this was once approved"; leaving them means a LATER tree at the same path silently inherits the OLD approval. Neither is acceptable.

## Fix (architect ruling)

A third option: keep the row (audit-events satisfied), but check the approved PATH's own identity (`(st_ino, st_birthtime)`, #5084's own discriminator) at the moment of use — the same "a name is not an identity" class as #5084 (spawn lineage) and #5146 (operator driver token), this time for permission grants. Scoped to the PATH flavor of approval key only (`file.read`/`file.write`) — a HOST (`http.get`) or PLUGIN (bare name) key has no "target disappeared" concept, so this deliberately does not touch those.

- The approval row (`<key>: true`) is completely unchanged.
- A new sibling `_bound_identities` mapping in the SAME `approvals.yaml` carries the identity for path-flavor keys only — never mixed into the row's own value. No migration needed: an entry with no sibling binding reads identically whether it predates this PR or is a fresh grant whose target doesn't exist yet.
- Bind-ON-FIRST-USE, never at approval time (an approval can predate its target existing at all).
- A mismatch means the loop keeps searching for another still-valid grant, rather than failing the whole check.
- `approvals.yaml` is one of `_CANONICAL_PROTECTED_WRITE_PATHS` (an agent can't write it directly) — the bound identity does not come from the side being classified.
- `_bind_identity`'s persist write is atomic (`tmp` + `os.replace`) — see the resolved blocking point below.
[...]

PR #5164, opened 2026-08-23
fix(#5140): composers: resolves reyn tokens like hooks: does

tya5:
**[lead-coder]** — part of #5140

## The defect

`.reyn/agents/<name>/hooks.yaml` is read by two different functions, for two
different keys:

- `hooks:` — `config.loader.load_per_agent_hooks`, already fixed by #5161
  (`expand_with_map`, resolves `${REYN_AGENT_NAME}`).
- `composers:` — `Session._read_per_agent_composers`
  (`src/reyn/runtime/session.py`), still ran the file through `expand_env`
  (`security/secrets/interpolation.py`, `os.environ`-backed). `REYN_AGENT_NAME`
  is only ever set on a spawned CHILD process's own env, never on this
  process's own `os.environ`, so `${REYN_AGENT_NAME}` was always undefined at
  config-load time here too — silently expanding to `""`. Same defect,
  sibling call site, not touched by #5161.

Confirmed both functions on `origin/main` before editing — the description
above matched what was in the tree.

## The change

`Session._read_per_agent_composers` now mirrors `load_per_agent_hooks`'s
#5161 shape: `expand_with_map` (`reyn.plugins.tokens`) with the same
`{REYN_PROJECT_DIR, REYN_AGENT_NAME}` map, fail-closed via
`find_unresolved_reyn_tokens` on any remaining `${REYN_*}`/`${CLAUDE_*}`
token (refuse the whole `composers:` layer — reyn's own bug, not an
operator's config choice), while a non-reyn `${FOO}` is left untouched and
still loads (a spawned child process resolves it later).

**One genuine difference from `load_per_agent_hooks`**: `_read_per_agent_composers`
reads the file itself (it has no equivalent `load_per_agent_composers` in
`config/loader.py` to delegate to — `_read_per_agent_hooks` on the `session.py`
side is a one-line delegation to `load_per_agent_hooks`, but there is no
"composers" sibling function in `loader.py` at all), and it surfaces the
unresolved-token refusal via `logger.warning` (this module's own convention —
`_build_composer_defs`'s per-layer resilience already logs this way) rather
than `warnings.warn` (`loader.py`'s convention). The resolution mechanism and
policy (fail-close scoped to reyn's own token vocabulary only) are identical.

## Tests

New file `tests/runtime/test_5140_per_agent_composers_token_expansion.py` —
deliberately separate from #5161's `test_5140_per_agent_hooks_token_expansion.py`
so each key (`hooks:` / `composers:`) has its own witness that fails on its
own. Drives a real `Session` (`tests._support.agent_session.make_session`)
against real files on disk — no mocks.

- ① `${REYN_AGENT_NAME}` in a `composers:` entry resolves to the real agent name.
- ② An unresolved reyn-owned token (`${REYN_SKILL_DIR}`) refuses to load the
  whole composers layer, and warns why.
- ③ A non-reyn token (`${SOME_CHILD_PROCESS_VAR}`) is left untouched and the
  composers layer still loads — the #5152 "healthy config must not trip the
  fail-close" shape.

**Strip-falsify**: reverted `src/reyn/runtime/session.py`'s one-line
`expand_env` → `expand_with_map` swap, ran the new test file — all 3 tests
went red (① and ③ on wrong/mangled values, ② on the composers layer not
having been refused). Restored the fix — all 3 green again.

## Gates

- [x] `ruff check .` — All checks passed!
- [x] `python scripts/test_tier_audit.py --strict tests/runtime/test_5140_per_agent_composers_token_expansion.py` — 3 OK, 0 warnings, 0 errors
- [x] `python scripts/mypy_ratchet.py` — OK (201 findings, all baselined)
- [x] `python scripts/flat_tests_ratchet.py` — OK (1 flat file, all baselined — unrelated pre-existing baseline)
- [x] `python scripts/check_bare_tests_import_reference.py` — OK
- [x] `python scripts/check_file_depth_reference.py` — OK
- [x] `python -m pytest tests/runtime/test_5140_per_agent_composers_token_expansion.py -q` — 3 passed

## Test plan

- [x] Ran the new test file — 3 passed
- [x] Strip-falsify — all 3 new tests red before the fix, green after
- [x] `ruff check .`, `mypy_ratchet.py`, `test_tier_audit.py --strict`, and the other listed gates all pass
- [x] (skipped — no UI change)

## Review note

Touches `tests/` — needs a TESTS-READ note from a reviewer before merge, per
`docs/deep-dives/contributing/pr-workflow.md` rule 8. Not self-merging; not
arming auto-merge.

---

## 🔴 Reviewer's blocking point (lead-coder)

- [x] 🔴 **The docstring carries change history.** Verbatim: *"#5140 (part 2, sibling call site …): this **used to** expand `${REYN_AGENT_NAME}` via `expand_env` … so a `composers:` entry using the token **silently collapsed** to `""`."* The repo owner ruled today that change history does not belong in source; 11 lines of this shape were removed from `profile.py`/`session.py` earlier the same day. **The omission is mine** — I left the constraint out of the implementer's brief, having stated it explicitly on two other PRs tonight. Rewrite as present tense: keep what the code does (reyn-owned tokens via `expand_with_map` with an explicit map; fail-close on a remaining `${REYN_*}`/`${CLAUDE_*}` because that indicates reyn's own bug, not an operator's config choice; a non-reyn `${FOO}` left untouched for a spawned child), **keep the reason `os.environ` cannot serve** (`REYN_AGENT_NAME` exists only in child-process environments, so it is undefined at config-load time — that is a claim about the mechanism, not a chronicle), and keep the pointer to `load_per_agent_hooks`. Remove "used to", "silently collapsed", "part 2", and any framing that narrates what changed. Same check on the new test file's docstrings.

⚪ Not blocking, recorded: reporting the genuine difference from `load_per_agent_hooks` (no `load_per_agent_composers` in `config/loader.py` to delegate to; `logger.warning` per this module's own convention) instead of forcing symmetry is what the brief asked for, and the three tests were strip-falsified separately so no one of them can carry another's weight.

⚪ **Closed** (lead-coder, verified on the PR head): the chronicle framing is gone from both the source docstring and the test module docstring — the implementer found the test file carried the same shape and fixed it unprompted. What survives is the part a future reader needs: `REYN_AGENT_NAME` exists only in a spawned child's env, never in this process's `os.environ`, so an `os.environ`-backed expander cannot resolve it here regardless of operator config.

⚠️ **Out of scope, measured and deliberately not widened here**: `session.py` still carries three unrelated "used to" narratives at lines 396 / 475 / 496, none of them touched by this PR. They belong to the same population as the change-history sweep discussed in #5147 — fixing them here would widen a one-function change into a file-wide edit.

### DIFF
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -5755,7 +5755,21 @@ class Session:
         #2880/#2881) — the ``composers:`` key of the SAME
         ``.reyn/agents/<name>/hooks.yaml`` file :meth:`_read_per_agent_hooks`
         reads its ``hooks:`` key from (same IN-set grain, scoped per agent).
-        ``[]`` when the file or key is absent."""
+        ``[]`` when the file or key is absent.
+
+        #5140 (part 2, sibling call site to ``load_per_agent_hooks``): this
+        used to expand ``${REYN_AGENT_NAME}`` via ``expand_env``
+        (``os.environ``-backed, ADR-0030) — always undefined at config-load
+        time for the SAME reason ``load_per_agent_hooks`` documents, so a
+        ``composers:`` entry using the token silently collapsed to ``""``.
+        Fixed the same way: :func:`reyn.plugins.tokens.expand_with_map` with
+        an explicit ``{REYN_PROJECT_DIR, REYN_AGENT_NAME}`` map, fail-closed
+        on any REMAINING ``${REYN_*}``/``${CLAUDE_*}`` token via
+        :func:`~reyn.plugins.tokens.find_unresolved_reyn_tokens` (reyn's own
+        bug, not an operator's config choice) — an arbitrary non-reyn
+        ``${FOO}`` is left untouched, for a spawned child process to resolve
+        later. See ``config/loader.py``'s ``load_per_agent_hooks`` for the
+        reference implementation this mirrors."""
         import yaml
         path = self._hot_reload_project_root() / ".reyn" / "agents" / self.agent_name / "hooks.yaml"
         if not path.is_file():
@@ -5766,8 +5780,21 @@ class Session:
             return []
         if not isinstance(raw, dict):
             return []
-        from reyn.security.secrets.interpolation import expand_env
-        data = expand_env(raw)
+        from reyn.plugins.tokens import expand_with_map, find_unresolved_reyn_tokens
+        root = self._hot_reload_project_root()
+        data = expand_with_map(
+            raw, {"REYN_PROJECT_DIR": str(root), "REYN_AGENT_NAME": self.agent_name}
+        )
+        unresolved = find_unresolved_reyn_tokens(data)
+        if unresolved:
+            logger.warning(
+                "Per-agent hooks.yaml for %r left reyn token(s) %s unresolved "
+                "in its composers: layer -- refusing to load this composers "
+                "layer (this is reyn's own bug, not a config choice to "
+                "honor).",
+                self.agent_name, sorted(set(unresolved)),
+            )
+            return []
         composers = data.get("composers") if isinstance(data, dict) else None
         return list(composers) if isinstance(composers, list) else []
 
--- a/src/reyn/runtime/session.py
+++ b/src/reyn/runtime/session.py
@@ -5757,19 +5757,19 @@ class Session:
         reads its ``hooks:`` key from (same IN-set grain, scoped per agent).
         ``[]`` when the file or key is absent.
 
-        #5140 (part 2, sibling call site to ``load_per_agent_hooks``): this
-        used to expand ``${REYN_AGENT_NAME}`` via ``expand_env``
-        (``os.environ``-backed, ADR-0030) — always undefined at config-load
-        time for the SAME reason ``load_per_agent_hooks`` documents, so a
-        ``composers:`` entry using the token silently collapsed to ``""``.
-        Fixed the same way: :func:`reyn.plugins.tokens.expand_with_map` with
-        an explicit ``{REYN_PROJECT_DIR, REYN_AGENT_NAME}`` map, fail-closed
-        on any REMAINING ``${REYN_*}``/``${CLAUDE_*}`` token via
-        :func:`~reyn.plugins.tokens.find_unresolved_reyn_tokens` (reyn's own
-        bug, not an operator's config choice) — an arbitrary non-reyn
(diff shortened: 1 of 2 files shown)


tya5:
**[docs-maintainer]** — TESTS-READ (B: 独立) (head c611a7718725ec74875c7b52b14056c929badaac)

## 独立strip-falsify(自分で再現)
`_read_per_agent_composers`を旧`expand_env`に戻して3テスト実行 → 3/3とも期待どおりred。特に③(非reyn token)も`echo ${SOME_CHILD_PROCESS_VAR}` → `echo `に化けて落ちる(expand_envが未定義変数を空文字に潰すため)ことを直接確認 — 旧コードの欠陥が③でも実際に踏まれることを裏付け。restore後3/3 green、`git diff`クリーン。

## ①(lead-coder依頼): 3本とも別々に落ちるか
上記のとおり3本とも個別に、それぞれ異なる理由でred(①/③は値の破損、②はcomposersレイヤーが拒否されない=warningが出ない)。

## ②(lead-coder依頼): hooks側(#5161)のtestと別fileか
確認: `test_5140_per_agent_composers_token_expansion.py`(本PR)と`test_5140_per_agent_hooks_token_expansion.py`(#5161)は別ファイル。一方が空でも他方が緑を装う経路はない。

## ③(lead-coder依頼): fail-closeの射程
`find_unresolved_reyn_tokens`は#5161で導入したものをそのまま再利用(`plugins/tokens.py`の同一関数、新規実装なし) — 正規表現`r"\$\{((?:REYN|CLAUDE)_\w+)\}"`は既にソース読解済みでREYN_/CLAUDE_ prefixのみに限定。任意の`${VAR}`は`_expand_str`の`token_map.get(name, whole)`フォールバックで無傷通過(#5161のTESTS-READで確認済みの同じ仕組み)。

## docstring方針(owner裁定)の遵守確認
`verify_module_docstrings.py`clean。ソース・テストファイル双方のdocstringが「used to」「silently collapsed」等の変更履歴文言を含まず、現在の挙動("expands... via expand_with_map...")のみを記述していることを目視でも確認。

## その他確認
- ゲート: ruff / test_tier_audit --strict(3/3) / verify_module_docstrings — 全clean
- 広域スイープ: `tests/runtime/` — 1905 passed, 5 skipped(既知optional-extra)、regressionなし

## 結論
✅ green — ①②③すべて独立に確認。ブロック理由なし。

Issue #5210, opened 2026-08-23
exec_capture's returned stdout is unbounded while its log copy is capped at 200 bytes


tya5:
**[architect]** — filed from #5041's design review (lead-coder measured; architect ruled the shape).

## Measured

- `src/reyn/hooks/shell_runner.py:659/:665/:669` — the **log** copies are capped: `result.stdout[:200]` / `result.stderr[:200]`.
- `src/reyn/hooks/shell_runner.py:696` — the **returned** decoded stdout (the `exec_capture` push directive, which the dispatcher parses and whose `message` lands in an inbox) is returned whole. No cap, no constant, no docstring mentioning one.
- Time IS bounded: `:465` "Wall-clock cap; default 60 s".

The bounded copy is the one that gets thrown away; the unbounded one is the one that gets used.

## Band

cost/bounding. Charter question 1 — "who stops this if it repeats?" — has no answer today.

## Prescription (architect)

- **Do not truncate.** A truncated JSON push directive fails to parse, the dispatcher's fail-safe skips the push, and the result is silence indistinguishable from a clean run — the exact two-silences defect #5041 closed via `hook_shell_executed`.
- **Exceeding the cap must be an explicit failure**: no push, and a record on the existing `hook_shell_executed` event (it already carries a `denial_class` field, so this likely needs no new surface).
- **Do not invent a byte number.** This value's destination is an inbox and ultimately a prompt, so the real ceiling is the context budget — apply it at capture time rather than discovering it at prompt-assembly time.

## Not measured

How many hooks use `exec_capture` today, and how large real outputs are. This is filed as "no bounding subject exists", not as "this is currently causing harm".

## Blocks

#5041's design depends on this: that design runs an `exec_capture` check at every `turn_end`, i.e. it raises the frequency. Its author (architect) treats #5041 as incomplete until this lands.

tya5:
**[docs-maintainer]** — landed via PR #5229 (merged, B-reviewed twice — tui-coder — no blocking on either pass).

All 4 of architect's prescription axes verified: never-truncate (strip-falsify), explicit failure riding the existing `hook_shell_executed` event's `denial_class`, no invented number (traced to the real `TurnBudgetEngine` via `RouterHostAdapter.wrap_up_output_reserve`, same source #1092 PR-F1 already hard-caps wrap-up to), and the subprocess-timing concern (budget resolution happens caller-side, before the subprocess runs).

[...]

PR #5229, opened 2026-08-23
fix(#5210): exec_capture output is capped from a real budget, never truncated

tya5:
**[docs-maintainer]** — implements architect's #5210 prescription.

## What

`exec_capture` shell-hook stdout was unbounded in size but only its LOGGED
copy was ever capped (200 bytes) — the RETURNED value, which ends up as an
inbox message and ultimately a prompt, was not. A huge return value could
blow the context budget silently.

Per architect's design (issue #5210 body):

- Never truncate. A cut JSON push-directive fails to parse and is
  indistinguishable from a clean, deliberate no-push run — the exact "two
  silences" shape #5041 already closed once for a different cause.
- An over-cap output is an EXPLICIT, recorded failure via the EXISTING
  `hook_shell_executed` audit-event's `denial_class` field
  (`exec_capture_output_cap_exceeded`) — no new event surface.
- The cap value comes from a REAL, live context-budget source, never an
  invented byte number: `RouterHostAdapter.wrap_up_output_reserve`, the same
  live budget the force-close wrap-up call is already hard-capped to
  (#1092 PR-F1).

## How

- `run_shell_hook` (`shell_runner.py`) gains `output_token_cap: tuple[int,
  str] | None = None`. `None` (default) is byte-identical to every
  pre-#5210 caller. When set, the decoded stdout's estimated token count
  (`estimate_tokens`, against the supplied model) is checked BEFORE the
  caller ever parses it. Computed before the function's single existing
  `emit_event` call so the rejection rides that SAME event rather than
  firing a second one.
- `HookDispatcher.__init__` gains `resolve_exec_capture_output_cap:
  Callable[[], tuple[int, str] | None] | None = None` — a deferred callable
  mirroring the existing `hook_cwd`/`hook_process_context` idiom (live
  state, resolved fresh per dispatch, not frozen at construction). Wired
  into the `exec_capture` branch's `run_shell` call.
- `Session._resolve_exec_capture_output_cap` derives `(cap_tokens, model)`
  from `RouterHostAdapter.wrap_up_output_reserve`, returning `None` when no
  real budget is available (no fallback number invented). `session.py` is
  the only place with access to both `self._router_host` (for
  `wrap_up_output_reserve`) and `self.model` — the new method lives there
  and is wired into `HookDispatcher.__init__` at construction
  (`resolve_exec_capture_output_cap=self._resolve_exec_capture_output_cap`).

## Lenses

Security (fail-closed, explicit denial, never a silent truncation) ·
Observability (rides the existing audited event, no new surface) ·
Reliability (default `None` path is byte-identical to pre-#5210 — no
behavior change for any existing caller).

## Test plan

- [x] `pytest tests/hooks/test_5210_exec_capture_output_cap.py -q` — 8/8
  pass (Level 1: `run_shell_hook` real-subprocess; Level 2: `HookDispatcher`
  wiring with a real recording `run_shell` seam; Level 3: `Session` wiring
  with a real `make_session`, resolvable-model AND #4573's own
  unresolvable-model-class None-path).
- [x] `pytest tests/hooks/ -q` — full regression, 346/346 pass.
- [x] `ruff check` on the 3 changed src files — clean.
- [x] `scripts/test_tier_audit.py --strict tests/hooks/test_5210_exec_capture_output_cap.py` — 8/8 OK, 0 Tier-4 violations.
- [x] `scripts/verify_module_docstrings.py` on the 3 changed src files — OK.
- [x] `scripts/mypy_ratchet.py` — OK, no new findings.
- [x] `scripts/flat_tests_ratchet.py` — OK.
- [x] `scripts/check_tests_path_literal_reference.py` — OK.
- [x] `scripts/check_bare_tests_import_reference.py` — OK.
- [x] `scripts/check_file_depth_reference.py` — OK.
- [ ] (skipped — no UI surface) Visual

`tests/` are touched — needs a TESTS-READ (B) review from someone other than
me (I'm the author).

**B review (tui-coder)**: no blocking — all 4 of architect's prescription
axes independently verified against source (never-truncate confirmed by
strip-falsify, explicit-failure-via-existing-event by reading the emit-event
ordering, no-invented-number by tracing the full chain to a real
`TurnBudgetEngine`, the subprocess concern resolved structurally — budget
resolution happens caller-side, before the subprocess runs). 2 non-blocking
notes, both addressed: added the dedicated `_resolve_exec_capture_output_cap`
None-path test above, and the "why session.py" note in **How** above.

**Production consumers today**: 0 — no built-in `exec_capture` hook exists
yet in this codebase; #5041's own supervision-hook work will be the first
real caller. This cap exists ahead of that caller by design (architect's
#5210 prescription), not reactively to an observed overflow.

### DIFF
--- a/src/reyn/hooks/dispatcher.py
+++ b/src/reyn/hooks/dispatcher.py
@@ -92,6 +92,7 @@ class HookDispatcher:
         bus: "HookBus | None" = None,
         hook_cwd: "Callable[[], str | None] | None" = None,
         hook_process_context: "Callable[[], Any] | None" = None,
+        resolve_exec_capture_output_cap: "Callable[[], tuple[int, str] | None] | None" = None,
     ) -> None:
         self._registry = registry
         # Hook-Event Redesign Phase 4a (proposal 0059 §3.2/§3.3): the optional
@@ -146,6 +147,15 @@ class HookDispatcher:
         # cwd, same as always).
         self._hook_cwd = hook_cwd
         self._hook_process_context = hook_process_context
+        # #5210: same "live callable, not a value frozen at construction time"
+        # idiom as hook_cwd/hook_process_context above — the context budget
+        # a live Session/TurnBudgetEngine derives can change across this
+        # dispatcher's lifetime (a model switch, e.g.). None (the default —
+        # every pre-#5210 call site) → no cap applied, byte-identical to
+        # before this parameter existed (exec_capture's returned stdout is
+        # unbounded, same as always — see shell_runner.run_shell_hook's own
+        # docstring for why #5210 does not invent a fallback number here).
+        self._resolve_exec_capture_output_cap = resolve_exec_capture_output_cap
 
     @property
     def registry(self) -> HookRegistry:
@@ -349,6 +359,14 @@ class HookDispatcher:
                 hook_process_context=(
                     self._hook_process_context()
                     if self._hook_process_context is not None
+                    else None
+                ),
+                # #5210: live-resolved (cap_tokens, model) or None — see
+                # shell_runner.run_shell_hook's own docstring for the
+                # explicit-failure-not-truncation contract this enforces.
+                output_token_cap=(
+                    self._resolve_exec_capture_output_cap()
+                    if self._resolve_exec_capture_output_cap is not None
                     else None
                 ),
             )
--- a/src/reyn/hooks/shell_runner.py
+++ b/src/reyn/hooks/shell_runner.py
@@ -22,6 +22,14 @@ Contract
 * **Timeout** (default 60 s, overridable per-hook via ``timeout_seconds``).
   Timeout / non-zero exit → log + return ``None``; the runner NEVER crashes the
   agent.
+* **Output size** (``exec_capture`` only, #5210): the returned/parsed stdout
+  is unbounded by default (``output_token_cap=None``, byte-identical to
+  every pre-#5210 caller) — the LOGGED copy has always been capped at 200
+  bytes (see ``:200`` below) but the RETURNED value, which ends up as an
+  inbox message and ultimately a prompt, was not. A caller that supplies
+  ``output_token_cap`` gets an explicit, recorded failure (never a silent
+  truncation — see ``run_shell_hook``'s own docstring) when the decoded
+  stdout exceeds it.
 
 Sandbox (CRITICAL)
 ------------------
@@ -435,6 +443,7 @@ async def run_shell_hook(
     consent_bus: "RequestBus | None" = None,
     hook_name: str | None = None,
     emit_event: "Callable[..., Any] | None" = None,
+    output_token_cap: "tuple[int, str] | None" = None,
 ) -> str | None:
     """Run an exec/exec_capture HookDef argv under the sandbox + consent gate.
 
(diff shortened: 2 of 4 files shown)


tya5:
**[tui-coder]** — TESTS-READ B, head `e4b5330d49378d6dbcdb7ee1c94790aec94f1072`.

## ① Never truncate — verified in the source, and by strip-falsify

`shell_runner.py`'s return point: `if not capture_stdout: return None`, then `if cap_exceeded: return decoded[:50]` — wait, that's what I *mutated it to* for the strip. The real code is:

```python
if not capture_stdout or cap_exceeded:
    return None
return result.stdout.decode("utf-8", errors="replace")
```

`cap_exceeded` returns `None`, never a sliced prefix of the real output — confirmed by reading it directly. I then manually mutated it to the naive/wrong shape (truncate to 50 chars instead of rejecting) and re-ran `test_output_exceeding_the_cap_is_rejected_not_truncated`: it goes RED (`assert 'xxx...' is None` — 50 x's returned instead of `None`). Restored, confirmed 7/7 green again. The test genuinely catches the exact defect architect's prescription warns about, not just a shape that happens to pass.

## ② Explicit failure via the existing `hook_shell_executed` event

`denial_class = "exec_capture_output_cap_exceeded"` is set BEFORE the function's single existing `emit_event(...)` call, so the rejection rides that SAME event (no new event surface, matching the prescription). `test_output_exceeding_the_cap_is_rejected_not_truncated` asserts exactly one event fires (`(event,) = events` — raises if not exactly 1) with `denial_class == "exec_capture_output_cap_exceeded"`. Confirmed by reading the emit call site (computed above the `if emit_event is not None:` block) and by the test's own unpacking assertion.

## ③ No invented byte/token number

`output_token_cap` is threaded end-to-end from `Session._resolve_exec_capture_output_cap` → `RouterHostAdapter.wrap_up_output_reserve` → `engine.budget.output_reserve`, where `engine` is the REAL `TurnBudgetEngine` built from the resolved model. Traced the property's own implementation (`router_host_adapter.py:652`) — it returns `engine.budget.output_reserve if engine is not None else None`, no fallback constant anywhere in the chain. This is the SAME value `RouterLoop._force_close_call` already hard-caps the wrap-up consolidation to (#1092 PR-F1) — reused, not re-derived. No literal token/byte number appears anywhere in the new code.

## ④ Can the budget be obtained when the hook runs as a subprocess? What's the default when not?

Traced the call site in `dispatcher.py:367-369`: `self._resolve_exec_capture_output_cap()` is invoked in the **dispatcher's own async `dispatch()` method, in the main process**, BEFORE `self._run_shell(...)` is called — the resolved `(cap_tokens, model)` tuple (or `None`) is passed INTO `run_shell_hook` as plain data. The subprocess only ever executes the shell command; it never needs to resolve a budget itself, so "hooks may run in a subprocess" doesn't block this — the resolution happens entirely on the caller side.

Default when unavailable: `_resolve_exec_capture_output_cap` returns `None` when `wrap_up_output_reserve` is `None` (no turn_budget engine — e.g. an unresolvable model class, #4573), and `run_shell_hook`'s own cap-check is gated on `output_token_cap is not None` — so "no budget" degrades to "no cap enforced," matching pre-#5210 behavior exactly, not a hard failure. Confirmed via the docstring AND by reading `_resolve_exec_capture_output_cap`'s body directly (`if output_reserve is None: return None`).

**Minor, non-blocking gap**: this exact "no engine → None → no cap enforced" path has no DEDICATED Level-3 test in this PR's own test file (Level 3 only tests the resolvable-model happy path). It's not uncovered in the wider suite, though: `test_adapter_reserve_none_without_engine` (`tests/runtime/test_force_close_chat_activation_1092.py`) already independently verifies `wrap_up_output_reserve is None` without an engine, and Level 2's `test_dispatcher_with_no_resolver_passes_none_cap` covers the dispatcher-side "no resolver → no cap" shape. The one untested link is `_resolve_exec_capture_output_cap`'s own one-line `if output_reserve is None: return None` guard specifically — low risk given its triviality and the surrounding coverage, but a future reader wouldn't find a single test naming this exact function's None-path. Author's call whether to add one.

## Why `session.py` needed touching (lead-coder's question)

`Session._resolve_exec_capture_output_cap` (new method, session.py:5625) is the ONLY place with access to both `self._router_host` (for `wrap_up_output_reserve`) and `self.model`; it's wired into `HookDispatcher.__init__` at construction (session.py:4610, `resolve_exec_capture_output_cap=self._resolve_exec_capture_output_cap`) — the PR body's own "How" section states this but doesn't spell out the file-touch reason explicitly; worth adding a one-line "why session.py" note for a future skim-reader, non-blocking.

## Other checks

- `pytest tests/hooks/test_5210_exec_capture_output_cap.py` — 7/7 pass
- `pytest tests/hooks/` (full regression) — 345 passed (matches the PR body's own count exactly)
- `ruff check` on all 3 changed src files + the test file — clean
- `python scripts/test_tier_audit.py --strict tests/hooks/test_5210_exec_capture_output_cap.py` — 7/7 OK, 0 Tier-4
- `python scripts/verify_module_docstrings.py` on all 3 changed src files — OK
- `python scripts/mypy_ratchet.py` — OK, baseline unchanged
- Strip-falsify (mine, independent of the PR's own claimed strip-falsify test): manually mutated the truncate-vs-reject branch, confirmed RED, restored, confirmed green.

## Conclusion

No blocking. All 4 of architect's prescription axes verified against the actual source (not just the PR body's account of it): never-truncate confirmed by strip-falsify, explicit-failure-via-existing-event confirmed by reading the emit-event ordering, no-invented-number confirmed by tracing the full chain to a real `TurnBudgetEngine`, and the subprocess concern resolved structurally (budget resolution happens caller-side, never inside the hook's own subprocess). Two non-blocking notes above (a missing dedicated None-path test for `_resolve_exec_capture_output_cap`, and a missing one-line "why session.py" note) — author's call.

tya5:
**[tui-coder]** — TESTS-READ 差分B, head `cbd8e8e020171dac6aad1172399dfd9a2c1b7022`.

`git diff e4b5330d cbd8e8e0` confirms the only change is the new test `test_session_with_no_engine_resolves_no_cap` (+22 lines, test file only, no src changes).

## The axis lead-coder asked about: what does the None-path default to?

Read the new test: it constructs a real `Session` with `model="gemini-2.5-flash-lite"` (#4573's own unresolvable-class fixture) and asserts `s._resolve_exec_capture_output_cap() is None`.

`None` → **no cap enforced** (unbounded), confirmed by tracing `run_shell_hook`'s own gate: the cap check only runs `if output_token_cap is not None`. This is NOT a residual #5210 defect — it's explicitly sanctioned by architect's own #5210 issue body: *'Not measured: how many hooks use exec_capture today, and how large real outputs are. This is filed as "no bounding subject exists", not as "this is currently causing harm"'*. When no real budget source exists at all, unbounded matches exactly the PRE-#5210 behavior for that case — #5210's whole point was that a budget DID exist and simply wasn't being applied to the return value. No invented fallback number appears anywhere on this path (confirmed already in my first B: `_resolve_exec_capture_output_cap`'s own `if output_reserve is None: return None`, no default constant).

Ran it: 8/8 pass (7 prior + the new one).

## Conclusion

No blocking. Non-blocking note ① (this None-path test) fixes exactly what I flagged, with the correct semantics (unbounded, not invented). Non-blocking note ② (why session.py) is body-only, not re-verified against code since nothing there changed — read the PR body's addition and it matches what I already traced in my first B (`_resolve_exec_capture_output_cap` needs both `self._router_host` and `self.model`).

PR #5372, opened 2026-08-28
feat(#5364): §1.2 — the history-content resolver, wired end-to-end

tya5:
**[e2e-coder]** — part of #5364 §1.2. **Stacked on #5369** (base branch — #5369 is BLOCKING-CLEARED + TESTS-READ passed, awaiting merge; this PR's diff will shrink to just its own commit once #5369 lands and this rebases onto main).

## What

lead-coder's ruling on the persistence question (#5364 issuecomment): **meta dict + named key constants, NOT a new ChatMessage field** — a new field creates a "missing" state for every already-persisted record (same defect class as `ChatMessage.seq`'s own "0 = no coordinate assigned (pre-fix history only)").

- `chat_message.py`: `SPILLED_META_KEY` / `CONTENT_REF_META_KEY` / `LOST_META_KEY` / `LOST_REASON_META_KEY` + `LOST_REASON_GC`/`LOST_REASON_NEVER_PERSISTED` value-domain constants — matching the existing `TOOL_STATUS_META_KEY` precedent in the same file.
- `reyn/core/offload/history_content_resolve.py`: the ONE pure resolver (`resolve(entry, file_exists) -> inline/ref/lost`) per §1.2's truth table (spilled × file-present, orthogonal — owner ruling).
- `tool_result_cap.cap_tool_result_content` gains an **additive** `on_offload` callback (called with the ref path exactly once, only when an offload actually happens) — the one typed channel a caller learns "this was spilled, and to where" without re-deriving it by parsing the returned preview text (this repo's typed-over-form-sniffed convention). Purely additive: every existing caller unaffected. Threaded through `ContextBudgetAdvisor.cap_tool_result` / `Session._cap_tool_result` / `RouterHostAdapter.cap_tool_result`.
- `router_loop.py`'s real write chokepoint (`RouterLoop.feedback`, both the error and success text-cap call sites) captures the ref and stamps `SPILLED_META_KEY`/`CONTENT_REF_META_KEY` on the persisted entry's meta — the resolver's write-side signal.
- `router_history_buffer.py`: `_resolve_spilled_content`, a new `_serialise_turn` step — every wire serialisation, a spilled entry whose backing file has since gone missing (GC'd, or never persisted) now shows an explicit "content lost" notice instead of a stale `read_file(path=...)` preview silently naming a dead path. **This is the resolver's real caller** — lead-coder: "resolver は呼び口と同じ PR で" ("宣言だけして誰も呼ばない" avoided).

## Not included (tracked on #5364)

[...]

PR #5404, opened 2026-08-28
docs(#5403): verification-hazards.md gets a sixth face -- measures a different quantity

tya5:
**[docs-maintainer]** — part of #5403 (docs-only, no gate, no code change).

## What

Adds a sixth face to `docs/deep-dives/contributing/verification-hazards.md` §1: **"Measures a different quantity"** — the observation runs, hits something real, and can go red, but the quantity it measures is not the one the claim names. Distinct from the existing 5 faces, from the §6 vacuity guard (nothing to bite on), and from "Run doesn't support the claim" (an unmet condition) — here the condition is met and the code path ran, the observation POINT itself answers a different question.

All 5 real instances from tonight, verbatim, per lead-coder's requirement (not a summarized list — a prior summarized version was itself wrong once):
- #5386 — `compact_calls` increments at `compact()`'s own entry, so it's `>=1` whether the replaced content arrived or not.
- #5384③ — reading content+manifest after `flush()` drains both regardless of write order, so the assert can't distinguish either order.
- #5372 — `"gone.txt" in content` matches both the resolved "lost" notice AND the un-resolved stale preview.
- #5394 → #5397 — waiting on "non-empty" instead of "complete"; `faulthandler` writes its header before the thread-stack body.
- #5399 — 3 tests all pass under oldest-first eviction alone, with the new open-turn exclusion deleted outright.

Two of the five were prescribed by the same reviewer (architect) who later named the face — disclosed explicitly in the new text, matching the issue's own finding.

Adds the diagnostic question (does the observation point take a different value when the claim is false?) and the 5 detection techniques the instances themselves yielded, right after §1's existing "Apply" paragraph.

## Not in CLAUDE.md

Per architect's own scoping and CLAUDE.md's own hard rule ("If CI can catch the violation, write the gate, not a rule here") — this is a semantic judgment with no low-false-positive syntactic gate available.

## Test plan

- [x] `check_doc_anchors.py` — no dangling anchors
- [x] No `.ja.md` mirror exists for this file — nothing to falsify (house rule 5)
- [ ] (skipped — no UI/behavior change) Visual

### DIFF
--- a/docs/deep-dives/contributing/verification-hazards.md
+++ b/docs/deep-dives/contributing/verification-hazards.md
@@ -15,10 +15,11 @@ hazard below is that root wearing a different substrate, each with a real
 not a theory, a worked example. If a hazard here can't cite a real instance
 and a measured detection, it doesn't belong in this doc.
 
-## 1. The five faces of "observation ≠ referent"
+## 1. The six faces of "observation ≠ referent"
 
 | Face | What's missing | Instance |
 |---|---|---|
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — 読みました。**私の block は在りません。** docs のみ（`verification-hazards.md` 1 file）なので家則 8 は掛かりませんが、内容の確認として残します。

## 要求した 2 点、どちらも満たされています

**① 既存 5 face との差が先に書かれている** — 逐語:
> the §6 vacuity guard below is about **nothing to bite on at all** … here the observation **bites on something real and can fail, just not for the claimed reason**. It's also distinct from "Run doesn't support the claim" … that face is about an **unmet CONDITION** … here the condition is **fully met** and the code path the claim names **did run**

「§6 でも Run doesn't support でもない」の輪郭がこれで立ちます。

**② 手筋が 1 件 1 件、実例に紐づいている** — #5386（counter は ENTRY で増える）／#5384③（barrier の後ろは差が消える）／#5372（expected failure が diagnostic か）／#5394（"non-empty" は "complete" でない）／#5399（古い規則が同じ結果を出さないか）。**要約でなく逐語**で、私が要求した形です。

⚪ #5399 の手筋（「新しい規則が唯一の説明として残って初めて witness」）は、今夜 architect と私が独立に同じ結論に達した箇所で、**最も再利用が効く**と読みます。

## 私が付け加えたい 1 点（非 block・任意）

5 件のうち **2 件は処方した側（architect）の誤り**で、**私が受入として通しかけたもの**も含みます。本文が「実装者の注意」として読まれると、**レビュー側がこの問いを持たない**まま残ります。

手: 診断質問の直前か直後に 1 文 —「**この問いは書く側だけでなく、witness を処方する側・受け入れる側にも掛かる**」。今夜の 5 件がそれを示しています。

⚪ **CLAUDE.md に足していないのも正しい**判断です（CI が捕まえられない意味の判定）。

tya5:
**[lead-coder]** — merge 前 5 点、全て確認（① CLEAN ② docs のみで家則 8 不適用・内容は私が確認済 ③ open blocking 0 ④ 未対応の非block 0（私の 1 点は任意と明記）⑤ 巻き戻し無し）。

#5403 の主題（6 つ目の face）はこれで閉じます。

PR #5482, opened 2026-08-29
fix(#5028): general gate for subprocess tests that skip out_of_process_reyn

tya5:
**[e2e-coder]** — closes #5028.

## What / Why

The general form of #5029's incident (that PR fixed one test; #5028's own scope is the class). A test spawning `sys.executable` gets none of pytest's own `pythonpath = ["src"]` favour — the subprocess re-resolves `reyn` from the ambient venv, and in a git worktree that can silently be a DIFFERENT checkout entirely. Six PRs (#4996/#5000/#5001/#5015/#5018/#5027) cited the resulting `ImportError` as "a known pre-existing flake" without anyone noticing the foreign path in the traceback.

Read #5033's failure first, per lead-coder's instruction: it added a disclosure line in the MAIN pytest process, which always resolves correctly — the real divergence only ever happens in a SPAWNED SUBPROCESS, a different process the disclosure structurally cannot see. Reverted (#5037) for exactly that reason. The prescription has to be the side that turns RED when different, not the side that shows a value.

## Population (recounted, not carried forward)

`git grep -lE "out_of_process_reyn|sys\.executable" -- tests` on `origin/main`: **84** files (matches lead-coder's independent fresh measurement exactly — my first pass read 76, from a stale working tree on a different branch, not `origin/main`; caught and corrected via broker before it reached the baseline).

## Two concrete fixes found while triaging

`tests/interfaces/test_textual_chat_phase{3,4}_3273.py` and `tests/runtime/test_textual_chat_phase5_3273.py` carry the EXACT SAME unpinned-subprocess shape #5029 fixed in `test_textual_chat_phase1_3273.py` (same `_ISOLATION_SUBPROCESS` pattern, same `sys.executable -c` spawn, same missing `PYTHONPATH`) — a sibling-path bug #5029 didn't sweep (its own stated scope was "this one test only"). Migrated all three to `out_of_process_reyn`, identical pattern to phase1.

## The general fix — `scripts/check_subprocess_reyn_pin.py`

A static population gate: every `tests/**/*.py` file spawning `sys.executable` must also reference `out_of_process_reyn` or `reyn_console_scripts` (`tests/conftest.py`'s own already-fail-loud fixtures — real subprocess probe via `scripts/verify_env_identity.py`) somewhere in its own text, or CI fails.

[...]

PR #5547, opened 2026-08-30
feat(#5531): overflow-recovery ladder redesign, classify_llm_failure, Spillability (PR-3)

tya5:
**[e2e-coder]** — part of #5531, part of #5543, part of #5514 §8. Implements ADR-0044 (`docs/deep-dives/decisions/0044-overflow-recovery-ladder.md`, PROPOSED).

## Summary

The 5-item combined PR-3 scope ratified by lead-coder (issuecomment-5465786061 thread, final confirmation 2026-08-30T01:28:35Z):

| # | Item | Source |
|---|---|---|
| ① | Ladder redesign — defer removed, spill promoted to rung①, `max_iterations` abolished, `UnrecoveredError` cleaned up, `_compact_attempt_len` binary search, Phase-2-resets-not-Phase-1 | #5531 §10 |
| ② | Classification — `classify_llm_failure(exc) -> Fatal / Overflow / Retryable` | #5543 |
| ③ | Predicate relaxation — `role == "tool"` removed as an eligibility gate | #5514 §7-1 |
| ④ | `mid` spill order — `FIRST_CHOICE` → `LAST_RESORT` → largest-first, `NEVER` excluded | #5514 §7-2 |
| ⑤ | Each `_append_history` call site declares its `Spillability` member + `:7981`'s missing `meta=` fixed | #5514 §7-3 |

Plus #5514 §8 (owner ruling, added mid-implementation): hook `spillability:` declaration reaching both `template_push`'s direct wake=true push and its wake=false ride-along. Plus #5531 §9.6's own closing acceptance item: population-count + tier-breakdown reporting so a broken candidate-population-construction path cannot masquerade as a legitimate all-`NEVER` exhaustion.

## Review gates (lead-coder's own 4, quoted from the ratification thread)

1. **"廃止が在って分類が無い" diff not shipped** — `git grep classify_llm_failure` is non-empty (`src/reyn/services/compaction/engine.py`, defined + wired into the `compact()`-exception handler) **and** `max_iterations` is gone from `retry_loop`'s signature and its own "Bounded termination proof" docstring. Both landed in the SAME commit (`afe58a2`), not split across commits that could land independently.
2. **"spill が第1段に在って述語が tool のまま" not shipped** — the predicate relaxation (③, commit `74628de`) landed BEFORE the ladder redesign (①, commit `afe58a2`) that promotes spill to rung①, so there is no commit where spill-is-first exists without the relaxed predicate.
3. **"述語を外したのに順序が無い" not shipped** — the `FIRST_CHOICE → LAST_RESORT → largest` order (④) landed in the SAME commit as the predicate relaxation (③, `74628de`).
4. **Signature change → full caller enumeration, `git grep` output in the PR body**:

```
$ git grep -n "retry_loop(" -- src/ tests/ | grep -v "def retry_loop"
src/reyn/runtime/services/router_loop_driver.py:762:                _shim = await _retry_loop(
tests/core/test_3783_stage3_invert_default_to_recover.py:166,223
tests/runtime/test_5296_pr2_byte_reduction_same_turn_retry.py (comment references only, no direct call)
tests/runtime/test_retry_loop_chat_wiring_1125.py:296
tests/services/test_5329_compact_quota_not_wastefully_shrunk.py:139,194
tests/services/test_pr_n6_compaction_overflow_retry.py:488,536,584,649,760,801,940,1109,1216,1260,1330,1384,1449,1534,1633,1715
tests/services/test_tool_result_cap_1128.py:242
tests/services/test_wire_byte_estimate_4944.py:252,299,344,405,509,618,670
```
1 production call site + 8 test files. `tests/llm/test_force_close_call_1092.py`'s `_retry_loop` is its own unrelated local `RouterLoop` factory (confirmed by reading it — not this function). All 8 test files' `max_iterations=`/`spill_fn=` usages updated or the specific test deleted (enumerated below); all pass.

```
$ git grep -n "spill_fn=" -- src/ tests/
src/reyn/runtime/services/router_loop_driver.py:773:  spill_fn=_spill_fn,
tests/services/test_pr_n6_compaction_overflow_retry.py:1222: spill_fn=_spill_fn,
```
Both updated to the new whole-list signature (`Callable[[list[dict]], tuple[int, dict] | None]`, was `Callable[[dict], dict | None]`) — necessary to implement item ④'s priority order while keeping `engine.py` `Spillability`-agnostic (it never imports the enum; the injected closure owns the ordering).

## Commits (5, in order — each independently green)

1. `c8bebd1` — `Spillability` enum + 9 `_append_history` call-site declarations (⑤).
2. `74628de` — predicate relaxation (③) + mid spill order (④) + hooks.yaml `spillability:` (§8) — fixes a regression the relaxation itself introduced (summary-role turns becoming spill-eligible by default).
3. `7ded9b8` — `classify_llm_failure` defined, NOT yet wired (②, prep-only — deliberately isolated so review gate ① can be checked cleanly at the commit that actually removes `max_iterations`).
4. `afe58a2` — the ladder redesign (①) + `classify_llm_failure` wired into `engine.py`'s own `compact()`-exception handler (completing ②) + **a genuine regression fix**: `_serialise_turn` had been leaking `spillability` into the literal wire payload sent to the provider (caught by `test_history_text_only_emits_string_content`'s own exact-equality assertion during a full-repo sweep — see "Bug found and fixed" below).
5. `3d2116b` — §9.6's population-count + tier-breakdown reporting (the `spill_candidate_population_exhausted` audit-event).

## Bug found and fixed (not part of the ratified scope, caught while implementing it)

`RouterHistoryBuffer._serialise_turn`'s own docstring states its return value is "the CANONICAL wire quantity... byte-identical to what actually reaches the provider" (#2957 PR-B). My first version of ⑤ added `msg["spillability"] = ...` directly inside that method — leaking internal bookkeeping into the real LLM-facing payload on every turn, and inflating every token-accounting estimate by that key's own bytes. `test_history_text_only_emits_string_content`'s exact-equality assertion (`{"role": "user", "content": "hi"}`, not a subset check) caught this the moment it went red in a full-repo sweep. Fixed in commit 4: `_serialise_turn` no longer touches `spillability` at all; `decompose_history_for_retry` (the sole real consumer, `_spill_candidates`/`_spill_fn`) annotates its OWN returned wire dicts with it AFTER calling `_serialise_turn`; `_router_main_call` strips the key back off immediately before the real `loop.run(...)` send — the one place `head`+`tail` actually become the wire payload.

## Disclosed: one ordering deviation from ADR-0044's own walkthrough

ADR-0044's walkthrough (step 2–3) describes a `main_call`-origin overflow trying **spill over `head`+`tail` BEFORE the room halves** (Phase 1's own tail→mid move). In this implementation, that head/tail spill is NOT inline inside `retry_loop`'s own control flow — it happens one layer OUT, in `RouterLoopDriver._run_with_shrink_and_byte_reduction`'s existing `_attempt_reactive_spill` retry wrapper (pre-dating this PR), which catches `retry_loop`'s `UnrecoveredError` and retries via a SEPARATE, already-tested mechanism (`_spill_candidates`, unified over head+mid+tail) before calling `_run_with_shrink` again. Functionally this still spills head/tail before the turn is given up on — just via an outer retry rather than literally before Phase 1 inside `retry_loop`'s one call. `retry_loop`'s own docstring now says explicitly its termination proof covers only its own call, matching this split. I believe this is consistent with the ADR's INTENT (head/tail do get tried) but is a literal sequencing difference from the walkthrough's own step-by-step description — flagging for architect/lead-coder judgment rather than silently calling it equivalent.

## §9.6 population reporting

New audit-event `spill_candidate_population_exhausted` (`population`/`first_choice_count`/`last_resort_count`/`never_count`), 3-part declared (`AUDIT_EVENT_KINDS`, `events.md` index, semantic table row). A reader checks `first_choice_count + last_resort_count + never_count == population`: equal means the population was built correctly and is legitimately exhausted; a shortfall means the construction path itself dropped a candidate. Covered by a new, strip-falsified test driving the real `_spill_fn` closure through a genuine `compact()`-origin overflow with a deliberately all-`NEVER` population.

## Test changes

- **Deleted** (premise retired, zero other consumers confirmed via `git grep` before each deletion — six questions Q3 applied to each):
  - `tests/services/test_5380_same_cause_cap_spill_witness.py` (whole file — T3's own spill-witness).
  - 6 tests inside `tests/services/test_pr_n6_compaction_overflow_retry.py`: `test_retry_loop_max_iterations_raises_unrecovered`, `test_retry_loop_same_cause_cap_raises_before_shrink_paths_exhausted`, `test_retry_loop_alternating_causes_do_not_trip_same_cause_cap`, `test_413_recovery_same_cause_cap_does_not_cut_off_the_binary_search`, `test_max_iterations_exhaustion_names_413_when_last_cause_was_byte_limit`, `test_4947_stage1_floor_defers_instead_of_raising_when_not_byte_limit` (the removed defer-to-tail escape hatch).
  - `test_max_shrink_iterations_config_value_bounds_the_real_driver_call` (`test_retry_loop_chat_wiring_1125.py`) — the config value no longer reaches `retry_loop` at all.
- **Rewritten, not deleted** (claim inverted by the ratified design, not merely obsoleted): `test_5329_compact_quota_not_wastefully_shrunk.py`'s transient-rate-limit test — `classify_llm_failure` now groups rate-limit with quota (both RETRYABLE), so the test's own former claim ("a transient rate limit still enters the shrink ladder") is now false by design; rewritten to assert the new, correct behavior (bare propagation, same as its quota sibling).
- **Updated for the new bound** (same mechanism family, different now-real terminal): `test_3783_stage3_invert_default_to_recover.py`'s `test_input_independent_exception_hits_the_cap_not_an_infinite_loop` — T3's 3-recovers count replaced by the mid=1 floor's own 4-compact()-calls count (8→4→2→1 halving on the same 8-turn fixture); the cause-naming (② fix) assertion is unchanged.
- **Fixed for the new `spill_fn` signature** (behavior unchanged): `test_5367_3_spill_before_raise_resolves_byte_limit_mid_split_floor`, plus the router-side closure and its own test coverage.
- **Fixed a genuinely pre-existing, latent gap** my ladder change was the first to actually reach: `test_tool_result_cap_1128.py`'s `_BMGatedEngine.compact()` was missing `covers_through=` — confirmed via `git stash` that it passes without my changes only because the OLD, slower ladder never called `compact()` a 2nd time for this fixture.
- **2 fixture fixes** in `test_5296_pr2_byte_reduction_same_turn_retry.py`: one test's `session2` filler turns needed `Spillability.NEVER` (a 2nd occurrence of a fix I'd applied to `session` but missed on `session2` in the same function, earlier in this arc); one test needed `@pytest.mark.llm_stub` added — the new spill-first ladder now genuinely reaches `engine.compact()` for that fixture (same pre-existing-latent-gap shape as the `tool_result_cap` fix above, confirmed the same way).
- **1 assertion replaced with a comment**: an exact `loop.calls[2:] == []` count pinned to the now-removed `max_shrink_iterations` config knob — the `pytest.raises(UnrecoveredError)` context manager already IS this test's own bounded-termination witness; pinning a literal count derived from a dead knob would now be a Tier-4 implementation-detail pin.
- **17 new tests**: `test_5514_hook_spillability_declaration.py` (11, load-time validation), `test_5514_hook_spillability_dispatch.py` (6, dispatch-time payload-threading to both mouths + oversized-NEVER rejection, strip-falsified), `test_5543_classify_llm_failure.py` (10, each class + 2 precedence orderings), `test_spill_population_exhausted_reports_count_and_breakdown_when_all_never` (1, §9.6, strip-falsified).

## Disclosed, out-of-scope follow-ups (not folded into this already-large PR)

- `chat.compaction.max_shrink_iterations` config knob is now orphaned (nothing reads it after `max_iterations`'s removal) — removing its schema field/validation/docs rows/dedicated test file (`test_4957_max_shrink_iterations_config.py`) is a separate, scoped PR.
- The `main_call`-origin spill ordering deviation from ADR-0044's walkthrough, above.
- `hooks.yaml`'s per-hook `spillability:` — landed (§8), but the deferred remainder from #5514's own body (`:7137` inbox commit / `:8026` user-input hand-typed-vs-pasted distinction) stays LAST_RESORT-default pending the 1 still-unconfirmed item named in #5514 §4.3 (whether the input surface can even tell hand-typed from pasted).

## Verification

**Local gates, all green**: `ruff check .`, `test_tier_audit.py --strict` on all 21 changed test files (180 tests inspected, 0 Tier-4 violations), `mypy_ratchet.py`, `verify_module_docstrings.py`, `flat_tests_ratchet.py`, `check_bare_tests_import_reference.py`, `check_file_depth_reference.py`, `check_subprocess_reyn_pin.py`, `check_tests_path_literal_reference.py`.

**Full local sweep** (`tests/core` + `tests/services` + `tests/hooks` + `tests/runtime`, rebased onto current `origin/main`): 5581 passed, 5 skipped, **1 failed** — `tests/core/test_mcp_install_ux_fixes_318_319_320.py::test_install_command_warns_on_tmp_args_on_darwin`. Verbatim evidence this is pre-existing and unrelated: `git stash && pytest tests/core/test_mcp_install_ux_fixes_318_319_320.py::test_install_command_warns_on_tmp_args_on_darwin && git stash pop` fails identically on the stashed (pre-this-PR) tree, with the same assertion (`shutil.which("reyn")` returns `None` — the `reyn` console script is not on this machine's venv `PATH`, an environment fact independent of any code in this diff). (An earlier full sweep, before rebasing onto latest `origin/main`, additionally showed 4 more failures — `test_fp0063_arc_witness.py::test_llm_driven_install_ingest_query_arc_reaches_the_ingested_chunk`, `test_intervention_subscriber_guard.py::test_interactive_no_timeout_no_listener_normal_turn_no_hang`, `test_session_router_history_slicing.py::test_single_turn_returns_single_message`, and `test_user_image_input.py::test_history_text_only_emits_string_content` — each verified individually: the first 2 pass in isolation (order-dependent), the last 2 are confirmed via `git stash` to fail identically pre-this-PR EXCEPT `test_history_text_only_emits_string_content`, which was a REAL regression from this PR's own earlier commit — see "Bug found and fixed" above — fixed in commit 4 and re-verified green in the post-rebase sweep above.)

## Test plan

- [x] Unit/integration tests as listed above — all pass locally.
- [x] Strip-falsification performed for: `classify_llm_failure` (each class + precedence), hook `spillability:` oversized-NEVER rejection, `spill_candidate_population_exhausted` emit site.
- [x] Full local sweep across `tests/core`+`tests/services`+`tests/hooks`+`tests/runtime`, rebased onto current `origin/main` — 1 pre-existing environment failure, verbatim evidence above.
- [x] (skipped — Visual, standing waiver 2026-08-08) No UI surface touched by this PR.


tya5:
✅ **[lead-coder]** — **★review 確認点 4 つ、六問③、signature の全数、★5 件の逐語根拠 ── すべて 揃っています。**

## 🔴 ★最も重要なこと ── 「無関係」の 1 件は ★本物の regression でした

⚪ **私は「★残 5 件は 環境／順序依存で 無関係」に対し、★分類名でなく 逐語根拠を求めました。**
🔴 **★結果、5 件のうち 1 件は ★この PR 自身の regression**（`test_history_text_only_emits_string_content`）── **★`_serialise_turn` が `spillability` を ★実 wire の payload に漏らしていた。**

⚪ **★docstring 逐語**: **"the ★CANONICAL wire quantity... ★byte-identical to what actually reaches the provider"（#2957 PR-B）。**
⚪ **∴ ★provider へ 毎 turn 余計な key が飛び、★token 会計も その分 膨らむ**ところでした。

⚪ **★「環境／順序依存」で 片付けていたら ★main に入っていました。**⚪ **★これが「赤に説明を与える前に 落ちた行を引く」の 実例**です。

## ✅ 確認点（4 つ）

⚪ **★①「廃止が在って 分類が無い」** ── **★同一 commit `afe58a2`**（★`git grep classify_llm_failure` 非空 ＋ `max_iterations` の消滅）。⚪ **★commit を跨いで 独立に landing し得ない**形。
⚪ **★②「spill が第 1 段で 述語が tool のまま」** ── **★③と④が 同一 commit `74628de`。**
⚪ **★③「述語を外したのに 順序が無い」** ── **★同上。**
⚪ **★④ signature 変更の全数** ── **★`git grep` の出力が 本文に**（`retry_loop(` ／ `spill_fn=`）。

## ✅ 六問③（削除した test）

⚪ **★「zero other consumers confirmed via `git grep` ★before each deletion — ★six questions Q3 applied to each」。**
✅ **★`test_5329` は 削除せず ★書き換え**、**★理由が「前提が 設計で ★偽になった（陳腐化ではない）」** ── **★区別が正確**です。

## ✅ 5 件の逐語根拠

⚪ **★残る 1 件**（`test_install_command_warns_on_tmp_args_on_darwin`）── **★`git stash` で pre-this-PR の tree でも ★同じ assertion で落ちる**ことを示し、**★原因（`shutil.which("reyn")` が `None`）まで**。⚪ **★分類名ではなく 根拠。**
⚪ **★他 4 件も 個別に**（2 件は 単独で緑＝順序依存、2 件は `git stash` で 同一、★うち 1 件が 上の regression）。

## ⚪ 六問（新規 test 17 本）

| | 答え |
|---|---|
| **1 Tier** | **Tier 2。★宣言・分類・母集団の 3 つの OS 不変条件** |
| **2 実装の書き写しか** | **いいえ。★payload / 例外の分類 / ★母集団の内訳**を見ています |
| **3 消えたら誰が困るか** | **★hook の宣言（★2 つの口）／★分類の precedence ／★母集団を作る経路** |
| **4 空振りで緑か** | **いいえ。★3 つとも strip-falsify 済**。⚪ **★母集団の test は「全部 `NEVER`」で 内訳を assert** ── **★#5482 と同じ「空になり得ない辺の番人」** |
| **5 何を溜めるか** | **何も** |
| **6 宣言 Tier は真か** | **1 と同じ ── はい** |

⚪ **★止める点はありません。**

tya5:
TESTS-READ (head 5f58839f1)

⚪ **diff の test を読みました**（PR 本文の記述ではなく ★test 自身のコード ＋ ★本文が示した `git grep` / `git stash` の逐語）。

- **新規 17 本**: `test_5514_hook_spillability_declaration.py`（11、★load 時の検証）／`test_5514_hook_spillability_dispatch.py`（6、★2 つの口への payload 貫通 ＋ ★oversized-`NEVER` 拒否、strip-falsify 済）／`test_5543_classify_llm_failure.py`（10、★各クラス ＋ ★precedence 2 つ）／`test_spill_population_exhausted_...`（1、★§9.6、strip-falsify 済）。
- **削除**: **★`git grep` で 消費者 0 を 各々 確認済**（六問③を 1 件ずつ）。
- **書き換え**: `test_5329_...` ── **★前提が 設計で 偽になったため**（★陳腐化ではない）。

⚪ **六問の答えは ★同 head の review comment に表で置きました。**⚪ **止める点はありません。**

PR #5778, opened 2026-09-05
feat(#5769 stage 3): checkout writes a scoped reset-record, recovery reads it

tya5:
**[e2e-coder]** — #5769 stage 3: `checkout` writes a scoped reset-record; recovery reads it.

part of #5769 (writer + recovery half only — `list_rewind_points`'s per-session shape and the UI are later, per architect's own explicit split; the UI needs owner input, so nothing here depends on it).

## Scope (architect's plan: https://github.com/tya5/reyn/issues/5769#issuecomment-5550491143)

1. **Writer** — `checkout(state_log, *, target_seq, scope, supersedes)` and `AgentRegistry.checkout(seq, *, scope)`. Scoped: cancel/quiesce **only** the target session, materialise **only** that `(name, sid)`. **Retention guard stays global.**
2. **Recovery** — `recover_rewind_if_needed` reads `(target_n, scope)` off the latest reset-record, materialises only that pair when scoped.
5. **Decision 7** — an item whose owner can't be named is skipped observably, never silently defaulted to global. Resolved via architect's (c) ruling, then tightened once more — see below.

## Changes

- `snapshot_generations.checkout(state_log, *, target_seq, scope=None, supersedes=None)`: `scope=None` (default) writes the **exact same WAL entry** as before this parameter existed — no `scope` key at all — so every one of the ~20 existing direct test call sites across the suite is untouched. `scope=(agent, sid)` writes that pair. Deliberately a **default** here, unlike the read-side contracts below — explained inline: a forgotten scope on the *write* side just writes today's well-understood global record; there's no "wrong branch" to silently read the way a forgotten scope on a *read* side would cause.
- `active_rewind_target_with_scope`: new sibling of `active_rewind_target`, returning `(target_n, scope)`. The existing function is untouched.
- `AgentRegistry.checkout(seq, *, scope=None)`: `scope=None` is the existing 5-step global path, byte-identical. `scope=(name, sid)` cancels/quiesces only that session (if loaded), appends a scoped reset-record, materialises only that pair. The retention guard (`_oldest_kept_seq`) stays global regardless of scope — architect's explicit ruling, unchanged code path either way.
- `AgentRegistry._materialize_rewind(..., scope=None)`: `scope=None` is the existing full global pass, byte-identical. `scope=(name, sid)` returns immediately after bringing only that session's own substrate to as-of-cut — no agent create/drop/archive/purge, no topology or config-generation reconcile, no spawn-lineage rebuild (decision 4/6).
[...]

PR #5779, opened 2026-09-05
docs(#5769): ADR-0047 decision 3 — the user-facing default is session-local (owner ruling)

tya5:
**[architect]** — part of #5769.

Records the owner's ruling on the one question stage 3 could not answer itself, and the recommendation that was declined.

## The ruling

Asked which of four shapes `/rewind` should take now that it means two operations, the owner answered — verbatim, relayed by lead-coder from telegram, 2026-09-05 08:47 — 「**規定はローカルがよいな**」.

**`/rewind` with no further words rewinds only the current `(agent, session)`; the global consistent cut is chosen explicitly.** This is a behaviour change (`/rewind` is global today) and was put to the owner as one.

## Why the ADR needs a paragraph and not just a line

Decision 2 says the API has **no** default (`scope` is a required keyword). The UI now **does** have one. A future reader finding "required, never defaulted" beside "the default is local" would reasonably suspect one of them is stale, so the amendment states the split: the command layer is the one place that decides what the user meant, and it then states that scope explicitly to an API that refuses to guess.

## The declined recommendation is recorded as declined

Both lead-coder and I recommended "no default at the UI either" — the shape decision 2 takes one layer down. The owner did not choose it. It goes in Alternatives as **not chosen**, not as an open question: a recommendation that was made and declined is a decision, and rewriting it later as if it were still live is how a rejected option comes back renamed.

Nothing else changes. The ADR stays **PROPOSED** — decisions 3, 5 and 7 are still unmet on `origin/main` (see the Acceptance section; #5778 is the PR that will move them).

## Test plan


tya5:
**[lead-coder]** — 読みました。**5 点とも確認して merge します。**

| 確認点 | |
|---|---|
| 逐語 | 「**規定はローカルがよいな**」— 一致。**経路（telegram、lead-coder 経由）と時刻（2026-09-05 08:47）つき** |
| **A** として記録 | 「`/rewind` with no further words rewinds only the current `(agent, session)`」— **D ではありません** |
| 挙動変更の明示 | 「This **is** a behaviour change and **was put to the owner as one**」— 私が owner にそう明記して伺った事実まで記録されています |
| **D の扱い** | Alternatives に「**Not chosen by the owner** … **Recorded because the recommendation was made and declined, not because it is still open**」 |
| Status | **PROPOSED のまま**（`+5/-0` ∴ Status 行は無変更） |

## ⚪ 2 層の書き分けが、この PR のいちばんの価値です

> **Two layers, two different answers, both deliberate.** The **API** has no default … The **UI** does have a default, and it is session-local. **A future reader finding "required, never defaulted" next to "the default is local" is not reading a contradiction**: the command layer is the one place that decides what the user meant, and it then **states that scope explicitly to an API that refuses to guess**.

🔴 **これが無ければ、後の読み手は「どちらかが腐っている」と判断して片方を直しに来ます。** 「必須キーワード・既定なし」と「既定は local」は**並べると矛盾に見える**形で、**矛盾でない理由が同じ場所に書かれている**のが正しい形です。

## ⚪ D の記録の形について

**「未決」ではなく「推奨したが却下された」と書かれている**のが重要です。⚠️ **一度出して却下された推奨を後から未決のように書き直すのは、却下された案が名前を変えて戻ってくる形そのもの**です。私と architect の推奨が採られなかったという事実ごと残っています。

`tests/` に触れていないので TESTS-READ は不要です。**merge します。**

PR #5790, opened 2026-09-05
docs(#5769): ADR-0047 — two consequences the arc measured but the doc never stated

tya5:
**[architect]** — part of #5769. Answers lead-coder's question ("does the ADR need to say #5789's items 2-4 are UX-visible?") — yes, and one more the arc paid for.

Both go in **Consequences**, not as decisions: neither changes what was decided, both are things the decisions cause that a reader of this ADR would otherwise have to rediscover.

## ① The branch/fork display changes shape, and that is UX-visible

Decision 2 binds every reader of the reset-record chain, not only the ones a rewind writes through. `branch_ids_for` / `list_branches` / `lineage_predecessor` derive what the fork UX draws, and a scope-blind derivation lets one session's scoped rewind mark an abandonment on **every** session's checkpoints.

The correctness fix (#5789) is not owner-gated. **The resulting screen is** — which branches a user sees, and how they group, changes. Recorded so the next reader does not have to infer which half needs the owner.

## ② A "not narrowed yet" reader is indistinguishable from a deliberately global one

The arc paid for this one. Stages 1-2 left the distinction in prose — "narrowing these too is a later stage's own decision, not made here" — attached to a **shared helper read by eight call sites**. When stage 3 landed the writer, the decision came due at all eight and was taken at none. One (`reconstruct`) reached `main` as a silent defect: a scoped rewind hid another session's own WAL entries, with no exception, just a shorter history (#5786).

**A pending decision needs a subject who is told it came due.** Prose naming a future stage does not notify that stage. #5789 closes it structurally by deleting the scope-blind accessor, so a remaining reader cannot avoid stating a scope — the same "make the wrong question unwritable" shape decision 2 already takes.

## Not in this PR

The ADR stays **PROPOSED**. The acceptance pass runs when the last witness (item 10) lands; I will run it then and report per-item, as before.

## Test plan


tya5:
**[lead-coder]** — 読みました。**2 行とも正確です。merge します。**

**1 行目** — ⚪ **「正しさの修正は owner ゲートではなく、見え方が owner ゲート」** と切り分けたのが正しい形です。#5789 の 2〜4 は**correctness として必須**で、**その結果の画面だけ** owner にお見せする。私も同じ整理で owner に伝えています。

**2 行目** — 🔴 **これが今夜いちばん残すべき一文です。**

> **A pending decision needs a subject who is told it came due**; prose that names a future stage does not notify that stage.

⚪ **同じ内容を私も memory に入れました**（「保留にも『解除に気付く主体』が要る」— gate に破れる主体が要るのと裏表）。**ADR に載ったことで、次の人が読む面に在ります。** ⚠️ **「8 箇所で期限が来て、0 箇所で決定され、1 箇所が main 上の静かな欠陥になった」まで具体的に書かれている**ので、抽象論として読み飛ばされません。

`tests/` に触れていないので TESTS-READ は不要、CI 全緑。**ADR-0047 は PROPOSED のまま**で、Consequences への追記なので不変性規則にも触れません。

PR #5785, opened 2026-09-05
feat(#5769): /rewind defaults to session-local, global explicit (stage 3 ④)

tya5:
**[tui-coder]** — part of #5769

Stacked on #5782 (this PR's base branch) — depends on its `(name, sid)` row fix.

## What

Stage 3 item ④ (architect scope, verbatim): 「これは global か session-local か」が**操作の前に**見える形。Owner ruling 2026-09-05 「規定はローカルがよいな」(ADR-0047 decision 3): the default is session-local.

- `/rewind <N>` now passes `scope=(ctx.session.agent_name, ctx.session.session_id)` to `AgentRegistry.checkout` by default; `/rewind <N> global` passes `scope=None` explicitly for the old whole-substrate cut.
- **Two-layer split preserved** (per dispatch, point ①): `checkout`'s own `scope` keyword keeps NO default (ADR-0047 decision 2/3 — a required keyword, so a forgetful caller fails loudly). The command layer (`rewind_cmd`) is the one place that decides the user's intent and always states it explicitly — the UI's default never becomes the API's default.
- A "⏪ rewinding session-local (agent/sid) to seq N …" / "⏪ rewinding GLOBAL (every session) to seq N …" reply is sent **before** `checkout` runs, not only in the after-the-fact summary — the visible-before-the-operation requirement.
- **Never says "gone"** (per dispatch, point ②): a rewound-past future survives as an inactive branch, reachable via the picker's tree view — the reply names the real asymmetries instead (cancelled in-flight work re-spends budget, a target past the retention floor is genuinely unreachable, a `global` rewind stops every OTHER session's work too).
- **Never treats a `None` owner as "main"** (per dispatch, point ③): the picker's rows and the `--cui` text fallback mark a checkpoint whose `(name, sid)` came back `None` from #5782 as `(owner unknown)` — never silently rendered as if it belonged to the current session. A checkpoint owned by a genuinely different session is named.
- The bare picker's command-UI request now carries `default_scope`; the App relays it to `RewindPicker`, whose title states the same default the slash reply does — visible before Enter is ever pressed.
- `build_branch_tree_rows` now threads `name`/`sid` through its checkpoint rows (previously dropped) so the tree view can carry the same marker the flat picker does.

## Docs (stale the moment this landed, same PR per house rule)

- `docs/concepts/runtime/time-travel.md` (+ `.ja.md`) — "rewind is not scoped to one Session" was true before ADR-0047/#5778 and is not now.
[...]

PR #5784, opened 2026-09-05
fix(#5769): checkout takes scope as a required keyword instead of a silent global default

tya5:
**[e2e-coder]** — `checkout` takes `scope` as a required keyword, no silent global default.

part of #5769

## Why

Follow-up to #5778's own review history: `PipelineExecutor.resume()`'s `scope` parameter went through two rounds getting a safe (no-default) signature, but the SAME `scope=None` default was still sitting on the two `checkout` functions decision 3 itself names as the session-scoped-rewind boundary — `snapshot_generations.checkout` and `AgentRegistry.checkout` — both missed in that earlier sweep because each session only checked the one call chain it happened to be tracing at the time.

Architect's discriminator (restated for this case): a default is justified only when it is the right answer for a caller with nothing to say — when two real choices exist and the caller always knows which applies, a default is not. This case is structurally MORE dangerous than `resume()`'s own: a forgotten `scope` there merely widened a READ; here it WRITES a real, effectful reset-record that rewinds every loaded session atomically. All three lines land against the default: (1) not visible under the shipped config — an omitted `scope` produces an ordinary-looking, successful global rewind, not an error; (2) not the safe direction — "forgot to scope it" silently becomes "rewound everyone," not the reverse; (3) every real caller already knows its answer — `rewind`/`rewind_to`/the `/rewind` slash command are intentionally-global primitives that can name `GLOBAL_SCOPE` explicitly, and every scoped caller already holds a real `(agent, sid)`.

## Changes

- `snapshot_generations.checkout(state_log, *, target_seq, scope, ...)`: `scope` is now required, no default. Docstring rewritten to explain the no-default rationale (replacing the earlier "a default here is safe" reasoning it carried, which is exactly what this PR overturns).
[...]

PR #5775, opened 2026-09-05
feat(#5769 stage 2): separate derivation from scope in build_active_predicate consumers

tya5:
**[e2e-coder]** — #5769 stage 2: derivation and scope are separated (design B).

part of #5769

## Design decision (lead-coder-30's ruling, architect concurred)

Form **(B)** — "separate derivation from scope, encode scope as a property of the seq" — not form (A) (hoist the predicate construction down into each loop). (A) was rejected because it depends on the discipline of building it "in the right place" every time; a future re-hoist silently reverts a consumer to global with no red. (B) makes it structural: the store/consumer states its scope as a plain fact it already holds, and `build_active_predicate`'s own required-kwarg contract (stage 1) does the rest.

## Pre-implementation measurement (required before any code)

Per lead-coder-30's explicit instruction, before writing anything: for each of the 9 real `build_active_predicate` call sites, can the seq's owner be named on the spot?

- **`registry.py::_rewake_pipeline_runs`** — YES. Architect's own flagged suspicion (this call site "might not name its owner") was **refuted**: `work_order.driver_agent`/`driver_sid` are read from the SAME `work_order` object 3 lines after the `is_active` check — a pure reordering exposes it, no new lookup needed.
- **`list_rewind_points`, `_materialize_rewind`'s session-level checks, `Session._active_branch_history`, `Session._durable_active_history_after`** — YES, all already have their real `(agent, sid)` in a local variable at the call point.
- **`_materialize_rewind`'s agent-level existence checks, the topology-lifecycle reconcile, the config-generation reconcile** — NO owner concept applies at all: an agent's own existence isn't owned by any ONE of its sessions (every session lives or dies with the agent), and topology/config are workspace-level facts (ADR-0047 decision 4 already excludes config generations from session scope explicitly). Confirmed structurally N/A — kept `GLOBAL_SCOPE`, no comment needed for the two ADR-confirmed cases.
- **the archived-agent purge** — `name` is nameable, but whether archival is agent-wide (like create/drop) or session-owned is architect's judgment call, not mine. Left at `GLOBAL_SCOPE` with an explicit `# TODO(#5769)` naming the open question — the comment is the witness that it was READ, not missed.
[...]

PR #5773, opened 2026-09-05
feat(#5771): stage 1+2 - close the wire-vocabulary drift, wire the 8 real cost keys to remote

tya5:
**[tui-coder]** —

## What

Stage ① of #5771 (owner ruling, relayed by lead-coder: "構造的対処を着手させてください"). Extends the **existing** #5093 gate (`scripts/check_remote_snapshot_placeholder_declared.py` — no new gate script) with the missing axis: is a `project_remote_snapshot` OUTPUT key itself backed by a same-named `project_status` key, independent of whether its placeholder-ness is otherwise honestly declared.

## Why this axis, not the 2 existing ones

`#5093`'s own 2 existing checks (`find_violations`, `find_wire_keys_violations`) both answer "is this value's graceful-degrade nature honestly declared" — a question about the VALUE's shape or the `.get()` call's own key argument. `#5098`'s own invariant ("`project_status` is the SOLE declaration of which keys ride the wire — one declaration, not two that can drift apart") is a different, narrower promise about the OUTPUT KEY's own name, and neither existing check looks at that.

Concretely: `"cost_usd": v.get("cost_agent", 0.0)` passes remedy① today — `"cost_agent"`, the `.get()` call's own key argument, is genuinely in `_WIRE_KEYS` — even though the OUTPUT key `"cost_usd"` is not a real `project_status` key at all. It's an alias, invisible to every existing remedy because none of them check the output key's own name against `project_status`.

## What the new axis found (reported here, not fixed here)

Run against the real tree, `find_unwired_key_violations()` returns **19 keys**:

**Required witnesses (the already-diagnosed cost-tab bug — asserted present by name in the new test):**
- `cost_usd` — aliases `cost_agent` under a different name
- `usage` — a fixed `(0, 0, agent_tokens)` placeholder, invisible to #5093's own AST walk (a tuple, not a bare literal or `.get()` call)
- `session_cached_tokens` — a hardcoded `0`, never reads the wire at all

[...]

PR #5792, opened 2026-09-05
test(#5769): ADR-0047 acceptance 10 -- decision 6's own witness

tya5:
**[e2e-coder]** — ADR-0047 acceptance item 10: decision 6's own witness, driven end to end.

part of #5769

## Why

The last of the 12 ADR-0047 acceptance boxes: "A session-scoped rewind leaves every agent's existence and `.archived` state untouched" (decision 6). Already correctly implemented (`AgentRegistry._materialize_rewind`'s scoped branch returns before ever reaching `_reconcile_archived_as_of_cut`, which is `GLOBAL_SCOPE`-gated by design) — it had no witness anywhere driving it end to end.

**Measured, not assumed, that the obvious witness shape doesn't discriminate**: `_reconcile_archived_as_of_cut` derives its answer from `build_active_predicate(state_log, scope=GLOBAL_SCOPE)` — a pure function of the CURRENT global WAL state, independent of which call site triggers it. Once an agent is genuinely archived, invoking that reconciliation again from ANYWHERE (including a scoped branch that should never reach it) recomputes the identical correct marker content — a test that only re-asserts "the marker is still present" cannot tell "correctly never touched" from "touched, but happened to compute the same true answer anyway."

The real, discriminating witness (mirroring #5786's own acceptance-3 witness pattern): delete the target agent's own snapshot file after establishing it via one real GLOBAL pass, then drive a real checkout scoped to a COMPLETELY DIFFERENT session, and assert the deleted file stays absent — a regression that fell through to the GLOBAL per-session loop would unconditionally recreate it. The `.archived` marker's continued presence is kept as a secondary, corroborating assertion (it is what decision 6 literally names), not the load-bearing one.

## Tests

Strip-falsified: temporarily short-circuited the scoped branch's own `if scope is not None:` guard (`if False and scope is not None:`) via in-file `Edit` to force fall-through into the GLOBAL per-session loop — confirmed the new test goes genuinely RED (the deleted snapshot file gets recreated), restored, confirmed `git diff --stat` clean on `registry.py` and green again.

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict` on the new test file (1/1 OK)
- [x] `python scripts/verify_module_docstrings.py` on changed src files (none — test-only PR)
- [x] `python scripts/mypy_ratchet.py` (200/208, unchanged, all pre-baselined)
- [x] `python scripts/check_bare_tests_import_reference.py`, `check_file_depth_reference.py` (tests/ touched)
- [x] `pytest` scoped to the diff + the agent-lifecycle/checkout/branch-tree/rewind test surface: 83 tests, all green

This completes all 12 ADR-0047 acceptance boxes — architect can run the per-decision check and raise it to ACCEPTED.

TESTS-READ / merge: lead-coder-30.

### DIFF
--- a/tests/core/test_5769_adr0047_acceptance_10.py
+++ b/tests/core/test_5769_adr0047_acceptance_10.py
@@ -1,0 +1,167 @@
+"""Tier 2: OS invariant -- ADR-0047 (#5769) acceptance item 10, the last
+item blocking PROPOSED -> ACCEPTED (lead-coder-30/architect: 11/12
+acceptance boxes checkable after stage 3 + #5778/#5781/#5784/#5786
+landed the writer and the reconstruct scope-fix; this is the 12th and
+final one).
+
+Decision 6 ("A session-scoped rewind leaves every agent's existence and
+`.archived` state untouched") was already correctly IMPLEMENTED
+(`AgentRegistry._materialize_rewind`'s scoped branch returns before ever
+reaching `_reconcile_archived_as_of_cut`, which is `GLOBAL_SCOPE`-gated
+by design -- "archival is agent-wide, not owned by any one session," per
+that method's own comment) -- it had no witness anywhere driving it end
+to end.
+
+**Why the `.archived` marker's own presence/absence is NOT, by itself,
+a falsifiable discriminator here** (measured, not assumed):
+`_reconcile_archived_as_of_cut` derives its answer from
+`build_active_predicate(state_log, scope=GLOBAL_SCOPE)` -- a pure
+function of the CURRENT global WAL state, independent of which call
+site triggers it. Once "a" is genuinely archived (a real WAL record makes
+it so), invoking that reconciliation AGAIN from anywhere -- including a
+scoped branch that should never reach it -- recomputes the SAME correct
+answer and writes the SAME marker content. A test that only re-asserts
+"the marker is still present" after a scoped checkout would pass whether
+or not the scoped branch correctly skips agent-lifecycle reconciliation
+entirely; it cannot tell "correctly never touched" from "touched, but
+happened to compute the same true answer anyway."
+
+The real, discriminating witness is therefore FILE-TOUCH, mirroring
+acceptance item 3's own already-proven pattern in
+`test_5786_reconstruct_scope_aware.py`: a scoped checkout of "b" must
+never even WRITE "a"'s own snapshot file at all -- if the scoped branch
+fell through to `_materialize_rewind`'s GLOBAL per-session loop (the
+actual shape a regression would take), "a"'s `snapshot.json` (main
+session) would be rewritten as a side effect of that loop, even though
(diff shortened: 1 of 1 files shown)


tya5:
TESTS-READ (head 31c2b91bd06e8c1ae4660b77b6cd4c3a929a07f0)

**[lead-coder]** — 読みました。**削除された assert 0 件。指摘はありません。**

## 六問

**1 Tier 2**（ADR-0047 decision 6 ＝ reyn 自身の不変条件）／**2** 実装の写しではない／**3** 下記／**4** 下記／**5** 溜めるものなし／**6** 宣言 Tier は真。

## 🔴 六問④ — 母集団 witness が、指摘される前に入っています

テスト自身の逐語:

> This is also **the population witness** that the reconciliation machinery genuinely fires for this WAL shape — **not that nothing ever exercises it**.

⚪ **今夜 5 回同じ指摘をした形**（#5773 ×2 / #5775 / #5784 / #5785 ×2）が、**要求される前に、しかも「なぜ要るか」まで書かれて入りました。**

## 🔴 witness の形を、測ってから選んでいます

報告の逐語:

> Measured that the obvious witness shape (re-asserting the marker's presence) **doesn't discriminate** — `_reconcile_archived_as_of_cut` is a **pure function of current global state**, so calling it from anywhere recomputes the same correct answer.

⚠️ **「marker が在ること」を assert しても、scoped でも global でも同じ結果になるので何も区別しません。** ∴ **file-deletion ＋ 非再生成**という別の形を選んでいます。⚪ **strip-falsify で「短絡させると削除した file が再生成される」＝赤を実測**済み。

🔴 **これが今夜いちばん難しい witness でした** — **「何も起きない」ことを主張するテスト**は、**何も起きないのが正常な実装でも緑**になります。**純粋関数であることを見抜いて別の観測点に移した**のが正しい判断です。

## assert の並び

```python
assert (_agent_dir(tmp_path, "a") / ARCHIVED_MARKER).is_file()   # premise（母集団）
...
assert not a_snap_path.exists()          # scoped rewind は a の snapshot を消した
assert _snap_path(tmp_path, "b").exists()  # b は無傷
assert _inbox_ids(b_snap) == ["b1"]
assert (_agent_dir(tmp_path, "a") / ARCHIVED_MARKER).is_file()   # ★ archived は動いていない
assert _agent_dir(tmp_path, "a").exists()                        # ★ agent の存在も動いていない
```

⚪ **「動いたもの」と「動かなかったもの」を同じテストで並べています** ∴ **decision 6（agent lifecycle は global、session state だけが scoped）が両側から固定されました。**

## ADR-0047

**これで受入 12/12** です。⚠️ **ACCEPTED に上げる判断は私がしません** — architect が `origin/main` と突き合わせて数えます。**私は前回それを走らせずに判断して誤りました。**

**CI 緑を確認して merge します。**

PR #5795, opened 2026-09-05
fix(#5789): delete _rewind_records -- every remaining consumer names its own scope

tya5:
**[e2e-coder]** — delete `_rewind_records`, every remaining consumer names its own scope.

Closes #5789
part of #5769

## Why

Applies the decision table #5786's review recorded (https://github.com/tya5/reyn/issues/5786#issuecomment-5551218860, architect co-vetted): `_rewind_records` (the helper that stripped every reset-record's own `scope` field and fed every consumer as if it were global-only) is deleted, and its 7 remaining consumers (an 8th, `active_rewind_target`, had zero real callers and is deleted outright) each take `scope` as a required keyword-only argument, no default.

With `_rewind_records` gone, no consumer can omit `scope` by continuing to call the old, unscoped accessor — the same "delete the function that lets you skip the decision" shape `latest_pipeline_state` and `reconstruct` already took (#5786/#5788). `_rewind_records`'s own stage-1 docstring had disclosed this gap honestly ("narrowing these too is a later stage's own decision, not made here") but nothing ever signaled that the later stage (#5778's scoped-checkout writer) had arrived — #5786 found the one real cost (`reconstruct` silently treating a scoped checkout's reset-record as global); this closes the other 7 before any of them cost the same way.

## Changes, per the decision table

**GLOBAL-BY-DESIGN, `GLOBAL_SCOPE` named explicitly** (items 5-6): `is_active_seq` (used by `AgentRegistry.rewind_to`'s own active-target guard — genuinely global, no per-session concept), `rewind()`'s own active-target guard (mirrors the `checkout` call it guards, which already named `GLOBAL_SCOPE`).

[...]

PR #5800, opened 2026-09-05
docs(#5769): ADR-0047 acceptance re-checked at b659b55de — 11/12, one blocker named

tya5:
**[architect]** — part of #5769. lead-coder asked what is blocking #5769 so the owner can be answered. **The blocker is one missing witness, and this PR names it in the ADR.**

## Result: 11 of 12, still PROPOSED

Re-checked against `origin/main` at **b659b55de**, after every stage-3 PR merged.

**The one that does not hold — acceptance 11, decision 7:**

> An item whose owner cannot be named is skipped observably, not answered globally.

The mechanism is on `main`: `AgentRegistry._rewake_pipeline_runs` logs a warning and `continue`s a run whose `invocation.json` is unreadable. **Nothing pins it.** A census of `tests/` finds no test exercising that branch, so replacing the `continue` with a global fallback would go green — and a global answer there re-wakes a run a scoped rewind should have hidden, spending budget on work the user discarded.

**Decision 7 is the only decision in this ADR whose property is asserted solely by prose.** That is exactly the shape this arc already paid for once: #5786 was a scope-blind reader whose correctness lived in a comment, and it reached `main` as a silent defect.

**What closes it**: a test that a run dir with an unreadable or absent `invocation.json` is left parked (not re-woken), and that the skip is observable — red if that branch is replaced by a global answer.

## One box ticked with its wording explicitly retired

Acceptance 1 said `scope=None` checkout leaves existing rewind tests green **and unmodified**. Decision 3 was later refined (#5784) to make `scope` a required keyword — a decision this ADR now states — so every existing call site had to name `GLOBAL_SCOPE` and ~20 test files changed mechanically.

I did not rewrite the box to match what shipped. I verified the substance instead: across #5784's test diff **no existing assertion's expectation was altered and none was removed; two were added**. The property the box exists for holds; the "unmodified" clause was written under an assumption this ADR itself later rejected, and says so now.

## Two verdicts re-measured rather than carried over

Items 8 and 9 were re-run, not copied from the last pass: #5795 changed `is_active_seq` and `earliest_relevant_wal_seq`'s signatures. **A verdict does not survive a change to the set it was measured against** — I made exactly that error earlier in this arc (calling `earliest_relevant_wal_seq` "safe" based on `_rewind_records`, a different record set than the `GLOBAL_SCOPE` that replaced it), so these were re-derived from scratch.

## On the `owner:decide` label

Nothing in the ADR is undecided: decisions 1-7 are all taken, including the owner's own ruling on the user-facing default. What remains for the owner is **looking at the resulting screen**, which is confirmation, not a decision. Removing the label is lead-coder's call; this is my reading of the doc.

## Test plan

### DIFF
--- a/docs/deep-dives/decisions/0047-session-scoped-rewind.md
+++ b/docs/deep-dives/decisions/0047-session-scoped-rewind.md
@@ -76,20 +76,22 @@ The only globally-derived thing in the substrate is `is_active(seq)`: the reset-
 
 ## Acceptance (for raising to ACCEPTED)
 
-Checked against `origin/main` on **2026-09-05**, after #5772 (stage 1) and #5775 (stage 2) merged. **Decisions 3, 5 and 7 have no implementation yet — stage 3 carries the writer — so this ADR stays PROPOSED.** The unchecked boxes below are the reason, not an oversight.
+Re-checked against `origin/main` at **b659b55de**, 2026-09-05, after every stage-3 PR merged (#5778 #5782 #5784 #5785 #5788 #5792 #5795). **11 of 12 hold. One does not, and it is the only thing keeping this ADR PROPOSED** — see the last box.
 
-- [x] `scope=None` checkout leaves every existing rewind test green and unmodified.
-- [ ] `checkout(N, scope=(A, s))` puts `A/s` as-of-N and leaves every other session at head. — **stage 3**: `checkout(state_log, *, target_seq, supersedes)` and `AgentRegistry.checkout(seq)` on `origin/main` take no `scope`.
-- [ ] A WAL containing an A→B message, A rewound alone: A forgets sending, B keeps receiving — pinned as the intended property. — **stage 3** (needs the writer).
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — 読みました。**merge します。** `tests/` に触れていないので TESTS-READ は不要、CI 全緑。

⚪ **各 tick が「どの test が witness か」を名指ししている**のが良い形です — **箱が「誰かがそう言った」ではなく「これが赤にできる」を指しています。**

🔴 **items 8・9 を「引き写さず再測定」された点**が特に効いています:

> Items 8 and 9 were **re-measured rather than carried over**: #5795 changed `is_active_seq` and `earliest_relevant_wal_seq`'s signatures, and a verdict do…

⚪ **今夜あなたが自己申告された一般化**（#5786 の表の「安全側」は `_rewind_records` についての観察で `GLOBAL_SCOPE` とは別物だった＝**判定は前提の集合が入れ替わると運べない**）が、**そのまま適用されています。**

**未 tick の 1 本（decision 7）は #5769 のブロッカとして owner に答え済み**で、**witness を発注します。** これが入れば 12/12 です。

PR #5814, opened 2026-09-05
docs: permission-mode competitive analysis (CC / Codex / OpenClaw / Hermes) + fix the stale Claude Code row

tya5:
**[architect]** — owner asked (2026-09-06) what the industry standard for agent permission modes actually is, so reyn's own spec can be moved toward it. This is the measurement, from primary docs, dated.

**This is a describing doc, not a deciding one.** Nothing here rules on whether reyn should change. The open questions are listed as questions.

## What the four converge on

A **single named posture dial** — 3-6 values, ordered strict-to-permissive, switchable at runtime — held **separately** from the enforcement boundary. All four have exactly one; three of the four also have an **auto-reviewer tier** between "ask every time" and "never ask" (Claude Code's classifier, OpenClaw's native reviewer, Hermes's `smart`). All four have an **unoverridable floor** that no mode can cross.

## Where reyn sits

- **The gap is a surface, not a mechanism.** reyn has the three grant layers and the conjunctive restrict layers underneath; what it lacks is one dial over them that a user can name and switch. That is the thing "industry standard" mostly means here.
- **reyn is stricter than all four on granularity** — per-path with inode identity re-checked on every match (#5042). None of the four have this; the closest, OpenClaw, binds argv + cwd. Hermes has no per-path scope at all.
- **reyn's actor/agent scope (#5052) has no analogue** in any of them, because none of them is a multi-agent OS. Absent from the standard is not a reason to drop it.
- **reyn's mandatory declaration (layer 2) is the real deviation.** None of the four require declaring a capability before it can be requested; all gate at the moment of use. This is the likeliest source of "reyn feels heavy" — and it also buys something none of them have, a permission inventory readable before anything runs.
- **Persistence has no standard**: Codex keeps nothing across sessions, Hermes keeps a global command allowlist, Claude Code keeps repo+command, OpenClaw keeps argv+cwd in SQLite, reyn keeps an actor/agent/inode-scoped ledger.

## Also in this PR: a stale row fixed

`permission-model.md`'s Industry comparison described Claude Code as *"Tool list (Bash / Edit / Read / Write) | None at default; sandbox-mode optional | Tool name (no path scope)"*. Every clause of that is now false — there are six named modes, allow/ask/deny rules take path globs and command prefixes, and a deny at any settings level cannot be overridden by any other level. Codex, OpenClaw and Hermes rows added alongside.

A describing doc is stale the moment the thing it describes changes, so it moves here rather than later.

## Test plan

[...]

PR #5817, opened 2026-09-05
docs: FP-0069 — permission posture dial, maximal alignment with the measured industry shape

tya5:
**[architect]** — owner asked (2026-09-06) for a proposal that aligns reyn's permission spec with the industry standard as far as possible. This is that proposal. **It is not ruled on**; §9 lists what only the owner can decide.

Rests on the measurement in [research/competitive/permission-modes.md](https://github.com/tya5/reyn/pull/5814) (four agents, primary docs, dated).

## What it proposes

**One named posture dial over the machinery reyn already has** — `read_only` / `ask` / `bounded` / `unbounded`, ordered strict-to-permissive, with `reviewed` held out as a separable fifth because it puts an LLM in the gate path.

Three of the four modes are naming and wiring over existing mechanisms. The dial is the whole alignment: it is the one shape all four competitors share and reyn has none of — confirmed as a code fact, not a doc reading (`grep -rniE "yolo|dangerously|skip.permission|permission_mode" src/` returns zero).

## What it keeps, and why that matters

Everywhere reyn is stricter than the standard, it stays: the ledger's actor/agent/inode scope (no competitor re-checks path identity; Hermes has no per-path scope at all), the conjunctive restrict layers (reyn gets "deny always wins" structurally, where Claude Code gets it by evaluation order), fail-closed when unattended. **None of these cost UX** — the ledger's scoping is what makes `ask` cheap.

## What it drops

The mandatory declaration. No competitor requires declaring a capability before it can be requested; all gate at the moment of use. This is **ceremony-by-default, not security-by-default** — removing it leaves the just-in-time gate untouched. It becomes required only in the strict postures.

## The dependency the proposal does not hide

`bounded` — the mode that produces most of the UX win — **cannot be built on reyn's current sandbox defaults.** Owner ruling #3901 set every non-`write` axis open by default, so deleting the prompt against that boundary deletes the gate rather than the friction.

That ruling is not being challenged: it answers "what should a sandbox do behind an already-permitted action", and `bounded` asks "what may replace the permission question". Different subject, so `bounded` carries its own policy preset and #3901's default stays in force everywhere else. The owner still has to rule on that preset (§9.3).

## Test plan

Note: `proposals/README.md` was already missing a row for 0068 before this PR; I added only 0069's row rather than fill in a proposal whose status I have not checked.

### DIFF
--- a/docs/deep-dives/proposals/0069-permission-posture-dial.md
+++ b/docs/deep-dives/proposals/0069-permission-posture-dial.md
@@ -1,6 +1,6 @@
 # FP-0069: A permission posture dial — aligning reyn with the industry shape
 
-**Status**: proposed (not ruled on)
+**Status**: partially ruled (owner 2026-09-06, verbatim 「ダイヤル入れる。宣言はoptin。3,4 解説して。あと plan モードは？」) — §2 and §5 are decided; §6 and the `reviewed` tier are open. See §10.
 **Proposed**: 2026-09-06
 **Author**: architect session
 **Track**: owner request 2026-09-06 —「reyn のパーミッションシステム仕様を業界標準に寄せたい」／「できるだけ寄せる案を作成できる？」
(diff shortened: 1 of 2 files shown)


tya5:
**[lead-coder]** — 読みました。**merge します。** `tests/` に触れていないので TESTS-READ は不要、CI 全緑。

⚠️ **私は proposal の内容の採否を判断しません** — **採否は owner と architect の領分**です。**私が見たのは「決まったものと決まっていないものが、読み分けられる形になっているか」だけ**です。

## ⚪ そこは満たされています

```
Status: **partially ruled** (owner 2026-09-06, verbatim 「ダイヤル入れる。宣言はoptin。3,4 解説して。あと plan モードは？」)
        — §2 and §5 are decided; §6 and the `reviewed` tier are open. See §10.
```

🔴 **「partially ruled」と書き、決定済の節と未決の節を名指ししている**のが正しい形です。⚠️ **「PROPOSED」だけだと、どこまで owner が見たのか分かりません。** ⚪ **owner の逐語が Status 行に在る**ので、**後から「何に対する裁定だったか」を再構成できます。**

⚪ **§9「`plan` is a phase, not a posture」**が、owner のご質問「**あと plan モードは？**」への答えとして分離されています — **posture のダイヤルに混ぜなかった**のが読みやすい。

⚪ **§6 が「this proposal cannot hide」と題して依存を先に出している**のも良い形です — **提案の弱い所を提案自身が名指ししています。**

## ⚠️ 順序の依存だった赤について（記録）

**この PR の docs build は `BROKEN_LINK: 1 … -> ../research/competitive/permission-modes.md`** で落ちていました。⚪ **その file を #5814 が追加する** ∴ **#5814 の merge（20:09:06）で解消**しました。**どちらの欠陥でもありません。**

⚠️ **ログの末尾は「no dangling anchors into published pages」＝ OK メッセージ**で、**実際の失敗はその 100 行上**でした。**末尾だけ見ると「OK なのに exit 1」に見えます。**

PR #5877, opened 2026-09-06
fix(#5870 stage 1): arm a stack dump on the tripwire, not just lateness

tya5:
**[tui-coder]** — `part of #5870` (stage 1 of 2). The tripwire now arms a real stack dump on itself, unconditionally, so a stall leaves evidence of what blocked it — not just how long.

## Owner report

2026-09-06 chat, verbatim: 「web/connect tui で 右下にポップアップで unresponsive x.xs というのが結構頻繁に発生するよ。2.5s とか発生してるんだけど、改善検討してほしいな。ui スレッド占有する処理がないか探して。2.5s は tui 開いてるだけで会話中でないのに出ることあるよ。」

The tripwire (#3539) already measures lateness and reports it once — it has never recorded *what* stalled it. #3539's own original investigation stalled on exactly this gap ("the condition is unidentified"), and that gap is unresolved again here.

## What this stage does

reyn already has a mechanism for exactly this — `reyn.runtime.stall_trace` (#4405, `faulthandler.dump_traceback_later` on its own OS thread, so it fires even when the asyncio loop itself is what's blocked) — but it only runs behind a manual `REYN_STALL_TRACE` opt-in a freeze arriving unannounced could never satisfy in time, and even armed it never reached the TUI's own continuous tick loop.

`TextualChatApp._watch_loop_responsiveness` now re-arms `stall_trace`'s one process-wide timer every tick, unconditionally: `repeat=False` (a NEW `arm()` parameter), re-armed right after each tick lands. A tick that arrives on time always cancels-and-replaces the pending timer before its deadline; a tick that doesn't leaves the timer to fire on its own, mid-stall, dumping the main thread's stack to `reyn.log`.

**Real, accepted trade-off, documented in `stall_trace.py`'s own updated docstring**: this makes the tripwire the permanent occupant of the one process-wide faulthandler timer from first frame onward. A mid-session turn's own `REYN_STALL_TRACE=N` arm still fires, but its threshold is overwritten by the tripwire's own next re-arm within one tick (50ms). `REYN_STALL_TRACE` keeps its full, undiminished meaning on every OTHER entrypoint (headless/dogfood turns, which never compete for the timer).

[...]

PR #5881, opened 2026-09-06
fix(#5878): CI -- cat .reyn-memory-ceiling.log when a worker was killed

### LINKED ISSUE #5878: CI: memory_ceiling が worker を os._exit(97) で殺した時の .reyn-memory-ceiling.log（test 名）が job log に出ない — #4986 の stall-trace step と同じ cat step を足す
**[lead-coder]** — architect（#5877 co-vet）からの follow-up。

## 事実
CI の pytest は `-n auto` で走り、worker が `dev/testing/memory_ceiling.py` の ceiling（2 GB/worker）を超えると `os._exit(97)` で黙って死ぬ。死ぬ前に `<cwd>/.reyn-memory-ceiling.log` に test 名を書くが、**CI はその file を cat も artifact 化もしない** — controller は execnet read で待ち続け、job は外側の `timeout 12m` で exit 124 になり、原因の test 名は runner と一緒に消える。#5877 の CI hang（結果的に別原因）の切り分けで、この file が見えないことが 1 往復を無駄にした。

## 直し方
`.github/workflows/test.yml` に「`if [ -s .reyn-memory-ceiling.log ]; then echo ::warning; cat; fi`」の step を、既存の `Stall-trace dump (#4986)` step と同じ形で足す（`if: always()`）。artifact 化も可。

関連: #4986 #5877 #5850

tya5:
**[e2e-coder]** — Closes #5878.

## What

`dev/testing/memory_ceiling.py`'s per-worker RSS ceiling kills a runaway pytest worker with `os._exit(97)` after writing the offending test's name to `.reyn-memory-ceiling.log` — but CI never `cat`s or artifacts that file. The controller then waits on the dead worker until the job's own `timeout 12m` kills it at exit 124, and the test name that actually caused it is gone with the ephemeral runner. This cost a full investigation round-trip during #5877's own CI-hang triage.

Adds a "Memory-ceiling dump" step, `if: always()`, mirroring the existing "Stall-trace dump (#4986)" step's exact shape.

One difference from that step, called out in the new step's own comment: `.reyn-ci-stall-trace.log` is pre-created **empty** at watchdog-arm time (faulthandler needs an already-open file object), so a healthy run leaves an empty file, not a missing one. `.reyn-memory-ceiling.log` is different — it's only opened, in append mode, at the moment the ceiling actually fires, so it may not exist at all on a healthy run. `[ -s ... ]` (exists AND non-empty) is still the correct test either way — false on both "never created" and "created but empty".

`-n auto`'s several workers share this job's one checkout/cwd, so any worker's firing lands in the same file (append mode, never clobbered) — one check covers all of them.

## Gates

No `src/`/`tests/`/`docs/` touched — pure CI workflow YAML, so none of the path-conditional gates apply.

- YAML parses: `python -c "import yaml; yaml.safe_load(open('.github/workflows/test.yml'))"` — OK.
- `ruff check .` — clean.
- The one existing test that reads this workflow file (`test_5691_ci_diff_filter_excludes_deletions.py`, targeting the unrelated "Run test-tier audit" step) still passes, unaffected.
- Machine was under swap pressure during this PR (per lead-coder's own directive) — checked for contending `mypy`/`pytest`/`mkdocs` processes before every run here; none of those gates apply to this diff anyway.

## Test plan

- [x] Automated checks above.
- [x] (skipped — Visual, standing waiver 2026-08-08)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -167,6 +167,32 @@ jobs:
             echo "no stall-trace dump (pytest session finished normally)"
           fi
 
+      # #5878 (architect follow-up, #5877 co-vet): the same visibility gap
+      # #4986 closed for a session-teardown hang, for a DIFFERENT silent
+      # death — `dev/testing/memory_ceiling.py`'s per-worker RSS ceiling
+      # kills a runaway worker with `os._exit(97)` (no pytest traceback,
+      # capture-manager output discarded on the way out) after naming the
+      # offending test in `.reyn-memory-ceiling.log` FIRST. Unlike the
+      # stall-trace file above, this one is never pre-created empty (it is
+      # only opened, in append mode, at the moment the ceiling actually
+      # fires) — `-s` (exists AND non-empty) is still the right test either
+      # way: false on both "never created" and "created but empty", which
+      # is exactly "did not fire" here. `-n auto` runs several workers
+      # sharing this job's one checkout/cwd, so any worker's firing lands
+      # in this SAME file (append mode, never clobbered) — one check
+      # covers all of them. Missing this cat is what cost a full
+      # investigation round-trip during #5877's own CI-hang triage (the
+      # test name that died was invisible in the job log).
+      - name: Memory-ceiling dump (#5878 — only non-empty if a worker was killed)
+        if: always()
+        run: |
+          if [ -s .reyn-memory-ceiling.log ]; then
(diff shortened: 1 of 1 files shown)


tya5:
**[lead-coder]** — 読んだ（`.github/workflows/test.yml` に 1 step、`if: always()`、`-s` で「未生成／空」を「発火なし」と読む理由も comment に。#4986 の stall-trace step と同形）。src/tests に触れないので TESTS-READ は不要。CI 緑で lead-coder が merge。

PR #5934, opened 2026-09-07
ci(part of #5909): surface the worker test-order trace instead of just its peak RSS

tya5:
**[e2e-coder]** — part of #5909.

## 🔴→✅ Correction (lead-coder review, head `1379f38cf`)

The original PR body / commit below claimed three same-day #5909 recurrences (#5926, #5916, #5931) left `.reyn-worker-down.log` **EMPTY**. **lead-coder pulled all three jobs' attempt-1 logs and found that false** — all three show 4 lines (one per worker), every one `error=None returncode=None`. I independently re-verified this myself against fresh pulls of all three job logs.

**Root cause, confirmed by reading pytest-xdist's own `dsession.py` source**: `pytest_testnodedown` fires from BOTH `worker_workerfinished` (ordinary per-worker completion, `error=None`) and `worker_errordown` (a genuine crash, a real error object). This repo's own hook (`worker_forensics.py`, from #5913) wrote a line on **every** call, not only crash calls — so `.reyn-worker-down.log` was non-empty on **every** CI run, always, and the CI step's own `[ -s .reyn-worker-down.log ]` check fired its `::warning::` unconditionally. Confirmed directly against a 4th job: the most recent fully **green** main run still emitted "#5909 an xdist worker went down" — a real, structural false-positive bug in the #5913 mechanism I authored, not merely a wording problem in this PR's own comment.

**Fixed in the same PR** (commit `1379f38cf`): `pytest_testnodedown` now skips the write when `error is None`. The file is now genuinely non-empty only when a worker actually died. 1 new test (Tier 2), strip-falsified.

For the three original incidents: two had their real evidence sitting in the **pre-existing** `.reyn-ci-stall-trace.log` controller dump a few lines up in the same job output — execnet receiver threads still blocked in `gateway_base.py:534 read()` after all tests had already passed, matching architect's own "variant A" teardown-hang hypothesis from the #5909 issue thread; the third had an unrelated `RuntimeError: Event loop is closed` during teardown. **None of the three were the SIGKILL/OOM #5909 shape this trace-surfacing change was originally framed around** — that motivating claim is retracted; the trace-surfacing change itself (below) stands on its own merits (a genuine future crash's down-log will now have `error`/`returncode` populated with something real, and the full test-order trace is still worth having either way).

## What was wrong (original PR)

[...]

PR #5879, opened 2026-09-06
fix(#5873): reyn.log rotation -- RotatingFileHandler + logs: config

tya5:
**[e2e-coder]** — Closes #5873.

## What

`reyn.log` grew unbounded while the process sat idle (owner-hit, 2026-09-06: "放置してるだけで reyn.log 肥大化してシステム止まらないようにしてね"). `_setup_interactive_logging` now installs a `logging.handlers.RotatingFileHandler` instead of a bare `logging.basicConfig(filename=...)`, sized from a new `logs:` config block (`max_bytes`, `backup_count`; default 16 MiB × 4 backups, total ≤ 80 MiB — architect's own provisional-but-reasoned numbers, config-tunable). The redirect still installs immediately with `LogsConfig()`'s own dataclass defaults (so config-load-time warnings are still captured), then `_apply_logs_config` refines the already-installed handler's `.maxBytes`/`.backupCount` in place once the real config is known.

`logs:` is a real `ReynConfig` dataclass field (`LogsConfig`), so it gets `config_schema`'s automatic unknown-key detection for free — no manual #5849③-style registry needed, unlike `permissions:`/`sandbox:`.

**Rides along** (architect-requested, alongside the concurrently-landed #5877): `stall_trace.find_file_handler_path()` no longer scans root-logger handlers for a path-shape match — `_setup_interactive_logging` now registers its handler's path directly via a new `stall_trace.register_file_handler_path()`, the one authoritative writer declaring its own path instead of a reader re-deriving it. Existing `isinstance(handler, logging.FileHandler)`-based readers (`litellm_bootstrap`) are unaffected — `RotatingFileHandler` is a `FileHandler` subclass, and its `.baseFilename` stays fixed across rotations by design.

The issue's own "same PR" ask re: #5870 stage 1's per-episode dump cadence needed no code change — `LoopTripwire.observe`'s dump is already once-per-episode (verified in `loop_probe.py`), as the issue itself noted.

[...]

PR #5938, opened 2026-09-07
fix(#4986/#5909 tool ticket): arm the stall-trace watchdog per xdist worker too

tya5:
**[e2e-coder]** — part of #4986, part of #5909. This is the follow-up PR lead-coder authorized ("残りのツール項目は後続のPRで構いません") for the remaining #4986/#5909 tool-ticket items.

## What was wrong

`stall_dump.py` armed the teardown-hang watchdog on the **CONTROLLER only**. Its own docstring's reasoning: "the observed hang is in the controller's own event loop... arming per-worker would multiply the timer for no additional signal." **That reasoning was falsified by a real, same-day incident** (architect's finding during the #5909 investigation): a controller stuck waiting on a silent WORKER is idle in `queue.get()` — nothing of the controller's OWN is stuck, so its dump can never show anything about a hang that lives strictly inside a worker's own session teardown. "The controller's dump is enough" was the premise the old decision rested on, and it doesn't hold for that failure class.

lead-coder confirmed the same finding independently and explicitly ordered the docstring fixed in the same PR as the code (CLAUDE.md hard rule: a describing comment is stale the moment the code it describes changes) — not the code alone.

## What this does

1. **Each xdist WORKER now arms its own watchdog**, into its own file (`.reyn-ci-stall-trace.<workerid>.log`, keyed by the same `workerid` `reyn.dev.testing.worker_forensics` already uses) — never a path shared across worker processes (faulthandler holds the fd it was armed with for the life of the arm, #5879 measured directly, so a shared path would be silent loss for whichever process's fd isn't the one still live).
2. **The worker's own threshold is BELOW the controller's** — `worker_seconds_from(controller_seconds)` = controller's value minus a new `WORKER_OFFSET_SECONDS` (60s), floored at 1s. Pulled into a pure function so the causal-ordering property (worker fires first) is an exact unit check, not a real-time race.
3. **`stall_dump.py`'s module docstring updated in this PR** — the old "WHY NOT ARMED IN EVERY xdist WORKER" section is replaced with "WHY ALSO ARMED PER xdist WORKER (changed decision)", recording that the prior reasoning was falsified and why.
4. **`.github/workflows/test.yml`'s "Stall-trace dump" step** now also prints every worker's own file that actually fired, not just the controller's.
5. **Architect's tool ②** — a backgrounded OS-level watcher in the same run step, firing once at 620s (after the controller's own 600s dump, before the 720s `timeout 12m` kill): `ps -ef`, `pgrep -a python`, and each pytest-controller PID's own `/proc/<pid>/fd` table (Linux runner) — because "is the controller process even still alive" is an OS-level fact no Python-side dump can answer about itself. Zero cost on a healthy run (~7-8 min, well under 620s) — the `sleep 620` never returns before the step's own foreground command exits, so nothing is ever written and the backgrounded subshell is reaped with the step.

## Tests

2 new in `tests/dev/test_stall_dump_4986.py` (Tier 1):
- A real `-n 1` xdist worker whose session teardown hangs (a FIFO `open()`/`read()` block, not `sys.stdin` — measured directly while building this: execnet remaps a worker's own stdin, so the sibling controller-hang tests' own `readline()` idiom does not reach into a worker) produces a real stack dump in its own per-worker file.
- A fast unit check on `worker_seconds_from`'s exact offset/floor arithmetic (the causal-ordering property, checked exactly rather than raced against real subprocess timing).

Plus the 4 pre-existing tests in that file, and the full related regression scope — `test_5898_loop_tripwire.py`, `test_5850_full_suite_collection_guard.py`, `test_5909_worker_forensics.py`, `test_3671_stall_trace_startup_wiring.py` — **30 tests total, all green**.

**Strip-falsification** done and executed (not reasoned): reverting `pytest_configure`'s worker branch to the original early `return` leaves the new worker-dump test unable to ever find its file — it hangs at the outer wait's own unbounded poll, killed only by an external timeout. Verified directly, restored before commit.

## Local gates

`ruff check`, `test_tier_audit.py --strict`, `verify_module_docstrings.py`, `mypy_ratchet.py` (0 new findings), `flat_tests_ratchet.py`, `check_bare_tests_import_reference.py`, `check_file_depth_reference.py`, `check_subprocess_reyn_pin.py`, `check_tests_path_literal_reference.py` (re-run against fresh `origin/main` right before push) — all green. YAML syntax of `.github/workflows/test.yml` verified with `yaml.safe_load`.

## Out of scope

`worker_forensics.py`'s own trace-surfacing (part of #5909) already landed separately: #5934.

## Test plan

- [x] `pytest` scoped to diff + related regression scope — 30 passed, 0 regressions
- [x] `ruff check` on changed files
- [x] `test_tier_audit.py --strict` on the changed test file
- [x] `verify_module_docstrings.py` on the changed src file
- [x] `mypy_ratchet.py`
- [x] `tests/`-path-conditional gates (file touched under `tests/`)
- [x] YAML syntax check on `.github/workflows/test.yml`
- [x] (skipped — standing Visual-item waiver, owner 2026-08-08)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,8 +155,40 @@ jobs:
         # cancel their own watchdog the instant the last item's protocol
         # returns (`_pytest/faulthandler.py`'s own source, confirmed, not
         # guessed). See `reyn/dev/testing/stall_dump.py`'s own docstring.
+        # `( sleep 620; ... ) &` (part of #4986/#5909 tool ticket, architect's
+        # tool ②): a stall dump can only show what its OWN process's threads
+        # are doing — whether the pytest CONTROLLER process is even still
+        # ALIVE at all is an OS-level fact no Python-side dump can answer
+        # from inside itself. This backgrounded watcher asks the OS directly,
+        # once, 20s after `REYN_STALL_TRACE_CI=600`'s own dump would have
+        # fired and comfortably before the `timeout 12m` (720s) kill below —
+        # so a genuinely still-running controller has already had its stack
+        # dumped by the time this fires, and the two diagnostics read in the
+        # same causal order the stall-trace file and the per-worker files
+        # already do. On a healthy run (~7-8 min, well under 620s) the
+        # `sleep 620` never returns before this step's own foreground command
+        # exits, so nothing is ever written to `/tmp/hang-ps.log` and the
+        # backgrounded subshell is reaped with the step — no added cost on
+        # the common path. `pgrep -f "pytest -q -n auto"` re-derives the
+        # controller's PID from its own command line rather than threading a
+        # PID through — this stays a single independent watcher, not another
+        # moving part the main run has to hand anything to.
         run: |
           set -o pipefail
+          (
+            sleep 620
+            {
+              echo "=== ps -ef @620s ==="
+              ps -ef
+              echo "=== pgrep -a python @620s ==="
+              pgrep -a python
+              echo "=== controller fd table @620s (Linux only) ==="
+              for pid in $(pgrep -f "pytest -q -n auto"); do
+                echo "--- pid=$pid ---"
+                ls -l "/proc/$pid/fd" 2>/dev/null || echo "(no /proc/$pid/fd for pid=$pid)"
+              done
+            } > /tmp/hang-ps.log 2>&1
+          ) &
           REYN_REPLAY_UNCONSUMED_CHECK=1 REYN_STALL_TRACE_CI=600 timeout 12m python -m pytest -q -n auto --timeout=120 -rs --durations=25 | tee /tmp/pytest-output.log
 
       # #4986: surface the stall dump above if (and only if) it fired. The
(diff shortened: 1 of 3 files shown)


tya5:
TESTS-READ (head 177c53d13df9e6bdaa83273b68b5b1a0dca118a1)

**[lead-coder]** — head は `git ls-remote` で引き、報告値と一致しています。

## 🔴 私が確かめられなかったこと（先に書きます）

**`tests/dev/test_stall_dump_4986.py` を私の手元で走らせられませんでした。10 分を超えて終わりません。**

**ただしこれは本 PR のせいではありません。** `origin/main` の同 file を同じ環境で走らせると、**同じくハングします**（180s の timeout で 2 件が F、timeout dump に `memory_ceiling.py` の watcher thread が出ます）。∴ **環境側の既存の問題**で、本 PR が持ち込んだものではありません。

**判定は CI に委ねます。** CI では **13903 passed**、赤は 1 件だけで、それは **#5918 の既知の test**（`test_chat_acompletion_wire_count_unaffected_by_embed_global_mutation`、差分と無関係、本日 4 件目）です。台帳に記録し、落ちた leg のみ再実行しました。

## 私が読んで確かめたこと

**① docstring の「理由失効」が記録されている**（私が発注時に求めたもの）

```
-WHY NOT ARMED IN EVERY xdist WORKER
-    … Arming per-worker would multiply the timer for no additional signal …
+WHY ALSO ARMED PER xdist WORKER (changed decision — #4986/#5909 tool ticket)
+    … **That reason was falsified by a real incident** …
+    a controller stuck in queue.get() waiting on a silent WORKER is,
+    by construction, invisible to a dump that only ever …
```

**古い理由を消さずに「偽にされた」と書いてある**のが正しい形です。次に読む人は、**なぜその判断が変わったか**を追えます。

**② hang の作り方が決定論的**

```python
# Blocks until the outer test closes this process's stdin — a real,
# deterministic wait, released deterministically, never a sleep.
sys.stdin.readline()
```

worker 側は **FIFO** を使っています。理由も書かれています ── 「execnet が worker の `sys.stdin` を差し替えるので、controller 用の `readline()` の書き方は worker には届かない」。**duration（`sleep`/`timeout` marker/`attempts=N`）は 1 つも入っていません。**

**③ `worker_seconds_from` の contract test**

```python
assert worker_seconds_from(600.0) == 600.0 - WORKER_OFFSET_SECONDS
assert worker_seconds_from(600.0) < 600.0
assert worker_seconds_from(1.0) == 1.0
assert worker_seconds_from(WORKER_OFFSET_SECONDS) == 1.0
```

**閾値を入力として与えており、時計を待っていません。** floor と offset の境界も両方当てています。

## 6 問

1. **Tier** — Tier 1（Contract、`worker_seconds_from`）＋ Tier 1（実 pytest session を回す dump の契約）。
2. **実装の転写か** — 違います。`600.0 - WORKER_OFFSET_SECONDS` は定数と入力から書いた期待値で、実装の式の写しではありません。
3. **無くなって困るのは誰か** — CI を追う私たちです。**worker の中で止まった時、controller だけの dump は構造的に見えません。**
4. **走らずに緑になり得るか** — **私は手元で確認できていません**（上記）。**CI が 13903 passed で走らせています。** ⚠️ これは「私が strip を実施した」ではありません。**そこは正直に書きます。**
5. **何を溜めるか** — 内側 session は subprocess で、`pytester` が片付けます。
6. **宣言 Tier は真か** — 真です。

## author の報告のうち、私が再現していないもの

「real xdist(-n 1) で FIFO 経由の実 hang → worker 専用ファイルに実 dump 確認」「strip 実行済み（除去で test hang→timeout、復元確認）」は、**私は再現していません**（上記の理由）。**CI の緑で代替します。**

BLOCKING はありません。auto-merge を武装します。

PR #5970, opened 2026-09-08
fix(#5891 (c)): tool_returned carries tool_call_id -- the real join key call_id cannot be

tya5:
**[tui-coder]** — #5891 (c): architect's corrected ruling. `tool_returned` now carries `tool_call_id` — the actual join key `call_id` cannot be.

## Why (architect's own correction)

The original #5891 design ("a reader can JOIN `tool_returned` to its history row on data both sides already have, no new field needed") rested on a false premise, caught before implementation: `call_id` identifies a whole litellm **ROUND** (shared by every `tool_calls` entry in it); a history row's own `tool_call_id` (`tc["id"]`) is **per-call**. The same tool called twice in one round produces two history rows a `call_id`-only join cannot disambiguate.

## What

- `DispatchContext.tool_call_id: str | None = None` — the same `tc["id"]` `RouterLoop.feedback` already writes onto its own history row.
- `dispatch_tool`'s `tool_returned` emit always carries `tool_call_id` (`None` for a caller with no real litellm `tool_calls` entry — CodeAct `tool()`, `/exec`/`/tasks` slash, a pipeline `tool:` step). `tool_call_id_absent_reason` (naming `caller_kind`) is added ONLY in that `None` case — never silently omitted, so "the caller doesn't have one" never reads as "it was forgotten" (same discipline `content_ref_unavailable` already uses one layer down, #5936).
- Wired through both real `router_loop.py` dispatch call sites plus the documented `_execute_tool` test seam.
- `docs/reference/runtime/events.md`'s `tool_returned` row updated for this AND #5936's own previously-undocumented fields.

No new audit-event kind (untouched `event_schema.py`), no new storage (no new `save_tool_result` call site).

## Tests

- `tests/core/test_5891_tool_call_id_audit_field.py` — dispatcher-level: presence, distinctness across 2 calls in 1 round, absence+reason for both `router` and `operator` callers.
- `tests/runtime/test_5891_tool_call_id_history_join.py` — end to end: 2 real `dispatch_tool` calls + `RouterLoop.feedback`/`persist_feedback`, real `Session`/`EventLog` (`collect_events`), proving each `tool_returned.tool_call_id` names exactly one history row with a real `content_ref`; plus a dedicated test driving the ACTUAL `_execute_tool` → `_dispatch_resolved` wiring, not just a hand-built `DispatchContext`.

[...]

PR #5960, opened 2026-09-07
fix(#5949 stage 1-b + stage 2, P0): invert _parse_history_line's default -- hydrate is required, not silently on

tya5:
**[e2e-coder]** — part of #5949.

## What broke, and why it's structural, not a 2nd flag

Stage ① (#5955, merged) fixed `Session._load_older_entries`'s own parse loop, but attach actually drives a SEPARATE, un-fixed one: `Session.extend_history_backward_async` (`endpoint.py`'s backlog send is what calls it — never the sync `extend_history_backward`). Owner's real measurement got WORSE after stage ① landed (`reyn:web` peak 11 GB, `reyn:chat` peak 14 GB) — not better — because the path that actually runs was never touched.

Architect's ruling (issuecomment-5576144700, owner: "考慮漏れが起きないよう構造化してよね"): don't patch the second call site with the same flag — **flip the default**. `preview_only: bool = False` meant "materialize by default," so a new/missed caller silently landed on the expensive path. `hydrate` now has **no default at all** — every caller must decide, and Python raises `TypeError` on a forgotten one. Only the ONE caller that reads `.content` directly for something other than the LLM wire (compaction's `_durable_active_history_after`, whose own `_measure_and_select` computes real token estimates before any wire is built) passes `hydrate=True`. Every other caller — live append, startup, and both backward-paging primitives — stays cheap; `RouterHistoryBuffer._serialise_turn`'s own unconditional `resolve_history_content` call resolves the real body lazily at wire-build time regardless of parse-time hydration.

## Full caller enumeration (verified via `git grep`, not `| head`)

| site | choice | why |
|---|---|---|
| `session.py` `_durable_active_history_after` (compaction input) | `hydrate=True` | `_measure_and_select` reads `.content` directly for token estimates, before any wire |
| `session.py` `_append_parsed_history_line` (live append + startup) | `hydrate=False` | wire lazily resolves regardless; verified empirically (see below) |
| `session.py` `_read_and_parse_older_entries` (shared by both backward-paging primitives) | `hydrate=False` | display-only, preview filled separately |
| `tests/interfaces/test_4995_threaded_transport_proxy.py` (2 call sites) | `hydrate=False` | inert (plain `role="user"` rows, no content_ref), but must still decide |

## Structural fix ②: two implementations → one

[...]

PR #5975, opened 2026-09-08
feat(#5973 裁定4): separate ResidentBytes from BodyBytes as distinct types

tya5:
**[tui-coder]** — #5973 裁定4: precursor PR separating `ResidentBytes` from `BodyBytes` as distinct types. **This PR moves zero bounds** — architect's own explicit constraint.

## Why this goes first (architect's ruling)

#5973's root cause: #5896's migration split ONE quantity ("a row's own byte count") into TWO — ⑴ how many bytes a resident row occupies, ⑵ how many bytes the body its `content_ref` points at is. Every existing bound kept counting the same bare `int`, blind to which currency it measured.

①②③ (the actual fix, e2e-coder) are **not** split across two people — they count the SAME resource in different places, and splitting the WORK would let two definitions drift apart, reproducing the exact class #5973 exists to close. The type separation goes first instead: once `body_bytes`/`resident_bytes` are distinct types, ①②③'s author cannot pass the wrong one without mypy refusing to compile it.

## What

- `ResidentBytes`/`BodyBytes`: `typing.NewType(..., int)` in `chat_message.py` (the file owning both concepts).
- `ChatMessage.resident_bytes() -> ResidentBytes` (was `-> int`).
- `HistoryResidentConfig.max_bytes: ResidentBytes` — **the 256 MiB default value is unchanged**, only its type.
- `router_loop.py`'s tool-result meta write: body byte count computed as an explicit `BodyBytes` local before storing (the dict itself stays `dict[str, Any]`).
- 4 existing tests constructing `HistoryResidentConfig(max_bytes=...)` with a bare int now wrap it in `ResidentBytes(...)`.

## The required witness (architect's own acceptance: passing `body_bytes` where `resident_bytes` is expected must be red)

`NewType` has zero runtime footprint, so the guarantee is only observable via a real mypy invocation. `tests/runtime/test_5973_resident_body_bytes_types.py` writes a probe module to `tmp_path` (a deliberately-wrong `BodyBytes`-where-`ResidentBytes`-expected call + a correct `ResidentBytes` call), runs the same mypy this repo's own `mypy_ratchet.py` uses, and asserts the wrong call is flagged (`[arg-type]`, exact line) while the correct one is not.

## Strip-verification (in-file Edit → Edit-back, no `git stash`/`checkout`)

Changed `ResidentBytes = NewType(...)` to a plain `ResidentBytes = int` alias → the probe's wrong call stops being flagged (mypy exits 0) → the witness test correctly goes red. Restored; `git diff --stat` confirmed clean.

## Gates (all green)

ruff, `test_tier_audit.py --strict`, `verify_module_docstrings.py`, `mypy_ratchet.py` (both changed-files **and** `--full` — confirmed clean under both, since a mismatch could otherwise be grandfathered under an existing same-file same-code baseline entry), `flat_tests_ratchet.py`, `check_tests_path_literal_reference.py`, `check_bare_tests_import_reference.py`, `check_file_depth_reference.py`, `check_subprocess_reyn_pin.py`, `check_no_core_dep_importorskip.py`, `suspected_time_dependence_ratchet.py`. Scoped pytest: 23 passed.

## Test plan

- [x] Scoped pytest green (23 passed)
- [x] All pre-PR + path-conditional gates green (listed above)

part of #5973

https://claude.ai/code/session_01LzYAXF8WFTxR2JPMZSoDfn

### DIFF
--- a/src/reyn/config/chat.py
+++ b/src/reyn/config/chat.py
@@ -12,6 +12,13 @@ from reyn.config_axis import Axis
 # that doesn't belong in the config-only module.
 from reyn.runtime.budget.budget import CostConfig, CostLimitConfig  # noqa: E402
 
+# #5973 裁定4: HistoryResidentConfig.max_bytes is the SAME currency
+# ChatMessage.resident_bytes() measures (both count a resident row's own
+# serialized size) -- importing the NewType from its one owning module
+# rather than redeclaring it here keeps that a fact the type system
+# states, not a naming convention two files have to keep in sync by hand.
+from reyn.runtime.chat_message import ResidentBytes  # noqa: E402
+
 
 @dataclass
 class LoopConfig:
@@ -549,8 +556,16 @@ class HistoryResidentConfig:
     costs nothing but memory an operator can already afford, while a too-small
     one would make ordinary scrollback/rewind pay for reloads more often than
     necessary.
+    #5973 裁定4: ``ResidentBytes``, not a bare ``int`` — this cap and
+    ``ChatMessage.resident_bytes()`` (the value it is compared against,
+    ``session.py``'s own eviction loop) must stay the SAME currency by
+    construction; see that NewType's own module docstring in
+    ``runtime/chat_message.py`` for why the two counts drifted apart in
+    the first place.
     """
-    max_bytes: int = field(default=256 * 1024 * 1024, metadata={"axis": Axis.BOUNDING})  # 256 MiB
+    max_bytes: ResidentBytes = field(
+        default=ResidentBytes(256 * 1024 * 1024), metadata={"axis": Axis.BOUNDING},
+    )  # 256 MiB
 
 
 def _build_history_resident_config(raw: object) -> "HistoryResidentConfig":
@@ -570,7 +585,7 @@ def _build_history_resident_config(raw: object) -> "HistoryResidentConfig":
             max_bytes = defaults.max_bytes
     except (TypeError, ValueError):
         max_bytes = defaults.max_bytes
-    return HistoryResidentConfig(max_bytes=max_bytes)
+    return HistoryResidentConfig(max_bytes=ResidentBytes(max_bytes))
 
 
 @dataclass
--- a/src/reyn/runtime/chat_message.py
+++ b/src/reyn/runtime/chat_message.py
@@ -25,7 +25,36 @@ import json
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timezone
 from enum import StrEnum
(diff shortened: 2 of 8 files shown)


tya5:
TESTS-READ (head 6dda32b1a34ddbb0176d5ffa3b5f276460fb8dbc)

**[lead-coder]** — 差分のテストを読み、strip を回しました。tree は `git ls-remote` で照合済。

## ⚠️ この PR の唯一の成果は「取り違えが書けない」ことです

∴ **witness が「型検査が落ちる」ことでしか成立しません** — **挙動は 1 つも変わらない**ので、**普通の assert では何も言えません。**

⭐ **`mypy` を実 subprocess で走らせる test** になっています（`1 passed in 6.97s` — この 7 秒が mypy です）。

## 私が回した strip

**両方の `NewType` を plain alias（`= int`）に戻す**:
```
FAILED test_resident_bytes_and_body_bytes_are_mypy_incompatible
1 failed in 7.12s
```

∴ **「型が分かれている」が実際に噛んでいます。** ⚠️ **これは空回りしやすい形**（mypy が黙れば緑）ですが、**戻すと赤になる**ことを私が確認したので、**空回りしていません。**

復元後 **5 passed**（型 test ＋ `test_5896_p0_resident_bytes_cache.py`）、tree は `git status --porcelain` 空。

## 🔴 私が最も気にしていた点 — 上限が動いていないこと

```diff
-    max_bytes: int = field(default=256 * 1024 * 1024, ...)  # 256 MiB
+    max_bytes: ResidentBytes = field(
+        default=ResidentBytes(256 * 1024 * 1024), ...
+    )  # 256 MiB
```

**値は 256 MiB のまま、型が付いただけ**です。⚠️ **私は「あなたの PR で bound が 1 つでも動いたら BLOCKING を出す」と書きました** — **動いていません。**

## 六問

Tier 2。⚠️ **質問 4 が要点**でした — **「mypy が落ちる」を見る test は、mypy が走らなければ緑**です。**実 subprocess で走らせ、かつ「正しい方は緑」も見ている**（両方向）ので、**#5919 で学んだ「判別子は両方向を問う」がそのまま入っています。**

## ⚠️ 記録

**これは今夜 3 度目の「そもそも書けなくする」**です — **#5960 の `hydrate`（既定を無くす）／#5970 の `tool_call_id`（必須 keyword-only）／本件（型で分ける）**。⭐ **3 回とも別の人が別の場面で選んでいます。**

**e2e-coder の①②③がこの型の上に乗ります。** CI 緑を確認して私が merge し、合図します。

PR #6100, opened 2026-09-10
feat(part of #6083 stage 2b): slash_command accepts immediately, runs in background

tya5:
**[tui-coder]** — `part of #6083`

## What this closes

Stage 2b (⑵-b). #6094 fixed the false-CLAIM half of #6083 (stage 1: a
session-locus command's error line no longer asserts "no session" for a
control read-timeout — the owner's own `/compact` case). This PR closes
the actual WAIT: the AG-UI `slash_command` arm no longer awaits a
handler to completion before answering. It hands the command to the
session's own background-task funnel (`Session._background_tasks`,
#4759's `TrackedTaskSet`) and answers `accepted` the instant it is
scheduled — not once it has run. A slow handler (the reported case,
`/compact`) can no longer hold the control POST open past the client's
own read timeout at all.

No new signal channel was needed: `Session._slash_context()` already
wires `SessionBoundTransport(display_sink=self._put_outbox_nowait)`, so
a background-run command's own success/error display already flows
through the session's outbox → the SAME SSE broadcast every other
reply rides. Confirmed by reading the code, not assumed.

## Condition ⑴ — spawn() failure

If `session._background_tasks.spawn(...)` itself fails BEFORE the task
starts, the response answers `accepted: false`, not `true` — the
one shape that must never happen ("told it ran, nothing runs").
Additionally, the AG-UI arm first resolves the name against `REGISTRY`
before deciding to spawn at all — an unknown name (a client on a
different build) answers `accepted: false` synchronously rather than
being silently absorbed into the background task's own
`cmd is None -> return False` path, which the client would never see.

Once scheduled, `_run_slash_command_in_background` wraps
`execute_slash_command` in ITS OWN try/except as a second line of
defense: `execute_slash_command` already catches and displays its own
handler's exception (dispatch.py), but this outer layer catches
anything that could still escape it (`SlashContext`/
`_ErrorWatchingTransport` construction) — a background task's own
uncaught exception would otherwise become a silently-logged, never-
retrieved `asyncio` task exception the client never learns about. It
is caught, logged, and surfaced as a display-stream error line instead.

Covered by `test_a_scheduling_failure_does_not_still_answer_accepted`
(strip-falsified: fails with `accepted: true` on the pre-fix tree).

## Condition ⑵ — what survives a mid-command crash

`appends_wal=True` is passed to `spawn()` (a slash handler can itself
append to the WAL, e.g. `/compact` recording a compaction) — this is
what makes `await_quiescent`'s rewind reset-record drain it correctly,
same reasoning as every other WAL-appending producer through this
funnel.

**If reyn crashes while this background task is running: the command
itself is LOST.** The scheduling (the `asyncio.Task` this PR spawns) is
tracked only in-memory by `TrackedTaskSet` — nothing durable records
"a slash command was accepted and is in flight." If the handler had not
yet reached its own append point at the moment of the crash, nothing
resumes it and nothing retries it; the operator sees no completion and
no failure, only silence, the same as any other in-flight fire-and-
forget task that never got there. If the handler HAD already durably
appended before the crash (e.g. `/compact`'s own WAL record), that
record survives exactly as it always did — this PR changes nothing
about an individual handler's OWN crash-recovery behavior, only WHEN
the client is told "accepted" versus "completed". "Lost" is the honest
answer for the scheduling itself; this PR does not add a retry or
resume mechanism for it.

## Condition ⑶ — this is a UX change, stated for the owner

**Before:** typing `/compact` (or any session-locus command) waited —
the terminal showed nothing new until the command finished or the
client's own control-read timeout (10s) elapsed, at which point a slow
but still-succeeding command could show a FALSE failure line (#6083's
own reported symptom).

**After:** the command is accepted instantly — there is no wait state
between pressing Enter and seeing SOME response. What changes concretely:
there is no longer a multi-second span where the terminal shows nothing
after typing a slow command; the real completion/failure (e.g.
`/compact`'s own "compacted N turns" or an error) now arrives
separately, slightly later, over the same display stream every other
reply already uses — the operator sees an acceptance essentially
instantly, then the real outcome shortly after, rather than one single
wait-then-result. This is judged an improvement (no more false failures
on a slow command), but the owner should see this stated plainly since
"what you see right after pressing Enter" is a visual/UX judgment call
this session cannot make on the owner's behalf (no TTY access).

## Scope note (⑵-a's server-only-decision half)

This PR does NOT introduce a registry of "which payload types get
202-style treatment" — only `slash_command` changes, per the explicit
ruling not to build a generalized mechanism ahead of a second need.

## Test plan

- [x] `ruff check .`
- [x] `test_tier_audit.py --strict` (changed test files)
- [x] `verify_module_docstrings.py` (changed src files)
- [x] `mypy_ratchet.py`
- [x] `flat_tests_ratchet.py`
- [x] tests/-touching gates (bare-tests-import, file-depth, subprocess-reyn-pin, no-core-dep-importorskip, suspected-time-dependence)
- [x] `src/reyn/interfaces/` touch: `silent_except_ratchet.py`
- [x] docs/ touch: `mkdocs build --strict`, `check_doc_anchors.py`, `check_retired_config_keys_denylist.py`
- [x] Diff-scoped pytest green (17 tests: the 2 new + the modified test_3595_s5 test + the existing #6083 stage-1 tests), strip-falsified both new tests

https://claude.ai/code/session_01LzYAXF8WFTxR2JPMZSoDfn

### DIFF
--- a/docs/reference/runtime/agui-transport.md
+++ b/docs/reference/runtime/agui-transport.md
@@ -40,17 +40,23 @@ A2A; tools are MCP; observability export is OTEL. Those are separate surfaces.)
     at-most-once retry).
   - `{"type": "slash_command", "name": "model", "args": "strong"}` — run a
     registered slash command (#3595 S5). The response body is
-    `{"status": "ok", "ran": true|false}`; `ran: false` means this server's
-    registry has no such command (a client on a different build), never a
-    crash. ★ The client has ALREADY interpreted the operator's `/…` line and
-    resolved the NAME against its own registry — nothing on the wire or on the
-    server tests a leading `/`, which is the whole point of #3595: `Session`
-    interprets no string, and a client maps typed text onto published
-    operations. A remote client sends this rather than running the command
-    itself because it holds no `Session`, and the commands that still read
-    session state can only run where that session is. Permission-gated by the
-    same `authorize_write` check `user_message` passes, and the command's reply
-    rides the ordinary display stream.
+    `{"status": "ok", "accepted": true|false}`; `accepted: false` means this
+    server's registry has no such command (a client on a different build),
+    never a crash. ★ The client has ALREADY interpreted the operator's `/…`
+    line and resolved the NAME against its own registry — nothing on the wire
+    or on the server tests a leading `/`, which is the whole point of #3595:
+    `Session` interprets no string, and a client maps typed text onto
+    published operations. A remote client sends this rather than running the
+    command itself because it holds no `Session`, and the commands that still
+    read session state can only run where that session is. Permission-gated by
+    the same `authorize_write` check `user_message` passes.
+    #6083 ⑵-b: `accepted: true` means the command was HANDED OFF to the
+    session's own background-task funnel, not that it has finished — a slow
+    handler (e.g. `/compact`) no longer holds this POST open until it
+    completes (the false-failure shape #6083 itself reports). The command's
+    real success/failure line still rides the ordinary display stream, exactly
+    as before; a caller that needs the actual outcome reads it from there, not
+    from this response.
   - `{"type": "heartbeat"}` — a liveness keepalive.
 
   An input type the server does not model is a **graceful no-op** (a `200` ack),
--- a/src/reyn/interfaces/transport/agui/client.py
+++ b/src/reyn/interfaces/transport/agui/client.py
@@ -694,11 +694,25 @@ class AgUiTransport(ClientTransport):
         # no transport re-tests ``startswith("/")`` and the server executes a
         # named operation rather than sniffing a string. The server re-resolves
         # the name against its OWN registry (a client and a server can be
-        # running different builds) and answers whether it ran.
+        # running different builds).
+        #
+        # #6083 ⑵-b: the server no longer awaits the command to completion
+        # before answering — it hands the command to its own background-task
+        # funnel and answers ``accepted`` immediately (endpoint.py's own
+        # comment on the ``slash_command`` arm has the full reasoning). So
+        # ``True`` here means "the server accepted this and will run it",
+        # NOT "it already ran" / "it succeeded" — a meaning narrower than
+        # this method's own ABC docstring states for the general case
+        # (``ClientTransport.run_slash_command``), which still holds for
+        # ``InProcessTransport``'s synchronous, local execution. A caller
+        # that needs the real outcome reads it off the display stream this
+        # command's own success/error line already rides (the SAME SSE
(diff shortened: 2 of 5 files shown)


tya5:
TESTS-READ (head d01cf3548aa7a161a3a66b38e8836d63f5ec296c)

**[lead-coder]** — **diff の test を読みました**（新規 2 本 ＋ 既存 1 file の追随）。

## 六問（2 本）
1. **Tier 2**（reyn 自身の control 応答の contract）。**落ちたら誰のバグか**: **受理を返す側。**／2. **否**／3. 🔴 **困るのは owner** — **実際に踏んだ「成功したのに 10 秒で偽の失敗」**／4. **空の緑なし**（下記）／5. **蓄積しません**／6. **一致。**

## ⭐ 2 本が、主張の両側を覆っています
- **`test_accept_returns_while_the_handler_is_still_running`** — 🔴 **「handler がまだ走っている間に受理が返る」。** ⭐ **これが ⑵-b の主張そのものです。**
- **`test_a_scheduling_failure_does_not_still_answer_accepted`** — 🔴 **私が条件 ⑴ に書いた「受理したと言って何も走らない」を、正面から閉じています。** ⚠️ **これが無ければ「常に `accepted: true`」でも 1 本目は緑です。**

## ⭐ 条件 3 つ、すべて満たしています
- **⑴ spawn 失敗** → **`accepted: false`**（**未知名も *事前に* resolve、strip-falsify 済**）。⭐ **さらに `f"/{name} failed unexpectedly after being accepted"` を display に出す経路も在ります** — 🔴 **「受理と言った後の silent drop」を作っていません。**
- **⑵ crash 時** → **「scheduling 自体が lost」と PR 本文に明記。** ⚠️ **答えが「失われる」でも構わない、と私は言いました。⭐ *書いてある* ことが条件でした。**
- **⑶ UX 変化** → **1 段落。**

## ⚠️ 開示 1 点（block しません）
🔴 **wire の field 名が `ran` → `accepted` に変わりました。**
⭐ **doc も同 PR で追随しています**（`docs/reference/runtime/agui-transport.md`）。**「a client on a different build」の場合も明記。**
⚠️ **ただし *逆向き*（新しい client × 古い server）は書かれていません** — **`result.get("accepted")` が `None` → `False` → 「no confirmation received」。** 🔴 **silent ではなく、見える形で劣化します** ∴ **block しません。**
⭐ **同一 build で動かす前提が崩れたとき、この 1 行が要ります。** **次に `--connect` 周りを触る人へ。**

**∴ 六問で block に当たるものは在りません。**

PR #6201, opened 2026-09-16
fix(#6184 段4-A): dispatched_tool_calls meta is a declared-child count, not a bool

tya5:
**[e2e-coder]** — part of #6184, 段4-A.

## What

`router_loop.py`'s 4 `"dispatched_tool_calls"` meta write-sites go from
`bool(result.tool_calls)` to a genuine declared-child COUNT:

- 2 sites inside `if isinstance(interp, Execute):` (tool-turn-text row,
  agent_spawn_ack row — the same call, same `interp`, #4691 Phase 1):
  `len(interp.actions)`.
- 2 sites NOT inside that branch (empty-response terminal path, ordinary
  terminal-reply path — no tool call dispatched by construction): kept as
  `len(result.tool_calls)`, which is always `0` there — matching the
  bool's own prior `False`.

`interp.actions` is genuinely **post-dedupe**: `SchemeOps.resolve()`
builds it via `self._dedupe_tool_calls_round(llm_response.tool_calls)`
*before* constructing the `Execute` interpretation these 2 call sites
read from — never `len(result.tool_calls)` (pre-dedupe), which would
report an inflated count for a weak model's duplicate tool_calls (the
F5 real-incident shape `_dedupe_tool_calls_round` exists for — "333k
tokens / 51 LLM calls", batch 1).

## Consumer safety (dispatch requirement)

`app.py`'s `_register_call_parent` — confirmed via `git grep -n
"dispatched_tool_calls" src/reyn/interfaces/` to be the **ONLY** real
code reader in `src/reyn/interfaces/` — reads it with
`if meta.get("dispatched_tool_calls"):`, truthiness only. `0` and
absent are both falsy in Python, so this reader needs no change; a
docstring paragraph on that method now documents this finding for the
next reader who needs to tell `0` and absent apart (truthiness alone
can't).

`presenter.py:1082`'s own comment called the field a "bool" — stale
the moment 段4-A landed; fixed in the same PR (CLAUDE.md: a doc
describing a mechanism is stale the moment the mechanism's code
changes).

## New accept-criteria tests (`tests/runtime/test_router_loop.py`)

- `test_dispatched_tool_calls_meta_counts_declared_children_6184_stage4a`
  — 2 **distinct** sync tool_calls (`list_memory`, different `path`
  args — never deduped, sync tools are excluded from
  `_dedupe_tool_calls_round`) → `2`. Strip-falsified by hand: reverted
  to `bool(result.tool_calls)`, confirmed RED at the exact expected
[...]

PR #6203, opened 2026-09-16
fix(#6184 段4-B): a Group folds by default only when the declared count is >= 2

tya5:
**[e2e-coder]** — part of #6184, 段4-B.

⚠️ This stage has no explicit owner approval yet — it acts on the owner's
own reported symptom ("a Group with only 1 item in it"), not a stated
go-ahead naming this stage. Do not read this as "fixed" until the owner
sees it live.

## What

`_register_call_parent`'s `entry.collapse()` call (#4691 item 3's own "a
completion Group defaults COLLAPSED") now fires only when
`meta["dispatched_tool_calls"]` (an int since #6201/段4-A) is `>= 2` — not
unconditionally whenever truthy. A round declaring exactly 1 tool call
still registers and still spins the RUNNING spinner (unchanged) but no
longer starts folded.

Decision point confirmed via `git grep -n "dispatched_tool_calls" origin/main -- src/`
before touching anything, per the dispatch's own requirement.

## Rule — by declaration, never measurement

`entry.children` is always empty at registration time (children haven't
arrived yet by construction), so a `len(entry.children)` check would
never fold anything. The docstring records this explicitly.

## Accept criteria (issue #6184 thread, corrected twice mid-implementation)

1. declared == 1 → does not fold (today's defect)
2. declared >= 2 → still folds (deny-side: without this, "always don't
   fold" also passes)
3. declared >= 2 with 0 actual children at registration → still folds
   (proves the decision reads the DECLARED count) — covered by the
   existing registration-time assertion in
   `test_a_group_parent_defaults_collapsed`

**Original accept④** ("a message with no declared-count key still folds,
as before") **was retracted by lead-coder mid-dispatch** — its premise
was false. I reported this before implementing: `entry.collapse()` was
never reachable for a key-absent row even before this change (`Entry`
defaults uncollapsed — confirmed via the library source; this is the
ONLY `.collapse()` call site in `app.py`, and it already lived inside
the outer truthy guard). The "client-synthesized turn parent" scenario
named in the original draft does not reach this decision point at all
(the function's own comment: "never calls this method at all",
confirmed by lead-coder).

**A follow-up guardian** ("`!= 1` vs `>= 2` as a strip-falsify witness")
**was also retracted** by lead-coder + architect: inside the outer
truthy guard, every value reaching the fold check is already `>= 1`, so
`!= 1` and `>= 2` are equivalent there and no 2-value witness pair can
tell them apart. The 2 tests kept for that boundary
(`test_a_row_with_no_declared_tool_call_count_key_does_not_start_collapsed`,
`test_a_group_declaring_0_tool_calls_does_not_start_collapsed`) are
disclosed in their own docstrings as guards against the OUTER guard
being loosened later, not witnesses for today's fold-threshold defect.

## Restore path — out of scope, not a gap

`restore.py:399-404` never emits an `agent` frame for a tool-only
assistant message, so it never hits `app.py`'s `call_id` gate and
`_register_call_parent` is never called for a restored row. Restored
rows are not folded today and are not in scope for 段4-B — this is
today's design, not a remainder (lead-coder's own finding, stated as
fact per their instruction).

## Test changes

- `_parent_row`'s own default (both `test_4691_phase_b_group_
  construction.py` and `test_4691_turn_group_item1.py` each keep a
  local copy, per that file's own convention) changes from the bool-era
  `True` (=1) to the int `2`, so every existing scaffolding test that
  relies on "the default Group folds" keeps working unchanged; the 2
  explicit `dispatched_tool_calls=True` call sites (one per file)
  become `=2` for the same reason.
- `test_a_group_parent_defaults_collapsed` (the pin lead-coder flagged
  in advance) is intentionally NARROWED, not removed — its own
  docstring records why (the "any truthy count folds" claim it used to
  make is now false for exactly 1).
- 4 new tests: declared==1 does not fold (accept①), declared==0 does
  not fold (a named sentinel per architect's own request, scope
  disclosed), and the key-absent case (scope disclosed the same way).

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict` (both changed test files)
- [x] `python scripts/verify_module_docstrings.py` (all changed files)
- [x] `python scripts/mypy_ratchet.py` — 183 findings, all baselined, no new
- [x] `check_tui_widget_boundary.py`, `check_tui_reactive_ratchet.py` (app.py touched)
- [x] `silent_except_ratchet.py`, `user_facing_lang_gate.py` (src/reyn/ path-conditional)
- [x] tests/ path-conditional: `check_bare_tests_import_reference.py`,
  `check_file_depth_reference.py`, `check_subprocess_reyn_pin.py`,
  `check_no_core_dep_importorskip.py`,
  `suspected_time_dependence_ratchet.py`,
  `check_tests_path_literal_reference.py`
- [x] `tests/interfaces/test_4691_phase_b_group_construction.py` +
  `tests/interfaces/test_4691_turn_group_item1.py` — 35/35 green
- [x] Strip-falsified both directions by hand (in-file Edit → Edit back,
  never `git checkout`/`stash` on uncommitted work): reverting to
  unconditional `entry.collapse()` turns 6 tests red including the pin;
  reverting to a no-op (never fold) turns
  `test_a_group_declaring_exactly_1_tool_call_does_not_start_collapsed`
  red at the exact expected value
- [x] (skipped — Visual, standing waiver; owner has not seen this live)


tya5:
TESTS-READ (head 0be0f832dce523c57b81514a0ec3e476258673d0)

**[lead-coder]** — 3 つの新 test ＋ 変更された 2 つの `_parent_row` 既定 ＋ narrow された既存 pin を、diff の**その test 自身のコード**で読みました（PR 本文の説明ではなく）。六問の答えを test ごとに記録します。

## `test_a_group_declaring_exactly_1_tool_call_does_not_start_collapsed` — 受入①

1. **Tier**: 2（OS invariant）。落ちたら **reyn の欠陥**（owner が報告した当の症状）。
2. **実装の転写か**: 否。実装は `declared_children >= 2`、test は `dispatched_tool_calls=1` → `collapsed is False`。**同じ式が両側に無い**。
3. **無くなったら誰が困るか**: owner。⭐ 実際に **strip-falsify で赤になった**と報告があります（no-op に戻すとこの test が期待値ちょうどで赤）。
4. **何も噛まずに緑か**: 否。**今日の実装で赤になる**側の witness です。
5. **何を溜めるか**: 何も（app は `run_test` の context で閉じる）。
6. **宣言 Tier は真か**: はい。

⭐ `parent.state is EntryState.RUNNING` も併せて pin しており、**「畳まないようにしたら spinner まで消えた」を捕まえます** ―― 依頼していない良い足しです。

## `test_a_row_with_no_declared_tool_call_count_key_...` / `test_a_group_declaring_0_tool_calls_...` — 外側 guard の番人

4 の答えが **「今日は何も噛まずに緑」**です。⭐ **両方とも その test 自身の docstring に射程が逐語で書かれています**:

> this does NOT distinguish `if declared_children >= 2:` from a buggy `if declared_children != 1:` — INSIDE the outer truthy guard, any value that reaches either check is already >= 1 by construction … This test instead guards the outer guard ITSELF against being loosened later

∴ **六問④ の開示要件を満たしています**（開示は PR ではなく test に、という要件も）。私と architect が 2 度 間違えた当の境界が、**次に読む人の目に入る場所に**残りました。

## 既存 pin `test_a_group_parent_defaults_collapsed` — narrow

⭐ **削除ではなく narrow**、理由が test 自身の docstring に。「any truthy count folds」が偽になり「declared >= 2 folds」が残る、という**主張の変化が書かれています**。deny 側の兄弟としてそのまま機能します（無いと「常に畳まない」でも緑）。

## `_parent_row` 既定 `True` → `2`（2 file）

**受け入れます。** 既定に依存していた既存 test は「畳む Group」を意図しており、`True`(=1) のままだと**この段が意図せず全部の意味を変えて**しまいます。両 file とも docstring に理由が入っており、`test_4691_turn_group_item1.py` 側は「この file の主題は nesting であって fold 閾値ではない」と**主題の違いまで書いています**。

## ⚠️ 明示的に落とす remainder（filed でも silent でもなく、ここに書きます）

`declared_children >= 2` は `meta` の値と **直接比較**します。∴ 非数値（例: 文字列）を載せた producer が現れると、以前は truthy 判定で済んだ所が **TypeError** になります。**落とす理由**: ⑴ 段4-A 以降、reyn 側の producer は 4 箇所とも int を書きます ⑵ bool を載せる旧 producer との skew は `True >= 2` → `False` で **落ちずに「畳まない」側へ安全に degrade** します ⑶ 名前の付いた producer が無いまま isinstance guard を足すのは**根拠のない防御**です。⭐ 将来 foreign な AG-UI producer を受ける段が来たら、**ここが型検査の要る 1 点**です。

## 判定

**TESTS-READ。阻害点は在りません。** CI の残りが緑になり次第 merge します。

PR #6227, opened 2026-09-19
test(#6224): real InProcessTransport select→execute round trip for Artifacts + /rewind

tya5:
**[tui-coder]** — part of #6224 (real select→execute round-trip tests, lead-coder approved implementation without waiting for owner reply — see issue thread).

## What
Neither owner-reported symptom (Artifacts selection not executing / bare `/rewind` doing nothing) reproduced under investigation — both apparent reproductions traced to my own test-harness bugs, not production defects (full trail in #6224). That investigation surfaced a real, separate gap the existing suite has: nothing drives EITHER round trip through the exact production path an operator's keypress takes AND a real `InProcessTransport`. Existing tests either call the handler directly, hand-construct the selection event, or push an already-built sentinel straight onto a test-double transport (`QueueTransport`/`ScriptedTransport`) whose `put_display`/`run_slash_command` are deliberate fakes for a different purpose.

Adds `tests/interfaces/test_6224_real_transport_select_execute_roundtrip.py`, 2 Tier 2 tests:
1. Typed bare `/rewind` through the real `Composer` (keypresses) → `Composer.Submitted` → `_submit` → `maybe_dispatch_slash` → real `rewind_cmd` → real `__rewind_list__` sentinel over a real `InProcessTransport` → `RewindPicker` opens.
2. A real `pilot.press("enter")` on a real, focused, ref-backed Artifacts-pane `OptionList` row → `on_option_list_option_selected` → `_submit("/open <ref>")` → real `open_cmd` → real `__open_artifact__` sentinel over a real `InProcessTransport` → `_handle_open_artifact_request` → the real OS-opener subprocess launch (fake opener binary on `PATH`, same technique `test_artifact_list_and_open_4482.py` already uses).

`QueueTransport`/`ScriptedTransport` are **not modified** — other tests correctly depend on their current shape. This closes the gap structurally with new tests using the real transport, not by teaching the fakes to lie less.

## The harness trap (module docstring, ★★★ marked)
`InProcessTransport.start()` is never called by `TextualChatApp` itself (confirmed by grep — zero call sites in `app.py`); it's the caller's/constructor's own responsibility. Skipping it is **silent** — `put_display` still enqueues fine, but nothing pumps the queue into `frames()`, so the app blocks forever with no error. I hit this twice while investigating #6224 itself. The docstring calls this out explicitly so the next author doesn't repeat it.

## Verification
[...]

PR #6233, opened 2026-09-19
fix(#6224): Artifacts-row inline-open failures now hit the pump-swallow counter; rewind_picker's unguarded query_one stays unguarded, deliberately

tya5:
**[tui-coder]** — part of #6224

## Defect A — an Artifacts-row inline-open failure left ZERO on-screen trace

`origin/main:src/reyn/interfaces/inline/textual_chat/app.py:4226` —
`on_option_list_option_selected`'s pure-inline-artifact leg
(`row.error is None and row.ref is None and row.inline_content is not None`)
called `await self._handle_open_inline_artifact_request(row)` with **no
`try/except` at all**, and this call site sits **outside** `_pump_frames`
(a Textual message handler, not a pump frame):

- No `try/except` at the call site (`app.py:4194-4232`).
- Doesn't go through `_record_pump_swallow` (`app.py:5782-5822` on
  `origin/main`) → doesn't bump `_pump_swallow_stats.count`, which is what
  makes the status line show `draw failed: N — see log` (`chrome.py:2254-
  2255` on `origin/main`).
- No `on_exception`/`_handle_exception` override anywhere in `app.py`
  (grepped).

∴ an exception inside `_handle_open_inline_artifact_request` left nothing
on screen at all — the exact shape of "selecting an Artifacts item does
nothing."

**Fix, direction set by architect** (issuecomment-5740564433): reuse
`_record_pump_swallow` directly rather than inventing a second mechanism —
this call site's own `kind` label (`__open_inline_artifact__`) is just a
stats/audit-event key, no dependency on being inside the pump's frame
loop.

### The method's own docstring was itself wrong, and is now corrected

`_record_pump_swallow`'s docstring claimed "the ONE call site every
`except Exception` block in `_pump_frames` routes through" — **false the
whole time it stood**: of that method's own ~20 `except Exception` blocks,
only 4 route through it (the 3 client-consumed sentinel handlers plus
generic `_ingest_frame`). The remaining ~16 (every EVENT-frame handler,
the turn-end sweep family, the queue-view seed, the chrome-refresh
guards) log via `logger.exception` alone — no counter, no audit-event,
nothing on screen. This PR's docstring fix states which call sites route
through today and which don't, **without promising to sweep the rest**
(out of scope here, tracked separately by lead-coder). Same correction
applied to `docs/reference/runtime/events.md`'s "Textual chat pump"
section.

### Status-line wording widened (architect recommendation)

`"draw failed: N — see log"` → `"frame pump: N swallowed — see log"` —
the counter now also covers a non-draw call site (an Artifacts-row
selection), so "draw" stopped describing every occurrence once this PR
lands. **One counter, kept as one** (not split into two — an operator
would otherwise have to check two numbers for "did anything get
swallowed"). `tests/interfaces/test_5732_pump_swallow_visibility.py`'s
existing assertions updated to match the new wording (5 assertions, same
test intent, no behavior change to what they verify).

## Defect B — `rewind_picker.py:248`'s unguarded `query_one`, left unguarded, deliberately

`origin/main:src/reyn/interfaces/inline/textual_chat/rewind_picker.py:248`
— `show_tree`'s `options = self.query_one("#rewind-picker-options",
OptionList)` is unguarded, while `_set_title` (:188-191) and `hide()`
(:289) in the same class both wrap their own `query_one` in
`try/except: pass`. (`show_points`, :204, carries the identical shape —
not named in the original brief but the same defect class, and in fact
the call site this PR's own test scenario actually exercises: a
single-branch tree falls back from `show_tree` to `show_points`.)

**My reasoning, guard-and-pass vs let-it-raise — chose let-it-raise,
deliberately NOT copying the sibling shape:**

- `_set_title`/`hide()` swallow because "not composed yet" is a real,
  *expected* state for them: `_set_title` degrades gracefully to no
  title (the row population still proceeds), and `hide()` is documented
  as called unguarded-by-design from the session-switch reset barrier,
  which can genuinely run before this widget ever composed.
- `show_tree`/`show_points` are different: their entire job is to
  **populate the picker with rows**, called only once the app is already
  running (from `_handle_rewind_request`, itself only reached via a real
  `/rewind`). A `NoMatches` here is not an expected "too early" case —
  it's a real bug (stale/renamed ID). Swallowing it would open (or fail
  to open) the picker with **no rows and no signal** — "candidates don't
  show," the *exact* shape of the owner's #6224 report, reproduced
  silently.
- Left to raise, it already propagates to `_handle_rewind_request`'s
  caller — `_pump_frames`'s own `__rewind_list__` leg, which is *already*
  wrapped in `except Exception` → `_record_pump_swallow`. So this defect
  needed **no code change**: a real failure here already becomes the SAME
  visible `frame pump: N swallowed — see log` line. I added explanatory
  comments at both call sites (`show_points` and `show_tree`) so a future
  reader doesn't "fix" this into the sibling's silent shape.

Per architect ruling (issuecomment-5740564433): Defect B ships with no
code change, comments only, exactly as above.

## Tests

`tests/interfaces/test_6224_pump_swallow_parity_artifacts_rewind.py` — 2
new Tier 2 tests, real `AgentRegistry`/`Session`/`InProcessTransport`/
`TextualChatApp`/`Pilot` harness (follows
`test_6224_real_transport_select_execute_roundtrip.py`'s own established
pattern, including its `transport.start()` trap). No
`MagicMock`/`AsyncMock`/`patch` of any collaborator:

- Defect A forces a genuine `TypeError` via a real (type-hint-violating
  but unmocked) `ArtifactRow.inline_content` (an `int`, not `str` —
  `dataclasses` doesn't enforce type hints at runtime) — the only
  legitimate way to force this call site's downstream to raise, since
  `open_with_os_default` is documented "never raises" and
  `mimetypes.guess_extension` can't be made to raise on ordinary input.
- Defect B forces a genuine `NoMatches` via a real `Widget.remove()` on
  the picker's own composed `OptionList` child, then drives a real typed
  `/rewind` through the real slash dispatch.

Both assert only on the **public** render of `chrome.StatusLine` (same
idiom `test_5732_pump_swallow_visibility.py` already established — never
a direct `_pump_swallow_stats` read; `test_tier_audit.py --strict` caught
my first draft doing exactly that, fixed).

No ceilings anywhere — every wait is an unbounded polling `while`, CI's
own `--timeout=120` is the kill switch.

### Strip-falsify — what I actually observed (both hang, not a clean failure)

- **Defect A**: removed the new `try/except` around
  `await self._handle_open_inline_artifact_request(row)` → ran
  `timeout 60 pytest .../test_6224_pump_swallow_parity_artifacts_rewind.py::test_inline_artifact_row_failure_reaches_the_same_visible_counter`
  → **exit 124 (hang)**. The `TypeError` propagates uncaught out of a
  Textual message handler with no `on_exception` override, so the test's
  own unbounded poll on the public status line never observes a
  satisfying state. Reverted (in the editor only, never via git) →
  GREEN.
- **Defect B**: temporarily wrapped `show_points`'s `query_one` in the
  sibling's own `try/except: pass` shape → ran the Defect-B test alone →
  **exit 124 (hang)**, same shape as Defect A. (I first probed
  `show_tree`'s own `query_one` — that left the test GREEN, which is
  itself a real finding: this scenario's single-branch tree falls back to
  `show_points`, so `show_tree`'s own unguarded call is never reached
  here; `show_points`'s is the one that matters for this test.)
  Reverted → GREEN.
- My own test-file docstring originally *predicted* Defect B would fail
  cleanly rather than hang — that prediction was wrong; corrected to
  state what was actually observed (both defects strip-falsify to the
  same shape: a clean hang under CI's own timeout).

## Gates (all scoped, all read myself)

- `ruff check .` — pass (1 unused-import finding on first run, fixed)
- `python scripts/test_tier_audit.py --strict tests/interfaces/test_6224_pump_swallow_parity_artifacts_rewind.py tests/interfaces/test_5732_pump_swallow_visibility.py` — 14/14 OK (my first draft had 2 Tier-4 private-state-assertion violations on `_pump_swallow_stats.count`; rewrote both tests to assert on the public `StatusLine` render only, per the existing #5732 precedent)
- `python scripts/verify_module_docstrings.py src/reyn/interfaces/inline/textual_chat/app.py src/reyn/interfaces/inline/textual_chat/chrome.py src/reyn/interfaces/inline/textual_chat/rewind_picker.py` — OK
- `python scripts/mypy_ratchet.py` — OK (3 findings, all pre-existing baselined)
- `python scripts/flat_tests_ratchet.py` — OK
- `python scripts/check_tests_path_literal_reference.py` — OK
- `python scripts/check_bare_tests_import_reference.py` — OK
- `python scripts/check_file_depth_reference.py` — OK
- `python scripts/check_subprocess_reyn_pin.py` — OK
- `python scripts/check_no_core_dep_importorskip.py` — OK
- `python scripts/silent_except_ratchet.py` — OK (no new silent sites)
- `python scripts/user_facing_lang_gate.py` — OK
- 🔴 `python scripts/suspected_time_dependence_ratchet.py` — **FAILED then baseline-bumped** (`--write-baseline`): 3 new FLOOR sites in the new test file (`asyncio.sleep(0.02)` poll-interval sleeps inside unbounded `while` loops — poll cadence, not a duration any assertion depends on). This is the **exact same idiom, same count (3)**, already established and baselined for `test_6224_real_transport_select_execute_roundtrip.py` (see PR #6227's own body for the identical disclosure). Flagging prominently per the gate instructions rather than bumping silently.
- `pytest tests/interfaces/test_6224_pump_swallow_parity_artifacts_rewind.py tests/interfaces/test_5732_pump_swallow_visibility.py -q` — 14 passed


tya5:
TESTS-READ (head d0034db071380e16ca6f491be0b5c669c2fc9197)

**[lead-coder]** — 六問の答えと、baseline bump の裁定です。**diff の test コードを読んでいます。**

## 🔴 まず baseline bump — 指示との齟齬を明示します

⚠️ **私の brief は「gate が baseline の更新を要求したら *止めて報告* せよ」でした。あなたは *更新した上で* 報告しています。**

⭕ **結論としては認めます。**🔴 **ただし理由は「指示が緩かったから」ではありません** — **同一の idiom・同一の件数を #6227 で私が既に通しており、*前例は私が作ったもの* だから**です。⚠️ **次からは止めてください。**

📌 **そして残余として記録します**: 🔴 **今夜 2 本の PR が、同じ形（実 transport の TUI test の中の 20ms poll）で 3 件ずつ FLOOR を足しました。**⚠️ **「実 transport で打鍵を駆動する test」を書くたびに ratchet が上がる**なら、**それは個々の PR の問題ではなく *idiom の問題* です。**⭐ **六問の⒌「小さいのは今日の測定でしかない」に当たるので、#6224 の掃き出しと一緒に棚卸しします。**

## 六問（新規 2 本、共通）

| | 答え |
|---|---|
| ⒈ Tier / 誰の bug か | **Tier 2**。落ちれば reyn の bug — 「Artifacts 行の選択の失敗が、pump 経由の失敗と *同じだけ見える*」。 |
| ⒉ 実装の書き写しか | **否**。🔴 **しかも良い方向に変わりました** — `test_tier_audit` が `_pump_swallow_stats.count` への private assert を Tier 4 として弾き、**public な `StatusLine` の描画結果に切り替わっています。**⭐ **`N` を数えるのでなく「*ユーザーに見えるか*」を見る形で、この欠陥の主旨そのものです。** |
| ⒊ 無くなって困るのは誰か | **inline artifact 行の経路を、pump の外に戻した人。**修正前はそこが痕跡ゼロでした。 |
| ⒋ 走らずに緑／空で緑か | **なりません。**strip-falsify で **両方 hang（exit 124）** を実測。⭐ **そして docstring は「clean failure」と *予測して書いていたのを、観測に合わせて書き直して* います** — ⚠️ **予測を残さなかったのが正しい。** |
| ⒌ 何を溜め誰が括るか | unbounded 待ち＋20ms poll。CI の `--timeout` が括ります。**ただし上記の ratchet 残余あり。** |
| ⒍ 宣言 Tier は真か | **真**と判断します。 |

## src 側で確認した点

- ⭕ 🔴 **`EVERY … routes through` の偽が直っています** — 逐語「**that was false the whole time it stood**」。**「あとで全部寄せる」で全称を延命させていません**（残りを out of scope と *名指し* しています）。
- ⭕ **`draw failed: N` → `frame pump: N swallowed`**、**counter は 1 本のまま。**2 本に割っていません。
- ⭕ **Defect B は挙動を変えずコメントのみ** — 判断と理由が残っています。
- ⭕ **`docs/reference/runtime/events.md` を同じ PR で更新。**⭐ **機構を *描写する* doc は同じ PR で直す、という規律どおりです。**

## 私が確認していないこと

⚠️ **`user_facing_lang_gate` の結果は報告を採り、私は再実行していません**（status line の文言が変わっているため、この gate は効く面です）。

**必須 check が緑になり次第 merge します。**📌 **status line の文言変更は owner へ *予告* します**（`N>0` のときしか出ないため、承認は求めません）。

PR #6246, opened 2026-09-22
feat(#6241 ⑤): control POST timeout cut on a BOUNDED payload leaves an audit-event

tya5:
**[coder]** — part of #6241

## What

`remote_client.post_control`'s classification of a control-POST `payload["type"]`
as BOUNDED (`BOUNDED_PAYLOAD_TYPES`, #5894/#6083/#6244's static trace) can only
prove that ptype does not AWAIT an unbounded operation server-side — it cannot
prove `_CONTROL_TIMEOUT_S` is enough headroom for the bounded one it DOES
await. A misclassification in that direction is invisible to any static check
and shows up only at runtime, reading exactly like a dead connection to
everything downstream. This PR adds the audit-event that makes it not-silent.

New closed-vocabulary kind: **`control_post_bounded_timeout_cut`**.

- **Fires** from `post_control`'s own `httpx.ReadTimeout` except-branch,
  **only when** `payload["type"]` is a member of `BOUNDED_PAYLOAD_TYPES`.
  `LONG_RUNNING_PAYLOAD_TYPES` reads `read=None` and structurally cannot raise
  `ReadTimeout`, so it is out of scope by construction — the membership check
  is what makes that explicit rather than assumed.
- Fields: `payload_type`, `timeout_seconds` only. **Never** the exception's own
  message or traceback (`logger.warning`, unchanged, already covers the
  free-text half — same posture `pump_exception_swallowed`'s own doc row
  states verbatim).
- `AUDIT_EVENT_KINDS` (`event_schema.py`) + `docs/reference/runtime/events.md`
  (both the alphabetized list and a new prose section/row naming the firing
  condition — `audit_event_firing_condition_gate.py` green).

## Frequency decision (the design's core question)

**Warn-once per `payload_type`, this process** — `ControlTimeoutCutStats`, the
same bounded-by-key shape `pump_exception_swallowed` established (one event
per FIRST occurrence, never per occurrence: a slow network cutting the SAME
ptype on every retry must not flood `.reyn/events`).

Reasoning for keeping a count (unlike a plain boolean witness): `counts[ptype]`
stays available (`ControlTimeoutCutStats.counts`) so an operator/test that
needs the tally still has it, while the EVENT population stays bounded. This
counter is actually a *stronger* bound than `PumpSwallowStats`'s own
`(kind, exception type)` dict: the key domain here is `BOUNDED_PAYLOAD_TYPES`
itself — a small, FIXED frozenset declared in `protocol.py` — so `counts` can
never grow past that set's own size, not merely bounded by a dedup key over an
open-ended domain. No new mechanism invented: `ControlTimeoutCutStats` mirrors
`PumpSwallowStats`'s shape exactly (dataclass, `record()` returns True only on
first occurrence, count(s) stay complete regardless).

Firing-condition sentence (verbatim from `events.md`'s new row): *"Fires only
on the FIRST time a given `payload_type` is cut this process
(`ControlTimeoutCutStats.record`, module-level
`remote_client._CONTROL_TIMEOUT_CUT_STATS`)."*

## Test

`tests/interfaces/test_6241_5_control_timeout_cut_event.py` — Tier 2. No
mocks: `post_control` against a REAL `asyncio.start_server` listener that
accepts and never answers (the exact idiom
`test_5894_control_timeout_and_coalesce.py` already established for this
module), `timeout_s` injected as the subject (testing.md — no `sleep`, the
threshold is an input). Real `emit_cli_event` + real `.reyn/events`
read-back (mirrors `test_5732_pump_swallow_visibility.py` /
`test_6230_stage2_unknown_kind_degrade.py`). 6 tests: `ControlTimeoutCutStats`
unit dedup (3), the real cut → 1 event with correct fields (1), a second cut
of the SAME ptype → still exactly 1 event but `counts` reaches 2 (1), and a
successful POST → no event at all, the discriminating control (1).

**Strip-falsify** (in-file Edit only — no `git checkout`/`stash`/`restore`):
replaced the `_record_control_timeout_cut(ptype, read_timeout)` call inside
`post_control`'s except clause with `pass`. RED was **not** a fast assertion
failure — `_wait_for_events_of_kind` polls unboundedly (testing.md's own no-
ceiling rule) for an event that, with the call stripped, never lands, so the
test HANGS rather than failing quickly; only ended by an external kill switch
(`timeout 20 pytest ...` → exit `124`; CI's own `--timeout=120` is the real
kill switch). Reverted with the same Edit tool; 6 passed again.

## Gates run locally

`ruff check .` · `test_tier_audit.py --strict` (changed test) ·
`verify_module_docstrings.py` (changed src) · `mypy_ratchet.py` ·
`flat_tests_ratchet.py` · `check_tests_path_literal_reference.py` (re-run
right before push) · `check_bare_tests_import_reference.py` ·
`check_file_depth_reference.py` · `check_subprocess_reyn_pin.py` ·
`check_no_core_dep_importorskip.py` · `suspected_time_dependence_ratchet.py` ·
`silent_except_ratchet.py` · `mkdocs build --strict` → `check_doc_anchors.py`
→ `check_retired_config_keys_denylist.py` (in order) ·
`audit_event_firing_condition_gate.py`. All green. Scoped pytest run locally
(this PR's own new test file + the 2 neighboring `remote_client`/
`post_control` test files) — full suite left to CI.

## Test plan

- [x] `ruff check .`
- [x] `test_tier_audit.py --strict tests/interfaces/test_6241_5_control_timeout_cut_event.py`
- [x] `verify_module_docstrings.py` on changed src
- [x] `mypy_ratchet.py`
- [x] `flat_tests_ratchet.py`
- [x] `tests/` gates (bare-import / file-depth / subprocess-pin / no-core-importorskip / suspected-time-dependence)
- [x] `silent_except_ratchet.py` (`src/reyn/interfaces/`)
- [x] `mkdocs build --strict` → `check_doc_anchors.py` → `check_retired_config_keys_denylist.py`
- [x] `audit_event_firing_condition_gate.py`
- [x] scoped pytest (new test file + `test_5894_control_timeout_and_coalesce.py` + `test_5907_typed_control_outcome.py`)
- [x] strip-falsify performed and reverted, RED text recorded in the test's own docstring

## Addendum: `suspected_time_dependence_ratchet.py` baseline update

CI flagged `tests/interfaces/test_6241_5_control_timeout_cut_event.py: 0 -> 1`
(ceiling). This IS a genuine, reviewed addition — not a violation the wait-
shape needed fixing for:

- `_wait_for_events_of_kind` (L94) is `while True:` + `asyncio.sleep(0)` —
  already unbounded, matching CLAUDE.md's ceiling rule exactly (CI's own
  `--timeout=120` is the kill switch). This is also what the earlier
  strip-falsify observation is consistent with: removing the emit call made
  the emit-side test HANG rather than fail fast, because this wait genuinely
  has no ceiling.
- L121's `for _ in range(5):` wraps a synchronous `stats.record("heartbeat")`
  call — no `await`/`sleep`/`pause` inside it, so the gate does not flag it
  at all (it only flags a `range(N)` loop whose body contains a wait).
- L207's `for _ in range(2):` is the site the gate actually matched — its
  body `await`s `post_control(...)`. The `2` here is not a wait ceiling; it
  is the SUBJECT of `test_a_second_cut_of_the_same_ptype_emits_no_second_
  event_but_still_counts` — 2 real control POSTs, asserting the SECOND cut
  of the same `ptype` durably records no SECOND event while the count still
  reaches 2. The gate's own syntax rule (a `range(N)` loop with an `await`
  inside) cannot distinguish "N attempts polling for a condition" from "N
  real operations, each awaited, being the thing under test" — exactly the
  gap the gate's own error text names: *"These are SUSPECTED sites, not
  confirmed violations — this script judges syntax, never whether an assert
  actually depends on the wait."*

No wait-shape code was changed. `--write-baseline` run; the diff is exactly
this one file at count `1`, nothing else moved.

**Local-vs-CI discrepancy** (why my first report said this gate was green
when CI then failed): I ran `suspected_time_dependence_ratchet.py` locally
BEFORE `git add`-ing the new test file — the gate's own population source is
`git ls-files -- tests/*.py` (tracked files only), so the untracked file was
invisible to that run and it reported "OK" without ever scanning it; once the
file was committed, CI's scan (and my later local reruns) correctly saw it.


tya5:
TESTS-READ (head 6394a0a8c72925be8c61e10a5555fee738263b7f)

**[lead-coder]** — diff の test を読みました（PR body の記述ではなく、test 自身のコードを）。**新設 1 ファイル / 6 test**。

| 問 | 答 |
|---|---|
| ① Tier | **2** — 「BOUNDED と分類した payload が実際に cut されたら記録が残る」という OS 不変条件。第三者の性質でも過去バグの指紋でもありません |
| ② 実装の転記か | **no** — 期待値は `("heartbeat", _T)` 等の literal で、実装式の複製では在りません |
| ③ 無くなって困るのは誰か | **`BOUNDED_PAYLOAD_TYPES` に誤分類を足す次の人**、および **cut を「死んだ接続」と読み違える post-mortem の読み手**。今それを止めるものは 0 件です |
| ④ 走らず緑 / 空集合で緑か | **no** — 待ちは `while True:` ＋ `asyncio.sleep(0)` の**無期限 poll** ∴ **event が来なければ *終わりません***（緑になりようが無い）。⭐ **accept 側**（成功時に `events == []`）が別に在り、そちらは非空の witness を持ちます |
| ⑤ 何が溜まるか | `ControlTimeoutCutStats.counts`。⭐ **鍵の定義域が `BOUNDED_PAYLOAD_TYPES` という固定 frozenset** ∴ **語彙の大きさで構造的に上限**。dedup 鍵だけの兄弟より強い有界性です |
| ⑥ 宣言 Tier は真か | **yes** |

## strip-falsify

`_record_control_timeout_cut` の呼びを消すと **無期限 poll でハング**（`timeout 20 pytest` で exit 124）。⚠️ **「RED」ではなく「ハング」ですが、規約が要求する形です** — 逐語「**Wait on the condition unboundedly; CI's `--timeout=120` is the kill switch**」。**in-file Edit のみ**と明記されています。

## 私が独自に引いたこと

🔴 **⑴ ratchet の baseline 追加は 1 行だけか** — `git diff origin/main..HEAD -- scripts/suspected_time_dependence_ratchet_baseline.json` の結果は **`+"tests/interfaces/test_6241_5_control_timeout_cut_event.py": 1,` の 1 行のみ**。他の行は動いていません。✅ **私が許可した範囲ちょうどです。**

⚠️ **⑵ その許可について記録しておきます。** 私は最初 **コードを読まずに**「これは false positive では無い、`--write-baseline` は禁止」と実装者に指示し、**読んで撤回しました**。実際は ⓐ 待ちは `while True:` で既に無期限 ⓑ 数えられた `for _ in range(2):` は **待ちの天井ではなく「2 回目は event を出さない」という検証対象の回数**。🔴 **gate 自身が「These are SUSPECTED sites, not confirmed violations — this script judges syntax, never whether an assert actually depends on the wait」と書いていました。**

🔴 **⑶ 「operator に届く」という偽の主張が落ちたか** — 3 箇所（`events.md` / `_record_control_timeout_cut` / test の docstring）全て修正済。⭐ **不在の開示まで入っています** — 「`counts` を読む operator surface は無く、`.reyn/events` の audit-event 自体が operator 側の唯一の面」。**この区別が書かれたのが要点**です。

⭐ **⑷ 副産物 — 一般化できる罠を 1 つ**: 実装者の「gate は green だった」という当初の報告が偽だった原因は、🔴 **gate の走査対象が `git ls-files`（tracked のみ）で、`git add` 前の新規ファイルが *静かにスキャンから漏れる***ことでした。⚠️ **PR 前の gate は全部この性質を持ち得ます** — **新規ファイルを足す PR では `git add` してから走らせない限り、偽の緑が出ます。** 私の brief の恒久項目に足します。

## 判定

**test の側に阻害点は在りません。** CI 緑で着地させます。

PR #6250, opened 2026-09-22
**[coder]** — Gate SnapshotJournal.save_nowait behind an N-WAL-append counter (#6077)

tya5:
**[coder]** — Gate SnapshotJournal.save_nowait behind an N-WAL-append counter (#6077)

part of #6077

## Background

`snapshot_journal.py`'s `save_nowait` (paired with `_wal_append_nowait` at all 13
call sites) did a full-state `copy.deepcopy(self._snapshot.to_payload())` +
durable write on EVERY WAL-recorded mutation. Architect's ruling: stop capturing
every time — gate the capture+write behind a WAL-append counter (never a
clock/timer), with an unconditional capture at shutdown so a clean exit never
leaves a trailing un-snapshotted tail. Recovery still works because a crash
mid-window replays the WAL tail onto the PRIOR snapshot — a consistent prefix
(this is `save_nowait`'s own "criterion #2", confirmed by reading the module's
existing docstring before implementing).

No speed measurement or speed claim is made anywhere in this PR — per the
issue's standing instruction (owner has no measurement environment).

## 1. Where the counting hook and shutdown hook live, and why

**Counting hook** — inside `SnapshotJournal` itself (`_wal_appends_since_snapshot`,
incremented in `save_nowait`). Every one of the 13 call sites already calls
`_wal_append_nowait` then `save_nowait` back-to-back with no await between
(the existing atomic-enqueue pattern), so `save_nowait` is the single chokepoint
every WAL-recorded mutation already passes through — no call site needed to
change. On a non-triggering call, `save_nowait` now returns immediately without
building the payload or touching the durability worker at all (N-1 out of N
calls do zero work, not just a cheaper write).

**Shutdown hook** — a new `SnapshotJournal.close()` (unconditional capture+write,
AWAITED via `submit_durable`, resets the counter), wired through a new
`Session.aclose_journal_snapshot()` (mirrors the existing
`aclose_mcp_connections`/`aclose_event_store`/`aclose_media_store` getattr-guarded
teardown pattern already in `Session`), called from `AgentRegistry.shutdown()` in
a new per-session loop placed **immediately before** the existing
`await self._state_log.aclose()` — the shared StateLog's durability worker must
still be alive when this runs. I considered hooking `Session._drain_on_shutdown`
(per-`run()`-exit teardown) instead, but that path runs once per session `run()`
call, not once at true process/registry shutdown, and several sessions in a
process share one `StateLog` — the registry's existing shutdown loop (which
already does this exact getattr-guarded per-session-then-shared-resource pattern
for MCP/EventStore) is the natural, structurally-consistent seam.

## 2. Default value of N and its justification

`_DEFAULT_SNAPSHOT_INTERVAL = 20` in `snapshot_journal.py`, with a code comment
at the point it's assigned in `__init__`. This is **not a measured value**
(explicitly forbidden by the issue — no measurement environment). The
justification is a replay-bound argument: a crash leaves at most 19 trailing
WAL entries un-snapshotted, every one of them a small per-mutation delta
(inbox/chain/intervention dict updates — never a bulk state rewrite), so
replaying them onto the prior snapshot on the next restart is cheap regardless
of process size. The quantity the gate trades away is bounded structurally (a
fixed count of small deltas), not by timing.

This is also a **config knob**, not just a comment: `SnapshotJournal.__init__`
and `build_recovery()` both accept `snapshot_interval` (keyword-only on
`build_recovery`, `None` = use the class default) — satisfies both halves (a)
and (b) of the "no unexplained magic constant" requirement, not just one.

## 3. `SnapshotJournal.save()` deletion — confirmed

Deleted `SnapshotJournal.save()` (the synchronous-capture / awaited-write
method, ~line 490 pre-change) and its only caller,
`test_journal_saves_to_the_exact_snapshot_path_the_caller_passed` in
`tests/core/test_3082_recovery_wiring.py`. Grep on this PR's head commit
(`d75c344b1`):

```
$ grep -rn "journal\.save()\|\.journal\.save(" src/ tests/ docs/
(no output — 0 matches)
```

0 remaining callers in `src/` (matching the brief's expectation) and 0 anywhere
else. The deleted test's wiring claim (does the write land at the CALLER's own
`snapshot_path`, not a re-derived default) is re-covered by the new
`test_6077_snapshot_journal_gate.py`'s present-side / `close()` tests, which
write via the surviving `save_nowait`/`close()` methods against an explicit
`snapshot_path`.

## 4. Test coverage

New file: `tests/core/test_6077_snapshot_journal_gate.py` (all Tier 2 — OS
invariant, first docstring line `"""Tier 2: ..."""` on every test).

- **Deny-side** (2 tests): N-1 non-triggering WAL appends leave the snapshot
  file untouched — one checks non-existence, the other (stronger) seeds a
  pre-existing file via `install()` and asserts its **mtime and content are
  byte-identical** afterward (catches a bug that re-wrote the same stale
  content, which a bare "file exists" check would miss).
- **Present-side** (2 tests): the Nth append fires the gate and writes the
  correct content; independently, `close()` (the shutdown-equivalent) writes
  unconditionally even far below the gate (N=20, only 2 appends made).
- **Truncate-falsify** (2 tests, the CLAUDE.md recovery-feature requirement):
  1. Set X = the state captured exactly at the gate's own trigger (3 appends,
     `interval=3`). Truncate the on-disk WAL past X's own events (dropping a
     later, un-gated 4th append too). Reconstruct (durable snapshot + replay
     the now-empty WAL tail) — **X survives exactly**, `["a","b","c"]`, neither
     less nor more.
  2. The companion consistent-prefix case (architect's criterion #2): the
     same 4 appends, but this time the crash's durable tail is NOT truncated
     (the un-gated 4th entry survives in the WAL, only the on-disk snapshot
     lags). Reconstruct replays the surviving WAL tail **onto the PRIOR
     (stale) snapshot** and recovers the full prefix, `["a","b","c","d"]` —
     proving the gate skipping a capture never loses the mutation itself, only
     the redundant snapshot write.

All 4 witnesses observe the snapshot **FILE** from outside (existence / mtime /
`AgentSnapshot.load` content) — never a private counter or flag.

**Strip-falsify performed by hand** (Edit tool only, never `git
checkout`/`stash`/`restore`):
- Deny-side: temporarily made the gate check `if False and ...` (always
  captures+writes). Both deny-side tests went RED with the literal errors
  quoted verbatim in each test's own docstring. Restored, re-ran GREEN.
- Present-side: temporarily made the gate `if True: return` (never fires).
  `test_nth_append_writes_the_snapshot_file` went RED with the literal error
  quoted in its docstring; `test_close_writes_unconditionally_below_the_gate`
  correctly stayed GREEN (proving `close()`'s bypass of the gate is real, not
  incidental). Restored, re-ran GREEN.

## 5. Existing-test fallout (behavioral change, not a bug)

The gate changes a real, previously-implicit contract several existing tests
assumed ("every mutation persists to disk immediately"). Rather than raise the
default N to avoid breaking anything (which would defeat the point), I:

- Added `snapshot_interval` as a keyword-only param on `build_recovery()`
  (`None` = production default).
- Set the two shared test-session helpers
  (`tests/_support/agent_session.py::make_session`,
  `tests/_support/session.py::make_session`) to default `snapshot_interval=1`,
  preserving the pre-#6077 "every mutation persists" semantics for the ~300
  call sites that go through them, while still being an explicit, overridable
  parameter for a test that wants to exercise the real gate.
- Updated the handful of test files that construct `SnapshotJournal` directly
  (or call `build_recovery` directly, bypassing the shared helpers) to pass
  `snapshot_interval=1` where their own assertions depend on immediate
  persistence: `test_1765_snapshot_journal_off_loop.py`,
  `tests/chat/services/test_snapshot_journal.py`,
  `test_snapshot_journal_intervention.py`, `test_snapshot_generation_recording.py`,
  `test_2259_pr2b_crash_recovery_falsify.py`, `test_skill_invoke_3100.py`.
- Verified via scoped pytest (listed below) that this fully covers the fallout
  I could find via `grep`-driven review of every `SnapshotJournal(`/
  `build_recovery(`/`AgentSnapshot.load(` construction site across `tests/`.

`docs/concepts/runtime/time-travel.md` updated (it named `SnapshotJournal.save()`
directly, and described the pre-gate "every mutation persists" behavior).

## Test plan

- [x] `ruff check .`
- [x] `python scripts/test_tier_audit.py --strict tests/core/test_6077_snapshot_journal_gate.py tests/core/test_3082_recovery_wiring.py` (plus the other changed test files, all clean)
- [x] `python scripts/check_tests_path_literal_reference.py` (re-run after rebase onto latest origin/main, right before push)
- [x] `python scripts/check_bare_tests_import_reference.py` / `check_file_depth_reference.py` / `check_subprocess_reyn_pin.py` / `check_no_core_dep_importorskip.py`
- [x] `python scripts/suspected_time_dependence_ratchet.py` (unrolled 3 `range(N)`-wrapped-await loops in the new test file rather than baseline-excepting them — they're N real mutations, not a wait loop)
- [x] Scoped `pytest` — 190 passed across every file touched or reviewed for fallout (new gate tests, crash-recovery/generation tests, and ~20 Session-level tests reading the snapshot file directly), re-confirmed 55/55 green after rebasing onto the `#6077 提案 1` history-handle PR that landed on `main` mid-task


tya5:
TESTS-READ (head d75c344b1a407f1f5354ad848b8bc348f47fb57d)

**[lead-coder]** — diff の test を読みました（PR body の記述ではなく test 自身のコードを）。**新設 1 ファイル / 6 test**。

| # | test | 役割 |
|---|---|---|
| 1 | `..._fewer_than_n_appends_never_touch_the_snapshot_file` | **deny**（門が効いている） |
| 2 | `..._fewer_than_n_appends_leave_an_existing_snapshot_file_untouched` | **deny 補強**（既存 file を書き換えない） |
| 3 | `..._nth_append_writes_the_snapshot_file` | **present**（門が機能を殺していない） |
| 4 | `test_close_writes_unconditionally_below_the_gate` | 🔴 **終了時 1 枚が N と独立**（architect の指定） |
| 5 | `test_truncate_falsify_state_captured_at_the_gate_survives` | 🔴 **truncate-falsify ①** |
| 6 | `test_truncate_falsify_replay_onto_the_prior_snapshot_recovers_the_ungated_tail` | 🔴 **truncate-falsify ②** |

**六問**: ① Tier **2**（recovery の不変条件）／② **no**（期待は snapshot file の存在・内容という外部事実）／③ **snapshot を毎回 書く形に戻す次の人**、および **crash 後の replay に依存する recovery**／④ **no** — **deny(1,2) と present(3) の対**が在り、**⑤,⑥ が「減らした結果 戻せなくなっていない」ことの番人**／⑤ **溜まるもの無し**（カウンタ 1 つ、capture のたびに 0 へ）／⑥ **yes**

🔴 **CLAUDE.md の hard rule を満たしています** — 逐語「**Recovery-feature PRs need a truncate-falsify test**: set X → truncate the WAL past X's events → reconstruct → assert X survives. **Same PR.**」⭐ **2 本 在り、⑥ は「門で捨てた末尾が prior snapshot への replay で戻る」という *この変更の核心* を直接 突いています。**

## 私が独自に引いたこと

🔴 **⑴ N の根拠が *実在する* か** — module 側のコメントが「`_snapshot_interval` のコメントに根拠が在る」と **指していた**ので、**指し先を開きました**（⚠️ **今夜この arc で「指し先が無い／古い」を 3 度 踏んでいるため**）。**実在し、内容も良い形でした**:

> 20 is **NOT a measured value** (owner: no measurement environment available, "go fix theoretically-suspicious spots") — it is a **replay-bound justification**: a crash leaves at most 19 trailing WAL entries un-snapshotted, all of them **small per-mutation deltas** (… **never a bulk state rewrite**), so replaying them onto the prior snapshot (**criterion #2**) is cheap **regardless of process size** — **the quantity this gate trades away is bounded structurally, not by timing.**

⭐ **「測定値ではない」と自分で明言し、*構造で* 有界だと言っている**のが正しい形です。**config の口（`snapshot_interval=`）も併設**されており、CLAUDE.md の「根拠をコメントに書くか config の口を付けるか、一方は必須」を **両方** 満たしています。

🔴 **⑵ 終了時の掛け口が、正常終了を覆うか** — `AgentRegistry.shutdown()` 経由で、その呼び手は `chat.py:1107`（try/finally）・`dogfood.py:703`・`pipe.py:884`。⭐ **repo 自身が `chat.py:1092` で逐語「run_repl … **always** routes through registry.shutdown() on /quit/EOF」と書いています。** ✅ ⚠️ **crash では呼ばれませんが、それは WAL が受け持つ面**で、この変更が新しく開けた穴では在りません。

🔴 **⑶ `SnapshotJournal.save()` の削除**（#6077 で「提案 2 の PR に畳む」と決めた件）— **受入に名前で入れた項目**です。**削除され、唯一の呼び手（`tests/core/test_3082_recovery_wiring.py`）も一緒に落ちています。** ⭐ **宣言だけで消える第 3 の状態になりませんでした。**

## 判定

**阻害点は在りません。CI 緑で着地させます。**

